PR #1, opened 2016-07-12
Add analytics code

kwalrath:
@tbeck & @filiph is this a good way to add the analytics code?

1. Will it make its way into every page?
2. Is it optimal?

It's the same code that we use on the existing www.dartlang.org, but included in a different way since this site is implemented differently.


kwalrath:
I'm going to merge this now, but feel free to improve it.

PR #18, opened 2016-07-20
Fix the dartino codelab link

mit-mit:
Was linking to the source file, would prefer to link to the actual website


kwalrath:
LGTM. Thanks!

(No CLA needed for this Googler; we should get this straightened out.)

PR #96, opened 2016-07-25
Linux download instructions differ by processor

whesse:
The debian package repo is only available for 64-bit intel.  There is an issue filed to make it available for arm as well.

### DIFF
--- a/src/install/linux.md
+++ b/src/install/linux.md
@@ -9,14 +9,14 @@ js:
   defer: true
 ---
 
-If you're using Debian/Ubuntu, you can choose two options,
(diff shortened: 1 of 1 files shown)


src/install/linux.md
```diff
@@ -9,14 +9,14 @@ js:
   defer: true
 ---
 
-If you're using Debian/Ubuntu, you can choose two options,
+If you're using Debian/Ubuntu on amd64 (64-bit Intel), you can choose two options,
```
kwalrath: From a brief web search (e.g. http://www.x86-64.org/about.html), it looks like amd64 should be AMD64.

kwalrath:
LGTM, although it'd be nice to fix the capitalization.

PR #122, opened 2016-08-30
Update native-extensions.md

zanderso:
This update is to match the changes here: https://codereview.chromium.org/2285223003/


kwalrath:
Thanks for the PR, @zanderso! When should this go live? (What release is it valid for?)

zanderso:
It will be in the 1.20 stable release after https://codereview.chromium.org/2285223003/ lands. I'm not sure how things are usually updated, but I would guess that it's alright if it doesn't go live until 1.20 stable is out.

kwalrath:
We could publish it when that CL lands, and just add "as of 1.20" somewhere in the description.

zanderso:
The CL has landed, and I've updated the text to say "as of 1.20".

PR #123, opened 2016-08-30
Adding a page on static analysis options.

Sfshaza:
There are a couple outstanding issues/questions, but please review.

@srawlins  @pq  @bwilkerson  @kwalrath 

Staged here:  https://sz-www-1.firebaseapp.com/guides/language/analysis-options


bwilkerson:
> Dart tools, such as the Dart Dev Compiler (DDC), dartanalyzer, flutter analyze, and the IntelliJ IDEs, all use the analysis server to evaluate your code.

That (mostly) isn't true. IntelliJ uses server, but server and all of the other tools listed there use the 'analyzer' package, and it's the 'analyzer' package that has support for the analysis options file.

> Note that some of the linter rules don’t play well with strong mode.

Do we know which ones? We should fix or remove these lints.

> To enable a linter rule, add linter: to the analysis options file, followed by rule:.

"rules:"

> In future, related lint rules may be coalesced into meta rules. See Issue 99: Meta linter rules for more information.

It seems strange to me to discuss a possible future enhancement in this kind of documentation.

> Or perhaps you rely on code generated from a package that you don’t own—it runs, but generates errors during static analysis.

I don't understand this example. If the code is outside the directory that was opened / analyzed we won't show errors in that code by default. Users don't need to exclude such files.

> TBD: Something about the language options…

The language options are:
- enableAssertInitializer - enables having an 'assert' in the initializer list for a constructor; defaults to `false`
- enableAsync - enables the 'async' support (used by Fletch to disable 'async' support); defaults to `true`, but shouldn't be included
- enableGenericMethods - enables generic method support (the real support, not the ddc comment syntax which is enabled for strong mode); defaults to `false`
- enableStrictCallChecks - enables strictly following the specification when generating warnings on "call" methods (fixes dartbug.com/21938); defaults to `false`
- enableSuperMixins - enables support to allow mixins to inherit from types other than Object, and to reference `super`; defaults to `false`

We will add additional flags for other language features as they are implemented and remove them when they are enabled by default. I'm not sure we want to document temporary flags like these.

Sfshaza:
Thanks for your review, @bwilkerson!

The following lint rules caused problems when updating some of our samples and, on @jmesserly's recommendation, I commented them out:

#    - always_declare_return_types
#    - always_specify_types
#    - avoid_as

Regarding file exclusion: Maybe I could have expressed it better, but I was thinking of the two situations where I had to exclude files when I converted our examples:

1. In the dart-up-and-running repo, we had some examples that were intentionally broken.
2. The "Nest of Pirates" sample uses the RPC package to generate the client API.  You include the generated API with your client code. Gustav wrote this package over a year ago and included it in a server codelab that I wrote. I'm not sure if anyone is still supporting this package. The generated code causes a number of errors/warnings, so I excluded it. That's the sort of thing I was referring to.

We do sometimes link to issues in our docs so that people can track bugs of interest. Since you seem to feel it's premature to mention the meta linter rules, I commented that out.

Thanks for clarification on the language: field. I am not covering it, though I am considering adding a single sentence stating that it's for experimental features and not to worry about it. I can't be the only person to notice it and wonder.  :)

Thanks! Let me know if this looks better! I hadn't realized the relationship between the analysis server and the analyzer package, so I hope I got it right this time.

bwilkerson:
lgtm

> The following lint rules caused problems ...

What kind of problems? We should probably open issues for them.

> Regarding file exclusion: ...

Those are both reasonable uses of the feature. I don't expect many users have code with intentional errors in it, so I might drop that. But it's definitely worth mentioning generated code as something to be excluded.

> ... adding a single sentence stating that it's for experimental ...

That seems reasonable.

pq:
Looks amazing!  Hats off @Sfshaza !

>>The following lint rules caused problems ...

> What kind of problems? We should probably open issues for them.

Yes.  Absolutely!

Needless to say

image

Sfshaza:
Thanks, @bwilkerson and @pq for your review!

@kwalrath, I'm ready for your review now!

src/_guides/language/analysis-options.md
```diff
@@ -0,0 +1,268 @@
...
+You can also configure the linter, one of the analyzer's plugins,
+to ensure that your code complies with the
+[Dart Style Guide](/guides/language/effective-dart/style)
+and other suggested guidelines
+in [Effective Dart](/guides/language/effective-dart).
+Dart tools such as the
+[Dart Dev Compiler (DDC)](https://github.com/dart-lang/dev_compiler),
+[`dartanalyzer`](https://github.com/dart-lang/sdk/tree/master/pkg/analyzer_cli#dartanalyzer),
+[`flutter analyze`](https://flutter.io/debugging/#the-dart-analyzer),
+and the [IntelliJ IDEs](https://www.dartlang.org/tools/jetbrains-plugin)
```
kwalrath: IntelliJ -> JetBrains
(delete "the"?)

src/_guides/language/analysis-options.md
```diff
@@ -0,0 +1,268 @@
...
+use the analyzer package to evaluate your code.
+
+This document explains how to customize the behavior of the analyzer
+using an `.analysis_options` file. If you want to
```
kwalrath: I'd pronounce this "dot analysis options", meaning an -> a.

But maybe just call it "analysis options" (no code font)?

src/_guides/language/analysis-options.md
```diff
@@ -0,0 +1,268 @@
...
+add static analysis to your tool, see the
+[analyzer package](https://pub.dartlang.org/packages/analyzer) docs and the
+[Analysis Server API Specification](https://htmlpreview.github.io/?https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/doc/api.html).
+
+
+## The analysis options file
+
+The `.analysis_options` (or `.analysis_options.yaml`) file,
```
kwalrath: You can give it either name? If true, say so.

src/_guides/language/analysis-options.md
```diff
@@ -0,0 +1,268 @@
...
+written in the YAML language, is typically placed in the same
```
kwalrath: "is... placed" is passive voice. Try to make it active.

src/_guides/language/analysis-options.md
```diff
@@ -0,0 +1,268 @@
...
+directory as the pubspec file at the top of the package.
+If no file is present, the analyzer walks up the directory
+tree, looking for one.
```
kwalrath: So the analyzer can walk out of your package? How far down in the package can the file go? (Where does the analyzer start looking for the file? Can you have multiple analysis options files in a single package? If so, which one takes precedence?)

src/_guides/language/analysis-options.md
```diff
@@ -0,0 +1,268 @@
...
+
+Here's a sample YAML analysis_options file:
```
kwalrath: You say "YAML" a lot. :)

analysis_options -> analysis options here?

src/_guides/language/analysis-options.md
kwalrath: I think you could delete "written in the YAML language" here. Just cover it once, maybe in the paragraph about YAML being sensitive to whitespace.

src/_guides/language/analysis-options.md
```diff
@@ -0,0 +1,268 @@
...
+    - avoid_empty_else
+    - cancel_subscriptions
+    - close_sinks
+{% endprettify %}
+
+YAML is sensitive to whitespace&mdash;don't use tabs in a YAML file,
+and use 2 spaces to denote each level of indentation.
+
+<div class="alert alert-info" markdown="1">
+Note: You might come across a `language:` tag in an analysis options file.
```
kwalrath: Note: -> **Note:**

kwalrath: Didn't we used to use `<aside>` instead of `<div>` for this kind of thing? (Maybe we changed...)

src/_guides/language/analysis-options.md
```diff
@@ -0,0 +1,268 @@
...
+`strong-mode: true` in the analysis-options file:
+
+{% prettify yaml %}
+analyzer:
+  strong-mode: true
+{% endprettify %}
+
+This flag defaults to false. Instead of specifying `true`
+you can use the following flags to look for specific types
+of implicit casting, on top of the standard strong mode checks.
```
kwalrath: Later you say that both flags default to true, which makes it sound like you're not *adding* to strong mode checks, but *removing* standard strong mode checks.

Or maybe the "This flag defaults to true" statements should be something like "This flag, when present, defaults to true"?

src/_guides/language/analysis-options.md
```diff
@@ -0,0 +1,268 @@
...
+analyzer:
+  strong-mode:
+    - no-implicit-downcasts: false
+    - no-implicit-dynamic: false
+{% endprettify %}
+
+## Enabling linter rules
+
+The analyzer package also provides a code linter.
+A [wide variety](http://dart-lang.github.io/linter/lints/)
```
kwalrath: Expand the link to include "linter rules", since "wide variety" by itself doesn't describe the link target.

src/_guides/language/analysis-options.md
```diff
@@ -0,0 +1,268 @@
...
+of linter rules are available. Linters tend to be
+non-denominational&mdash;rules don't have to agree with each other.
+For example, some rules are more appropriate for library packages
+and others are designed for Flutter apps.
+Note that some of the linter rules don't play well with strong mode.
+Linter rules can have a false positive, unlike a static analyzer.
```
kwalrath: There seems to be a singular-plural mismatch here. How about:

Linter rules can have false positives, unlike static analysis.

src/_guides/language/analysis-options.md
```diff
@@ -0,0 +1,268 @@
...
+    - no-implicit-downcasts: false
+    - no-implicit-dynamic: false
+{% endprettify %}
+
+## Enabling linter rules
+
+The analyzer package also provides a code linter.
+A [wide variety](http://dart-lang.github.io/linter/lints/)
+of linter rules are available. Linters tend to be
+non-denominational&mdash;rules don't have to agree with each other.
```
kwalrath: non- -> non
(one word, no hyphen)

src/_guides/language/analysis-options.md
kwalrath: We could flow a little better from the previous sentence to this one. An exercise for you. :)

src/_guides/language/analysis-options.md
```diff
@@ -0,0 +1,268 @@
...
+For example, some rules are more appropriate for library packages
+and others are designed for Flutter apps.
+Note that some of the linter rules don't play well with strong mode.
+Linter rules can have a false positive, unlike a static analyzer.
+
+To enable a linter rule, add `linter:` to the analysis options file,
+followed by `rules:`.
+On subsequent lines, specify the rules that you want to apply,
+prefixed with a dash. For example:
```
kwalrath: a dash -> dashes

src/_guides/language/analysis-options.md
```diff
@@ -0,0 +1,268 @@
...
+
+{% comment %}
+Join the discussion list for linter enthusiasts:
+
+* [linter-discuss (TBD)](xxx)  Doesn't exist yet...
+{% endcomment %}
+
+Use the following resources to learn more about static analysis in Dart:
+
+* [Strong Mode](https://github.com/dart-lang/dev_compiler/blob/master/STRONG_MODE.md#strong-mode)
```
kwalrath: Mode -> mode

src/_guides/language/analysis-options.md
```diff
@@ -0,0 +1,268 @@
...
+
```
kwalrath: Delete the blank line (here and after every other item in this list)

src/_guides/language/analysis-options.md
```diff
@@ -0,0 +1,268 @@
...
+* [Dart Linter](https://github.com/dart-lang/linter#linter-for-dart)
```
kwalrath: Linter -> linter

src/_guides/language/analysis-options.md
```diff
@@ -0,0 +1,268 @@
...
+
+* [Linter Rules](http://dart-lang.github.io/linter/lints/)
```
kwalrath: Linter Rules -> Dart linter rules
?

kwalrath:
I had a bunch of picky little comments, because... you know me. But I think this is a great, well-written doc.

Sfshaza:
I've incorporated your edits, @kwalrath, thanks! I've added a graphic showing a sample directory structure with two analysis options files. I also ran the formatter on some files, but that's not related to this issue.

pq:
LGTM! 👍

src/_guides/language/analysis-options.md
```diff
@@ -121,19 +132,19 @@ analyzer:
...
 ## Enabling linter rules
 
-The analyzer package also provides a code linter.
-A [wide variety](http://dart-lang.github.io/linter/lints/)
-of linter rules are available. Linters tend to be
-non-denominational&mdash;rules don't have to agree with each other.
+The analyzer package also provides a code linter. A wide variety of
+[linter rules](http://dart-lang.github.io/linter/lints/)
+are available. Linters tend to be
+non denominational&mdash;rules don't have to agree with each other.
```
kwalrath: It should be just "nondenominational", one word. (Actually, "non-denominational" is used often enough that it's justifiable, e.g. see https://en.wikipedia.org/wiki/Non-denominational, which uses both forms. But  "non denominational" would just be wrong.)

Most "non" words are hyphen free, as you can see from the big list at  http://www.merriam-webster.com/dictionary/non-.

src/_guides/language/analysis-options.md
```diff
@@ -255,14 +266,9 @@ Join the discussion list for linter enthusiasts:
...
 
-* [Strong Mode](https://github.com/dart-lang/dev_compiler/blob/master/STRONG_MODE.md#strong-mode)
-
-* [Dart Linter](https://github.com/dart-lang/linter#linter-for-dart)
-
-* [Linter Rules](http://dart-lang.github.io/linter/lints/)
-
+* [Strong mode](https://github.com/dart-lang/dev_compiler/blob/master/STRONG_MODE.md#strong-mode)
+* [Dart linter](https://github.com/dart-lang/linter#linter-for-dart)
+* [Dart linter Rules](http://dart-lang.github.io/linter/lints/)
```
kwalrath: Rules -> rules

src/_guides/language/analysis-options.md
```diff
@@ -121,19 +132,19 @@ analyzer:
```
kwalrath: As I told you in person, these flags don't make sense to me. To enable both flags, you set them to false? And then there's the double-negative: `no-implicit-*: false`. So confusing.

kwalrath:
Very close! In addition to my comments above, we also discussed the need to change "package root" in the figure to "project root".

Or I suppose you could just delete "package root" and the line from it.

Sfshaza:
@jmesserly, could you review the section on strong mode? Thanks!

@kwalrath, I believe this incorporates all of your feedback.

Staged here: https://sz-www-1.firebaseapp.com/guides/language/analysis-options

Sfshaza:
@kwalrath 

Staged: https://sz-www-1.firebaseapp.com/guides/language/analysis-options#specifying-strong-mode

kwalrath:
LGTM!

pq:
:metal:

PR #127, opened 2016-09-02
Update mixins.md

anders-sandholm:
Tried to implement feedback from https://github.com/floitschG and https://github.com/lrhn on the previous change.


filiph:
Thanks!

kwalrath:
I was surprised that three back ticks in a row (within a line) would work, but they do! (https://www.dartlang.org/articles/language/mixins) I'll probably change them to use one....

Issue #70, opened 2016-07-21
What Not to Commit page needs updating wrt files created by dart2js?


Sfshaza:
_From @ArgentiApparatus on October 22, 2015 13:9_

https://www.dartlang.org/tools/private-files.html
https://github.com/dart-lang/www.dartlang.org/blob/master/src/site/tools/private-files.markdown

The file patterns given for files created by *dart2js* are:

```
*.dart.js
*.js_
*.js.deps
*.js.map
```

However, running *dart2js* (SDK 1.12) on one of my dart projects results in:

```
> dart2js .\web\main.dart

> ls out.*
out.js
out.js.deps
out.js.map
```

Presumably newer versions of *dart2js* use the extension `.js` instead of `dart.js` and do not generate `.js_` files. 

I note that *dart2js* may also produce `out.info.json` if the `--dump-info` option is specified.

Should the file patterns given for files created by *dart2js* be updated to:

```
*.js
*.js.deps
*.js.map
*.info.json
```

?

_Copied from original issue: dart-lang/www.dartlang.org#1496_

Sfshaza:
_From @sethladd on October 22, 2015 15:46_

cc @sigmundch Would you have a moment to provide some thoughts on which files we should be ignoring?

@ArgentiApparatus thanks for the issue! Once we get confirmation, we'd welcome a PR to update the doc.

Sfshaza:
_From @ArgentiApparatus on October 22, 2015 16:11_

I'd be glad to.

BTW I have a pull request in for Stagehand to update .gitignore templates that may be effected by the outcome of this issue: https://github.com/google/stagehand/pull/281

Sfshaza:
_From @sigmundch on October 22, 2015 20:3_

Good questions. I think we should keep the *.dart.js and *.js_, but sounds great to add the *.info.json.

**Regarding `*.js` vs `*.dart.js`**

Adding *.js to the .gitignore might be too aggressive as it is common for dart projects to sometimes contain source files written in JavaScript.

You are correct that calling dart2js without any flags will generate `out.js`, but tools that invoke dart2js (and often people do this by hand too) specify an output name such as:
```bash
dart2js -omain.dart.js main.dart
```

`pub-build` does that by default, in the past I believe the editor used to do this too, not sure if WebStorm does it as well, but it might.

[...]

PR #133, opened 2016-09-07
Updating the "What Not to Commit" page.

Sfshaza:
Fixing: https://github.com/dart-lang/site-www/issues/70

@sigmundch, could you review?

Staged here: https://sz-www-2.firebaseapp.com/guides/libraries/private-files


src/_guides/libraries/private-files.md
```diff
@@ -112,3 +114,16 @@ that make debugging easier.
 For more information on generating and debugging
 JavaScript source for Dart apps, see the
 [dart2js documentation]({{site.webdev}}/tools/dart2js).
+
+### *.js
```
sigmundch: interesting, I wonder if this is too general - is it commonly because of "out.js", would it make sense to just exclude that file only?

src/_guides/libraries/private-files.md
```diff
@@ -112,3 +114,16 @@ that make debugging easier.
...
+
+Calling dart2js without any flags generates `out.js`, but the
+tools that invoke dart2js often specify an output name, such as:
+
+{% prettify none  %}
+dart2js -o main.dart.js main.dart
+{% endprettify %}
+
+Don't commit JavaScript files generated by dart2js. Just be careful
```
sigmundch: An idea - what if we combine this and the previous section so that it's all one idea?

For example:
> Don't commit JavaScript files generated by dart2js. Just be careful that you don't accidentally exclude source files written in JavaScript.

> Calling dart2js without any flags generates `out.js`, but the tools that invoke dart2js often specify an output name, such as:

> {% prettify none  %}
dart2js -o main.dart.js main.dart
{% endprettify %}

> For more information on generating and debugging JavaScript source for Dart apps, see the JavaScript source for Dart apps, see the [dart2js documentation]({{site.webdev}}/tools/dart2js). [dart2js documentation]({{site.webdev}}/tools/dart2js).

sigmundch:
changes lgtm, just some suggestions

PR #135, opened 2016-09-08
Adding strong mode info to the "Why Dart Types are Optional" article

### LINKED ISSUE #86: Add strong mode info to types article
_From @kwalrath on November 6, 2015 22:29_

https://www.dartlang.org/articles/why-dart-types/ is one of our first articles. We might want to mention Strong Mode in there.

**How to use strong mode**

With the command line [analyzer](https://github.com/dart-lang/analyzer_cli#dartanalyzer), `--strong-mode` gets you into strong mode.  With the analysis server, put

```
analyzer:
  strong-mode: true
```

into your .analysis_options file, and you should start seeing strong mode errors in your IDE.  

**More info**

* https://pub.dartlang.org/packages/analyzer
* https://github.com/dart-lang/dev_compiler/blob/master/STRONG_MODE.md

There are still quite a few rough edges - the error messages are still very DDC specific, the lack of generic methods (very soon!) makes it hard to make strong mode happy in some cases, and there are still some warts in the implementation. 

_Copied from original issue: dart-lang/www.dartlang.org#1517_

Sfshaza:
@kwalrath, staged here:
https://sz-www-1.firebaseapp.com/articles/design-decisions/why-dart-types

Closes #86

I also fixed the formatting.

### DIFF
--- a/src/_articles/design-decisions/why-dart-types.md
+++ b/src/_articles/design-decisions/why-dart-types.md
@@ -10,37 +10,56 @@ _Written by Eli Brandt<br>
 December 2011_
 
 Dart uses types in a way that might seem strange.  Most popular languages
-  that offer a type system use it very differently.  If you're familiar with
-  types in Java, Haskell, Scala, or any statically typed language, you may
-  wonder why on earth Dart makes the choices that it does.
+that offer a type system use it very differently.  If you're familiar with
+types in Java, Haskell, Scala, or any statically typed language, you may
+wonder why on earth Dart makes the choices that it does.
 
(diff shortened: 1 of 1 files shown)


kwalrath:
Is it true that the only content that changed was (1) adding the aside about strong mode and (2) adding a comment to one sample? If so, LGTM.

(Even with ?w=1, GitHub made it look like much more changed than just whitespace. So annoying.)

Sfshaza:
Yes, that's the only change, plus changing one of the links to use markdown instead of HTML. The entire thing was formatted badly.

Issue #130, opened 2016-09-07
anonymous functions unexplained in language tour


ben4ever:
Anonymous functions are talked about in [this article](https://www.dartlang.org/resources/dart-tips/dart-tips-ep-7) but they aren't explained in the [language tour](https://www.dartlang.org/guides/language/language-tour).

Sfshaza:
You can see the new section here:
https://www.dartlang.org/guides/language/language-tour#anonymous-functions

PR #138, opened 2016-09-13
Added a section on anonymous functions.

Sfshaza:
@kwalrath. @leafpetersen, can you look at this? Kathy and I weren't sure that "anonymous function" is the correct term.

Issue: https://github.com/dart-lang/site-www/issues/130

Staged: https://sz-www-1.firebaseapp.com/guides/language/language-tour#anonymous-functions


src/_guides/language/language-tour.md
```diff
@@ -1096,8 +1097,59 @@ var loudify = (msg) => '!!! ${msg.toUpperCase()} !!!';
...
 
+This example uses an anonymous function.
+More about those in the next section.
 
-### Lexical scope {#lexical-scope}
+### Anonymous functions
+
+Most functions are named, such as `main()` or `printElement()`.
+You can also create a nameless function,
+called an _anonymous function_.
```
leafpetersen: Calling it an "anonymous function" is fine.  I'd say something like "called an anonymous function, or sometimes a lambda or closure", since either of the latter two terms have become fairly widely used.

src/_guides/language/language-tour.md
```diff
@@ -1096,8 +1097,59 @@ var loudify = (msg) => '!!! ${msg.toUpperCase()} !!!';
...
+An anonymous function is typically invoked once, or only a few times.
```
leafpetersen: Not entirely sure that this is true, but not too important either way.

src/_guides/language/language-tour.md
```diff
@@ -1096,8 +1097,59 @@ var loudify = (msg) => '!!! ${msg.toUpperCase()} !!!';
...
+You might assign an anonymous function to a variable so that,
+for example, you can add or remove it from a collection.
```
leafpetersen: This isn't the most common use of these, but it's a valid one.  Might be worth describing in text the use case that you use in the example below?

src/_guides/language/language-tour.md
```diff
@@ -1096,8 +1097,59 @@ var loudify = (msg) => '!!! ${msg.toUpperCase()} !!!';
...
+
+An anonymous function looks similar to a named function&mdash;an optional parameter
```
leafpetersen: Not sure if you want to go into it here or not, but it can be multiple parameters.  Also the parameters can have types.

leafpetersen:
lgtm.

Issue #134, opened 2016-09-08
Creation of a Videos page


packtpartner:
Since videos is the new tool that is catching up in the learning of developers, I would suggest if you could create a Videos page just like the Books page (https://www.dartlang.org/resources/books)

Packt has published a new video on "[Learning Dart](https://www.packtpub.com/web-development/learning-dart-video)" this year. We would appreciate if you could feature this as it intends to help with the learning of the Dart community.

Would love to hear your thoughts on this.

packtpartner:
@leafpetersen could you help me out on this please?

Sfshaza:
Perhaps, in the short term, we could extend the Books page to "Books and Videos".

packtpartner:
@Sfshaza thanks for updating the page and adding the video. Highly appreciated

PR #142, opened 2016-09-15
Expanding the books page to also include videos.

Sfshaza:
And adding a video published by Packt.

Fixed: https://github.com/dart-lang/site-www/issues/134

Staged: https://sz-www-2.firebaseapp.com/resources/books

@kwalrath, please review


src/resources/books.md
```diff
@@ -1,14 +1,14 @@
...
 layout: default
 permalink: /resources/books
-title: Books about Dart
+title: Books and Videos about Dart
 description: Read all about it! Here's a collection of books about Dart.
 toc: false
 ---
 
-Here's a collection of books about Dart.
+Here's a collection of books and videos about Dart.
```
kwalrath: Should we qualify the "videos" part somehow? We're not including our videos (although maybe we should point to them from here). And I assume we won't include short videos. We're really talking about major resources here. I'm not sure what words to use, though...

src/resources/books.md
```diff
@@ -1,14 +1,14 @@
...
 [Let us know](https://github.com/dart-lang/site-www/issues)
-if you find another book that we should add.
+if you find another resource that we should add.
```
kwalrath: My recent dive into the style guide has informed me that we should almost always put the qualifier before the "do this" thing. So:

If you ...., [let us know](...).

src/resources/books.md
Sfshaza: I thought about this, too. We don't want to include  youtube videos or Dart Tips, for example. It seems like we shouldn't have to explain it but I could, perhaps, add "packaged" videos. Not loving it, though. I could say, "professionally produced books and videos", but I'm not loving that, either. I think we should leave it be and just vet any thing that comes in.

Sfshaza:
Staged: https://sz-www-2.firebaseapp.com/resources/books

@kwalrath

src/_tutorials/dart-vm/cmdline.md
```diff
@@ -29,7 +29,11 @@ prevpage:
 <aside class="alert alert-info" markdown="1">
 <strong>Prerequisite:</strong>
 
+<<<<<<< Updated upstream
```
kwalrath: Uh oh, fix this!

kwalrath:
LGTM!

Issue #149, opened 2016-09-27
print styles need work


kwalrath:
When you print out a page, you see all kinds of stuff that you shouldn't—headers, footers, and more.

kwalrath:
Partially fixed by #150. I should apply these fixes to webdev.dartlang.org, as well.

Some other print format issues I noticed:

* Notes should have a box around them, but don't. They just look indented for no reason (especially in notes where there's no bold title to lead off... we really should make a point of doing that). For an example, see https://www.dartlang.org/tools/pub/cmd.

* In Canary, at least, the printout often has blank pages at the end. For example, https://www.dartlang.org/tools/pub/cmd has 3 pages of content followed by 4 blank pages.

Sfshaza:
Another issue. Highlighted text in code blocks doesn't display.

kwalrath:
Another issue: local URLs print (in parens) after their link, but external URLs don't. If we were to differentiate, the opposite should be true. I'm inclined to just underline the link text and NOT provide the URL in the printed version. The external link icon could probably be dropped, too.

kwalrath:
I'm working on some fixes... should have a PR soon.

kwalrath:
The issues might be fixed... but I'm not sure. I'd love for someone else to take a look and see what we could do to improve.

Sfshaza:
I've looked at several pages on www.dartlang. I brought up the print panel, and viewed the preview there - I didn't actually print.

* There are still no boxes around notes. (same as reported)
* Local URLs still display, but you only see the link icon for external links. (same as reported)

Overall, I think the output looks pretty good!

Sfshaza:
Oh, the table in the Get Started page doesn't look that great, but I doubt this is a page that will be printed much.

kwalrath:
There might still be a few ways we can improve print styles (maybe code highlighting?), but we can file individual bugs for those.

PR #150, opened 2016-09-28
Don't display navigation when printing

kwalrath:
To see the changes, go to the staging site and view a page. Here's one I've used for testing, but any other page should work as well (especially if it's long):

https://kw-www-dartlang-1.firebaseapp.com/guides/language/effective-dart/style

Then view the print version of the page. The easiest way is to just bring up the Chrome print dialog, which has a preview.

Compare it to the existing version of the page, which has bad print styles (#149).

@Sfshaza could you please take a look?


Sfshaza:
Looks much better!

LGTM

PR #152, opened 2016-09-30
Update zones.md

matanlurey:
`scheduleTask` was renamed to `scheduleMicrotask` a long time ago :)


filiph:
Thanks. Confirmed this was the only place where `scheduleTask` appeared on dartlang.org. Will check webdev.dartlang.org now.

Issue #13, opened 2016-07-14
Can't permalink to a style rule for Effective Dart


jmesserly:
Previously you could link to a particular style rule:
https://web.archive.org/web/20160702200714/https://www.dartlang.org/effective-dart/usage/#do-use-adjacent-strings-to-concatenate-string-literals

The permalink is easily accessible on the right of each rule. We use this all the time for Dart Readability reviews in Google, to point folks at a specific item.

In the new site it doesn't seem possible:
https://www.dartlang.org/guides/language/effective-dart/usage

The ids are still there on each item, if you open up web inspector. So it's possible to make a permalink, but not easy.

jmesserly:
Source code for Effective Dart seems to be: https://github.com/dart-lang/site-www/blob/master/src/_guides/language/effective-dart/

Sfshaza:
OK, it sounds like we need to add this back. (The links are still there in the source, but you have to view source to find it.)

How would you prefer this feature to work?
1. A generated TOC at the beginning of the file?
2. Hovering over the header to see a hash on the right that you click?

Sfshaza:
Also filed here:  https://github.com/dart-lang/site-www/issues/9  (closed as a duplicate)

jmesserly:
Oh nice! Yeah either way works. Slight preference for option 2, but whatever is easier :)

har79:
Could we have both? I would really appreciate a TOC for finding rules. I often want to reference a rule I know exists but can't even remember which page of effective-dart it's in, let alone where on the page, and end up with four tabs and Ctrl-F to find it. A TOC would make this a lot quicker and make it easier to remember the list of rules in each page.

jmesserly:
yeah, good point the groupings into different sections has never made a lot of sense. Most style guides put everyone on one page for that reason (easy to search)

kwalrath:
In the short term, we plan to produce a TOC at https://www.dartlang.org/guides/language/effective-dart that lists and links to _all_ of the rules.

Let's see how that goes, and then we'll add anchors if need be.

jmesserly:
sounds great! thank you :)

har79:
That would be great, thanks!

gmesch:
So if I want to link to the section I'm looking at, I have to go to the TOC, find the headline of the same section again, and copy the link from there? I admit this doesn't sound so convincing.

filiph:
Here's the quick fix @kwalrath mentioned above:
[...]

PR #153, opened 2016-10-07
Add TOCs to effective dart pages.

har79:
Staged at https://hc-www-dartlang-1.firebaseapp.com/guides/language/effective-dart/style

The TOC can be configured through the kramdown option toc_levels, but seems to default to exactly what we would want.

May help alleviate the issues in https://github.com/dart-lang/site-www/issues/13 although I hope to have a full fix for that later today.


Sfshaza:
LGTM

PR #155, opened 2016-10-10
Prevent headings from being hidden under the main header bar.

har79:
Staged: https://hc-www-dartlang-1.firebaseapp.com/guides/language/effective-dart

When clicking on the anchor link beside a heading it was being positioned at the top of the screen and obscured by the header bar. Adding a positive padding, causes it to load, unobscured, below the header bar and an opposing negative margin prevents any change to the page layout. This was required for both the heading itself and the anchor as they each have IDs which sometimes differ subtly. The JS which was used on page load to scroll to the heading is now no longer necessary and has been removed.

I also, restricted all the anchor link related CSS to only apply to the central content section, and extracted the sizes to variables.


Sfshaza:
Excellent! LGTM.

parlough:
I know this is quite a bit late, but I found that this solution causes some issues with highlighting text and clicking links above the headings and the anchors as it essentially creates an invisible box over the above text. For most cases on the website this doesn't occur due to their being enough padding before the headers, but there are a few locations, such as the following on https://dart.dev/resources/books:

image

I couldn't think of a great alternative solution without using JavaScript as I don't have a lot of CSS experience. Perhaps you have a suggestion?

har79:
Good catch, thanks!

So, there is actually now a css property for exactly this case: `scroll-margin`. Unfortunately it's not yet supported by Edge. Also, unfortunately `pointer-events` doesn't work onpseudo elements or it would be possible to set that to allow clicking through to the link below.

Good news though! I think I've found an alternative solution that should work across browsers and still allow clicking through to links.

kwalrath:
Thank you, @parlough & @har79! I'd never noticed the issue, but sure enough, I can't click the **Flutter books** link in that page.

kwalrath:
(Would one of you mind opening an issue for this?)

har79:
Done: https://github.com/dart-lang/site-www/issues/2230

PR #156, opened 2016-10-12
Remove unnecessary "there might be bugs" warning

matanlurey:
Either we have this feature or we don't - if there aren't any major bugs yet we shouldn't have this warning. It's been a while since 1.9...

/cc @filiph


Sfshaza:
LGTM!

filiph:
LGTM, good find.

PR #157, opened 2016-10-17
Fix Dartium + Mac links

kevmoo:
Also note the cleanup of the package dir crazy

Careful when you commit next that nothing leaks through


src/install/archive/web/download_archive.dart
```diff
@@ -258,3 +276,21 @@ void addVersion(String channel, Map<String, String> version) {
     });
   }
 }
+
+final RegExp regexp120 = new RegExp(r'^(\d+)\.(\d+)\.');
+
+// TODO: use pkg/pub_semver
+bool is120OrGreater(String versionString) {
+  var thing = regexp120.firstMatch(versionString);
```
whesse: Would "match" be a better name than "thing"?

Sfshaza:
It would be a better name, Bill! I went ahead and merged and will fix it separately.

Sfshaza:
It's live! (I also renamed thing.)

https://www.dartlang.org/install/archive

PR #187, opened 2016-12-09
Update redirects

filiph:


filiph:
Actually, hold it. I want to sort the redirects and remove duplicates.

filiph:
Ok now it's ready for review.

firebase.json
```diff
@@ -16,43 +16,38 @@
...
-        "source": "/articles/numeric-computation/",
-        "destination": "/articles/dart-vm/numeric-computation",
+        "source": "/+lexicalscope",
+        "destination": "/guides/language/language-tour#lexical-scope",
         "type": 301
       },
       {
-        "source": "/articles/simd/",
-        "destination": "/articles/dart-vm/simd",
+        "source": "/+pub",
```
kevmoo: What does the `+` do here? Just curious.

kwalrath: IIRC, @sethladd created a bunch of /+ shortcuts. For example, near me I see an old poster for lexical scope. At the bottom of it is the URL dartlang.org/+lexicalscope.

firebase.json
```diff
@@ -721,7 +726,7 @@
         "type": 301
       },
       {
-        "source": "/pub/tools/editor",
```
kwalrath: How do you know whether the original URL was editor/ or editor.html?

filiph: We don't :(. Were they different pages at some point?

kwalrath: I don't recall. I was just wondering if you had a trick.

Issue #180, opened 2016-12-05
Change optional param separator from : to = in language tour (and elsewhere)


kevmoo:
_From @sethladd on December 2, 2016 20:47_

The 1.21 Dart language support "Allow = as well as : as separator for named parameter default values."

We could mention that in https://www.dartlang.org/guides/language/language-tour

_Copied from original issue: dart-lang/sdk#27976_

lrhn:
If you change anything, I would just change them all to `=`. It's more "future-safe" since we are likely to deprecate `:` in v1.50+ and remove it in v2.0.

sethladd:
Thanks for the advice!

kwalrath:
We'll need a note about : for people who are using older releases/code.

kwalrath:
I'm going to close this, since I can't think of other uses of `:` (nor an easy way to search for them).

PR #188, opened 2016-12-09
Remove optional named params section

mohamedhayibor:
* Also updated code sample to a positional parameter one

This is a start to address https://github.com/dart-lang/site-www/issues/180

Will  iterate upon any feedback. Thanks for your time.

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -889,35 +889,7 @@ A function can have two types of parameters: required and optional. The
 required parameters are listed first, followed by any optional
 parameters.
 
-
-### Optional parameters
-
-Optional parameters can be either positional or named, but not both.
-
-#### Optional named parameters
-
-When calling a function, you can specify named parameters using
-<code><em>paramName</em>: <em>value</em></code>. For example:
-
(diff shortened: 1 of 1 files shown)


kwalrath:
Thanks for the PR! My understanding is that we shouldn't remove anything, but just change the featured syntax (for specifying default values for named parameters) from `:` to `=`. We should also have a note that in SDKs before 1.21, named parameters used `:` instead of `=`, but that `:` for named parameters is likely to be phased out.

The note is important because the pubspec should specify ^1.21 for the SDK if the code uses `=` to specify named parameter default values. (We might want to mention the SDK dependency explicitly... We haven't done that in the past, but perhaps we should.)

I was actually planning to work on this issue today, but I'm happy to let you do the work if you'd like to.

mohamedhayibor:
Thanks for the explanation @kwalrath . I'll let you do the modifications. I commented on the gist used If that could save you a second or two: https://gist.github.com/Sfshaza/d988cfce0a54c6853799

I'm always nervous to close a PR though. Is it fine if I do?

I'll hunt out other issues to tackle 😄

kwalrath:
Thanks very much, @mohamedhayibor! It's fine to close your own PR. But since I'm here...

Issue #181, opened 2016-12-05
Request to add generic methods to the Language Tour


kevmoo:
_From @sethladd on December 2, 2016 20:46_

https://www.dartlang.org/guides/language/language-tour doesn't seem to have generic methods.

The 1.21 changelog mentions it support generic methods. https://github.com/dart-lang/sdk/blob/master/CHANGELOG.md

Here is the "informal specification" : https://gist.github.com/eernstg/4353d7b4f669745bed3a5423e04a453c

Thanks!

_Copied from original issue: dart-lang/sdk#27975_

Sfshaza:
Here is some info:
https://github.com/dart-lang/sdk/blob/master/pkg/dev_compiler/doc/GENERIC_METHODS.md

kwalrath:
This should probably go in the [Generics section](https://www.dartlang.org/guides/language/language-tour#generics) of the language tour, but doing it well might require some thought. Also, some other parts of the language tour might be affected.

A quick fix would be to add a note to the Generics section that refers to [Using Generic Methods](https://github.com/dart-lang/sdk/blob/master/pkg/dev_compiler/doc/GENERIC_METHODS.md) (which should probably itself refer to the informal spec).

kwalrath:
I'm going to close this, but feel free to reopen this (or create a new issue) if you think of a better way/place to handle this.

PR #199, opened 2016-12-13
Document generic methods in the language tour

kwalrath:
Partial fix for #181.

Staged at: https://kw-www-dartlang-1.firebaseapp.com/guides/language/language-tour#using-generic-methods

The example (https://dartpad.dartlang.org/a02c53b001977efa4d803109900f21bb) could probably be better. I was aiming for something simple that showed a variety of ways to use a type parameter.

cc @sethladd @kevmoo

### DIFF
--- a/.firebaserc
+++ b/.firebaserc
@@ -1,5 +1,6 @@
 {
   "projects": {
-    "default": "www-dartlang-org"
+    "default": "www-dartlang-org",
+    "kw-staging": "kw-www-dartlang-1"
   }
 }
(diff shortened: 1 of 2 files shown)


kevmoo:
Added @munificent and @leafpetersen – first pass looks good to me, though.

kwalrath:
Thanks, @leafpetersen. This should be published within a few minutes.

sethladd:
Thanks @kwalrath !

PR #201, opened 2016-12-13
Adding a page on writing sound Dart code.

Sfshaza:
This version includes some restructuring from the Google doc version. I've removed all references to covariance and contravariance, which are terms that many programmers don't seem to know. 

Thanks for your feedback. It is much appreciated!

@kwalrath  @prose  @dgrove @srawlins  @leafpetersen @munificent


src/_guides/language/sound-dart.md
```diff
@@ -0,0 +1,671 @@
...
+
+A sound type system has several benefits:
+
+* Revealing type-related bugs at compile time.<br>
+  A sound type system forces code to be unambiguous about its types,
+  so type-related bugs that might be tricky to find at runtime are
+  revealed at compile time.
+
+* More readable code.<br>
+  Code is easier to read because you can rely on a value actually haing
```
filiph: haing -> having

src/_guides/language/sound-dart.md
```diff
@@ -0,0 +1,696 @@
...
+{% endcomment %}
+
+<aside class="alert alert-info" markdown="1">
+**Note:** The terms "sound Dart", "strong mode Dart", and "type safe Dart"
+are sometimes used interchangeably. _Strong mode_ is Dart's implementation
+of a sound type system. With strong mode enabled,
+Dart is a type safe language. "Classic Dart" refers to Dart
+before soundness was added to the language.
+
+Eventually, soundness won't be optional in the Dart language, and strong mode
```
leafpetersen: I don't think we've officially announced the "won't be optional part of this yet".

src/_guides/language/sound-dart.md
```diff
@@ -0,0 +1,696 @@
...
+<div class="fails-sa" markdown="1">
+{% prettify dart %}
+void fn(List<int> a) {
+  print(a);
+}
+
+main() {
+  var list = [];
+  list.add(1);
+  list.add(2);
```
leafpetersen: I think this example would be a better introduction to soundness if the code was something that worked, but that also didn't really match the documented intent.  So for example, using `list.add("2")`.

src/_guides/language/sound-dart.md
```diff
@@ -0,0 +1,696 @@
...
+class Animal {
+  void chase(Animal a) {}
+  [[highlight]]HoneyBadger[[/highlight]] get parent => ...
+}
+{% endprettify %}
+</div>
+
+<div class="fails-sa" markdown="1">
+{% prettify dart %}
+class Animal {
```
leafpetersen: Likewise?

src/_guides/language/sound-dart.md
```diff
@@ -0,0 +1,696 @@
...
+
+<div class="passes-sa" markdown="1">
+{% prettify dart %}
+// Inferred as if you wrote <int>[].
+List<int> listOfInt = [];
+
+// Inferred as if you wrote <double>[3.0].
+var listOfDouble = [3.0];
+
+// x is inferred as int, double using downward information.
```
leafpetersen: Typo: should be "x is inferred as double using"

src/_guides/language/sound-dart.md
```diff
@@ -0,0 +1,696 @@
...
+
+If you use [DartPad](/tools/dartpad) to write and test code, you can
+enable strong mode by checking the **Strong mode** box in the lower
+right corner. Note that DartPad doesn't support the implicit casts flag,
+implicit dynamic flag, or enabling linter rules. For this functionality
+you must use DDC or IntelliJ for now.
+
+## Substituting types
+
+When can you replace a type with a subtype or a supertype?
```
leafpetersen: Ok, this terminology is ok, but I think it's confusing without some development.  This isn't really about substituting types, it's about substituting things which have one type for things which have a difference one.  So maybe something like:

"When you override a method, you are replacing something of one type (the old method) with something of a new type (the new method).  Similarly, when you pass an argument to a function, you are replacing something which has one type (a parameter with a declared type) with something that has another type (the actual argument).  When can you replace something with which has one type with something that has a subtype or a supertype?"

src/_guides/language/sound-dart.md
```diff
@@ -0,0 +1,696 @@
...
+The annotation applies to a single parameter and can be placed on the
+superclass or subclass method.
+Usually the superclass method is the best place to put it.
+
+## Strong mode vs. checked mode
+
+You may be famililar with the Dart compiler's checked mode feature.
+In checked mode, the compiler inserts dynamic type assertions and
+generates a warning if the types don't match up. For example,
+the following line of code generates a compile-time warning in checked mode:
```
leafpetersen: run-time instead of compile-time?

leafpetersen: Or are you using "checked mode" as a shorthand for "classic Dart"?

src/_guides/language/sound-dart.md
```diff
@@ -0,0 +1,696 @@
...
+This code raises no issues when run in checked mode, but generates
+numerous errors when analyzed under strong mode.
+
+## Other resources
+
+The following resources have further information on sound Dart and
+strong mode:
+
+* [Sound Dart](https://www.youtube.com/watch?v=DKG5CMyol9U) - Leaf
+  Peterson's talk from 2016 Dart Summit
```
leafpetersen: Petersen

src/_guides/language/sound-dart.md
lrhn: I'll have to object to the entire paragraph. It's mixing different usages of the word "substitution", and in an OO setting you should only ever use one meaning: Lisskov's Substitution Principle, which says that you can always use an instance of a subclass where an instance of a superclass is expected. (More pedantically, it says that any visible property of instances of the supertype must also hold for instances of the subtype).

This is substitution at the *value* level - a value of one type can be used instead of a value of another type. We represent that in the language by making subtype values assignable to supertype'd variables. We *allow* that assignment because the principle tells us that it's safe.

When you "override", you are not replacing anything. You *create* a subclass which has a different type for that method. You want/need the subclass to be a subtype, therefore you are required to make the new method type substitutable for the supertype method's type.

Passing a parameter to a function is not substitution at all (unless we are doing this as lambda calculus). It's about assignability - you assign a value to a variable.

Now, it's late and I haven't read the document yet, so I'm not sure what the section is supposed to say, so I don't have brilliant suggestion for what to write instead. I just know that this paragraph got me completely confused about what it is trying to say, probably because "substituting types" is meaningless to me :)

leafpetersen: > When you "override", you are not replacing anything

Sure you are.  You're creating a version of the superclass with some methods replaced with new ones.  :)

This section is technically suspect, but I think the goal is to give some intuition without getting into subtyping and assignability per se, but it's hard for me to see where to draw the line.  Maybe something closer to this:

"When you override a method, you are replacing something of one type (the old method) with something of a new type (the new method). Similarly, when you pass an argument to a function, you are assigning something which has one type (the actual argument) to a location that was expecting something of some other type (the function parameter). When can you override a method of one type  with a method of another type?  And when can you assign something of one type to a location of another type?

When thinking about types, it helps to think in terms of producers and consumers."

etc.

Honestly though, @munificent is much better at sorting out how to express these things in  an accurate but understandable to non-Klingon-speakers way.  Bob?

Sfshaza:
I hadn't seen Leaf asking you to comment, Bob. I did merge this PR, but if
you have more feedback, I'd be happy to incorporate!

munificent:
Oh, sorry. I've been swamped with other stuff, and I figured this was in good hands with you and Leaf. :)

Sfshaza:
No problem, but I'm about to send out a PR for a Sound Dart FAQ. Leaf's
already lgtm'd it. Did you want to take a look?

PR #203, opened 2016-12-17
fix sample code which can't work

joe-re:
"Write Command Line App" chapter's Sample code doesn't work.
I guess, it is wrong. fix it.


joe-re:
I signed it

PR #209, opened 2016-12-28
Sound Dart Guide -> Sound Dart

kwalrath:
We don't usually say "Guide" in titles. There's no reason to do so here, is there?


Sfshaza:
lgtm

PR #214, opened 2017-01-12
Define soundness more correctly

thejameskyle:
Hi!

So quick nit-picky suggestion. I don't want to come off as pedantic or anything like that. I just want to make this suggestion to make sure that people are learning the correct definition of soundness.

Implicit nullability is not really a "sound" behavior of a language. Unless of course the language accounts for that nullability everywhere. I don't know where Dart falls on this, but I do think this small blurb leaves people with the wrong impression.

I think it might also be useful to give a more complete definition of soundness, maybe even contrasting it with completeness.

I'm working on a new documentation for [Flow](https://flowtype.org/) right now and here's what I wrote about this topic:

> In type systems, ***soundness*** is the ability for a type checker to catch every single error that *might* happen at runtime. This comes at the cost of sometimes catching errors that will not actually happen at runtime.
> 
> On the flip-side, ***completeness*** is the ability for a type checker to only ever catch errors that *would* happen at runtime. This comes at the cost of sometimes missing errors that will happen at runtime.
> 
> In an ideal world, every type checker would be both sound *and* complete so that it catches *every* error that *will *happen at runtime.

If this is useful to you, feel free to use any or all of that.

Thanks 😸

### DIFF
--- a/src/_guides/language/sound-dart.md
+++ b/src/_guides/language/sound-dart.md
@@ -83,11 +83,15 @@ DartPad url: https://dartpad.dartlang.org/3c7c95683f0c06be8326a2fd3975cd19
 
 ## What is soundness?
 
-Soundness is about the relationship between the code that you write and
-the values that show up in the code at runtime. In a sound language,
-if the static type of an expression is `int`, once you evaluate the
-experssion to a value, you are guaranteed to get an `int` and nothing
-else.
+Soundness is the ability for a compiler to catch every error that may
+happen at runtime. In a sound language, the possible types of all
+values are known statically and guaranteed to be the same types at
+runtime.
+
+In Dart, if the static type of an expression is `int`, once you evaluate
+the expression to a value, you are guaranteed to either get an `int` or
+`null`. Dart will ensure that your code works with both an `int` or
+`null`.
 
 Dart was created as an optionally typed language and is not sound.
 For example, it is valid to create a list in Dart that contains
--- a/src/_guides/language/sound-dart.md
+++ b/src/_guides/language/sound-dart.md
@@ -87,7 +87,7 @@ Soundness is about the relationship between the code that you write and
 the values that show up in the code at runtime. In a sound language,
 if the static type of an expression is `int`, once you evaluate the
 experssion to a value, you are guaranteed to get an `int` and nothing
-else (except possibly null).
+else.
 
 Dart was created as an optionally typed language and is not sound.
 For example, it is valid to create a list in Dart that contains
--- a/src/_guides/language/sound-dart.md
+++ b/src/_guides/language/sound-dart.md
@@ -83,15 +83,19 @@ DartPad url: https://dartpad.dartlang.org/3c7c95683f0c06be8326a2fd3975cd19
 
 ## What is soundness?
 
-Soundness is the ability for a compiler to catch every error that may
-happen at runtime. In a sound language, the possible types of all
-values are known statically and guaranteed to be the same types at
-runtime.
+*Soundness* is about ensuring your program can't get into certain
+invalid states. A sound *type system* means you can never get into
+a state where an expression evaluates to a value that doesn't match
+the expression's static type. For example, if an expression's static
+type is `String`, at runtime you are guaranteed to only get a string
+when you evaluate it.
 
-In Dart, if the static type of an expression is `int`, once you evaluate
-the expression to a value, you are guaranteed to either get an `int` or
-`null`. Dart will ensure that your code works with both an `int` or
-`null`.
+Strong mode, like the type systems in Java and C# is sound. It
+enforces that soundness using a combination of static checking
+(compile errors) and runtime checks. For example, assigning a `String`
+to `int` is a compile error. Casting an `Object` to a string using
+`as String` will fail with a runtime error if the object isn't a
+string.
 
 Dart was created as an optionally typed language and is not sound.
 For example, it is valid to create a list in Dart that contains


dherman:
Soundness is relative to a set of promised invariants. A sound type system does not guarantee the absence of *all bugs ever* -- ML doesn't guarantee you won't divide by zero, for example, and Haskell doesn't guarantee that you won't send an HTTP response to the wrong socket. "Bug" is not even a formally defined concept.

Soundness means that you have formally classified a set of runtime states that you call type errors, promise that a well typed program will never enter those states, and then always keep that promise. As long as Dart's sound mode does not promise the absence of null dereferences, implicit nullability doesn't have to violate soundness.

Sfshaza:
@munificent @leafpetersen

munificent:
Thanks for chiming in!

@thejameskyle, your pull request is correct in that because *all* types are nullable in Dart, even `int`, and the type system and runtime account for that, it's not strictly necessary to call it out as special there. I think I encouraged @Sfshaza to add the parenthetical because most people coming to Dart are coming from languages where an int cannot be `null`.

Permitting or prohibiting `null` there doesn't interfere with soundness because, like @dherman notes, soundness is always with respect to some set of promises. As long as we don't, say, promise to prohibit nulls *and* then not check for them, we're sound.

My understanding is also that it's not strictly necessary to perform *all* checks statically in order to be called sound. I think most languages consider it OK to defer some of those checks to runtime. As long as your program can never reach a state for a promised invariant does not hold, you're sound. Preventing reaching that state by emitting a compile error and refusing to generate code is one way, but you could also throw an exception before the invariant would break, abort the process, turn the user's machine off, set their house on fire, etc.

I think that looser definition is necessary if you want to consider C# and Java sound (which most people do) because both of those defer array covariance checks until runtime (ClassCastException and  ArrayTypeMismatchException). Also, explicit casts may fail at runtime in either language.

By that token, JS is sound too because it doesn't let you're code get into an undefined state at runtime. You can't misinterpret bits in memory and crash the VM. JS just defers *all* the checks to runtime. :)

C isn't sound.—you can cast to `void*` and then back to an unrelated type and then all bets are off with your program. Rust and C# aren't sound within unsafe blocks, I think.

I do find it frustrating that programmers don't have clear definitions in their heads for "sound" and "type safe". We can pick one, of course, but that doesn't change all of the brains that already have a definition that doesn't agree with ours. :-/

> As long as Dart's sound mode does not promise the absence of null dereferences, implicit nullability doesn't have to violate soundness.

Exactly right. Dart's type system does treat `null` as a value of all types (it's the bottom type). Because it allows that, and because `null` doesn't actually support every method defined ever, it means the runtime has to check for a `null` receiver before invoking a method, which it does. But I still thought it was worth drawing a little attention to the fact that all types are nullable in Dart, since it's unusual for "primitive" types.

(We *are* working on seeing if we can retrofit non-nullable types into the type system. If we can make that work, then we won't need the parenthetical at all anymore. :) )

As far as a completeness goes, I thought that was undecidable in general for any Turing-complete language? Isn't that what Gödel proved?

thejameskyle:
Yup, I already agreed that implicit nullability is fine as long as the language accounts for it:

> Implicit nullability is not really a "sound" behavior of a language. Unless of course the language accounts for that nullability everywhere.

But I just want to raise the issue that you are defining "soundness" for the first time for readers here, I think it's important that you leave them with a more generalized description.

Of course soundness and completeness are not a definitive thing for verifying a program's validity, but they have well defined goals.

I think this small bit about nullability only serves to confuse the readers initial understanding of soundness. It would be more useful to break this section apart into:

1. Explain soundness in a general sense
2. Explain what soundness means in Dart

I think this would improve understanding of both soundness and Dart.

thejameskyle:
Here's a quick rough update with some suggested changes.

Sfshaza:
@thejameskyle, thanks for your attention to this! If, in fact, Dart's definition of soundness varies from the generally accepted definition of soundness, then I'm very much OK with this clarification. I defer to @munificent and @leafpetersen , of course.  I also agree that finding a good definition of soundness on the web is not easy, as I certainly looked.  :)

src/_guides/language/sound-dart.md
```diff
@@ -83,11 +83,15 @@ DartPad url: https://dartpad.dartlang.org/3c7c95683f0c06be8326a2fd3975cd19
 
 ## What is soundness?
 
-Soundness is about the relationship between the code that you write and
-the values that show up in the code at runtime. In a sound language,
-if the static type of an expression is `int`, once you evaluate the
-experssion to a value, you are guaranteed to get an `int` and nothing
-else (except possibly null).
+Soundness is the ability for a compiler to catch every error that may
+happen at runtime. In a sound language, the possible types of all
```
munificent: Catching *every* error would be nice, but isn't what even sound type systems claim to do.

How about:

```markdown
*Soundness* is about ensuring your program can't get into certain invalid states. A sound
*type system* means you can never get into a state where an expression evaluates to a value
that doesn't match the expression's static type. For example, if an expression's static
type is `String`, at runtime you are guaranteed to only get a string when you evaluate it.

Strong mode, like the type systems in Java and C# is sound. It enforces that soundness
using a combination of static checking (compile errors) and runtime checks. For example,
assigning a `String` to `int` is a compile error. Casting an `Object` to a string using
`as String` will fail with a runtime error if the object isn't a string.
```

I switched from `int` to `String` here because those are nullable in most languages, so I think we don't need to call out the surprising `null` case then.

What do you think?

thejameskyle: That sounds a lot better

munificent: Groovy, thanks for bringing this up!

@Sfshaza, do you want to merge this in?

munificent:
🤘

thejameskyle:
😃

PR #218, opened 2017-01-19
Directly explain why article is obsolete (via link)

filiph:
This is just a proposal.


kwalrath:
I like the direct link, but I think we also need to link to the Sound Dart doc prominently, since that's what they should be reading instead.

kwalrath:
I'm going to close this and create a new PR.

kwalrath:
See #221.

PR #220, opened 2017-01-23
Add rule about returning void in a setter

### LINKED ISSUE #131: Please add to "Effective Dart" a recommendation to omit "void" in setters
bad:

``` dart
void set foo(Foo value) {...}
```

good:

``` dart
set foo(Foo value) {...}
```

The `void` is pointless noise.  Please formalize this in [Effective Dart](https://www.dartlang.org/guides/language/effective-dart) so that we have something official to point to in code review.  I got a soft promise from rnystrom@google.com that he would do this sometime in 10/2016.

kwalrath:
Fixes #131

### DIFF
--- a/src/_guides/language/effective-dart/design.md
+++ b/src/_guides/language/effective-dart/design.md
@@ -744,9 +744,9 @@ Future<bool> install(PackageId id, String destination) {
 With types, all of this is clarified.
 
 
-### DON'T specify that a setter returns `void`.
+### DON'T specify a return type for a setter.
 
-Although the `void` type is technically correct, it's pointless noise.
+The type system infers `void` for all setters automatically.
 
 <div class="bad">
 {% prettify dart %}
--- a/src/_guides/language/effective-dart/design.md
+++ b/src/_guides/language/effective-dart/design.md
@@ -743,6 +743,24 @@ Future<bool> install(PackageId id, String destination) {
 
 With types, all of this is clarified.
 
+
+### DON'T specify that a setter returns `void`.
+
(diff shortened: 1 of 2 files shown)


kwalrath:
Oops, I also need to update /guides/language/effective-dart to link to the new rule.

kwalrath:
OK, I've updated the TOC. Ready for review.

src/_guides/language/effective-dart/design.md
```diff
@@ -743,6 +743,24 @@ Future<bool> install(PackageId id, String destination) {
 
 With types, all of this is clarified.
 
+
+### DON'T specify that a setter returns `void`.
```
munificent: This reads to me like it might be discouraging settings *in general*. How about:

```markdown
### DON'T specify a return type for a setter.
```

src/_guides/language/effective-dart/design.md
```diff
@@ -743,6 +743,24 @@ Future<bool> install(PackageId id, String destination) {
...
+
+Although the `void` type is technically correct, it's pointless noise.
```
munificent: Saying it's "technically correct" might lead readers to believe it adds some information, but it really doesn't. The language infers `void` for setters anyway, so it adds nothing. Maybe just say:

> The type system infers `void` for all setters automatically.

kwalrath:
Review comments incorporated. Thanks, @munificent!

kwalrath:
Published:

https://www.dartlang.org/guides/language/effective-dart/design#dont-specify-a-return-type-for-a-setter

PR #221, opened 2017-01-24
revise "obsolete" wording

kwalrath:
* Added “(Obsolete)” to the description to make it show up in the
article index and summary text.
* We could use “**This page is obsolete** because” in all obsolete
pages.

Should we also mark [Optional Types in Dart](https://www.dartlang.org/articles/language/optional-types) as obsolete?

Staged:
https://kw-www-dartlang-1.firebaseapp.com/articles/design-decisions/why-dart-types
https://kw-www-dartlang-1.firebaseapp.com/articles/design-decisions
https://kw-www-dartlang-1.firebaseapp.com/articles#design-decisions

### DIFF
--- a/src/_articles/design-decisions/why-dart-types.md
+++ b/src/_articles/design-decisions/why-dart-types.md
@@ -1,7 +1,7 @@
 ---
 layout: article
 title: "Why Dart Types Are Optional and Unsound"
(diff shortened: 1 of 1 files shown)


kwalrath:
I'm going to merge this, but we can (of course) iterate.

PR #233, opened 2017-02-03
Fix the good example of default optional parameters

stefanchrobot:
- Removed the "new" since that this was not valid Dart code and I assume the examples were supposed to illustrate declarations, not calls.
- Updated the functions to be aligned with the current API.
- Removed the Duration example, since:
  - it does not add much extra value,
  - it's confusing regarding the syntax (the suggested syntax is "=", not ":"),
  - there's already a lot of code since the DateTime went multiline.

### DIFF
--- a/src/_guides/language/effective-dart/design.md
+++ b/src/_guides/language/effective-dart/design.md
@@ -939,9 +939,15 @@ pass later one. You're better off using named arguments for that.
 
 <div class="good">
 {% prettify dart %}
-new String.fromCharCodes(Iterable source, [int start = 0, int end])
-new DateTime(year, month, day, [hours, minutes, seconds, milliseconds])
-new Duration({days: 0, hours: 0, minutes: 0, seconds: 0, milliseconds: 0})
+String.fromCharCodes(Iterable<int> charCodes, [int start = 0, int end])
+DateTime(int year,
+         [int month = 1,
+          int day = 1,
+          int hour = 0,
+          int minute = 0,
+          int second = 0,
+          int millisecond = 0,
+          int microsecond = 0])
 {% endprettify %}
 </div>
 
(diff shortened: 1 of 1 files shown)


src/_guides/language/effective-dart/design.md
```diff
@@ -939,9 +939,15 @@ pass later one. You're better off using named arguments for that.
...
-new Duration({days: 0, hours: 0, minutes: 0, seconds: 0, milliseconds: 0})
+String.fromCharCodes(Iterable<int> charCodes, [int start = 0, int end])
+DateTime(int year,
+         [int month = 1,
+          int day = 1,
+          int hour = 0,
+          int minute = 0,
+          int second = 0,
+          int millisecond = 0,
+          int microsecond = 0])
```
munificent: Can you format this using dartfmt? Consistency is good. :)

src/_guides/language/effective-dart/design.md
```diff
@@ -939,9 +939,15 @@ pass later one. You're better off using named arguments for that.
 
 <div class="good">
 {% prettify dart %}
-new String.fromCharCodes(Iterable source, [int start = 0, int end])
-new DateTime(year, month, day, [hours, minutes, seconds, milliseconds])
-new Duration({days: 0, hours: 0, minutes: 0, seconds: 0, milliseconds: 0})
```
munificent: I think the Duration example is useful, though switching to use `=` would be good.

It shows an alternative API where it makes more sense to use named parameters. The idea is that DateTime should be positional because it's not meaningful to have a DateTime that specifies, say, an hour without specifying the day and month too. For Duration, though, it is reasonable to omit any of the parameters, which is why they are named.

src/_guides/language/effective-dart/design.md
stefanchrobot: Sure! I just copy-pasted it from here ;-)
https://api.dartlang.org/stable/1.21.1/dart-core/DateTime/DateTime.html

src/_guides/language/effective-dart/design.md
stefanchrobot: Ack. I shall submit the changes later this day.
I'll try to look for some other places in dart:core that would show the same ideas but are shorter.

PR #234, opened 2017-02-03
Changes for 1.22

kwalrath:
Add covariant & assert-with-message to the language tour.

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -86,7 +86,7 @@ apps:
 <div class="alert alert-info" markdown="1">
 **Note:**
 Our code follows the conventions in the
-[Dart Style Guide.](/guides/language/effective-dart/style).
+[Dart style guide](/guides/language/effective-dart/style).
 For example, we use two-space indentation.
 </div>
 
@@ -105,6 +105,10 @@ mind:
     clarifies your intent and enables static checking by tools, but it’s
     optional. (You might notice when you’re debugging your code that
     variables with no specified type get a special type: `dynamic`.)
+    Recent changes to the Dart language and tools
+    make writing type safe code possible,
+    with many expected benefits;
+    see [Sound Dart](/guides/language/sound-dart) for details.
 
 -   Dart parses all your code before running it. You can provide tips to
     Dart—for example, by using types or compile-time constants—to catch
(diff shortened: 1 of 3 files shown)


src/_guides/language/language-tour.md
```diff
@@ -158,38 +162,48 @@ mind:
...
 {:.table .table-striped .nowrap}
 
 <sup>1</sup> Words with the superscript **1**
-are *built-in identifiers*. Avoid using
+are **built-in identifiers**. Avoid using
 built-in identifiers as identifiers, and never use
 them for class or type names. Built-in identifiers exist to ease
 porting from JavaScript to Dart. For example,
 if some JavaScript code has a variable named `factory`, you don't have
 to rename it when you port the code to Dart.
```
eernstg: We could make an adjustment here: 'Avoid using built-in identifiers as identifiers. It is a compile-time error to use them for class or type names, and it is a source of confusion if they are used as names of variables or functions.'

The point is that the hard fact is stated as such, and the broader style rule about not using these names to declare other things is motivated. This will match the actual experience of developers who do it anyway.

src/_guides/language/language-tour.md
```diff
@@ -158,38 +162,48 @@ mind:
...
+added after Dart's 1.0 release.
 You can't use `async`, `await`, or `yield` as
 an identifier in any function body marked with `async`, `async*`, or `sync*`.
 For more information, see
 [Asynchrony support](#asynchrony-support).
 
-All other words in the keyword table are <em>reserved words</em>.
+<sup>3</sup> Words with the superscript **3**
+are newer, limited reserved words related to **type safety**
+added in 1.22 or later releases.
```
eernstg: `covariant` would simply be marked as a built-in identifier, so there should not be a need to introduce a new category of identifiers here, just use `{bii}` on covariant in the table.

src/_guides/language/language-tour.md
```diff
@@ -158,38 +162,48 @@ mind:
...
+Use `covariant` to modify method or field types, as described in
+[Overriding members](#overriding-members).
```
eernstg: I guess this reference to #overriding-members would not be needed: None of the other "more-or-less-reserved" words don't have their own links. The description in 'Overriding members' would still be there, of course, and the reader would presumably just search for 'covariant' and find it like that.

src/_guides/language/language-tour.md
```diff
@@ -2891,32 +2921,89 @@ class SmartTelevision extends Television {
...
+    // ...
   }
+  // ...
 }
 {% endprettify %}
 
-You can use the `@override` annotation to indicate that you are
-intentionally overriding a member:
+To narrow the type of a method parameter while maintaining type safety,
+use the `covariant` keyword.
```
eernstg: As I mentioned in the email message, `covariant` is useless in Dart 1. This document never says 'strong mode' anywhere, so maybe we shouldn't mention `covariant` here at all?

Otherwise, it should be described in a separate section about strong mode, and the only reference to `covariant` in the Dart 1 part should be `You can add the modifier `covariant` to parameter declarations in instance methods (including setters and operators); it has no effect. It is only allowed because it can make the same code usable in Dart 1 as well as in strong mode.'.

eernstg: Ah, of course we have to mention `covariant` in this document, because it is indeed a built-in identifier&mdash;so developers _will_ see an error if they try to define a typedef or a class with that name, so they do need to know about it.

But we only need to say that it has no effect, and that it only exists due to compatibility with strong mode.

src/_guides/language/language-tour.md
kwalrath: Done.

src/_guides/language/language-tour.md
kwalrath: I've drastically reduced the coverage of covariant. Now we just point to the Sound Dart docs.

I've removed the examples, so I'll close the PR that added them.

src/_guides/language/language-tour.md
kwalrath: Done.

src/_guides/language/language-tour.md
kwalrath: I've simplified the description.

kwalrath:
Thanks for the review! I've drastically reduced the coverage of `covariant` in the language tour as a result.

We might want to wait on committing these changes until 1.22 is released.

PR #236, opened 2017-02-05
Improve the function call example in Sample Code

stefanchrobot:
Cookies are delicious delicacies, but this went too far.


kwalrath:
@filiph maybe it's time for us to grow up?

kwalrath:
Thanks very much for the PR, @stefanchrobot!

PR #237, opened 2017-02-08
Add "brew cleanup dart" to Mac upgrade commands

alanhzhao:
Mac Spotlight will continue to see old versions of Dartium, so brew cleanup is useful to remove these


kwalrath:
Thanks!

PR #239, opened 2017-02-15
Update jetbrains-plugin.md

devoncarew:
- replace two links that had specific (outdated) IntelliJ versions with generic, forwarding links
- updated a link that had 404'd

A general question - do we have a link checker for this site?


kwalrath:
Thanks for the PR. We actually do have a link checker (e.g. see https://api.travis-ci.org/jobs/201604088/log.txt?deansi=true, which shows a single 404: www.jetbrains.com/idea/documentation/migration_faq.html), but a link failure doesn't break the build or trigger a nastygram. I'm not sure why not. I'll file a bug to look into this.

Issue #238, opened 2017-02-14
rename the .analysis_options files to analysis_options.yaml


kwalrath:
See https://github.com/dart-lang/sdk/issues/28385, https://github.com/google/stagehand/pull/362.

PR #241, opened 2017-02-15
Renaming .analysis_options to analysis_options.yaml

Sfshaza:
Fixes https://github.com/dart-lang/site-www/issues/238

@kwalrath 

I will take care of the code repos next.


src/_guides/language/analysis-options.md
```diff
@@ -47,10 +47,8 @@ add static analysis to your tool, see the
 
 ## The analysis options file
 
-Place the analysis options file at the root of the package,
-in the same directory as the pubspec file.
-By convention, the file is called `.analysis_options`, but it
-can also be named `analysis_options.yaml` (without the leading dot).
+Place the analysis options file, `analysis_options.yaml`,
```
kwalrath: We should have a note about this change, especially since `.analysis_options` files might be around for a while.

Basically help them understand (1) that they might see the . name and (2) that they should change the name if they do see it, since we expect support for it to go away.

Sfshaza:
I asked Dan R and he's not sure exactly when support goes away, so I didn't mention a release number.  Please review, @kwalrath.

src/_guides/language/analysis-options.md
```diff
@@ -47,10 +47,14 @@ add static analysis to your tool, see the
...
+at the root of the package, in the same directory as the pubspec file.
+
+<aside class="alert alert-info" markdown="1">
+**Note:** The `.analysis_options` file (note the leading dot) is no longer
+preferred and, at some point, will not be supported. Please change
+the file name to avoid problems in the future.
```
kwalrath: file name -> filename (could do this after committing)
but actually... we shouldn't assume they have files named .analysis_options. Maybe:

`<aside class="alert alert-warning" markdown="1">`
**Breaking change:** The conventional name for the analysis options file used to be `.analysis_options`
(note the leading dot and missing `.yaml` suffix). We expect support for the `.analysis_options` name to go away in a future release, so we recommend that you **change your `.analysis_options` files to `analysis_options.yaml`.**

Issue #245, opened 2017-02-20
Change wording in grey boxes


anders-sandholm:
Suggestion from @kasperl : I'd suggest changing two of the prominently featured grey boxes. 

Change:

  Dart webdev
  Build browser apps.

to:

  Web
  Build browser apps.

and:

  Dart VM
  Build scripts, servers.

to:

  Command-line
  Build scripts, servers.

Sfshaza:
@anders-sandholm @filiph @kwalrath 

This has been languishing because we didn't reach consensus. Should we schedule a meeting to discuss?

filiph:
Happy to meet!

FWIW I agree with the changes (sorry for taking so long to reply).

srawlins:
"Webdev" has changed to "Web", but "Dart VM" has not yet been changed to "Command-line"

chalin:
Has a consensus been reached?

The _What next?_ section of the [Get Started](https://www.dartlang.org/guides/get-started) page currently refers to the platforms as **_Mobile_**, **_Web_**, and **_Server_**:

> <img width="420" alt="screen shot 2018-06-24 at 11 25 21" src="https://user-images.githubusercontent.com/4140793/41820544-57cdeea2-77a1-11e8-9cf6-d9b56fdd8436.png">

Given the "client-optimized" focus for Dart, it seems most reasonable (and complementary) to refer to the other platform as _Server_ rather than just command-line.

Thoughts?

cc @kevmoo

anders-sandholm:
I'd like to propose - as I believe Patrice also suggests above - that we remain consistent (with the What next) and rename "Dart VM" to "Server" on the platform page as well (and keep the existing subtitle "Write a command-line app or server").

chalin:
Sounds good. Whatever terminology we settle on, we should be consistent across pages:

- [ ] [Get Started page](https://www.dartlang.org/guides/get-started)
- [ ] [Platforms page](https://www.dartlang.org/guides/platforms)
- [ ] Sidenav - all pages (when viewed from wide displays) other than the front page.

Should we be changing the `/dart-vm` URLS to be `/server` (with appropriate redirects, of course)?

anders-sandholm:
Happy for us to use 'Server' or 'server-side app' depending on context. 
Changing the URLs is not necessary (and higher risk, I guess), but probably a nice thing to do as well while we're at it.

chalin:
Closing this issue since it has been addressed. Followup actions will be tracked via #965.

PR #246, opened 2017-02-20
Changing "Dart VM" to "Command line" in several key places.

Sfshaza:
Fixes issue: https://github.com/dart-lang/site-www/issues/245

Should we go further than changing "Dart VM" on just the grey box on the front page? In this PR, I've changed "Dart VM" in the footer, side nav, and on the Dart VM home page, as well as the grey box. I changed "Web" in the footer, as well as the grey box.

Take a look and we'll discuss.

Staged: 
https://sz-www-1.firebaseapp.com/  (grey boxes)
https://sz-www-1.firebaseapp.com/dart-vm  (footer, side-nav, and title)

Btw: "command line" isn't generally hyphenated unless used as an adjective, for example, "the command-line interface", but "type the following at the command line."  

English...such a silly language.

@anders-sandholm  @kwalrath  @filiph

### DIFF
--- a/src/_data/side-nav.yml
+++ b/src/_data/side-nav.yml
@@ -52,7 +52,7 @@
       children:
       - title: "API Reference"
         permalink: http://docs.flutter.io/flutter/
-    - title: "Dart VM"
+    - title: "Command line"
       permalink: /dart-vm
       children:
       - title: "Install Dart"
--- a/src/_includes/page-footer.html
+++ b/src/_includes/page-footer.html
@@ -18,9 +18,9 @@
         <div class="content">
           <h4>Technologies</h4>
           <ul>
-            <li><a href="{{site.webdev}}">Dart webdev</a></li>
+            <li><a href="{{site.webdev}}">Web</a></li>
             <li><a href="{{site.webdev}}/angular">AngularDart</a></li>
-            <li><a href="{{site.dart_vm}}">Dart VM</a></li>
+            <li><a href="{{site.dart_vm}}">Command line</a></li>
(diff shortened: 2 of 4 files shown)


anders-sandholm:
Wauw, that was fast (and early?). 

Thinking some more about this I believe "AngularDart" qualifies to be in the grey box rather than "Web" or "Webdev".

Otherwise LGTM - but would like to also hear Filip's opinion because he has thought more about how we name things on the site.

Sfshaza:
Yeah, it's one of those days I woke up at 2am.  ;)

Thanks for that feedback!

I will wait for Filip and Kathy to weigh in on Tuesday. (Today is a holiday
in the U.S. - Presidents Day, which was created to celebrate Abraham
Lincoln and George Washington... ahem)

kwalrath:
I worry a bit that people will confuse "Command line" with command-line tools—like pub, dart2js, etc. The explanation helps, but the less people have to read, the better.

Another possibility is to say some variation of "Server side".

Or we could reconsider the gray boxes entirely. I think they're easy to miss, not parallel, would be better with icons, ... lots of stuff we could improve if it's a priority.

filiph:
Sorry for missing this. My 2 cents:

* "Command line" is a bit underselling, imho. Do we have space to say "Command line & server"? "CLI & server"? "Standalone"?
  * I agree that "Dart VM" is not very helpful to outside users, so happy for any change.
* I like "Web". "Angular" would be too forceful. I think we could add another gray box for AngularDart itself: "AngularDart: Build Angular apps".

Sfshaza:
I spoke to Filip yesterday and there's no resolution yet on the front page. I'm closing this for now.

PR #254, opened 2017-03-15
Fix code example for "Using generic methods"

ben4ever:
This fixes the following:
1. The proper syntax for the assign-if-null operator is `??=` instead of `?=`.
2. `??=` cannot be used in combination with declaring a variable (i.e. within the same statement). The reason for this might be because the variable is always `null` at the point of declaration anyway.

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -3373,7 +3373,8 @@ A newer syntax, called _generic methods_, allows type arguments on methods and f
 {% prettify dart %}
 [[highlight]]T[[/highlight]] first[[highlight]]<T>[[/highlight]](List<[[highlight]]T[[/highlight]]> ts) {
   // ...Do some initial work or error checking, then...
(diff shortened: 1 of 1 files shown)


kwalrath:
Is there a discussion of this change somewhere? I'd like to know more about it.

ben4ever:
Sorry, I didn't provide any information in my first comment. I just updated the comment now.

kwalrath:
Looking at the [dartpad](https://dartpad.dartlang.org/a02c53b001977efa4d803109900f21bb) referred to just above the sample code, it should indeed be `=` instead of `?=`.

Good catch! Would you mind changing the PR?

ben4ever:
Ah right, didn't even notice that link to the dartpad example.

I simply changed the `?=` to `=` now. Or did you want me to include the
```dart
  if (!(ts?.length > 0)) {
    // ts is null or empty
    return null;
  }
```
part as well for a more extensive example?

kwalrath:
How you have it now is good. Thanks very much!

Issue #249, opened 2017-03-10
Add redirects for most common 404s


kwalrath:
Add redirects for these pages:

* /docs/dart-up-and-running/contents/ch01
* /articles/mocking-with-dart (and/or edit the referrer, http://stackoverflow.com/questions/20477587/mocking-with-dart-how-to-test-that-a-function-passed-as-a-parameter-was-called)
* /editor

kwalrath:
https://www.dartlang.org/articles/mocking-with-dart still gives a 404. (But the stack overflow article has been updated, so that's good.)

PR #263, opened 2017-04-01
Add redirects.

Sfshaza:
Fixes https://github.com/dart-lang/site-www/issues/249

Please review, @kwalrath

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -452,7 +452,7 @@
       },
       {
         "source": "/docs/dart-up-and-running/contents/ch02",
-        "destination": "/resources/books",
+        "destination": "/guides/language/language-tour",
         "type": 301
       },
       {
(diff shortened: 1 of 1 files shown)


firebase.json
```diff
@@ -446,6 +451,11 @@
         "type": 301
       },
       {
+        "source": "/docs/dart-up-and-running/contents/ch02",
+        "destination": "/resources/books",
```
kwalrath: Shouldn't the destination be /guides/language/language-tour, just like the next redirect?

kwalrath:
Thanks for doing this! Have you staged this?

firebase.json
filiph: pinging @Sfshaza

Sfshaza:
Thanks for reminding me about this, @filiph!

PR #269, opened 2017-05-05
Update analysis-options.md

devoncarew:
Update our docs for the analysis options file:

- make a bit more clear the distinction between ignoring a rule and changing its severity
- also call out that you can use the same mechanism to change the severity of lints as well


Sfshaza:
LGTM!

PR #286, opened 2017-05-24
chore(firebase serve): use `FIREBASE_TOKEN` and `--project` option

chalin:
As firebase/firebase-tools#302 explains, starting with firebase-tools 3.6, firebase serve will require authentication credentials.

To be able to serve locally, one needs a token, _and_ a project must be specified, without these the local serve fails, as is shown in [this Travis log](https://travis-ci.org/chalin/site-www/builds/235775690):
> Error: No project active, but project aliases are available.

The corresponding `site-webdev` change was done via https://github.com/dart-lang/site-webdev/pull/654.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,11 @@ cache:
   bundler: true
   directories:
   # - $HOME/[path]
-rvm:
-- 2.2.3
+rvm: [2.3.0]
+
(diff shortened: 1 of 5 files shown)


chalin:
cc @kevmoo @filiph

PR #287, opened 2017-05-25
Document "// ignore_for_file" in analysis_options.md

davidmorgan:
I wasn't quite sure the best way to document this -- there are now a lot of ways of suppressing rules! Suggestions welcome.

This is a new addition -- do we need to call out which SDK version it's available from?

Thanks.

### DIFF
--- a/src/_guides/language/analysis-options.md
+++ b/src/_guides/language/analysis-options.md
@@ -257,6 +257,21 @@ analyzer:
 This analysis options file instructs the analysis tools to ignore
 the TODO rule.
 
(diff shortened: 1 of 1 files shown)


Sfshaza:
This looks great, @davidmorgan! Thanks for submitting a PR!

Yes, can you please add the version of the SDK when this became available? Something along the lines of:

```
Alternatively, as of Dart 1.23 you can ignore a specific rule for a specific file using
an `ignore_for_file` comment:
```

Otherwise, lgtm

davidmorgan:
Done.

PR #307, opened 2017-06-21
Adding function-type syntax to the language tour.

Sfshaza:
Fixes https://github.com/dart-lang/site-www/issues/302

Staged:
https://sz-www-1.firebaseapp.com/guides/language/language-tour#functional-types
https://sz-www-1.firebaseapp.com/guides/language/language-tour#typedefs-and-generic-functions

Please review @floitschG @kwalrath

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -1049,7 +1049,7 @@ fields.  To do this, replace a function's name with `Function`
 in its declaration. For example, the type of `void foo() {}`
 would be `void Function() {}`.
 This syntax was originally added to [refer to types in the typedef of a
-generic method](#generics-and-generic-functions), but can be used
+generic method](#typedefs-and-generic-functions), but can be used
 wherever a type can be written. For example:
 
 {% prettify dart %}
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -1042,6 +1042,26 @@ src="{{site.custom.dartpad.embed-dart-prefix}}?id=d988cfce0a54c6853799&horizonta
 </iframe>
 {% endcomment %}
 
+### Functional types
+
+Dart allows you to specify function types in places like variables and
+fields.  To do this, replace a function's name with `Function`
+in its declaration. For example, the type of `void foo() {}`
+would be `void Function() {}`.
+This syntax was originally added to [refer to types in the typedef of a
+generic method](#generics-and-generic-functions), but can be used
+wherever a type can be written. For example:
+
+{% prettify dart %}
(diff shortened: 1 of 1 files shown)


src/_guides/language/language-tour.md
```diff
@@ -1042,6 +1042,26 @@ src="{{site.custom.dartpad.embed-dart-prefix}}?id=d988cfce0a54c6853799&horizonta
 </iframe>
 {% endcomment %}
 
+### Functional types
+
+Dart allows you to specify function types in places like variables and
```
chalin: A statement like

> Dart allows you to specify function types in places like variables and fields.

only really makes sense under the implicit context of comparing the original (<1.24) anonymous function-type syntax with the new.

Dart has always had syntax for anonymous function-types, but the new syntax has no restrictions (as far as I can tell): you can use the new anonymous function-type syntax _anywhere_ a type annotation can be written. That is not the case with the original syntax which, in particular, cannot be used to annotate a global variable, local variable, or a class field.

IMHO, to be complete this new section should document both the new and the original syntax (in case the reader comes across the original syntax --- it will take some time before existing code is rewritten to use the new syntax, if at all).

How about opening this section something like this:

> ### Anonymous function-types
>
> In Dart, you can use an anonymous function-type as a type annotation. 
> Here are some examples:
> 
> [Example of, say, a callback function parameter using the original syntax, and a global variable using the new syntax, followed by a statement saying that both syntaxes will be explained next]
>
> You can write the anonymous type corresponding to any function.
> To do this, replace a function's name with `Function` in its declaration. For example,
> the type of `void foo() {...}` can be written as`void Function()`. ...
> This `Function`-keyword based syntax can be used anywhere a type annotation is permitted.
>
> In contrast, the original anonymous function-type syntax ...

src/_guides/language/language-tour.md
```diff
@@ -3854,6 +3874,32 @@ main() {
...
+This syntax uses an assignment operator in conjunction with
+`Function()` in the declaration.
+
+Compare the following typedef statements:
+
+|--------------------------+------------------------------------------------|
+| Typedef example          | Description                                    |
+|--------------------------|------------------------------------------------|
+| <code>typedef Foo<T>(T param); | A function that takes `dynamic`. |
+| <nobr><code>typedef Foo = Function<T>(T param); | A generic function that types some type parameter T, and a parameter of that type. |
```
chalin: The descriptions will need to be revisited. I have some comments and questions, but I'll post them to the original issue (https://github.com/dart-lang/site-www/issues/302).

Sfshaza:
OK, after ignoring (vacay followed by Dart 2.0 work), then thought, then group discussion, I'm going a different way on this. A new PR will be forthcoming.

PR #309, opened 2017-06-22
Doing some requested cleanup

Sfshaza:
Removing two optional types articles and renaming Sound Dart docs to Strong Mode docs.

@dgrove @kevmoo @kwalrath @filiph


kwalrath:
I'll need to update webdev to match.

Sfshaza:
I have a PR for webdev, too.

PR #310, opened 2017-06-23
Cleaning up the FAQ

Sfshaza:
Please review, @wmleler @filiph 

I'm about to start an email thread where we can discuss this in further detail.

Staged: https://sz-www-1.firebaseapp.com/faq

### DIFF
--- a/src/faq.md
+++ b/src/faq.md
@@ -75,6 +75,19 @@ about the VM question in more detail.
 
 ## Language
 
+{% comment %}
+### Q. Isn't Dart a lot like Java?
+
+TODO: provide an up-to-date answer
+{% endcomment %}
+
+### Q. How does Dart relate to Go?
+
+Dart and Go are both language projects started at Google, but they
+are independent and have different goals. As a result,
+they make different choices, and the languages have very different
+natures, even while we all try to learn from each others' work.
+
 ### Q. Why isn't Dart more like Haskell / Smalltalk / Python / Scala / other language?
 
 Various reasons, depending on the language being asked about.
@@ -82,7 +95,7 @@ Various reasons, depending on the language being asked about.
 For languages that are quite different from JavaScript: it's important for Dart
 to compile to efficient JavaScript.  Our experience in GWT is that if the source
 language is too different from JavaScript, it creates some cases where complex
-output code is needed to emulate the source language's behavior.  This can cause
+output code is needed to emulate the source language's behavior. This can cause
 performance to vary in ways that are not transparent to the programmer.
 
 For languages that are less mainstream: we expect that modeling Dart on these
@@ -94,6 +107,14 @@ think we've got our hands full introducing a new language at all.
 For languages that are "more dynamic" than Dart: Dart deliberately trades off
 some of this arbitrary runtime modification for the goal of better performance
 and tools.
+
+Q. Why isn't Dart syntax more exciting?
+
+We did throw in some nice syntactic features such as `this.` constructor args
+and `=>` for one-line functions, but we'd agree that Dart chooses
+familiarity over excitement. One team member's personal testimonial:
+"I wish it had a little more razzle dazzle but I can't deny that
+literally on my first day of writing Dart code, I was productive in it."
 
 ### Q. Does Dart have reflection capabilities?
 
@@ -129,6 +150,11 @@ Dart is a type-safe language if you opt in to [strong mode].
 Strong mode enables a sound static type system and is used by
 [Flutter] (a cross-platform mobile development system),
 and the Dart Development Compiler ([DDC]).
+
+Even in strong mode, though, you can annotate any variable with
+`dynamic` if you need the flexibility of a dynamic language.
+`dynamic` can contain any type. Of course, that removes many of
+the benefits of a type-safe language for that variable.
 
 Without strong mode enabled, Dart is not a type-safe language.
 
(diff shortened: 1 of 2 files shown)


src/faq.md
```diff
@@ -129,39 +95,6 @@ For languages that are "more dynamic" than Dart: Dart deliberately trades off
 some of this arbitrary runtime modification for the goal of better performance
 and tools.
 
-### Q. Why isn't Dart syntax more exciting?
```
filiph: I think this question can stay.

src/faq.md
```diff
@@ -190,55 +123,23 @@ appreciation through careful design choices and fair communication about them.
...
-has also shown that full statically typed languages are sometimes too rigid,
-and we wanted Dart development to be more flexible for a wide range of
-developers.
+### Q. Is Dart a type-safe language?
 
-### Q. Why is the type system designed to be unsound?
+Dart is a type-safe language if you opt in to [strong mode].
+Strong mode enables a sound static type system and is used by
+[Flutter] (a cross-platform mobile development system),
+and the Dart Development Compiler ([DDC]).
```
filiph: Suggestion (so that people still know there's dynamic as an opt-in):

"Even in strong mode, though, you can annotate any variable with `dynamic` if you need the flexibility of a dynamic language. `dynamic` can contain any type. Of course, that will remove many of the benefits of a type-safe language for that variable."

src/faq.md
```diff
@@ -297,26 +193,7 @@ For more information, see [Dart Tools].
...
-Yes! See
-[Dart and Google Cloud Platform]
-for details.
-
-### Q. What CPU architectures does the Dart VM support?
-
-The Dart VM works on IA-32, x64, MIPS, ARMv5TE, ARMv6, ARMv7, and
-ARM64 processors.
-
-### Q. What are some real-world production deployments of Dart?
```
filiph: Maybe leave this and link to who uses dart, plus point out, for example, the news about AdWords, AdSense or AdMob. Maybe even say that 75%+ Google revenue flows through these apps? 

That _is_, by the way, a very frequently asked question.

src/faq.md
```diff
@@ -75,40 +75,6 @@ about the VM question in more detail.
...
-
-* Java supports public, protected, package protected, and private. Dart supports
-public and library-private.
-
-* Dart is purely object oriented. Java has objects and primitives.
-
-* The Dart language has mixins, optional static types, named parameters, and
-more.
-
-### Q. How does Dart relate to Go?
```
filiph: This could stay, no?

src/faq.md
```diff
@@ -75,40 +75,6 @@ about the VM question in more detail.
 
 ## Language
 
-### Q. Isn't Dart a lot like Java?
```
filiph: Let's add a TODO to explain the current differences from Java. It's still a valid question to ask, and there definitely are differences — if nothing else, the terseness and the more developer-friendly type system.

Sfshaza:
Thanks for your feedback, @filiph! 

Staged: https://sz-www-1.firebaseapp.com/faq

src/faq.md
```diff
@@ -129,38 +108,13 @@ For languages that are "more dynamic" than Dart: Dart deliberately trades off
...
-development and sometimes at run time.
-
-So, for example, Dart isn't likely to support evaluating a string as code in the
-current context, but it may support loading that code dynamically into a new
-isolate.  Dart isn't likely to support adding fields to a value, but it may
-(through a mirror system) support adding fields to a class, and you can
-effectively add methods using `noSuchMethod()`.  Using these features
-will have a runtime cost; it's important to us to minimize the cost for programs
-that don't use them.
+Q. Why isn't Dart syntax more exciting?
```
filiph: This lacks the `###` at the start, so it's not a heading.

src/faq.md
```diff
@@ -297,26 +219,13 @@ For more information, see [Dart Tools].
...
+using [Flutter].
 
 ### Q. What are some real-world production deployments of Dart?
 
-The [Dart 1.0 announcement][announcement] lists a few, like
-internal Google apps, external Google apps,
-Mixbook, blossom.io, Soundtrap, Mandrill, and more.
-Many more have been released or are in development.
-See [Who Uses Dart] for more.
+Google AdWords, AdSense, and AdMob all use Dart.  In fact,
```
filiph: [nit] not sure about the "in fact" — we could drop it

Sfshaza:
Thanks, @filiph. Both items addressed.
Staged: https://sz-www-1.firebaseapp.com/faq

kwalrath:
LGTM. I say ship it!

src/faq.md
wmleler: (Changes to Filip's addition) A _sound_ type system is not the same as a _static_ type system, so you can still have the benefits of dynamic languages. Even in strong mode, in Dart you can declare any variable as `dynamic`. While you get the benefit of a dynamic language, the tradeoff is less type safety for that variable. But the decision is yours. For example, you might build a prototype with many dynamic types, but then go through and specify more specific types before you release that code. Dart also supports _type inferencing_, to reduce the amount typing information you have to supply.

wmleler:
Dang, did my comments miss out on getting into this?

src/faq.md
```diff
@@ -190,55 +144,28 @@ appreciation through careful design choices and fair communication about them.
...
-
-We want to maintain the feel of a dynamically typed language, which is familiar
-to web developers. Mandatory types don't fit with that goal. Experience
-has also shown that full statically typed languages are sometimes too rigid,
-and we wanted Dart development to be more flexible for a wide range of
-developers.
+### Q. Is Dart a type-safe language?
 
-### Q. Why is the type system designed to be unsound?
+Dart is a type-safe language if you opt in to [strong mode].
```
wmleler: Technically, type safety is not actually the same thing as type soundness. And indeed, Dart is not actually type safe. And just to confuse things, "strongly typed", although it is frequently used, doesn't have a good definition, so "strong mode" can be a bit confusing.

src/faq.md
```diff
@@ -190,55 +144,28 @@ appreciation through careful design choices and fair communication about them.
...
+Strong mode enables a sound static type system and is used by
```
wmleler: Would be nice to have a short definition of soundness, like "a sound type system guarantees that an expression on one type cannot produce a value of another type" (_no surprises_!). Here's a reasonable link talking about soundness -- http://papl.cs.brown.edu/2014/safety-soundness.html

PR #313, opened 2017-06-29
Avoid firebase deploy issue

kwalrath:
See dart-lang/site-webdev/pull/788 for more information.

### DIFF
--- a/deploy/deploy-firebase.sh
+++ b/deploy/deploy-firebase.sh
@@ -3,4 +3,4 @@
 : ${FIREBASE_PROJECT:=default}
 
 echo "================ Deploy to Firebase ($FIREBASE_PROJECT) ========================"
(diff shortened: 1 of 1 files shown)


kwalrath:
Oops, wait, I put the args in the wrong order.

kwalrath:
Deploy failed: https://travis-ci.org/dart-lang/site-www/builds/248489011

From @chalin:

> Two issues with site-www:
> 
> (1) Liquid Warning: Liquid syntax error (line 14): Expected id but found pipe in "{{ book.authors || array_to_sentence_string }}" in resources/books.md

(We've had this for a while, but haven't gotten around to fixing it because it never seemed to break anything.)

> (2) Deployment is still failing, with error message:
> Your CLI authentication needs to be updated to take advantage of new features.
> ⚠ Please run firebase login --reauth
> ...
> I think it might be a node version issue.

(At first, Patrice thought we might need a new token, but it turns out he updated the token for site-www at the same time as the one for site-webdev (on May 24).)

PR #324, opened 2017-07-12
Add link to install page to side nav

chalin:
While the site landing page has an "Install Dart" button, no other page
has a link to the "Install Dart" page. This minor PR make it easier to gain access to
the Install Dart page from anywhere on the site.

cc @Sfshaza

### DIFF
--- a/src/_guides/get-started.md
+++ b/src/_guides/get-started.md
@@ -20,7 +20,7 @@ src="{{site.custom.dartpad.embed-dart-prefix}}?horizontalRatio=70&verticalRatio=
 Note that DartPad supports only a few core libraries.
 If you want to use other libraries,
 such as dart:io or libraries from packages,
-you'll need to install an SDK.
+you'll need to [install](/install) an SDK.
 
 
 <div class="get-started-table__text">


src/_data/side-nav.yml
```diff
@@ -6,6 +6,8 @@
       permalink: /samples
     - title: Codelabs
       permalink: /codelabs
+    - title: Install Dart
+      permalink: /install
```
kwalrath: Thanks for doing this!

I'm thinking this should be under TOOLS instead. (A 3-item QUICK START is already long.)

Sfshaza: I can also see it under TOOLS. Otherwise, it looks good to me.

filiph:
I slightly prefer putting this under TOOLS as well.

Another option is to rename TOOLS in the sidebar to "INSTALL" and below it, have "Dart SDK" and "Tools".

chalin:
Let's call the current side-nav organization solution (A).

I juggled with some of the other options you've suggested but none of them seem quite right. E.g.
- (B) The Dart SDK isn't a tool, so it doesn't seem right to put it under "TOOLS".
- (C) @filiph, I like your suggestion of changing the heading to "INSTALL", but then there is a bit of a clash since the "Tools" page describes the tools rather than the steps to follow to install them.

(D) Another option might be to swap the "quick start" and "get started" heading to have, e.g.:
- **GET STARTED**
  - [Quick Start](#)
  - [Sample Code](#)
  - [Codelabs](#)
  - [Install](#)

(E) Maybe, most simply, we could just rename the first heading to something like:
- **TRY DART**
  - [Get started](#)
  - [Sample Code](#)
  - [Codelabs](#)
  - [Install](#)

(F) Finally, we could rework the side-nav code to allow headings to be links. Currently we have headings with single children (like TOOLS, TESTING and COMMUNITY) that could be coalesced (effectively migrating the child link to the heading and eliminating the child item). Under this organization we could have INSTALL as a top level heading / entry.

My preferred quickfix is (E), with (F) as a longer-term solution. WDYT?

filiph:
Thanks, Patrice, for the write-up! FWIW, I don't like (F) because it would mean that a) we are inconsistent, as some sidenav headings are links and some aren't (which is really confusing in a sidenav), or b) we force "landing pages" for all the headings (which is a lot of transactional content).

D and E has the problem of giving too many options — which is not great for new users. Especially since 'Install' is already included in the various 'Get started' pages (for web, Flutter, standalone), and over there the install steps are tailored to the target.

So let's circle back. Which users are asking for the additional link? Do people get lost trying to install Dart when getting started? Or is this more for existing users coming back to Dart and trying to re-install (e.g. on a new computer)? If so, where do these people land if not on the homepage?

cc @lukechurch

chalin:
Thanks for the feedback Filip.

> since 'Install' is already included in the various 'Get started' pages (for web, Flutter, standalone)

Right, but the "Get started" page on _this_ site does not currently contain a link to the Install page. One of the changes in this PR is to rectify that (sub-)issue.

If we can't agree on how to update the side-nav, I'd be ok to add a link only to the "Get started" page.

chalin:
Funny thing, I just noticed that there already _is_ a link to the Install page in the sidenav: it is tucked under the PLATFORMS >> Dart VM submenu.

So, I've simplified this PR. It now only impacts the Get Started page, hyperlinking the word "install" that was already there, so that it can lead the reader to the install page.

Are we ok with this simplified change.

(I have more comments about the sidenav, but I'll create an issue for those.)

chalin:
I know that there is a lot going on atm, and that this isn't really important (so that we don't need to deal with it now), but here is a followup issue: #327

chalin:
@kwalrath: is this simplified PR good to go?

Issue #328, opened 2017-07-14
Rename "Dart VM" to "Server" in sidenav and elsewhere


chalin:
We'll probably want to adjust permalinks (`/dart-vm` to `/server`) and add the corresponding firebase redirect rules.

cc @kwalrath @Sfshaza

Sfshaza:
Yes, that makes sense.

kwalrath:
Done.

PR #329, opened 2017-07-14
reorg/simplify sidenav, better matching webdev

### LINKED ISSUE #327: simplify sidenav
Might we want to reorganize the site-www sidenav to better match the webdev sidenav. Here are snapshots:

site-www (current) | site-webdev | site-www (proposed)
-- | -- | --
screen shot 2017-07-13 at 1 38 06 pm |  screen shot 2017-07-13 at 1 07 04 pm | screen shot 2017-07-13 at 1 35 01 pm

This is in part a followup to the discussions of #324.

cc @filiph @kwalrath @Sfshaza

chalin:
This fixes all aspects mentioned in #327, _except_ for Kathy's suggested rename of "Dart VM" to "Server" because that will require quite a bit of work both to the sidenav and permalinks. I've created a separate issue to track that (#328).

Staged at https://dartlang-org-dev.firebaseapp.com

Fixes #327

### DIFF
--- a/src/_data/side-nav.yml
+++ b/src/_data/side-nav.yml
@@ -1,8 +1,3 @@
-- title:
-  children:
-  - title: Get Started
-    permalink: /guides/get-started
-
 - title: Language
   children:
     - title: "Overview"
--- a/src/_data/side-nav.yml
+++ b/src/_data/side-nav.yml
@@ -15,6 +15,8 @@
           permalink: /guides/language/effective-dart/usage
         - title: Design
           permalink: /guides/language/effective-dart/design
+    - title: Sample Code
+      permalink: /samples
 
 - title: Libraries
   children:
@@ -55,8 +57,6 @@
     permalink: /tutorials
   - title: "Articles"
     permalink: /articles
-  - title: Sample Code
-    permalink: /samples
   - title: "Dart Tools"
     permalink: /tools
   - title: "Community and Support"
--- a/src/_data/side-nav.yml
+++ b/src/_data/side-nav.yml
@@ -1,11 +1,7 @@
-- title: Quick Start
+- title:
   children:
-    - title: Get Started
-      permalink: /guides/get-started
-    - title: Sample Code
-      permalink: /samples
-    - title: Codelabs
-      permalink: /codelabs
+  - title: Get Started
+    permalink: /guides/get-started
 
 - title: Language
   children:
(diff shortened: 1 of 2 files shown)


Sfshaza:
I like the top of the new side nav, with Get Started by itself, though Samples are one of our most heavily used pages and it is now buried - I had to scroll to see it. It used to be that the tabs at the top matched the categories in the side nav, and this PR breaks that. I looked and saw that the tabs and categories no longer match on webdev either, so maybe it's not that important.

So, yes, I'm OK with it.

filiph:
If you can somehow add the "Get Started" link to have classes "btn btn-primary", it will look like this:

screen shot 2017-07-14 at 13 12 10

Source:

```
<ul>
    <li>
      <a href="/guides/get-started" class="btn btn-primary" title="Get Started">Get Started</a>
    </li>
</ul>
```

If you like this, we might want to hardcode that button there (and remove the link from sidenav.yml).

Sfshaza:
Oh, I like that, @filiph!

filiph:
Oh wait, but that will break the 'currently active' functionality of the sidenav. So we probably want to have it through sidenav.yml, not hardcoded.

filiph:
Here's how it would look like with "btn-default" instead of "btn-primary", fyi. I think I like btn-default more.

screen shot 2017-07-14 at 13 17 36

chalin:
Following Filip's suggestion, the last commit makes "Get Started" a btn-default. When the page is active, the "active" class is applied to the button, to look like this:

When viewing Get Started | When viewing another page
-- | --
screen shot 2017-07-15 at 11 24 07 am | screen shot 2017-07-15 at 11 24 53 am

The staging server has been updated.

If we agree on this change, let me know if you'd like the same change done to webdev.

chalin:
@filiph - Any feedback about the [latest iteration on this PR](https://github.com/dart-lang/site-www/pull/329#issuecomment-315542046)?

@Sfshaza - About the Samples link. What if we put it under LANGUAGE?  It seems like an appropriate heading, and that would put it at the top, like this:
> - [Get Started]
> - LANGUAGE
>   - Overview
>   - Tour
>   - Effective Dart >
>   - Sample Code

Sfshaza:
Yes, to putting samples under language.  thx

chalin:
Sample link moved. See the updated staging server: Staged at https://dartlang-org-dev.firebaseapp.com/guides/get-started

@kwalrath - is this approved?
Also, pls confirm whether you think that we should do the same for webdev.

kwalrath:
> Also, pls confirm whether you think that we should do the same for webdev.

Yes, let's keep the navigation as parallel as possible.

PR #330, opened 2017-07-14
Add a CONTRIBUTING.md file

kwalrath:
The site-www version of dart-lang/site-webdev#820.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,26 @@
+# Contributing :heart:
+
+Thanks for thinking about helping with [www.dartlang.org][www]!
+Here are a couple of ways that you can contribute:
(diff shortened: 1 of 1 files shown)


LarkAscending:
It looks good - one small piece of feedback is that at least on the Flutter team, reviewers have flagged comments for removal before merge. This is an informal practice, not written down to my knowledge (and maybe doesn't apply to content in this repo).

kwalrath:
> It looks good - one small piece of feedback is that at least on the Flutter team, reviewers have flagged comments for removal before merge. This is an informal practice, not written down to my knowledge (and maybe doesn't apply to content in this repo).

@LarkAscending interesting! I'll follow up on this separately...

PR #334, opened 2017-07-18
No more Dartium downloads after 1.24

### LINKED ISSUE #331: install/archive page should be changed for Dartium versions
There will no longer be Dartium versions on the dev channel, and the stable 1.24 channels should fetch Dartium with version numbers 1.25.0-dev.6.x, as x increases.  Even though we are releasing 1.25.0-dev.7.0, and perhaps further, before 2.0, Dartium builds of those versions will not be released.

Perhaps the page should just be changed to have a separate stable Dartium section, and the existing stable and dev sections should work unchanged, just not including Dartium builds.

@terrylucas @whesse @Sfshaza @kwalrath

kwalrath:
Also made the code strong mode compliant. (dartdevc successfully compiles it.)
If we have any Dartium dot releases we'll need to update the text & code again.

Staged: https://kw-www-dartlang-1.firebaseapp.com/install/archive

Fixes #331.

### DIFF
--- a/src/install/archive/web/download_archive.dart
+++ b/src/install/archive/web/download_archive.dart
@@ -2,6 +2,8 @@ import 'dart:async';
 import 'dart:convert';
 import 'dart:html';
 
+import 'package:pub_semver/pub_semver.dart';
+
 const String storageApiBase =
     "https://www.googleapis.com/storage/v1/b/dart-archive/o";
 const String storageBase = "https://storage.googleapis.com/dart-archive";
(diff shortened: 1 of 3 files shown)


src/install/archive/web/download_archive.dart
```diff
@@ -95,9 +97,9 @@ Future getListing(String channel, String respString) async {
   // Format is lines of "channels/stable/release/\d+/".
   Iterable<Future> versionRequests = versions.map((String path) async {
     try {
+      // XXX: We get 403/404s for dev/release/1.XY.0 where XY > 16.
```
whesse: I know why this is.  We were requested to copy the stable versions to the dev channel, so that the dev channel would have 1.x.0-dev.foo followed by 1.x.0 followed by 1.x+1.0-dev.foo.  And when copying the files over, I think we missed giving the READ: ALLUSERS ACL to the new copy.  We could fix this.

kwalrath: It'd be great to fix this, since it results in console errors.

kwalrath:
> LGTM. Does this assume that new versions of Dartium will be put into the stable patch versions of 1.24, and named with the 1.24.x version of that patch version? This would not actually be correct, since the version they would report internally would be 1.25.0-dev.6.y. But if this is how we have to do it, OK.
I'm willing to change this code to whatever makes sense. Just didn't want to prematurely gum it up for something that might not happen...

PR #340, opened 2017-07-27
Adding a new page for 2.0, removing some files, and other cleanup.

Sfshaza:
I've added the Dart 2.0 page I was telling you about and deleted a couple of articles. Please review!

Staged: 
FAQ:  https://sz-www-1.firebaseapp.com/faq
Dart 2.0:  https://sz-www-1.firebaseapp.com/dart-2.0

Thanks!


Sfshaza:
@wmleler, please review the new page. Thx!

src/dart-2.0.md
```diff
@@ -0,0 +1,95 @@
...
+
+---
+
+You can implement these changes now,
+but keep watching this page for updates. We'll let you know what you can
+do now, and what needs to wait for Dart 2.0.
+
+## Changes in Dart 2.0
+
+Dart 1.x is an optionally typed language, but Dart 2.0 requires static types.
```
sethladd: you can still use dynamic with d2. Is it accurate to say "requires static types" ?

We probably want to mention dynamic, type inference, and sound type system.

e.g. this is still valid: "var x = 1;"

src/dart-2.0.md
```diff
@@ -0,0 +1,95 @@
...
+Thanks to [strong mode](/guides/language/sound-dart),
+the Dart analyzer can find bugs earlier in the development cycle.
```
sethladd: ", and types are checked at runtime" ?

src/dart-2.0.md
```diff
@@ -0,0 +1,95 @@
...
+The analyzer is smart about types and can sometimes
+[infer](/guides/language/sound-dart#type-inference) a type,
```
sethladd: I believe the runtimes also infer. (or, will, when they implement d2)

src/dart-2.0.md
```diff
@@ -0,0 +1,95 @@
...
+so you don't have to manually type every variable.
+
+Migrate your code to strong mode now, and you may find some bugs in the
+process! You'll also find it much easier to transition to Dart 2.0.
```
sethladd: maybe more "may identify some lurking bugs that will now be more easily identifiable and addressable" :) . -- a bit more "phew, found that sneaky bug before it hurt users" :)

src/dart-2.0.md
```diff
@@ -0,0 +1,95 @@
...
+For more information, see [A stronger Dart for
+everyone](http://news.dartlang.org/2017/06/a-stronger-dart-for-everyone.html)
+and [Strong Mode Dart.](/guides/language/sound-dart)
+
+Flutter is already strong mode compliant, which contributes to Flutter's
+fast development cycle. If you develop on Flutter, you are good to go.
```
sethladd: not sure that we'd draw a line between strong mode and fast development cycle. it's probably more accurate to say "which has helped Flutter build their entire framework and widgets with Dart at scale"

src/dart-2.0.md
```diff
@@ -0,0 +1,95 @@
...
+  for faster compilation speeds with dartdevc. Place library code that won't be
+  _directly_ imported by your `web` code, or by other packages,
+  under `lib/src`. You can do this now. The
+  [Stagehand](https://github.com/google/stagehand/tree/master/templates)
+  templates have already been updated to this directory structure.
+
+### Mobile (Flutter)
+
+The Flutter SDK is already strong mode compliant.
+We have no other updates at this time.
```
sethladd: maybe add something like "and developers using Flutter should already be opted-into strong mode"

src/dart-2.0.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Servers and command line
+
+Strong mode Dart is available as of Dart 1.15.
+To use strong mode in Dart 1.x,
+[you must enable it.](/guides/language/sound-dart#how-to-enable-strong-mode)
+
+We have no other updates at this time.
+
+## Removed from Dart 2.0
```
sethladd: love this. thank you!

src/dart-2.0.md
wmleler: I agree. It is very confusing (even if it is somewhat technically accurate) to say that D2 "requires static types". What is confusing is that D2 does not require the user to specify these static types if they can be determined automatically through type inference. As Seth points out, "var x = 1;" is valid because D2 determines that x is an integer and (statically) types it as such. Languages like Java and Pascal require types to be explicit and require more work by the programmer, while D2 does not.

Also confusing is the fact that in D2 the type "dynamic" is a static type. The user can declare a type (or it can be inferred) as dynamic. That declares (statically) that the type can be dynamic, so it can actually be any type at runtime. Likewise, in D2 a num can be either an int or a double at runtime. Again, a static type that behaves like a dynamic type at runtime.

The bottom line is that we need to make it clear that while D2 is statically typed, the provision of type inference and types (like dynamic and num) that behave in dynamic ways at runtime, takes much of the burden of specifying types away from the developer. You could even say that D2 offers the advantages of a strongly typed language, while only requiring slightly more work on the part of the programmer (compared to D1).

Even further, D2 has a sound type system, which is even better than strongly typed (a statically typed language is not necessarily sound, One famous example is the C language, which didn't distinguish a pointer from an int or an array, among other things).

I can help wordsmith this if you want.

BTW, this is a short but good overview of typing -- http://blog.steveklabnik.com/posts/2010-07-17-what-to-know-before-debating-type-systems

sethladd: The big change in Dart 2 is that var changes from "dynamic" to "infer me!". This is a HUGE change.

this is valid in Dart 2:

dynamic x = 1;
x = "String";

This is invalid in Dart 2:

var x = 1;
x = "string";

See https://dartpad.dartlang.org/c7d9bb2240ffaba75122a6347a096d29

src/dart-2.0.md
kevmoo: True! If you have an arg of type dynamic and you try to assign it to a `String foo = arg;` you'll get runtime checks

src/dart-2.0.md
kevmoo: Yup! Same inference semantics everywhere. Is nice!

src/dart-2.0.md
kevmoo: +1

src/dart-2.0.md
```diff
@@ -0,0 +1,95 @@
...
+
+The following features are removed from Dart 2.0:
+
+Checked mode
+: Strong mode replaces checked mode. To learn how they are different, see
+  [What is the difference between strong mode and checked
+  mode?](/guides/language/sound-faq#how-is-it-different-than-checked-mode)
+
+Dartium
+: The dartdevc compiler performs fast compiles to JavaScript,
```
kevmoo: Do we have an article about getting started w/ dartdevc? Or is that coming later?

kwalrath:
My general feedback is that this page should be more scannable. A web dev, flutter dev, or command-line dev should be able to easily see exactly what they have to do, all in one place.

kwalrath:
Another bit of feedback: It shouldn't be too hard for devs to see what's changed since the last time they looked. That doesn't matter this time around, but let's think about what people will want to see once we've added an item that matters to them.

src/_guides/language/analysis-options.md
```diff
@@ -301,7 +301,7 @@ Join the discussion list for linter enthusiasts:
 
 Use the following resources to learn more about static analysis in Dart:
 
-* [Strong mode](https://github.com/dart-lang/dev_compiler/blob/master/STRONG_MODE.md#strong-mode)
+* [Strong Mode Dart](/guides/language/sound-dart)
```
kwalrath: All the other links aren't literal titles, so I don't think you need to change this title here. (More and more, I'm against using titles in links, since titles tend to change.)

src/_guides/language/sound-faq.md
```diff
@@ -264,7 +264,11 @@ The following image shows examples of type inference in strong mode clean code:
 <a name="is-it-optional"></a>
 ### I liked Dart before. Can I still write Dart code the way I always did?
 
-Yes, because strong mode is optional, you can write Dart code as before.
+Strong mode is optional in Dart 1.x, so you can write code as before.
```
kwalrath: This could be a bit clearer. Also, shouldn't we link to the new page? Maybe:

Yes, but we recommend migrating your code now. For help on getting ready for Dart 2.0, see [/dart-2.0](Dart 2.0 Updates).

_(Yeah, I broke my rule about not using page names in links, but I couldn't think of a better way to refer to the page.)_

src/dart-2.0.md
kwalrath: We do have dartdevc docs: https://webdev.dartlang.org/tools/dartdevc

src/dart-2.0.md
```diff
@@ -0,0 +1,138 @@
...
+Also, in Dart 2.0, a `num` can either be an
+`int` or a `double` at runtime.
+The main difference is that, in optionally-typed Dart, the analyzer
+infers the `dynamic` type, but in Dart 2.0, both the analyzer and
+the runtime try to infer something more specific than `dynamic`.
+
+While Dart 2.0 is statically typed, type inference and types like
+`dynamic` and `num` take much of the burden of specifying types away
+from the developer. Dart 2.0 offers the advantages of a strongly typed
+language, while only requiring slightly more work compared to Dart 1.x.
```
kwalrath: only requiring -> requiring only

kwalrath: You could change _slightly_ to something like _a bit_ to avoid two -lys in a row.

src/dart-2.0.md
```diff
@@ -0,0 +1,138 @@
...
+title: Dart 2.0 Updates
+description: How Dart 2.0 is different than 1.x, and how you can prepare.
+---
+
+This page has information about changes that are coming in Dart 2.0.
+
+<aside class="alert alert-info" markdown="1">
+**Flutter:** You don't need to do anything to prepare Flutter code for Dart 2.0.
+If changes become necessary, watch this page for updates.
+</aside>
```
kwalrath: I'd move this below the list (just above "What's different") and rephrase it slightly. Maybe:

**Flutter developer?**
You ... necessary, we'll add Flutter details to this page.

(I can't think of a good way to phrase it without "we" and future tense. Basically, I was trying to avoid the "if something happens that you could learn about by watching this page, watch this page" conundrum.)

src/dart-2.0.md
```diff
@@ -0,0 +1,138 @@
...
+What's different in Dart 2.0:
+
+* [Obsolete features](#obsolete features)
+* [Strong mode and static typing](#strong-mode-and-static-typing)
+
+## Migration tips: web
+
+If you develop for the web, you can start migrating to Dart 2.0 now:
+
+Convert your code to be strong mode compliant
```
kwalrath: Add a period?

src/dart-2.0.md
```diff
@@ -0,0 +1,138 @@
...
+: For more information,
+  see [Strong mode and static typing](#strong-mode-and-static-typing) below.
+
+Optimize the structure of your libraries
```
kwalrath: Add a period?

src/dart-2.0.md
```diff
@@ -0,0 +1,138 @@
...
+will eventually be phased out.
+{% endcomment %}
+
+## Migration tips: servers and command-line scripts
+
+If you develop servers or command-line scripts,
+you can start migrating to Dart 2.0 now:
+
+* Convert your code to be strong mode compliant. For more information,
+  see [Strong mode and static typing](#strong-mode-and-static-typing) below.
```
kwalrath: Make this a `<dl>`, like the web one.

src/dart-2.0.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Obsolete features
+
+The following features are redundant and will not be supported by Dart 2.0:
```
kwalrath: will not be supported by -> won't be in?

src/dart-2.0.md
```diff
@@ -0,0 +1,138 @@
...
+  performs fast compiles to JavaScript,
+  thanks to strong mode. Instead of Dartium, you’ll use Chrome
+  or other standard browsers for testing. For information on
+  Dartium's removal, see [A stronger Dart for
+  everyone.](http://news.dartlang.org/2017/06/a-stronger-dart-for-everyone.html)
+
+## Strong mode and static typing
+
+<aside class="alert alert-info" markdown="1">
+**Note about Flutter:**
```
kwalrath: Use the same **bold text** you have in the first note. I think starting with "Flutter" is probably a good idea, for scannability. I suggested **Flutter developer?** but **Flutter note:** would work, too.

PR #357, opened 2017-08-14
Make sure /dart-2.0 shows up as Content Type: html

filiph:
Without this fix, Firebase serves dartlang.org/dart-2.0 as “Content-Type: false”. The fix forces text/html on that particular page.

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -12,6 +12,15 @@
             "value": "*"
           }
         ]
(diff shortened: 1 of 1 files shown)


chalin:
Maybe a firebase bug (where it interprets the URL "extension" of "0" as false).

cc @kwalrath

PR #361, opened 2017-08-16
Adding info on the new function type syntax.

Sfshaza:
Adding a mention of the function type syntax introduced in Dart 1.24.

Fixes: https://github.com/dart-lang/site-www/issues/302

Staged: https://sz-www-1.firebaseapp.com/guides/language/language-tour#typedefs

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -3915,14 +3915,17 @@ main() {
 {% endprettify %}
 
 <aside class="alert alert-info" markdown="1">
-**Version note:**
-A new form of function types was introduced in Dart 1.24. Unless you define
-fields that are function types, or you're doing complex things with
-generic methods, like passing them around, or defining an argument
-that's a function type that's also generic, you may not care about the
-new function-type syntax. For more information, see
-[Generic Function Type
-Alias](https://gist.github.com/eernstg/ffc7bd281974e9018f10f0cb6cfee4aa).
+**New function type syntax:**
+Dart 1.24 introduced a new form of function types, the [generic function type
+alias.](https://github.com/dart-lang/sdk/blob/master/docs/language/informal/generic-function-type-alias.md)
+You might use this feature if you pass around generic methods,
+define fields that are function types,
+or define arguments with generic function types.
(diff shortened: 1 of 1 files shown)


src/_guides/language/language-tour.md
```diff
@@ -3914,6 +3914,16 @@ main() {
...
 
+<aside class="alert alert-info" markdown="1">
+**Version note:**
+A new form of function types was introduced in Dart 1.24. Unless you define
+fields that are function types, or you're doing complex things with
+generic methods, like passing them around, or defining an argument
+that's a function type that's also generic, you may not care about the
+new function-type syntax. For more information, see
+[Generic Function Type
+Alias](https://gist.github.com/eernstg/ffc7bd281974e9018f10f0cb6cfee4aa).
```
kwalrath: When you go there, it points you to https://github.com/dart-lang/sdk/blob/master/docs/language/informal/generic-function-type-alias.md.

src/_guides/language/language-tour.md
```diff
@@ -3914,6 +3914,16 @@ main() {
 }
 {% endprettify %}
 
+<aside class="alert alert-info" markdown="1">
+**Version note:**
+A new form of function types was introduced in Dart 1.24. Unless you define
```
kwalrath: The second sentence was hard for me to parse. Also, "Version note" didn't necessarily seem like the best title, unless you start the sentence with "Dart 1.24"...but even then, this is more a note about new syntax than about versions. An example would be nice, too. Maybe something like this (I've probably used incorrect wording, but you can fix that!):

---------
**New function type syntax:** Dart 1.24 introduced a new form of function types, the [generic function type alias.](https://github.com/dart-lang/sdk/blob/master/docs/language/informal/generic-function-type-alias.md) You might use this new feature if you pass around generic methods, define fields that are function types, or define arguments with generic function types. Here's an example of using the new syntax:

```
typedef F = List<T> Function<T>(T);
```
---------

src/_guides/language/language-tour.md
Sfshaza: That is a change since yesterday.  :)

Issue #296, opened 2017-06-06
graphic doesn't match text


kwalrath:
https://www.dartlang.org/guides/libraries/create-library-packages#importing-library-files says "The following graphic shows how to import lib/src/foo/a.dart from both lib and web."

But the web example imports lib/foo/a.dart, not lib/src/foo/a.dart. Also, we shouldn't encourage people to import anything from lib/src into web/main.dart.

I noticed this while I was working on the webdev dartdevc docs, where I'm spelling out the rules for imports.

/cc @jakemac53

kwalrath:
Also, for accessibility's sake, we should probably put the code in text, not [just] the image.

faruk13:
@kwalrath  this issue can be fixed by updating the text from lib/src/foo/a.dart to lib/foo/a.dart in appropriate place right? Correct me if I am wrong, beginner here.

kwalrath:
Yes, @faruk13, I think only the text update is necessary.

PR #363, opened 2017-08-18
updated create-library-packages.md

faruk13:
updated create-library-packages.md for correct import of the web example

### DIFF
--- a/src/_guides/libraries/create-library-packages.md
+++ b/src/_guides/libraries/create-library-packages.md
@@ -145,7 +145,7 @@ inside, or outside, of lib.
 When in doubt, use the `package:` directive; it works in all cases.
 
 The following graphic shows how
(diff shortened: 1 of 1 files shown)


faruk13:
@googlebot I signed it!

ben4ever:
Hmm, wasn't the original state of the file correct with `lib/src/foo/a.dart` or am I wrong?

I think the actual problem is the graphic which is shown below the text. I think the `import` statement on the right side in the graphic should say `import 'package:my_package/src/foo/a.dart` instead.

Or do I misunderstand something here?

kwalrath:
The way the pub works with dartdevc, we want to discourage importing from the `src` directory. But you're right, there's a src in the graphic. I'll remove that. Once I do that, #296 will be fixed.

PR #364, opened 2017-08-19
Redirect /)

### LINKED ISSUE #362: Redirect /)
Our top 404s lately have been for `dartlang.org/)`, mostly from youtube & disqus. We probably can't fix them at the source, so I'm going to try redirecting them to the homepage. I'd like to not have the terminal `/`, but I'm not sure that's practical.

kwalrath:
Fixes #362 

Test by going to: https://kw-www-dartlang-1.firebaseapp.com/)
It redirects to: https://kw-www-dartlang-1.firebaseapp.com

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -1118,6 +1118,11 @@
         "source": "/angular/cheatsheet",
         "destination": "https://webdev.dartlang.org/angular/cheatsheet",
         "type": 301
+      },
+      {
+        "source": "/)",
+        "destination": "/",
+        "type": 301
       }
     ]
   }


kwalrath:
I did a couple of linkcheck runs (all internal links, and then the old site URLs) to make sure this didn't mess up anything else.

chalin:
Interesting. Do they explain the rational for the `/)` notation? Will you want to do the same for site-webdev?

kwalrath:
It wouldn't hurt to do this for webdev, but I wasn't planning to... It's rare, especially compared to our other 404s (API docs, mostly).

I think the /) links happen when someone puts a link to the site (in this case, dartlang.org) in parentheses, and for whatever reason the ending parenthesis is included in the URL. We keep an eye on 404s, and for www.dartlang.org, we've done well enough with legit 404s that ones like this are now at the top of the list.

chalin:
Oh! `/)` is the actual URL you are receiving. I thought that it was a special pattern recognized by firebase. In that case, you can ignore my comments from earlier.

kevmoo:
I wish there was a response code which was "sure, I'll redirect it – but ya'll should fix your incoming link!"

PR #371, opened 2017-08-22
Update library import figure & add a note

### LINKED ISSUE #296: graphic doesn't match text

kwalrath:
Fixes #296 

Staged at https://kw-www-dartlang-1.firebaseapp.com/guides/libraries/create-library-packages#importing-library-files


src/_guides/libraries/create-library-packages.md
```diff
@@ -149,6 +149,15 @@ to import `lib/foo/a.dart` from both lib and web.
 
 {% img 'libraries/import-lib-rules.png' %}
 
+<aside class="alert alert-info" markdown="1">
+**Note:**
+Although the lib graphic shows `lib/bar/b.dart` using a relative import
+(`import '../../foo/a.dart'`),
```
chalin: The path should be `../foo/a.dart` (it was wrong before too).

kwalrath:
Thanks for catching that, @chalin! Fixed.

PR #374, opened 2017-09-07
Clarify docs about sound Dart

### LINKED ISSUE #360: Dart 2.0: remarks about type inference need adjustment
In the [dart-2.0 page, lines 107-108](https://github.com/dart-lang/site-www/blame/master/src/dart-2.0.md#L107-L108) it is written that:

> In Dart 2.0, both the analyzer and the runtime try to infer something more specific than `dynamic`.

This sentence needs to be revised because I doubt that there is any runtime type inference. (Type inference is used to attribute _static_ types.)

cc @wmleler @sethladd @kwalrath @kevmoo 

Also, in [sound-dart](https://www.dartlang.org/guides/language/sound-dart#type-inference) it is written:

> When the analyzer can’t infer the type, the `dynamic` type is assigned.

Technically, the analyzer _always_ infers a type, it's just that sometimes it can't infer anything more specific than `dynamic`; i.e., the sentence from the dart-2.0 page quoted above is more accurate, so maybe consider using it (once reference to "runtime" is removed :).

kwalrath:
Staged at:
https://kw-www-dartlang-1.firebaseapp.com/dart-2.0
https://kw-www-dartlang-1.firebaseapp.com/guides/language/sound-dart

Fixes #360

### DIFF
--- a/src/_guides/language/sound-dart.md
+++ b/src/_guides/language/sound-dart.md
@@ -345,7 +345,7 @@ However, the app throws an exception at runtime because it is an error
 to assign a list of Dogs to a list of Cats.
 
 <aside class="alert alert-info" markdown="1">
-**Note:** As of release 1.21, only dartdevc implements these runtime checks,
+**Note:** As of release 1.24, only dartdevc implements these runtime checks,
 but support in other tools is coming.
 </aside>
 
@@ -353,11 +353,13 @@ but support in other tools is coming.
 
 Does strong mode Dart mean that you _always_ have to specify a type?
 
-Actually, no. Strong mode Dart supports type inference. In some cases,
-the analyzer can infer types for fields, methods, local variables,
+No. Although _types_ are mandatory in strong mode,
+type _annotations_ are optional.
+The analyzer can infer types for fields, methods, local variables,
 and generic type arguments.
 
-When the analyzer can't infer the type, the `dynamic` type is assigned.
+When the analyzer doesn't have enough information to infer
+a specific type, it uses the `dynamic` type.
 
 How does type inference work with collections and generics?
 For example, what happens when you use `var` with maps or lists
--- a/src/dart-2.0.md
+++ b/src/dart-2.0.md
@@ -2,7 +2,7 @@
 layout: default
 permalink: /dart-2.0
 title: Dart 2.0 Updates
-description: How Dart 2.0 is different than Dart 1.x, and how you can prepare.
+description: How Dart 2.0 is different from Dart 1.x, and how you can prepare.
 ---
 
 This page has information about changes that are coming in Dart 2.0,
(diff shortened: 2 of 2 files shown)


src/dart-2.0.md
```diff
@@ -78,13 +78,20 @@ Dartium
...
 
-Dart 1.x is an optionally typed language, but types aren't optional in
-Dart 2.0. With both static and runtime type checks, Dart 2.0 has a sound
-type system.
+In Dart 1.x, types are optional. You can remove all type annotations
+from a Dart 1.x program without affecting its behavior.
+
+In Dart 2.0, _types_ are mandatory,
+but type _annotations_ are still optional.
+Wherever the type annotation is absent, tools infer the type.
```
chalin: How about:
- Wherever the type --> When a type

src/dart-2.0.md
```diff
@@ -104,12 +111,13 @@ As the example shows, you can use the static `dynamic` type
 to indicate that the runtime type is unknown.
 One big difference between optionally typed Dart and Dart 2.0
 is that, in the former, the analyzer infers the `dynamic` type.
-In Dart 2.0, both the analyzer and
-the runtime try to infer something more specific than `dynamic`.
+In Dart 2.0, tools like the analyzer can infer
+more specific types than `dynamic`,
```
chalin: I'd suggest a slight rewording because "more specific types than dynamic" doesn't quite read right for me.

> ... can [something-goes-here] infer types more specific than `dynamic`

Candidates for something-goes-here:
- often
- generally
- more easily

src/dart-2.0.md
```diff
@@ -104,12 +111,13 @@ As the example shows, you can use the static `dynamic` type
...
+following rules in the language specification.
 
-While Dart 2.0 is statically typed, type inference and types like
-`dynamic` and `num` (which can be `int` or `double`)
-take away much of the burden of specifying types.
+While Dart 2.0 is statically typed, type inference and
+less precise types such as `dynamic` and `num`
+take away much of the burden of specifying exact types.
```
chalin: Why not simply:
> While Dart 2.0 is statically typed, type inference takes away much of the burden of writing type annotations.

src/_guides/language/sound-dart.md
```diff
@@ -345,19 +345,21 @@ However, the app throws an exception at runtime because it is an error
...
 
 ## Type inference
 
 Does strong mode Dart mean that you _always_ have to specify a type?
 
-Actually, no. Strong mode Dart supports type inference. In some cases,
-the analyzer can infer types for fields, methods, local variables,
+No. Although _types_ are mandatory in strong mode,
+type _annotations_ are optional.
+The analyzer can infer types for fields, methods, local variables,
```
chalin: Can the analyzer really (reliably) infer the types for fields?

src/dart-2.0.md
kwalrath: I think we wanted to let people know that they could be specifically unspecific... that if they knew the type could be any object or any number, they could indicate it.

chalin: Ok (though that which takes away the burden is the type inference).

src/_guides/language/sound-dart.md
kwalrath: (asked again in email, just now)

bwilkerson: I don't believe that we infer the types for all fields, but will for final / const fields. @leafpetersen Am I remembering correctly?

chalin: For final fields that seems reasonable (assuming that they are initialized via an initializing expression, not through the constructor).

leafpetersen: Yes, we infer types for fields.  We do this via two techniques.  

The first is override inference: if a field is declared with a type in a superclass, and then overridden without a declared type, then the override will inherit the type from the overridden field.

If override inference does not apply, then we will use the inferred type from the RHS to infer a type for the LHS.  If inferring the type of the RHS requires (recursively) the type of the thing that we are attempting to infer, then it is a static error.

leafpetersen: This was our original proposal for toplevel inference:

https://github.com/dart-lang/sdk/blob/8f2be46aac7705ab6a1024a795aaf977519a8be2/docs/language/informal/toplevel-inference.md

It did not allow inference for instance fields: only static fields and toplevel variables.  That proved too restrictive, and so we generalized it to what I described above.  This generalized version hasn't been specified yet.

kwalrath:
OK, sounds like we're done (for now).

chalin:
Yes, thanks! Looking forward to having this documented (in the new spec?) once things settle down.

PR #376, opened 2017-09-08
Require an upper bound for SDK constraints

### LINKED ISSUE #375: sdk constraint section should suggest an upper bound
https://www.dartlang.org/tools/pub/pubspec#sdk-constraints

Pub is (probably) soon going to stop letting you publish versions of packages that don't have an upper bound constraint on the sdk (because its not safe at all generally). The docs should probably also suggest one so everybody doesn't run into this on new packages :).

For now the suggestion should probably be `>=1.x.x <2.0.0` (note that ^ syntax isn't supported for sdk constraints)

kwalrath:
Also improved wording of Flutter constraint discussion.

Grepping for site-www files that contain `sdk:`, I didn't find any other bad examples.

Staged: https://dartlang-org-dev.firebaseapp.com/tools/pub/pubspec#sdk-constraints

Fixes #375

### DIFF
--- a/src/tools/pub/pubspec.md
+++ b/src/tools/pub/pubspec.md
@@ -217,16 +217,26 @@ versions of it.
 A package can specify that using an *SDK constraint*. This goes inside a
 separate top-level `environment` field in the pubspec and uses the same
 [version constraint](/tools/pub/dependencies#version-constraints) syntax as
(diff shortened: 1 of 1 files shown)


kwalrath:
Thanks, @kevmoo.

Issue #293, opened 2017-06-02
Redirect top 404s


kwalrath:
Our top 404s for the last month (excluding some that I hope are fixed by the addition of https://www.dartlang.org/guides/testing):

1. /downloads/archive
1. /)
1. /docs/pub-package-manager
1. /download
1. /tools/debian
1. /tools/download-editor
1. /dartium
1. /docs
1. /codelabs/polymer

kwalrath:
In the past 30 days, the top few 404s (none of them worryingly high) were:

* /language-tour
* /api
* /dartium/home
* /tools/download-editor
* /codelabs/polymer
* /events/2013/flight-school
* /tools/debian
* /2.0

Sfshaza:
WOW, flight school?!?!

chalin:
@kwalrath - given that #384 has landed, can this be closed?

PR #384, opened 2017-09-23
Fix top 404s

kwalrath:
See #293

I've confirmed that these redirects work (link checker & interactive).

Staged at https://kw-www-dartlang-1.firebaseapp.com.


chalin:
LGTM

PR #390, opened 2017-10-03
Allow line breaks in links

### LINKED ISSUE #222: Allow line breaks in links
In guides/libraries/useful-libraries, a table has the following text:

```[stagehand command](http://stagehand.pub/)```.

Annoyingly, the whole link is kept on a single line, which usually makes the line before very short. This might be OK in regular text, but when width is limited, it's a problem. (It doesn't matter whether "stagehand" is code font or not.)

This might be related to https://github.com/dart-lang/site-webdev/issues/279.

undecidabot:
Fixes #222.

### DIFF
--- a/src/_assets/stylesheets/main.scss
+++ b/src/_assets/stylesheets/main.scss
@@ -56,17 +56,10 @@ a, button {
    text-decoration: none;
  }
  &.external {
-  position: relative;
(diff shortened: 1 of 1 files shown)


kwalrath:
I staged it locally, and it looks great. Thank you, @undecidabot!

PR #400, opened 2017-10-18
Update wording of "foundation" blurb

filiph:


filiph:
cc @hchiam: Merging this since we have approval from Kathy and I don't want to forget about this — but if you are strongly against, I'm happy to revert/amend. Thanks again for nudging us to do this!

hchiam:
@filiph No worries, and thanks for following up! I re-read this sentence in the context of the tooling goal, and I think your wording is great.

PR #405, opened 2017-10-22
Move from firebase to superstatic for serve

filiph:
For serve_local.sh, `firebase` is an overkill, and it needs logged-in user. `superstatic` is Firebase’s CLI for serving that doesn’t require login.

This is more convenient when you have more than one Firebase accounts and would need to switch to the correct one anytime you want to test the site.

More importantly, it unlocks he possibility to test the site (`linkcheck`) on Travis, as is the case in the site-webdev repo.


kwalrath:
Why was linkcheck in deploy? Was that a holdover from before you published https://pub.dartlang.org/packages/linkcheck?

chalin:
Good move &mdash; we've been using superstatic for site-webdev for a little while now.

README.md
```diff
@@ -33,10 +33,10 @@ Contributions welcome!
 1. Install `npm` by installing [Node.js](https://nodejs.org/en/); only needed to install firebase tools.
    Consider using [nvm](http://nvm.sh) to manage node versions;
    your **node version** must be **v6 or higher.**
-1. Install Firebase:
+1. Install Firebase & superstatic:
 
     ```
-    npm install -g firebase-tools     # might require sudo
+    npm install -g firebase-tools superstatic     # might require sudo
```
chalin: I don't hink that `sudo` would be required.

PR #410, opened 2017-10-26
Improve print styles

### LINKED ISSUE #149: print styles need work

undecidabot:
Fixes #149 

The following changes were made:
* Display a dotted underline under links
* Remove external link icon and link path
* Display a box (border) around notes
* Display dashed top and bottom borders around code blocks 

Please tell me if you want any further changes to be done. If it looks good, I'll port the changes to the webdev site as well

### DIFF
--- a/src/_assets/stylesheets/_overwrites.scss
+++ b/src/_assets/stylesheets/_overwrites.scss
@@ -359,7 +359,7 @@ body.homepage .get-started-section {
 
 	/* Display underlines under links */
 	a {
-		text-decoration: dotted underline;
+		text-decoration: underline;
 	}
 
 	/* Remove external link icon and link path */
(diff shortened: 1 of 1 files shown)


kwalrath:
Sorry for the delay, I was out of town. I'll take a look now!

kwalrath:
Thanks so much! This is a big improvement. I'd like a couple of minor tweaks:

* Change the code block border to be just like the note border. (Unless there's a reason to differentiate between the two? Maybe there's a problem if notes have embedded code blocks?)

* Change the link decorator to be a plain (not dotted) underline (and make sure there's some space between the underline and letters without descenders).

I might also like to hide the "Open Dartpad" links. Any ideas on the best way to do that?

kwalrath:
Maybe we could make the code block border lighter? In general, I like having solid lines, rather than dotted lines, because the dots often aren't rendered well.

kwalrath:
Looks great! So much better than what we had before. (I'll attach a before-and-after picture below.)

Someone just created a PR for some CSS on webdev, so let me check on that first (but yes, I'd love these changes there, too).

As for the Dart for JS Programmers article... I understand. If you happen to get anywhere (notes, anything), I'd love to see them, but I completely understand that you have other demands on your time.

Thanks very much for the contribution! As the photo shows, the printed version is much clearer and cleaner now.

img_20171113_133930
_(The old print formatting is on the left.)_

Issue #407, opened 2017-10-23
Code excerpts should be auto-updated and tested


chalin:
... and ideally, as part of every build, similar to what is done for https://github.com/dart-lang/site-webdev.

Eventually, all code excerpts in the site should be subject to testing. We'll be progressively covering site pages (so that the code excerpts they contain are tested). Here is an ordered list provided by @kwalrath in a separate thread:

- [x] Homepage (no.1 page on the site, with ~26% of views **[-> 29%]**) - #448
- [x] Dart 2 (no.3, ~6%) - #621
- [x] LANGUAGE > Tour (no.2, ~9%) - #414, #424, #427
- [x] LANGUAGE > Sample Code (no.11, 1.7% **[-> no.13]**) - #453
- [x] Library Tour (no.15) - #440, #443, #445
- [x] LANGUAGE  > Effective Dart (style is no.17, usage no.22, doc no.24, design no.26) - #474, #480
- [x] Tutorial: Asynchronous Programming: Futures (no.18) - #538 (also see  #512)
- [x] Tutorial (VM): Get Started (no.29 **[-> no.20]**) - #742
- [x] Strong Mode Dart (no.25) - #590, #603 
- [x] Tutorial (VM): Write HTTP Clients and Servers (no.12 **[-> no.25]**) - #463, #470
- [x] ~Cookbook: Dart by Example (no.42 **[-> no.31]**)~ #687
- [x] Tutorial: Asynchronous Programming: Streams (no.34); ~waiting on #512~ - #813
- [ ] Tutorial (VM): Write Command-line Apps (no.52)
- [ ] Native Extensions article #980
- [ ] Various articles (though some articles are quite dated). We should consider moving the content of the important ones into real docs.
  - await-async (no.45)
  - mixins (no.61)
  - serialization (no.64)
  - creating-streams (no.65)
  - zones (no.67, 0.25%)
- [ ] [Customize Static Analysis](https://www.dartlang.org/guides/language/analysis-options)

Possibly related issues:
- [x] Automated release testing #50
- [x] Use dartfmt on all samples in the language/library tour #190
- [ ] Document code coverage utility #52

kwalrath:
I just took a look at analytics for the past week, when we've gotten a lot of new visitors, and here are my conclusions:

* The VM Get Started tutorial should probably be tested, given its relatively high(er) readership.
* We should consider adding tests for the following pages, which have small amounts of code:
[...]

PR #411, opened 2017-10-26
Example code copied from up-and-running repo

chalin:
These are the original source files, unchanged, from up-and-running repo's `code` folder. These are the sources of the language and library tours.

Contributes to #407


chalin:
I want this to be just a copy  of the code that was in the other repo, w/o any changes.

(Since then that file has been renamed to `analysis_options.yaml`)

PR #414, opened 2017-11-01
Setup language-tour.md for auto code-refresh from examples/language-tour sources

chalin:
**Reviewers**:

You'll probably want to focus your review efforts on **[`src/_guides/language/language-tour.md` while ignoring whitespace diffs](https://github.com/dart-lang/site-www/pull/414/files?w=1#diff-cd4b0043d2416c7c2db7fe09bba74060)**:
- The code diffs that you will see are the result of having refreshed the code excerpts from their originating sources. (After this PR is merged, I'll set this up so that it happens automatically.)
- In very few places I adjusted the Tour prose and/or examples -- but usually only after discussing the proposed change with you via email.

TL;DR

The purpose of this PR has been to adapt the **Language Tour**, and its example sources, so that we can use the [code_excerpt_updater](https://github.com/chalin/code_excerpt_updater) to auto-refresh code excerpts in the Language Tour markdown.

For the most part, the language tour markdown already contained embedded paths to the source files it uses, and the source files contained some form of code excerpt markers. Consequently, most changes made in this PR are "lexical" and minor:

- Sources under `examples/language-tour`: changed code excerpt marker syntax from
  - `BEGIN`/`END` to 
  - `#docregion`/`#enddocregion`.
- `src/_guides/language/language-tour.md`: changed inline code excerpt comments
  - `<!-- path/to/source -->` to 
  - `<?code-excerpt? "path/to/source"?>` processing instructions.

Doc regions were manually adjusted to ensure that the code excerpts remained unchanged (except for genuine updates done to the code since the markdown was written).

In a few cases files were relocated because they were in a wrong folder (e.g., deprecated was moved to metadata).

Contributes to #407

What is left to do: I haven't touched code excepts that are in the Gists used with DartPad.


src/_guides/language/language-tour.md
```diff
@@ -284,9 +284,9 @@ Instance variables can be `final` but not `const`.
 
 Here's an example of creating and setting a final variable:
 
-<!-- language-tour/final-initialization/bin/main.dart -->
+<?code-excerpt "language-tour/final-initialization/bin/main.dart"?>
 {% prettify dart %}
-final name = 'Bob'; // Or: final String name = 'Bob';
+final String name = 'Bob'; // Or: final String name = 'Bob';
```
kevmoo: Why adding types?

src/_guides/language/language-tour.md
```diff
@@ -296,22 +296,22 @@ Where you declare the variable, set the value to a compile-time constant
 such as a number or string literal, a const
 variable, or the result of an arithmetic operation on constant numbers:
 
-<!-- language-tour/const/bin/main.dart -->
+<?code-excerpt "language-tour/const/bin/main.dart"?>
 {% prettify dart %}
-const bar = 1000000;       // Unit of pressure (dynes/cm2)
-const atm = 1.01325 * bar; // Standard atmosphere
+const int bar = 1000000; // Unit of pressure (dynes/cm2)
```
kevmoo: ....and here?

src/_guides/language/language-tour.md
```diff
@@ -452,30 +452,30 @@ const msUntilRetry = secondsUntilRetry * msPerSecond;
...
 single or double quotes to create a string:
 
-<!-- language-tour/quoting/bin/main.dart -->
+<?code-excerpt "language-tour/quoting/bin/main.dart"?>
 {% prettify dart %}
-var s1 = 'Single quotes work well for string literals.';
-var s2 = "Double quotes work just as well.";
-var s3 = 'It\'s easy to escape the string delimiter.';
-var s4 = "It's even easier to use the other delimiter.";
+String s1 = 'Single quotes work well for string literals.';
```
kevmoo: ...and here...

src/_guides/language/language-tour.md
chalin: Type annotations like these are appearing because the code excerpt is now synced with its originating source file. https://github.com/dart-lang/dart-up-and-running-book/pull/53 is the PR in which most of these type annotations were introduced.

Here is a direct link to the change made to the code excerpt you commented on: https://github.com/dart-lang/dart-up-and-running-book/pull/53/files#diff-9c88f184c7d08c6ebcfb3ff1909212db:

```diff
- final name = 'Bob'; // Or: final String name = 'Bob';
+ final String name = 'Bob'; // Or: final String name = 'Bob';
  // name = 'Alice';  // Uncommenting this causes an error
  // END(final_initialization)
  
- final name2 = 'Roberta';
+ final String name2 = 'Roberta';
  ...
  ```

In this, and many other cases, I believe that type annotations were added because the `type_annotate_public_apis` analyzer rule.

If this isn't what we want, then we'll need to drop that analyzer rule and fix the source files. Let me know what you prefer.

Comment (kevmoo) [reply src/_guides/language/language-tour.md]: This is kinda tough.

If we're just showing folks "here's how you set a variable", that code – IMHO – should be in a `main` method or something.

These are more tricky. v2 inference will mean even public fields – like these – won't strictly need types.

@munificent – thoughts here?

Comment (chalin) [reply src/_guides/language/language-tour.md]: > ... that code – IMHO – should be in a main method or something.

Agreed. I can review each of the code diffs (that add extra type annotations), and remove the annotation if it isn't meaningful for that particular example.

File: src/_guides/language/language-tour.md (hunk 3 above)
Comment (chalin) [reply src/_guides/language/language-tour.md]: These variable declarations appear in the section on "Strings", which may be why the variables are declared as `String`. (The same is true of the use of `int` and `double` above, in the sections on "Numbers".)

File: src/_guides/language/language-tour.md [hunk 4]
```diff
@@ -651,9 +651,9 @@ usually just call them *lists*.
 Dart list literals look like JavaScript array literals. Here’s a simple
 Dart list:
 
-<!-- language-tour/list-literal/bin/main.dart -->
+<?code-excerpt "language-tour/list-literal/bin/main.dart"?>
 {% prettify dart %}
-var list = [1, 2, 3];
+List<int> list = [1, 2, 3];
```
chalin: Ditto: IMHO, this type annotation helps emphasize that we are in the section talking about Lists.

chalin:
PTAL at the latest commit. I've removed only two type annotations: for each of the examples of `final` and `const`, I've left one declaration w/o a type and kept one declaration with a type annotation.

I've looked at all other cases, and the places where type annotations were added make sense to me (see my inline comments). Let me know what you think.

kwalrath:
I don't think we should use types anywhere that we wouldn't in our own well-written code. If need be, we can mention the type in a comment or the text.

examples/analysis_options.yaml
```diff
@@ -3,10 +3,10 @@
 
 analyzer:
   strong-mode: true
-  exclude:
-    - language-tour/reference/*.dart
-    - library-tour/reference/*.dart
-    - library-tour/async-await/bin/main.dart
+  # exclude:
```
filiph: Why not just remove these?

src/_guides/language/language-tour.md
```diff
@@ -27,15 +27,15 @@ more details about a language feature.
 
 The following code uses many of Dart’s most basic features:
 
-<!-- language-tour/basic-dart-program/bin/main.dart -->
+<?code-excerpt "language-tour/basic-dart-program/bin/main.dart"?>
 {% prettify dart %}
 // Define a function.
-printNumber(num aNumber) {
+void printNumber(num aNumber) {
```
filiph: I think, in language tour, we can actually make omissions like this (no `void`) for the sake of smoother beginnings. Does that break the code-excerpt pipeline? If this is because of a lint, maybe we want to use [`// ignore_for_file` annotations](https://www.dartlang.org/guides/language/analysis-options#ignoring-specific-analysis-rules)?

src/_guides/language/language-tour.md
filiph: But — the original page doesn't have the type annotations, right? https://www.dartlang.org/guides/language/language-tour#strings

Let's keep it that way.

src/_guides/language/language-tour.md
filiph: I would just leave `var`.

src/_guides/language/language-tour.md
chalin: > I think, in language tour, we can actually make omissions like this (no void) for the sake of smoother beginnings.

Ok

> Does that break the code-excerpt pipeline?

Not at all.

> If this is because of a lint, ...

Yes. I'm thinking that we have too many lint rules. Maybe we should opt for the same list that we recommend for Angular: https://github.com/dart-lang/stagehand/blob/master/templates/web-angular/analysis_options.yaml

@kwalrath @filiph WDYT?

examples/analysis_options.yaml
chalin: Gone

src/_guides/language/language-tour.md
chalin: > Let's keep it that way.

Ok, I'll revert those code changes that were done before the code was copied over into this repo

src/_guides/language/language-tour.md
chalin: Ok.

src/_guides/language/language-tour.md
filiph: I think in general it's good to have as many lint rules turned on for samples as possible (because by default, samples should be exceptionally "standard" code). I'd rather we keep them on but that we're not afraid to use `// ignore_for_file` where it makes sense.

chalin: SGTM

src/_guides/language/language-tour.md
```diff
@@ -2111,24 +2136,30 @@ non-null object—not just Exception and Error objects—as an exception.
...
 {% prettify dart %}
 throw 'Out of llamas!';
 {% endprettify %}
 
+<div class="alert alert-info" markdown="1">
+  **Note:** In production code, you'll usually throw types that implement
+  [Error]({{site.dart_api}}/dart-core/Error-class.html) or
+  [Exception]({{site.dart_api}}/dart-core/Exception-class.html).
+</div>
+
```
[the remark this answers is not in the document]
kwalrath: Looks good, but let's rewrite slightly to be more precise and avoid the unnecessary future tense:

**Note:** Production-quality code usually throws types that...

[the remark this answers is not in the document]
chalin: Done

Issue #416, opened 2017-11-02
Trim down and reorganize examples code


chalin:
@kwalrath - A few questions about the code that came from the up-and-running repo:

- Some of the examples contain much more code than what is used in the tour code excerpts. Could we trim this down (before we start writing tests for all the code excerpts - #407)?
- Is it necessary to keep the current example subfolder structure? Some very small code excerpts could be combined (of course, provided that there is a meaningful grouping).
- Each example (except for those under `reference`) has its own `main()` etc. Our main goals is to test the code excerpts, not necessarily run them. So in some cases, the extra code used to run the excerpt could be dropped in favor of having tests instead.

Thoughts?

kwalrath:
Feel free to do anything that'll make the code easier to deal with and test. You have carte blanche.

kwalrath:
Actually, one caveat... We might want to convert some/most of the examples to be DartPads. Those might need to be complete examples (or easily convertible to complete examples).

chalin:
Completed via #445, #443, #440, ... .

PR #427, opened 2017-11-13
Language tour example tests + Travis test job setup

### LINKED ISSUE #428: Omit the function return type
In [this section](https://www.dartlang.org/guides/language/language-tour#functions), the function return type in the second example should be omitted.

```dart
isNoble(int atomicNumber) {
  return _nobleGases[atomicNumber] != null;
}
```

chalin:
### What's in this PR

- See `examples/README.md` for details. Essentially, `examples` is now a Dart package from which all sample tests will be run. This PR contains the new source for the Language Tour.
- A few tests are run in chrome, so Travis setup has been adapted so that chrome is installed and configured to run tests.

---
### What to review

The main file to review is `src/_guides/language/language-tour.md`. Most of the changes to that file are updated `<?code-excerpt?>` paths, but a few of the code excerpts have been updated too. I'll be adding individual comments shortly to explain the most significant of those changes.

The remaining files in this PR are either new/adapted scripts in support of analyzing and testing the Language Tour sources, or the new Language Tour sources and their tests. (The Language Tour code excerpts aren't really new, they've just been reorganized and enhances with tests.)

---

**Note**: Eventually the "original" example sources (`examples/language_tour ` and `examples/library_tour `) will be deleted, once we are sure that all relevant parts (including some that are only in Gists) have been integrated into the new package.

Contributes to #407 and #416.
Fixes #428.

screen shot 2017-11-13 at 09 54 49

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,0 +1,71 @@
+# Examples
+
+This package contains all of the sources that appear in the Language Tour and the Library Tour, including the smallest of code excerpts.
+
+## Original sources
+
+Sources originally taken from the Up-and-Running repo (but reworked to contain doc regions)
+are in these folders:
+
+- `language_tour` (now obsolete, see below)
+- `library_tour`
+
+These original sources don't have tests, and practically each code excerpt is an independent
+package (with its own pubspec).
+
+## New sources
+
+Consolidated and reworked versions of the original sources are being developed.
+So far, the Language Tour sources have been completely rewritten, and are found
+in these folders:
+
+- `lib/language_tour`
+- `test/language_tour`
+
+As can be expected, Travis jobs run the
+
+- Analyzer over both `lib` and `test`
+- Tests under `test`
+
+### Source organisation
+
+The new Language Tour sources are under `lib/language_tour` and `test/language_tour`.
+Below each of these folders, you'll find either a file or a folder with a name
+that matches a top-level section of the Language Tour. This file/folder contains
+all of the code excerpt sources for that section.
+
+Under `lib` you might find a function defined as:
+
+```dart
+void miscDeclAnalyzedButNotTested(bool c) {
+  {
+    // #docregion foo-1
+    var foo = 1
+    // #enddocregion foo-1
+  }
+
+  {
+    // #docregion foo-2
+    var foo = 2
+    // #enddocregion foo-2
+  }
+
+  // ...
+}
+```
+
+That is, code regions are placed with a (headless) code block. This ensures
+that code region declarations (which somethings re-declare the same entity)
+don't clash.
+
+### Where to find a code excerpt
+
+Generally speaking, a small code excerpt that is subject to testing is embedded
+directly in its test. Code regions that are subject to analysis only
+are in `lib`. Larger examples are in `lib`, and their tests (if any) under `test`.
+
+### Why not test all?
+
+Not all code excerpts are tested because some are just fragments of anything
+useful, and others illustrate features that would required significant test
+scaffolding to be written and the effort isn't worth the small gain.
--- a/examples/README.md
+++ b/examples/README.md
@@ -55,7 +55,7 @@ void miscDeclAnalyzedButNotTested(bool c) {
 ```
 
 That is, code regions are placed with a (headless) code block. This ensures
-that code region declarations (which somethings re-declare the same entity)
+that code region declarations (which sometimes re-declare the same entity)
 don't clash.
 
 ### Where to find a code excerpt
@@ -69,3 +69,26 @@ are in `lib`. Larger examples are in `lib`, and their tests (if any) under `test
 Not all code excerpts are tested because some are just fragments of anything
 useful, and others illustrate features that would required significant test
 scaffolding to be written and the effort isn't worth the small gain.
+
+### _$print_
+
+All code excerpts use `$print()` instead of `print()`. The global `$print`
+variable (defined in `lib/util/print.dart`) is overwritten when running
+tests so that we can capture print output. By default `$print == print`.
+
+Of course, when displaying code excerpts in the Language Tour page,
+we show `print()` instead. It is `scripts/refresh-code-excerpts.sh` that does
+a global replace of `$print()` by `print()`.
+
+### Code highlights
+
+Code segments enclosed in the special syntax [!...!] will be highlighted.
+For example, the code
+
+    ```
+      int [!foo!] = bar;
+    ```
+
+will display as
+
+  <code>int <mark>foo</mark> = bar;</code>
--- a/examples/analysis_options.yaml
+++ b/examples/analysis_options.yaml
@@ -3,8 +3,9 @@
 
 analyzer:
   strong-mode: true
-#   exclude:
-#     - path/to/excluded/files/**
+  exclude:
+    - language_tour/**
+    - library_tour/**
 
 linter:
   rules:
--- a/examples/analysis_options.yaml
+++ b/examples/analysis_options.yaml
@@ -4,6 +4,7 @@
 analyzer:
   strong-mode: true
   exclude:
+    # Exclude the "old" sources. Only the new sources need be analyzed.
     - language_tour/**
     - library_tour/**
 
--- a/examples/lib/language_tour/classes/employee.dart
+++ b/examples/lib/language_tour/classes/employee.dart
@@ -1,0 +1,47 @@
+// ignore_for_file: unnecessary_cast
+// #docplaster
+import '../util/print.dart';
+
+Map getDefaultData() => {}; // stub
+
+// #docregion
+class Person {
+  String firstName;
+
+  Person.fromJson(Map data) {
+    $print('in Person');
+  }
+}
+
+// #docregion method-then-constructor
+class Employee extends Person {
+  // #enddocregion ,
+  Employee() : super.fromJson(getDefaultData());
+  // ···
+  // #enddocregion method-then-constructor
+  // #docregion
+  // Person does not have a default constructor;
+  // you must call super.fromJson(data).
+  Employee.fromJson(Map data) : super.fromJson(data) {
+    $print('in Employee');
+  }
+// #docregion method-then-constructor
+}
+// #enddocregion method-then-constructor
+
+void main() {
+  var emp = new Employee.fromJson({});
+  // Prints:
+  // in Person
+  // in Employee
+
+  // #docregion emp-is-Person
+  if (emp is Person) {
+    // Type check
+    emp.firstName = 'Bob';
+  }
+  // #enddocregion emp-is-Person
+  // #docregion emp-as-Person
+  (emp as Person).firstName = 'Bob';
+  // #enddocregion emp-as-Person
+}
--- a/scripts/browser.sh
+++ b/scripts/browser.sh
@@ -1,0 +1,16 @@
+#!/usr/bin/env bash
+
+set -e -o pipefail
+
+[[ -z "$NGIO_ENV_DEFS" ]] && . ./scripts/env-set.sh > /dev/null
+
+if [[ $TASK == *test* ]]; then
+  set -x
+  sh -e /etc/init.d/xvfb start
+  #t=0; until (xdpyinfo -display :99 &> /dev/null || test $t -gt 10); do sleep 1; let t=$t+1; done
+  # content_shell --run-layout-test --disable-gpu-early-init --disable-gpu
+  sleep 3
+  set +x
+else
+  echo "We don't launch chrome / content_shell for TASK $TASK"
+fi
--- a/scripts/dartfmt.sh
+++ b/scripts/dartfmt.sh
@@ -4,6 +4,8 @@
 
 set -e -o pipefail
 
+cd `dirname $0`/..
+
 DARTFMT="dartfmt -w"
 EXAMPLES=examples
 
@@ -23,6 +25,18 @@ $DARTFMT -l 60 \
 $DARTFMT -l 62 examples/language_tour/flow/exceptions/web/main.dart
 
 $DARTFMT -l 75 examples/language_tour/flow/break-continue/web/main.dart
+
+# New code
+
+$DARTFMT -l 60 \
+  examples/lib/language_tour/classes/immutable_point.dart \
+  examples/lib/language_tour/classes/logger_with_main.dart \
+  examples/lib/language_tour/classes/no_such_method.dart \
+  examples/lib/language_tour/classes/orchestra.dart \
+  examples/test/language_tour/built_in_types_test.dart \
+  examples/test/language_tour/functions_test.dart
+
+$DARTFMT -l 65 examples/lib/language_tour/exceptions.dart
 
 # If any files were changed, then exit 1:
 REFORMATTED_FILES=$(git status --short)
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -27,7 +27,7 @@ more details about a language feature.
 
 The following code uses many of Dart’s most basic features:
 
-<?code-excerpt "language-tour/basic-dart-program/bin/main.dart"?>
+<?code-excerpt "test/language_tour/basic_test.dart"?>
 {% prettify dart %}
 // Define a function.
 printNumber(num aNumber) {
@@ -98,8 +98,7 @@ mind:
 
 -   Everything you can place in a variable is an *object*, and every
     object is an instance of a *class*. Even numbers, functions, and
-    `null` are objects. All objects inherit from the
-    [Object]({{site.dart_api}}/{{site.data.pkg-vers.SDK.channel}}/dart-core/Object-class.html) class.
+    `null` are objects. All objects inherit from the [Object][] class.
 
 -   Specifying static types (such as `num` in the preceding example)
     clarifies your intent and enables static checking by tools.
@@ -205,7 +204,7 @@ You can't use reserved words as identifiers.
 
 Here’s an example of creating a variable and assigning a value to it:
 
-<?code-excerpt "language-tour/creating-a-variable/bin/main.dart"?>
+<?code-excerpt "lib/language_tour/variables.dart (var-decl)"?>
 {% prettify dart %}
 var name = 'Bob';
 {% endprettify %}
@@ -218,7 +217,7 @@ reference to a String object with a value of “Bob”.
 Uninitialized variables have an initial value of `null`. Even variables
 with numeric types are initially null, because numbers are objects.
 
-<?code-excerpt "language-tour/numbers-are-objects/bin/main.dart"?>
+<?code-excerpt "test/language_tour/variables_test.dart (var-null-init)"?>
 {% prettify dart %}
 int lineCount;
 assert(lineCount == null);
@@ -250,7 +249,7 @@ update-for-dart-2
 You have the option of adding static types to your variable
 declarations:
 
-<?code-excerpt "language-tour/static-types/bin/main.dart"?>
+<?code-excerpt "lib/language_tour/variables.dart (static-types)"?>
 {% prettify dart %}
 String name = 'Bob';
 {% endprettify %}
@@ -284,7 +283,7 @@ Instance variables can be `final` but not `const`.
 
 Here's an example of creating and setting a final variable:
 
-<?code-excerpt "language-tour/final-initialization/bin/main.dart"?>
+<?code-excerpt "lib/language_tour/variables.dart (final)"?>
 {% prettify dart %}
 final name = 'Bob'; // Without a type annotation
 // name = 'Alice';  // Uncommenting this causes an error
@@ -297,7 +296,7 @@ Where you declare the variable, set the value to a compile-time constant
 such as a number or string literal, a const
 variable, or the result of an arithmetic operation on constant numbers:
 
-<?code-excerpt "language-tour/const/bin/main.dart"?>
+<?code-excerpt "lib/language_tour/variables.dart (const)"?>
 {% prettify dart %}
 const bar = 1000000; // Unit of pressure (dynes/cm2)
 const double atm = 1.01325 * bar; // Standard atmosphere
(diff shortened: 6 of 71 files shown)


src/_guides/language/language-tour.md
```diff
@@ -308,13 +307,13 @@ You can also use it to create constant _values_,
...
 {% prettify dart %}
 // Note: [] creates an empty list.
-// const [] creates an empty, immutable list (EIA).
-var foo = const []; // foo is currently an EIA.
-final bar = const []; // bar will always be an EIA.
-const baz = const []; // baz is a compile-time constant EIA.
+// const [] creates an empty, immutable list (EIL).
+var foo = const []; // foo is currently an EIL.
+final bar = const []; // bar will always be an EIL.
+const baz = const []; // baz is a compile-time constant EIL.
```
chalin: "EIL" seemed like a better abbreviation for "empty, immutable list" than "EIA".

src/_guides/language/language-tour.md
```diff
@@ -489,22 +487,23 @@ units.
...
-var s1 = 'String ' 'concatenation'
-         " works even over line breaks.";
-assert(s1 == 'String concatenation works even over '
-             'line breaks.');
+var s1 = 'String '
+    'concatenation'
+    " works even over line breaks.";
+assert(s1 ==
+    'String concatenation works even over '
+    'line breaks.');
```
chalin: I can't seem to coerce *dartfmt* to give us the same formatting as before. If you think that it is important to preserve the original formatting, let me know and I'll add a few lines of custom perl to fix it after *dartfmt* has been applied to all the sources.

Btw, the formatting script is `scripts/dartfmt.sh`.

src/_guides/language/language-tour.md
```diff
@@ -1203,33 +1196,27 @@ The code block that follows contains the function's body:
...
 prints a string that includes the value at the specified index.
 
+<?code-excerpt "test/language_tour/functions_test.dart (anonymous-function)"?>
 {% prettify dart %}
-var list = ['apples', 'oranges', 'grapes', 'bananas', 'plums'];
-list.forEach((i) {
-  print(list.indexOf(i).toString() + ': ' + i);
+var list = ['apples', 'bananas', 'oranges'];
+list.forEach((item) {
+  print('${list.indexOf(item)}: $item');
```
chalin: IMHO we didn't need a fruit list that was that long. I also switched to using string interpolation (since the point of that example is to illustrate anonymous functions, and we'd already covered string interpolation).

src/_guides/language/language-tour.md
```diff
@@ -1358,6 +1347,13 @@ void main() {
 All functions return a value. If no return value is specified, the
 statement `return null;` is implicitly appended to the function body.
 
+<?code-excerpt "test/language_tour/functions_test.dart (implicit-return-null)"?>
+{% prettify dart %}
+foo() {}
+
+assert(foo() == null);
+{% endprettify %}
+
```
chalin: I added a short code excerpt in support of the prose.

src/_guides/language/language-tour.md
```diff
@@ -1586,16 +1581,33 @@ As you’ve already seen, you can assign values using the `=` operator.
...
+  b ??= value; // Assign value to b if b is null;
+  // otherwise, b stays the same
+  print('After: a == $a, b == $b');
+}
+
+main() {
+  assignValues(0, 0, 1);
+  assignValues(null, null, 1);
+}
+{% endprettify %}
```
chalin: This isn't a new code excerpt shown in the page (notice that it is inside a comment). Rather, it is just showing the code that is (should be) in the Gist.

src/_guides/language/language-tour.md
```diff
@@ -1697,36 +1709,33 @@ When you need to assign a value
...
-{% comment %}
-https://dartpad.dartlang.org/f2c8d82ce5d0dd533fe2
-https://gist.github.com/f2c8d82ce5d0dd533fe2
-{% endcomment %}
-
+<?code-excerpt "test/language_tour/operators_test.dart (if-null-alt)" replace="/\d//g"?>
 {% prettify dart %}
 // Slightly longer version uses ?: operator.
-String toString() => msg == null ? super.toString() : msg;
+String playerName(String name) => name != null ? name : 'Guest';
```
chalin: Just a reminder that offline, we had agreed to replace the old example using `super.toString()` by a simpler one that doesn't make use of `super`.

src/_guides/language/language-tour.md
```diff
@@ -2473,18 +2475,17 @@ name) constructor.
...
 
   Point(this.x, this.y);
 
   // Named constructor
-  Point.fromJson(Map json) {
-    x = json['x'];
-    y = json['y'];
+  [!Point.origin()!] {
+    x = 0;
+    y = 0;
```
chalin: A slightly simpler named constructor example (we use the `Point.fromJson()` constructor elsewhere).

src/_guides/language/language-tour.md
```diff
@@ -2591,21 +2592,14 @@ Besides invoking a superclass constructor, you can also initialize
...
-      : x = json['x'],
-        y = json['y'] {
-    print('In Point.fromJson(): ($x, $y)');
-  }
+// Initializer list sets instance variables before
+// the constructor body runs.
+Point.fromJson(Map json)
+    : x = json['x'],
+      y = json['y'] {
+  print('In Point.fromJson(): ($x, $y)');
```
chalin: I trimmed the code excerpt to show only the constructor being discussed (since the full class is shown repeated times earlier).

src/_guides/language/language-tour.md
```diff
@@ -2973,28 +2962,27 @@ class Person {
 }
 
 // An implementation of the Person interface.
-class Imposter implements Person {
-  // We have to define this, but we don't use it.
-  final _name = "";
+class Impostor implements Person {
+  get _name => '';
```
chalin: IMHO, this is the proper why to implement the Person interface `_name` element. `Person._name` is a final field, so the implicit interface contains only a getter named `_name`. (And this way it avoids the partly awkward comment of "We have to define this, but we don't use it."

src/_guides/language/language-tour.md
```diff
@@ -3014,9 +3002,9 @@ class Television {
   // ···
 }
 
-class SmartTelevision extends Television {
+class SmartTelevision [!extends!] Television {
   void turnOn() {
-    super.turnOn();
+    [!super!].turnOn();
```
chalin: I had started adding some extra highlighting to some of the examples, but stopped. We can add more in a separate pass.

When the code excerpt is non trivial (like in this case), judiciously placed highlights help focus the eye of the reader.

src/_guides/language/language-tour.md
```diff
@@ -3721,18 +3704,11 @@ Although an async function might perform time-consuming operations,
 it returns immediately—before
 any of its body executes.
 
-<?code-excerpt "language-tour/async-await/bin/main.dart (checkVersion-lookUpVersion)"?>
-{% prettify dart %}
-Future checkVersion() async {/* ... */}
-
-Future lookUpVersion() async {/* ... */}
-{% endprettify %}
-
```
chalin: I removed the code excerpt because, IMHO, it did nothing to support the prose.

src/_guides/language/language-tour.md
```diff
@@ -3853,17 +3829,16 @@ Click the run button ( {% img 'red-run.png' %} ) to execute the code.
...
 
-void main() {
-  var wf = new WannabeFunction();
-  var out = wf("Hi", "there,", "gang");
-  print('$out');
-}
+var wf = new WannabeFunction();
+var out = wf("Hi", "there,", "gang");
+
+main() => print(out);
```
chalin: I moved `var out` outside of `main()` so that its value can be tested.

examples/lib/language_tour/classes/employee.dart
```diff
@@ -0,0 +1,47 @@
...
+import '../util/print.dart';
+
+Map getDefaultData() => {}; // stub
+
+// #docregion
+class Person {
+  String firstName;
+
+  Person.fromJson(Map data) {
+    $print('in Person');
```
chalin: All code excerpts use `$print()` instead of `print()`. The global `$print` variable (defined in `../util/print.dart`) is overwritten when running tests so that we can capture print output.

Of course, when displaying code excerpts in the Language Tour, we show `print()` instead. The substitution is done by `scripts/refresh-code-excerpts.sh`.

src/_guides/language/language-tour.md
kwalrath: Good catch. Looks like we did a global search-and-replace of array to list, and missed this consequence.

src/_guides/language/language-tour.md
kwalrath: This is fine.

src/_guides/language/language-tour.md
kwalrath: Is this in the Return values section?

src/_guides/language/language-tour.md
```diff
@@ -2206,10 +2212,9 @@ your exception handler needs the exception object.
 
 You can specify one or two parameters to `catch()`.
 The first is the exception that was thrown,
-and the second is the stack trace
-(a [StackTrace]({{site.dart_api}}/{{site.data.pkg-vers.SDK.channel}}/dart-core/StackTrace-class.html) object).
+and the second is the stack trace (a [StackTrace][] object).
 
-<?code-excerpt "language-tour/flow/exceptions/web/main.dart (try-catch-2)" replace="/\(e.*?\)/[!$&!]/g"?>
+<?code-excerpt "lib/language_tour/exceptions.dart (try-catch-2)" replace="/\(e.*?\)/[!$&!]/g"?>
```
kwalrath: Please add something about `[!` to the README. (It could point elsewhere... I just want to make sure that every weird bit of syntax we use in these examples is easily findable in the README.)

src/_guides/language/language-tour.md
```diff
@@ -2677,14 +2670,15 @@ If your class produces objects that never change, you can make these
...
 class ImmutablePoint {
-  final num x;
-  final num y;
-  const ImmutablePoint(this.x, this.y);
   static final ImmutablePoint origin =
       const ImmutablePoint(0, 0);
+
+  final num x, y;
+
+  const ImmutablePoint(this.x, this.y);
```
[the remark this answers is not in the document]
kwalrath: It's a little convoluted (for the point the text is making), but I suppose ordering it like real code is better.

src/_guides/language/language-tour.md
kwalrath: Sounds good. Also, thanks for fixing the spelling of impostor! Hard to believe it went unnoticed for so long...

src/_guides/language/language-tour.md
kwalrath: We need to add a "Dart 2.0 note" here about this behavior changing, as described in https://github.com/dart-lang/sdk/blob/master/docs/newsletter/20170915.md#synchronous-async-start.

examples/README.md
```diff
@@ -0,0 +1,71 @@
...
+    var foo = 2
+    // #enddocregion foo-2
+  }
+
+  // ...
+}
+```
+
+That is, code regions are placed with a (headless) code block. This ensures
+that code region declarations (which somethings re-declare the same entity)
```
kwalrath: somethings -> sometimes

examples/analysis_options.yaml
```diff
@@ -3,8 +3,9 @@
 
 analyzer:
   strong-mode: true
-#   exclude:
-#     - path/to/excluded/files/**
+  exclude:
+    - language_tour/**
```
kwalrath: Add a comment about why these are excluded?

scripts/dartfmt.sh
```diff
@@ -24,6 +26,18 @@ $DARTFMT -l 62 examples/language_tour/flow/exceptions/web/main.dart
...
+
+$DARTFMT -l 60 \
+  examples/lib/language_tour/classes/immutable_point.dart \
+  examples/lib/language_tour/classes/logger_with_main.dart \
+  examples/lib/language_tour/classes/no_such_method.dart \
+  examples/lib/language_tour/classes/orchestra.dart \
+  examples/test/language_tour/built_in_types_test.dart \
+  examples/test/language_tour/functions_test.dart
+
+$DARTFMT -l 65 examples/lib/language_tour/exceptions.dart
```
kwalrath: Why the different line lengths for different examples?

(We should figure out what the optimum line length is, testing with various screen widths. But that can wait.)

src/_guides/language/language-tour.md
```diff
@@ -1742,9 +1751,9 @@ allows you to write more fluid code.
 
 Consider the following code:
 
-<?code-excerpt "language-tour/cascade-operator/web/main.dart"?>
+<?code-excerpt "test/language_tour/browser_test.dart (cascade-operator)"?>
 {% prettify dart %}
-querySelector('#button') // Get an object.
+querySelector('button') // Get an object.
```
kwalrath: Why `button` instead of `#button`? If it's because it seems like a bad name, how about something like `#confirm_button`?

src/_guides/language/language-tour.md
```diff
@@ -1757,15 +1766,17 @@ might be returned.
 
 The previous example is equivalent to:
 
+<?code-excerpt "test/language_tour/browser_test.dart (cascade-operator-example-expanded)"?>
 {% prettify dart %}
-var button = querySelector('#button');
+var button = querySelector('button');
```
kwalrath: same question

src/_guides/language/language-tour.md
```diff
@@ -3853,17 +3829,16 @@ Click the run button ( {% img 'red-run.png' %} ) to execute the code.
...
+  String call(String a, String b, String c) => '$a $b $c!';
 }
 
-void main() {
-  var wf = new WannabeFunction();
-  var out = wf("Hi", "there,", "gang");
-  print('$out');
-}
+var wf = new WannabeFunction();
+var out = wf("Hi", "there,", "gang");
```
kwalrath: " -> '

While you're in there... The sample code should use `'` instead of `"`, for consistency (except where we're discussing `"` specifically).

src/_guides/language/language-tour.md
chalin: Note (and link) added.

examples/lib/language_tour/classes/employee.dart
chalin: Done

src/_guides/language/language-tour.md
```diff
@@ -1586,16 +1581,33 @@ As you’ve already seen, you can assign values using the `=` operator.
...
 use the `??=` operator.
 
+<?code-excerpt "test/language_tour/operators_test.dart (assignment)"?>
 {% prettify dart %}
-a = value;   // Assign value to a
-b ??= value; // If b is null, assign value to b;
-             // otherwise, b stays the same
+a = value; // Assign value to a
+b ??= value; // Assign value to b if b is null;
+// otherwise, b stays the same
```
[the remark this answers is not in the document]
chalin: done

src/_guides/language/language-tour.md
chalin: done

examples/README.md
chalin: Fixed, thanks!

scripts/dartfmt.sh
chalin: In the previous pass over the code, I determined the line lengths by experimentation. My goal was to ensure that we could re-run dartfmt and preserving the existing layout of the code excerpts.

examples/analysis_options.yaml
chalin: > Exclude the "old" sources. Only the new sources need be analyzed.

Done.

src/_guides/language/language-tour.md
chalin: If there is a single button on the page, using `button` as a selector is just fine. Using `#confirm_button` would be better, though longer, and it doesn't add anything to the point being made in the prose (IMHO).

src/_guides/language/language-tour.md
chalin: same answer :)

src/_guides/language/language-tour.md
chalin: Done for this example. And I double checked that I caught that needed switch in all other examples (except where it was more convenient to use "... like 'this' ...").

examples/README.md
```diff
@@ -0,0 +1,94 @@
...
+we show `print()` instead. It is `scripts/refresh-code-excerpts.sh` that does
+a global replace of `$print()` by `print()`.
+
+### Code highlights
+
+Code segments enclosed in the special syntax [!...!] will be highlighted.
+For example, the code
+
+    ```
+      int [!foo!] = bar;
```
filiph: Nice and succinct. I wonder if that's going to clash with things like `var falseValues = [!true, false]`. Does it require the closing `!]`? If so, maybe mention that here so that people's worries are addressed. Ideally, link to the script that does the rewriting.

examples/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+Not all code excerpts are tested because some are just fragments of anything
+useful, and others illustrate features that would required significant test
+scaffolding to be written and the effort isn't worth the small gain.
+
+### _$print_
+
+All code excerpts use `$print()` instead of `print()`. The global `$print`
+variable (defined in `lib/util/print.dart`) is overwritten when running
+tests so that we can capture print output. By default `$print == print`.
```
filiph: This might need a bit more motivation here. Maybe show a code excerpt where you're testing the input, or just link to a dart file that uses `$print` and is most likely to survive forever.

Also, should we _always_ use `$print`, even in excerpts that are not tested (yet)? If so, clarify the language — "you should always use $print because ...".

examples/lib/language_tour/classes/employee.dart
```diff
@@ -0,0 +1,47 @@
...
+  Employee.fromJson(Map data) : super.fromJson(data) {
+    $print('in Employee');
+  }
+// #docregion method-then-constructor
+}
+// #enddocregion method-then-constructor
+
+void main() {
+  var emp = new Employee.fromJson({});
+  // Prints:
```
filiph: This is not some kind of a test, right? It's just a doc. I'm confused why we use `$print` above. Is that tested elsewhere?

filiph: Oh, never mind, I see that we do actually test this from a `classes_test.dart`. In that case, all good. Maybe I'd remove the comments here since it seems like these are not shown or tested anywhere and will get out of sync easily.

scripts/browser.sh
```diff
@@ -0,0 +1,16 @@
...
+
+set -e -o pipefail
+
+[[ -z "$NGIO_ENV_DEFS" ]] && . ./scripts/env-set.sh > /dev/null
+
+if [[ $TASK == *test* ]]; then
+  set -x
+  sh -e /etc/init.d/xvfb start
+  #t=0; until (xdpyinfo -display :99 &> /dev/null || test $t -gt 10); do sleep 1; let t=$t+1; done
+  # content_shell --run-layout-test --disable-gpu-early-init --disable-gpu
```
filiph: This script doesn't actually launch a browser or anything else, does it? Just xvfb.

scripts/dartfmt.sh
filiph: Should we just make the first `$DARTFMT` call (line 13) have `-l 60`? Or do more samples require the default (80)?

examples/README.md
```diff
@@ -0,0 +1,94 @@
+# Examples
+
+This package contains all of the sources that appear in the Language Tour and the Library Tour, including the smallest of code excerpts.
+
```
filiph: I think you'll want to add a writer-oriented manual. Something like

1. Add sample code as `.dart` files to `examples/directory/`
2. Annotate it with `#docregion`
3. Add the same code as excerpts to your `.md` articles
4. Annotate it with `<annotation>`

chalin:
Thank you for the feedback @filiph!

> I'm fine with this going to prod immediately if you have a look afterwards.

Ok, great, I'll merge this now and address the points you raise in the next wave of changes (for the Library Tour).

examples/README.md
chalin: Switching to a writer-oriented perspective is an excellent idea.

scripts/browser.sh
chalin: Right

scripts/dartfmt.sh
kwalrath: IIRC (which I might not), the maximum code line length for most pages we checked was 68. I'd be happy with a default 65- or 60-character line length.

examples/README.md
chalin: Good example of a potential clash, though it does seem unlikely that that will arise. If it does arise, we can adjust the highlight markers that our code_excerpt_updater tool uses (or make it a command line option -- similar to how AngularTS allows interpolation brackets (`{{`, `}}`) to be changed).

Or, if you prefer, I could change it now to `[![` ... `]!]`, which cannot clash with valid Dart code.

> Does it require the closing !]? 

Yes.

> If so, maybe mention that here so that people's worries are addressed.

Ok.

> Ideally, link to the script that does the rewriting.

Good idea.

examples/lib/language_tour/classes/employee.dart
chalin: > Maybe I'd remove the comments here since it seems like these are not shown or tested anywhere

This code shows up in DartPad IIRC (note that I haven't addressed DartPad code syncs yet; that will be done in a subsequent PR).

scripts/dartfmt.sh
chalin: Most samples are fine when the default 80 is used.

kwalrath: I'd like to avoid horizontal scrolling, wherever possible. 80 characters used to be too long for the window width I usually use.

chalin: > I'd be happy with a default 65- or 60-character line length.

Ok, I can try 60 for all samples and show you the result.

src/_guides/language/language-tour.md
filiph: While what Patrice says is true, this does rub me the wrong way. I'd consider it a potential error (or source of later error) if I saw this. 

How about `querySelector('#confirm')` if you want to stay terse? In my opinion, `#button` is also fine, as is `#confirm_button`.

examples/README.md
filiph: If it does require the closing `!]`, I think we're good. I can't imagine `!]` in regular Dart code outside some weird string literal (which, I can live with).

Still suggest documenting this here.

examples/lib/language_tour/classes/employee.dart
filiph: Okay. Let's keep it then.

scripts/browser.sh
filiph: I'd suggest renaming it to `xvfb.sh` or something similar, then. Unless you have a concrete plan of uncommenting that `content_shell` line.

scripts/dartfmt.sh
chalin: See #431, and let me know what you think.

scripts/browser.sh
chalin: Ok

src/_guides/language/language-tour.md
chalin: Ok, I'll switch to `#confirm` instead of `button` as selector.

src/_guides/language/language-tour.md
chalin: I'll switch to `#confirm` instead of `button` as selector here too.

examples/README.md
chalin: I'll followthrough via #436.

PR #431, opened 2017-11-16
[TEST ONLY] Language tour examples with dartfmt -l 60 applied

chalin:


examples/lib/language_tour/built_in_types.dart
```diff
@@ -6,7 +6,8 @@ void miscDeclAnalyzedButNotTested() {
     // #docregion integer-literals
     int x = 1;
     int hex = 0xDEADBEEF;
-    int bigInt = 34653465834652437659238476592374958739845729;
+    int bigInt =
+        34653465834652437659238476592374958739845729;
```
kwalrath: We picked this number because it fit on one line without overflowing...

examples/test/language_tour/exceptions_test.dart
```diff
@@ -26,14 +27,22 @@ void main() {
...
+            'main() finished handling ${e.runtimeType}.');
       }
     }
     // #enddocregion rethrow
 
     main();
     expect(
-        printLog.toString().contains('misbehave() partially handled'), isTrue);
-    expect(printLog.toString().contains('main() finished handling'), isTrue);
+        printLog
```
kwalrath: interesting, this version is much clearer!

src/_guides/language/language-tour.md
```diff
@@ -3320,7 +3335,8 @@ the list is probably a mistake. Here’s an example:
 {% prettify dart %}
 var names = new List<String>();
 names.addAll(['Seth', 'Kathy', 'Lars']);
-names.add(42); // Fails in checked mode (succeeds in production mode).
+names.add(
+    42); // Fails in checked mode (succeeds in production mode).
```
kwalrath: this is an unfortunate break

chalin:
I could determine suitable width increments but currently site-www bakes in code formatting so it wouldn't do us much good.  I can think of some CSS magic workaround, but that will require some experimentation (and extra tooling). Is this something that you'd like for me to look into, or shall we put this on ice until you come back?

kwalrath:
Let's ice this. It seems like a good thing to think about in December.

PR #440, opened 2017-11-20
Library tour example tests, up to but excluding dart:html

chalin:
Contributes to #407 and #416.

cc @Sfshaza

### DIFF
--- a/examples/lib/library_tour/async/future.dart
+++ b/examples/lib/library_tour/async/future.dart
@@ -1,0 +1,103 @@
+// ignore_for_file: unused_element, type_annotate_public_apis
+// #docregion import
+import 'dart:async';
+// #enddocregion import
+
+void miscDeclAnalyzedButNotTested() {
+  final args = <String>[];
+  Future<String> findEntryPoint() async => 'entrypoint';
+  Future<int> runExecutable(
+          String entryPoint, List<String> args) async =>
+      0;
+  Future<int> flushThenExit(int exitCode) async => 0;
+
+  {
+    // #docregion runUsingFuture
+    runUsingFuture() {
+      // ...
+      findEntryPoint().then((entryPoint) {
+        return runExecutable(entryPoint, args);
+      }).then(flushThenExit);
+    }
+    // #enddocregion runUsingFuture
+  }
+
+  {
+    // #docregion runUsingAsyncAwait
+    runUsingAsyncAwait() async {
+      // ...
+      var entryPoint = await findEntryPoint();
+      var exitCode = await runExecutable(entryPoint, args);
+      await flushThenExit(exitCode);
+    }
+    // #enddocregion runUsingAsyncAwait
+  }
+
+  {
+    Future catchExample() async {
+      // #docregion catch
+      var entryPoint = await findEntryPoint();
+      try {
+        var exitCode = await runExecutable(entryPoint, args);
+        await flushThenExit(exitCode);
+      } catch (e) {
+        // Handle the error...
+      }
+      // #enddocregion catch
+    }
+  }
+
+  final url = 'humans.txt';
+  Future costlyQuery(String url) async {}
+  Future expensiveWork(dynamic value) async {}
+  Future lengthyComputation() async {}
+
+  {
+    Future f() {
+      // #docregion then-chain
+      Future result = costlyQuery(url);
+      result
+          .then((value) => expensiveWork(value))
+          .then((_) => lengthyComputation())
+          .then((_) => print('Done!'))
+          .catchError((exception) {
+        /* Handle exception... */
+      });
+      // #enddocregion then-chain
+      return null;
+    }
+  }
+
+  {
+    Future f() async {
+      // #docregion then-chain-as-await
+      try {
+        final value = await costlyQuery(url);
+        await expensiveWork(value);
+        await lengthyComputation();
+        print('Done!');
+      } catch (e) {
+        /* Handle exception... */
+      }
+      // #enddocregion then-chain-as-await
+    }
+  }
+
+  bool elideBody = true;
+  {
+    Future f() async {
+      // #docregion wait
+      Future deleteLotsOfFiles() async => elideBody;
+      Future copyLotsOfFiles() async => elideBody;
+      Future checksumLotsOfOtherFiles() async => elideBody;
+
+      await Future.wait([
+        deleteLotsOfFiles(),
+        copyLotsOfFiles(),
+        checksumLotsOfOtherFiles()
+      ]);
+      print('Done with all the long steps!');
+      // #enddocregion wait
+    }
+  }
+}
--- a/examples/lib/library_tour/async/future.dart
+++ b/examples/lib/library_tour/async/future.dart
@@ -94,7 +94,7 @@ void miscDeclAnalyzedButNotTested() {
       await Future.wait([
         deleteLotsOfFiles(),
         copyLotsOfFiles(),
-        checksumLotsOfOtherFiles()
+        checksumLotsOfOtherFiles(),
       ]);
       print('Done with all the long steps!');
       // #enddocregion wait
--- a/scripts/analyze-and-test-examples.sh
+++ b/scripts/analyze-and-test-examples.sh
@@ -34,7 +34,9 @@ echo
 travis_fold start analyzeAndTest.tests.vm
 echo Running VM tests ...
 
-pub run test --reporter expanded --exclude-tags=browser | tee $LOG_FILE
+TEST="pub run test" # --reporter expanded
+
+$TEST --exclude-tags=browser | tee $LOG_FILE
 LOG=$(grep 'All tests passed!' $LOG_FILE)
 if [[ -z "$LOG" ]]; then EXIT_STATUS=1; fi
 travis_fold end analyzeAndTest.tests.vm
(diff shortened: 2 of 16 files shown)


chalin:
I forgot to mention that like for the Language Tour, the main file to review in this PR is `src/_guides/libraries/library-tour.md`. Most of the changes to that file are updated `<?code-excerpt?>` paths, but a few of the code excerpts have been updated too.

src/_guides/libraries/library-tour.md
```diff
@@ -984,6 +974,49 @@ Future objects appear throughout the Dart libraries, often as the object
 returned by an asynchronous method. When a future *completes*, its value
 is ready to use.
 
+
+#### Basic usage
```
chalin: This subsection isn't new, I just moved it up so that basic usage (explaining `Future.then()` etc) comes just before "Using await".

src/_guides/libraries/library-tour.md
```diff
@@ -1111,24 +1102,39 @@ In the preceding example, the methods run in the following order:
...
+try {
+  final value = await costlyQuery(url);
+  await expensiveWork(value);
+  await lengthyComputation();
+  print('Done!');
+} catch (e) {
+  /* Handle exception... */
+}
+{% endprettify %}
+
```
chalin: Given that the subsection before this one already covers "Using await", it seemed to make sense to include a version of the example code that uses await rather than `then()` chaining

scripts/analyze-and-test-examples.sh
```diff
@@ -34,7 +34,9 @@ echo
 travis_fold start analyzeAndTest.tests.vm
 echo Running VM tests ...
 
-pub run test --reporter expanded --exclude-tags=browser | tee $LOG_FILE
+TEST="pub run test" # --reporter expanded
```
filiph: What are we disabling here with commenting out `--reporter expanded`, exactly? Might want to just delete it or make a separate comment. This way it looks like a malformed explanation comment.

src/_guides/libraries/library-tour.md
```diff
@@ -76,9 +75,7 @@ assert(double.parse('1.2e+2') == 120.0);
 {% endprettify %}
 
 For more information, see the API documentation for
-[int,]({{site.dart_api}}/{{site.data.pkg-vers.SDK.channel}}/dart-core/int-class.html)
-[double,]({{site.dart_api}}/{{site.data.pkg-vers.SDK.channel}}/dart-core/double-class.html) and
-[num.]({{site.dart_api}}/{{site.data.pkg-vers.SDK.channel}}/dart-core/num-class.html) Also see
+[int,][int] [double,][double] and [num.][num] Also see
```
filiph: nit: `[num][num].` I think the period and the commas should be outside the link.

src/_guides/libraries/library-tour.md
```diff
@@ -536,18 +525,20 @@ Although Map doesn’t implement Iterable, you can get Iterables from it
 using the Map `keys` and `values` properties.
 </div>
 
-Use `isEmpty` to check whether a list, set, or map has no items:
+Use `isEmpty` or `isNotEmpty` to check whether a list, set, or map has items:
```
filiph: I wonder if we should add a comment explaining why the core library even has `isNotEmpty`. Something like "Practice showed that developers use `!list.isEmpty` a lot and so — after careful consideration — we decided to add `isNotEmpty`. It's redundant, but developers find it handy, readable and pragmatic."

cc @kwalrath 

Consider this non-blocking. I'd love to get Kathy's opinion once she gets back.

src/_guides/libraries/library-tour.md
filiph: Also, I believe you can just say `[num][]` in this case.

src/_guides/libraries/library-tour.md
filiph: FWIW, I'm not sure this is a good move. Most developers, most of the time, will want to use `async/await`. Maybe we want to call this section "low-level usage" and move it back?

src/_guides/libraries/library-tour.md
filiph: Just bookmarking in case you do decide to go with my suggestion to move "Basic usage" back to where it was. Not sure if it changes anything about the reasoning here. (I think it doesn't.)

src/_guides/libraries/library-tour.md
```diff
@@ -1111,24 +1102,39 @@ In the preceding example, the methods run in the following order:
...
+Future checksumLotsOfOtherFiles() async => /* ... */
 
-Future.wait([deleteDone, copyDone, checksumDone])
-    .then((List values) {
-      print('Done with all the long steps');
-    });
+await Future.wait([
+  deleteLotsOfFiles(),
+  copyLotsOfFiles(),
+  checksumLotsOfOtherFiles()
```
filiph: Nice. 

Style-wise, this looks like a good place for a trailing comma (dartfmt will be forced to always put each element on its own line, and moving elements around is much easier).

src/_guides/libraries/library-tour.md
```diff
@@ -1176,15 +1183,15 @@ The equivalent code with await expressions,
 including an asynchronous for loop (`await for`),
 looks more like synchronous code:
 
-<!-- dart-tutorials-samples/cmdline/bin/dgrep.dart -->
+<?code-excerpt "lib/library_tour/async/stream.dart (await-for)" replace="/await for/[!$&!]/g"?>
 {% prettify dart %}
-main(List<String> arguments) async {
-  ...
+Future main(List<String> arguments) async {
```
filiph: I wonder if we can elide the return type `Future` in this section, just to keep the code at a minimum.

src/_guides/libraries/library-tour.md
```diff
@@ -2695,3 +2668,63 @@ sampling of what you can install using pub.
 
 To learn more about the Dart language, see
 [A Tour of the Dart Language](/guides/language/language-tour).
+
+[AnchorElement]: {{site.dart_api}}/{{site.data.pkg-vers.SDK.channel}}/dart-html/AnchorElement-class.html
```
filiph: Thanks for doing this! It really makes the markdown read much better.

src/_guides/libraries/library-tour.md
chalin: Yes, good point. Done.

src/_guides/libraries/library-tour.md
chalin: While we elided return type annotations for `main()` functions at the start of the Language Tour to keep the code as light as possible, IMHO, our Library Tour should illustrate the recommended practice of specifying the return type of public APIs.

Generally speaking, I find it easy to gloss over the "async" keyword, whereas the "Future" annotation at the start of the method signature is usually grabs one attention :).

I'd vote to keep it. If you'd still like for me to remove it, let me know.

src/_guides/libraries/library-tour.md
chalin: That's what I figured, but the presentation seemed off on first reading. I.e., the "Using await" section, when first, uses an example of "Basic usage" to motivate the use of await.

Anyhow, I'll revert to the original order.

src/_guides/libraries/library-tour.md
chalin: > Not sure if it changes anything about the reasoning here. (I think it doesn't.)

Agreed. Leaving this as is.

scripts/analyze-and-test-examples.sh
chalin: Ah, I had meant to clean that out, thanks. (Running `pub run test` from a script makes `--reporter expanded` the default.)

Done.

examples/lib/library_tour/async/future.dart
```diff
@@ -0,0 +1,103 @@
+// ignore_for_file: unused_element, type_annotate_public_apis
+// #docregion import
+import 'dart:async';
+// #enddocregion import
+
+void miscDeclAnalyzedButNotTested() {
```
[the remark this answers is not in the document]
chalin: Ok, I'll add a few lines to the README when it is revised to have adopt writer-oriented perspective. How's this:

> Most code excerpts should be tested, unless they are so incomplete that they require extensive supplemental support infrastructure (which _isn't_ being shown in the tour), just to get something executable.

IMHO, it makes more sense to invest in testing complete samples, when they exist, than to have to create considerable support code. Also, once the code samples are setup for processing of `docregion`s, then we could replace some of the tour code excerpts with snippets from complete samples.

src/_guides/libraries/library-tour.md
chalin: Right, but those render in the HTML as external links, which have an arrow icon afterwards:
> screen shot 2017-11-21 at 08 12 21

Kathy much prefers having the punctuation before the "external" icon.

src/_guides/libraries/library-tour.md
chalin: SGTM, I'll add this to a list of topic to review with Kathy when she returns.

chalin:
@filiph, thanks for the quick feedback! Changes submitted.

src/_guides/libraries/library-tour.md
kwalrath: I don't want to add a comment this to the tour, since it generally tries to be as succinct as possible. If there's a place that talks about designing APIs (e.g. Effective Dart), _that_ might be an appropriate place to mention this.

src/_guides/libraries/library-tour.md
kwalrath: I'd prefer to remove the type before main(). It doesn't seem to help, and it obscures the method name.

src/_guides/libraries/library-tour.md
chalin: There is _some_ form of justification given in this Effective Dart Usage Dont': https://www.dartlang.org/guides/language/effective-dart/usage#dont-use-length-to-see-if-a-collection-is-empty

Shall I make "Use `isEmpty` or `isNotEmpty`" a link to this entry?

PR #443, opened 2017-11-22
Library tour example tests from dart:html

chalin:
Contributes to #407 and #416.

Continuation of #440

### DIFF
--- a/examples/test/library_tour/html_test.dart
+++ b/examples/test/library_tour/html_test.dart
@@ -1,0 +1,212 @@
+// #docplaster
+@Tags(const ['browser'])
+@TestOn('browser')
+import 'dart:async';
+import 'dart:html';
+import 'package:test/test.dart';
+import 'package:examples/util/logger.dart';
+
+void main() {
+  test('querySelector', () {
+    final html =
+        // #docregion anchor-html
+        '<a id="example" href="http://example.com">link text</a>';
+    // #enddocregion anchor-html
+    document.body.appendHtml(html);
+
+    // #docregion href
+    var anchor = querySelector('#example') as AnchorElement;
+    anchor.href = 'http://dartlang.org';
+    // #enddocregion href
+    // The code above should throw no exceptions.
+  });
+
+  test('querySelectorAll', () {
+    // #docregion os-html
+    final html = '''<!-- In HTML: -->
+    <p>
+      <span class="linux">Words for Linux</span>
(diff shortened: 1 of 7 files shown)


src/_guides/libraries/library-tour.md
```diff
@@ -1758,93 +1761,48 @@ Traditionally, AJAX-style apps make heavy use of HttpRequest. Use
 HttpRequest to dynamically load JSON data or any other resource from a
 web server. You can also dynamically send data to a web server.
 
-The following examples assume all resources are served from the same web
-server that hosts the script itself. Due to security restrictions in the
-browser, the HttpRequest class can’t easily use resources that are
-hosted on an origin that is different from the origin of the app. If you
-need to access resources that live on a different web server, you need
-to either use a technique called JSONP or enable CORS headers on the
-remote resources.
```
chalin: This section is about "Using HTTP resources with HttpRequest", so I've simplified the examples by dropping use of JSON (since JSON is covered in the `dart:convert` section further below in this tour.

src/_guides/libraries/library-tour.md
```diff
@@ -1758,93 +1761,48 @@ Traditionally, AJAX-style apps make heavy use of HttpRequest. Use
...
-    print('License: $license');
-  } catch (e) {
-    print("$xmlUri doesn't have correct XML formatting.");
+Future main() async {
+  HttpRequest req = await HttpRequest.request(
+    url,
+    method: 'HEAD',
+  );
+  if (req.status == 200) {
+    // Successful URL access...
```
chalin: Again, this example has been simplified, focusing on the `HttpRequest.request()` call. For variability, I've shown a HEAD method call.

examples/test/library_tour/html_test.dart
```diff
@@ -0,0 +1,212 @@
...
+
+    try {
+      initWebSocket();
+      expect(await t.first, contains('Received message'));
+    } finally {
+      await wsStream.close();
+    }
+  });
+
+  group('xxxx', () {
```
filiph: Is this supposed to be here? Should the name be a bit too descriptive? Just checking.

chalin: Oops, no; the 'xxxx' entries are just templates I use when copy-pasting new code excerpts. They'll be removed in the next round (and by the time we're done with the Library Tour they should be gone for good).

PR #453, opened 2017-12-01
Tests for samples

chalin:
Contributes to #407 

Staged at: https://dartlang-org-staging-0.firebaseapp.com/samples

### DIFF
--- a/src/_samples/index.md
+++ b/src/_samples/index.md
@@ -1,7 +1,6 @@
 ---
-title: "Sample Code"
-description: "Examples of idiomatic Dart with links to larger examples."
-permalink: /samples
+title: Sample Code
+description: Examples of idiomatic Dart with links to larger examples.
 ---
 
 This collection is not exhaustive&mdash;it's just a brief
@@ -34,6 +33,7 @@ also want to check out the following pages.
 
 First things first:
 
+<?code-excerpt "tours/test/samples_test.dart (hello-world)"?>
 {% prettify dart %}
 void main() {
   print('Hello, World!');
@@ -44,6 +44,7 @@ void main() {
 
 We're not explicitly typing the variables here. We could, but type inference works for us.
 
+<?code-excerpt "tours/test/samples_test.dart (var)"?>
 {% prettify dart %}
 var name = 'Voyager I';
 var year = 1977;
@@ -61,6 +62,7 @@ var image = {
 
 No surprises here.
 
+<?code-excerpt "tours/test/samples_test.dart (control-flow)"?>
 {% prettify dart %}
 if (year >= 2001) {
   print('21st century');
@@ -88,6 +90,7 @@ including `break` and `continue`, `switch` and `case`, and `assert`.
 
 As a best practice, we're specifying the type of the function's argument and return value here. You don't need to do that, though.
 
+<?code-excerpt "tours/test/samples_test.dart (functions)"?>
 {% prettify dart %}
 int fibonacci(int n) {
   if (n == 0 || n == 1) return n;
@@ -101,6 +104,7 @@ A shorthand (_fat arrow_) syntax is handy for functions that contain a single st
 It's especially useful when functions are passed as arguments,
 but it also means that Hello World can be made even [shorter](https://gist.github.com/filiph/8a5e3e845acdafe2ea928fd257a46859).
 
+<?code-excerpt "tours/test/samples_test.dart (fat arrow)"?>
 {% prettify dart %}
 flybyObjects.where((name) => name.contains('turn')).forEach(print);
 {% endprettify %}
@@ -126,16 +130,18 @@ including how the documentation tooling works.
 
 ## Imports
 
+<?code-excerpt "tours/test/samples_test.dart (import)" replace="/'.*?' as \w+/'path\/to\/my_other_file.dart'/g"?>
 {% prettify dart %}
 // Importing core libraries
 import 'dart:async';
 import 'dart:math';
 
 // Importing libraries from external packages
-import 'package:angular2/angular2.dart';
+import 'package:test/test.dart';
 
 // Importing files
 import 'path/to/my_other_file.dart';
+import '../lib/samples/spacecraft.dart';
 {% endprettify %}
 
 [Read more](/guides/language/language-tour#libraries-and-visibility) about libraries and visibility in Dart,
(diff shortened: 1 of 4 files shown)


src/_samples/index.md
```diff
@@ -126,23 +130,26 @@ including how the documentation tooling works.
...
 
+<?code-excerpt "test/samples_test.dart (import)" replace="/'.*?' as \w+/'path\/to\/my_other_file.dart'/g"?>
 {% prettify dart %}
 // Importing core libraries
 import 'dart:async';
 import 'dart:math';
 
 // Importing libraries from external packages
-import 'package:angular2/angular2.dart';
+import 'package:test/test.dart';
```
chalin: I figured that external package would do and so went with `test`.  Avoiding angular here means that we avoid the delays on precompiling it when `pub get/upgrade` is run.

src/_samples/index.md
```diff
@@ -252,26 +268,37 @@ abstract class Describable {
...
 
-Any class extending `Describable` has the `describeWithEmphasis()` method, which calls the extender's implementation of `describe()`.
+Any class extending `Describable` or using it as a mixin has the `describeWithEmphasis()` method, which calls the extender's / base class implementation of `describe()`.
+
+<?code-excerpt "lib/samples/spacecraft.dart (abstract use)" replace="/abstract/[!$&!]/g"?>
+{% prettify dart %}
+class Probe extends Spacecraft with Describable {
+  Probe(String name) : super(name, new DateTime.now());
+}
+{% endprettify %}
```
chalin: I've added example example of the use of the abstract class, though as a mixin.

src/_samples/index.md
```diff
@@ -203,9 +215,10 @@ class Orbiter extends Spacecraft {
 
 Mixins are a way of reusing code in multiple class hierarchies. The following class can act as a mixin.
 
+<?code-excerpt "lib/samples/spacecraft.dart (mixin)"?>
 {% prettify dart %}
-class Manned {
-  int astronauts;
+class Piloted {
```
kwalrath: Nice change!

src/_samples/index.md
```diff
@@ -226,11 +240,12 @@ class Orbiter extends Spacecraft with Manned {
...
 ## Interfaces
 
-Dart has no `interface` keyword. Instead, all classes implicitly define an interface. Therefore, you can `implement` any class.
+Dart has no `interface` keyword. Instead, all classes implicitly define an interface. Therefore, you can implement any class.
 
+<?code-excerpt "lib/samples/spacecraft.dart (implements)"?>
 {% prettify dart %}
 class MockSpaceship implements Spacecraft {
-  // ...
+  // ···
```
kwalrath: I just noticed the changes from `...` to `···`. It looks strange to me because of the vertical alignment—like you're showing single-quote characters, rather than  an ellipsis (which I expect to look like [or be] three periods).

src/_samples/index.md
```diff
@@ -252,26 +268,37 @@ abstract class Describable {
 }
 {% endprettify %}
 
-Any class extending `Describable` has the `describeWithEmphasis()` method, which calls the extender's implementation of `describe()`.
+Any class extending `Describable` or using it as a mixin has the `describeWithEmphasis()` method, which calls the extender's / base class implementation of `describe()`.
```
kwalrath: extender's / base class
->
extender or base class

(I usually avoid "/" in text, since it can be ambiguous. https://developers.google.com/style/slashes)

src/_samples/index.md
kwalrath: LGTM. @filiph what do you think?

src/_samples/index.md
```diff
@@ -252,26 +268,37 @@ abstract class Describable {
...
-  await new Future.delayed(const Duration(seconds: 1));
+const aSecond = const Duration(seconds: 1);
+// ···
+Future<Null> printWithDelay(String message) [!async!] {
+  await new Future.delayed(aSecond);
   print(message);
 }
 {% endprettify %}
 
 The code above is equivalent to:
```
kwalrath: code -> method

(because you added the code for const aSecond)

src/_samples/index.md
```diff
@@ -252,26 +268,37 @@ abstract class Describable {
...
 
 ## Async
 
 You can avoid callback hell and make your code much more readable by using `async` and `await`.
 
+<?code-excerpt "test/samples_test.dart (async)" replace="/async/[!$&!]/g"?>
 {% prettify dart %}
-Future<Null> printWithDelay(String message) async {
-  await new Future.delayed(const Duration(seconds: 1));
+const aSecond = const Duration(seconds: 1);
```
kwalrath: I keep reading "aSecond" as "a second [something]", so maybe:

`aSecond` -> `oneSecond`

src/_samples/index.md
filiph: I'm not sure if this alone helps explain how this works. If we're adding sample code, I'd suggest adding:

```
var probe = new Probe("probe");  // TODO(chalin): some better name
probe.describeWithEmphasis();  // Method taken from the Describable class.
```

Also not sure whether introducing a mixin in the example is a good idea. It might be confusing that we're suddenly showing a mixin.

src/_samples/index.md
chalin: Wow, you are right. Ok, I'll have to make a sweeping change of these.

src/_samples/index.md
chalin: Ok, then I'll just pull that out.

src/_samples/index.md
chalin: (Since the mixin has been dropped, I've reverted to the original sentence.)

src/_samples/index.md
chalin: Good catch. Fixed.

src/_samples/index.md
chalin: Done

src/_samples/index.md
chalin: I've reverted to using dots in the samples code. I'll address other instances separately.

chalin:
Post-review edits done. Kathy gave a 👍 to merging, so here goes ...

chalin:
@kwalrath - btw, the samples included here are from the main [Samples](https://www.dartlang.org/samples) page. They are, in a sense, a "mini sample tour" with links to larger samples. That is why I put the code for these mini samples under the `examples/tours` folder. Larger samples will be under `examples`, but under their own packages.

PR #457, opened 2017-12-04
Rename example file

natebosch:
In the new pub website `lunch.dart` won't be treated as the example.
Update instead to use one of the filenames that can be identified and
shown as an example on the pub site.

https://github.com/dart-lang/pub-dartlang-dart/blob/8ae285f2b846b0e3bf1098162318f6c52eff5776/app/lib/shared/utils.dart#L219


kevmoo:
Great find, @natebosch !

chalin:
@kevmoo, shall I merge?

kevmoo:
Please!

PR #463, opened 2017-12-07
Tests for dart-vm httpserver tutorial code excerpts

chalin:
Contributes to #407, covering the **Tutorial (VM): Write HTTP Clients and Servers (no.12)**

- Staged at https://dartlang-org-staging-0.firebaseapp.com/tutorials/dart-vm/httpserver
- Current https://www.dartlang.org/tutorials/dart-vm/httpserver

Some code excerpts have note markers. It was a bit tricky to deal with those in the context of code excerpts that get refreshed. I've experimented with a few approaches and settled on the use of comment markers `/*1*/`, `/*2*/`, etc. Let me know what you think.


examples/httpserver/bin/hello_world_server.dart
```diff
@@ -4,22 +4,24 @@
...
+  print('Listening on localhost:${server.port}');
 
-  try {
-    await for (HttpRequest request in requestServer) {
-      request.response
-        ..write('Hello, world!')
-        ..close();
-    }
-  } catch (e) {
-    print(e.toString);
```
chalin: The code excerpt in the tutorial page didn't contain a try-catch statement. I've reverted the code to match the excerpt, because

- Catching an exception and only printing a message is bad practice. (Note only that, but the code is wrong; it prints the value of the toString function).
- There are better examples of error handling elsewhere in the tutorial.

kwalrath:
Looking now...

src/_tutorials/dart-vm/httpserver.md
```diff
@@ -555,66 +560,55 @@ The `close()` method sends the request to the server
...
+    'ship': 'Millennium Falcon',
+    'weakness': 'smuggling debts'
+  };
+
+  var request = await new HttpClient()
+      [!/*1*/!] .post(InternetAddress.LOOPBACK_IP_V4.host, 4049, 'file.txt');
+  [!/*2*/!] request.headers.contentType = ContentType.JSON;
+  [!/*3*/!] request.write(JSON.encode(jsonData));
+  [!/*4*/!] HttpClientResponse response = await request.close();
+  [!/*5*/!] await for (var contents in response.transform([!UTF8.decoder!])) {
```
kwalrath: Why is only `UTF8.decoder` highlighted, and none of the other code that used to be highlighted?

src/_tutorials/dart-vm/httpserver.md
```diff
@@ -555,66 +560,55 @@ The `close()` method sends the request to the server
...
-The `close()` method sends the request to the server and, when complete,
-returns an [HttpClientResponse][] object
-that's assigned to the `reponse` variable.
-
-<span class="code-note">6</span>
-The response from the server is encoded in UTF-8.
-Use a transformer defined in the `dart:convert` library
-to convert the data into regular Dart string format.
+{:.code-notes}
+1. The `post()` method requires the host, port, and the path to the requested
```
kwalrath: Why did you remove the discussion of `request` and `await`?

src/_tutorials/dart-vm/httpserver.md
```diff
@@ -555,66 +560,55 @@ The `close()` method sends the request to the server
...
+   contains the request headers. For some headers, like `contentType`,
+   HttpHeaders has a property specific to that header. For other headers, use
+   the `set()` method to put the header in the HttpHeaders object.
+
+3. The client writes data to the request object using `write()`. The encoding,
+   JSON in this example, matches the type specified in the [ContentType][]
+   header.
+
+4. The `close()` method sends the request to the server and, when complete,
+   returns an [HttpClientResponse][] object that's assigned to the `reponse`
```
kwalrath: reponse -> response

(old typo!)

src/_tutorials/dart-vm/httpserver.md
```diff
@@ -555,66 +560,55 @@ The `close()` method sends the request to the server
...
+   variable.
+
+5. The UTF-8 response from the server is decoded. Use a transformer defined in
```
kwalrath: I liked the original wording. Why the change?

src/_tutorials/dart-vm/httpserver.md
```diff
@@ -629,13 +623,14 @@ for example, the URI has no query string
...
 
-<strong>Bonus code:</strong>
-If you would like to see some client code that
-makes GET requests,
-check out the code for
-[number_guesser.dart.]({{gh-path}}/bin/number_guesser.dart)
-It's a standalone client for the number thinker server
-that makes periodic guesses until it guesses correctly.
+<aside class="alert alert-info" markdown="1">
+  If you would like to see some client code that
```
kwalrath: We can make this shorter (not your text, I know) and add a title that makes the transition clearer. (I don't know, offhand, why this section about POST requests is pointing you to a GET example.) Something like this (but better, this is a WAG):

**Curious about how GET and POST differ?** To see client code that makes GET requests instead of POST requests, look at .... Compare it to....

src/_tutorials/dart-vm/httpserver.md
```diff
@@ -656,66 +651,67 @@ join() method in Stream to concatenate the string values of those chunks.
...
+import 'dart:convert';
+
+Future main() async {
+  var server = await HttpServer.bind(InternetAddress.LOOPBACK_IP_V4, 4049);
+  await for (var req in server) {
+    ContentType contentType = req.headers.contentType;
+    HttpResponse response = req.response;
+
+    if (req.method == 'POST' &&
+        [!contentType != null && /*1*/!]
```
kwalrath: Could this be shortened to something like `'application/json' == contentType?.mimeType`?

kwalrath: It's easy to miss the `/*1*/` when there's no break between the highlighted code and the footnote number. Would it be too horrible to NOT highlight the character before the `/*`?

src/_tutorials/dart-vm/httpserver.md
```diff
@@ -780,35 +776,38 @@ It responds to all requests by returning the contents of the
...
 
+<?code-excerpt "httpserver/bin/mini_file_server.dart"?>
 {% prettify dart %}
+import 'dart:async';
 import 'dart:io';
 
-main() async {
+File targetFile = new File('index.html');
+
+Future main() async {
```
kwalrath: I kind of hate seeing `main` annotated with a type. It obscures the entry point for little to no benefit, as far as I can see.

Does the analyzer complain if we leave out the type?

examples/httpserver/bin/hello_world_server.dart
kwalrath: Makes sense. We'll need to update the dart-tutorial-samples repo to match, of course. (But that's true everywhere.)

src/_tutorials/dart-vm/httpserver.md
```diff
@@ -930,40 +932,48 @@ calls `bindSecure()` using
...
-     }
-   }
+import 'dart:async';
+import 'dart:io';
+
+Future main() async {
+  var certificateChain =
+      Platform.script.resolve('server_chain.pem').toFilePath();
+  var serverKey = Platform.script.resolve('server_key.pem').toFilePath();
+  [!var serverContext = new SecurityContext(); /*1*/!]
```
kwalrath: This new syntax is so much easier for me to read!

chalin: I'm very glad that you like it. I juggled with a few options, but this was the simplest to deal with and using /*#*/ seems natural enough a marker for code.

chalin:
> fix the formatting of the command-line snippets? "Wrote data for Han Solo." is unnecessarily colorized.

Let's address this separately. I've created #466 to track this issue.

examples/httpserver/bin/basic_writer_client.dart
```diff
@@ -20,11 +21,11 @@ Future main() async {
   };
 
   var request = await new HttpClient()
-      .post(InternetAddress.LOOPBACK_IP_V4.host, 4049, '/file.txt');
-  request.headers.contentType = ContentType.JSON;
-  request.write(JSON.encode(jsonData));
-  HttpClientResponse response = await request.close();
-  await for (var contents in response.transform(UTF8.decoder)) {
+      /*1*/ .post(InternetAddress.LOOPBACK_IP_V4.host, 4049, 'file.txt');
```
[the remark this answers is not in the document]
chalin: done

src/_tutorials/dart-vm/httpserver.md
```diff
@@ -154,21 +152,22 @@ listening, and responding.
 
 ## Binding a server to a host and port {#binding}
 
-_Example for this section: hello_world_server.dart._
+_Example for this section:_
+[hello_world_server.dart.]({{gh-path}}/bin/hello_world_server.dart)
 
 The first line of code in `main()` uses `HttpServer.bind()` to create an
```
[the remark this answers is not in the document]
chalin: done

src/_tutorials/dart-vm/httpserver.md
```diff
@@ -154,21 +152,22 @@ listening, and responding.
...
-...
+var server = await HttpServer.bind(
+  InternetAddress.LOOPBACK_IP_V4,
+  4040,
+);
 {% endprettify %}
 <div class="prettify-filename">hello_world_server.dart</div>
 
 The code uses `await` to call the `bind` method asynchronously.
-When the bind is successful, the new HttpServer object is assigned
```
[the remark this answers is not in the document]
chalin: The name of the declared variable changed and so that line of prose required editing. I decided to just drop that sentence because IMHO: (1) The code self explanatory. (2) This is especially the case given that the paragraph immediately before the code snippet reads:
> The first statement in `main()` uses `HttpServer.bind()` to create an [HttpServer][] object and bind it to a host and port.

(Otherwise, the statement is essentially describing how a variable declaration with an initializer works.)

src/_tutorials/dart-vm/httpserver.md
```diff
@@ -204,17 +203,17 @@ The server begins listening for HTTP requests using `await for`.
 For each request received, the highlighted code is executed for that
```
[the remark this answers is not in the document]
chalin: Yes, that is better (and yes I had forgotten to add the highlighting in that snippet). I've changed it to:

> For each request received, the code sends a "Hello, world!" response.

src/_tutorials/dart-vm/httpserver.md
```diff
@@ -271,38 +270,32 @@ The form also specifies the URL, which includes the port number,
...
-[[note]]3[[/note]]   <select [[highlight]]name="q"[[/highlight]]>
-        <option value="0">0</option>
-        <option value="1">1</option>
-        <option value="2">2</option>
-        ...
-        <option value="9">9</option>
-      </select>
-[[note]]4[[/note]]   <input [[highlight]]type="submit"[[/highlight]] value="Guess">
-    </form>
+<form [!action="http://localhost:4041"!] [!method="GET"!]>
```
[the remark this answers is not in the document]
chalin: No, it is because we can't put HTML comments inside tags (which is effectively what we would need to do here). 

An alternative could be to have two markers on a single line like this:
```html
<form action="http://localhost:4041" method="GET"> <!--1--> <!--2-->
  <select name="q"> <!--3-->
    <option value="0">0</option>
    <option value="1">1</option>
    <option value="2">2</option>
    <!-- ··· -->
    <option value="9">9</option>
  </select>
  <input type="submit" value="Guess"> <!--4-->
</form>
```

That doesn't seem very readable.

Other, more elaborate solutions (e.g., involving tooltip popups) might be possible, but I didn't think that it is worth it for just a single example. WDYT?

src/_tutorials/dart-vm/httpserver.md
```diff
@@ -271,38 +270,32 @@ The form also specifies the URL, which includes the port number,
...
-kinds of requests include POST, PUT, and DELETE.
-
-<span class="code-note">3</span>
-Any element within the form that has a name becomes
-a parameter in the query string.
-
-<span class="code-note">4</span>
-When pressed, the submit button formulates
-the request based on the content of the form and sends it.
+- The form's `action` attribute is assigned the
```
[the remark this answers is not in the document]
chalin: Yes! Done.

src/_tutorials/dart-vm/httpserver.md
```diff
@@ -271,38 +270,32 @@ The form also specifies the URL, which includes the port number,
...
+  URL to send the request to.
+- The form's `method` attribute defines
+  the kind of request, here a `GET`. Other common
+  kinds of request include POST, PUT, and DELETE.
+- Any element within the form, like `<select>`, that has a `name` becomes
```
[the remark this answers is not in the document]
chalin: done

src/_tutorials/dart-vm/httpserver.md
```diff
@@ -555,66 +560,55 @@ The `close()` method sends the request to the server
...
+  Map jsonData = {
+    'name': 'Han Solo',
+    'job': 'reluctant hero',
+    'BFF': 'Chewbacca',
+    'ship': 'Millennium Falcon',
+    'weakness': 'smuggling debts'
+  };
+
+  var request = await new HttpClient()
+      [!/*1*/!] .post(InternetAddress.LOOPBACK_IP_V4.host, 4049, 'file.txt');
```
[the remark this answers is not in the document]
chalin: Yes, in part because of line issues, but also because dartfmt doesn't make it easy to place the first marker close to the `post(` tokens. I've moved the markers. Let me know what you think about their new positions.

src/_tutorials/dart-vm/httpserver.md
chalin: Actually, now that the `/*#*/` markers are at the end of the line, it can be easy to miss them. I've added some highlighting to all the lines. WDYT?

src/_tutorials/dart-vm/httpserver.md
chalin: For the same reasons that I removed a similar line earlier in the page: see [this comment](https://github.com/dart-lang/site-www/pull/463/files#r155930657):

1. The code is self-explanatory.
2. We've already explained what `await` does.
3. I don't think that we need to describe how a variable assignment occurs.

Let me know if you'd like for me to add it back.

src/_tutorials/dart-vm/httpserver.md
chalin: Good catch! (This is another reason why I like to avoid mentioning variable names in the prose.)

Actually, I've dropped the last part of the sentence. It now reads:

> The `close()` method sends the request to the server and, when complete,  returns an [HttpClientResponse][] object.

The prose should be commenting on the essential parts of the code and skip uninteresting details.

src/_tutorials/dart-vm/httpserver.md
chalin: The original statement was (highlight mine):

> The response from the server is **encoded** in UTF-8.

Most of the code notes refer to actions performed by the `basic_writer_client`. It just felt a little off to mention an **encoding** action (which is performed elsewhere), when the transform being applied here by the basic writer client is **decoding**.

Let me know what you prefer.

src/_tutorials/dart-vm/httpserver.md
chalin: I've dropped the \<aside> (falling back to regular paragraphs), replacing it with:

> The client in this example makes REST-compliant POST requests.
>
> To see client code that makes REST-compliant GET requests
look at [number_guesser.dart.]({{gh-path}}/bin/number_guesser.dart)
It's a standalone client for the number thinker server
that makes periodic guesses until it guesses correctly.

WDYT?

src/_tutorials/dart-vm/httpserver.md
chalin: Yes! Done.

chalin: done

src/_tutorials/dart-vm/httpserver.md
chalin: Yes, the analyzer will complain, specifically about the following lint rule:

**&nbsp;&nbsp;lint • Type annotate public APIs at ... • type_annotate_public_apis**

With the increased importance of static type annotations in Dart 2, I feel that we get more benefits from enforcing the `type_annotate_public_apis` lint than any gains we achieve (like more terse code) from excluding it.

Without the return type, I find it easy to quickly scan the method signature `main() async { ... }` and "read" `main() { ... }`. The return type sticks out like a warning sign, alerting the reader to the fact that main is asynchronous. 

> It obscures the entry point

I find that the code highlighting helps make it just as easy to locate the entry point, because `main()` is in black font, whereas the tokens around it are more lightly colored.

chalin:
Updates completed. Staging server refreshed.

examples/httpserver/bin/hello_world_server.dart
chalin: #467

chalin: Would you prefer that we point to the files in this repo? The files themselves will contain `#docregion` and `#docplaster` tags. As for a zip of the projects, it can be relatively easy to create a zip of the sources with those tags removed. We could then host those project .zip files from dartlang.org (along with the rest of the site files).

src/_tutorials/dart-vm/httpserver.md
```diff
@@ -5,7 +5,6 @@ description: Communicate over the internet
 prevpage:
   url: /tutorials/dart-vm/cmdline
   title: "Dart-VM: Write Command-Line Apps"
-css: ["httpserver.css"]
 ---
 {% capture gh-path -%}
   https://github.com/dart-lang/dart-tutorials-samples/blob/master/httpserver
```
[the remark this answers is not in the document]
chalin: > we should probably update that repo

Yes, I'll do that for now (after this PR is accepted), and we'll keep this URL as is (so that readers will be directed to sources w/o `#docregion` tags).

examples/httpserver/bin/hello_world_server_secure.dart
```diff
@@ -14,10 +14,10 @@
 import 'dart:async';
 import 'dart:io';
 
+String certificateChain = 'server_chain.pem';
+String serverKey = 'server_key.pem';
+
 Future main() async {
-  var certificateChain =
-      Platform.script.resolve('server_chain.pem').toFilePath();
-  var serverKey = Platform.script.resolve('server_key.pem').toFilePath();
```
chalin: This server script, with its use of `Platform.script`, doesn't lend itself well to testing. We don't use `Platform.script` in any other app, and it is beside the point that we are trying to illustrate here, so I dropped it.

src/_tutorials/dart-vm/httpserver.md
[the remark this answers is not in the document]
kwalrath: Yep, makes sense to keep that repo for the cleaned up version of these files.

examples/httpserver/bin/hello_world_server.dart
kwalrath: I don't feel strongly about where we point to, as long as (1) those tags aren't visible and (2) we're careful about keeping everything in sync.

src/_tutorials/dart-vm/httpserver.md
kwalrath: I see `contentType?.mimeType == 'application/json'` in the new code.

From ancient code reviews (don't remember whether they were even in Dart), I remember being told to use the constant as the LHS, so you don't have to worry about invoking a method on null. Is that not a concern?

chalin: No, that isn't a problem in Dart.

[Operator `==` is defined in `Object`](https://api.dartlang.org/stable/1.24.2/dart-core/Object/operator_equals.html), and hence it is defined for all types, including `Null`.

src/_tutorials/dart-vm/httpserver.md
kwalrath: I think my objection was to the passive voice. How about:

A transformer defined in the ... library converts the data...

examples/httpserver/bin/basic_writer_client.dart
```diff
@@ -20,11 +21,11 @@ Future main() async {
   };
 
   var request = await new HttpClient()
-      .post(InternetAddress.LOOPBACK_IP_V4.host, 4049, '/file.txt');
-  request.headers.contentType = ContentType.JSON;
-  request.write(JSON.encode(jsonData));
-  HttpClientResponse response = await request.close();
-  await for (var contents in response.transform(UTF8.decoder)) {
+      .post(InternetAddress.LOOPBACK_IP_V4.host, 4049, 'file.txt'); /*1*/
```
kwalrath: At the window width I use when reviewing (just barely 3-column), this line is cut off. The last fully visible character is the `;`.

examples/httpserver/bin/basic_writer_client.dart
```diff
@@ -20,11 +21,11 @@ Future main() async {
...
+  request.headers.contentType = ContentType.JSON; /*2*/
+  request.write(JSON.encode(jsonData)); /*3*/
+  HttpClientResponse response = await request.close(); /*4*/
+  await for (var contents in response.transform(UTF8.decoder /*5*/)) {
```
kwalrath: At my window width, I can't see anything to the right of the `))`.

src/_tutorials/dart-vm/httpserver.md
```diff
@@ -128,21 +121,26 @@ In the code for the hello world server,
...
-
-main() async {
-  var requestServer =
-      await HttpServer.bind(InternetAddress.LOOPBACK_IP_V4, 4040);
-  print('listening on localhost, port ${requestServer.port}');
-
-  await for (HttpRequest request in requestServer) {
-    request.response..write('Hello, world!')
-                    ..close();
+import 'dart:io';
```
kwalrath: Highlight is missing on this line.

src/_tutorials/dart-vm/httpserver.md
kwalrath: I think the main reason to keep it would be to emphasize that the returned value is a HttpClientRequest, since that isn't visible in the code. (You mention that type later, but you don't explicitly tie it to this variable.)

src/_tutorials/dart-vm/httpserver.md
```diff
@@ -555,66 +561,54 @@ The `close()` method sends the request to the server
...
+  Map jsonData = {
+    'name': 'Han Solo',
+    'job': 'reluctant hero',
+    'BFF': 'Chewbacca',
+    'ship': 'Millennium Falcon',
+    'weakness': 'smuggling debts'
+  };
+
+  var request = await new HttpClient()
+      .[!post!](InternetAddress.LOOPBACK_IP_V4.host, 4049, 'file.txt'); [!/*1*/!]
```
kwalrath: I can't see anything after the `;` in this line.

src/_tutorials/dart-vm/httpserver.md
```diff
@@ -656,66 +649,66 @@ join() method in Stream to concatenate the string values of those chunks.
...
-      } else {
-        req.response..statusCode = HttpStatus.METHOD_NOT_ALLOWED
-                    ..write("Unsupported request: ${req.method}.")
-                    ..close();
+import 'dart:async';
+import 'dart:io';
+import 'dart:convert';
+
+Future main() async {
+  var server = await HttpServer.bind(InternetAddress.LOOPBACK_IP_V4, 4049);
```
kwalrath: The last visible characters on this line are `V4,`.

src/_tutorials/dart-vm/httpserver.md
```diff
@@ -656,66 +649,66 @@ join() method in Stream to concatenate the string values of those chunks.
...
+  await for (var req in server) {
+    ContentType contentType = req.headers.contentType;
+    HttpResponse response = req.response;
+
+    if (req.method == 'POST' &&
+        [!contentType?.mimeType == 'application/json'!] [!/*1*/!]) {
+      try {
+        var [!jsonString = await req.transform(UTF8.decoder).join();!] [!/*2*/!]
```
kwalrath: Last visible characters: `();`

src/_tutorials/dart-vm/httpserver.md
```diff
@@ -727,11 +720,11 @@ allows POST and OPTIONS requests from any origin.
 Use CORS headers with caution,
 because they can open your network up to security risks.
 
+<?code-excerpt "httpserver/bin/note_server.dart (addCorsHeaders)"?>
 {% prettify dart %}
 void addCorsHeaders(HttpResponse response) {
   response.headers.add('Access-Control-Allow-Origin', '*');
-  response.headers.add(
-      'Access-Control-Allow-Methods', 'POST, OPTIONS');
+  response.headers.add('Access-Control-Allow-Methods', 'POST, OPTIONS');
```
kwalrath: line is cut off

src/_tutorials/dart-vm/httpserver.md
kwalrath: The difference between the type and the method name isn't very noticeable to me:

image

src/_tutorials/dart-vm/httpserver.md
kwalrath: Much better, thanks!

src/_tutorials/dart-vm/httpserver.md
kwalrath: @filiph what do you think we should do? I don't like turning off lints. Maybe we can get a special case for "main" in the lint?

src/_tutorials/dart-vm/httpserver.md
chalin: Yes, good point. I've added a type annotation to the local request variable. Note that I'm not setting a precedent here:

- A little earlier in the page, in the first `number_thinker.dart` excerpt of the "Listening for and handling requests" section, the local variable `server` has a type annotation.
- The `number_guesser.dart` client also has its local `request` variable typed.

src/_tutorials/dart-vm/httpserver.md
chalin: Just noting that this issue isn't gating. (We are still discussing this offline.)

examples/httpserver/bin/basic_writer_client.dart
chalin: fixed

examples/httpserver/bin/basic_writer_client.dart
chalin: This one is a bit trickier to fix. I've reworked the code, PTAL.

src/_tutorials/dart-vm/httpserver.md
chalin: Added back.

src/_tutorials/dart-vm/httpserver.md
chalin: I've reworked the code, PTAL.

src/_tutorials/dart-vm/httpserver.md
chalin: fixed

src/_tutorials/dart-vm/httpserver.md
chalin: fixed

src/_tutorials/dart-vm/httpserver.md
chalin: fixed

src/_tutorials/dart-vm/httpserver.md
```diff
@@ -702,14 +702,14 @@ Future main() async {
    characters can be encoded over multiple bytes. The join() method puts the
    chunks together.
 
-3. The URL for the request is
+3. The data sent by the client is JSON formatted. The server decodes it using
+   the JSON codec available in the [dart:convert][] library.
+
+4. The URL for the request is
```
chalin: FYI, the text of these bullets has not changed. I've just swapped the steps because of a reordering of statements in the code.

PR #465, opened 2017-12-09
Minor copy edit of pubspec page

kwalrath:
A recent change made the TOC a bit awkward. This fixes that, and a few more things I happened to notice.

Staged at:

https://kw-www-dartlang-1.firebaseapp.com/tools/pub/pubspec


Sfshaza:
Lots of little changes! I wonder if we should replace the Polymer example?

Overall, LGTM!

kwalrath:
Good point about the polymer examples! I've filed #469 to track that.

PR #474, opened 2017-12-13
code-excerpts for Effective Dart: styles, docs, usage

chalin:
Contributes to #407, covering Effective Dart, Styles.

- Staged at https://dartlang-org-staging-0.firebaseapp.com/guides/language/effective-dart/style
- Current https://www.dartlang.org/guides/language/effective-dart/style

### DIFF
--- a/src/_guides/language/effective-dart/documentation.md
+++ b/src/_guides/language/effective-dart/documentation.md
@@ -1,14 +1,13 @@
 ---
 layout: guide
 title: "Effective Dart: Documentation"
-description: "Clear, helpful comments and documentation."
-
+description: Clear, helpful comments and documentation.
 nextpage:
   url: /guides/language/effective-dart/usage
-  title: "Usage"
+  title: Usage
 prevpage:
   url: /guides/language/effective-dart/style
-  title: "Style"
+  title: Style
 ---
 
 It's easy to think your code is obvious today without realizing how much you
@@ -32,12 +31,11 @@ generated documentation.
 
 ### DO format comments like sentences.
 
-<div class="good">
+{:.good-style}
 {% prettify dart %}
 // Not if there is nothing before it.
 if (_chunks.isEmpty) return false;
 {% endprettify %}
-</div>
 
 Capitalize the first word unless it's a case-sensitive identifier. End it with a
 period (or "!" or "?", I suppose). This is true for all comments: doc comments,
@@ -45,23 +43,21 @@ inline stuff, even TODOs. Even if it's a sentence fragment.
 
 ### DON'T use block comments for documentation.
 
-<div class="good">
+{:.good-style}
 {% prettify dart %}
 greet(name) {
   // Assume we have a valid name.
   print('Hi, $name!');
 }
 {% endprettify %}
-</div>
 
-<div class="bad">
+{:.bad-style}
 {% prettify dart %}
 greet(name) {
   /* Assume we have a valid name. */
   print('Hi, $name!');
 }
 {% endprettify %}
-</div>
 
 You can use a block comment (`/* ... */`) to temporarily comment out a section
 of code, but all other comments should use `//`.
@@ -80,19 +76,17 @@ before a declaration and uses the special `///` syntax that dartdoc looks for.
 Using a doc comment instead of a regular comment enables [dartdoc][] to find it
 and generate documentation for it.
 
-<div class="good">
+{:.good-style}
 {% prettify dart %}
 /// The number of characters in this chunk when unsplit.
 int get length => ...
 {% endprettify %}
-</div>
 
-<div class="bad">
+{:.bad-style}
 {% prettify dart %}
 // The number of characters in this chunk when unsplit.
 int get length => ...
 {% endprettify %}
-</div>
 
 For historical reasons, dartdoc supports two syntaxes of doc comments: `///`
 ("C# style") and `/** ... */` ("JavaDoc style"). We prefer `///` because it's
@@ -119,7 +113,7 @@ other parts of the library.
 The first paragraph of any doc comment is a brief, user-oriented description
 ending with a period. As you can see below, it is often not a complete sentence.
 
-<div class="good">
+{:.good-style}
 {% prettify dart %}
 /// Defines a flag.
 ///
@@ -127,15 +121,13 @@ ending with a period. As you can see below, it is often not a complete sentence.
 /// there is already an option using abbreviation [abbr]. Returns the new flag.
 Flag addFlag(String name, String abbr) { ... }
 {% endprettify %}
-</div>
 
-<div class="bad">
+{:.bad-style}
 {% prettify dart %}
 /// Starts a new block as a child of the current chunk. Nested blocks are
 /// handled using their own independent [LineWriter].
 ChunkBuilder startBlock() { ... }
 {% endprettify %}
-</div>
 
 The description should help the reader understand whether this API might
 be useful to them, compared to similar-sounding APIs. Don't just repeat the API
@@ -145,7 +137,7 @@ name&mdash;tell the reader something they don't already know.
 
 The doc comment should focus on what the code *does*.
 
-<div class="good">
+{:.good-style}
 {% prettify dart %}
 /// Returns `true` if every element satisfies the [predicate].
 bool all(bool predicate(T element)) { ... }
@@ -153,7 +145,6 @@ bool all(bool predicate(T element)) { ... }
 /// Starts the stopwatch if not already running.
 void start() { ... }
 {% endprettify %}
-</div>
 
 ### PREFER starting variable, getter, or setter comments with noun phrases.
 
@@ -161,7 +152,7 @@ The doc comment should stress what the property *is*. This is true even for
 getters which may do calculation or other work. What the caller cares about is
 the *result* of that work, not the work itself.
 
-<div class="good">
+{:.good-style}
 {% prettify dart %}
 /// The current day of the week, where `0` is Sunday.
 int weekday;
@@ -169,7 +160,6 @@ int weekday;
 /// The number of checked buttons on the page.
 int get checkedCount { ... }
 {% endprettify %}
-</div>
 
 If there's both a setter and a getter, comment only the getter. That way,
 dartdoc will treat it like a variable.
@@ -181,25 +171,23 @@ program. They describe the type's invariants, establish the terminology it uses,
 and provide context to the other doc comments for the class's members. A little
 extra effort here can make all of the other members simpler to document.
 
-<div class="good">
+{:.good-style}
 {% prettify dart %}
 /// A chunk of non-breaking output text terminated by a hard or soft newline.
 ///
 /// ...
 class Chunk { ... }
 {% endprettify %}
-</div>
 
 ### CONSIDER including code samples in doc comments.
 
-<div class="good">
+{:.good-style}
 {% prettify dart %}
 /// Returns the lesser of two numbers.
 ///
 ///     min(5, 3); // 3.
 num min(num a, num b) { ... }
 {% endprettify %}
-</div>
 
 Humans are great at generalizing from examples, so even a single code sample
 makes an API easier to learn.
@@ -209,28 +197,26 @@ makes an API easier to learn.
 If you surround things like variable, method, or type names in square brackets,
 then dartdoc will look up the name and link to its docs.
 
-<div class="good">
+{:.good-style}
 {% prettify none %}
 Throws a [StateError] if ...
 
 similar to [anotherMethod], but ...
 {% endprettify %}
-</div>
 
 You can also link to constructors using `new` followed by the constructor:
 
-<div class="good">
+{:.good-style}
 {% prettify none %}
 To create a point, call [new Point] or use [new Point.polar] to ...
 {% endprettify %}
-</div>
 
 ### DO use prose to explain parameters, return values, and exceptions.
 
 Other languages use verbose tags and sections to describe what the parameters
 and returns of a method are.
 
-<div class="bad">
+{:.bad-style}
 {% prettify dart %}
 /// Defines a flag with the given name and abbreviation.
 ///
@@ -241,12 +227,11 @@ and returns of a method are.
 ///     the given name or abbreviation.
 Flag addFlag(String name, String abbr) { ... }
 {% endprettify %}
-</div>
 
 The convention in Dart is to integrate that into the description of the method
 and highlight parameters using square brackets.
 
-<div class="good">
+{:.good-style}
 {% prettify dart %}
 /// Defines a flag.
 ///
@@ -254,7 +239,6 @@ and highlight parameters using square brackets.
 /// there is already an option using abbreviation [abbr]. Returns the new flag.
 Flag addFlag(String name, String abbr) { ... }
 {% endprettify %}
-</div>
 
 ### AVOID redundantly mentioning types in doc comments.
 
@@ -266,21 +250,19 @@ Tell the reader something they *don't* already know.
 
 ### DO put doc comments before metadata annotations.
 
-<div class="good">
+{:.good-style}
 {% prettify dart %}
 /// _Deprecated: Use [newMethod] instead._
 @deprecated
 oldMethod();
 {% endprettify %}
-</div>
 
-<div class="bad">
+{:.bad-style}
 {% prettify dart %}
 @deprecated
 /// _Deprecated: Use [newMethod] instead._
 oldMethod();
 {% endprettify %}
-</div>
 
 
 ## Markdown
@@ -376,7 +358,7 @@ think.
 When documenting a member for a class, you often need to refer back to the
 object the member is being called on. Using "the" can be ambiguous.
 
-<div class="good">
+{:.good-style}
 {% prettify dart %}
 class Box {
   /// The value this wraps.
@@ -386,5 +368,4 @@ class Box {
   bool get hasValue => _value != null;
 }
 {% endprettify %}
-</div>
 
--- a/src/_guides/language/effective-dart/documentation.md
+++ b/src/_guides/language/effective-dart/documentation.md
@@ -9,6 +9,7 @@ prevpage:
   url: /guides/language/effective-dart/style
   title: Style
 ---
+<?code-excerpt replace="/=. blockEllipsis;/{ ... }/g;/ellipsis;/.../g;/\/\*(\s*\.\.\.\s*)\*\//$1/g"?>
 
 It's easy to think your code is obvious today without realizing how much you
 rely on context already in your head. People new to your code, and
@@ -32,6 +33,7 @@ generated documentation.
 ### DO format comments like sentences.
 
 {:.good-style}
+<?code-excerpt "misc/lib/effective_dart/docs_good.dart (comments-like-sentences)"?>
 {% prettify dart %}
 // Not if there is nothing before it.
 if (_chunks.isEmpty) return false;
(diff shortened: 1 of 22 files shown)


kwalrath:
@munificent I'll review this, but would you also like to take a look? (Some examples changed.) You can see the diffs starting at https://github.com/dart-lang/site-www/pull/474/files#diff-94aed4fc336e2c2c764ff10fc6160047L1.

src/_guides/language/effective-dart/documentation.md
```diff
@@ -209,28 +209,28 @@ makes an API easier to learn.
...
 
-<div class="good">
+{:.good-style}
+<?code-excerpt "misc/lib/effective_dart/docs_good.dart (identifiers)"?>
 {% prettify none %}
-Throws a [StateError] if ...
-
-similar to [anotherMethod], but ...
+/// Throws a [StateError] if ...
+/// similar to [anotherMethod], but ...
```
chalin: I added leading `///`, otherwise the prettifier would colorize the content of the comments (as if it were code).

src/_guides/language/effective-dart/documentation.md
```diff
@@ -209,28 +209,28 @@ makes an API easier to learn.
...
-</div>
 
 You can also link to constructors using `new` followed by the constructor:
 
-<div class="good">
+{:.good-style}
+<?code-excerpt "misc/lib/effective_dart/docs_good.dart (new)"?>
 {% prettify none %}
-To create a point, call [new Point] or use [new Point.polar] to ...
+/// To create a point, call [new Point] or use [new Point.polar] to ...
```
chalin: I added leading `///`, otherwise the prettifier would colorize the content of the comments (as if it were code).

src/_guides/language/effective-dart/style.md
```diff
@@ -87,47 +86,49 @@ in that form. Using underscores as the separator ensures that the name is still
...
 
-<div class="good">
+{:.good-style}
+<?code-excerpt "misc/lib/effective_dart/style_lib_good.dart (import-as)" replace="/(package):examples[^']*/$1:angular_components\/angular_components/g"?>
 {% prettify dart %}
-import 'dart:json' as json;
 import 'dart:math' as math;
-import 'package:javascript_utils/javascript_utils.dart' as js_utils;
+import 'package:angular_components/angular_components'
+    as angular_components;
```
chalin: Switching to `angular_components` seemed more du jour :)

src/_guides/language/effective-dart/style.md
```diff
@@ -362,34 +356,34 @@ Doing so avoids the [dangling else][] problem.
 
 [dangling else]: http://en.wikipedia.org/wiki/Dangling_else
 
-<div class="good">
+{:.good-style}
+<?code-excerpt "misc/lib/effective_dart/style_good.dart (curly-braces)"?>
 {% prettify dart %}
-if (true) {
-  print('sanity');
+if (isWeekDay) {
```
chalin: The analyzer was (rightfully) reporting dead code. To avoid having to turn that check off, I changed the example.

src/_guides/language/effective-dart/style.md
```diff
@@ -362,34 +356,34 @@ Doing so avoids the [dangling else][] problem.
...
 These are typically used for "guard" code that returns or breaks if the
 condition is met. But they're also fine for expressions, as long as the entire
 `if` statement and the expression fit on one line.
 
-<div class="good">
+{:.good-style}
+<?code-excerpt "misc/lib/effective_dart/style_good.dart (one-line-if-expr)"?>
 {% prettify dart %}
-if (parameter == null) parameter = defaultValue;
+if (parameter > limit) parameter = defaultValue;
```
chalin: `if (foo == null) foo = expr` would more succinctly be written as `foo ??= expr`, so I changed the example a bit (since we're meant to be illustrating single-line `if` statements).

src/_guides/language/effective-dart/style.md
```diff
@@ -450,114 +444,114 @@ but not for specifying generic types. Both `is` and `is!` are considered single
...
 *not* have spaces around it.
 
-<div class="good">
+{:.good-style}
+<?code-excerpt "misc/lib/effective_dart/style_good.dart (bin-op)"?>
 {% prettify dart %}
 average = (a + b) / 2;
 largest = a > b ? a : b;
 if (obj is! SomeType) print('not SomeType');
-optional([parameter = defaultValue]) { ... }
```
chalin: This section is about "DO place spaces around binary and ternary operators." The `optional(...)` seemed to add nothing that wasn't already illustrated by the previous examples. If this assumption is wrong, let me know.

src/_guides/language/effective-dart/style.md
```diff
@@ -450,114 +444,114 @@ but not for specifying generic types. Both `is` and `is!` are considered single
...
 ### DO place the opening curly brace (`{`) on the same line as what it follows.
 
-<div class="good">
+{:.good-style}
+<?code-excerpt "misc/lib/effective_dart/style_good.dart (open-curly-brace-same-line)" replace="/(Foo)2/$1/g"?>
 {% prettify dart %}
 class Foo {
   method() {
-    if (true) {
+    if (someCondition) {
```
chalin: To avoid dead-code warnings from the analyzer.

src/_guides/language/effective-dart/style.md
```diff
@@ -566,35 +560,35 @@ var bobLikesIt = isDeepFried ||
...
 ### DO place the `.` on the next line in a multi-line expression.
 
-<div class="good">
+{:.good-style}
+<?code-excerpt "misc/lib/effective_dart/style_good.dart (multi-dot)"?>
 {% prettify dart %}
-someVeryLongVariable.withAVeryLongProperty
-    .aMethodOnThatObject();
+someVeryLongVariableName.withAVeryLongPropertyName
+    .aReallyLongMethodName(args);
```
chalin: There are two sections using very long identifier names. This was changed to help unify the use of variables names in both sections.

src/_guides/language/effective-dart/style.md
```diff
@@ -628,29 +622,29 @@ mapInsideList([
...
-    list,
-    literal
-  ];
 }
+
+var compoundsWithLongNames = [
+  buckminsterfullerene,
+  dodecahedrane,
+  olympiadane
+];
```
chalin: Rather than force line breaks, I opted for the use of long variable names. If you'd prefer that I revert to the original list elements, let me know.

src/_guides/language/effective-dart/style.md
```diff
@@ -662,43 +656,43 @@ switch (fruit) {
...
 ### PREFER indenting continued lines with at least four spaces.
 
-<div class="good">
+{:.good-style}
+<?code-excerpt "misc/lib/effective_dart/style_good.dart (four-spaces)"?>
 {% prettify dart %}
-someLongObject.aReallyLongMethodName(longArg, anotherLongArg,
-    wrappedToNextLine);
+someVeryLongVariableName.aReallyLongMethodName(
+    arg, anotherArg, wrappedToNextLine);
```
chalin: There are two sections using very long identifier names. This was changed to help unify the use of variables names in both sections.

src/_guides/language/effective-dart/usage.md
```diff
@@ -30,59 +31,58 @@ form&mdash;you do not need to use `+` to concatenate them. Just like in C and
...
 {% endprettify %}
-</div>
 
-<div class="bad">
+{:.bad-style}
+<?code-excerpt "misc/lib/effective_dart/usage_bad.dart (adjacent-strings-literals)"?>
 {% prettify dart %}
-raiseAlarm(
-    'ERROR: Parts of the spaceship are on fire. Other ' +
+raiseAlarm('ERROR: Parts of the spaceship are on fire. Other ' +
```
chalin: This is how dartfmt formats this now.

src/_guides/language/effective-dart/usage.md
```diff
@@ -30,59 +31,58 @@ form&mdash;you do not need to use `+` to concatenate them. Just like in C and
...
 'Hello, $name! You are ${year - birth} years old.';
 {% endprettify %}
-</div>
 
-<div class="bad">
+{:.bad-style}
+<?code-excerpt "misc/lib/effective_dart/usage_bad.dart (string-interpolation)"?>
 {% prettify dart %}
-'Hello, ' + name + '! You are ' + (year - birth) + ' years old.';
+'Hello, ' + name + '! You are ' + (year - birth) + ' y...';
```
chalin: I've elided the last string because it's full value isn't essential to this example. Then the full string is given, dartfmt spreads this long String expression onto multiple lines.

src/_guides/language/effective-dart/usage.md
```diff
@@ -371,28 +372,27 @@ class Circle {
   }
 
   void _recalculate() {
-    _area = math.PI * _radius * _radius,
+    _area = math.PI * _radius * _radius;
```
chalin: Found a first syntax error! Which is pretty good, given the number of excerpts in this site!

src/_guides/language/effective-dart/usage.md
```diff
@@ -371,28 +372,27 @@ class Circle {
...
 {% prettify dart %}
 class Circle {
   num radius;
 
+  Circle(this.radius);
+
   num get area => math.PI * radius * radius;
   num get circumference => math.PI * 2.0 * radius;
-
-  Circle(this.radius);
```
chalin: I moved the constructor so as to follow our own guidelines :) ... otherwise the analyzer (w/ lint rules enabled) complains.

src/_guides/language/effective-dart/usage.md
```diff
@@ -567,23 +568,21 @@ class Box {
...
+<?code-excerpt "misc/lib/effective_dart/usage_good.dart (param-dont-shadow-field-ctr-init)"?>
 {% prettify dart %}
 class Box extends BaseBox {
   var value;
 
   Box(value)
       : value = value,
-        super(value)
-      {}
+        super(value);
```
chalin: I dropped `{}` so as to follow our own guidelines :) ... otherwise the analyzer (w/ lint rules enabled) complains.

src/_guides/language/effective-dart/usage.md
```diff
@@ -721,22 +720,32 @@ then. That always happens after all initializers are run regardless of where
 useful. In fact, [dartdevc]({{site.webdev}}/tools/dartdevc)
 *requires* that it appear last.
 
-<div class="good">
+{% comment %}
+Question for Kathy: I've added a "Dart 2.0 note" below.
+Do we need to keep the last sentence above?
+{% endcomment %}
```
chalin: Question for @kwalrath

src/_guides/language/effective-dart/usage.md
```diff
@@ -721,22 +720,32 @@ then. That always happens after all initializers are run regardless of where
...
+    : super(style), // Error in Dart 2.0
+      _children = children;
 {% endprettify %}
-</div>
 
+{% comment %}update-for-dart-2{% endcomment %}
+<div class="alert alert-danger" markdown="1">
+**Dart 2.0 note:** The call to `super()` must be _last_ in Dart 2.0, otherwise
+an error is reported.
+</div>
```
chalin: @kwalrath - I've added a Dart 2.0 note here. Is this ok?

src/_guides/language/effective-dart/usage.md
```diff
@@ -721,22 +720,32 @@ then. That always happens after all initializers are run regardless of where
...
-</div>
 
-<div class="bad">
+{:.bad-style}
+<?code-excerpt "misc/lib/effective_dart/usage_bad.dart (super-first)"?>
 {% prettify dart %}
 View(Style style, List children)
-    : super(style),
-      _children = children {
+    : super(style), // Error in Dart 2.0
```
chalin: Do we also want this comment in the code?

src/_guides/language/effective-dart/usage.md
```diff
@@ -811,27 +813,27 @@ instead of throwing the same exception object using `throw`.
...
 
-<div class="bad">
+{:.bad-style}
+<?code-excerpt "misc/lib/effective_dart/usage_bad.dart (rethrow)" replace="/throw e/[!$&!]/g"?>
 {% prettify dart %}
 try {
   somethingRisky();
-} catch(e) {
-  if (!canHandle(e)) throw e;
+} catch (e) {
```
chalin: The `catch(e)` to `catch (e)` change here (and in the next few examples below) was done by dartfmt.

src/_guides/language/effective-dart/usage.md
```diff
@@ -811,27 +813,27 @@ instead of throwing the same exception object using `throw`.
...
-<div class="good">
+{:.good-style}
+<?code-excerpt "misc/lib/effective_dart/usage_good.dart (rethrow)" replace="/rethrow/[!$&!]/g"?>
 {% prettify dart %}
 try {
   somethingRisky();
-} catch(e) {
-  if (!canHandle(e)) rethrow;
+} catch (e) {
+  if (!canHandle(e)) [!rethrow!];
```
chalin: There are many lines in this example. I've added highlighting to help focus the reader's attention.

src/_guides/language/effective-dart/usage.md
```diff
@@ -872,29 +875,28 @@ Future<bool> doAsyncComputation() {
...
 It's easy to get in the habit of using `async` on any function that does
 anything related to asynchrony. But in some cases, it's extraneous. If you can
 omit the `async` without changing the behavior of the function, do so.
 
-<div class="good">
+{:.good-style}
+<?code-excerpt "misc/lib/effective_dart/usage_good.dart (unnecessary-async)"?>
 {% prettify dart %}
-Future afterTwoThings(Future first, second) {
+Future afterTwoThings(Future first, Future second) {
```
chalin: The analyzer complained when `second` was (implicitly) declared as `dynamic`.

src/_guides/language/effective-dart/usage.md
```diff
@@ -872,29 +875,28 @@ Future<bool> doAsyncComputation() {
...
 }
 {% endprettify %}
-</div>
 
-<div class="bad">
+{:.bad-style}
+<?code-excerpt "misc/lib/effective_dart/usage_bad.dart (unnecessary-async)"?>
 {% prettify dart %}
-Future afterTwoThings(Future first, second) async {
+Future afterTwoThings(Future first, Future second) async {
```
chalin: The analyzer complained when `second` was (implicitly) declared as `dynamic`.

src/_guides/language/effective-dart/usage.md
```diff
@@ -919,11 +922,8 @@ Future asyncError() async {
   throw 'Error!';
 }
 
-Future asyncValue() async {
-  return 'value';
-}
+Future asyncValue() async => 'value';
```
chalin: Changed to arrow function, per our own guidelines.

src/_guides/language/effective-dart/style.md
```diff
@@ -1,15 +1,14 @@
...
 nextpage:
   url: /guides/language/effective-dart/documentation
-  title: "Documentation"
+  title: Documentation
 prevpage:
   url: /guides/language/effective-dart
-  title: "Overview"
+  title: Overview
 ---
+<?code-excerpt  replace="/ellipsis;/.../g;/\/\/!//g;/\/\*(\s*\.\.\.\s*)\*\//$1/g"?>
```
kwalrath: My regex fu is lacking. What does this do?

chalin: There are three (semicolon separated) replace expressions:

- `/ellipsis;/.../g`, replaces `ellipsis;` with `...`.
- `/\/\/!//g`, erases `//!`, which is the marker I use to force a line break when `dartfmt` doesn't want to add one :).
- `/\/\*(\s*\.\.\.\s*)\*\//$1/g`: essentially replaces `/* ... */` with `...`.

src/_guides/language/effective-dart/style.md
```diff
@@ -420,27 +415,26 @@ main() {
...
 
 ### DON'T place a space between the declared name of a method, operator, or setter and its parameter list.
 
-<div class="good">
+{:.good-style}
+<?code-excerpt "misc/lib/effective_dart/style_good.dart (args-etc-no-spaces)"?>
 {% prettify dart %}
-bool convertToBool(arg) { ... }
-bool operator ==(other) { ... }
+log(arg) { ... }
```
kwalrath: Why did this change?

chalin: In this case, the method return type annotation (`bool`) seemed to distract from the "DON'T place a space between the declared name of a method, operator, or setter and its parameter list" guideline, so I switched to a method whose return type doesn't need to be shown.

src/_guides/language/effective-dart/usage.md
kwalrath: But it's OK to have `num radius` above? It seems weird to split the property declarations.

src/_guides/language/effective-dart/usage.md
kwalrath: I like the note, and I think we should remove that sentence about dartdevc.

src/_guides/language/effective-dart/usage.md
kwalrath: I'm on the fence about this. I suppose people might miss the note, but it's not super clear from the comment what exactly causes the error.

src/_guides/language/effective-dart/usage.md
kwalrath: I like the note, but I'd emphasize _must_ instead of "last", or maybe _must be last_.

src/_guides/language/effective-dart/usage.md
```diff
@@ -811,27 +813,27 @@ instead of throwing the same exception object using `throw`.
 `rethrow` preserves the original stack trace of the exception. `throw` on the
 other hand resets the stack trace to the last thrown position.
 
-<div class="bad">
+{:.bad-style}
+<?code-excerpt "misc/lib/effective_dart/usage_bad.dart (rethrow)" replace="/throw e/[!$&!]/g"?>
```
kwalrath: I think we shouldn't emphasize bad code (because human brains are funny), so let's not highlight "throw e".

src/_guides/language/effective-dart/usage.md
kwalrath: I do like emphasizing `rethrow`.

src/_guides/language/effective-dart/usage.md
chalin: The rule is [sort_constructors_first](http://dart-lang.github.io/linter/lints/sort_constructors_first.html):

> DO sort constructor declarations before method declarations.

So field declarations ahead of the constructor are ok. I sometimes feel that getters and setters seem like they should be on par with fields, but they are not.

src/_guides/language/effective-dart/usage.md
```diff
@@ -847,21 +849,22 @@ the language. They make a huge improvement in readability code and let you use
 all of the built-in control flow structures of the language within your
 asynchronous code.
 
-<div class="good">
+{:.good-style}
+<?code-excerpt "misc/lib/effective_dart/usage_good.dart (async-await)"?>
```
kwalrath: Let's highlight "async" and "await".

src/_guides/language/effective-dart/usage.md
chalin: done

src/_guides/language/effective-dart/usage.md
chalin: I was on the fence too. I'll drop the code comment.

src/_guides/language/effective-dart/usage.md
chalin: I had previously only emphasized _must_. Ok, let's emphasize "_must be last_".

src/_guides/language/effective-dart/usage.md
chalin: Highlight removed.

src/_guides/language/effective-dart/usage.md
chalin: Highlight added.

PR #476, opened 2017-12-14
${HOME} instead of ~ works with grinder or gradle

MikeMitterer:


kwalrath:
Thanks for improving the docs, @MikeMitterer!

chalin:
Also, how about using the shorter `$HOME` expression?

MikeMitterer:
@chalin Yes - ~ doesn't work in some cases. E.g. grinder. These tools need a full path to the bin dir. (I think Gradle had the same problem)
`$HOME` - would work too. I prefer the curly-brace... it avoids parsing problems but these case are very rare.

PR #480, opened 2017-12-15
Tests and code-excerpts for Effective Dart Design

chalin:
Contributes to #407, covering Effective Dart Desgin.

- Staged at https://dartlang-org-staging-0.firebaseapp.com/guides/language/effective-dart/design
- Current https://www.dartlang.org/guides/language/effective-dart/design

### DIFF
--- a/examples/misc/lib/effective_dart/design_bad.dart
+++ b/examples/misc/lib/effective_dart/design_bad.dart
@@ -8,7 +8,7 @@ void miscDeclAnalyzedButNotTested() {
   (errors, monsters, subscription) {
     // #docregion code-like-prose
     // Telling errors to empty itself, or asking if it is?
-    if (errors.empty) {/* ... */}
+    if (errors.empty) {/*-...-*/}
 
     // Toggle what? To what?
     subscription.toggle();
@@ -19,7 +19,7 @@ void miscDeclAnalyzedButNotTested() {
 
     Iterable theCollectionOfErrors;
     // #docregion code-like-prose-overdone
-    if (theCollectionOfErrors.isEmpty) {/* ... */}
+    if (theCollectionOfErrors.isEmpty) {/*-...-*/}
 
     monsters.producesANewSequenceWhereEach((monster) => monster.hasClaws);
     // #enddocregion code-like-prose-overdone
--- a/examples/misc/lib/effective_dart/design_bad.dart
+++ b/examples/misc/lib/effective_dart/design_bad.dart
@@ -1,6 +1,6 @@
 // ignore_for_file: type_annotate_public_apis, unused_element, unused_local_variable
 
-import 'dart:math' as math;
+import 'dart:math';
 import 'package:dartlang_examples_util/ellipsis.dart';
 import 'design_good.dart';
 
@@ -115,11 +115,8 @@ class Color {
 class Point {
   num x, y;
   Point(this.x, this.y);
-  static Point polar(num theta, num radius) {
-    return new Point(
-        radius * math.cos(theta), //!
-        radius * math.sin(theta));
-  }
+  static Point polar(num theta, num radius) =>
+      new Point(radius * cos(theta), radius * sin(theta));
 }
 // #enddocregion named-ctr
 
@@ -133,17 +130,15 @@ class C<Foo> {
 }
 
 //----------------------------------------------------------------------------
+// ignore_for_file: annotate_overrides
 
 // #docregion eq-dont-check-for-null
 class Person1 {
   final String name;
   // #enddocregion eq-dont-check-for-null
   Person1(this.name);
-
-  @override
   int get hashCode => ellipsis;
   // #docregion eq-dont-check-for-null
-  @override
-  operator ==(other) => other != null && other is Person && ellipsis;
+  operator ==(other) => other != null && ellipsis;
 }
 // #enddocregion eq-dont-check-for-null
--- a/examples/misc/lib/effective_dart/design_bad.dart
+++ b/examples/misc/lib/effective_dart/design_bad.dart
@@ -1,0 +1,149 @@
+// ignore_for_file: type_annotate_public_apis, unused_element, unused_local_variable
+
+import 'dart:math' as math;
+import 'package:dartlang_examples_util/ellipsis.dart';
+import 'design_good.dart';
+
+void miscDeclAnalyzedButNotTested() {
+  (errors, monsters, subscription) {
+    // #docregion code-like-prose
+    // Telling errors to empty itself, or asking if it is?
+    if (errors.empty) {/* ... */}
+
+    // Toggle what? To what?
+    subscription.toggle();
+
+    // Filter the monsters with claws *out* or include *only* those?
+    monsters.filter((monster) => monster.hasClaws);
+    // #enddocregion code-like-prose
+
+    Iterable theCollectionOfErrors;
+    // #docregion code-like-prose-overdone
+    if (theCollectionOfErrors.isEmpty) {/* ... */}
+
+    monsters.producesANewSequenceWhereEach((monster) => monster.hasClaws);
+    // #enddocregion code-like-prose-overdone
+  };
+
+  () {
+    // #docregion cascades
+    var buffer = new StringBuffer0() //!
+        .write('one')
+        .write('two')
+        .write('three');
+    // #enddocregion cascades
+  };
+
+  () {
+    // #docregion type_annotate_public_apis
+    install(id, destination) => ellipsis;
+    // #enddocregion type_annotate_public_apis
+  };
+
+  () {
+    // #docregion func-expr-no-param-type
+    var names = people.map((Person person) {
+      return person.name;
+    });
+    // #enddocregion func-expr-no-param-type
+  };
+
+  () {
+    // #docregion avoid-dynamic
+    dynamic lookUpOrDefault(String name, Map map, dynamic defaultValue) {
+      var value = map[name];
+      if (value != null) return value;
+      return defaultValue;
+    }
+    // #enddocregion avoid-dynamic
+  };
+
+  // #docregion avoid-Function
+  bool isValidString(String value, Function predicate) => ellipsis;
+  // #enddocregion avoid-Function
+
+  (int start) =>
+      // #docregion avoid-mandatory-param
+      string.substring(start, null);
+  // #enddocregion avoid-mandatory-param
+}
+
+//----------------------------------------------------------------------------
+
+class StringBuffer0 {
+  StringBuffer0 write(dynamic _) => null;
+}
+
+//----------------------------------------------------------------------------
+// ignore_for_file: one_member_abstracts
+
+// #docregion one-member-abstract-class
+abstract class Predicate<E> {
+  bool test(E element);
+}
+// #enddocregion one-member-abstract-class
+
+//----------------------------------------------------------------------------
+
+// #docregion class-only-static
+class DateUtils {
+  static DateTime mostRecent(List<DateTime> dates) {
+    return dates.reduce((a, b) => a.isAfter(b) ? a : b);
+  }
+}
+
+class _Favorites {
+  static const mammal = 'weasel';
+}
+// #enddocregion class-only-static
+
+//----------------------------------------------------------------------------
+
+// #docregion class-only-static-exception
+class Color {
+  static const red = '#f00';
+  static const green = '#0f0';
+  static const blue = '#00f';
+  static const black = '#000';
+  static const white = '#fff';
+}
+// #enddocregion class-only-static-exception
+
+//----------------------------------------------------------------------------
+
+// #docregion named-ctr
+class Point {
+  num x, y;
+  Point(this.x, this.y);
+  static Point polar(num theta, num radius) {
+    return new Point(
+        radius * math.cos(theta), //!
+        radius * math.sin(theta));
+  }
+}
+// #enddocregion named-ctr
+
+//----------------------------------------------------------------------------
+// ignore_for_file: avoid_return_types_on_setters
+
+class C<Foo> {
+  // #docregion avoid_return_types_on_setters
+  void set foo(Foo value) {/* ... */}
+  // #enddocregion avoid_return_types_on_setters
+}
+
+//----------------------------------------------------------------------------
+
+// #docregion eq-dont-check-for-null
+class Person1 {
+  final String name;
+  // #enddocregion eq-dont-check-for-null
+  Person1(this.name);
+
+  @override
+  int get hashCode => ellipsis;
+  // #docregion eq-dont-check-for-null
+  @override
+  operator ==(other) => other != null && other is Person && ellipsis;
+}
+// #enddocregion eq-dont-check-for-null
--- a/examples/misc/lib/effective_dart/design_good.dart
+++ b/examples/misc/lib/effective_dart/design_good.dart
@@ -21,7 +21,7 @@ void miscDeclAnalyzedButNotTested() {
   (Iterable errors, Iterable monsters) {
     // #docregion code-like-prose
     // "If errors is empty..."
-    if (errors.isEmpty) {/* ... */}
+    if (errors.isEmpty) {/*-...-*/}
 
     // "Hey, subscription, cancel!"
     subscription.cancel();
--- a/examples/misc/lib/effective_dart/design_good.dart
+++ b/examples/misc/lib/effective_dart/design_good.dart
@@ -39,7 +39,7 @@ void miscDeclAnalyzedButNotTested() {
     // #enddocregion omit-verb-for-bool-param
   };
 
-  (DoubleLinkedQueue queue, window, connection) {
+  (Queue queue, window, connection) {
     // #docregion verb-for-func-with-side-effect
     list.add(ellipsis);
     queue.removeFirst();
@@ -48,9 +48,10 @@ void miscDeclAnalyzedButNotTested() {
     // #enddocregion verb-for-func-with-side-effect
   };
 
-  () {
+  (test) {
     // #docregion noun-for-func-returning-value
     list.elementAt(3);
+    list.firstWhere(test);
     string.codeUnitAt(4);
     // #enddocregion noun-for-func-returning-value
   };
--- a/examples/misc/lib/effective_dart/design_good.dart
+++ b/examples/misc/lib/effective_dart/design_good.dart
@@ -1,0 +1,281 @@
+// ignore_for_file: type_annotate_public_apis, unused_element, unused_local_variable
+// #docplaster
+
+import 'dart:async';
+import 'dart:isolate';
+import 'dart:math' as math;
+import 'dart:typed_data';
+
+import 'package:dartlang_examples_util/ellipsis.dart';
+
+dynamic element, key, value;
+ByteBuffer bytes;
+DateTime dateTime;
+List list;
+Map map;
+String string;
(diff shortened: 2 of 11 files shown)


src/_guides/language/effective-dart/design.md
```diff
@@ -98,24 +99,26 @@ When in doubt about naming, write some code that uses your API, and try to read
...
+<?code-excerpt "misc/lib/effective_dart/design_good.dart (code-like-prose)"?>
 {% prettify dart %}
 // "If errors is empty..."
-if (errors.isEmpty) ...
+if (errors.isEmpty) { ... }
 
-// "Hey, _subscription, cancel!"
-_subscription.cancel();
+// "Hey, subscription, cancel!"
+subscription.cancel();
```
chalin: For consistency, I dropped the leading `_` because we use `subscription` elsewhere in the page.

src/_guides/language/effective-dart/design.md
```diff
@@ -230,9 +235,10 @@ Those kinds of members should be named using an imperative verb phrase that
 clarifies the work the member performs.
 
 {:.good-style}
+<?code-excerpt "misc/lib/effective_dart/design_good.dart (verb-for-func-with-side-effect)" replace="/;//g"?>
 {% prettify dart %}
-list.add()
-queue.removeFirst()
+list.add(...)
+queue.removeLast()
```
chalin: Using `removeLast()` because it is an actual `Iterable` method name (of course, that assumes that is a dequeue).

src/_guides/language/effective-dart/design.md
```diff
@@ -357,14 +368,16 @@ with a meaningless name like `call` or `invoke`, there is a good chance you
 just want a function.
 
 {:.good-style}
+<?code-excerpt "misc/lib/effective_dart/design_good.dart (one-member-abstract-class)"?>
 {% prettify dart %}
-typedef bool Predicate(item);
+typedef bool Predicate<E>(E element);
```
chalin: Note: added generics (to make the example more realistic/complete).

src/_guides/language/effective-dart/design.md
```diff
@@ -357,14 +368,16 @@ with a meaningless name like `call` or `invoke`, there is a good chance you
...
 {% endprettify %}
 
 {:.bad-style}
+<?code-excerpt "misc/lib/effective_dart/design_bad.dart (one-member-abstract-class)"?>
 {% prettify dart %}
-abstract class Predicate {
-  bool test(item);
+abstract class Predicate<E> {
+  bool test(E element);
```
chalin: Adding generics to match the use over `Predicate<E>`.

src/_guides/language/effective-dart/design.md
```diff
@@ -944,9 +975,10 @@ than the index of the last item.
 This is consistent with core libraries that do the same thing.
 
 {:.good-style}
+<?code-excerpt "misc/test/effective_dart_test.dart (param-range)" replace="/expect\(//g; /, \/\*\*\//  \/\//g; /\);//g"?>
 {% prettify dart %}
-[0, 1, 2, 3].sublist(1, 3) // [1, 2].
-'abcd'.substring(1, 3)     // "bc".
+[0, 1, 2, 3].sublist(1, 3)  // [1, 2]
+'abcd'.substring(1, 3)  // 'bc'
```
chalin: This is the only snippet that is actually tested. The nice thing is that each entire line (including the expression value in comments), is excerpt from the test.

src/_guides/language/effective-dart/design.md
```diff
@@ -1003,24 +1035,27 @@ The language specifies that this check is done automatically and your `==`
...
 class Person {
   final String name;
+  // ...
+  @override
+  [!operator ==!](other) => other is Person && name == other.name;
 
-  operator ==(other) =>
-      other is Person && name == other.name;
+  @override
+  int get [!hashCode!] => ...
```
chalin: In the guideline just above we suggest "When you define ==, you also have to define hashCode", so it seems best to illustrate that in the "good" example.

src/_guides/language/effective-dart/design.md
```diff
@@ -1003,24 +1035,27 @@ The language specifies that this check is done automatically and your `==`
...
 class Person {
   final String name;
-
-  operator ==(other) =>
-      other != null &&
-      other is Person &&
-      name == other.name;
+  // ···
+  @override
+  operator ==(other) => [!other != null!] && other is Person && ...
```
chalin: I'm unsure that you'll like the highlight here, but it puts emphasis on the bad part.

chalin:
@kwalrath - btw, some code blocks consist of a list of names, or extremely simple expressions, like this:

```
 {% prettify dart %}
 pageCount             // A count (of pages).
 ConversionSink        // A sink for doing conversions.
 ChunkedConversionSink // A ConversionSink that's chunked.
 CssFontFaceRule       // A rule for font faces in CSS.
 {% endprettify %}
```

For such fragments, that are extremely unlikely to ever become invalid, I've chosen to not extract them. As a consequence, they won't be analyzed, but I don't think that will be an issue.

kwalrath:
@lrhn, would you like to review these changes? (Either way is fine. @munificent won't be reviewing them.)

src/_guides/language/effective-dart/design.md
```diff
@@ -511,12 +528,14 @@ class Point {
...
 {:.bad-style}
+<?code-excerpt "misc/lib/effective_dart/design_bad.dart (named-ctr)"?>
 {% prettify dart %}
 class Point {
   num x, y;
   Point(this.x, this.y);
   static Point polar(num theta, num radius) {
-    return new Point(radius * math.cos(theta),
+    return new Point(
+        radius * math.cos(theta),
```
lrhn: We should do what the formatter does, and it does:
```dart
class Point {
  num x, y;
  Point(this.x, this.y);
  static Point polar(num theta, num radius) {
    return new Point(radius * math.cos(theta), radius * math.sin(theta));
  }
}
```
or, preferably, use a `=>` function:
```dart
class Point {
  num x, y;
  Point(this.x, this.y);
  static Point polar(num theta, num radius) =>
      new Point(radius * math.cos(theta), radius * math.sin(theta));
}
```
If the line still gets too long, maybe remove the `math.` prefixes, which aren't necessary.

src/_guides/language/effective-dart/design.md
```diff
@@ -716,13 +735,15 @@ With types, all of this is clarified.
 The type system infers `void` for all setters automatically.
```
lrhn: Is this text aimed at Dart 2 or Dart 1? 
(Because this line was incorrect for Dart 1, but the code still uses `new` in the `Point` example above, and many other places too, which it shouldn't for Dart 2).

src/_guides/language/effective-dart/design.md
```diff
@@ -944,9 +975,10 @@ than the index of the last item.
 This is consistent with core libraries that do the same thing.
 
 {:.good-style}
+<?code-excerpt "misc/test/effective_dart_test.dart (param-range)" replace="/expect\(//g; /, \/\*\*\//  \/\//g; /\);//g"?>
 {% prettify dart %}
-[0, 1, 2, 3].sublist(1, 3) // [1, 2].
-'abcd'.substring(1, 3)     // "bc".
+[0, 1, 2, 3].sublist(1, 3)  // [1, 2]
```
lrhn: The formatter leaves only one space between the end-of-line and the `//`,  so you shouldn't need to insert an extra space here.

src/_guides/language/effective-dart/design.md
```diff
@@ -1003,24 +1035,27 @@ The language specifies that this check is done automatically and your `==`
 method is called only if the right-hand side is not `null`.
 
 {:.good-style}
+<?code-excerpt "misc/lib/effective_dart/design_good.dart (eq-dont-check-for-null)" replace="/operator ==|hashCode/[!$&!]/g"?>
 {% prettify dart %}
 class Person {
   final String name;
+  // ...
+  @override
```
lrhn: Why add `@override`. The style guide does not recommend using `@override`. There is nothing "good" or "bad" about using it, so I don't think it should be added to a "good" section.

src/_guides/language/effective-dart/design.md
lrhn: I'd also include the actual code: `=> name.hashCode`.
That said, there is an `// ...` above, so we could just say that the `hashCode` declaration is implied.

src/_guides/language/effective-dart/design.md
lrhn: This is actually a somewhat bad example because `other != null && other is Person` is already redundant because `null is! Person`. It's bad code, independently of it being in a `==` declaration.

The bad thing here is to compare to `null` at all in `==` because we know the argument will never be `null`. If possible, we should focus on that.

I don't have a good example because I very rarely make a `==` method that doesn't start with an `is` check.

examples/misc/lib/effective_dart/design_bad.dart
```diff
@@ -0,0 +1,149 @@
+// ignore_for_file: type_annotate_public_apis, unused_element, unused_local_variable
+
+import 'dart:math' as math;
```
lrhn: Feel free to import dart:math without a prefix. I would :)

examples/misc/lib/effective_dart/design_bad.dart
```diff
@@ -0,0 +1,149 @@
...
+    // #docregion func-expr-no-param-type
+    var names = people.map((Person person) {
+      return person.name;
+    });
+    // #enddocregion func-expr-no-param-type
+  };
+
+  () {
+    // #docregion avoid-dynamic
+    dynamic lookUpOrDefault(String name, Map map, dynamic defaultValue) {
```
lrhn: The example seems silly now that we have generic methods, and with type inference, you sometimes *have* to write `dynamic` to get `dynamic`.

In this example, there is no advantage to writing `dynamic`, you could write `Object` and the function works just as well.

I'm not sure the "don't write dynamic" recommendation actually applies to Dart 2.

src/_guides/language/effective-dart/design.md
chalin: Right, I was looking at the `List` API. My objective was to stay within the scope of types offered by `dart:*` packages. I just noticed that [DoubleLinkedQueue](https://api.dartlang.org/stable/latest/dart-collection/DoubleLinkedQueue-class.html) is a part of `dart:collections`, and it has both remove methods as you mention. So I'll revert this change.

src/_guides/language/effective-dart/design.md
chalin: I prefer your suggestion to include `=> name.hashCode`.

src/_guides/language/effective-dart/design.md
chalin: > The bad thing here is to compare to `null` at all in `==` because we know the argument will never be `null`. If possible, we should focus on that.

To focus on that I suggest that we elide the rest of the expression, like this:

```dart
operator ==(other) => other != null && ...
```

examples/misc/lib/effective_dart/design_bad.dart
chalin: That is my preference as well. I kept the name because that is how the examples had been written. I'll drop the prefix.

chalin: I've dropped it here and for the other Effective Dart pages.

examples/misc/lib/effective_dart/design_bad.dart
chalin: FYI, these versions of the pages are for 1.24.

> I'm not sure the "don't write dynamic" recommendation actually applies to Dart 2.

I've added a note to that effect in the page. We'll be revisiting the pages when creating the Dart 2 version of the site.

src/_guides/language/effective-dart/design.md
lrhn: LGTM

src/_guides/language/effective-dart/design.md
```diff
@@ -98,24 +99,26 @@ When in doubt about naming, write some code that uses your API, and try to read
 it like a sentence.
 
 {:.good-style}
+<?code-excerpt "misc/lib/effective_dart/design_good.dart (code-like-prose)"?>
 {% prettify dart %}
 // "If errors is empty..."
-if (errors.isEmpty) ...
+if (errors.isEmpty) { ... }
```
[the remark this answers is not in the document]
chalin: Use of braces is only to make it consistent with the other Effective Dart pages (that have 28 instance of the use of `{ ... }`).

@kwalrath - what is your opinion about this? Should we replace `{ ... }` by `...` in all pages?

src/_guides/language/effective-dart/design.md
chalin: Done. Dropped `math.` and switched to using `=>`.

src/_guides/language/effective-dart/design.md
chalin: Dart 1

> Because this line was incorrect for Dart 1

Which line, this one? "The type system infers `void` for all setters automatically."?

src/_guides/language/effective-dart/design.md
chalin: Fixed.

src/_guides/language/effective-dart/design.md
chalin: I've been trying to avoid disabling any of the lint rules unless necessary.
Anyhow, I've removed `@override`s.

chalin:
Updates done. Staging server refreshed.

src/_guides/language/effective-dart/design.md
lrhn: Yes. Dart 1 had no inference. The actual return type would be `dynamic` if you didn't write anything.

chalin: Ah, of course, when interpreting "infers" for Dart 1, I took it non-technically in the sense that Dart 1 would assume the return type to be `void`, but it being `dynamic` makes more sense.

For Dart 1, how about if we just drop the prose? @kwalrath, any suggestions?

How are the semantics different in Dart 2?

chalin: I just checked and this guideline was added via #220 by @kwalrath and approved by @munificent. So this was meant to be for Dart 2 only?

chalin: Hmm, #220 was done in January of this year. Dart was still at 1.21 back then.

src/_guides/language/effective-dart/design.md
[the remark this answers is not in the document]
kwalrath: For the most part, yes. I certainly wouldn't want to add two unnecessary lines to every example that happens to use `if`, and unless we're talking about `if` syntax specifically, emphasizing the braces isn't necessary. If those cases where we do want to talk about the braces, I guess we should use the extra two lines.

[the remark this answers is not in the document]
chalin: Ok, I'll make the change.

src/_guides/language/effective-dart/design.md
kwalrath: It fixed bug #131, which was requested by a readability reviewer. The wording might not be correct, but it reflects Dart practices within Google.

kwalrath: I'm fine with dropping the explanatory prose, as long as we keep the guideline.

chalin: Dropping the prose but keeping the guideline.

kwalrath:
I tried to take a look at the refreshed page, but it only partially uploaded:
https://dartlang-org-staging-0.firebaseapp.com/guides/language/effective-dart/design

(It doesn't show up for people in SF & Mountain View; it does show up in Seattle & Montreal.)

This appears to be a CDN issue; let's NOT redeploy to this staging server until I give the all-clear.

The workaround is to add a query parameter (any query parameter) to the end:
https://dartlang-org-staging-0.firebaseapp.com/guides/language/effective-dart/design?0

chalin:
@lrhn - is this good to go?

examples/misc/test/library_tour/math_test.dart
```diff
@@ -8,47 +8,47 @@ void main() {
...
   });
 
   test('constants', () {
     expect(() {
       // #docregion constants
       // See the Math library for additional constants.
-      print(math.E); // 2.718281828459045
-      print(math.PI); // 3.141592653589793
-      print(math.SQRT2); // 1.4142135623730951
+      print(E); // 2.718281828459045
```
lrhn: Use lower-case constants here: e, pi, sqrt2.

examples/misc/lib/effective_dart/design_good.dart
```diff
@@ -0,0 +1,281 @@
...
+    queue.removeFirst();
+    window.refresh();
+    connection.downloadData();
+    // #enddocregion verb-for-func-with-side-effect
+  };
+
+  () {
+    // #docregion noun-for-func-returning-value
+    list.elementAt(3);
+    string.codeUnitAt(4);
```
lrhn: Consider adding `list.firstWhere(test)` as an example that doesn't end in `At`.

src/_guides/language/effective-dart/usage.md
```diff
@@ -331,8 +331,8 @@ class Circle {
 
   Circle(num radius)
       : radius = radius,
-        area = math.PI * radius * radius,
-        circumference = math.PI * 2.0 * radius;
+        area = PI * radius * radius,
```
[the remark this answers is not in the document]
chalin: I've replaced all dart:math E, PI, SQRT2 by e, pi, sqrt2.

src/_guides/language/language-tour.md
```diff
@@ -1406,10 +1406,10 @@ way:
 <?code-excerpt "misc/test/language_tour/operators_test.dart (precedence)"?>
 {% prettify dart %}
 // Parens improve readability.
```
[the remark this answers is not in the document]
chalin: Agreed, ... and fixed.

examples/misc/lib/effective_dart/design_good.dart
```diff
@@ -0,0 +1,281 @@
...
+
+  (entryPoint, message, elements, pattern) {
+    // #docregion omit-verb-for-bool-param
+    Isolate.spawn(entryPoint, message, paused: false);
+    new List.from(elements, growable: true);
+    new RegExp(pattern, caseSensitive: false);
+    // #enddocregion omit-verb-for-bool-param
+  };
+
+  (DoubleLinkedQueue queue, window, connection) {
```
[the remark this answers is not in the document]
chalin: Done.

examples/misc/lib/effective_dart/design_good.dart
chalin: Good suggestion. Done.

src/_guides/language/effective-dart/usage.md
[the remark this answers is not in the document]
chalin: Ah, ... but the lowercase names are for Dart 2 only, right (or has this change been introduced in 1.24.3)?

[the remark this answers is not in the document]
lrhn: Ack, no, that's not until 2.0.0-dev.8.0.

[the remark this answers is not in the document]
chalin: Ok, I've pulled out those changes (and saved them for the site's dev branch). I'll go ahead and merge the rest.

PR #486, opened 2017-12-18
chore: markdown and some CSS cleanup of Strong-Mode Dart pages

chalin:
This is preparation for continued work over #407.

@kwalrath - maybe we should rename the pages (and permalinks of) `sound-dart.md` and `sound-problems.md` to `strong-mode-dart.md` and `strong-mode-problems.md`?

In addition to fairly minor markdown cleanup, this PR has slight changes to the styles for the "passes/fails static analysis" classes. Notice the check mark and cross as the start of teach short message; also the "fails" case text is shorter.

screen shot 2017-12-18 at 08 49 23

### DIFF
--- a/src/_assets/stylesheets/main.scss
+++ b/src/_assets/stylesheets/main.scss
@@ -811,8 +811,8 @@ main .content {
 @include dart-style-for(good-style, $alert-success-bg, 'good');
 @include dart-style-for(bad-style, $alert-danger-bg, 'bad');
 
-@include dart-style-for(passes-sa, $alert-success-bg, '\2714  passes static analysis');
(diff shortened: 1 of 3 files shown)


kwalrath:
I'm inclined not to rename the pages because once Dart 2 is out, there will be no strong mode. (Everything will be sound/strong.) That's why we named the page that way originally.

As for the text changes, I'd like for the red text to be a little more clear and parallel to the "passes" text. (The new red text made it sound like the code itself was errors/warnings.) Maybe:

* static analysis: success
* static analysis: error/warning

chalin:
Good suggestion. Update done:
screen shot 2017-12-18 at 14 09 02

PR #491, opened 2017-12-19
chore: README and build script updates

chalin:


scripts/env-set.sh
```diff
@@ -2,10 +2,69 @@
 #
 # Until site-www has it's own setup, use the one from site-webdev
```
kwalrath: Delete this comment?

PR #493, opened 2017-12-19
chore: consolidate and document code-excerpt replace regexp conventions, and other misc updates

chalin:
1. Unify, consolidate and document code-excerpt replace regexp conventions
2. Misc markdown cleanup
3. Add new (2.0.0-dev.13.0) analyzer ignore flag `assignment_to_final_local`
4. Fix typo in script

This PR essentially has no user site-page-visible changes other than the code-excerpt changes done in step 1 above: e.g. `/* ... */` uniformly replaced by just `...`.

This PR requires the most recent version of https://github.com/chalin/code_excerpt_updater, so please re-activate it (by rerunning `./scripts/before-install.sh`, for example).

### DIFF
--- a/examples/util/lib/ellipsis.dart
+++ b/examples/util/lib/ellipsis.dart
@@ -1,2 +1,2 @@
 dynamic blockEllipsis; // Use as replace="/=. blockEllipsis;/{ ... }/g"
-dynamic ellipsis; // Use as replace="/ellipsis;/.../g"
+dynamic ellipsis; // Use as replace="/ellipsis;?/.../g"
--- a/src/_guides/language/effective-dart/design.md
+++ b/src/_guides/language/effective-dart/design.md
@@ -6,7 +6,7 @@ prevpage:
   url: /guides/language/effective-dart/usage
   title: Usage
 ---
-<?code-excerpt replace="/\/\/!//g;/ellipsis;?/.../g;/\/\*(\s*\.\.\.\s*)\*\//$1/g;/\{\/\*-(\s*\.\.\.\s*)-\*\/\}/$1/g;/([A-Z]\w*)\d\b/$1/g"?>
+<?code-excerpt replace="/([A-Z]\w*)\d\b/$1/g"?>
 
 Here are some guidelines for writing consistent, usable APIs for libraries.
 
--- a/src/_tutorials/language/streams.md
+++ b/src/_tutorials/language/streams.md
@@ -1,22 +1,19 @@
 ---
 title: "Asynchronous Programming: Streams"
-description: "Learn how to consume single-subscriber and broadcast streams."
-
+description: Learn how to consume single-subscriber and broadcast streams.
 prevpage:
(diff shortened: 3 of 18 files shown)


examples/util/lib/ellipsis.dart
```diff
@@ -1,2 +1,2 @@
 dynamic blockEllipsis; // Use as replace="/=. blockEllipsis;/{ ... }/g"
-dynamic ellipsis; // Use as replace="/ellipsis;/.../g"
+dynamic ellipsis; // Use as replace="/ellipsis;?/.../g"
```
kwalrath: Why does `ellipsis` have the `?` but `blockEllipsis` doesn't?

Do we use blockEllipsis anywhere?

src/_guides/language/effective-dart/design.md
```diff
@@ -6,7 +6,7 @@ prevpage:
   url: /guides/language/effective-dart/usage
   title: Usage
 ---
-<?code-excerpt replace="/\/\/!//g;/ellipsis;?/.../g;/\/\*(\s*\.\.\.\s*)\*\//$1/g;/\{\/\*-(\s*\.\.\.\s*)-\*\/\}/$1/g;/([A-Z]\w*)\d\b/$1/g"?>
+<?code-excerpt replace="/([A-Z]\w*)\d\b/$1/g"?>
```
kwalrath: When we have regexps, could you add a comment explaining them?

src/_tutorials/language/streams.md
```diff
@@ -1,22 +1,19 @@
...
-description: "Learn how to consume single-subscriber and broadcast streams."
-
+description: Learn how to consume single-subscriber and broadcast streams.
 prevpage:
   url: /tutorials/language/futures
   title: "Asynchronous Programming: Futures"
 ---
 
 <div class="panel" markdown="1">
+  <h4>What's the point?</h4>
```
kwalrath: Why `<h4>` instead of `####`?

examples/util/lib/ellipsis.dart
chalin: I've removed all current uses of `blockEllipsis`, but left it there in case we might need it later.

src/_tutorials/language/streams.md
chalin: So that we can have it all nicely indented under the `<div>`. I find that it makes the entire file easier to skim and read.

src/_guides/language/effective-dart/design.md
chalin: This one is used in a few files. I've explained it's purpose in the README.

chalin: To answer your question more directly :) ... yes, I will explain them, most certainly the global ones, unless their intent is obvious (like `/,//g` erases commas). Is that ok?

chalin:
I'll merge for now, but can add extra comments in subsequent passes.

PR #504, opened 2018-01-05
Fix bad formatting in homepage

kwalrath:
Some additional changes have cropped up in _main-example.html, for some
reason.

@chalin can you take a look at why those additional changes happened? Maybe I generated _main-example.html in the wrong way? I did it like this:

```
dart create_site_main_example.dart
``` 

Staged: https://kw-www-dartlang-1.firebaseapp.com/
Compare to: https://www.dartlang.org/


kwalrath:
It looks correct now.

PR #510, opened 2018-01-12
add more async/generator info to the language tour

### LINKED ISSUE #336: Add async*, sync*, yield* info to the language tutorial
The info in [Dart Language Asynchrony Support: Phase 2](https://www.dartlang.org/articles/language/beyond-async) was never added to the language tour.

cc: @kwalrath

kwalrath:
Fixes #336 

Staged at https://kw-www-dartlang-1.firebaseapp.com/guides/language/language-tour#asynchrony-support

### DIFF
--- a/examples/misc/test/language_tour/async_test.dart
+++ b/examples/misc/test/language_tour/async_test.dart
@@ -1,0 +1,54 @@
+// ignore_for_file: type_annotate_public_apis
+import 'dart:async';
+
+import 'package:test/test.dart';
+import 'package:dartlang_examples_util/print_matcher.dart' as m;
+
+void main() {
+  test('syncGenerator', () {
+    // #docregion sync-generator
+    Iterable naturalsTo(n) sync* {
+      int k = 0;
+      while (k < n) yield k++;
+    }
+    // #enddocregion sync-generator
+
+    main() {
+      print(naturalsTo(3));
+    }
+
+    expect(main, m.prints('(0, 1, 2)'));
+  });
+
+  test('asyncGenerator', () {
+    // #docregion async-generator
+    Stream asynchronousNaturalsTo(n) async* {
+      int k = 0;
+      while (k < n) yield k++;
+    }
+    // #enddocregion async-generator
+
+    main() async {
+      print(await asynchronousNaturalsTo(3).last);
+    }
+
+    expect(main, m.prints('2'));
+  });
+
+  test('recursiveGenerator', () {
+    // #docregion recursive-generator
+    Iterable naturalsDownFrom(n) sync* {
+      if (n > 0) {
+        yield n;
+        yield* naturalsDownFrom(n - 1);
+      }
+    }
+    // #enddocregion recursive-generator
+
+    main() {
+      print(naturalsDownFrom(3));
+    }
+
+    expect(main, m.prints('(3, 2, 1)'));
+  });
+}
--- a/examples/misc/test/language_tour/async_test.dart
+++ b/examples/misc/test/language_tour/async_test.dart
@@ -10,6 +10,7 @@ void main() {
       int k = 0;
       while (k < n) yield k++;
     }
+
     // #enddocregion sync-generator
     expect(naturalsTo(3), [0, 1, 2]);
   });
--- a/examples/misc/test/language_tour/async_test.dart
+++ b/examples/misc/test/language_tour/async_test.dart
@@ -2,7 +2,6 @@
 import 'dart:async';
 
 import 'package:test/test.dart';
-import 'package:dartlang_examples_util/print_matcher.dart' as m;
 
 void main() {
   test('syncGenerator', () {
@@ -12,15 +11,10 @@ void main() {
       while (k < n) yield k++;
     }
     // #enddocregion sync-generator
-
-    main() {
-      print(naturalsTo(3));
-    }
-
-    expect(main, m.prints('(0, 1, 2)'));
+    expect(naturalsTo(3), [0, 1, 2]);
   });
 
-  test('asyncGenerator', () {
+  test('asyncGenerator', () async {
     // #docregion async-generator
     Stream asynchronousNaturalsTo(n) async* {
       int k = 0;
@@ -28,11 +22,7 @@ void main() {
     }
     // #enddocregion async-generator
 
-    main() async {
-      print(await asynchronousNaturalsTo(3).last);
-    }
-
-    expect(main, m.prints('2'));
+    expect(await asynchronousNaturalsTo(3).toList(), [0, 1, 2]);
   });
 
   test('recursiveGenerator', () {
@@ -45,10 +35,6 @@ void main() {
     }
     // #enddocregion recursive-generator
 
-    main() {
-      print(naturalsDownFrom(3));
-    }
-
-    expect(main, m.prints('(3, 2, 1)'));
+    expect(naturalsDownFrom(3), [3, 2, 1]);
   });
 }
--- a/examples/misc/test/language_tour/async_test.dart
+++ b/examples/misc/test/language_tour/async_test.dart
@@ -10,8 +10,8 @@ void main() {
       int k = 0;
       while (k < n) yield k++;
     }
-
     // #enddocregion sync-generator
+
     expect(naturalsTo(3), [0, 1, 2]);
   });
 
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -3806,8 +3806,9 @@ because UI frameworks send endless streams of events.
 
 An asynchronous for loop has the following form:
 
+<?code-excerpt "misc/lib/language_tour/async.dart (await-for)"?>
 {% prettify dart %}
-await for (variableDeclaration in expression) {
+await for (varOrType identifier in expression) {
   // Executes each time the stream emits a value.
 }
 {% endprettify %}
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -3724,11 +3724,11 @@ make sure `await` is in an async function.**
 For example, to use `await` in your app's `main()` function,
 the body of `main()` must be marked as `async`:
 
-<?code-excerpt "misc/lib/language_tour/async.dart (main)"?>
+<?code-excerpt "misc/lib/language_tour/async.dart (main)" replace="/async|await/[!$&!]/g"?>
 {% prettify dart %}
-Future main() async {
+Future main() [!async!] {
   checkVersion();
-  print('In main: version is ${await lookUpVersion()}');
+  print('In main: version is ${[!await!] lookUpVersion()}');
 }
 {% endprettify %}
 
@@ -3830,11 +3830,11 @@ make sure the `await for` is in an async function.**
 For example, to use an asynchronous for loop in your app's `main()` function,
 the body of `main()` must be marked as `async`:
 
-<?code-excerpt "misc/lib/language_tour/async.dart (number_thinker)"?>
+<?code-excerpt "misc/lib/language_tour/async.dart (number_thinker)" replace="/async|await for/[!$&!]/g"?>
 {% prettify dart %}
-Future main() async {
+Future main() [!async!] {
   // ...
-  await for (var request in requestServer) {
+  [!await for!] (var request in requestServer) {
     handleRequest(request);
   }
   // ...
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -3649,23 +3649,9 @@ a possibly time-consuming operation
 (such as I/O),
 without waiting for that operation to complete.
 
-Several language features support asynchronous programming,
-as the following table shows.
-These features let you write asynchronous code that
+The `async` and `await` keywords support asynchronous programming,
+letting you write asynchronous code that
 looks similar to synchronous code.
-A related language feature lets you synchronously generate objects.
-
-|--------------+------------------------------------------------------------|
-|              |`async`|`await`|`await for`|`sync*`|`async*`|`yield`|`yield*`|
-|--------------|------------------------------------------------------------|
-| Handle values from **Futures**.|X|X| | | | | |
-| Handle values from **Streams**.|X| |X| | | | |
-| Implement a function that returns an **Iterable**<br>(a synchronous generator). | | | |X| |X| |
-| Implement a function that returns a **Stream**<br>(an asynchronous generator). | | | | |X|X| |
-| Implement a recursive generator.                               | | | | | | |X|
-{:.table .table-striped}
-
-The most commonly used asynchrony feature is handling Futures.
 
 
 <a id="await"></a>
@@ -3755,6 +3741,23 @@ Although an async function might perform time-consuming operations,
 it returns immediately—before
 any of its body executes.
 
+
+<aside class="alert alert-info" markdown="1">
+  **[Dart 2.0](/dart-2.0) difference**:
+    Async functions won't return immediately.
+    Instead, an async function will continue executing until
+    it either suspends or completes.
+    For more information, see the
+    [synchronous async start discussion](https://github.com/dart-lang/sdk/tree/master/docs/newsletter/20170915.md#synchronous-async-start)
+    in the
+    [Dart Language and Library Newsletter.](https://github.com/dart-lang/sdk/tree/master/docs/newsletter#dart-language-and-library-newsletters)
+{% comment %}
+More info:
+https://github.com/dart-lang/sdk/issues/29906
+https://github.com/dart-lang/sdk/issues/22909
+{% endcomment %}
+</aside>
+
 Adding the `async` keyword to a function makes it return a Future.
 For example, consider this synchronous function,
 which returns a String:
@@ -3778,7 +3781,7 @@ Dart creates the Future object if necessary.
 
 
 <a id="await-for"></a>
-### Handling values from Streams
+### Handling Streams
 
 When you need to get values from a **Stream**,
 you have two options:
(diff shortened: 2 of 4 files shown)


kwalrath:
@chalin, ptal

src/_guides/language/language-tour.md
```diff
@@ -3636,11 +3641,6 @@ for advice on how to implement a library package.
 <a id="asynchrony"></a>
 ## Asynchrony support
 
-Dart has several language features
-to support asynchronous programming.
-The most commonly used of these features are
-`async` functions and `await` expressions.
-
 Dart libraries are full of functions that
 return Future or Stream objects.
```
chalin: Add link to types on first occurrence in this section on asynchrony?

> return [Future][] or [Stream][] objects.

[Future]: tbc
[Stream]: tbc

src/_guides/language/language-tour.md
```diff
@@ -3649,17 +3649,20 @@ a possibly time-consuming operation
...
-When you need to use a value represented by a Future,
-you have two options:
+The `async` and `await` keywords support asynchronous programming,
+letting you write asynchronous code that
+looks similar to synchronous code.
+
 
-* Use `async` and `await`
-* Use the [Future API](/guides/libraries/library-tour#future)
+<a id="await"></a>
```
chalin: I'm not seeing any links to this anchor.

If you do want to keep this anchor, maybe move it to L3666 just above the paragraph that starts with the following?

> Code that uses `async` and `await` is asynchronous, ...

src/_guides/language/language-tour.md
```diff
@@ -3649,17 +3649,20 @@ a possibly time-consuming operation
...
+### Handling Futures
 
-Similarly, when you need to get values from a Stream,
+When you need to use a value represented by a **Future**,
```
chalin: How about:

> When you need the result of a completed Future, ...

src/_guides/language/language-tour.md
```diff
@@ -3649,17 +3649,20 @@ a possibly time-consuming operation
...
 you have two options:
 
-* Use `async` and an _asynchronous for loop_ (`await for`)
-* Use the [Stream API](/guides/libraries/library-tour#stream)
+* Use `async` and `await` to handle the most common case:
```
chalin: I would drop "most common case: waiting until the Future's value is available" because `async` and `await` are usable to help catch errors too. So maybe just:

* Use `async` and `await`

src/_guides/language/language-tour.md
```diff
@@ -3693,6 +3696,36 @@ try {
...
+<?code-excerpt "misc/lib/language_tour/async.dart (repeated-await)"?>
+{% prettify dart %}
+var entrypoint = await findEntrypoint();
+var exitCode = await runExecutable(entrypoint, args);
+await flushThenExit(exitCode);
+{% endprettify %}
+
+In <code>await <em>expression</em></code>,
+the value of <code><em>expression</em></code> is usually a Future;
+if it isn't, then the value is automatically wrapped in a Future.
```
chalin: L3711-3712: Is such detail really relevant in a Language Tour?

src/_guides/language/language-tour.md
```diff
@@ -3649,17 +3649,20 @@ a possibly time-consuming operation
...
 
-Similarly, when you need to get values from a Stream,
+When you need to use a value represented by a **Future**,
 you have two options:
 
-* Use `async` and an _asynchronous for loop_ (`await for`)
-* Use the [Stream API](/guides/libraries/library-tour#stream)
+* Use `async` and `await` to handle the most common case:
+  waiting until the Future's value is available.
+* Use the [Future API](/guides/libraries/library-tour#future).
```
chalin: I would have expected "Future API" to link to the Future API entry. I actually was a bit challenged to find a link to the Future API entry :).

How about making "Future API" link to the API entry and then mention that the reader can also consult the library tour section on Futures.

examples/misc/test/language_tour/async_test.dart
```diff
@@ -0,0 +1,54 @@
...
+      int k = 0;
+      while (k < n) yield k++;
+    }
+    // #enddocregion sync-generator
+
+    main() {
+      print(naturalsTo(3));
+    }
+
+    expect(main, m.prints('(0, 1, 2)'));
```
chalin: Good tests! Note that we don't need `main()` functions, so they can be dropped.
The generator functions can be tested directly, e.g.:

```dart
    expect(naturalsTo(3), [0, 1, 2]);
```

examples/misc/test/language_tour/async_test.dart
```diff
@@ -0,0 +1,54 @@
...
+  });
+
+  test('asyncGenerator', () {
```
chalin: Make the callback `async`:

> test('asyncGenerator', () async {

examples/misc/test/language_tour/async_test.dart
```diff
@@ -0,0 +1,54 @@
...
+      int k = 0;
+      while (k < n) yield k++;
+    }
+    // #enddocregion async-generator
+
+    main() async {
+      print(await asynchronousNaturalsTo(3).last);
+    }
+
+    expect(main, m.prints('2'));
```
chalin: Drop `main()` and then expect this:

```dart
    expect(await asynchronousNaturalsTo(3).toList(), [0, 1, 2]);
```

examples/misc/test/language_tour/async_test.dart
```diff
@@ -0,0 +1,54 @@
...
+        yield* naturalsDownFrom(n - 1);
+      }
+    }
+    // #enddocregion recursive-generator
+
+    main() {
+      print(naturalsDownFrom(3));
+    }
+
+    expect(main, m.prints('(3, 2, 1)'));
```
chalin: Drop `main()` and then expect this:

```dart
    expect(naturalsDownFrom(3), [3, 2, 1]);
```

chalin: After these changes are made, the print_matcher import can be dropped too.

chalin:
Later we should look into #512.

src/_guides/language/language-tour.md
kwalrath: It's the original ID for this section. I figured I'd leave it as-is, in case anyone had links to it.

src/_guides/language/language-tour.md
```diff
@@ -3730,56 +3780,29 @@ Note that the function's body doesn't need to use the Future API.
...
-the value of <code><em>expression</em></code> is usually a Future;
-if it isn't, then the value is automatically wrapped in a Future.
-This Future object indicates a promise to return an object.
-The value of <code>await <em>expression</em></code> is that returned object.
-The await expression makes execution pause until that object is available.
-
-**If `await` doesn't work, make sure it's in an async function.**
-For example, to use `await` in your app's `main()` function,
-the body of `main()` must be marked as `async`:
+<a id="await-for"></a>
```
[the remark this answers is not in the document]
kwalrath: As for "await", I was just keeping the original section ID.

src/_guides/language/language-tour.md
```diff
@@ -3693,6 +3696,36 @@ try {
...
+{% endprettify %}
+
+In <code>await <em>expression</em></code>,
+the value of <code><em>expression</em></code> is usually a Future;
+if it isn't, then the value is automatically wrapped in a Future.
+This Future object indicates a promise to return an object.
+The value of <code>await <em>expression</em></code> is that returned object.
+The await expression makes execution pause until that object is available.
+
+**If `await` doesn't work, make sure it's in an async function.**
```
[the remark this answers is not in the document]
kwalrath: We should be more precise... we meant that you'd get a compile-time error. (It's an easy mistake to make.)

src/_guides/language/language-tour.md
kwalrath: Where else would it go?

src/_guides/language/language-tour.md
```diff
@@ -3730,56 +3780,29 @@ Note that the function's body doesn't need to use the Future API.
...
+* Use the [Stream API](/guides/libraries/library-tour#stream).
 
-<a id="await-for"></a>
-### Using asynchronous for loops with Streams
+<aside class="alert alert-warning" markdown="1">
+**Note:**
+Asynchronous for loops aren't always appropriate.
+Use them only if it's OK for your program to do nothing
+while it waits for values from a single stream.
+</aside>
```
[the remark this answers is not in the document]
kwalrath: I think futures are easier to comprehend than streams. When we were converting all of our examples from the Future API to async-await, we had no issues. It was harder to convert the examples that used Streams. I remember wanting to warn people about the stream issues, but I left out the warning (wish I could remember why).

src/_guides/language/language-tour.md
kwalrath: I've modified the link text to clarify where it's going.

src/_guides/language/language-tour.md
```diff
@@ -3636,11 +3641,6 @@ for advice on how to implement a library package.
...
 ## Asynchrony support
 
-Dart has several language features
-to support asynchronous programming.
-The most commonly used of these features are
-`async` functions and `await` expressions.
-
 Dart libraries are full of functions that
 return Future or Stream objects.
 These functions are _asynchronous_:
```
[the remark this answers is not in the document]
kwalrath: I agree. Do you think we need to spell it out?

src/_guides/language/language-tour.md
```diff
@@ -3730,56 +3780,29 @@ Note that the function's body doesn't need to use the Future API.
...
-Future main() async {
-  checkVersion();
-  print('In main: version is ${await lookUpVersion()}');
-}
-{% endprettify %}
+When you need to get values from a **Stream**,
+you have two options:
 
+* Use `async` and an _asynchronous for loop_ (`await for`).
+* Use the [Stream API](/guides/libraries/library-tour#stream).
```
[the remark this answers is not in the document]
kwalrath: I modified the link text, as I did for Future.

src/_guides/language/language-tour.md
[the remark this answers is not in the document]
kwalrath: Actually, we did have the warning, but only in the library tour (https://www.dartlang.org/guides/libraries/library-tour#using-an-asynchronous-for-loop). It's arguably more appropriate here.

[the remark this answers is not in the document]
kwalrath: I started cutting down the Future and Stream sections of the library tour—there's a lot of duplication, as you noted—but decided to leave that for a future PR.

src/_guides/language/language-tour.md
```diff
@@ -3730,56 +3780,29 @@ Note that the function's body doesn't need to use the Future API.
...
+Asynchronous for loops aren't always appropriate.
+Use them only if it's OK for your program to do nothing
+while it waits for values from a single stream.
+</aside>
 
 An asynchronous for loop has the following form:
 
-<pre>
-<b>await for</b> (<em>variable declaration</em> in <em>expression</em>) {
+```dart
```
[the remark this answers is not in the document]
kwalrath: I started doing that, then realized I'd need a real stream, and that seemed like too much trouble for code that doesn't actually do anything.

kwalrath:
I've addressed most of your comments, although I suspect I'm missing something about generator terminology/conventions. Please take another look.

kwalrath:
Oops, I prematurely deleted some code from the library tour tests.

src/_guides/language/language-tour.md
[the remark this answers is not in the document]
chalin: I juggled with some optional prose and in the end felt that the current simple text is fine (for the Language Tour).

src/_guides/language/language-tour.md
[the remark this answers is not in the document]
chalin: Good change to clarify that we're referring to a compile-time error.

src/_guides/language/language-tour.md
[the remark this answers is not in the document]
chalin: The new prose for the note is much better. I how about a slight simplification:

> Before using `await for`, ensure that it makes the code clearer
> and that you really do want to wait for _all_ of the stream's results
> (since some streams, like UI event streams. are potentially endless).

src/_guides/language/language-tour.md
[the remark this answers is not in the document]
chalin: Since you won't be testing the code, but only subjecting to the analyzer, then you don't need to create stream; e.g. something like this could be added to the end of `examples/misc/lib/language_tour/async.dart`:

```dart
  (Stream expression) async {
    await for (var variableDeclaration in expression) {
      // Executes each time the stream emits a value.
    }
  };
```

**Note** the added `var` keyword.

src/_guides/language/language-tour.md
```diff
@@ -3824,6 +3850,58 @@ and
 [Dart Language Asynchrony Support: Phase 2](/articles/language/beyond-async),
 and the [Dart language specification](/guides/language/spec).
 
+
+<a id="generator"></a>
+## Generators
+
+When you need to lazily produce a sequence of values,
+consider using a _generator function_.
+Dart has built-in support for two kinds of generator functions:
```
chalin: Function should be singular in:
> for two kinds of generator function:

src/_guides/language/language-tour.md
```diff
@@ -3819,6 +3842,53 @@ and
...
+  int k = 0;
+  while (k < n) yield k++;
+}
+{% endprettify %}
+
+To implement an **asynchronous** generator, use `async*` and `yield`:
+
+<?code-excerpt "misc/test/language_tour/async_test.dart (async-generator)"?>
+{% prettify dart %}
+Stream asynchronousNaturalsTo(n) async* {
```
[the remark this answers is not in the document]
chalin: Could you add a inlined `{% comment %}` to this effect?

src/_guides/language/language-tour.md
```diff
@@ -3819,6 +3842,53 @@ and
...
+  int k = 0;
+  while (k < n) yield k++;
+}
+{% endprettify %}
+
+If your generator is recursive, improve performance using `yield*`:
```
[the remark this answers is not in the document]
chalin: I notice that you didn't go with the change I suggested, but there is a subtle difference between the current prose and what I suggested.

I _think_ that performance is improved even in the case that we have _mutually_ recursive generator functions invoking each other in a (tail-)recursive manner. Hence the use of the plural in my suggested rewording was purposeful :) I.e., "The performance of recursive generator functions".

src/_guides/language/language-tour.md
chalin: Given the proposed change in semantics of async functions (which won't block unless they need to), that statement (about non-Future expressions being wrapped) might not even be true anymore. I.e., it is an "implementation detail" of any given compiler or interpreter to wrap or not a non-Future expression.

I'd replace this entire paragraph by the simpler prose along the lines of:

> The expression <code>await <em>expression</em></code> makes execution pause, if necessary, until the value of _expression_ is available

src/_guides/language/language-tour.md
```diff
@@ -3698,6 +3701,37 @@ try {
...
+For example, to use `await` in your app's `main()` function,
+the body of `main()` must be marked as `async`:
+
+<?code-excerpt "misc/lib/language_tour/async.dart (main)"?>
+{% prettify dart %}
+Future main() async {
+  checkVersion();
+  print('In main: version is ${await lookUpVersion()}');
+}
+{% endprettify %}
```
chalin: At a quick glance it is hard to find the `await` inside the string interpolation expression. I suggest highlighting `async` and `await` in this excerpt (similar to what is done in previous excerpts).

src/_guides/language/language-tour.md
kwalrath: I think my wording is still correct. In your wording, I'm not sure about the "if necessary". I think (but should check) that await always pauses.

src/_guides/language/language-tour.md
kwalrath: Wow, this took me down a grammatical rathole! It turns out, opinions differ. The text and comments at [this grammar girl post](http://www.quickanddirtytips.com/education/grammar/sort-kind-or-type) imply that it should/can be plural (especially in US English). [This stack exchange topic](https://english.stackexchange.com/questions/29843/types-of-followed-by-singular-or-plural), on the other hand, is definitely on the "it should be singular" side. I'm going to leave it as is.

src/_guides/language/language-tour.md
kwalrath: I'm going to leave this as-is (since it's been in the published version for quite a while), but I'll file a bug to remind me to check.

chalin: Ok, let's drop the "if necessary". That leaves us with:

> The expression <code>await <em>expression</em></code> makes execution pause until the value of _expression_ is available

chalin: 👍 Will be tracked via #521

kwalrath:
OK, *now* I think we're ready to go. Staging server is updated.

examples/misc/test/language_tour/async_test.dart
```diff
@@ -0,0 +1,41 @@
...
+void main() {
+  test('syncGenerator', () {
+    // #docregion sync-generator
+    Iterable naturalsTo(n) sync* {
+      int k = 0;
+      while (k < n) yield k++;
+    }
+
+    // #enddocregion sync-generator
+    expect(naturalsTo(3), [0, 1, 2]);
```
chalin: You'll want to move the blank line that is at the end of the code excerpt and place it _after_ the `#enddocregion sync-generator` tag.

PR #511, opened 2018-01-13
Update language tour: Dart 2 changes to @proxy & noSuchMethod

### LINKED ISSUE #442: Update noSuchMethod explanation for Dart 2
from https://twitter.com/MatanLurey/status/933378071947702272

Sounds like `@proxy` will go away in Dart 2.

kwalrath:
Fixes #442

Also changed "Dart 2.0" to "Dart 2" everywhere in the language tour.

Staged at https://kw-staging-dartlang-2.firebaseapp.com/guides/language/language-tour#nosuchmethod

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -110,7 +110,7 @@ mind:
 -   In [strong mode](/guides/language/sound-dart), static and runtime
     checks ensure that your code is type safe, helping you catch bugs
     in development, rather than at runtime. Strong mode is optional in
-    Dart 1.x, but not optional in [Dart 2.0](/dart-2.0).
+    Dart 1.x, but not optional in [Dart 2](/dart-2.0).
 
 -   Dart parses all your code before running it. You can provide tips to
     Dart—for example, by using types or compile-time constants—to catch
@@ -2929,7 +2929,7 @@ update-for-dart-2
 {% endcomment %}
 
 <aside class="alert alert-info" markdown="1">
-  **[Dart 2.0](/dart-2.0) difference**:
+  **[Dart 2](/dart-2.0) difference**:
   Only abstract classes can have abstract methods.
   An error is reported otherwise.
 </aside>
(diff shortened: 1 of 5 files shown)


src/_guides/language/language-tour.md
```diff
@@ -3053,10 +3053,26 @@ class A {
...
 {% endprettify %}
 
+<aside class="alert alert-info" markdown="1">
+  **[Dart 2](/dart-2.0) difference**:
+  In Dart 2, you **can't invoke** an unimplemented method **unless**
+  the receiver is of type `dynamic` and
+  any of its **supertypes define that method** (even if abstractly).
+  For more information, see the
+  [informal feature specification.](https://github.com/dart-lang/sdk/blob/master/docs/language/informal/nosuchmethod-forwarding.md)
+</aside>
```
chalin: I don't think that this is accurate and/or necessary. Let's see what feedback we get from my comments in https://github.com/dart-lang/site-www/issues/442#issuecomment-357721649.

src/_guides/language/language-tour.md
```diff
@@ -110,7 +110,7 @@ mind:
 -   In [strong mode](/guides/language/sound-dart), static and runtime
     checks ensure that your code is type safe, helping you catch bugs
     in development, rather than at runtime. Strong mode is optional in
-    Dart 1.x, but not optional in [Dart 2.0](/dart-2.0).
+    Dart 1.x, but not optional in [Dart 2](/dart-2.0).
```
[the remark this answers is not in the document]
kwalrath: It looks like we'll be making that change ("Dart 2" instead of "Dart 2.0"), but I'm checking to make sure.

src/_guides/language/language-tour.md
kwalrath: How do you interpret the feedback you got? As far as I can tell, this note is still necessary (though it could be tweaked).

src/_guides/language/language-tour.md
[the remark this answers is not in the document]
kwalrath: And the word is in: it's "Dart 2". I'll make the edits.

src/_guides/language/language-tour.md
eernstg: We need an adjustment here: '(1) the receiver has static type `dynamic`; or (2) the dynamic type of the receiver has an implementation of `noSuchMethod` which is different from the one in class `Object`, and the static type of the receiver has that method (declared as abstract or with a body, both will suffice).'

I think we should mention the implementation of `noSuchMethod` here, because otherwise we'd expect the readers to discover on their own that the dynamic type of the receiver would have had a compile-time error _unless_ it has such a declaration of `noSuchMethod`, and I'm not sure they'll think of that automatically.

kwalrath:
@chalin ptal

src/_guides/language/language-tour.md
eernstg: Oops: If that method is declared with a body in one of the superclasses then we will _not_ "call an unimplemented method" after all, because it is implemented!

So we might want to say '(2) the dynamic type of the receiver has an implementation of `noSuchMethod` which is different from the one in class`Object`, the static type of the receiver has the method, and the receiver class does not have an implementation of the method.'

We may or may not wish to give an example (after all, this is just a short remark in the given context), but here's one: 

```dart
class A { foo() {}} 
class C implements A { noSuchMethod(i) => 42; }
main () => print(new C().foo() == 42); // 'true'
```

PR #516, opened 2018-01-19
Simplified front page

anders-sandholm:
Tim and I have been talking about ways that we could simplify the website and in particular the front page. Here's and initial attempt.

Would like feedback from Tim on the contents.

Would like feedback from website experts (looking at you Filip and Kathy) on how to fix alignment issues and not having enough room for the text that gets displayed when clicking underlined code. 

Bear with me if I am doing this the wrong way. Much if the mechanics of this is either new to me or rusty. There should be a staging server at https://as-dartlang-1.firebaseapp.com/


kwalrath:
I'm not terribly familiar with the CSS for this page. I can help you tweak it, but your guess is as good as mine.

chalin:
If you all agree on the content I can fix any CSS that needs fixing. Also, I have a pending PR to fix/improve the "tooltips" that are shown when we click underlined text (#6).

As for content, if you want to show only **three cards** on the main page, I'd recommend showing the following three (plus accompanying text); they match top three main categories in the sidenav:

- [Language](https://www.dartlang.org/guides/language)
- [Libraries](https://www.dartlang.org/guides/libraries)
- [Platforms](https://www.dartlang.org/guides/platforms) - accompanying prose would mention Web, Mobile and Server.

kwalrath:
@chalin if you could independently work on the CSS/implementation (while we finalize the content), that'd be great.

chalin:
FYI, #526 has the new tooltip work.

kwalrath:
I'm going to implement this in another PR.

PR #523, opened 2018-01-21
Spec updates for Dart 2

### LINKED ISSUE #514: Dart 2.0 -> Dart 2
Not a big deal, but it's a global change of terminology. We should only use "Dart 2" when we're differentiating from Dart 1.x; otherwise it's just plain Dart.

### LINKED ISSUE #522: Add more links to updates from spec page
We're making lots of changes to the language that aren't reflected in the 4th spec linked to by https://www.dartlang.org/guides/language/spec. We should point to those and perhaps to tracking bugs and the language newsletters, too.

Related: #516

/cc @anders-sandholm @timsneath

kwalrath:
Staged: https://kw-www-dartlang-1.firebaseapp.com/guides/language/spec
Compare to: https://www.dartlang.org/guides/language/spec

* Added a section to the spec page for Dart 2 updates.
* Removed an obsolete section about DEPs.
* Changed _Dart 2.0_ to _Dart 2_ everywhere in the site.

Fixes #514 
Fixes #522 

/cc @timsneath

### DIFF
--- a/src/_guides/language/spec.md
+++ b/src/_guides/language/spec.md
@@ -2,7 +2,6 @@
 layout: default
 title: "Dart Language Specification"
 description: "The Dart language specification and proposed changes."
-toc: false
 ---
 
 Download the _Dart Programming Language Specification_ from
@@ -16,6 +15,20 @@ For a gentler introduction to the Dart language, see the
 
 You can find previous editions of the specification at
 [Standard ECMA-408](http://www.ecma-international.org/publications/standards/Ecma-408-arch.htm).
+
+
+## Dart 2 changes
+
+Dart 2 is changing the Dart language in many ways,
+some of which are not backward-compatible.
+For information about what is changing and the implementation status,
+see the following resources:
+
+* [Dart 2 Updates](/dart-2.0)
+* [Dart Language and Library Newsletters](https://github.com/dart-lang/sdk/tree/master/docs/newsletter#dart-language-and-library-newsletters)
+* [Language specification source file (TeX format)](https://github.com/dart-lang/sdk/blob/master/docs/language/dartLangSpec.tex)
+* [Informal specifications](https://github.com/dart-lang/sdk/tree/master/docs/language/informal)
+
 
 ## Changes in the 4<sup>th</sup> edition
 
@@ -58,11 +71,3 @@ the following new language features:
   Implemented in 1.6. For details, see the language tour:
   [Lazily loading a library](/guides/language/language-tour#deferred-loading).
 {% endcomment %}
-
-## Proposed enhancements to the Dart language
-
-Interested in learning about new features that may be added to Dart?
-
-You can track all Dart Enhancement Proposals (DEPs) in the
-[dart_enhancement_proposals](https://github.com/dart-lang/dart_enhancement_proposals)
-repo on GitHub.
--- a/src/_guides/language/spec.md
+++ b/src/_guides/language/spec.md
@@ -24,7 +24,7 @@ some of which are not backward-compatible.
 For information about what is changing and the implementation status,
 see the following resources:
 
-* [Dart 2 Updates](/dart-2.0)
+* [Dart 2 Updates](/dart-2)
 * [Dart Language and Library Newsletters](https://github.com/dart-lang/sdk/tree/master/docs/newsletter#dart-language-and-library-newsletters)
 * [Language specification source file (TeX format)](https://github.com/dart-lang/sdk/blob/master/docs/language/dartLangSpec.tex)
 * [Informal specifications](https://github.com/dart-lang/sdk/tree/master/docs/language/informal)
--- a/src/dart-2.0.md
+++ b/src/dart-2.0.md
@@ -1,17 +1,17 @@
 ---
 layout: default
 permalink: /dart-2.0
-title: Dart 2.0 Updates
-description: How Dart 2.0 is different from Dart 1.x, and how you can prepare.
+title: Dart 2 Updates
+description: How Dart 2 is different from Dart 1.x, and how you can prepare.
 ---
 
-This page has information about changes that are coming in Dart 2.0,
+This page has information about changes that are coming in Dart 2,
 and how you can migrate your code from Dart 1.x.
 
 <aside class="alert alert-warning" markdown="1">
 **Don't rely on 2.0 until it's stable.**
-Dart 2.0 pre-releases will have a series of incompatible changes.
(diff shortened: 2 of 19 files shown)


src/_guides/language/spec.md
```diff
@@ -17,6 +16,20 @@ For a gentler introduction to the Dart language, see the
...
+## Dart 2 changes
+
+Dart 2 is changing the Dart language in many ways,
+some of which are not backward-compatible.
+For information about what is changing and the implementation status,
+see the following resources:
+
+* [Dart 2 Updates](/dart-2.0)
+* [Dart Language and Library Newsletters](https://github.com/dart-lang/sdk/tree/master/docs/newsletter#dart-language-and-library-newsletters)
+* [Language specification source file (TeX format)](https://github.com/dart-lang/sdk/blob/master/docs/language/dartLangSpec.tex)
```
chalin: Maybe put the TeX format last :). Few folks will want to read TeX sources.

Btw, it wouldn't be too difficult to setup a site with a built PDF version of the spec.

src/dart-2.0.md
```diff
@@ -1,17 +1,17 @@
...
+title: Dart 2 Updates
+description: How Dart 2 is different from Dart 1.x, and how you can prepare.
 ---
 
-This page has information about changes that are coming in Dart 2.0,
+This page has information about changes that are coming in Dart 2,
 and how you can migrate your code from Dart 1.x.
 
 <aside class="alert alert-warning" markdown="1">
 **Don't rely on 2.0 until it's stable.**
```
chalin: Possibly change this to:

> **Don't rely on Dart 2 until it's stable.**

chalin:
Might you consider renaming `/dart-2.0` to `/dart-2` and adding a redirecting rule?

chalin:
If you do decide to rename the file, then you can drop the special setting of the headers for `/dart-2.0` in `firebase.json`.

src/_guides/language/spec.md
kwalrath: I actually put the TeX third to bury it. The last item in a list is easier to see, in my experience.

Could you open an (sdk?) issue for publishing a PDF version of the current externally?

src/dart-2.0.md
kwalrath: Good catch

kwalrath: Good catch

kwalrath: Good catch

src/_guides/language/spec.md
chalin: Here is the issue: https://github.com/dart-lang/sdk/issues/31957

chalin: If you agree with what is proposed https://github.com/dart-lang/sdk/issues/31957#issuecomment-359309323, then this link could be used for the most recent PDF of the Language Specification: https://github.com/chalin/dart-spec-and-grammar/blob/master/doc/dartLangSpec.pdf

kwalrath:
I'm going to do the rename in a separate PR.

src/dart-2.md
```diff
@@ -1,17 +1,17 @@
...
 <aside class="alert alert-warning" markdown="1">
-**Don't rely on 2.0 until it's stable.**
-Dart 2.0 pre-releases will have a series of incompatible changes.
-Although we appreciate your willingness to prepare for and test Dart 2.0,
+**Don't rely on Dart 2 until it's stable.**
+Dart 2 pre-releases will have a series of incompatible changes.
+Although we appreciate your willingness to prepare for and test Dart 2,
 please don't bet your livelihood on a pre-release.
 For information on potential changes, see
 [Dart Language & Library Newsletters.](https://github.com/dart-lang/sdk/blob/master/docs/newsletter/README.md#dart-language-and-library-newsletters)
```
chalin: Rather than link only to the news letters, this should probably now link back to /guides/language/spec#dart-2-changes.

chalin:
Consider:

> the following is valid code in both Dart 1.x and Dart 2:

Now that we made the switch to using the term "Dart 2", it would seem more consistent to refer to "Dart 1" as well (rather than 1.x).

src/_guides/language/spec.md
chalin: @kwalrath - if you prefer I could publish the Language Spec PDF on dartlang.org, and update it periodically. We could then change the link to something like:

* [Language specification](#link-to-PDF-on-dartlang) ([TeX source](https://github.com/dart-lang/sdk/blob/master/docs/language/dartLangSpec.tex))

chalin:
Also please see https://github.com/dart-lang/site-www/pull/523#discussion_r162839369 and  https://github.com/dart-lang/site-www/pull/523#discussion_r162959530 -- I'm adding a top-level comment with links here because it is easy to miss feedback on threads marked as "outdated".

src/_guides/language/spec.md
kwalrath: Let me ask the people who care about where we should publish the PDF.

kwalrath:
Re: terminology
For now, at least, we're going to continue with "Dart 1.x" and "Dart 2".

src/_guides/language/spec.md
anders-sandholm: If do that we should clearly mark it as "draft" as it is WIP. 
Would like the "informal" readme file to give a good overview of current semantics (i.e., what can be found as an informal spec and what has already been added to the draft 5th version of the spec - and hence can be found in the PDF if we publish it.

anders-sandholm:
LGTM

src/_guides/language/spec.md
chalin: > If we do that, we should clearly mark it as "draft" as it is WIP.

The first page of the PDF has "5th edition draft" right under the title. Is that enough? Otherwise I can probably add "DRAFT" on each page if you prefer.

> Would like the "informal" readme file to give a good overview of current semantics

To be clear, which informal readme are you referring to?

chalin: Also note that https://www.dartlang.org/dart-2 has a summary of the main Dart 2 updates. Might a link to that page be sufficient?

anders-sandholm: Adding a draft watermark or something would be ideal.

I was referring to https://github.com/dart-lang/sdk/blob/master/docs/language/informal/README.md

Ideally, that file would have an updated table of the status of each file/part the directory to give an easy overview.

chalin: > Adding a draft watermark or something would be ideal.

Done: https://github.com/chalin/dart-spec-and-grammar/blob/master/doc/dartLangSpec.pdf

PR #525, opened 2018-01-21
fix existing page popovers in site pages

### LINKED ISSUE #524: Popovers issues: hidden pop-up, and clicking scrolls to top
Popovers are currently used in these pages:

- https://www.dartlang.org/tutorials/dart-vm/cmdline
- https://www.dartlang.org/tutorials/libraries/shared-pkgs

There are two issues with the current implementation.

1. Clicking on a popover scrolls the page to the top. That shouldn't happen.
2. Popovers are easy to miss on narrow displays because the popup is shown to the right of the triggering text and it can be hidden unless the reader knows to scroll left.

As an example of (2), here is a popover for the dart:async import:
> screen shot 2018-01-21 at 15 21 44

But it can't be seen unless you scroll left:
> screen shot 2018-01-21 at 15 22 01

Fixing this is a precursor to fixing #6.

chalin:
Fixes #524, and behaves nicely on mobile.

Replaced all `<a href="#" class="dart-popover" data-toggle="popover" ...>...</a>` elements by appropriate `<span>` elements.

Popovers show when hovering over the trigger text, or clicking (which is necessary on mobile).

Clicking no longer moves the page, and popups show more nicely on the top of the trigger text; e.g.:

screen shot 2018-01-21 at 15 40 12

This PR has no content change other than required edits to obsolete `pubspec.yaml` code excerpts found in the shared-pkgs page.

### DIFF
--- a/src/_assets/javascripts/main.js
+++ b/src/_assets/javascripts/main.js
@@ -109,7 +109,12 @@ $(document).on('ready', function(){
 
 
   // Popovers
-  $('[data-toggle="popover"], .dart-popover').popover();
+  $('[data-toggle="popover"]').popover({
+    container: 'body',
+    html: true,
+    placement: 'top',
+    trigger: 'focus hover',
+  });
 
   // open - close mobile navigation
   $('#menu-toggle').on('click', function(e) {
--- a/src/_assets/stylesheets/_overwrites.scss
+++ b/src/_assets/stylesheets/_overwrites.scss
@@ -382,3 +382,9 @@ body.homepage .get-started-section {
 pre code {
   white-space: pre;
 }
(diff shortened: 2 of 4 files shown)


chalin:
Staged at

- https://dartlang-org-staging-0.firebaseapp.com/tutorials/dart-vm/cmdline
- https://dartlang-org-staging-0.firebaseapp.com/tutorials/libraries/shared-pkgs

kwalrath:
The behavior is much better on mobile! I'll take a look at the behavior on desktop later, but I trust that you've gotten it right.

kwalrath:
LGTM on desktop, too.

PR #526, opened 2018-01-22
Improved tooltips for frontpage code excerpt and prose 

chalin:
- A preliminary step to #6, which adds support for HTML in the tooltip text.
- Continuation of #525.
- Staged at https://dartlang-org-staging-0.firebaseapp.com
- This eliminates the sometimes awkwardly placed front page "footnote" area that was used to display tooltips.
   screen shot 2018-01-22 at 09 19 24
   The main problem with this area is that it was invisible on most mobile devices until the user scrolled.

---

This is what the new tooltip/popovers look like:

screen shot 2018-01-22 at 09 13 54

screen shot 2018-01-22 at 09 13 44

screen shot 2018-01-22 at 09 12 51


chalin:
Dotted underlines are back. Staging server updated.

PR #533, opened 2018-01-26
Simplify the homepage

kwalrath:
Builds on work done in #516.

Staged at: https://kw-www-dartlang-1.firebaseapp.com

### DIFF
--- a/src/_assets/stylesheets/main.scss
+++ b/src/_assets/stylesheets/main.scss
@@ -758,6 +758,7 @@ main .content {
   text-align: center;
   h1, h2, h3, h4, h5, h6 {
     margin-top: 0;
+    margin-bottom: 1em;
   }
 }
 
--- a/src/_assets/stylesheets/main.scss
+++ b/src/_assets/stylesheets/main.scss
@@ -39,7 +39,7 @@ h1, h2, h3, h4, h5, h6 {
 h1, .h1 { font-size: $font-size-h1; margin-top: 0; margin-bottom: 0.67em;}
 h2, .h2 { font-size: $font-size-h2;}
 h3, .h3 { font-size: $font-size-h3;}
-h4, .h4 { font-size: $font-size-h4; margin-bottom: 0; }
+h4, .h4 { font-size: $font-size-h4;}
 h5, .h5 { font-size: $font-size-h5;}
 h6, .h6 { font-size: $font-size-h6;}
 
@@ -758,7 +758,6 @@ main .content {
   text-align: center;
   h1, h2, h3, h4, h5, h6 {
     margin-top: 0;
-    margin-bottom: 1em;
   }
 }
 
--- a/src/index.html
+++ b/src/index.html
@@ -35,9 +35,11 @@ description: Dart is an application programming language that’s easy to learn,
                 <li> Delightful, flexible <span data-toggle="popover" data-content="These range from <a href='/tools'>general-purpose tools</a> to framework-specific ones like the <a href='https://flutter.io/getting-started'><code>flutter</code> tool</a>.">tooling</span>
                 </li>
               </ul>
-              <p><a href="/guides/get-started" class="frontpage-slogan__get-started btn btn-default btn-lg">Get Started</a> </p>
+              <p><a href="/guides/get-started" class="frontpage-slogan__get-started btn btn-default btn-lg">Get started</a> </p>
             </h3>
-            <p> <em>Click <span data-toggle="popover" data-content="Try clicking dotted underlines in the text above and in the code sample.">underlined text</span> to learn more.</em> </p>
+            <p> <em>Click the
+              <span data-toggle="popover" data-content="In the text above and in the code sample, click dotted underlines to get more details.">underlined text</span>
+            to learn more.</em> </p>
         </div>
       </div>
     </div>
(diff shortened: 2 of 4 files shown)


src/index.html
```diff
@@ -26,54 +26,42 @@
...
+
+              <ul>
+                <li> A <span data-toggle="popover" data-content="Dart started out being optimized for web apps, and is evolving to provide great support for mobile apps. Dart also runs on the command line and server-side.">client-optimized</span> language </li>
+                <li> Rich, powerful <span data-toggle="popover" data-content="For mobile apps, we recommend <a href='https://flutter.io'>Flutter</a>. For desktop apps, <a href='https://webdev.dartlang.org/angular'>AngularDart</a>.">frameworks</span> </li>
+                <li> Delightful, flexible <span data-toggle="popover" data-content="These range from <a href='/tools'>general-purpose tools</a> to framework-specific ones like the <a href='https://flutter.io/getting-started'><code>flutter</code> tool</a>.">tooling</span>
+                </li>
+              </ul>
+              <p><a href="/guides/get-started" class="frontpage-slogan__get-started btn btn-default btn-lg">Get Started</a> </p>
             </h3>
-            <p><a href="/guides/get-started" class="frontpage-slogan__get-started btn btn-default btn-lg">Get Started</a> <a href="/install" class="btn btn-lg">Install Dart</a></p>
```
chalin: I liked having the Install Dart button on the main page (and there is enough view real estate to host it).

src/_assets/stylesheets/main.scss
```diff
@@ -758,6 +758,7 @@ main .content {
   text-align: center;
   h1, h2, h3, h4, h5, h6 {
     margin-top: 0;
+    margin-bottom: 1em;
```
chalin: I wouldn't recommend making this style change for all heading levels inside a `card`-styled element. This impacts pages other than the front page (e.g., Languages overview).

The source of the problem is that the h4 element is assigned a margin-bottom of 0 globally (that shouldn't have been done).

Anyhow, see below: I suggest using `<h3>` elements instead (because they look better, and that will allow us to fix the mis-styling of `<h4>` elements later).

src/index.html
```diff
@@ -26,54 +26,42 @@
...
-            <a href="https://pub.dartlang.org/" class="chip no-automatic-external">
-              <h3>Pub</h3>
-              <p>Find packages.</p>
-            </a>
-            <a href="{{site.webdev}}" class="chip no-automatic-external">
-              <h3>Dart webdev</h3>
-              <p>Build browser apps.</p>
+          <div class="card-grid">
+            <a href="/guides/language/language-tour" class="card no-automatic-external">
+              <h4>Language tour</h4>
```
chalin: These look better as `<h3>` elements (which is the heading level used, e.g., in the Languages Overview page) -- and that is aside from the `<h4>` margin-bottom problem mentioned above.

src/index.html
kwalrath: The problem with having the install button there is that you don't want to install Dart that way if you use Flutter.

src/_assets/stylesheets/main.scss
kwalrath: I wanted the font size of the card headings to match the font size of the Get Started button. I'll try fixing the `<h4>` styling, since that's bothered me for a while.

src/index.html
chalin: Ok, makes sense.

src/_assets/stylesheets/main.scss
chalin: > I wanted the font size of the card headings to match the font size of the Get Started 

I just checked and it is the same size when the heading is `<h3>`.

kwalrath: I see the h4 ("Library tour" in the picture below) as 21px and the h3 ("Dart packages" below) as 24px. The button is actually 20px, but I couldn't see the difference.

image

kwalrath:
ptal (and see https://github.com/dart-lang/site-www/pull/533#discussion_r164221073)

kwalrath:
I'm going to wait for approval from @anders-sandholm or @timsneath before merging.

src/index.html
anders-sandholm: Agree with @kwalrath on this one. Would be nice if there was a single fat "install" button that made sense, though.

src/index.html
```diff
@@ -26,54 +26,42 @@
...
+              <p>A walk through Dart's major language features.</p>
             </a>
-            <a href="https://flutter.io/" class="chip no-automatic-external">
-              <h3>Flutter</h3>
-              <p>Build mobile apps.</p>
+            <a href="/guides/libraries/library-tour" class="card no-automatic-external">
```
anders-sandholm: I'd prefer having the "Platform" (https://www.dartlang.org/guides/platforms) as one of the three boxes instead of the library tour. Ideally as the last one after Pub

src/index.html
```diff
@@ -26,54 +26,42 @@
...
+              <h4>Library tour</h4>
+              <p>An example-driven introduction to Dart's core libraries.</p>
             </a>
-            <a href="/dart-vm" class="chip no-automatic-external">
-              <h3>Dart VM</h3>
-              <p>Build scripts, servers.</p>
+            <a href="https://pub.dartlang.org/" class="card no-automatic-external">
+              <h4>Dart packages</h4>
+              <p>An ecosystem of contributed libraries and tools.</p>
```
anders-sandholm: Might consider using (some of) the text from the Pub front-page "Find and use packages to build Flutter and web apps with Dart." instead.

src/index.html
```diff
@@ -26,54 +26,42 @@
...
+                <li> Rich, powerful <span data-toggle="popover" data-content="For mobile apps, we recommend <a href='https://flutter.io'>Flutter</a>. For desktop apps, <a href='https://webdev.dartlang.org/angular'>AngularDart</a>.">frameworks</span> </li>
+                <li> Delightful, flexible <span data-toggle="popover" data-content="These range from <a href='/tools'>general-purpose tools</a> to framework-specific ones like the <a href='https://flutter.io/getting-started'><code>flutter</code> tool</a>.">tooling</span>
+                </li>
+              </ul>
+              <p><a href="/guides/get-started" class="frontpage-slogan__get-started btn btn-default btn-lg">Get Started</a> </p>
             </h3>
-            <p><a href="/guides/get-started" class="frontpage-slogan__get-started btn btn-default btn-lg">Get Started</a> <a href="/install" class="btn btn-lg">Install Dart</a></p>
+            <p> <em>Click <span data-toggle="popover" data-content="Try clicking dotted underlines in the text above and in the code sample.">underlined text</span> to learn more.</em> </p>
```
anders-sandholm: I almost feel like we should get rid of this line now with the new popup.

kwalrath: When the feature was on mouse-over, I think we could've deleted it, but now that you have to click, I think we need a hint. (I suppose we could test this supposition, somehow.)

src/index.html
kwalrath: I'm OK with Platform as one of the 3, but I'd like the page to be more attractive. I'll file a bug for that.

src/index.html
kwalrath: That didn't turn out to fit well here, once I added the Platforms card.

kwalrath:
@anders-sandholm or @timsneath ptal.

anders-sandholm:
Awesome. LGTM. 
Thanks a lot @kwalrath for driving these changes.
FWIW, I still think we could do fine w/o the "instruction" sentence telling people to click the underlined text and it would make the page "cleaner"/simpler.

kwalrath:
Thanks, Anders! I'll file an issue to remind us to take another look at the instruction sentence...

PR #540, opened 2018-01-31
Update DartPad page

kwalrath:
* Update the screenshot.
* Add instructions for using Format button.
* Improve flow & navigation, now that we show 3rd-level heads in the TOC (e.g. added "Library support" header).
* Clarify library support.
* Add site URL to top of page.
* Add a `site.pub` macro (without adjusting other pages' references to pub.dartlang.org... I'll file a bug about that).
* Tweak the `site.dart_api` macro, which is the only one to have a `/` at the end. I searched for uses and didn't find any that depended on that.

Staged at: https://kw-staging-dartlang-2.firebaseapp.com/tools/dartpad


_config.yml
```diff
@@ -13,8 +13,9 @@ dartlang: ""
 webdev: "https://webdev.dartlang.org"
 dart_vm:  "/dart-vm"
 flutter:  "https://flutter.io"
-dart_api: "https://api.dartlang.org/"
+dart_api: "https://api.dartlang.org"
```
chalin: Good catch.

src/tools/dartpad.md
```diff
@@ -4,21 +4,29 @@ description: The tool that lets you interactively play with Dart in a browser.
...
+To get familiar with DartPad, try these steps:
+
+* [Run some samples.](#step-1-open-and-run)
+* [Create a simple command-line app.](#step-2-server)
+* [Create a simple web app.](#step-3-web)
+
+
+### Open DartPad, and run some samples {#step-1-open-and-run}
 
 <ol markdown="1">
```
chalin: Eventually, these should be converted to markdown lists (something I started doing in the ACX Lottery tutorial and elsewhere).

PR #541, opened 2018-01-31
Update csharp-samples.xml

Vtek:
- Updated csharp samples. Replace String.Format method with C# 6 string interpolation.


Vtek:
I signed it!

kwalrath:
Thanks very much for the update, @Vtek!

PR #543, opened 2018-01-31
Frontpage popovers: use auto placement mode

### LINKED ISSUE #534: Popovers are hidden at the lower end of the code area
When clicking popovers at the lower end of the code screen, all or some of it gets hidden (see image below). 

I guess one solution could be to place the popover in a way that it is visible, if at all possible. E.g., when a user clicks at the lower end of the screen, the popover appears above the clicked text as opposed to below.

image

chalin:
Fixes #534. I tried this "auto" mode before, but it wasn't working. I think that it might have been because the view hadn't yet correctly been set. Anyhow, "auto top" as a placement option is working well now.

### DIFF
--- a/src/_assets/javascripts/main.js
+++ b/src/_assets/javascripts/main.js
@@ -98,12 +98,7 @@ $(document).on('ready', function(){
     popovers.popover({
       container: viewport === 'body' ? 'body' : undefined,
       html: true,
-      placement:
-        // When in a subview, open popovers towards the bottom so that they
-        // can be viewed by scrolling the subview. Scrolling to view the
-        // popover is possible when the popover opens below the trigger element,
-        // not when it appears above.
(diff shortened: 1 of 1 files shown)


kwalrath:
Why is there a lighter background on the link targets? (I think you staged both this PR and #544 here, so I don't know which one caused the change.) It looks a little odd, especially in cases where you have two links that overlap vertically, like "Iterable" and "final":

image

Also, could you pick up my latest changes to this page?

kwalrath:
PS: The behavior seems to work great. (@anders-sandholm, agree?) It's only the styling I quibble with.

kwalrath:
Accepting this change, since it shouldn't affect the styles... #544 is probably the culprit.

chalin:
Yes, #544 introduces the style change.

PR #544, opened 2018-01-31
fix dropped popover clicks

### LINKED ISSUE #535: Popover clicks sometimes don't work
When clicking an underlined item, say "A", a popover appears.
When you click "B", the popover for A gets closed and the one for B appears... most of the time. 

When clicking around in the code it was easy for me to reproduce a case where when clicking B only the popover for A would disappear and no new popover for B would appear. In particular if you then keep clicking a new area, none of them would show a popup...

chalin:
Builds on #543.
Fixes #535.


chalin:
Concerning the style change mentioned in https://github.com/dart-lang/site-www/pull/543#issuecomment-362090064. The lighter background is the default style for links inside code elements. I had to change the popover base element to be links to get rid of the lost clicks. I thought that the lighter background made for a stronger visual cue that there is something of interest in the code.

I don't have a strong opinion either way. Let me know what you prefer. If it is the old style, then I'll add a CSS override.

PR #554, opened 2018-02-02
Add logos to Platform Overview page

### LINKED ISSUE #551: Add logos to Platform Overview page
I'm not a designer... but I'd love to see large Flutter, AngularDart, and Dart logos in each of the three boxes to lighten up the page (https://www.dartlang.org/guides/platforms)

chalin:
Fixes #551 

Also converted the page to markdown, since the div nesting was becoming too much.

Staged at https://dartlang-org-staging-0.firebaseapp.com/guides/platforms

---

screen shot 2018-02-02 at 14 27 51

---

screen shot 2018-02-02 at 14 25 46


kwalrath:
I'd like the boxes (and thus logos) to be smaller, so you can see more info in less space. Maybe have the smaller logo after the title/link?

Can we use a gray (clear) background for the logos?

chalin:
Btw, as you might have noticed, I haven't changed the width of the cards. The current image size IMHO is what looks best with the current card width.

As for the image bg, IMHO we need something to visually separate the text from the image "tab".

Here is what smaller logos might look like:

screen shot 2018-02-02 at 15 35 42

(I don't currently have a good Flutter logo with a transparent bg.)

I've sampled a few Google sites (such as https://developers.google.com/web/fundamentals/), and most have images first. Logos after the text, prototyped below, doesn't look as good IMHO:

screen shot 2018-02-02 at 15 33 05

Which option do you prefer?
And who can I talk to to get a Flutter logo w/ transparent bg?

Maybe @anders-sandholm wants to weigh in?

kwalrath:
I'm not a fantastically visual person, so I'm going to let Anders or someone else (@wmleler?) chime in. My main concern is that people can find the info they need without scrolling.

kwalrath:
I'm trying to look at other web pages for inspiration... The Popular list on https://developers.google.com/products/ has smallish logos to the side of the text. I rather like that.

Also, I don't think we have to stick with cards. They take up a lot of space for no benefit I can see in a static page.

wmleler:
I don't have a strong preference, but I like the layout from https://firebase.google.com/ (scroll down to "Solutions to common challenges"). No extra borders/cards, image first. Large size doesn't bother me as long as it scales down as the viewport shrinks.

anders-sandholm:
@chalin thanks for taking a stab at this. 

Looking at the mobile version on my phone, I'd agree with @kwalrath that smaller icons would be better (unless we can automatically scale things down on smaller displays).

Also happy to lose the cards. That could also bring us closer to the look and feel of the referenced sites earlier in this thread - like https://firebase.google.com/ and https://developers.google.com/products/ (plus at least temporarily eliminate the need for a transparent Flutter logo ;-) )

chalin:
Ok, how do you find the following? It is modelled after https://developers.google.com/products, where 100x100 icons are used:

---

screen shot 2018-02-05 at 11 00 28

---

screen shot 2018-02-05 at 10 59 39

---

chalin:
> Whether you're creating a mobile app, web app, command-line script, or server, there's

Change made. Merging.

PR #555, opened 2018-02-03
Simplify Get Started page

### LINKED ISSUE #552: Simplify the Get Started page
I love the simplification of the front page and hope that we could do something similar for this page...

I'd like to propose that we
1) turn the bullet list into a sentence: "This website (www.dartlang.org) contains information about the Dart _language_ and _core libraries_".
2) Move the "use case table" down below these two sentences and potentially replace it with the contents of the simpler "Platforms" page which has the same amount of information.
3) Remove the helpful flowchart. I don't think it's that helpful / adding much beyond the "plaforms" info.
4) I'd even consider removing the remaining three sections as well (books, samples, get help). The two latter are easily reachable from the menus.

kwalrath:
Also (re)moved stuff in the DartPad docs. Staged at:

https://kw-staging-dartlang-2.firebaseapp.com/guides/get-started
https://kw-staging-dartlang-2.firebaseapp.com/tools/dartpad

Fixes #552

### DIFF
--- a/src/_guides/get-started.md
+++ b/src/_guides/get-started.md
@@ -1,14 +1,15 @@
 ---
 layout: guide
 title: Get Started
-description: "A guide for getting started with Dart."
+description: "A guide to getting started with Dart."
 toc: false
 ---
 
 Jump right in. Use the embedded DartPad below to play with Dart and to experience the language and core APIs.
-[DartPad](/tools/dartpad) is a quick and easy way to
-become familiar with Dart language features.
-You can also open <a href="http://dartpad.dartlang.org" target="_blank">DartPad in a new window</a>.
+DartPad is a quick and easy way to
+become familiar with Dart language features
+([more info](/tools/dartpad)).
+You can also open <a href="http://dartpad.dartlang.org" target="_blank">DartPad in a new window.</a>
 
 <iframe
 src="{{site.custom.dartpad.embed-dart-prefix}}?horizontalRatio=70&verticalRatio=65"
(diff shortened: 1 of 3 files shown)


src/tools/dartpad.md
```diff
@@ -13,6 +13,16 @@ Here's what DartPad looks like:
 <img src="images/DartPadWindow.png" alt="DartPad screenshot" />
 
 
+## Library support
+
+DartPad supports [dart:* libraries]({{site.dart_api}}) that work with web apps;
+it doesn't support [dart:io]({{site.dart_api}}/stable/dart-io) or
+libraries from [packages.]({{site.pub}})
```
filiph: nit: "libraries from packages" --> "libraries from pub.dartlang.org"? I know it's technically less correct but "libraries from packages" sounds like "programs from files" to me.

Not feeling strongly about this, feel free to ignore.

src/_guides/get-started.md
```diff
@@ -20,63 +21,26 @@ src="{{site.custom.dartpad.embed-dart-prefix}}?horizontalRatio=70&verticalRatio=
...
-## Helpful flowchart
-
-{% comment %}
-The original Lucid Chart is here:
-https://www.lucidchart.com/documents/edit/0f170001-d5f4-4b17-8cdc-72d5575f2e78?shared=true&#
-Created by filiph, the Lucid app is available to Googlers in mystuff.
-{% endcomment %}
+To learn about the Dart [language](/guides/language) and
+[core libraries](/guides/libraries),
+stay on this site.
```
filiph: [nit]: `"Learn about the Dart [language] and [core libraries] on this site. Once you have a use case ...."`

src/tools/dartpad.md
kwalrath: I started implementing this, but I couldn't figure out how to make the next sentence work well if I didn't introduce "packages" here.

Issue #553, opened 2018-02-02
Consider using Travis build stages


kwalrath:
Can we use build stages to reduce the amount of resources we need for Travis? This could reduce our build time and help out other dart-lang projects.

Related: https://github.com/dart-lang/angular/pull/828

chalin:
I've looked into stages already and AFAIK, it wouldn't buy us much for our site-* repos. In particular, as is mentioned under [Data persistence between stages and jobs](https://docs.travis-ci.com/user/build-stages/#Data-persistence-between-stages-and-jobs):

> It is important to note that jobs do not share storage, as each job runs in a fresh VM or container.

natebosch:
They don't share storage directly, but they can share a cache
https://github.com/dart-lang/angular/blob/d78c9d9737c02b6622c84220d45e10273425914a/.travis.yml#L47-L52

jakemac53:
See https://docs.travis-ci.com/user/build-stages/warm-cache/ for the travis docs on warming up the cache

kwalrath:
From #559: This is going to take a while to setup, partly because we have a hybrid environment using Ruby (for Jekyll), Node (for tooling like the shredder), and Dart. The interplay with nvm, rvm and the cache is tricky to get right. Overall, this is unlikely to save much time since typically builds currently require about 7 min to complete.

Talking about this a bit more, it seems that optimizing the site-www build might not be as important as optimizing site-webdev: site-webdev builds take much longer (up to 2 hours, total time). We need to figure out what our goals are: how will we know when we've optimized the build process enough?

chalin:
In the meantime, I'll interleave improving the Travis build efficiency of site-webdev, along with the other site-www and site-webdev work. The build efficiency can be improved, it'll just take a little while before we get there (with all that is going on atm).

chalin:
I've tuned the build process and its use of **Travis caches** and here are the numbers:

- Old build process: .e.g., [**15 min (20 min total)**](https://travis-ci.org/dart-lang/site-www/builds/332063661)
- New build process takes, e.g., [**7.5 min (15 min total)**](https://travis-ci.org/dart-lang/site-www/builds/341910514).

I've also ensured that PRs only launch one Travis build (as opposed to two when no special configuration is used).

This site's build is too simple, there isn't **anything worth staging**. I'd suggest closing this issue.

WDYT?

chalin:
[...]

PR #559, opened 2018-02-06
[WIP] Use Travis stages

chalin:
WIP towards #553.

This is going to take a while to setup, partly because we have a hybrid environment using Ruby (for Jekyll), Node (for tooling like the shredder), and Dart. The interplay with nvm, rvm and the cache is tricky to get right. 

Overall, this is unlikely to save much time since typically builds currently require about 7 min to complete.

cc @kwalrath

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,12 +6,16 @@ cache:
   directories:
     - $HOME/.nvm
     - $HOME/.pub-cache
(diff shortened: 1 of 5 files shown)


chalin:
Closing, see https://github.com/dart-lang/site-www/issues/553#issuecomment-365958931.

PR #561, opened 2018-02-07
Adjustments to Strong Mode opening example

### LINKED ISSUE #560: Expected strong mode error not being reported by Dartpad
This is the opening example of the [Strong Mode page](https://www.dartlang.org/guides/language/sound-dart):

```dart
void fn(List<int> a) {
  print(a);
}

main() {
  var list = [];
  list.add(1);
  list.add("2");
  fn(list); // (*)
}
```

This example is available in a [DartPad Gist](https://dartpad.dartlang.org/3c7c95683f0c06be8326a2fd3975cd19).

According to the prose accompanying this example, a type error should be reported due to the unsound cast at (*). But DartPad reports no static or runtime errors when this code is analyzed or run.

This was discussed offline. This is what is happening in DartPad:

1. No static error is reported because such errors are now controlled by an analyzer option (`--no-implicit-casts`) that enables implicit casts by default.
2. No runtime error is reported because DartPad uses dart2js, and dart2js does not yet fully implement strong mode semantics.

The prose accompanying the example should be adjusted, and until DartPad fully implements strong mode semantics, the link to the DartPad Gist should probably be dropped.

cc @kwalrath @stereotype441 

(Related: #407)

chalin:
Fixes #560 

### DIFF
--- a/src/_guides/language/sound-dart.md
+++ b/src/_guides/language/sound-dart.md
@@ -66,8 +66,9 @@ The error, reported at runtime and by the analyzer (when
 highlights an unsound implicit cast from `List<dynamic>` to `List<int>`.
 The `list` variable has static type `List<dynamic>`. This is because the
 initializing declaration `var list = []` doesn't provide the analyzer with
-enough information for it to infer a type more specific than `dynamic`.
-The `fn()` function expects a list of type `int`, causing a mismatch of types.
+enough information for it to infer a type argument more specific than `dynamic`.
+The `fn()` function expects a parameter of type `List<int>`,
+causing a mismatch of types.
 
 When adding a type annotation (`<int>`) on creation of the list
 (highlighted below) the analyzer complains that a string argument can't be assigned to
--- a/src/_guides/language/sound-dart.md
+++ b/src/_guides/language/sound-dart.md
@@ -12,7 +12,7 @@ You'll learn how to use strong mode to enable soundness, as well as
 how to substitute types safely when overriding methods.
 
 <aside class="alert alert-info" markdown="1">
-  The terms "**sound** Dart", "**strong mode** Dart", and "**type safe** Dart"
+  The terms **sound** Dart, **strong mode** Dart, and **type safe** Dart
   are sometimes used interchangeably. _Strong mode_ is a sound static
   type system that uses a combination of static and runtime checks to
   ensure your code is type safe&mdash;that you can never see a value
@@ -24,7 +24,7 @@ how to substitute types safely when overriding methods.
   is the only full implementation of strong mode.
   VM and dart2js support are on their way.
 
-  "**Classic Dart**" refers to Dart before soundness was added to the language.
+  **Classic** Dart refers to Dart before soundness was added to the language.
 </aside>
 
 By writing sound Dart code today, you'll reap some benefits now,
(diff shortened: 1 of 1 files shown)


stereotype441:
LGTM assuming @kwalrath agrees.

src/_guides/language/sound-dart.md
```diff
@@ -37,41 +37,46 @@ developer experience.
...
+```nocode
+The argument type 'List' can't be assigned to the parameter type 'List<int>'
+```
+
+The error, reported at runtime and by the analyzer (when
+[implicit casts are disabled](#call-dartanalyzer-with-strong-mode-enabled)),
+highlights an unsound implicit cast from `List<dynamic>` to `List<int>`.
+The `list` variable has static type `List<dynamic>`. This is because the
+initializing declaration `var list = []` doesn't provide the analyzer with
+enough information for it to infer a type more specific than `dynamic`.
```
eernstg: Drive-by comment: At this point, maybe `type argument` is better than `type`, given that we could actually type an expression evaluating to that list as `dynamic`. Similarly, we could say `a list of type `List<int>`, ..` in line 70, in order to avoid some potential confusion among readers.

chalin: Good points. I've made both changes.

chalin:
@eernstg - I've created #563 to ensure a followup to your last comment (which I'd rather address separately from this PR).

src/_guides/language/sound-dart.md
```diff
@@ -12,7 +12,7 @@ You'll learn how to use strong mode to enable soundness, as well as
 how to substitute types safely when overriding methods.
 
 <aside class="alert alert-info" markdown="1">
-  **Note:** The terms "sound Dart", "strong mode Dart", and "type safe Dart"
+  The terms "**sound** Dart", "**strong mode** Dart", and "**type safe** Dart"
```
kwalrath: GLOBAL: I like the switch to boldface, but let's drop the quotation marks when we use bold.

src/_guides/language/sound-dart.md
```diff
@@ -24,7 +24,7 @@ how to substitute types safely when overriding methods.
   is the only full implementation of strong mode.
   VM and dart2js support are on their way.
 
-  "Classic Dart" refers to Dart before soundness was added to the language.
+  "**Classic Dart**" refers to Dart before soundness was added to the language.
```
kwalrath: "**Classic Dart**" -> **Classic** Dart
(no quotes, no bold on "Dart")

src/_guides/language/sound-dart.md
```diff
@@ -37,41 +37,47 @@ developer experience.
 Strong mode Dart adds only a few additional rules beyond that for classic
 Dart&mdash;mostly you clarify code where the types are ambiguous or
 incorrect. In fact, most strong mode errors can be fixed by adding type
-annotations to your Lists and Maps. The following code shows valid
-but unsound Dart code. The `fn()` function prints an integer list.
-The `main()` function creates a list of integers and passes it to `fn()`:
+annotations to your Lists and Maps.
+For example, in the following code the `fn()` function prints an integer list,
```
kwalrath: Let's split this into two paragraphs, so we don't bury the tip.

We might even boldface **adding type annotations to your Lists and Maps.** What do you think?

chalin: Paragraphs split.

> We might even boldface **adding type annotations to your Lists and Maps.** What do you think?

My first reaction would be to not put it in bold. I have a proposal to rework the initial paragraphs. Maybe we can revisit using bold or not then

PR #564, opened 2018-02-07
Strong Mode page: intro rework, add warning about DartPad

### LINKED ISSUE #563: Strong Mode page: clarify that DartPad uses Dart 1 dynamic semantics
From https://github.com/dart-lang/site-www/pull/561#pullrequestreview-94730866:

> ... The section 'Enable strong mode in DartPad' on https://www.dartlang.org/guides/language/sound-dart does not mention the dynamic semantics: It should say that the dynamic semantics of DartPad execution is Dart 1 semantics (and hence we get such behaviors as letting a variable/parameter of type `List<int>` refer to an instance of `List<dynamic>`) --- that's really crucial information to give the reader who wants to understand the text that's being modified in this PR, and the behavior of DartPad if they start experimenting with the example.

cc @stereotype441 @eernstg @kwalrath

chalin:
Fixes #563 

### DIFF
--- a/src/_guides/language/sound-dart.md
+++ b/src/_guides/language/sound-dart.md
@@ -11,20 +11,27 @@ This guide tells you why and how to write sound (type safe) Dart code.
 You'll learn how to use strong mode to enable soundness, as well as
 how to substitute types safely when overriding methods.
 
+_**Strong mode**_ is a sound static
+type system that uses a combination of static and runtime checks to
+ensure your code is type safe&mdash;that you can never see a value
+whose runtime type does not match its static type.
+With strong mode enabled (in an implementation that has both the
+static and runtime checks), Dart is a sound language.
+
 <aside class="alert alert-info" markdown="1">
-  The terms **sound** Dart, **strong mode** Dart, and **type safe** Dart
-  are sometimes used interchangeably. _Strong mode_ is a sound static
-  type system that uses a combination of static and runtime checks to
-  ensure your code is type safe&mdash;that you can never see a value
-  whose runtime type does not match its static type.
-  With strong mode enabled (in an implementation that has both the
-  static and runtime checks), Dart is a sound language.
+#### Terminology
+
+  - The terms **sound** Dart, **strong mode** Dart, and **type safe** Dart
+    are sometimes used interchangeably.
+  - **Classic** Dart refers to Dart before soundness was added to the language.
+</aside>
+
+<aside class="alert alert-warning" markdown="1">
+  **Warning:**
   Currently, the Dart dev compiler
   ([dartdevc,]({{site.webdev}}/tools/dartdevc) also known as _DDC_)
   is the only full implementation of strong mode.
   VM and dart2js support are on their way.
-
-  **Classic** Dart refers to Dart before soundness was added to the language.
 </aside>
 
 By writing sound Dart code today, you'll reap some benefits now,
@@ -95,6 +102,7 @@ Gist:  https://gist.github.com/3c7c95683f0c06be8326a2fd3975cd19
 DartPad url: https://dartpad.dartlang.org/3c7c95683f0c06be8326a2fd3975cd19
 
 Note: Can't use DP because the runtime doesn't implement strong mode yet:
+https://github.com/dart-lang/dart-services/issues/334
 [Try it in DartPad](https://dartpad.dartlang.org/3c7c95683f0c06be8326a2fd3975cd19).
 {% endcomment %}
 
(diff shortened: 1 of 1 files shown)


src/_guides/language/sound-dart.md
```diff
@@ -11,20 +11,27 @@ This guide tells you why and how to write sound (type safe) Dart code.
 You'll learn how to use strong mode to enable soundness, as well as
 how to substitute types safely when overriding methods.
 
+_**Strong mode**_ is a sound static
+type system that uses a combination of static and runtime checks to
+ensure your code is type safe&mdash;that you can never see a value
+whose runtime type does not match its static type.
+With strong mode enabled (in an implementation that has both the
+static and runtime checks), Dart is a sound language.
+
```
chalin: The text from lines 14-20 is not new text, it used to be inside the `<aside>`.

src/_guides/language/sound-dart.md
```diff
@@ -507,7 +515,15 @@ For more information on these flags, see
...
 If you use [DartPad](/tools/dartpad) to write and test code, you can
 enable strong mode by selecting the **Strong mode** box in the lower
-right corner. Note that DartPad doesn't support the implicit casts flag,
+right corner.
+
+<aside class="alert alert-warning" markdown="1">
+  **Warning:**
+  While DartPad implements strong mode _static checks_,
+  DartPad does **not** currently implement strong mode _runtime checks_.
+</aside>
```
chalin: @eernstg - how's this as a warning about the partial strong mode support in DartPad?  (FYI, I deliberately chose not to mention "Dart 1 runtime".)

src/_guides/language/sound-dart.md
kwalrath: But let's use just bold font, not bold italic.

src/_guides/language/sound-dart.md
```diff
@@ -11,20 +11,27 @@ This guide tells you why and how to write sound (type safe) Dart code.
...
+
 <aside class="alert alert-info" markdown="1">
-  The terms **sound** Dart, **strong mode** Dart, and **type safe** Dart
-  are sometimes used interchangeably. _Strong mode_ is a sound static
-  type system that uses a combination of static and runtime checks to
-  ensure your code is type safe&mdash;that you can never see a value
-  whose runtime type does not match its static type.
-  With strong mode enabled (in an implementation that has both the
-  static and runtime checks), Dart is a sound language.
+#### Terminology
```
kwalrath: This format is unusual. It's kind of like the summaries we have in some tutorials (e.g. https://www.dartlang.org/tutorials/language/streams), but the background is different _and_ this isn't a summary.

The usual way of formatting a note (which I think should work here) is like this:

**Terminology note:** The terms.... **Classic** Dart....

src/_guides/language/sound-dart.md
```diff
@@ -11,20 +11,27 @@ This guide tells you why and how to write sound (type safe) Dart code.
...
+
+  - The terms **sound** Dart, **strong mode** Dart, and **type safe** Dart
+    are sometimes used interchangeably.
+  - **Classic** Dart refers to Dart before soundness was added to the language.
+</aside>
+
+<aside class="alert alert-warning" markdown="1">
```
kwalrath: Let's put the warning first, since it's more important.

I try to avoid having two notes in a row... Maybe put the terminology note after the next paragraph?

src/_guides/language/sound-dart.md
```diff
@@ -11,20 +11,27 @@ This guide tells you why and how to write sound (type safe) Dart code.
...
+  **Warning:**
   Currently, the Dart dev compiler
   ([dartdevc,]({{site.webdev}}/tools/dartdevc) also known as _DDC_)
   is the only full implementation of strong mode.
   VM and dart2js support are on their way.
```
kwalrath: "on their way" could link to tracking bugs.

src/_guides/language/sound-dart.md
eernstg: I like it!  But maybe changing `currently` to `yet` would induce more hope, because it implies that DartPad definitely will support those checks.

chalin: Yes! Done.

src/_guides/language/sound-dart.md
chalin: > Let's put the warning first, since it's more important.

Done.

> Maybe put the terminology note after the next paragraph?

I agree if these were two notes, but a warning next to a note seems better than the alternative (I tried both). I'll leave it for now. We can move it later if you feel strongly about it.

src/_guides/language/sound-dart.md
chalin: I've searched through the SDK issues but didn't find anything appropriate. Maybe you can ask around?

Issue #562, opened 2018-02-07
Add more common strong-mode issues


matanlurey:
Page:
https://dart.dev/guides/language/sound-problems
https://github.com/dart-lang/site-www/blob/master/src/_guides/language/sound-problems.md

I'd love to see two more items here, and I'm happy to help send/edit:

* [x] `uses_dynamic_as_bottom` (contravariant function type parameters, etc)
* [x] Invalid type casts (`JSArray is not List<String>`)
* [ ] type parameter conflicts
* [ ] callable classes

/cc @leafpetersen @lrhn @srawlins @munificent

kwalrath:
If you could do a quick first pass (even just pointing to before and after examples, along with any other text you can easily put together), I'd be happy to finish it and publish it.

matanlurey:
From @leafpetersen, two more issues we should highlight:

> Type parameter conflicts:
>  - A class may not implement (even transitively, anywhere in its superclass hierarchy) the same generic interface with different type parameters.
>  - So no having List<int> *and* List<double> in your class hierarchy somewhere.
>  - You can have the same generic interface with *the same* type parameters
> 
> Callable classes:
>   - A class with a call method can still be "called" using the call syntax: o(arguments)
>   - A function can still be called with the ".call" syntax
>   - But if 
>      - you use an object o with static type C
>      - where C has a call method
>      - in a context where a function typed thing is expected
>      - then this will be treated as a tearoff of the call method, instead of just passing the object directly
> 
> There's a good comment on this bug: https://github.com/dart-lang/sdk/issues/32151
> 
> with concrete examples of how Dart 2 will differ from Dart 1

srawlins:
I think GitHub Markdown ate your generics on `List` and `List`.

chalin:
> I think GitHub Markdown ate your generics on List and List.

Fixed.

atsansone:
@matanlurey : This issue will be closed as the documentation has likely been updated in the last 6 years. If this issue still persists with the existing documentation, could you please file a new issue? Thanks.

PR #568, opened 2018-02-09
Update type cast errors section

matanlurey:
... based on @munificent's internal doc, which I think does a much better job than I originally did (plus lots of my terminology is actually related to `uses_dynamic_as_bottom`, *not* this class of error/warning, so I'll save it for a future PR).

Work towards https://github.com/dart-lang/site-www/issues/562.

### DIFF
--- a/src/_guides/language/sound-problems.md
+++ b/src/_guides/language/sound-problems.md
@@ -394,7 +394,7 @@ In other cases, however, Dart needs to insert a _runtime_ check. Consider:
 {:.passes-sa}
 {% prettify dart %}
 assumeStrings(List<Object> objects) {
-  List<String> [!strings = objects!]; // <!-- Implicit downcast
+  List<String> [!strings = objects!]; // <- Implicit downcast
   String string = strings[0];         // Expect to get a type 'String' out.
 }
 {% endprettify %}
--- a/src/_guides/language/sound-problems.md
+++ b/src/_guides/language/sound-problems.md
@@ -451,7 +451,7 @@ You can also more precisely type the local variable, and let inference help:
 
 In cases where you are working with a collection that you don't create, such
 as from JSON or an external data source, you can use the
-[`cast` method](https://api.dartlang.org/dev/dart-core/List/cast.html)
+[`cast` method](...) --> [cast()]({{site.dart_api}}/dev/dart-core/List/cast.html) method
 provided by `List` (and other collection classes).
 
 Here's an example of the preferred solution: tightening the object's type.
--- a/src/_guides/language/sound-problems.md
+++ b/src/_guides/language/sound-problems.md
@@ -387,7 +387,7 @@ main() {
  
 Since neither `List<int>` nor `List<String>` is a subtype of the other,
 Dart rules this out statically. You can see other examples of these
-static analysis errors in [Assigning Mismatched Types](##assigning-mismatched-types).
+static analysis errors in [Assigning Mismatched Types](#assigning-mismatched-types).
 
 In other cases, however, Dart needs to insert a _runtime_ check. Consider:
 
--- a/src/_guides/language/sound-problems.md
+++ b/src/_guides/language/sound-problems.md
@@ -369,56 +369,99 @@ the command-line Dart VM with `--strong` (where supported).
 
 ### Invalid casts
 
-In Dart 1.x `dynamic` was both a [top type][top_type_wiki] (also known as a
-_universal supertype_, or the _supertype of all other types_) and a
-[bottom type][bottom_type_wiki] (also known as the _zero_ or _empty_ type)
-depending on the context. This meant it was legal to assign, for example,
-an object of type `List<dynamic>` to a `List<String>`:
+Unlike Dart 1.x, Dart 2 enforces a sound type system. This means (roughly)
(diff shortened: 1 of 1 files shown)


src/_guides/language/sound-problems.md
```diff
@@ -369,56 +369,99 @@ the command-line Dart VM with `--strong` (where supported).
...
-  List<String> names = json['names'];
+  List<int> numbers = [1, 2, 3];
+  List<String> [!string = numbers!]; // <-- Static analysis error
 }
-```
+{% endprettify %}
+ 
+Since neither `List<int>` nor `List<String>` is a subtype of the other,
+Dart rules this out statically. You can see other examples of these
+static analysis errors in [Assigning Mismatched Types](##assigning-mismatched-types).
```
chalin: Anchor has one `#` too many

src/_guides/language/sound-problems.md
```diff
@@ -369,56 +369,99 @@ the command-line Dart VM with `--strong` (where supported).
...
+For example, this type error is detected at compile-time:
 
-```dart
+{:.fails-sa}
+{% prettify dart %}
 main() {
-  Map json = {'names': []};
-  List<String> names = json['names'];
+  List<int> numbers = [1, 2, 3];
+  List<String> [!string = numbers!]; // <-- Static analysis error
```
chalin: Note that, by default, this is a warning. It is only an error if `--no-implicit-casts` is specified as a command option.

I'd suggest dropping the comment since the code is highlighted and the entire code excerpt shows "static analysis: error/warning".

Or consider changing the comment to mention the type of error:

```dart
List<String> string = numbers; // warning: invalid_assignment
```

The prose that follows the code excerpt gives the details.

src/_guides/language/sound-problems.md
```diff
@@ -369,56 +369,99 @@ the command-line Dart VM with `--strong` (where supported).
...
+Since neither `List<int>` nor `List<String>` is a subtype of the other,
+Dart rules this out statically. You can see other examples of these
+static analysis errors in [Assigning Mismatched Types](##assigning-mismatched-types).
+
+In other cases, however, Dart needs to insert a _runtime_ check. Consider:
+
+{:.passes-sa}
+{% prettify dart %}
+assumeStrings(List<Object> objects) {
+  List<String> [!strings = objects!]; // <!-- Implicit downcast
```
chalin: At least drop the `!` from `<!--`.

(The general style used elsewhere in the docs, is IMHO that we avoid using an ASCII arrow prefix like `<--`. If you feel like removing them, great. If not I can sweep the code excerpts later.)

src/_guides/language/sound-problems.md
```diff
@@ -369,56 +369,99 @@ the command-line Dart VM with `--strong` (where supported).
...
+{% prettify dart %}
+  List[!<String>!] list  = []; // <-- Give it an explicit left-hand type.
+  list.add('a string');
+  list.add('another');
+  assumeStrings(list);
+{% endprettify %}
+
+In cases where you are working with a collection that you don't create, such
+as from JSON or an external data source, you can use the
 [`cast` method](https://api.dartlang.org/dev/dart-core/List/cast.html)
```
chalin:
```markdown
[`cast` method](...) --> [cast()]({{site.dart_api}}/dev/dart-core/List/cast.html) method
```

src/_guides/language/sound-problems.md
matanlurey: I'd prefer not to add any information around `--no-implicit-casts`, as it's not officially supported (and we aren't offering any support it).

matanlurey:
Done. PTAL @chalin @kwalrath

src/_guides/language/sound-problems.md
chalin: Oops, I just double checked: it _is_ an error even w/o the `--no-implicit-casts` (I had forgotten the `--strong` option when I did the test; w/o `--strong` it is just a warning). So the example is good as is. Apologies for the noise.

src/_guides/language/sound-problems.md
```diff
@@ -451,7 +451,7 @@ You can also more precisely type the local variable, and let inference help:
 
 In cases where you are working with a collection that you don't create, such
 as from JSON or an external data source, you can use the
-[`cast` method](https://api.dartlang.org/dev/dart-core/List/cast.html)
+[`cast` method](...) --> [cast()]({{site.dart_api}}/dev/dart-core/List/cast.html) method
```
chalin: Oops, maybe a copy-paste error? Delete the leading

```markdown
[`cast` method](...) --> 
```

matanlurey: Done... I think

chalin: I push the necessary tweak.

matanlurey:
Yay! @kwalrath could you merge?

chalin:
@kwalrath gave the thumbs up (offline). CLAs are ok. Merging.

matanlurey:
Yayyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy

chalin:
Thanks for the revision to this section @matanlurey!

PR #586, opened 2018-02-16
Enable no-implicit-casts: bug fix, and non-user-visible code changes

chalin:
These are code fixes necessary once the analyzer option `implicit-casts` is set to false. This helped identify one code bug, which has been fixed. Essentially, no other code change is visible in the docs (modulo Gists which aren't being updated for minor changes), either because the change was done to test code, or because a replace expressions was added to a code-excerpt instruction.


examples/misc/lib/effective_dart/usage_bad.dart
```diff
@@ -13,10 +13,9 @@ void miscDeclAnalyzedButNotTested() {
     // #enddocregion adjacent-strings-literals
   }
 
-  (name, year, birth) {
+  (String name, num year, num birth) {
     // #docregion string-interpolation
-    'Hello, ' + name + '! You are ' + (year - birth) + ' y...';
-
+    'Hello, ' + name + '! You are ' + (year - birth).toString() + ' y...';
```
chalin: This code excerpt was actually invalid, given that it is assumed `year` and `birth` are numbers (`.toString()` was missing).

PR #587, opened 2018-02-16
Enable no-implicit-casts: code fixes that are user visible

chalin:
@kwalrath - I'd like to set strong mode `implicit-casts` to false for the example code. It has already helped identify a bug in our effective Dart code (see #586).

#586 included the bug fix and non-user-visible changes.

I've kept user visible changes for this PR (actually, only the language tour is affect). The changes are minor, but worth it IMHO. PTAL

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -2601,7 +2601,7 @@ initializers with commas.
 {% prettify dart %}
 // Initializer list sets instance variables before
 // the constructor body runs.
-Point.fromJson(Map json)
+Point.fromJson(Map<String, num> json)
     : x = json['x'],
       y = json['y'] {
   print('In Point.fromJson(): ($x, $y)');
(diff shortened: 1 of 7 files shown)


src/_guides/language/language-tour.md
```diff
@@ -2601,7 +2601,7 @@ initializers with commas.
 {% prettify dart %}
 // Initializer list sets instance variables before
 // the constructor body runs.
-Point.fromJson(Map json)
+Point.fromJson(Map<String, num> json)
```
kwalrath: This is OK for now, but we're going to need to introduce generics earlier. I'll file an issue.

src/_guides/language/language-tour.md
```diff
@@ -2630,7 +2630,7 @@ class Point {
   final num y;
   final num distanceFromOrigin;
 
-  Point(x, y)
+  Point(num x, num y)
```
kwalrath: This looked really weird at first, but now I get it. (You aren't referring to the instance variables here; it looks weird because this isn't idiomatic Dart.)

PR #590, opened 2018-02-20
Strong mode code excerpt tests for main page

chalin:
Contributes to #407 

### DIFF
--- a/examples/misc/lib/language_tour/built_in_types.dart
+++ b/examples/misc/lib/language_tour/built_in_types.dart
@@ -6,6 +6,7 @@ void miscDeclAnalyzedButNotTested() {
     // #docregion integer-literals
     int x = 1;
     int hex = 0xDEADBEEF;
+    // ignore_for_file: 2, integer_literal_out_of_range
     int bigInt = 34653465834652437659238476592374958739845729;
     // #enddocregion integer-literals
   }
--- a/src/_guides/language/sound-dart.md
+++ b/src/_guides/language/sound-dart.md
@@ -215,7 +215,7 @@ The return type of a method in a subclass must be the same type or a
 subtype of the return type of the method in the superclass. Consider
 the getter method in the Animal class:
 
-<?code-excerpt "strong/lib/animal.dart (Animal-wo-eat)" replace="/Animal get.*/[!$&!]/g"?>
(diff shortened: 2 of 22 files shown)


src/_guides/language/sound-dart.md
```diff
@@ -601,31 +620,9 @@ In the following example, you can substitute
...
-  // Was: List<Cat> myCats = new List<Cat>();
-  List<Cat> myCats = new List<[!MaineCoon!]>();
-  Cat muffin = new Cat();
-  Cat winky = new Cat();
-  Cat bongo = new Cat();
-  myCats.addAll([muffin, winky, bongo]);
-
-  feedAnimals(myCats);
-}
+List<Cat> myCats = new List<[!MaineCoon!]>();
```
kwalrath: Before this, please add the comment that was there before:

```
  // Was: List<Cat> myCats = new List<Cat>();
```

examples/misc/lib/language_tour/built_in_types.dart
```diff
@@ -6,6 +6,7 @@ void miscDeclAnalyzedButNotTested() {
     // #docregion integer-literals
     int x = 1;
     int hex = 0xDEADBEEF;
+    // ignore_for_file: 2, integer_literal_out_of_range
     int bigInt = 34653465834652437659238476592374958739845729;
```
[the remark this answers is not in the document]
chalin: Good question. How about we keep it for this PR, and then generally decide how to document https://github.com/dart-lang/sdk/blob/master/sdk/lib/core/bigint.dart?

src/_guides/language/sound-dart.md
chalin: Sure but I don't think that it is relevant to the presentation. What is being contrasted in the code excerpt is the type argument to `List` on the LHS vs. the type argument to `List` on the RHS.

I'll leave out the comment for now and I'll revisit later if you feel the need.

kwalrath: The text "In the following example, you can substitute new List<Cat>() with new List<MaineCoon>()" refers to code (new List<Cat>()) that I actually looked above to find, since it seemed to be saying that I'd seen that code before.

chalin: Ah, you are right. #607 fixes that. (Given that #607 is a minor fix to this approved PR, I'll assume that it is ok to merge it once Travis is green.)

PR #592, opened 2018-02-21
 Reinstate Install Dart button + warn Flutter devs

### LINKED ISSUE #576: Install link should be more prominent on the home page
If I'm in a hurry I'd like to be able to visit dartlang.org and ctrl-f "install" and find what I'm looking for.

chalin:
This PR reinstates the frontpage Install Dart button which had been removed in #533. It also adds a clear visual warning for Flutter devs.

Fixes #576 

Staged at:
- https://dartlang-org-staging-0.firebaseapp.com
- https://dartlang-org-staging-0.firebaseapp.com/install

---

Here is the alert for Flutter devs:
> screen shot 2018-02-21 at 11 37 08

### DIFF
--- a/src/_assets/stylesheets/main.scss
+++ b/src/_assets/stylesheets/main.scss
@@ -985,6 +985,7 @@ li.card {
 .alert-with-image {
   display: table-row;
   > img {
+    margin-right: $content-padding * 2;
     width: $font-size-base * 3;
     @media(max-width: $screen-xs-min) { display: none; }
   }
--- a/src/_assets/stylesheets/main.scss
+++ b/src/_assets/stylesheets/main.scss
@@ -983,12 +983,15 @@ li.card {
 }
 
 .alert-with-image {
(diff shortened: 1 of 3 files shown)


chalin:
> If you want to use Dart for web or server-side development, install...

Done.

As for the sentence length, it seems ok to me (a tad on the long side IMHO, but still ok). WDYT?

kwalrath:
I think we can be even clearer for Flutter developers. Maybe:

**Important:** The Flutter SDK contains Dart. To develop Flutter apps, **don't install the Dart SDK separately.** Instead, [install Flutter]() or [upgrade Flutter]().

kwalrath:
(And I think you're right about sentence length: it's long but still OK.)

kwalrath:
Also... I don't think it should hold up this PR, but we should get a version of the Flutter logo that has a transparent background. @wmleler does such a beast exist?

PR #597, opened 2018-02-22
Add list of Dart 2 resources

kwalrath:
People were having trouble finding some of these, so I collected all the links that were already in the Dart 2 page (plus some) and listed them at the end.

Staged at: https://kw-www-dartlang-1.firebaseapp.com/dart-2#resources

/cc @anders-sandholm @kevmoo @filiph


anders-sandholm:
Awesome - thanks for putting the list together.

PR #623, opened 2018-03-03
Fixes typo: `substract` => `subtract`.

sxn:
Fixes a small typo on the [sound dart](https://www.dartlang.org/guides/language/sound-dart) page.


chalin:
Good catch, thanks!

PR #625, opened 2018-03-04
Remove toc from tutorials main page

### LINKED ISSUE #624: 'Dart Tutorials' page UI Issue
From URL: https://www.dartlang.org/tutorials

image

The menu bar on the right side has UI issue.
#toc .nav li {
    border-left: 3px solid #111111; //must be removed.
}

chalin:
Fixes #624

### DIFF
--- a/src/_guides/index.md
+++ b/src/_guides/index.md
@@ -1,7 +1,6 @@
 ---
-layout: default
-title: "Guides"
-description: "Read about the Dart language and libraries with this collection of guides."
+title: Guides
+description: Read about the Dart language and libraries with this collection of guides.
 permalink: /guides
 toc: false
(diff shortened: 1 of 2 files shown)


kwalrath:
That fixes the problem for this page, but I think there's a larger design issue of whether you should have that doubly thick line in the TOC. I'll file an issue.

chalin:
IMHO, the issue wasn't with the double line (which is there to indicate which subheading is active), it was because the cards contain headings and the sidenav was trying to show then as active.

kwalrath:
I interpreted it differently, based on CSS snippet and the orange rectangle around the line. @krishnatandon1208, could you tell us what you meant?

Issue #620, opened 2018-03-01
Rows are too close together in IDE list


kwalrath:
https://www.dartlang.org/tools needs more vertical space between rows.

PR #627, opened 2018-03-04
Add space between rows in IDE list

aymyers:
Add padding-bottom of 15px to IDE list items so they aren't so close together on https://www.dartlang.org/tools

In response to issue #620

### DIFF
--- a/src/_assets/stylesheets/main.scss
+++ b/src/_assets/stylesheets/main.scss
@@ -710,6 +710,7 @@ main .content {
   column-gap: 0;
   li {
     page-break-inside: avoid;
+    padding-bottom: 15px;
   }
 }
 .col3 {
--- a/src/_assets/stylesheets/main.scss
+++ b/src/_assets/stylesheets/main.scss
@@ -708,9 +708,11 @@ main .content {
   list-style: none;
   column-count: 2;
   column-gap: 0;
+  padding-left: $font-size-base / 2;
   li {
     page-break-inside: avoid;
-    padding-bottom: 15px;
+    padding: $font-size-base / 2;
+    font-size: smaller;
   }
 }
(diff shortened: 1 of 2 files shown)


src/_assets/stylesheets/main.scss
```diff
@@ -710,6 +710,7 @@ main .content {
   column-gap: 0;
```
chalin: For `.col2` I'd suggest adding:
```css
  padding-left: $font-size-base / 2;
```

src/_assets/stylesheets/main.scss
```diff
@@ -710,6 +710,7 @@ main .content {
...
   li {
     page-break-inside: avoid;
+    padding-bottom: 15px;
```
chalin: I'd prefer
```css
    padding: $font-size-base / 2;
    font-size: smaller;
```

chalin:
The extra CSS style changes above give make this look like this on mobile:

> screen shot 2018-03-04 at 15 50 23


A few other things we can do to clean this up (@aymyers, you don't have to do these unless you'd like the extra challenge :):

- Add `class="no-automatic-external"` to the anchors; that view is busy enough without the external link icon.
- Convert the `<ul>` HTML list markup to markdown. To add a class to the list, put `{:.col2}` just before the first markdown list entry.

@kwalrath - WDYT about this extra suggestions. @aymyers - feel free to ignore these suggestions (certainly at least until @kwalrath chims in).

chalin:
@kwalrath - I'll merge now. We can make any further adjustments in a followup PR.

kwalrath:
Thanks very much for the contribution, @aymyers!

aymyers:
You're welcome!

PR #648, opened 2018-03-12
Update language-tour.md

### LINKED ISSUE #572: In Dart 2, first arg to assert() can't be a function
See dart-lang/sdk#30326.

I think the only affected doc is the last paragraph of https://www.dartlang.org/guides/language/language-tour#assert.

anders-sandholm:
Update assert description to only allow expressions of type boolean (and not allow functions).
Fixes #572.

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -2122,8 +2122,8 @@ a post on http://news.dartlang.org.] -->
 </div>
 
 The first argument to `assert` can be any expression that
-resolves to a boolean value or to a function. If the expression’s value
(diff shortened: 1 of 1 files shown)


Sfshaza:
LGTM

kwalrath:
Should we make a point of updating the other branch to say this change will be coming? (If so, we shouldn't close the bug, but note that the dart 1 docs should be updated.)

chalin:
> Should we make a point of updating the other branch to say this change will be coming?

It has been my understanding that we only make changes to the 1.x branch if something is "broken".

anders-sandholm:
Merging then.

PR #665, opened 2018-03-14
Updating front page with Dart 2 core tenets

anders-sandholm:
Replace the old core goals with Dart 2's new core tenets.

@timsneath @dgrove FYI and/or for feedback.  

FWIW, I'm planning on an additional PR adding a section on "Client-Side Uses of Dart" mentioning/showcasing Flutter and Webdev.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,7 +1,7 @@
 ---
 layout: homepage
 title: Dart programming language
-description: Dart is an application programming language that’s easy to learn, easy to scale, and deployable everywhere.
+description: Dart is a language uniquely optimized for client-side development for web and mobile.
 ---
 <section id="code-sample-section">
 <div class="container">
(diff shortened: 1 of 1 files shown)


kwalrath:
I've staged this: https://kw-staging-dartlang-2.firebaseapp.com

src/index.html
```diff
@@ -78,22 +78,39 @@ <h4><a href="https://pub.dartlang.org/">Dart packages</a></h4>
     <div class="row">
       <div class="col-sm-5 col-lg-4">
         <div class="content">
-          <h1>Core goals</h1>
+          <h1>Dart’s Core Tenets</h1>
```
kwalrath: "Tenets" doesn't seem like a word most people know. What's wrong with "goals"?

src/index.html
```diff
@@ -78,22 +78,39 @@ <h4><a href="https://pub.dartlang.org/">Dart packages</a></h4>
...
+          <p>Client developers have to think about three platforms today: iOS, Android, and Web. The language needs to work 
+            well on all of them.</p>
+          <p>Dart supports compilation to native code (ARM, x86, etc.) for fast mobile performance as well as transpilation 
+            to efficient JavaScript for the web.</p>
+          <h2>Approachable</h2>
+          <p>The language can’t stray too far from the familiar if it wishes to be relevant for millions of developers.</p>
+          <p>Dart is approachable to many existing developers, thanks to its unsurprising object-oriented aspects and syntax 
+            that — according to our users — allows any C++, C#, Objective-C, or Java developer to be productive in a matter 
+            of days.</p>
+          <h2>Reactive</h2>
```
kwalrath: I wouldn't say "Dart is reactive", which is how the intro sets this up ("Dart ~is~ **helps you be** productive", "Dart is fast", "Dart is portable", "Dart is approachable", "Dart ~is~ **supports** reactive **programming**"). Some of the other adjectives don't work as well, but "reactive" stands out to me. Maybe "Flexible"? But it seems like you want "Reactive" in a big heading?

I don't have a good suggestion for how to fix this, and I suppose I could live with it if we can't come up with something better.

src/index.html
```diff
@@ -1,7 +1,7 @@
 ---
 layout: homepage
 title: Dart programming language
-description: Dart is an application programming language that’s easy to learn, easy to scale, and deployable everywhere.
+description: Dart is a language uniquely optimized for client-side development for web and mobile.
```
kwalrath: "uniquely" seems unnecessary and sounds like bragging to me. But if we really want to say this and will back it up, I'm OK with it. (This description shows up in search engine results and social shares.)

src/index.html
kwalrath: If you think "goals" implies something we might not have achieved, "principles" is more familiar than "tenets". But maybe we can be less lofty and just say something like "Why Dart?" or "What's special about Dart" or maybe just drop the heading entirely.

anders-sandholm: I like the idea of using "principles" instead. But actually, "Why Dart?" is probably a better way to both get attention and also a good headline for what we are trying to articulate in this section.

src/index.html
anders-sandholm: I'd like to try to reword the intro section so this is less about what Dart strictly speaking "is" and more about the key benefits you get from using Dart.

src/index.html
anders-sandholm: Happy to leave out uniquely for now. @kevmoo also reacted to the use of this word in a different context.

kwalrath:
Much better! I'll do a copy edit now, and see if I can add it as a commit to this PR.

kwalrath:
On second thought... If you have ideas of how to rewrite the feature descriptions to not be so principle-y... go ahead. (If you want me to have a go, I can. It's just not a simple copy edit.)

anders-sandholm:
Feel free to take a stab at it. 
Fine that it's not a simple copy edit. 
We can also move the iteration to docs if that's better / easier.

kwalrath:
I'm going to do this in a doc.

kwalrath:
Please take a look at the doc I shared, @anders-sandholm, @timsneath, and @dgrove.

kwalrath:
I've updated this PR with the changes from our doc and updated the staging server:

https://kw-staging-dartlang-2.firebaseapp.com

PTAL

src/index.html
```diff
@@ -80,38 +80,59 @@ <h4><a href="https://pub.dartlang.org/">Dart packages</a></h4>
...
+          </p>
+
           <h2>Portable</h2>
-          <p>Client developers have to think about three platforms today: iOS, Android, and Web. The language needs to work 
-            well on all of them.</p>
-          <p>Dart supports compilation to native code (ARM, x86, etc.) for fast mobile performance as well as transpilation 
-            to efficient JavaScript for the web.</p>
+          <p>
+            Dart compiles to native ARM and x86 code, so that Dart mobile apps
+            can run natively on iOS, Android, and beyond.
```
anders-sandholm: A little too much "native" in this sentence. I'd propose getting rid of the first one.

kwalrath: Good catch.

PR #666, opened 2018-03-14
Update pi_monte_carlo.dart

anders-sandholm:
Removing two occurrences of "new" because those are optional in Dart 2. 
The one "const" occurrence needs to stay.

Not sure if this will break any tests that we're running. I guess it depends on whether we're running the tests with a 2.0.0-dev* version or latest stable.

### DIFF
--- a/examples/misc/lib/pi_monte_carlo.dart
+++ b/examples/misc/lib/pi_monte_carlo.dart
@@ -40,9 +40,9 @@ Stream<double> computePi({int batch: 100000}) async* { //!tip("async*") //!tip("
 }
 
 Iterable<Point> generateRandom([int seed]) sync* { //!tip("sync*") //!tip("[int seed]") //!tip("Iterable")
-  final random = new Random(seed); //!tip("final")
+  final random = Random(seed); //!tip("final")
   while (true) {
-    yield new Point(random.nextDouble(), random.nextDouble()); //!tip("yield")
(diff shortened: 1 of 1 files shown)


chalin:
As can be seen from [this Travis log](https://travis-ci.org/dart-lang/site-www/jobs/353295863), we're using the latest dev release of Dart:

```nocode
$ dart --version
Dart VM version: 2.0.0-dev.36.0 (Tue Mar 13 11:04:03 2018 +0100) on "linux_x64"
```

But it would seem that new/const is still required by Dart VM?

```nocode
Analyzing lib, test...
  error • The function 'Random' isn't defined at lib/pi_monte_carlo.dart:43:18 • undefined_function
  error • The function 'Point' isn't defined at lib/pi_monte_carlo.dart:45:11 • undefined_function
  error • The type 'Iterable' implied by the 'yield' expression must be assignable to 'Iterable<Point>' at lib/pi_monte_carlo.dart:45:11 • yield_of_invalid_type
3 errors found.
```

We can either disable tests for this file (not my preferred solution), or wait until the Dart VM has support for optional new/const. WDYT?

cc @munificent

chalin:
I suppose that another alternative would be to try to run tests over the Flutter VM.

chalin:
FYI, I'm making progress on this (getting analysis and tests to pass over this modified example code). I'll post an update soon.

chalin:
Please see #669

kwalrath:
I'm going to close this in favor of #669.

Issue #507, opened 2018-01-11
Update language tour for Dart 2 changes


kwalrath:
* [x] Implicit new and const - lrn (dart-lang/sdk#30921) (partially fixed in #677)
* [x] Support assert in initializers - rnystrom (dart-lang/sdk#30968) (fixed in #677)
* [x] Fixed-sized integers limited to 64 bits - floitsch (dart-lang/sdk#28595, issue #591 ) (fixed in #677)
* [x] Generalized void - eernst (dart-lang/sdk#30176) (fixed [partially?] in #677)
* [ ] Eliminate fuzzy arrows - leafp (dart-lang/sdk#29630)
* [x] Start async functions synchronously - floitsch (fixed in #677)
* [x] @proxy & noSuchMethod (issue #442, pr #511)
* [x] First arg to assert can't be a function (issue #572)
* [x] Introduce generic syntax earlier (issue #588) (fixed in #677)
* [x] Asserts in initializers

kwalrath:
I'm working on the strong mode changes, which are pretty invasive.

sethladd:
This is so awesome to see!

kwalrath:
I don't know how to update the language tour for fuzzy arrows, so I'll be contacting @leafpetersen. We expect to start using implicit new/const everywhere for beta 3.

kwalrath:
We should cover asserts in initializers (and probably beef up the initializer coverage a bit).

kwalrath:
I'm starting on removing `new` and `const`.

chalin:
> I'm starting on removing `new` and `const`.

There is a PR pending for that: Remove all uses of "new" and "const" from "Effective Dart" #844. In particular, note this comment: https://github.com/dart-lang/site-www/pull/844#issuecomment-395498524. If we have the green light to go ahead with this, then great!

kwalrath:
> > I'm starting on removing `new` and `const`.
> 
> There is a PR pending for that: _Remove all uses of "new" and "const" from "Effective Dart"_ #844. In particular, note this comment: https://github.com/dart-lang/site-www/pull/844#issuecomment-395498524. If we have the green light to go ahead with this, then great!

Because Flutter's style (at least within constructors) is not to use `new`, I've been asked to avoid `new` whenever possible. The bar for Effective Dart is higher, because it sets conventions used everywhere.

kwalrath:
[...]

PR #677, opened 2018-03-15
Dart 2 changes to the language tour

kwalrath:
Contributes to #507. Staged at:

https://kw-www-dartlang-1.firebaseapp.com/guides/language/language-tour

Even though this page isn't completely done, if it's acceptable I'd like to commit and iterate.

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -9,6 +9,13 @@ This page shows you how to use each major Dart feature, from
 variables and operators to classes and libraries, with the assumption
 that you already know how to program in another language.
 
+<div class="alert alert-info" markdown="1">
+**Dart 2 note:**
+This tour is being updated for [Dart 2 changes](/dart-2),
+such as type safety, optional `new` and `const`,
+and changes to assertion support.
+</div>
+
 To learn more about Dart's core libraries, see
 [A Tour of the Dart Libraries](/guides/libraries/library-tour).
 
@@ -48,13 +55,14 @@ apps:
 
 <code>// <em>This is a comment.</em> </code>
 
-:   Use // to indicate that the rest of the line is a comment.
-    Alternatively, use /\* ... \*/. For details, see
-    [Comments](#comments).
+:   A way to make Dart ignore some content.
+    For more information, see [Comments](#comments).
 
 `num`
 
-:   A type. Some of the other built-in types are String, int, and bool.
+:   A type. A `num` variable's value can be either an `int` or a
+    `double` object.
+    Some of the other built-in types are `String`, `List`, and `bool`.
 
 `42`
 
@@ -86,9 +94,8 @@ apps:
 
 <div class="alert alert-info" markdown="1">
 **Note:**
-Our code follows the conventions in the
+This site's code follows the conventions in the
 [Dart style guide](/guides/language/effective-dart/style).
-For example, we use two-space indentation.
 </div>
 
 
@@ -101,21 +108,13 @@ mind:
     object is an instance of a *class*. Even numbers, functions, and
     `null` are objects. All objects inherit from the [Object][] class.
 
--   Specifying static types (such as `num` in the preceding example)
-    clarifies your intent and enables static checking by tools.
-    (You might notice when you’re debugging your code that
-    variables with no specified type get a special type: `dynamic`.)
-    Specifying static types is optional in Dart 1.x, however Dart is moving
-    towards being a fully type safe language.
+-   Although Dart is strongly typed, type annotations are optional
+    because Dart can often infer the type. In the code above, `number`
+    is inferred to be of type `int`. When you want to explicitly say
+    that no type is expected, use the special type `dynamic`.
 
--   In [strong mode](/guides/language/sound-dart), static and runtime
-    checks ensure that your code is type safe, helping you catch bugs
-    in development, rather than at runtime. Strong mode is optional in
-    Dart 1.x, but not optional in [Dart 2](/dart-2).
-
--   Dart parses all your code before running it. You can provide tips to
-    Dart—for example, by using types or compile-time constants—to catch
-    errors or help your code run faster.
+-   Dart supports generic types, like `List<int>` (a list of integers)
+    or `List<dynamic>` (a list of objects of any type). 
 
 -   Dart supports top-level functions (such as `main()`), as well as
     functions tied to a class or object (*static* and *instance
@@ -131,11 +130,11 @@ mind:
     private to its library. For details, see
     [Libraries and visibility](#libraries-and-visibility).
 
--   *Identifiers* can start with a letter or \_, followed by any
+-   *Identifiers* can start with a letter or underscore (\_), followed by any
     combination of those characters plus digits.
 
 -   Sometimes it matters whether something is an *expression* or a
-    *statement*, so we’ll be precise about those two words.
+    *statement*, so it helps to be precise about those two words.
 
 -   Dart tools can report two kinds of problems: warnings and errors.
     Warnings are just indications that your code might not work, but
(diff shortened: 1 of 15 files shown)


src/_guides/language/language-tour.md
```diff
@@ -2613,6 +2542,25 @@ Point.fromJson(Map<String, num> json)
 The right-hand side of an initializer does not have access to `this`.
 </div>
 
+During development, you can validate inputs by using `assert` in the
+initializer list.
+
+<?code-excerpt "misc/lib/language_tour/classes/point_alt.dart (initializer-list-with-assert)"?>
+{% prettify dart %}
+// Initializer list can use assert().
```
chalin: I'd suggest dropping the code comment and use highlighting on the snippet: `replace="/assert\(.*?\)/[!$&!]/g"`

src/_guides/language/language-tour.md
```diff
@@ -3323,34 +3204,28 @@ as E, T, S, K, and V.
...
 ### Why use generics?
 
-Because types are optional in Dart 1.x, you never *have* to use generics.
-You might *want* to, though, for the same reason you might want to use
-other types in your code: types (generic or not) let you document and
-annotate your code, making your intent clearer.
-
-{% include optional-types-2.0.html %}
+Generics are often required for type safety, but they have other benefits
+than just allowing your code to run:
```
chalin: other benefits --> more benefits
(otherwise the rest of the sentence doesn't work)

src/_guides/language/language-tour.md
```diff
@@ -3323,34 +3204,28 @@ as E, T, S, K, and V.
...
-the Dart VM in checked mode) can detect that assigning a non-string to
+you, your fellow programmers, and your tools can detect that assigning a non-string to
 the list is probably a mistake. Here’s an example:
 
 <?code-excerpt "misc/test/language_tour/generics_test.dart (why-generics)"?>
 {% prettify dart %}
 var names = new List<String>();
 names.addAll(['Seth', 'Kathy', 'Lars']);
-names.add(42); // Fails in checked mode (succeeds in production mode).
+names.add(42); // Error.
```
chalin: [nit] Drop the period:

-  // Error. --> // Error

(I believe that this change was only done in this file, not the original source.)

src/_guides/language/language-tour.md
```diff
@@ -3484,11 +3352,15 @@ void main() {
...
-  // Specifying any non-SomeBaseClass type results in a warning and, in
-  // checked mode, a runtime error.
+  // Specifying any non-SomeBaseClass type results in an error.
   // var objectFoo = new Foo<Object>();
 {% endprettify %}
 
+{% comment %}
+TODO: Confirm that non-SomeBaseClass is an error.
+Perhaps be more specific.
+{% endcomment %}
```
chalin: I confirm that it is an error: `'Object' doesn't extend 'SomeBaseClass'`. I'd suggest pulling `var objectFoo = new Foo<Object>();` out into it's own excerpt, uncommented, so that we can have the analyzer validate our claim that it is an error.

Would you like for me to make that change in this, or maybe a followup, PR?

If you'd like we can have the analyzer error automatically inlined in the prose (as we've done elsewhere in the docs), but I don't think that is necessary.

src/_guides/language/language-tour.md
```diff
@@ -3713,6 +3586,16 @@ Future checkVersion() [!async!] {
 }
 {% endprettify %}
 
+<aside class="alert alert-info" markdown="1">
+**Note:**
+Although an async function might perform time-consuming operations,
+it doesn't wait for those operations.
+Instead, the async function executes only until it encounters
+its first `await` expression.
+Then it returns a Future object,
```
chalin: [Style nit] Future --> `Future`

(I'm not sure that we are consistent with this convention. E.g., one of the changes you made early in this page was to use \<code> style for `String`, `int`, etc. but I don't think that we are systematic about it.)

src/_guides/language/language-tour.md
```diff
@@ -3713,6 +3586,16 @@ Future checkVersion() [!async!] {
 }
 {% endprettify %}
 
+<aside class="alert alert-info" markdown="1">
+**Note:**
+Although an async function might perform time-consuming operations,
+it doesn't wait for those operations.
+Instead, the async function executes only until it encounters
+its first `await` expression.
```
chalin: Good note. We should be able to refer the reader to a page with more details. How about using the link to https://github.com/dart-lang/sdk/blob/master/docs/newsletter/20170915.md#synchronous-async-start for now?

chalin:
FYI, I've committed a fix for analysis failure (adjusting expected analyzer output).

kwalrath:
Thanks for the review and the commit! I'll update this soon.

src/_guides/language/language-tour.md
```diff
@@ -48,13 +55,14 @@ apps:
...
-    [Comments](#comments).
+:   A way to make Dart ignore some content.
+    For more information, see [Comments](#comments).
 
 `num`
 
-:   A type. Some of the other built-in types are String, int, and bool.
+:   A type. A `num` variable's value can be either an `int` or a
+    `double` object.
+    Some of the other built-in types are `String`, `List`, and `bool`.
```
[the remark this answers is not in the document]
kwalrath: Yeah, I struggled with whether to change this. I wasn't sure why we used `num` in the first place.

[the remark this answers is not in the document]
kwalrath: I think that maybe we saw JavaScript programmers using no type when they should've used `num`. With type safety, that shouldn't be as much of an issue.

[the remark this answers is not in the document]
kwalrath: /cc @sethladd in case his memory is better.

src/_guides/language/language-tour.md
```diff
@@ -207,65 +191,60 @@ Here’s an example of creating a variable and assigning a value to it:
 
 <?code-excerpt "misc/lib/language_tour/variables.dart (var-decl)"?>
 {% prettify dart %}
-var name = 'Bob';
+var handle = 'Bob';
```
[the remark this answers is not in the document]
kwalrath: Yeah, I guess `name` is OK. It seemed weird to have a non-String name, but it's probably weirder to have a handle or reference.

src/_guides/language/language-tour.md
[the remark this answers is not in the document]
kwalrath: I'm renaming the function `printInteger` and changing the type to `int`.

src/_guides/language/language-tour.md
```diff
@@ -207,65 +191,60 @@ Here’s an example of creating a variable and assigning a value to it:
...
+The type of the `handle` variable is inferred to be String,
+but you can change that type by specifying it.
+If an object isn't restricted to a single type,
+specify the `Object` or `dynamic` type, following
+[design guidelines](/guides/language/effective-dart/design#do-annotate-with-object-instead-of-dynamic-to-indicate-any-object-is-accepted).
 
-Uninitialized variables have an initial value of `null`. Even variables
-with numeric types are initially null, because numbers are objects.
+{% comment %}
+**[PENDING: check!]**
```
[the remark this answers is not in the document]
kwalrath: I think I wanted to change on the Object vs. dynamic guidance, as it might change. I'll update the comment.

src/_guides/language/language-tour.md
kwalrath: Yes, if you could write that test, I'd appreciate it. My test fu isn't great. (I agree we don't need to inline the error.)

src/_guides/language/language-tour.md
kwalrath: We aren't consistent. At first I was trying to minimize the use of code font, especially in words that are already capitalized (this was the style at NeXT, but code font keeps creeping in.

src/_guides/language/language-tour.md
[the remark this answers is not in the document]
sethladd: works for me! :)

sethladd:
Thank you for updating the tour to Dart 2!!

Let us know how we can help advertise it when it's published :)

kwalrath:
ptal

src/_guides/language/language-tour.md
chalin: Working on the test now.

src/_guides/language/language-tour.md
```diff
@@ -2613,6 +2559,26 @@ Point.fromJson(Map<String, num> json)
 The right-hand side of an initializer does not have access to `this`.
 </div>
 
+During development, you can validate inputs by using `assert` in the
+initializer list.
+
+<?code-excerpt "misc/lib/language_tour/classes/extends.dart" replace="/extends|super/[!$&!]/g"?>
```
chalin: This was a superfluous `<?code-excerpt?>` instruction. I've removed it.

chalin:
Re. https://github.com/dart-lang/site-www/pull/677#discussion_r175220488: I've added the test; see the last commit. I've had to adjust the prose of that section. It now read as follows:
> screen shot 2018-03-16 at 19 37 44

kwalrath:
LGTM. Thanks, Patrice!

PR #680, opened 2018-03-17
Wrong boolean JS description fixed

ArsProgramma:
JS does treat values as true, not as false

**IMPORTANT!** Submit your PR against the right branch:
- `master` is for Dart 2-dev
- `1.x` is for Dart 1.x


ArsProgramma:
I signed it!

kwalrath:
This PR demonstrates that we should rephrase that sentence — the passive voice and double negative make it easy to misread.

Issue #684, opened 2018-03-19
Improve boolean true/false JS-vs-Dart description


kwalrath:
See PR #680. We should make the sentence active voice and clearer.

rsaxby:
Hello, are you still looking for help with this issue?

kwalrath:
Thanks for asking. We have a PR in flight (waiting for engineer feedback).

PR #686, opened 2018-03-19
Improved description of JS-vs-Dart boolean treatment. See #684

ArsProgramma:
Improved description of JS-vs-Dart boolean treatment. See Issue #684

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -544,11 +544,10 @@ To represent boolean values, Dart has a type named `bool`. Only two
 objects have type bool: the boolean literals `true` and `false`,
 which are both compile-time constants.
 
-When Dart expects a boolean value, only the value `true` is treated as
-true. All other values are treated as false. Unlike in JavaScript,
-values such as `1`, `"aString"`, and `someObject` are all treated as
-false.
+In JavaScript many values, such as `1`, `"aString"`, and `someObject` are treated as
+true, whereas values like null, 0 or empty string are treated as false.
 
+Dart in contrast treats every value except 'true' as false.
 Dart’s treatment of booleans is designed to avoid the strange behaviors
 that can arise when many values can be treated as true. What this means
 for you is that, instead of using code like


src/_guides/language/language-tour.md
```diff
@@ -544,11 +544,10 @@ To represent boolean values, Dart has a type named `bool`. Only two
 objects have type bool: the boolean literals `true` and `false`,
 which are both compile-time constants.
 
-When Dart expects a boolean value, only the value `true` is treated as
-true. All other values are treated as false. Unlike in JavaScript,
-values such as `1`, `"aString"`, and `someObject` are all treated as
-false.
+In JavaScript many values, such as `1`, `"aString"`, and `someObject` are treated as
+true, whereas values like null, 0 or empty string are treated as false.
```
kwalrath: I just realized that, with Dart 2, all of this should be unnecessary: you shouldn't be assigning non-boolean values to a boolean.

Let me check on this and get back to you.

kwalrath:
@leafpetersen should we remove all talk about converting non-booleans to booleans, or is that still a thing in Dart 2?

leafpetersen:
There are no boolean conversions in Dart 2 in the same sense as Dart 1, but there are implicit downcasts.  So in any place where you would have done boolean conversion in Dart 1, we
- require that the type is assignable to bool
- and if it is not actually bool insert an implicit cast to bool

so

```dart
Object reallyABool = true;
Object notABool = null;
// Implicit cast to bool, succeeds
if (reallyABool) print("Really a bool");
// Implicit cast to bool, fails
if(notABool) print("Shouldn't get here);
notABool = 0;
if (notABool) print("Still shouldn't get here);
```

kwalrath:
Just to clarify... When your code says "Shouldn't get here", does it not get there because it's an error, or because notABool is treated as false?

If the former, we should just remove the code that talks about assigning non-booleans to booleans. If the latter... I guess I'm surprised.

leafpetersen:
Right yes, sorry unclear example.  Nothing is ever converted from one object representation to a boolean.  Things which are not a boolean and are used in a boolean context result in a runtime error.

Better example, this code:

```dart
void test(Object b) {
  try {
    if (b) {
      print("$b is true");
    } else {
      print("$b is false");
    }
  } catch (e) {
    print("$b isn't a bool");
  }
}

void main() {
  test(true);
  test(false);
  test(null);
  test(new Object());
  test(3);
  test(0);
}
```

should print:

```
true is true
false is false
null isn't a bool
Instance of 'Object' isn't a bool
3 isn't a bool
0 isn't a bool
```

src/_guides/language/language-tour.md
kwalrath: It looks like we need to delete everything from "In JavaScript" up to "What this means for you is that, instead". Instead, it'll be something like:

> ...compile-time constants.
> 
> Dart's type safety means that you can't use code like `if (nonbooleanValue)` or `assert(nonbooleanValue)`. Instead, explicitly check for values, like this:
----------

@leafpetersen does that look correct?

leafpetersen: Looks reasonable to me.

kwalrath: @ArsProgramma would you like to make this change?

kwalrath:
Since we haven't heard from @ArsProgramma, I plan to make this fix myself. I'll close this PR when I start work.

Issue #632, opened 2018-03-07
Update Effective Dart to Dart 2


anders-sandholm:

mit-mit:
cc @munificent is this done?

kwalrath:
We're still waiting for Flutter to support whereType (#995), but I can remove the banner now. Everything else is done, I think.

kwalrath:
Banner removal is in #1138.

PR #688, opened 2018-03-19
Add banner to Effective Dart

kwalrath:
Added a banner like the following to all Effective Dart pages. These will be visible once we flip the site to Dart 2.

image

/cc @munificent


kwalrath:
Stopgap measure until #632 is fixed.

Issue #664, opened 2018-03-14
Review all tutorials for Dart 2


kwalrath:
Consider updating them or adding an "out of date" banner, for now.

- [x] [Language tutorials](https://www.dartlang.org/tutorials/language)
  - [x] https://www.dartlang.org/tutorials/language/futures - #512 suggests combining async/generator docs.
  - [x] https://www.dartlang.org/tutorials/language/streams - #512 suggests combining async/generator docs.
- [x] [Library tutorials](https://www.dartlang.org/tutorials/libraries)
  - [x] https://www.dartlang.org/tutorials/libraries/shared-pkgs - (OK for now: #692) 
- [x] [VM tutorials](https://www.dartlang.org/tutorials/dart-vm) (bannered: #691)
  - [x] https://www.dartlang.org/tutorials/dart-vm/cmdline
  - [x] https://www.dartlang.org/tutorials/dart-vm/get-started
  - [x] https://www.dartlang.org/tutorials/dart-vm/httpserver

chalin:
All pages mentioned in this issue have been updated; #512 can be addressed separately.

PR #691, opened 2018-03-19
Minimal changes to Dart VM tutorials

kwalrath:
Contributes to #664.

Staged at URLs like https://kw-www-dartlang-1.firebaseapp.com/tutorials/dart-vm/get-started.


src/_includes/tutorial-banner.html
```diff
@@ -0,0 +1,5 @@
+<div class="banner">
```
chalin: The blue banner is too easy to miss IMHO. Try:

> \<div class="banner alert alert-warning">

(This is what is used here: https://github.com/dart-lang/site-webdev/blob/5-dev/src/_includes/banner.html#L3.)

chalin: (Btw, this could be a .md file.)

PR #703, opened 2018-03-26
Remove pub build/serve & transformers

kwalrath:
@nex3 A couple of questions need answers before I can commit this PR (which will be in the Dart 2 version of www.dartlang.org):

* Does the `--packages-dir` flag still exist in Dart 2?
* Does `pub run` still have a `--mode` flag, and if so, how does that affect the [pub run docs](https://kw-www-dartlang-1.firebaseapp.com/tools/pub/cmd/pub-run)?

@filiph Are the Flutter notes in the pub and pub/cmd pages correct?

Staged in places like:

https://kw-www-dartlang-1.firebaseapp.com/tools/pub
https://kw-www-dartlang-1.firebaseapp.com/tools/pub/cmd
https://kw-www-dartlang-1.firebaseapp.com/tools/pub/cmd/pub-run
https://kw-www-dartlang-1.firebaseapp.com/tools/pub/glossary
https://kw-www-dartlang-1.firebaseapp.com/tools/pub/transformers


nex3:
> Does the `--packages-dir` flag still exist in Dart 2?

Yes, but it's hidden. I think it's safe to not mention it on the website.

> Does `pub run` still have a `--mode` flag, and if so, how does that affect the [pub run docs](https://kw-www-dartlang-1.firebaseapp.com/tools/pub/cmd/pub-run)?

The `--mode` flag for `pub run` will be removed when transformer support is removed.

kwalrath:
These changes are now ready for review, @filiph and @nex3.

firebase.json
```diff
@@ -1062,6 +1062,21 @@
         "destination": "/tools/faq#pub",
         "type": 301
       },
+      {
```
filiph: Would be better to redirect to something less confusing. Is there a source document that explains why we're done with transformers?

src/dart-2.md
```diff
@@ -1,65 +1,74 @@
...
+### Tools
+
+* Tools related to web development have changed.
+  For details, see the
+  [Dart 2 page for web developers.][webdev dart2]
+  * Pub no longer supports `build` and `serve`.
+  * Dartium is no longer supported.
+* Pub no longer supports transformers.
+  Instead, use builders. For more information, see
+  your platform documentation or the
```
filiph: Unclear what "your platform documentation" means in this context.

src/tools/pub/assets-and-transformers.md
```diff
@@ -1,397 +0,0 @@
----
```
filiph: As much as I love deleting code, do we really want to completely delete this? What about people who'll want to use Dart 1 for a few months longer? Can't we just paste a huge "This is Dart 1" banner over this?

src/tools/pub/cmd/index.md
```diff
@@ -81,56 +85,28 @@ or any modification of the pubspec.
...
 * The [`pub run`](/tools/pub/cmd/pub-run) command invokes a Dart script in your
   package, or in one of its dependencies.
 
 * The [`pub global`](/tools/pub/cmd/pub-global) command lets you work with
-  globally available packages that can be run when you are not currently inside
-  a package.
+  globally available packages.
 
-## Building and deploying apps and packages
+## Deploying packages and apps
```
filiph: Hmm, do we want to have an aside here that redirects people to the new way of _building_ packages? "The pub tool has historically been used for building as well. That is no longer true. Learn here how this changed."

I'm sure people will be coming here trying to learn about building because there are people (and articles) that will claim that `pub` is Dart's build tool for some time still.

src/tools/pub/index.md
```diff
@@ -1,14 +1,25 @@
 ---
-title: Pub Package and Asset Manager
-description: Use the pub tool to manage Dart's packages and assets.
+title: Pub Package Manager
```
filiph: Same here with adding an aside about the historical function of `pub` and where to find modern alternatives (like, I assume, `build_runner`).

src/tools/pub/transformers/aggregate.md
```diff
@@ -1,128 +0,0 @@
----
```
filiph: At some point, we might want to forward this to aggregate transformers in build_runner.
https://github.com/dart-lang/build/issues/926#issuecomment-371605646

But that's not ready yet.

I'm unsure that we want to hard-delete this. Aggregate transforming (aggregate building) is really useful. We're just at a point when one way is deprecated and the other way is not well documented yet.

src/tools/pub/transformers/index.md
```diff
@@ -1,326 +0,0 @@
----
-layout: default
```
filiph: Same here with deletion vs adding a big fat banner.

src/tools/pub/cmd/index.md
```diff
@@ -2,29 +2,33 @@
...
+and for deploying packages and command-line apps.
+
+<aside class="alert alert-warning"><div class="alert-with-image">
+  <img src="{% asset_path flutter-logo.png %}" alt="[Flutter]">
+  <div markdown="1">
+  **Important:**
+  The Flutter SDK has its own commands for managing and updating packages.
+  For more information, see
+  [Using Packages]({{site.flutter}}/using-packages/) and
+  [Upgrading Flutter]({{site.flutter}}/upgrading/)
```
filiph: I don't think you need the Upgrading Flutter link. The information how to upgrade is in the /using-packages link already.

filiph:
Oh, and the Flutter asides are correct, yes!

kwalrath:
@filiph I should've been clearer: These changes are only for the Dart 2 version of the site, which we will switch when Flutter switches to Dart 2. We'll keep a copy of the Dart 1.x version up at https://v1-dartlang-org.firebaseapp.com/ for an as-yet-undetermined length of time.

We'll have some UI for letting people switch between the sites. It might resemble what we have on webdev.dartlang.org (click AngularDart at the top, and you can switch to the other version).

firebase.json
kwalrath: We could possibly link to https://v1-dartlang-org.firebaseapp.com/tools/pub/assets-and-transformers and put a big note on there that transformers won't be supported in Dart 2.

There's also https://github.com/dart-lang/build/blob/master/build/README.md, but maybe the assets & transformers note should point there.

src/tools/pub/assets-and-transformers.md
kwalrath: I believe we want to delete as much as possible, to have a clean slate for Dart 2. Is that right, @kevmoo and @dgrove? We will, as I mentioned in another comment, keep https://v1-dartlang-org.firebaseapp.com/ (the 1.x fork of site-www) up for a while.

src/tools/pub/transformers/aggregate.md
kwalrath: Is this an issue if the v1 site is still available?

filiph:
Ohh, sorry. Yes, that makes sense. Can we point the redirects to the v1 site, then?

src/tools/pub/assets-and-transformers.md
kevmoo: I'm also nervous about just dropping this content – I wonder if having a place-holder page is better – which links to the v1 docs or the the v2 content?

firebase.json
kwalrath: I've created a new page, [/tools/pub/obsolete](https://kw-www-dartlang-1.firebaseapp.com/tools/pub/obsolete) that has a bit of explanation and a lot of links. What do you think?

kwalrath: (It doesn't, however, explain _why_ we dropped support for transformers. Do you think that's necessary?)

kevmoo: Awesome!

src/tools/pub/assets-and-transformers.md
kwalrath: I've created https://kw-www-dartlang-1.firebaseapp.com/tools/pub/obsolete to do this. What do you think?

kwalrath:
I was leery of pointing people directly to the v1 site (because they might not notice the site change), so I made the redirects be to https://kw-www-dartlang-1.firebaseapp.com/tools/pub/obsolete, which people can then use to get to the page they originally wanted.

kwalrath:
Ready for more feedback now.

kwalrath:
I've updated the config variables (@chalin) and think this might be ready to go. I've updated the staging server; you might need to use incognito mode to get the new redirects.

OK to merge?

kwalrath:
I just realized that the /tools/pub page has the same note as the /tools/pub/cmd page, and needs the same change. I'll do that now...

_config.yml
```diff
@@ -11,7 +11,9 @@ branch: master
 #   {{site.webdev}}/downloads
 
 dartlang: ""
+prev-dartlang: "https://v1-dartlang-org.firebaseapp.com"
```
chalin: If we are still going with the naming convention that we had agreed to, then this should be named `prev-url` (or at least we should have both names) so that we end up with:

```yaml
url: https://dartlang-org-dev.firebaseapp.com # this is from line 5 above
prev-url: https://v1-dartlang-org.firebaseapp.com
```

But as I mentioned, this isn't critical, it can be changed later as needed.

kwalrath: Oh, you mean literally `url`? I thought `url` was a stand-in for whatever we decided to call this current site. :)

kwalrath:
I think this is ready to go, so I'm going to merge. Please let me know if I need to improve anything.

PR #704, opened 2018-03-26
Add a couple of guidelines for using optional parameters.

munificent:


examples/misc/lib/effective_dart/usage_bad.dart
```diff
@@ -76,6 +76,26 @@ void miscDeclAnalyzedButNotTested() {
     // #enddocregion use-tear-off
   };
 
+  {
+    // #docregion default-separator
+    void insert(Object item, {int at: 0}) {
+      // ...
```
chalin: In other code excerpts where there are ellipses we use `/*...*/` in the code:

```dart
void insert(Object item, {int at = 0}) {
  /*...*/
}
```

When this code is extracted into the document, the `/*...*/` is replaced by `...`:

```dart
void insert(Object item, {int at = 0}) {
  ...
}
```

See line 55 above for an example.

examples/misc/lib/effective_dart/usage_bad.dart
```diff
@@ -76,6 +76,26 @@ void miscDeclAnalyzedButNotTested() {
...
+  {
+    // #docregion default-value-null
+    void error([String message = null]) {
+      if (message != null) {
+        stderr.write(message);
+      } else {
+        stderr.write("\n");
+      }
+    }
+    // #enddocregion default-value-null
```
chalin: Since the point of this example concerns parameters, how about making the body more compact?

```dart
void error([String message = null]) {
  stderr.write(message ?? '\n');
}
```

Also note the use of single quotes. Or, even more compact would be:

```dart
    void error([String message = null]) => //!<br>
        stderr.write(message ?? '\n');
```

(The special marker `//!<br>` gets removed. Its there to force dartfmt to insert a line break.)

examples/misc/lib/effective_dart/usage_good.dart
```diff
@@ -92,6 +92,26 @@ void miscDeclAnalyzedButNotTested() {
     // #enddocregion use-tear-off
   };
 
+  {
+    // #docregion default-separator
+    void insert(Object item, {int at = 0}) {
+      // ...
```
chalin: Suggestion: use `/*...*/`

examples/misc/lib/effective_dart/usage_good.dart
```diff
@@ -92,6 +92,26 @@ void miscDeclAnalyzedButNotTested() {
...
+
+  {
+    // #docregion default-value-null
+    void error([String message]) {
+      if (message != null) {
+        stderr.write(message);
+      } else {
+        stderr.write("\n");
+      }
+    }
```
chalin: To match the suggestion above:

```dart
    void error([String message]) =>
        stderr.write(message ?? '\n');
```

examples/misc/lib/effective_dart/usage_bad.dart
munificent: Done.

examples/misc/lib/effective_dart/usage_bad.dart
munificent: Great idea. Done.

examples/misc/lib/effective_dart/usage_good.dart
munificent: Done.

examples/misc/lib/effective_dart/usage_good.dart
munificent: Done. (But kept braced body. Feels a little weird to explicitly override the formatter in a document that tells you to trust the formatter... 😄)

chalin: Lol, no problem. Of course, we _are_ trusting the formatter. It's just that it can be tricky to balance code excerpt line length across many different and (mostly) unrelated code excerpts that _happen_ to be living in the same file. So we play with both the dartfmt line length argument and the occasional use of `//!<br>`. (All of this is FYI)

PR #705, opened 2018-03-26
Add a guideline for naming type parameters.

### LINKED ISSUE #30: Provide guidance for naming type parameters
_From @munificent on July 19, 2016 23:59_

Right now, the style guide is silent. I'm thinking maybe something like:

> DO name type parameters using UpperCamelCase.
> 
> CONSIDER using a single letter name, usually T, if its meaning is obvious. Otherwise, use a descriptive name.
> 
> CONSIDER using K and V for type parameters that represent a key and value, respectively.

_Copied from original issue: dart-lang/www.dartlang.org#1591_

munificent:
Fixes #30.

### DIFF
--- a/examples/misc/lib/effective_dart/design_good.dart
+++ b/examples/misc/lib/effective_dart/design_good.dart
@@ -190,6 +190,58 @@ class Task {
 
 //----------------------------------------------------------------------------
 
+// #docregion type-parameter-e
+class IterableBase1<E> {}
+
+class List1<E> {}
+
+class HashSet1<E> {}
+
+class RedBlackTree<E> {}
+// #enddocregion type-parameter-e
+
+// #docregion type-parameter-k-v
+class Map1<K, V> {}
+
+class Multimap<K, V> {}
+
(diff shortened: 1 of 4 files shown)


chalin:
The code excerpts should certainly be analyzed. As you'd expect, they can should be added to 

  `misc/lib/effective_dart/design_good.dart`

Add a numeric suffix to a class name to make it unique in the scope of the file. Such numeric suffixes get stripped out when the code excerpts are extracted. E.g.,

```dart
class IterableBase1<E> { /*...*/ }
class List1<E> { /*...*/ }
class HashSet1<E> { /*...*/ }
class RedBlackTree1<E> { /*...*/ }
```

To achieve ellipses `...` in the code excerpt, encode it as `ellipsis()`, like this:

```dart
class Future<T> {
  Future<S> then<S>(FutureOr<S> onValue(T value)) => ellipsis();
}
```

If you'd like, I can make the change to this PR, or (as Kathy suggested), we can open an issue and address it later. Let me know what you prefer.

PR #706, opened 2018-03-27
[1.x] Pub transformer warning

kwalrath:
1.x-only text to warn about the demise of pub build, pub serve, and pub support for transformers. Related to #703, which removes these docs from the Dart 2 version of the site.

This change will be visible on www.dartlang.org as soon as it's committed. 

Staged:
https://v1-dartlang-org.firebaseapp.com/tools/pub/assets-and-transformers
https://v1-dartlang-org.firebaseapp.com/tools/pub/transformers
https://v1-dartlang-org.firebaseapp.com/tools/pub#writing-transformers
https://v1-dartlang-org.firebaseapp.com/tools/pub/cmd#building-and-deploying-apps-and-packages
...
https://v1-dartlang-org.firebaseapp.com/tools/jetbrains-plugin#reporting-issues

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -10,7 +10,9 @@ branch: 1.x
 #   {{site.webdev}}/downloads
 
 dartlang: ""
+dartlang_2: "https://dartlang-org-dev.firebaseapp.com"
 webdev: "https://webdev.dartlang.org"
+webdev_dev: "https://webdev-dartlang-org-dev.firebaseapp.com"
 dart_vm:  "/dart-vm"
 flutter:  "https://flutter.io"
 dart_api: "https://api.dartlang.org"
--- a/_config.yml
+++ b/_config.yml
@@ -10,9 +10,9 @@ branch: 1.x
 #   {{site.webdev}}/downloads
 
 dartlang: ""
-dartlang_2: "https://dartlang-org-dev.firebaseapp.com"
+dev-dartlang: "https://dartlang-org-dev.firebaseapp.com"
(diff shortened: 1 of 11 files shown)


_config.yml
```diff
@@ -10,7 +10,9 @@ branch: 1.x
 #   {{site.webdev}}/downloads
 
 dartlang: ""
+dartlang_2: "https://dartlang-org-dev.firebaseapp.com"
```
chalin: FYI, I've been using the following YAML names in the webdev config:

- `url` is the main URL to this site. (It's defined on line 4 above.)
- `dev-url` for the URL to the dev version of the site.
- `prev-url` for the URL to the previous version of the site (which will make sense only once Dart 2 lands on master).

Of course, we don't have to stick to these names, but I thought that I'd mention it.

_config.yml
```diff
@@ -10,7 +10,9 @@ branch: 1.x
...
 webdev: "https://webdev.dartlang.org"
+webdev_dev: "https://webdev-dartlang-org-dev.firebaseapp.com"
```
chalin: Maybe `dev-webdev` (to follow the field naming contentions mentioned above, which use a hyphen rather than an underscore).

_config.yml
kwalrath: I wanted to avoid renaming this once 1.x is no longer the default site. Maybe dartlang-2?

kwalrath: I've done this rename, just in case.

filiph:
LGTM!

PR #711, opened 2018-03-28
Clarify single-line if rule.

munificent:
This comes up a lot so I think it's worth being extra explicit.

### DIFF
--- a/examples/misc/lib/effective_dart/style_bad.dart
+++ b/examples/misc/lib/effective_dart/style_bad.dart
@@ -10,3 +10,10 @@ class Dice {
   static final NUMBER_GENERATOR = new Random();
 }
 // #enddocregion const-names
+
+oneLineIf(dynamic overflowChars, dynamic other) {
+  // #docregion one-line-if-wrap
+  if (overflowChars != other.overflowChars)
+    return overflowChars < other.overflowChars;
+  // #enddocregion one-line-if-wrap
+};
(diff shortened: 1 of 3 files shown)


examples/misc/lib/effective_dart/style_bad.dart
```diff
@@ -10,3 +10,10 @@ class Dice {
   static final NUMBER_GENERATOR = new Random();
 }
 // #enddocregion const-names
+
+oneLineIf(dynamic overflowChars, dynamic other) {
```
chalin: I usually don't name the function, given that the `docregion` name is what is used.

examples/misc/lib/effective_dart/style_bad.dart
```diff
@@ -10,3 +10,10 @@ class Dice {
...
+  // #docregion one-line-if-wrap
+  if (overflowChars != other.overflowChars)
+    return overflowChars < other.overflowChars;
+  // #enddocregion one-line-if-wrap
+};
```
chalin: I think that the formatter is going to want a newline here ... if not, I'd suggest adding one :)

examples/misc/lib/effective_dart/style_bad.dart
munificent: This is a top-level function, so it needs a name.

examples/misc/lib/effective_dart/style_bad.dart
munificent: It's actually a straight-up syntax error since this is at the top level. Oops! Fixed.

PR #718, opened 2018-03-29
Add migration process to /dart-2

kwalrath:
Also rejiggered much of the page.
All this was spurred by the need to link to the dart2_fix tool.

Staged:
https://kw-staging-dartlang-2.firebaseapp.com/dart-2

### DIFF
--- a/src/dart-2.md
+++ b/src/dart-2.md
@@ -5,23 +5,41 @@ title: Dart 2 Migration
 description: How Dart 2 is different from Dart 1.x, and how you can convert your code to work with Dart 2.
 ---
 
-**This page is under construction.**
-It aims to help developers migrating from Dart 1.x to Dart 2.
+Here's the general process for migrating to Dart 2,
+from either Dart 1.x or an earlier version of Dart 2:
+
+1. Get an up-to-date version of the Flutter or Dart SDK.
+   * [Flutter SDK instructions][Flutter SDK install]
+   * [Dart SDK instructions][Dart SDK install] (VM or web)
+2. Upgrade the packages your app depends on.
+   * Flutter: [`flutter packages upgrade`][flutter package upgrade]
+   * Dart VM or web: [`pub upgrade`][pub upgrade]
+3. Run the [dart2_fix tool.][dart2_fix]
+4. Run the analyzer to find [compile-time errors][Fixing Common Type Problems].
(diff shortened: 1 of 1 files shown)


src/dart-2.md
```diff
@@ -5,23 +5,41 @@ title: Dart 2 Migration
...
+Here's the general process for migrating to Dart 2,
+from either Dart 1.x or an earlier version of Dart 2:
 
+1. Get an up-to-date version of the Flutter or Dart SDK.
+   * [Flutter SDK instructions][Flutter SDK install]
+   * [Dart SDK instructions][Dart SDK install] (VM or web)
+2. Upgrade the packages your app depends on.
+   * Flutter: [`flutter packages upgrade`][flutter package upgrade]
+   * Dart VM or web: [`pub upgrade`][pub upgrade]
+3. Run the [dart2_fix tool.][dart2_fix]
```
devoncarew: nit: the period should be outside of the link

devoncarew: and I see this in two links further down the page, so you may have a different opinion about where to put the periods :)

src/dart-2.md
```diff
@@ -5,23 +5,41 @@ title: Dart 2 Migration
...
+{% comment %}
+TODO:
+- talk about library vs. app migration?
+- talk more about versioning?
+{% endcomment %}
+
+
+## Differences
 
 ### Language and libraries
```
chalin: I missed this earlier: from a few lines below ...

    * [Email: Breaking Change: --preview-dart-2 turned on by default][Leaf's email]

The `--` gets converted to an m-dash when the markdown is process. Maybe code-format it?

* [Email: Breaking Change: `--preview-dart-2` turned on by default][Leaf's email]

[Leaf's email]: #

src/dart-2.md
kwalrath: Normally, I'd agree with you, but we have little icons for external links that look bad when they come before a period or comma.

src/dart-2.md
kwalrath: Oh! I hadn't noticed this. I'll fix that.

src/dart-2.md
chalin: (I've been considering writing a plugin to handle that.)

PR #719, opened 2018-03-30
Fix build

kwalrath:
The example code needed to be updated to reflect the code in the text. I think I restored what @munificent intended (plus made a couple of small changes, just because).

Staged at https://kw-www-dartlang-1.firebaseapp.com/guides/language/effective-dart/documentation#markdown.

### DIFF
--- a/src/_guides/language/effective-dart/documentation.md
+++ b/src/_guides/language/effective-dart/documentation.md
@@ -354,10 +354,10 @@ a flavor of what's supported:
 {% prettify dart %}
 /// This is a paragraph of regular text.
 ///
-/// This sentence has *two* _emphasized_ words (i.e. italics) and **two**
+/// This sentence has *two* _emphasized_ words (italics) and **two**
 /// __strong__ ones (bold).
 ///
-/// A blank line creates another separate paragraph. It has some `inline code`
+/// A blank line creates a separate paragraph. It has some `inline code`
 /// delimited using backticks.
 ///
 /// * Unordered lists.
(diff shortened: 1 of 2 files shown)


kwalrath:
For some reason, refresh-code-excerpts.sh wants to re-insert part of the code. (If I delete the code and run it, it looks right, but if I run it again, it starts inserting again right after the last header, starting with the triple back ticks.) Help, @chalin!

kwalrath:
It looks like the triple ticks are causing the problem, so I've removed them for now.

kwalrath:
I'm going to merge this, so that the build works again. Let me know if I should change anything.

src/_guides/language/effective-dart/documentation.md
```diff
@@ -372,21 +372,6 @@ a flavor of what's supported:
...
-///     .retain(its, formatting);
-/// ```
-///
-/// The code language (for syntax highlighting) defaults to Dart. You can
-/// specify it by putting the name of the language after the opening backticks:
-///
-/// ```html
-/// <h1>HTML is magical!</h1>
-/// ```
-///
```
munificent: It would be good not to lose these examples. Now that @chalin fixed the excerpt generator, can you put these back in?

kwalrath: He added them back in his PR. It turned out to be an issue with the closing tag.

munificent: 👍

PR #723, opened 2018-03-30
Remove whitespace guidelines. Dartfmt is the style guide.

### LINKED ISSUE #717: Remove detailed guidelines on whitespace and indentation?
Now that we have dartfmt, I don't think it's useful to write down a bunch of detailed explicit rules on whitespace. Especially given that in some cases, dartfmt's rules are too subtle to easily capture in prose.

The style guide already says, "DO use dartfmt to format your code." What do you think about removing all of the other formatting guidelines?

munificent:
Fix #717.


chalin:
Probably because of the cache. I've cleared all caches, hopefully that will solve the problem. Otherwise I'll investigate further.

Issue #654, opened 2018-03-13
Switch to Dart 2 site; link to Dart 1.x site


kwalrath:
Once Flutter switches to Dart 2, so will www.dartlang.org. The Dart 1.x version of the site will be temporarily available at https://v1-dartlang-org.firebaseapp.com.

---

Edit: @chalin added itemized list. Contributes to #637.

- [x] Adjust URL config values for `url`, `dev-url`, `prev-url`
  - [x] On master - #772
  - [x] On 1.x - #770
- [x] Add site-switching dropdown menu to main nav
  - [x] On master - #724
  - [x] On 1.x - #753
- [x] Add a banner to every page, similar to what we have for webdev. See @anders-sandholm request in https://github.com/dart-lang/site-www/pull/753#issuecomment-379233184
  - [x] On master - #765
  - [x] On 1.x - #764

kwalrath:
We'll probably want an easy way of switching between the two sites until Dart 2 is on the stable channel.

chalin:
Would you like a dropdown line on webdev?

kwalrath:
> Would you like a dropdown (under "Language") like on webdev?

That's probably the easiest way to go. Of course, libraries are also versioned, but most of the site changes are language related. @kevmoo & @filiph what do you say? See https://webdev.dartlang.org for an example of what Patrice is proposing.

chalin:
Switch complete! \o/

PR #724, opened 2018-03-30
Dart language-version menu for switching between site versions

chalin:
Addresses https://github.com/dart-lang/site-www/issues/654#issuecomment-376715048.

Feature allowing users to switch back and forth between Dart 1 and 2 versions of dartlang.org. The language selection menu looks like this:

> screen shot 2018-03-30 at 15 05 03

This PR adds the menu to master. A followup PR will add it to the 1.x branch.


chalin:
@kwalrath - URL updated and staging server refreshed.

chalin:
Any more feedback on this?

kevmoo:
Wondering if the header should be hint that the site is in "preview" mode...or something.

The check mark is weird. So this is the "state" we're in – but also a link to the guide.

I think this is improvement, though...

chalin:
> Wondering if the header should be hint that the site is in "preview" mode...or something.

It is my understanding that we won't officially be using the master branch version of the site until we switch dartlang.org over to Dart 2. At that point the menus will become:

- 1.24.3 (archive)
- 2.0.0 (stable)

> The check mark is weird. So this is the "state" we're in – but also a link to the guide.

Agreed about the weirdness of the link, but ... it is also my understanding that this menu is meant as a temporary measure, to be used for a few weeks? So this design is trying to be non-disruptive (e.g., not create new main nav entries).

I suppose that an alternative _might_ be to put the drop down under the "Dart". WDYT?

kwalrath:
@chalin could you please mock up having the pulldown under the icon?

chalin:
The logo links to the site landing page, so having a dropdown menu associated with the logo will suffer the same link weirdness issues.

As an alternative, I've prototyped having a _Version_ dropdown in the main nav bar:

> screen shot 2018-04-04 at 15 47 01

Staging server refreshed: https://dartlang-org-staging-0.firebaseapp.com

How does this fair against the previous prototype @kevmoo, et all?

Note:

- The _Version_ dropdown replaces the _Community_ link (we were running out of horizontal real estate).
- The _Version_ dropdown isn't visible on narrow displays.

mit-mit:
With Dart 2 becoming default in Flutter beta 2, I think "preview" is understatement. Can we use "beta" instead?

chalin:
Updated to use "beta". Staging server refreshed: https://dartlang-org-staging-0.firebaseapp.com

chalin:
> But the toggle disappears after selecting 1.24.3.

Right, selecting 1.24.3 redirects you to the v1 site, which doesn't yet have a Version dropdown. Once this PR is approved and merged, a similar change will be made to the v1 site -- then readers will be able to navigate back and forth between the v1 and v2 sites.

chalin:
The **main nav isn't visible on narrow displays**, so we'd lose the Version dropdown in those cases. I've added a Version menu to the sidenav. It looks like this when closed:

> screen shot 2018-04-05 at 09 22 01

After clicking the "dropdown" arrow, it looks like this:

> screen shot 2018-04-05 at 09 22 22

WDYT?

chalin:
Btw, the sidenav is always accessible. If we'd like to make switching site versions less obvious, we could remove it from the main nav.

chalin:
@kevmoo - the reason I prototyped the Version dropdown was because of the [link weirdness](https://github.com/dart-lang/site-www/pull/724#issuecomment-378374313) that you pointed out in the Language dropdown. It occurred to me that issue with the link could be fixed by adding an explicit link to the Language Overview page:

> screen shot 2018-04-05 at 10 35 14

This is also an opportunity to add a link to the Migration page too (as shown above). 
WDYT?

kevmoo:
Would it be better to just have a top-level version dropdown?

kevmoo:
Looking at https://v1-dartlang-org.firebaseapp.com/ – I like it!

Now just change the header "Version" to "Dart 2 (beta)"

mit-mit:
I like it too. And I would simply not bother with adding a toggle option in narrow browsers. Hopefully a fairly small amount of developers will need this toggle.

chalin:
Ok, thanks for the feedback.

@kevmoo - by changing "Version" to "Dart 2 (beta)" we'll have two "Dart" items in the main menu (the logo on the left being the other). Are you ok with that? Also, doesn't it seem strange to have a "Dart 2 (beta)" dropdown title with a menu entry "1.24.3 (stable)"?

kevmoo:
I think that's fine.

kwalrath:
@kevmoo, do you mean you were looking at https://dartlang-org-staging-0.firebaseapp.com/?

chalin:
After a little fine tuning with @kevmoo and @kwalrath, we've agreed on a "Dart 2" dropdown (in the main nav only), with a Migration menu item. Please see the [staging server](https://dartlang-org-staging-0.firebaseapp.com/).

PR #726, opened 2018-03-30
chore: add firebase dartlang-org-staging-1 project

chalin:


.firebaserc
```diff
@@ -3,6 +3,7 @@
     "default": "www-dartlang-org",
     "dev": "dartlang-org-dev",
     "staging-0": "dartlang-org-staging-0",
+    "staging-1": "dartlang-org-staging-1",
```
kwalrath: When should we use staging-0 & staging-1?

chalin: I had originally planned on using them as part of an auto-PR deploy scheme, but that will have to wait. So, use them whenever they are free :)

Issue #657, opened 2018-03-13
Move/copy webdev installation instructions to site-webdev


kwalrath:
Currently, developers have to bounce back and forth between webdev.dartlang.org and www.dartlang.org to get started. Instead, they should be able to stay on webdev, even if that means we duplicate content. Implementing this will be easier if we get infrastructure that lets us share files between sites. Once these instructions are VM only, we'll be able to simplify them.

/cc @filiph @anders-sandholm

kwalrath:
We'd need to copy most/all of the following pages:

https://www.dartlang.org/install
https://www.dartlang.org/install/archive
https://www.dartlang.org/install/linux
https://www.dartlang.org/install/mac
https://www.dartlang.org/install/windows

chalin:
Ok, I'll explore options tomorrow.

anders-sandholm:
I'd vote in favor of just copying contents rather than sharing for two reasons. 

1) It allows us to tailor the descriptions/wording to the context / use case at hand (cmd-line or web) although the actual actions / technical steps may be exactly the same. 

2) We may - over time - end up having separate SDKs for core and web.

chalin:
This issue was moved to dart-lang/site-webdev#1471

PR #727, opened 2018-03-30
Refactor Install page before moving a copy to webdev

chalin:
Contributes to #657.

Staged, see:

- https://dartlang-org-staging-1.firebaseapp.com/install
- https://dartlang-org-staging-1.firebaseapp.com/tools/sdk#install

---

_Install_ page is now like the _Platoforms_, but it directs the user to installation instructions appropriate for the chosen platform:

> screen shot 2018-03-30 at 16 13 37

---

The Web and VM installation instructions now make use of @filiph's compact OS-tabbed instructions (slightly enhanced with a bit of extra information):

> screen shot 2018-03-30 at 16 16 05

cc @JekCharlsonYu @anders-sandholm


src/tools/sdk/_linux.md
```diff
@@ -0,0 +1,41 @@
+#### Install using apt-get
```
kwalrath: We need some intro text before this, or else it isn't clear that you have a choice between apt-get, debian, and compiling from source. (It looks like the text is there in https://www.dartlang.org/install/linux, so it should be easy to grab. I guess formatting is the only issue. We lose a lot of navigation when we give up the page and its TOC.)

src/tools/sdk/index.md
```diff
@@ -0,0 +1,131 @@
...
+js:
+- url: /tools/sdk/archive/assets/install.js
+  defer: true
+---
+
+The Dart SDK has the libraries and command-line tools that you need to develop
+Dart apps. To learn about other tools you can use for Dart development, see
+[Dart Tools](/tools).
+
+The [DartPad][]{: target="_blank"} is a great way to learn how to write a simple
```
kwalrath: This paragraph (or maybe something more like the original)  belongs on the main Install page, not here.

src/tools/sdk/_windows.md
```diff
@@ -0,0 +1,33 @@
+#### Install using a setup wizard
+{: style="margin-top: 0"}
+
```
kwalrath: As for the Linux page, we need an overview here.

src/tools/sdk/_windows.md
```diff
@@ -0,0 +1,33 @@
+#### Install using a setup wizard
```
kwalrath: The style of these 4th-level headers looks strange to me under the smaller font of the tabs. Let's see if it looks better once we add introductory text.

src/tools/sdk/index.md
```diff
@@ -0,0 +1,131 @@
...
+Dart SDK.
+
+This site's documentation and examples use the
+[{{site.data.pkg-vers.SDK.vers}}][site SDK version]{:.no-automatic-external}
+release of the **Dart SDK**.
+
+## Install the SDK
+
+A package manager can help you easily install and update the Dart SDK.
+
```
kwalrath: Let's put the archive link here:

Alternatively, you can download zip files from the [Index of Downloads].

src/tools/sdk/index.md
```diff
@@ -0,0 +1,131 @@
...
+{% include_relative _windows.md %}
+</div>
+<div id="tab-sdk-install-linux" class="tabs__content" markdown="1">
+{% include_relative _linux.md %}
+</div>
+<div id="tab-sdk-install-mac" class="tabs__content" markdown="1">
+{% include_relative _mac.md %}
+</div>
+
+Alternatives:
```
kwalrath: Let's delete this paragraph and the bulleted list that follows.

src/tools/sdk/archive/index.md
```diff
@@ -3,9 +3,9 @@ title: "Index of Downloads"
 description: "Download specific stable and dev channel versions of the Dart SDK, Dartium, and the Dart API documentation."
 permalink: /install/archive
```
kwalrath: Update the permalink to /tools/sdk/archive, and (wherever the breadcrumbs are defined) make the previous page be Dart SDK, not Install Dart.

src/install.md
```diff
@@ -0,0 +1,40 @@
+---
+title: Install Dart
+description: The bundles that support the Dart language.
+permalink: /install
+toc: false
+---
+
+Choose your platform.
+
+{:.card-grid}
```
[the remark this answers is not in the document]
chalin: I believe that this is the same layout and spacing as the platforms page, but I'll double check.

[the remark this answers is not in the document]
chalin: I confirm that we're using the same `card-grid` and `card` styles in other pages, and that the cards have the same width, etc. E.g., see the cards on these pages:

- https://www.dartlang.org/guides/language
- https://www.dartlang.org/guides/libraries
- https://www.dartlang.org/guides/platforms

I don't suggest that we start making one-off changes. Later we might want to consider a site-wide revision of styling.

WDYT?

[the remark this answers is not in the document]
chalin: Given that the "platform" cards are mostly white, I suppose that I could make those narrower (on both the platform and install pages). Let me try something ... .

src/tools/sdk/index.md
chalin: I've added both the archive and "build from source" links here.

src/tools/sdk/archive/index.md
chalin: I've renamed this page to "Dart SDK Download", with a short title of just "Download", so that we get this heading w/ breadcrumbs:

> screen shot 2018-04-02 at 15 48 41

kwalrath: We have to be careful about how we refer to these pages. This title might be good, but we don't want to imply that it's the preferred way to download the SDK.

chalin:
Changes complete and staging server refreshed.

src/tools/sdk/archive/index.md
```diff
@@ -1,13 +1,14 @@
...
+description: Download specific stable and dev channel versions of the Dart SDK and the Dart API documentation.
 js:
-- url: /install/archive/out/web/download_archive.dart.js
+- url: /tools/sdk/archive/out/web/download_archive.dart.js
   defer: true
-- url: /install/archive/assets/install.js
+- url: /tools/sdk/archive/assets/install.js
   defer: true
 ---
+
```
kwalrath: Later this page refers to the "main Dart installation page". Let's just call it the "Dart SDK" page. We should also check for other references to that page.

src/tools/sdk/archive/index.md
```diff
@@ -1,13 +1,14 @@
 ---
-title: "Index of Downloads"
-description: "Download specific stable and dev channel versions of the Dart SDK, Dartium, and the Dart API documentation."
-permalink: /install/archive
+title: Dart SDK Download
+short-title: Download
```
kwalrath: Download -> Zip Files (in the title & short-title)
... or something else indicating that this isn't the primary download page.

src/tools/sdk/index.md
```diff
@@ -0,0 +1,122 @@
...
+[{{site.data.pkg-vers.SDK.vers}}][site SDK version]{:.no-automatic-external}
+release of the **Dart SDK**.
+
+## Install the SDK
+
+A package manager can help you easily install and update the Dart SDK.
+Don’t want to use a package manager? Other options are:
+
+* [Download the SDK manually](sdk/archive)
+* [Build the SDK from source][]
```
kwalrath: I'm afraid this bulleted list might tempt people away from package managers. How about:

...update the Dart SDK. Other options are [building the SDK from source] and [downloading the SDK as a zip file].

That way the download link is still easy to find (last thing in the paragraph), but not quite as tempting.

src/install.md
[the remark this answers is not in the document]
kwalrath: It looks better, thanks.

src/install.md
```diff
@@ -5,30 +5,38 @@ permalink: /install
 toc: false
 ---
 
-Choose your platform.
+The [DartPad][]{: target="_blank"} is a great way to learn how to write a simple
+app in the Dart language, once you are ready to develop real apps, you need the
+Dart SDK.
+
+Install the Dart SDK for your platform.
```
kwalrath: Dart SDK -> SDK

src/install.md
```diff
@@ -5,30 +5,38 @@ permalink: /install
 toc: false
 ---
 
-Choose your platform.
+The [DartPad][]{: target="_blank"} is a great way to learn how to write a simple
```
kwalrath: The -> Although

src/install.md
```diff
@@ -5,30 +5,38 @@ permalink: /install
...
+app in the Dart language, once you are ready to develop real apps, you need the
+Dart SDK.
```
kwalrath: the Dart SDK -> an SDK

src/tools/sdk/index.md
```diff
@@ -38,15 +38,6 @@ A package manager can help you easily install and update the Dart SDK.
...
-- Not using a package manager for your OS? No problem!
-  [Download](/install/archive)
-  zip files of the Dart SDK and docs.
-- Looking for an older version?
-  Check out our [zip file archive](/install/archive) for
-  previous versions of the Dart SDK.
-
 ## What's in the SDK
 
 The SDK directory includes a `lib` directory for the
```
kwalrath: SDK directory -> Dart SDK

It seems weird to talk about an SDK directory, especially when some package managers put stuff all over the place. E.g. my copy of dart is in ~/homebrew/bin, which links to ../Cellar/dart/2.0.0-dev.43.0/bin/dart, which has `bin` and `libexec` directories (the latter of which contains `bin` and `lib` directories).  I considered dropping the mentions of `bin` and `lib`, but on second thought they're kind of handy.

kwalrath: In general, many occurrences of "SDK" should perhaps be "Dart SDK". Some occurrences of "SDK" could perhaps be deleted ( e.g. "About release ~SDK~ channels and version strings").

src/tools/sdk/archive/index.md
```diff
@@ -40,7 +41,7 @@ production use.
 
 {% include_relative _archives_table.html channel="dev" %}
 
-## Direct download URLs
+## Download URLs
```
kwalrath: We should update the example URLs, either now or in a future PR.

src/install.md
```diff
@@ -5,30 +5,38 @@ permalink: /install
 toc: false
 ---
 
-Choose your platform.
+The [DartPad][]{: target="_blank"} is a great way to learn how to write a simple
+app in the Dart language, once you are ready to develop real apps, you need the
```
kwalrath: delete "in the Dart language"
you are -> you're

src/tools/sdk/archive/index.md
chalin: How about _Archive_?

src/tools/sdk/index.md
chalin: Good suggestion.

src/tools/sdk/archive/index.md
chalin: I've made some updates to the example URLs. We can make more once 2.0 final comes out.

src/tools/sdk/archive/index.md
chalin: > We should also check for other references to that page.

I've swept the docs and made the necessary changes.

chalin:
Thank you for the very good suggestions. I've made the necessary changes and updated the staging server.

firebase.json
```diff
@@ -1171,7 +1171,9 @@
         "source": "/dart-2.0",
         "destination": "/dart-2",
         "type": 301
-      }
+      },
+      { "source": "/install/**", "destination": "/install", "type": 301 },
+      { "source": "/tools/pub/**", "destination": "/toosl/pub", "type": 301 }
```
kwalrath: toosl -> tools
(but why add this redirect? it breaks all the URLs under pub, e.g. https://dartlang-org-staging-1.firebaseapp.com/tools/pub/cmd/pub-get)

src/_data/side-nav.yml
```diff
@@ -39,8 +39,6 @@
     - title: "Dart VM"
       permalink: /dart-vm
       children:
-      - title: "Install Dart"
```
kwalrath: Where will people find this page?

src/tools/pub/installing.md
```diff
@@ -1,49 +0,0 @@
----
```
kwalrath: Why are we getting rid of this page?

src/_includes/configure-path.html
```diff
@@ -1,21 +0,0 @@
-<div markdown="1">
```
kwalrath: Do we cover setting the path anywhere?

src/dart-vm/tools/index.md
```diff
@@ -1,12 +1,10 @@
...
-title: "Dart VM Tools"
-description: "The tools that you get when you download the Dart SDK."
-permalink: /dart-vm/tools
+title: Dart VM Tools
+description: The tools that you get when you download the Dart SDK.
 toc: false
 ---
 
-When you [install](/install) the
+When you [install](/tools/sdk#install) the
```
kwalrath: Let's remove the link on "install".

src/_data/side-nav.yml
chalin: We have a **top-level sidenav entry named _Install_** under RESOURCES. Why should _Dart VM_ be singled out to have an install submenu item? If we have one for Dart VM, for sake of uniformity, we might as well have one for Web and for Mobile. Instead, I think that folks should be encouraged to go to the main install page in all cases. WDYT?

src/tools/pub/installing.md
chalin: Because we don't _install_ pub. One installs the Dart SDK; the main pub page opens with:

> You can use the `pub` tool to manage Dart packages. Pub is one of the tools that you get when you [install](/tools/sdk#install) the Dart SDK.

In think that that is all we need.

The rest of the pub "installing" page was actually only about environment variable configuration. Which is what I renamed to page to.

src/_includes/configure-path.html
chalin: Whether / how one sets the path or not depends on (1) the technique used to install the SDK, (2) how the SDK is used and in which environment / shell it is used from.

For example, when using Homebrew under MacOS, one **shouldn't** adjust the path specifically for the Dart SDK. Doing so can lead to problem later on when Dart is upgraded and/or versions are switched.

Our previous instructions on how to set the path were incomplete at best, misleading in some cases.

firebase.json
chalin: Oops, clearly not the desired effect. Changed to / was meant to be:

```
      { "source": "/tools/pub/installing", "destination": "/tools/pub/environment-variables", "type": 301 }
```

src/_data/side-nav.yml
kwalrath: Ah, I thought you were deleting the one under RESOURCES.

src/_includes/configure-path.html
kwalrath: Let's just mention adding the SDK's bin directory to your path, and leave it up to people to find the instructions. (We added the path info because some beginners would get confused when they installed Dart but couldn't run the commands.)

chalin: Where do you want this mentioned?

chalin: (I'd suggest leaving any such mention for now and we can consider re-adding it, or directing folks to SO ... as the need arises. WDYT?)

firebase.json
```diff
@@ -1171,7 +1171,9 @@
         "source": "/dart-2.0",
         "destination": "/dart-2",
         "type": 301
-      }
+      },
+      { "source": "/install/**", "destination": "/install", "type": 301 },
```
kwalrath: I'm torn about whether to set up a separate redirect for the archive (/install/archive -> /tools/sdk/archive). I suspect that lots of people have links to the archive page, so it seems reasonable to take them to the page they want.

The only reason not to redirect /install/archive to /tools/sdk/archive is if we really DO want to force people to think — to look at the 3 SDK options and then hunt in the Dart SDK page for the archive.

I'm leaning toward redirecting /install/archive to /tools/sdk/archive, to avoid irritating the type of expert user who tends to use the archive.

src/tools/sdk/index.md
```diff
@@ -0,0 +1,120 @@
...
+This site's documentation and examples use the
+[{{site.data.pkg-vers.SDK.vers}}][site SDK version]{:.no-automatic-external}
+release of the **Dart SDK**.
+
+## Install the SDK
+
+A package manager can help you easily install and update the Dart SDK.
+Don’t want to use a package manager? Other options are
+[building the SDK from source][] and
+[downloading the SDK as a zip file](sdk/archive).
```
kwalrath: sdk/archive -> /tools/sdk/archive (or archive)?

I somehow ended up with the path .../tools/sdk/, and then this took me to .../tools/sdk/sdk/archive, which was a 404.

I wonder (a) why it's possible to have the `/` at the end of the URL (there's a bug about this somewhere) and (b) how many similar possible 404s exist in this site.

src/_includes/configure-path.html
kwalrath: I think we should mention adding the SDK's tools to your path in the /tools/sdk page, somewhere inside the "Install the SDK" section.

firebase.json
chalin: > leaning toward redirecting /install/archive to /tools/sdk/archive

That sounds reasonable.

chalin: FYI, we'll have both redirects. One for the archive link, and the other to catch the OS specific pages.

src/tools/sdk/index.md
chalin: Stange. I'm not seeing that. I've change the link to /tools/sdk/archive just in case.

src/_includes/configure-path.html
chalin: Oops, I didn't get the notice about this comment until now. I'll add it with my next PR.

src/tools/sdk/index.md
kwalrath: Ah, the related issue is on site-webdev: https://github.com/dart-lang/site-webdev/issues/1418

PR #728, opened 2018-03-30
Redo guidelines around getters and setters.

### LINKED ISSUE #298: Clarify guideline for when to define a getter
It's not clear if the bullet list for [the guideline](https://www.dartlang.org/guides/language/effective-dart/design#do-use-getters-for-operations-that-conceptually-access-properties) is meant to completely define which members are "conceptual properties" or whether it only comes into play *after* the user has decided the member is one.

The intent was to be the former, but the text doesn't state that, and I don't think the bullet list is complete enough for that to be true. In particular, according to the rules, `toString()` should be a getter. We should probably add another bullet or two stating that things like conversions or members where you want the reader to focus on the work done to produce the value more than the value itself can still be methods.

### LINKED ISSUE #569: in effective dart, advice on getters is a little vague
the [effective dart style guide](https://www.dartlang.org/guides/language/effective-dart/design#do-use-getters-for-operations-that-conceptually-access-properties) says
```
You should define a getter instead of a method when all of these are true:
- Does not take any arguments.
- Returns a value.
- Is side-effect free. Invoking a getter shouldn't change any externally-visible
  state (caching internally or lazy initialization is OK). Invoking the same
  getter repeatedly should return the same value unless the object is explicitly
  changed between calls.
```

the last point about "same value" needs some clarification.  in particular,
```dart
class MutableButDeepEquals {
  int value = 0;

  @override
  bool operator ==(other) =>
      other is MutableButDeepEquals && value == other.value;
}

class A {
  get getter => new MutableButDeepEquals();
}

var a = new A();
var x = a.getter;
var y = a.getter;
expect(x, y);   // so are x and y "the same value"?
x.value++;
expect(x, y);   // fails.  so x and y _were_ the same value, but no longer are??
```

on the other hand, there may be legitimate cases where x and y are distinct by some criterion, but not for the intended usage.  consider the case of a page object:
```dart
@EnsureTag('editor')
class EditorPO {
  @root
  PageLoaderElement _root;
  Future<String> get value => _root.visibleText;
}
```
here editor.value will generate a different Future each time it's called, but they are still equivalent according to how they are intended to be used, namely, get the String out of them.  so the advice should probably incorporate intended usage.  consider adding to the style guide something like
```
("same value" means can't be distinguished for all intended purposes
but doesn't literally need to be ==)
```

munificent:
This is a lot of new prose. :( Deciding when to use a getter versus a
method seems to be difficult for many Dart users, so I felt it was
worth going into detail.

I also tweaked a bunch of examples and added a couple of other related
guidelines.

Fix #298. Fix #569.


src/_guides/language/effective-dart/design.md
```diff
@@ -260,18 +260,56 @@ property, and should be named as such using a phrase that describes *what* the
...
 
 This guideline is deliberately softer than the previous one. Sometimes a method
 has no side effects but is still simpler to name with a verb phrase like
 `list.take()` or `string.split()`.
 
 
+### CONSIDER an imperative verb phrase for a function or method if you want to draw attention to the work it performs.
+
+When a member produces a result without any side effects, it should usually be a
+getter or a method with noun phrase name describing the result it returns.
```
kwalrath: noun -> a noun

src/_guides/language/effective-dart/design.md
```diff
@@ -260,18 +260,56 @@ property, and should be named as such using a phrase that describes *what* the
...
+However, sometimes the work required to produce that result is important. It may
+be prone to runtime failures, or use heavyweight resources like networking or
+file IO. In cases like this, where you want the caller to think about the work
```
kwalrath: IO -> I/O

src/_guides/language/effective-dart/design.md
```diff
@@ -260,18 +260,56 @@ property, and should be named as such using a phrase that describes *what* the
...
+<?code-excerpt "misc/lib/effective_dart/design_good.dart (verb-for-func-with-work)"?>
+{% prettify dart %}
+var table = database.downloadData();
+var packageVersions = packageGraph.solveConstraints();
+{% endprettify %}
+
+Note, though, that this guideline is softer than the previous two. The work an
+operation performs is often an implementation detail that isn't relevant to the
+caller, and performance and robustness boundaries change over time. Most of the
+time, name your members based on *what* it does for the caller, not *how* it
```
kwalrath: it does -> they do (x2)

src/_guides/language/effective-dart/design.md
```diff
@@ -260,18 +260,56 @@ property, and should be named as such using a phrase that describes *what* the
...
+does it.
+
+
+### AVOID starting a method name with `get`.
+
+In most cases, the method should be an actual getter whose name is whatever you
```
kwalrath: I found this hard to read. Maybe:

the method ... -> a method that you're tempted to name `getXyz` should be an actual getter with the name `xyz`. (and then drop the example, maybe)

(I see we use ___ else for a stand-in name, but that would be bad at the beginning of a member name.)

We could use italics to indicate that the name is a placeholder.

src/_guides/language/effective-dart/design.md
```diff
@@ -260,18 +260,56 @@ property, and should be named as such using a phrase that describes *what* the
...
+would put after the `get`. For example, instead of a method named
+`getBreakfastOrder()`, define a getter named `breakfastOrder`.
+
+Even if the member does need to be a method because it takes arguments or
+otherwise isn't a good fit for a getter, you should still avoid `get`. Like the
+previous guidelines states, you can give it a noun phrase name like
```
kwalrath: states -> state (or guidelines -> guideline)
give it -> use

src/_guides/language/effective-dart/design.md
```diff
@@ -260,18 +260,56 @@ property, and should be named as such using a phrase that describes *what* the
...
+`breakfastOrder()` or choose a verb that more usefully describes how it produces
```
kwalrath: how it -> how the method

("it" in this sentence is unclear)

src/_guides/language/effective-dart/design.md
```diff
@@ -658,50 +696,121 @@ can.
...
+button.canShow;
+dataSet.minimumValue;
 {% endprettify %}
 
-Unlike other languages, in Dart we don't require getters to be particularly fast
-or have certain complexity guarantees. Calling `length` on an Iterable may be
-`O(n)`, and that's OK.
-
 
 ### DO use a setter for operations that conceptually change a property.
```
kwalrath: a setter -> setters
(Or change the previous heading from plural to singular: DO use a getter for an operations that conceptually accesses a property.)

src/_guides/language/effective-dart/design.md
```diff
@@ -658,50 +696,121 @@ can.
...
 
-If the name of the method starts with `set` that's often a sign that it could be
-a setter. More specifically, use a setter instead of a method when it:
+The forces pushing you towards a setter versus a method are similar to when
```
kwalrath: This sentence seems to drop the "forces" metaphor midway through, and it's a bit disconcerting. Maybe:

Deciding between a setter versus a method is similar to deciding between a getter versus a method.

src/_guides/language/effective-dart/design.md
```diff
@@ -658,50 +696,121 @@ can.
...
+Deciding when a member should be a getter versus a method is a challenging,
+subtle, but important part of good API design, hence this very long guideline.
+Some other language's cultures shy away from getters. They only use them when
+the operation is almost exactly like a field — it does a miniscule amount of
+calculation on state that lives entirely on the object. Anything more complex or
+heavyweight than that gets `()` after the name to signal "computation goin' on
+here!" because a bare name after a `.` means "field".
+
+Dart is *not* like that. In Dart, *all* dotted names are member invocations that
+may do computation. Fields are the special-case — getters whose implementation
```
kwalrath: special-case -> special
getters -> they're getters

src/_guides/language/effective-dart/design.md
```diff
@@ -658,50 +696,121 @@ can.
...
+is provided by the language. In other words, getters are not "particularly slow
+fields" in Dart, fields are "particularly fast getters".
```
kwalrath: , -> ; (if you don't hate semicolons)
or even , -> :
The "," made this a little hard to parse initially.

src/_guides/language/effective-dart/design.md
```diff
@@ -658,50 +696,121 @@ can.
...
+
+Even so, choosing a getter over a method sends an important signal to the caller
+and you should use a getter when that's the signal you want to send. The signal,
```
kwalrath: I'd just delete the rest of the sentence after "caller" (~and...to send~).

src/_guides/language/effective-dart/design.md
```diff
@@ -658,50 +696,121 @@ can.
...
-  * **Does not take any arguments.**
-  * **Returns a value.**
-  * **Is side-effect free.** Invoking a getter shouldn't change any
-  externally-visible state (caching internally or lazy initialization is
-  OK). Invoking the same getter repeatedly should return the same value
-  unless the object is explicitly changed between calls.
+*   **The operation does not have user-visible side effects.** Accessing a real
+    field does not alter the object or any other state in the program. It
+    doesn't produce output, write files, etc. A getter should generally follow
+    that.
```
kwalrath: "that" doesn't seem like enough, but I don't know what else to use. "follow that precedent"? Maybe just delete the sentence?

src/_guides/language/effective-dart/design.md
```diff
@@ -658,50 +696,121 @@ can.
...
+    list.clear; // Modifies object.
+    {% endprettify %}
+
+*   **The operation is *idempotent*.** "Idempotent" is an odd word that, in this
+    context, basically means that calling the operation multiple times produces
+    the same result each time, unless some state is explicitly modified between
+    those calls. (Obviously, `list.length` produces different results if you add
+    an element to the list between calls.)
+
+    "Same" here does not mean a getter must literally produce an identical
```
kwalrath: Same -> Same result
(I'd lost track of _what_ was the same by this paragraph.)

src/_guides/language/effective-dart/design.md
```diff
@@ -658,50 +696,121 @@ can.
...
+    object on successive calls. Requiring that would force many getters to have
+    brittle caching, which negates the whole point of using a getter. It's
+    common, and perfectly fine, for a getter to return a new future or list each
+    time you call it. The important part is that the future completes to the
+    same value and the list contains the same elements.
```
kwalrath: and -> , and

src/_guides/language/effective-dart/design.md

src/_guides/language/effective-dart/design.md

src/_guides/language/effective-dart/design.md

src/_guides/language/effective-dart/design.md
munificent: Agree it was awkward. Tried something a little different.

src/_guides/language/effective-dart/design.md

src/_guides/language/effective-dart/design.md

src/_guides/language/effective-dart/design.md

src/_guides/language/effective-dart/design.md
munificent: 👍 Done.

src/_guides/language/effective-dart/design.md

src/_guides/language/effective-dart/design.md
munificent: Generally dislike ";", but I think it's best here.

src/_guides/language/effective-dart/design.md

src/_guides/language/effective-dart/design.md
munificent: Deleting it feels like it leaves it up to the reader to make the connection. Tried a different sentence.

src/_guides/language/effective-dart/design.md

src/_guides/language/effective-dart/design.md

kwalrath:
LGTM once you fix the test.

```
Analyzing lib, test...
  error â€¢ Local variable 'list' can't be referenced before it is declared at lib/effective_dart/design_good.dart:80:15 â€¢ referenced_before_declaration
1 error found.
```

examples/misc/lib/effective_dart/design_good.dart
```diff
@@ -36,28 +36,37 @@ void miscDeclAnalyzedButNotTested() {
...
-    list.elementAt(3);
-    list.firstWhere(test);
-    string.codeUnitAt(4);
+    var element = list.elementAt(3);
+    var first = list.firstWhere(test);
+    var char = string.codeUnitAt(4);
     // #enddocregion noun-for-func-returning-value
   };
 
+  (bool Function(dynamic) test) {
```
chalin: This anonymous function doesn't need any parameters (`test` isn't used in the body).
On the other hand, you could make `database` and `packageGraph` parameters.

munificent: Done.

Issue #721, opened 2018-03-30
Add redirects


kwalrath:
This week's 404 report showed the following problem URLs:

* /docs/pub-package-manager (never noticed this before, but it's apparently been around quite a while; it's probably from a very old version of dartlang.org)
* /eclipse (should redirect to /tools)
* /flutter (should redirect to flutter.io)
* /polymer-old (should redirect to https://webdev.dartlang.org/guides/polymer)
* /tools/pub/cmd/pub-build (should redirect to https://webdev.dartlang.org/tools/pub/pub-build)

PR #729, opened 2018-03-31
Redirects and link tweaks

kwalrath:
I ran linkcheck internally, externally, and against the old site's URLs
(deploy/urls/old_site_urls.txt).


kwalrath:
Fixes #721.

PR #730, opened 2018-03-31
Better async example.

### LINKED ISSUE #79: Inconsistencies in "Effective Dart: Usage" Async section
_From @lassedamgaard on October 29, 2015 10:1_

https://www.dartlang.org/effective-dart/usage/#prefer-asyncawait-over-using-raw-futures

This example uses `return await` which contradicts another recommendation:

``` dart
Future<bool> doAsyncComputation() async {
  try {
    var result = await longRunningCalculation();
    return await verifyResult(result);
  } catch(e) {
    log.error(e);
    return false;
  }
}
```

IMO it should be:

``` dart
Future<bool> doAsyncComputation() async {
  try {
    var result = await longRunningCalculation();
    return verifyResult(result);
  } catch(e) {
    log.error(e);
    return false;
  }
}
```

Furthermore, the 'bad' example using the `Future` API is overly verbose because it contradicts the recommendation to use tear-offs instead of closures and also unnecessarily returns a Future in `catchError` instead of just letting the return value get auto-wrapped in a Future:

``` dart
Future<bool> doAsyncComputation() {
  return longRunningCalculation().then((result) {
    return verifyResult(result);
  }).catchError((e) {
    log.error(e);
    return new Future.value(false);
  });
}
```

This could be written much cleaner using fat arrows as:

``` dart
  Future<bool> doAsyncComputation() =>
      longRunningCalculation().then(verifyResult).catchError((e) {
        log.error(e);
        return false;
      });
```

As such I don't think it's a good example of when async/await is much better than the Future API. Especially in cases where the whole method body is wrapped in a try catch and you only need to do 1 or two async things I often prefer to use the Future API in conjuction with fat arrows instead of async/await with try/catch. Now that's just my style but I think another example would drive the point home better.

_Copied from original issue: dart-lang/www.dartlang.org#1507_

munificent:
Also remove a bullet point that no longer applies with sync-async.

Fix #79.


src/_guides/language/effective-dart/usage.md
```diff
@@ -1047,7 +1047,7 @@ Future<int> countActivePlayers(String teamName) [!async!] {
 {% prettify dart %}
 Future<int> countActivePlayers(String teamName) {
   return downloadTeam(teamName).then((team) {
-    if (team == null) return 0;
```
kwalrath: Wow, I assumed that this would be auto-wrapped.

munificent: It will at runtime, but type inference gets a little confused when you have a function body that returns both an immediate value and a future, I think.

src/_guides/language/effective-dart/usage.md
```diff
@@ -1020,35 +1020,41 @@ The following best practices apply to asynchronous coding.
 
 ### PREFER async/await over using raw futures.
 
-Explicit asynchronous code is notoriously hard to read and debug, even when
-using a nice abstraction like futures. This is why we added `async`/`await` to
-the language. They make a huge improvement in readability code and let you use
-all of the built-in control flow structures of the language within your
-asynchronous code.
+Asynchronous code is notoriously hard to read and debug, even when using a nice
+abstraction like futures. This is why we added `async`/`await` to the language.
```
chalin: Do we really want to bring language evolution into this discussion? cc @kwalrath WDYT?

How about more simply:

> Asynchronous code is notoriously hard to read and debug, even when using a nice abstraction like futures. The `async`/`await` syntax can contribute to improving code readability while letting you use all of the Dart control flow structures within your async code.

examples/misc/lib/effective_dart/usage_bad.dart
```diff
@@ -152,6 +139,34 @@ void miscDeclAnalyzedButNotTested() {
 
 //----------------------------------------------------------------------------
 
+class Player {
+  bool get isActive => false;
+}
```
chalin: You can drop `class Player` since `usage_good.dart` is imported and it defines `Player`.

examples/misc/lib/effective_dart/usage_bad.dart
```diff
@@ -152,6 +139,34 @@ void miscDeclAnalyzedButNotTested() {
...
+
+class Team {
+  Future<List<Player>> get roster => null;
+  Future<Team> downloadTeam(String name) => null;
+
```
chalin: How about replacing L146-148 by the following (which can help us ensure that we are using the same API in both places):

```dart
class BadTeam extends Team {
```

(Unfortunately we have to keep the local `log` field otherwise it clashes with the `log` from `math.dart`.)

chalin:
Great example btw!

src/_guides/language/effective-dart/usage.md
kwalrath: Good point, adding async-await is ancient history.

munificent: Good call. Simplified it even further.

PR #734, opened 2018-04-02
Tweak style guidelines.

### LINKED ISSUE #85: Suggest to re-arrange program source to make dartfmt produce better output
_From @eneev-a on November 5, 2015 21:13_

Can we include in [Effective Dart: Style Guide](https://www.dartlang.org/effective-dart/style/) a suggestion to re-arrange program source when formatter produces questionable output?

If the program code is so tricky that it baffles the formatter, it might be worth trying to re-arrange code so that it doesn't.

A similar suggestion is done in [Effective Go: Formatting](https://golang.org/doc/effective_go.html#formatting).

_Copied from original issue: dart-lang/www.dartlang.org#1515_

### LINKED ISSUE #722: 'Effective Dart: Style' page issue: 80 char line limit exception
From URL: https://www.dartlang.org/guides/language/effective-dart/style

Under "AVOID lines longer than 80 characters.", it lists the following exception: "We make an exception for strings containing URIs—mainly imports and exports. Those can remain single-line strings even if they go over the line limit. This makes it easier to search source files for a given path.".

I often see code that has long URIs in comments, however this isn't covered by this exception. I'd suggest adding a few words about allowing long URIs in comments as well as strings.

munificent:
Add a guideline for modifying code to be dartfmt-friendly. Fix #85.

Include comments in the exception for long lines containing paths.
Fix #722.

### DIFF
--- a/src/_guides/language/effective-dart/style.md
+++ b/src/_guides/language/effective-dart/style.md
@@ -346,10 +346,10 @@ formatted output may still be hard to read.
 
 When that happens, reorganize or simplify your code. Consider shortening a local
 variable name or hoisting out an expression into a new local variable. In other
-words, the exact kinds of modifications you'd make if you were hand-formatting
-the code and struggled to get it to your liking. Think of dartfmt as a
-partnership where you work together, sometimes iteratively, to produce beautiful
-code.
+words, make the same kinds of modifications that you'd make if you were
+formatting the code by hand and trying to make it more readable. Think of
+dartfmt as a partnership where you work together, sometimes iteratively, to
+produce beautiful code.
 
 
 ### AVOID lines longer than 80 characters.
@@ -368,8 +368,8 @@ Note that dartfmt does 99% of this for you, but the last 1% is you. It does not
 split long string literals to fit in 80 columns, so you have to do that
 manually.
 
-We make an exception for comments strings containing URIs and file
-paths&mdash;mainly imports and exports. Those can remain on a single line even
+We make an exception for URIs and file paths. When those occur in comments or
+strings (usually in imports and exports), they may remain on a single line even
 if they go over the line limit. This makes it easier to search source files for
 a given path.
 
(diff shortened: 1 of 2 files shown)


src/_guides/language/effective-dart/style.md
```diff
@@ -337,6 +337,21 @@ for you.
...
+### CONSIDER changing your code to make it more formatter-friendly.
+
+The formatter does the best it can with whatever code you throw at it, but it
+can't work miracles. If your code has particularly long identifiers, deeply
+nested expressions, a mixture of different kinds of operators, etc. the
+formatted output may still be hard to read.
+
+When that happens, reorganize or simplify your code. Consider shortening a local
+variable name or hoisting out an expression into a new local variable. In other
+words, the exact kinds of modifications you'd make if you were hand-formatting
```
kwalrath: Let's make this a sentence and reword a bit for easier reading. Maybe:

In other words, make the same kinds of modifications that you'd make if you were formatting the code by hand and trying to make it more readable.

src/_guides/language/effective-dart/style.md
```diff
@@ -353,9 +368,10 @@ Note that dartfmt does 99% of this for you, but the last 1% is you. It does not
 split long string literals to fit in 80 columns, so you have to do that
 manually.
 
-We make an exception for strings containing URIs&mdash;mainly imports and
-exports. Those can remain single-line strings even if they go over the line
-limit. This makes it easier to search source files for a given path.
+We make an exception for comments strings containing URIs and file
```
kwalrath: Are we missing commas or conjunctions here? Also, do we need "mainly imports and exports"? Here's a stab at rewriting it (you can probably do better):

We make exceptions for file paths and for comments and strings that contain URIs.
One advantage of keeping paths on single line is that searching for a particular path is easier.

src/_guides/language/effective-dart/style.md
munificent: Done.

src/_guides/language/effective-dart/style.md
munificent: I did leave out a word there. The intent is to permit any combination of:

* comment containing a file path
* comment containing a URI
* string containing a file path
* string containing a URI

Reworded.

munificent:
Thanks!

PR #740, opened 2018-04-04
Reminder to add SDK bin to PATH when manually installing SDK

chalin:
Addresses https://github.com/dart-lang/site-www/pull/727#discussion_r178967869.


src/tools/sdk/index.md
```diff
@@ -20,6 +20,8 @@ A package manager can help you easily install and update the Dart SDK.
 Don’t want to use a package manager? Other options are
 [building the SDK from source][] and
 [downloading the SDK as a zip file](/tools/sdk/archive).
+If you use either of these options, remember to
```
kwalrath: I feel like even if you use a package manager, you might need to alter your path... but let's see how this flies.

PR #742, opened 2018-04-04
Dart VM, Get Started updates

chalin:
Contributes to #407 

Staged at https://sz-www-2.firebaseapp.com/tutorials/dart-vm/get-started

---

Also included in this PR is a style change that makes the "What's the point?" text look like this:

> screen shot 2018-04-04 at 13 49 23

... instead of this (notice how the bullet indentation seems off):

> screen shot 2018-04-04 at 13 50 01

If you like the new style, I'll update the other pages that have similar "What's the point?" mini-ToCs.

### DIFF
--- a/examples/misc/lib/vm_get_started.dart
+++ b/examples/misc/lib/vm_get_started.dart
@@ -1,0 +1,1 @@
+double milesToKM(double miles) => miles / 0.62;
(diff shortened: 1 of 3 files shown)


chalin:
> For some reason, the image in https://www.dartlang.org/tutorials/dart-vm/get-started#pub-package-manager has extra junk... Did you fix this elsewhere, and I missed it?

The funny thing is that those are two separate images. I dropped one of the images in this PR. (I'm unsure why the second image got included in the first place.)

PR #743, opened 2018-04-04
Revise the guidelines on type annotations.

### LINKED ISSUE #53: Move "CONSIDER omitting the types for local variables" to "Type Annotations" section
_From @ninachen on January 8, 2016 18:20_

https://www.dartlang.org/effective-dart/usage/#consider-omitting-the-types-for-local-variables

Currently, it's under "Variables" in Usage, although I often find myself looking for it under "[Type annotations](https://www.dartlang.org/effective-dart/design/#type-annotations)" in Design.

I can see arguments for putting it in either place, although it seems like it belongs more under Type annotations to me.

_Copied from original issue: dart-lang/www.dartlang.org#1542_

### LINKED ISSUE #392: Update style guide recommendations for dynamic parameters
The style guide [currently says](https://www.dartlang.org/guides/language/effective-dart/design#avoid-annotating-with-dynamic-when-not-required) "AVOID annotating with `dynamic` when not required". However, with the advent of strong mode, more and more code is being analyzed with the `--no-implicit-dynamic` flag, and it's possible that this flag will be enabled by default at some point. This flag explicitly disallows the pattern currently encouraged by the style guide.

See also dart-lang/sdk#27623, particularly [this comment](https://github.com/dart-lang/sdk/issues/27623#issuecomment-333987214).

### LINKED ISSUE #735: Effective Dart: Function, typedef, guidance
I'd like to see guidance around:

* Prefer the new typedef/function type syntax
* Consider using inline function types unless the typedef is strongly needed (with examples)
* Avoid type annotating anything plain `Function` (basically `dynamic` in Dart 2)
* Avoid implementing `Function` (?)

/cc @munificent

munificent:
This is the big one!

I revamped most of the guidelines around static types to be in line with
the more statically-typed inferred way people use Dart today. Specific
changes:

- Added more intro text explaining what type annotations are, and terms
  around "dynamic" and "inferred".

- Re-ordered some guidelines like the one for setters to group related
  ones together.

"DO type annotate public declarations."

- Changed it to allow omitting types when inference fills it in.
- Added some text around when it's good to rely on inference and when it
  isn't.

"PREFER type annotating private declarations."

- Allow omitting types if they can be inferred.
- Mentioned most people annotate private APIs.

"AVOID annotating types for local variables."

- Added qualifier that this only applies to initialized locals since
  locals without initializers are not inferred.

"AVOID annotating types on function expressions."

- Added qualifier around inference.
- Added more context about why this works.

"AVOID annotating with `dynamic` when not required."

- Totally flipped this. Now "PREFER" instead of "AVOID".
- Added explanatory text.

"AVOID annotating with `Function`."

- Added clarification now that "Function" is also used as a keyword for
  the new function type syntax.

"DON'T use the legacy typedef syntax."

- New guideline. With Dart 2, we now have two typedef syntaxes. The old
  one is deprecated.

"PREFER inline function types over typedefs."

- New guideline. Now that you don't *need* to create a typedef for
  function types, added a guideline suggesting using the new syntax.

"CONSIDER using the generalized function type syntax for parameters."

- New guideline. Similar to previous one but looser because some people
  still prefer the old function-typed parameter syntax.

"DO annotate with `Object` instead of `dynamic` to indicate any object
is accepted."

- Rewrote this to be clearer about when you should use each.

Fix #53. Fix #392. Fix #735.


chalin:
Wonderful!

Btw, under _Types_ >> _AVOID defining a one-member abstract class when a simple function will do._, might you want to change the typedef from the old to the new syntax?

Actually, maybe that recommendation could be moved somewhere else and then the section could be renamed from _Types_ to, say, _Classes_? WDYT?

(I'm starting a few of the changes you've made now, but I might only finish tomorrow.)

src/_guides/language/effective-dart/design.md
```diff
@@ -880,26 +880,47 @@ var buffer = new StringBuffer()
...
 
-In Dart 1.x, adding static types to your variables is optional.
+A "type annotation" is a place in your code where you write down the static type
+of a variable, field, function return type, parameter, etc. Type annotations are
+how you tell the type checker your constraints on the kinds of values that flow
+into different parts of your code.
 
-{% include optional-types-2.0.html %}
+In most places, Dart allows you to omit a type annotation and infers a type for
+you based on the nearby context (or defaults to the `dynamic` type). Inference
```
chalin: How about using a comma instead:

> ... based on the nearby context, or defaults to the `dynamic` type.

src/_guides/language/effective-dart/design.md
```diff
@@ -880,26 +880,47 @@ var buffer = new StringBuffer()
...
+is a powerful tool to spare you the effort of writing (and reading) types that
+are obvious or uninteresting, but it's still worth explicitly annotating in some
+cases. The guidelines here strike the best balance we've found between brevity
+and explicitness, flexibility and safety.
 
-{% comment %}
-update-for-dart-2.0
-{% endcomment %}
+The fact that Dart has both type inference and an actual `dynamic` type leads to
```
chalin: I'd suggest dropping "actual" from "a<del>n actual</del> `dynamic` type".

src/_guides/language/effective-dart/design.md
```diff
@@ -880,26 +880,47 @@ var buffer = new StringBuffer()
...
+some confusion about what it means to say code is "untyped". Does that mean the
+code is dynamically typed, or that you didn't *write* the type? To avoid that
+confusion, we avoid saying "untyped" and instead use:
```
chalin: > ... instead use <ins>the following terminology</ins>:

src/_guides/language/effective-dart/design.md
```diff
@@ -880,26 +880,47 @@ var buffer = new StringBuffer()
...
 
-### DO type annotate public APIs.
+*   If the code is "type annotated", the type was explicitly written in the
+    code.
 
-Type annotations are important documentation for how a library should be used.
-Annotating the parameter and return types of public methods and functions helps
-users understand what the API expects and what it provides.
+*   If the code is "inferred", no type annotation was written down, and Dart
```
chalin: > was written <del>down</del>,

src/_guides/language/effective-dart/design.md
```diff
@@ -880,26 +880,47 @@ var buffer = new StringBuffer()
...
+    successfully figured out the type on its own. Inference can fail, in which
+    case the guidelines don't consider that "inferred". In some places,
+    inference failure is a static error. In others, Dart uses `dynamic` as the
+    fallback type.
```
chalin: Hmm, I'd hesitate to use "inference failure" for anything other than an error. Using "dynamic" IMHO isn't a failure, and in fact it _might_ be what the user wants/expects.

src/_guides/language/effective-dart/design.md
```diff
@@ -880,26 +880,47 @@ var buffer = new StringBuffer()
...
+
+*   If the code is "dynamic", then its static type is the special `dynamic`
+    type. Code can be explicitly annotated `dynamic` or it can be inferred.
+
+*   If the code has any other type, it's not "dynamic". Code can end up with a
+    non-`dynamic` type by being explicitly annotated with some other type, or by
+    having Dart infer a non-`dynamic` type.
+
```
chalin: In all the cases above, rather than "If the code...", how about "If the declaration..." (because, I believe, type annotations appear in declarations, not in arbitrary code).

src/_guides/language/effective-dart/design.md
```diff
@@ -880,26 +880,47 @@ var buffer = new StringBuffer()
...
+In other words, whether some code is annotated or inferred is orthogonal to
+whether it is `dynamic` or some other type.
 
-Note that if a public API accepts a range of values that Dart's type system
-cannot express, then it is acceptable to leave that untyped. In that case, the
-implicit `dynamic` *is* the correct type for the API.
+### DO type annotate public declarations whose type isn't inferred.
```
chalin: > DO type annotate a public declaration whose type isn't inferred.

I know you've switched to plural elsewhere, but in the plural form is seems just a bit off.

src/_guides/language/effective-dart/design.md
```diff
@@ -908,37 +929,39 @@ install(id, destination) => ...
...
-
-### DON'T specify a return type for a setter.
-
-{:.bad-style}
-<?code-excerpt "misc/lib/effective_dart/design_bad.dart (avoid_return_types_on_setters)"?>
-{% prettify dart %}
-void set foo(Foo value) { ... }
-{% endprettify %}
+In some cases, especially with constants, Dart can infer the type for you. In
+that case, it's fine to omit the annotation:
```
chalin: How about:

> When Dart can infer the type for you, then it's fine to omit the annotation:
> ```
> const screenWidth = 640; // Infers "int"
> ```

src/_guides/language/effective-dart/design.md
```diff
@@ -908,37 +929,39 @@ install(id, destination) => ...
...
 
 {:.good-style}
-<?code-excerpt "misc/lib/effective_dart/design_good.dart (avoid_return_types_on_setters)"?>
+<?code-excerpt "misc/lib/effective_dart/design_good.dart (inferred)"?>
 {% prettify dart %}
-set foo(Foo value) { ... }
+const screenWidth = 640; // Infers "int" from 640.
 {% endprettify %}
```
chalin: I'd drop the "from 640".

src/_guides/language/effective-dart/design.md
```diff
@@ -908,37 +929,39 @@ install(id, destination) => ...
...
 
+You may still wish to explicitly annotate even when inference would fill in the
+same type. If the inference relies on values or declarations from other
+libraries, you may want to annotate the type in *your* declaration so that a
+change to that other library doesn't silently change the type of your own API
+without you realizing.
```
chalin: Yes! Thank you for mentioning that explicitly.

src/_guides/language/effective-dart/design.md
```diff
@@ -908,37 +929,39 @@ install(id, destination) => ...
...
 
-### PREFER type annotating private declarations.
+This guideline does *not* mean you can never use the dynamic type in a public
+signature. Sometimes `dynamic` is the best type. What it states is that when
+you want to use `dynamic`, you should do so *explicitly*.
```
chalin: The "What it states ..." is a bit confusing. How about dropping that sentence and after "Sometimes `dynamic` is the best type" directing/linking the reader to the corresponding guideline?

src/_guides/language/effective-dart/design.md
```diff
@@ -952,13 +975,76 @@ class CallChainVisitor {
 }
 {% endprettify %}
 
+This guideline is deliberately softer than the previous one because the benefits
+of static typing are slightly reduced when code is confined within a single
+library. In practice, though, most Dart users type annotate all declarations,
+public or private.
```
chalin: > In practice, though, most Dart users type annotate all declarations, public or private.

I'm glad to read that (it supports what I've been encouraging my students to do).

munificent:
> Btw, under Types >> AVOID defining a one-member abstract class when a simple function will do., might you want to change the typedef from the old to the new syntax?

Good catch, done.

> Actually, maybe that recommendation could be moved somewhere else and then the section could be renamed from Types to, say, Classes? WDYT?

Yup! I had that same thought and made that change here:
https://github.com/dart-lang/site-www/commit/6836e7d0c4bb1008de2dbc517baa4661a0a48ecd

src/_guides/language/effective-dart/design.md
```diff
@@ -952,13 +975,76 @@ class CallChainVisitor {
...
+Function bodies in modern code tend to be short, and the types of local
+variables are almost always obvious and inferrable from the initializing
+expression. Writing that type down is often simply visual noise. Instead, spend
+that energy choosing a great *name* for your local variable.
+
+{:.good-style}
+<?code-excerpt "misc/lib/effective_dart/design_good.dart (omit-types-on-locals)"?>
+{% prettify dart %}
+List<List<Ingredient>> possibleDesserts(Set<Ingredient> pantry) {
+  var desserts = <List<Ingredient>>[];
```
chalin: Make this `final`?

src/_guides/language/effective-dart/design.md
```diff
@@ -952,13 +975,76 @@ class CallChainVisitor {
...
+  for (var recipe in cookbook) {
+    if (pantry.containsAll(recipe)) {
+      desserts.add(recipe);
+    }
+  }
+
```
chalin: Remove this blank line (in both good and bad excerpts)?

src/_guides/language/effective-dart/design.md
```diff
@@ -952,13 +975,76 @@ class CallChainVisitor {
 }
 {% endprettify %}
 
+This guideline is deliberately softer than the previous one because the benefits
+of static typing are slightly reduced when code is confined within a single
+library. In practice, though, most Dart users type annotate all declarations,
+public or private.
+
+
+### AVOID annotating types for initialized local variables.
```
chalin: Could we emphasize "initialized" (say, styling it with italics)?

kwalrath:
I've staged this under https://kw-www-dartlang-1.firebaseapp.com/guides/language/effective-dart.

src/_guides/language/effective-dart/design.md
```diff
@@ -952,13 +975,76 @@ class CallChainVisitor {
...
-likely doesn't need types.
+Anonymous functions are almost always immediately passed to a method expecting a
+function of some type. (If it isn't used immediately, it's usually worth making
+it a named function declaration.) When a function expression is created in a
+typed context, Dart tries to infer the function's parameter types based on the
+expected type.
+
+For example, when you pass a function expression to `Iterable.map()`, your
+function's parameter type is inferred based on the type of callback that
+`map()` expects and you don't need to annotate it:
```
chalin: Maybe drop "and you don't need to annotate it"?

src/_guides/language/effective-dart/design.md
```diff
@@ -952,13 +975,76 @@ class CallChainVisitor {
...
+if (node is Constructor) {
+  parameters = node.signature;
+} else if (node is Method) {
+  parameters = node.parameters;
+}
+{% endprettify %}
+
 
-### AVOID annotating types on function expressions.
+### AVOID annotating inferrable parameter types on function expressions.
```
chalin: Emphasize _inferrable_?

src/_guides/language/effective-dart/design.md
```diff
@@ -974,79 +1060,260 @@ var names = people.map((Person person) {
...
-### AVOID annotating with `dynamic` when not required.
+### PREFER annotating with `dynamic` even when not required.
 
-{% comment %}
-update-for-dart-2.0
-https://github.com/dart-lang/site-www/pull/480#discussion_r157355593
-{% endcomment %}
+When you omit a type annotation, *most* of the time, Dart infers a correct
+static type for you. That word "most" is the alarming part. Sometimes, inference
+fails and silently uses `dynamic` instead. If you didn't realize that inference
```
chalin: I'm hoping that we can avoid the use of "fails" here. I'm not sure how to word it, but maybe using terminology like "defaults to `dynamic`"? (Will have to think about this more.)

src/_guides/language/effective-dart/design.md
```diff
@@ -974,79 +1060,260 @@ var names = people.map((Person person) {
...
+failed, you now have code that *looks* statically safe but actually has no
+type checking whatsoever.
 
-In most places in Dart, a type annotation can be omitted, in which case the type
-will automatically be `dynamic`. Thus, omitting the type annotation entirely is
-semantically equivalent but more terse.
+Sometimes, `dynamic` is what you want. Dart does support a dynamic type for good
+reasons. But a casual reader of your code who sees an annotation is missing has
```
chalin: > ... for good reason.

@kwalrath WDYT?

src/_guides/language/effective-dart/design.md
```diff
@@ -974,79 +1060,260 @@ var names = people.map((Person person) {
 });
 {% endprettify %}
 
+In rare cases, the surrounding context is not precise enough to provide a type
+for one or more of the function's parameters. In those cases, you may need to
+annotate.
+
 
-### AVOID annotating with `dynamic` when not required.
+### PREFER annotating with `dynamic` even when not required.
```
chalin: I'm very glad for this switch in the guideline (it matches my intuition better).

src/_guides/language/effective-dart/design.md
```diff
@@ -1005,79 +1091,260 @@ var names = people.map((Person person) {
...
+
+Using `dynamic` explicitly when that is the type you want makes that intention
+clear.
 
 {:.good-style}
-<?code-excerpt "misc/lib/effective_dart/design_good.dart (avoid-dynamic)"?>
+<?code-excerpt "misc/lib/effective_dart/design_good.dart (prefer-dynamic)"?>
 {% prettify dart %}
-lookUpOrDefault(String name, Map map, defaultValue) {
+T lookUp<T>(String name, Map<String, T> map, T defaultValue) {
```
chalin: The emphasis in this code excerpt are the type annotations so _maybe_ the body can be elided. If we keep the body, how about making it more null-operator smart (in both the good and the bad examples)?

```dart
  T lookUp<T>(String name, Map<String, T> map, T defaultValue) =>
      map[name] ?? defaultValue;
```

src/_guides/language/effective-dart/design.md
```diff
@@ -1005,79 +1091,260 @@ var names = people.map((Person person) {
...
-### AVOID annotating with `Function`.
+<aside class="alert alert-info" markdown="1">
+
+Before Dart 2, this guideline stated the exact opposite: *don't* annotate with
+`dynamic` when it is implicit. With the new stronger type system and type
+inference, users now expect Dart to behave like an inferred statically-typed
+language. With that mental model, it is an unpleasant surprise to discover that
+a region of code has silently lost all of the safety and performance of static
+types.
+
```
chalin: Nit: drop the blank lines and indent the content of the aside by 2 spaces.

src/_guides/language/effective-dart/design.md
```diff
@@ -1005,79 +1091,260 @@ var names = people.map((Person person) {
...
 
-The `Function` type is barely more precise than using no annotation at all. If
-you're bothering to annotate, it's better to use a precise function type that
-describes the signature and return type of the function.
+### AVOID annotating with only `Function`.
 
-If you are annotating a field, this does mean you have to create a typedef, but
-that's usually worth doing.
+The name `Function` by itself without any return type or parameter signature
+refers to the special `Function` type. This type is only marginally more useful
```
chalin: In the following, could we link `Function` to a page that covers more about it?

> ... refers to the special `Function` type.

src/_guides/language/effective-dart/design.md
```diff
@@ -1005,79 +1091,260 @@ var names = people.map((Person person) {
...
+a region of code has silently lost all of the safety and performance of static
+types.
+
+</aside>
+
 
-The `Function` type is barely more precise than using no annotation at all. If
-you're bothering to annotate, it's better to use a precise function type that
-describes the signature and return type of the function.
+### AVOID annotating with only `Function`.
```
chalin: How about make this a positive recommendation?

> ### PREFER full `Function` type annotations

src/_guides/language/effective-dart/design.md
```diff
@@ -1005,79 +1091,260 @@ var names = people.map((Person person) {
...
 {% prettify dart %}
-bool isValidString(String value, Function predicate) => ...
+bool isValid(String value, Function test) => ...
+{% endprettify %}
+
+In Dart 1, users sometimes used `Function` to avoid the necessary tedium of
+creating a typedef to use the function type in a field, variable, or generic
+type argument. Dart 2 has a [general function type syntax][fn syntax] that can
+be used anywhere a type annotation can occur, so this shortcut is no longer
+needed.
```
chalin: The Dart 1 sentence is quite a mouthful. But that got me thinking, why not just drop this entire paragraph? Do we _really_ need to try to explain the Dart 1 limitation in detail?

src/_guides/language/effective-dart/design.md
```diff
@@ -1005,79 +1091,260 @@ var names = people.map((Person person) {
...
+
+[fn syntax]: #prefer-inline-function-types-over-typedefs
+
+One exception to this guideline is if you want a type that represents the union
```
chalin: Since we now have a "PREFER annotating with `dynamic` ..." guideline, I think that it would make sense to pull out this "one exception" into its own guideline.

> ### PREFER annotation with `Function` only ...

Maybe complete with something along the lines of "... when the function arity isn't fixed/know"?

src/_guides/language/effective-dart/design.md
```diff
@@ -1005,79 +1091,260 @@ var names = people.map((Person person) {
...
+{:.good-style}
+<?code-excerpt "misc/lib/effective_dart/design_good.dart (function-arity)"?>
+{% prettify dart %}
+void handleError(void Function() operation, Function errorHandler) {
+  try {
+    operation();
+  } catch (err, stack) {
+    if (errorHandler is Function(Object)) {
+      errorHandler(err);
+    } else if (errorHandler is Function(Object, Object)) {
```
chalin: Cool, I didn't know that we could do that!

src/_guides/language/effective-dart/design.md
```diff
@@ -1005,79 +1091,260 @@ var names = people.map((Person person) {
...
+Dart has two notations for defining a named typedef for a function type. The
+original syntax looks like:
+
+{:.bad-style}
+<?code-excerpt "misc/lib/effective_dart/design_bad.dart (old-typedef)"?>
+{% prettify dart %}
+typedef int Comparison<T>(T a, T b);
+{% endprettify %}
+
+It has a couple of problems:
```
chalin: Mabye: It --> The original syntax

src/_guides/language/effective-dart/design.md
```diff
@@ -1005,79 +1091,260 @@ var names = people.map((Person person) {
...
+{% endprettify %}
+
+The new syntax can express anything the old syntax could express and more, and
+lacks the error-prone misfeature where a single identifier is treated as the
+parameter's name. The same function type syntax after the `=` in the typedef is
+also allowed anywhere a type annotation may appear, giving us a single
+consistent way to write function types anywhere in a program.
+
+The old typedef syntax is still supported to avoid breaking existing code, but
+is deprecated.
```
chalin: but is --> but it's

src/_guides/language/effective-dart/design.md
```diff
@@ -1005,79 +1091,260 @@ var names = people.map((Person person) {
...
+    void Function(Event) last;
+    for (var observer in _observers) {
+      observer(event);
+      last = observer;
+    }
+
+    return last;
+  }
+}
+{% endprettify %}
```
chalin: Personally, I find the `In a ...` comments distracting.

src/_guides/language/effective-dart/design.md
```diff
@@ -1005,79 +1091,260 @@ var names = people.map((Person person) {
...
+
+
+### PREFER inline function types over typedefs.
+
+In Dart 1, if you wanted to use a function type for a field, variable, or
+generic type argument, you had to first define a typedef for it. Dart 2 supports
+a function type syntax that can be used anywhere a type annotation is allowed:
+
+{:.good-style}
+<?code-excerpt "misc/lib/effective_dart/design_good.dart (function-type)"?>
```
chalin: There is a lot of code in the excerpt; consider adding highlighting of `Function` lexemes to help focus the reader's attention:

    <?code-excerpt "misc/..." replace="/Function/[!$&!]/g"?>

src/_guides/language/effective-dart/design.md
```diff
@@ -1005,79 +1091,260 @@ var names = people.map((Person person) {
...
 
 
 ### DO annotate with `Object` instead of `dynamic` to indicate any object is accepted.
 
-Some operations will work with any possible object. For example, a log method
-could take any object and call `toString()` on it. Two types in Dart permit all
-objects: `Object` and `dynamic`. However, they convey two different things.
+Some operations work with any possible object. For example, a `log()` method
+could take any object and call `toString()` on it. Two types in Dart permit
+values of all classes: `Object` and `dynamic`. However, they convey different
```
chalin: Maybe:

> ... permit values from any class.

Or more simply?

> ... permit any value.

src/_guides/language/effective-dart/design.md
```diff
@@ -1087,8 +1354,8 @@ void log(Object object) {
   print(object.toString());
 }
 
-// Only accepts bool or String, which can't be expressed in a type annotation.
-bool convertToBool(arg) {
+// Only accepts bool or String, which type system can't express.
```
chalin: > ... which <ins>the</ins> type system ...

src/_guides/language/effective-dart/design.md
```diff
@@ -1087,8 +1354,8 @@ void log(Object object) {
   print(object.toString());
```
chalin: _Maybe_:

```dart
void log(Object o) => print(o.toString());
```

chalin: I'd suggest dropping the "// Accepts any object" comment.

src/_guides/language/effective-dart/design.md
```diff
@@ -1087,8 +1354,8 @@ void log(Object object) {
...
 }
 
-// Only accepts bool or String, which can't be expressed in a type annotation.
-bool convertToBool(arg) {
+// Only accepts bool or String, which type system can't express.
+bool convertToBool(dynamic arg) {
   if (arg is bool) return arg;
   if (arg is String) return arg == 'true';
   throw new ArgumentError('Cannot convert $arg to a bool.');
```
chalin: _Maybe_:

```dart
  bool convertToBool(dynamic arg) => arg is bool
      ? arg
      : arg is String
          ? arg == 'true'
          : throw new ArgumentError('bool or String expected, got $arg');
```

src/_guides/language/effective-dart/design.md
munificent: Added a link to the core library class.

src/_guides/language/effective-dart/design.md
munificent: Changed it to "PREFER signatures in function type annotations". How's that?

We don't really have a name for the new function type syntax, so it's sort of hard to talk about.

src/_guides/language/effective-dart/design.md
munificent: Good point. I think the later guideline on preferring function types over typedefs covers this. Deleted.

src/_guides/language/effective-dart/design.md
munificent: That feels like overkill to me. This guideline is already a "PREFER" one, which implies there are exceptions and other PREFER/AVOID guidelines sometimes spell out those exceptions too.

src/_guides/language/effective-dart/design.md

src/_guides/language/effective-dart/design.md

src/_guides/language/effective-dart/design.md
munificent: Do you? I thought it would be good to spell out all of the cases and what each case represents. (In particular, I think using a function type as a type argument or a return type is non-obvious.)

src/_guides/language/effective-dart/design.md
munificent: Ah, good idea! What that, I think you're right that the comments are needless.

src/_guides/language/effective-dart/design.md

src/_guides/language/effective-dart/design.md

src/_guides/language/effective-dart/design.md
munificent: Hmm, I know that's a thing that works now, but I don't find it particularly idiomatic outside of setters and don't really want to encourage it.

src/_guides/language/effective-dart/design.md
kwalrath: I tried to figure out a way to do the emphasis using just word order, but I couldn't make it work.

src/_guides/language/effective-dart/design.md
munificent: Done, and turned the other into a real doc comment.

src/_guides/language/effective-dart/design.md
munificent: Nested ternary operators in a style guide? I'll pass. 😄

src/_guides/language/effective-dart/design.md
```diff
@@ -1005,79 +1091,260 @@ var names = people.map((Person person) {
...
+failed, you now have code that *looks* statically safe but actually has no
+type checking whatsoever.
 
-In most places in Dart, a type annotation can be omitted, in which case the type
-will automatically be `dynamic`. Thus, omitting the type annotation entirely is
-semantically equivalent but more terse.
+Sometimes, `dynamic` is what you want. Dart does support a dynamic type for good
+reasons. But a casual reader of your code who sees an annotation is missing has
+no way of knowing if you intended it to be `dynamic`, expected inference to fill
+in some other type, or you simply forgot to write the annotation.
```
kwalrath: you simply -> simply

src/_guides/language/effective-dart/design.md
```diff
@@ -1005,79 +1091,260 @@ var names = people.map((Person person) {
...
+
+Using `dynamic` explicitly when that is the type you want makes that intention
```
kwalrath: A little hard to read. How about:

When `dynamic` is the type you want, using it explicitly makes your intent clear.

src/_guides/language/effective-dart/design.md
```diff
@@ -1005,79 +1091,260 @@ var names = people.map((Person person) {
...
+clear.
```
kwalrath: I don't see `dynamic` in this example. Did it get overwritten?

src/_guides/language/effective-dart/design.md
kwalrath: Consider having a standard, searchable way to mark notes about changes in the guidelines. Maybe:

**Changed guideline:** Before...

src/_guides/language/effective-dart/design.md
chalin: Ok

src/_guides/language/effective-dart/design.md
chalin: That's good.

src/_guides/language/effective-dart/design.md
chalin: I prefer the highlights to the comments. It pulls in my eye and tickles my curiosity: "why are there so many highlighted regions? Oh, they are all different usages, cool!" :)

@kwalrath what's your take on this: leave the comments out or put them back in? (The comments would describe what kind of function type annotation it is: field, generic type argument, etc.)

src/_guides/language/effective-dart/design.md
chalin: Ok

src/_guides/language/effective-dart/design.md
```diff
@@ -1005,90 +1091,262 @@ var names = people.map((Person person) {
...
 
-If you are annotating a field, this does mean you have to create a typedef, but
-that's usually worth doing.
+</aside>
+
+
+### PREFER signatures in function type annotations.
+
+The identifier `Function` by itself without any return type or parameter
+signature refers to the [special `Function` type][fn class]. This type is only
```
chalin: We've been following the convention of putting the period together with the link text so that the "external link" icon is after the period.

Come to think of it, we'd usually only link the API entry identifier, like this (@kwalrath can weigh in as well):

    signature refers to the special [Function][] type. This type ...

    [Function]: {{site.dart_api}}/{{site.data.pkg-vers.SDK.channel}}/dart-core/Function-class.html

In this way we don't have to worry about the period. Also note the formula to use for API entries (I've put a reminder below too where [fn class] is defined).

src/_guides/language/effective-dart/design.md
```diff
@@ -1005,90 +1091,262 @@ var names = people.map((Person person) {
...
+marginally more useful than using `dynamic`. If you're going to annotate, prefer
+a full function type that includes the parameters and return type of the
+function.
+
+[fn class]: https://api.dartlang.org/stable/dart-core/Function-class.html
```
chalin: This needs to be:

    [Function]: {{site.dart_api}}/{{site.data.pkg-vers.SDK.channel}}/dart-core/Function-class.html

src/_guides/language/effective-dart/design.md
chalin: Ok

src/_guides/language/effective-dart/design.md
chalin: I think that this example could benefit from some highlighting too: how about using

    <?code-excerpt "..." replace="/Function(\(.*?\))?/[!$&!]/g"?>

src/_guides/language/effective-dart/design.md
chalin: McCarthy conditionals FTW! ;)

src/_guides/language/effective-dart/design.md
kwalrath: "for good reasons" sounds more natural to me.

("for good reason" is the actual saying, I think, but it sounds formal.)

src/_guides/language/effective-dart/design.md
kwalrath: Agreed that [Function] is the way to go.

kwalrath: I recently wrote a wiki page on how to do API links: https://github.com/dart-lang/site-www/wiki/Referring-to-API-docs.

I should add examples of links in Markdown.

chalin: Nice!

kwalrath:
@munificent I think I'll wait until you have all of Patrice's changes in, and then take a final look. Feel free to @kwalrath me if you'd like me to weigh in on something.

kwalrath:
Just redeployed... I'd accidentally written on top of it.

src/_guides/language/effective-dart/design.md
kwalrath: I definitely like the highlights, but it's easy to miss what the importance of each example is. The comments seem like a nice way to highlight that.

chalin:
> Just redeployed... I'd accidentally written on top of it.

FYI, this isn't an issue for me since I've been generating the site locally.

src/_guides/language/effective-dart/design.md

src/_guides/language/effective-dart/design.md

src/_guides/language/effective-dart/design.md

src/_guides/language/effective-dart/design.md

src/_guides/language/effective-dart/design.md
chalin: How about "expression" rather than "code"?

chalin: Actually, nm, "code" is fine.

src/_guides/language/effective-dart/design.md
chalin: How about:
- **Inferred**: no type annotation was written down, but Dart successfully figured out the type on its own. ...

src/_guides/language/effective-dart/design.md
```diff
@@ -880,26 +880,47 @@ var buffer = new StringBuffer()
...
-Type annotations are important documentation for how a library should be used.
-Annotating the parameter and return types of public methods and functions helps
-users understand what the API expects and what it provides.
+*   If the code is "inferred", no type annotation was written down, and Dart
+    successfully figured out the type on its own. Inference can fail, in which
+    case the guidelines don't consider that "inferred". In some places,
+    inference failure is a static error. In others, Dart uses `dynamic` as the
+    fallback type.
+
+*   If the code is "dynamic", then its static type is the special `dynamic`
```
chalin: How about:
- **Dynamic**: the code is explicitly annotated or inferred as `dynamic`, or inference failed and Dart uses `dynamic` as the fallback type.

chalin:
Sorry but GH lost some of my edits so I retyped them as "single comments". Put all together what I was suggesting was something like this:

> To avoid that confusion, we avoid saying "untyped" and instead use:
> - **Type annotated**: a type was explicitly written in the code.
> - **Inferred**: no type annotation was written down, but Dart successfully figured out the type on its own. ...
> - **Dynamic**: the code is explicitly annotated or inferred as `dynamic`, or inference failed and Dart uses `dynamic` as the fallback type.

WDYT? /cc @kwalrath

src/_guides/language/effective-dart/design.md
munificent: Simplified it to try to clarify. The intent of this paragraph isn't really about when you should use the dynamic type, it's just to try to make it clear that this guideline is about annotation, not dynamism.

After this PR lands, I may try to add some more guidance around when to you use the dynamic type (annotated or not).

src/_guides/language/effective-dart/design.md
munificent: We don't have any guidelines around using `final` versus `var` for locals. I think more code I see in reviews uses `var` than `final`, but I'm not sure.

src/_guides/language/effective-dart/design.md
munificent: I like having a blank line after blocks. :)

src/_guides/language/effective-dart/design.md

src/_guides/language/effective-dart/design.md
chalin: @kevmoo might have an opinion on this (Kevin hadn't you suggested, at some point, that we needed a guideline about `final`)?

src/_guides/language/effective-dart/design.md
munificent: As above, I do think "fails" is an OK word.

src/_guides/language/effective-dart/design.md
munificent: 👍 Done.

src/_guides/language/effective-dart/design.md

src/_guides/language/effective-dart/design.md
munificent: Yeah, I like the highlights too. Killed the comments.

src/_guides/language/effective-dart/design.md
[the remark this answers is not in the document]
munificent: There are two cases where inference can give you `dynamic` as a result:

```dart
// 1. Successfully infer dynamic.
var list = <dynamic>["stuff"];
var i = list[0]; // <-- Infer "dynamic" for i based on type of list.

// 2. Fail to infer, default to dynamic.
var a; // <-- Fail to infer a and use dynamic.
```

It's important to keep these cases distinct because they represent different things and have different user impact. Also, the second case is an error when it happens at the top-level or when you have `--no-implicit-dynamic` enabled, but the first case isn't.

src/_guides/language/effective-dart/design.md
munificent: Oh, dear, you're right. This example doesn't make sense anymore. (It used to before we had generic methods.) Came up with a new example.

src/_guides/language/effective-dart/design.md

src/_guides/language/effective-dart/design.md

src/_guides/language/effective-dart/design.md
```diff
@@ -880,26 +880,47 @@ var buffer = new StringBuffer()
...
-update-for-dart-2.0
-{% endcomment %}
+The fact that Dart has both type inference and an actual `dynamic` type leads to
+some confusion about what it means to say code is "untyped". Does that mean the
+code is dynamically typed, or that you didn't *write* the type? To avoid that
+confusion, we avoid saying "untyped" and instead use:
 
-### DO type annotate public APIs.
+*   If the code is "type annotated", the type was explicitly written in the
+    code.
```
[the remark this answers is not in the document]
munificent: I kind introducing the terms in the context of a sentence because I think it might help readers understand how the terms are used grammatically.

src/_guides/language/effective-dart/design.md
```diff
@@ -880,26 +880,47 @@ var buffer = new StringBuffer()
...
+    case the guidelines don't consider that "inferred". In some places,
+    inference failure is a static error. In others, Dart uses `dynamic` as the
+    fallback type.
+
+*   If the code is "dynamic", then its static type is the special `dynamic`
+    type. Code can be explicitly annotated `dynamic` or it can be inferred.
+
+*   If the code has any other type, it's not "dynamic". Code can end up with a
+    non-`dynamic` type by being explicitly annotated with some other type, or by
+    having Dart infer a non-`dynamic` type.
```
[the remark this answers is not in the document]
munificent: It's not strictly necessary, but I'm trying to cover all four combinations of explicit/inferred and dynamic/other-type.

src/_guides/language/effective-dart/design.md
munificent: See above.

munificent:
Thanks for the detailed review! I think I got everything.

kwalrath:
All right, I'm looking now. 👀

kwalrath:
I've updated the staged version: https://kw-www-dartlang-1.firebaseapp.com/guides/language/effective-dart/design

src/_guides/language/effective-dart/design.md
```diff
@@ -1005,90 +1089,254 @@ var names = people.map((Person person) {
...
 
-{% comment %}
-update-for-dart-2.0
-https://github.com/dart-lang/site-www/pull/480#discussion_r157355593
-{% endcomment %}
+When you omit a type annotation, *most* of the time, Dart infers a correct
+static type for you. That word "most" is the alarming part. Sometimes, inference
+fails and silently uses `dynamic` instead. If you didn't realize that inference
+failed, you now have code that *looks* statically safe but actually has no
+type checking whatsoever.
```
kwalrath: This first paragraph seems to say the opposite of what the title says. It seems to be saying "annotate, or else you'll get dynamic and you'll be sorry". It becomes clearer later what you mean, but it's still a bit of a shock.

I like the words _That word “most” is the alarming part. Sometimes, inference fails and silently uses dynamic instead. If you didn’t realize that inference failed, you now have code that looks statically safe but actually has no type checking whatsoever._ but they don't seem to belong here.

I also worry about "even when not required" (in the heading) being ambiguous. It could be interpreted as "prefer annotating with dynamic over annotating with more specific types". (Also, isn't `dynamic` worse, performance-wise? It seems like we might want to mention that somewhere, or link to somewhere that discusses it.)

How about something like this instead:

### PREFER annotating with `dynamic` when you want no type checking.

When you omit a type annotation, Dart usually infers a correct static type for you. But a casual reader of your code who sees an annotation is missing has no way of knowing if you intended it to be dynamic, expected inference to fill in some other type, or simply forgot to write the annotation.

When `dynamic` is the type you want, using it explicitly makes your intent clear.

src/_guides/language/effective-dart/design.md
```diff
@@ -1005,90 +1089,254 @@ var names = people.map((Person person) {
...
+{:.bad-style}
+<?code-excerpt "misc/lib/effective_dart/design_bad.dart (prefer-dynamic)"?>
+{% prettify dart %}
+mergeJson(original, changes) => ...
+{% endprettify %}
+
+
+<aside class="alert alert-info" markdown="1">
+
+Before Dart 2, this guideline stated the exact opposite: *don't* annotate with
```
kwalrath: I earlier suggested having a **Changed guideline:** intro to notes like these. Do you not want to call out changes to guidelines? **New guideline** might also be a handy thing for people to search for. Or maybe you could just post a news/medium article about what's changed, and we can refer people to that.

src/_guides/language/effective-dart/design.md
```diff
@@ -1005,90 +1089,254 @@ var names = people.map((Person person) {
...
+### DON'T specify a return type for a setter.
 
-The `Function` type is barely more precise than using no annotation at all. If
-you're bothering to annotate, it's better to use a precise function type that
-describes the signature and return type of the function.
+Setters always return `void` in Dart. Writing the word is pointless.
 
-If you are annotating a field, this does mean you have to create a typedef, but
-that's usually worth doing.
+{:.bad-style}
```
kwalrath: Why isn't the good example before the bad example?

src/_guides/language/effective-dart/design.md
```diff
@@ -1005,90 +1089,254 @@ var names = people.map((Person person) {
...
+    typedef bool TestNumber(num);
+    {% endprettify %}
+
+    Most users expect this to be a function type that takes a `num` and returns
+    `bool`. It is actually a function type that takes *any* object (`dynamic`)
+    and returns `bool`. The parameter's *name* (which isn't used for anything
+    except documentation in the typedef) is "num". This has been a
+    long-standing source of errors in Dart.
+
+The new syntax looks like:
```
kwalrath: like -> like this

(The colon should be preceded by a sentence: https://developers.google.com/style/colons)

src/_guides/language/effective-dart/design.md
```diff
@@ -1005,90 +1089,254 @@ var names = people.map((Person person) {
...
+It may still be worth defining a typedef if the function type is particularly
+long or frequently used. But in most cases, users want to see what the function
+type actually is right where it's used, and the function type syntax gives them
+that clarity.
+
+
+### CONSIDER using the generalized function type syntax for parameters.
+
+Dart has a special syntax when defining a parameter whose type is a function.
+Sort of like C, you surround the parameter's name with the function's return
```
kwalrath: C -> in C

(otherwise you're saying that "you" are sort of like C)

src/_guides/language/effective-dart/design.md
```diff
@@ -1005,90 +1089,254 @@ var names = people.map((Person person) {
...
+type and parameter signature:
+
+<?code-excerpt "misc/lib/effective_dart/design_bad.dart (function-type-param)"?>
+{% prettify dart %}
+Iterable<T> where(bool predicate(T element)) => ...
+{% endprettify %}
+
+Before Dart 2 added general function type syntax, this was the only way to give
```
kwalrath: general -> the general
?
(I stopped and reread due to the "the" being gone. Maybe you meant that, though.)

src/_guides/language/effective-dart/design.md
```diff
@@ -1005,90 +1089,254 @@ var names = people.map((Person person) {
...
 
-Some operations will work with any possible object. For example, a log method
+Some operations work with any possible object. For example, a `log()` method
 could take any object and call `toString()` on it. Two types in Dart permit all
-objects: `Object` and `dynamic`. However, they convey two different things.
+values: `Object` and `dynamic`. However, they convey different things.
 
-The `Object` annotation says "I accept any object, and I only require it to have
-the methods that `Object` itself defines."
+Using `dynamic` sends a more complex signal. It may mean that Dart's type system
```
kwalrath: "more" -> "" or "somewhat"

(It's unclear what it's more complex than.)

src/_guides/language/effective-dart/design.md
```diff
@@ -1005,90 +1089,254 @@ var names = people.map((Person person) {
...
+isn't sophisticated enough to represent the set of types that are allowed, or
+that the values are coming from interop or otherwise outside of the purview of
+the static type system, or you explicitly want runtime dynamism at that point in
```
kwalrath: or you -> or that you
(for parallelism)

src/_guides/language/effective-dart/design.md
```diff
@@ -911,26 +911,47 @@ var buffer = new StringBuffer()
...
+*   If the code is *inferred*, no type annotation was written, and Dart
+    successfully figured out the type on its own. Inference can fail, in which
+    case the guidelines don't consider that inferred. In some places, inference
+    failure is a static error. In others, Dart uses `dynamic` as the fallback
+    type.
+
+*   If the code is *dynamic*, then its static type is the special `dynamic`
+    type. Code can be explicitly annotated `dynamic` or it can be inferred.
+
+*   If the code has any other type, it's not dynamic. Code can end up with a
```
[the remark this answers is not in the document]
munificent: OK, deleted.

src/_guides/language/effective-dart/design.md
```diff
@@ -983,13 +1004,76 @@ class CallChainVisitor {
...
 
-### AVOID annotating types on function expressions.
 
-The value of function expressions is their brevity. If a function is complex
-enough that types are needed to understand it, it should probably be a function
-statement or a method. Conversely, if it is short enough to be an expression, it
-likely doesn't need types.
+### AVOID annotating types for initialized local variables.
+
+Function bodies in modern code tend to be short, and the types of local
```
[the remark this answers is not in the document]
munificent: Changed to "idiomatic".

src/_guides/language/effective-dart/design.md
munificent: Took a different stab at improving this.

src/_guides/language/effective-dart/design.md
munificent: I don't generally want to call them out, no. Most readers of the guidelines probably don't care about its tumultuous history. I only called out this one because it is an exact inversion of what it previously stated. In most other cases, the guidelines are simply new or only slightly tweaked in intent.

src/_guides/language/effective-dart/design.md
munificent: I think the context for this comment may have been changed.

src/_guides/language/effective-dart/design.md

src/_guides/language/effective-dart/design.md
munificent: Done!

src/_guides/language/effective-dart/design.md
munificent: Done. I really wish there was a canonical name for this syntax. I made up "generalized function type syntax", but that's a mouthful. :-/

src/_guides/language/effective-dart/design.md
munificent: Moved the description of Object up, so it's clear what dynamic is more complex than.

src/_guides/language/effective-dart/design.md

src/_guides/language/effective-dart/design.md
```diff
@@ -1368,17 +1345,17 @@ that clarity.
 ### CONSIDER using the generalized function type syntax for parameters.
```
lrhn: I'd just say "function type syntax" or "inline function types".
The "generalized" makes no sense on its own, it's relative to something, and that something isn't clear here (if anywhere).

src/_guides/language/effective-dart/toc.md
```diff
@@ -206,12 +205,16 @@
 
 **Type annotations**
 
-* <a href='/guides/language/effective-dart/design#do-type-annotate-public-apis'>DO type annotate public APIs.</a>
+* <a href='/guides/language/effective-dart/design#do-type-annotate-public-declarations-whose-type-isnt-inferred'>DO type annotate public declarations whose type isn't inferred.</a>
```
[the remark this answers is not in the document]
munificent: > Are any public types inferred? Public types would all come from top-level inference, which doesn't insert types as such (right?).

I think type level inference can fill in types, can't it? You should know! 😀 

```dart
final i = 123;
```

Here, `i` gets inferred as `int` at the top level, right?

PR #745, opened 2018-04-04
Add guideline for placing multiple classes in a single file.

### LINKED ISSUE #36: specify file organization conventions in style guide
_From @nicolasgarnier on August 26, 2014 15:19_

_From [sigmund@google.com](https://code.google.com/u/sigmund@google.com/) on December 30, 2013 20:26:08_

consider adding to the styleguide suggestions for file organization.

For example:
- avoid 1 file per class: this is common in other languages,  but Dart encourages that you put together things that are related to each other.

_Original issue: http://code.google.com/p/dart/issues/detail?id=15836_

_Copied from original issue: dart-lang/www.dartlang.org#804_

munificent:
Fix #36.

### DIFF
--- a/src/_guides/language/effective-dart/design.md
+++ b/src/_guides/language/effective-dart/design.md
@@ -476,6 +476,25 @@ can delete dead code. It can't do that if the member is public because it
 doesn't know if any code outside of its view is using it.
 
 
+### CONSIDER declaring multiple classes in the same library.
+
+Java and C# tie the organization of files to the organization of classes — each
+file may only define a single top level class. Dart does not have that
+limitation. Libraries are distinct entities separate from classes. It's
+perfectly fine for a single library to contain multiple classes, top level
+variables, and functions if they all logically belong together.
+
+Placing multiple classes together in one library can enable some useful
+patterns. Since privacy in Dart works at the library level, not the class, this
+is a way to define "friend" classes like you might in C++. Every class declared
+in the same library can access each other's private members, but code outside of
+that library cannot.
+
+Of course, this guideline doesn't mean you *should* dump all of your classes
+into a huge monolithic library, just that you are allowed to place more than one
+class in a single library.
+
+
 ## Classes
 
 Dart is a "pure" object-oriented language in that all objects are instances of


src/_guides/language/effective-dart/design.md
```diff
@@ -476,6 +476,25 @@ can delete dead code. It can't do that if the member is public because it
 doesn't know if any code outside of its view is using it.
 
 
+### CONSIDER declaring multiple classes in the same library.
+
+Java and C# tie the organization of files to the organization of classes — each
```
chalin: > ... classes — each ...

I usually prefer using `&mdash;` since it is then clearer that it _is_ an mdash.

src/_guides/language/effective-dart/design.md
```diff
@@ -476,6 +476,25 @@ can delete dead code. It can't do that if the member is public because it
...
+file may only define a single top level class. Dart does not have that
+limitation. Libraries are distinct entities separate from classes. It's
+perfectly fine for a single library to contain multiple classes, top level
+variables, and functions if they all logically belong together.
+
+Placing multiple classes together in one library can enable some useful
+patterns. Since privacy in Dart works at the library level, not the class, this
```
chalin: > ... not the class <ins>level</ins>, ...

src/_guides/language/effective-dart/design.md
```diff
@@ -476,6 +476,25 @@ can delete dead code. It can't do that if the member is public because it
 doesn't know if any code outside of its view is using it.
 
 
+### CONSIDER declaring multiple classes in the same library.
```
chalin: How about:

> ### CONSIDER declaring _related_ classes in the same library.

src/_guides/language/effective-dart/design.md
```diff
@@ -476,6 +476,25 @@ can delete dead code. It can't do that if the member is public because it
...
+perfectly fine for a single library to contain multiple classes, top level
+variables, and functions if they all logically belong together.
+
+Placing multiple classes together in one library can enable some useful
+patterns. Since privacy in Dart works at the library level, not the class, this
+is a way to define "friend" classes like you might in C++. Every class declared
+in the same library can access each other's private members, but code outside of
+that library cannot.
+
+Of course, this guideline doesn't mean you *should* dump all of your classes
```
chalin: Dump seems a bit harsh; maybe use "put". Actually, is this last sentence really needed since "... just that you are allowed to place more than one" has already been made as a point (actually it is _the_ central point).

Like other guidelines, I suppose that we are just trying to remind the reader to be judicious in making use of this recommendation. I'm not sure how or even if that needs to be articulated here.

src/_guides/language/effective-dart/design.md
chalin: Actually, I was having a hard time thinking about an alternative title for this guideline. The content of this guideline seems to be more informative (along the lines of "by the way, you _can_ do this ...") than anything else. I suppose that one good reason to put more than one class in a library is because they are related. Another reason might be simply because the library is small (but then, I suppose, that the classes would then _still_ be related -- coincidental cohesion). Anyhow, kinda thinking out loud.

src/_guides/language/effective-dart/design.md
munificent: Done. Fixed the others too (in this file). Also consistently fixed them to not use spaces on either side. (I personally prefer thin spaces around em dashes because no spaces looks too tight to me, but maybe I'm a little too obsessed with this kind of minutia...)

src/_guides/language/effective-dart/design.md

src/_guides/language/effective-dart/design.md
munificent: Yeah, the guideline is a little weird. It technically carries no actual prescriptive content. Anything *not* prohibited by a guideline is permitted by default, so this is telling you nothing. But I've had a number of users believe there is some implicit that you *cannot* have more than one class per file — and assert that to other users in code reviews! — so I thought it worth adding a guideline to make this explicit.

src/_guides/language/effective-dart/design.md
munificent: Changed to "put". Agreed that it's weird, but <shrug>.

PR #753, opened 2018-04-05
[1.x] Dart language-version menu for switching between site versions

chalin:
Cherry picked #724 + some minor adjustments.

---

The dropdown looks like this:

> screen shot 2018-04-05 at 17 37 22

cc @kevmoo

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -1,6 +1,6 @@
 # Site settings
 title: Dart
-description: "Dart is a platform for building structured apps. It includes a language, a VM, libraries, tools, and compilers to many targets, including JavaScript."
+description: Dart is a platform for building structured apps. It includes a language, a VM, libraries, tools, and compilers to many targets, including JavaScript.
 url: https://www.dartlang.org
 repo: https://github.com/dart-lang/site-www
 branch: 1.x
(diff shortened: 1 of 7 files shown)


kevmoo:
Love it!

kwalrath:
It looks great! Let's hold off on this until @anders-sandholm tells us it's OK to switch. (If it is, then of course we'll need to commit #752 as well.)

anders-sandholm:
LGTM. 

Except for the URL in the address bar (and the dropdown), it is hard to tell whether you're looking at the 1.x or 2 version of the site. It might make sense to throw in a banner for the entire 1.x site saying something along the lines of "This page is showing old contents for Dart 1.24. Go to dartlang.org for the latest Dart 2 related contents."

chalin:
@anders-sandholm - I've added an "Add banner to 1x"  work item to #654 to track that.

chalin:
- #764 added the banner w/ conditional logic to ensure that it will show up only after the server switch.
- @kwalrath and @anders-sandholm have given a thumbs up, so I'm merging this. Like for the banner, the menu will only appear after the switch.

PR #760, opened 2018-04-06
[WIP] chore(sharing): use site-shared git submodule

chalin:
Contributes to https://github.com/dart-lang/site-webdev/issues/637


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "site-shared"]
```
filiph: I had a long comment here about different ways to name this repo, but I think `site-shared` is the best possible name.

README.md
```diff
@@ -24,13 +24,19 @@ Install the following tools if you don't have them already.
...
-1. **Clone this repo** ([site-www][]) into the chosen directory by following
-   the instructions given in the GitHub help on [Cloning a repository][].
+> NOTE: This repo has a git _submodule_, which affects how you clone it.
+
+To **clone this repo** ([site-www][]), follow the instructions given in the
+GitHub help on [Cloning a repository][], and _choose one_ of the following
+submodule-cloning techniques:
+
+- Clone this repo and its submodule _at the same_: use the
+  `--recurse-submodules` option with the git clone command for this repo.
```
filiph: Let's add the full command here, both as an example and as a quick default. It's unlikely that we'll change the path of where the repo lives anytime soon.

_config.yml
```diff
@@ -99,10 +99,11 @@ custom:
 
 # Build settings
 source: src
+symlinked-sources: [site-shared/src] # [src/tools/sdk] also works
```
filiph: I don't understand the comment here.

README.md
chalin: Done

_config.yml
chalin: Done

PR #761, opened 2018-04-06
Update library-tour.md

lucyget:
Fixing "Implementing map keys" example so that two fields aren't declared on the same line.

### DIFF
--- a/examples/misc/lib/library_tour/core/hash_code.dart
+++ b/examples/misc/lib/library_tour/core/hash_code.dart
@@ -1,7 +1,8 @@
 // ignore_for_file: unrelated_type_equality_checks
 // #docregion
 class Person {
-  final String firstName, lastName;
+  final String firstName;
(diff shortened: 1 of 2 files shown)


kwalrath:
Good catch, @lucyget! Could you also update the example file, https://github.com/dart-lang/site-www/blob/master/examples/misc/lib/library_tour/core/hash_code.dart? (The build will fail unless the two match.)

chalin:
The original intent was to keep the code compact since those excerpts are [long enough](https://dartlang-org-dev.firebaseapp.com/guides/libraries/library-tour#implementing-map-keys) as it is (34 lines!). IMHO, having those two field declarations on the same line is fine.

kwalrath:
Ah OK. I thought we had a guideline about not putting stuff on one line, but I can't find it. I'll close this, but thanks very much @lucyget for making the effort!

PR #783, opened 2018-04-11
Add a bit more info to PR template

kwalrath:


kwalrath:
I'm a little afraid that some folks will end up on 1.x for some reason, and encounter this problem. But you're right that we probably don't have to worry about that...

Let's leave it up just a bit longer.

PR #808, opened 2018-04-20
More "Effective Dart" guidelines.

munificent:
- Don't return FutureOr.
- Use whereType().
- Don't do unneeded cast() or retype().

@lrhn, this includes one of your suggested guidelines. I'm still working out what to do for `cast()` and `retype()`. I'll do that in a different PR.


kwalrath:
All Travis builds are failing today, apparently due to the switch to 2.0.0-dev.49.0.

chalin:
Rebasing, and running [`git submodule update --init --remote`](https://github.com/dart-lang/site-webdev/wiki/Git-submodule-notes#pull-upstream-changes) should fix the issues.

src/_guides/language/effective-dart/design.md
```diff
@@ -1398,10 +1398,64 @@ bool convertToBool(dynamic arg) {
...
+[contravariant][] positions.* Parameters are contravariant and return types are
+covariant. In nested function types, this gets flipped&mdash;if you have a
+parameter whose type is itself a function, then the callback's return type is
+now in contravariant position and the callback's parameters are covariant. This
+means it's OK for a *callback's* type to return `FutureOr<T>`:
+
+[contravariant]: https://en.wikipedia.org/wiki/Covariance_and_contravariance_(computer_science)
+
+{:.good-style}
+<?code-excerpt "misc/lib/effective_dart/design_good.dart (future-or-contra)" replace="/FutureOr&lt;S&gt;/[!$&!]/g"?>
```
chalin: In the `replace` argument, instead of `&lt;S&gt;` use, e.g.:
- `<S\x3E`
- `.S.`
- or the following: 
\<?code-excerpt "misc/..." replace="/FutureOr<b><ins>\S+</ins></b>/[!$&!]/g"?>

src/_guides/language/effective-dart/design.md
```diff
@@ -1468,11 +1523,12 @@ Duration(
     int microseconds: 0});
 {% endprettify %}
 
-### AVOID mandatory parameters that permit nonce values.
+
+### AVOID mandatory parameters that permit sentinel values.
```
chalin: Did you mean "special" here too? (I've used "sentinel" in the context of collections search, but in this context it seems strange.)

src/_guides/language/effective-dart/usage.md
```diff
@@ -352,6 +352,83 @@ var ints = new List<int>.from(numbers);
...
+indirection and redundant runtime checking. Fortunately, the core library has a
+method for this exact use case:
+
+{:.good-style}
+<?code-excerpt "misc/lib/effective_dart/usage_good.dart (where-type)"?>
+{% prettify dart %}
+var objects = [1, "a", 2, "b", 3];
+var ints = objects.whereType<int>();
+{% endprettify %}
+
```
chalin: A few comments about this section:

1. I would add the three above excerpts to the tests file, and add a (hidden) test to ensure that the runtime type is `Iterable<Object>`, `Iterable<int>` and `Iterable<int>`, respectively.
2. Consider renaming `objects` to `iterable` (which was the name that you used in other examples involving iterables).
3. Consider linking to the API entry for `whereType()`:
> Fortunately, the core library has the [whereType()][] method for this exact use case:

    [whereType()]: {{site.dart_api}}/{{site.data.pkg-vers.SDK.channel}}/dart-core/Iterable/whereType.html

src/_guides/language/effective-dart/usage.md
```diff
@@ -352,6 +352,83 @@ var ints = new List<int>.from(numbers);
...
+This is terse, produces an Iterable of the desired type, and has no unnecessary
```
chalin: Iterable --> `Iterable` or [Iterable][]

src/_guides/language/effective-dart/usage.md
```diff
@@ -352,6 +352,83 @@ var ints = new List<int>.from(numbers);
...
+levels of wrapping.
+
+
+### DON'T use `cast()` or `retype()` if you are already performing an operation that can change the type.
+
+Often, a number of transformations on Iterables or Streams are combined. At the end, you want to produce an object with a certain type argument. Instead of tacking on a call to `cast()` or `retype()`, see if one of the existing operations can change the type.
```
chalin: on Iterables or Streams are --> on iterables or streams are

src/_guides/language/effective-dart/usage.md
```diff
@@ -352,6 +352,83 @@ var ints = new List<int>.from(numbers);
...
+
+If you are already calling `toList()`, replace that with a call to `List.from<T>()` where `T` is the type of resulting list you want.
```
chalin: Linking to the API entry for `List.from<T>()` would be nice:

    [List.from<T>()]: {{site.dart_api}}/{{site.data.pkg-vers.SDK.channel}}/dart-core/List/List.from.html

src/_guides/language/effective-dart/usage.md
```diff
@@ -352,6 +352,83 @@ var ints = new List<int>.from(numbers);
...
+<?code-excerpt "misc/lib/effective_dart/usage_bad.dart (cast-list)"?>
+{% prettify dart %}
+var stuff = <dynamic>[1, 2];
+var ints = stuff.toList().cast<int>();
+{% endprettify %}
+
+If you are calling `map()`, give it an explicit type argument so that it produces an iterable of the desired type. Type inference will often pick the correct type for you based on the function you pass to `map()`, but sometimes you need to be explicit.
+
+{:.good-style}
+<?code-excerpt "misc/lib/effective_dart/usage_good.dart (cast-map)" replace="/\(n as int\)/n/g"?>
```
chalin: Why `replace="/\(n as int\)/n/g"`?

src/_guides/language/effective-dart/usage.md
```diff
@@ -352,6 +352,83 @@ var ints = new List<int>.from(numbers);
...
+{% prettify dart %}
+var stuff = <dynamic>[1, 2];
+var reciprocals = stuff.map<double>((n) => 1 / n);
+{% endprettify %}
+
+{:.bad-style}
+<?code-excerpt "misc/lib/effective_dart/usage_bad.dart (cast-map)" replace="/\(n as int\)/n/g"?>
```
chalin: Ditto, why `replace="/\(n as int\)/n/g"` as opposed to showing the actual code?

src/_guides/language/effective-dart/usage.md
```diff
@@ -1156,6 +1233,54 @@ Future<bool> fileContainsBear(String path) async {
...
+However, if the value type is `Object` or a type parameter that could possibly
+be instantiated with `Object` then the two types overlap. `Future<Object>`
+itself implements `Object`, so `is Object` or `is T` where `T` is some type
+parameter that could be instantiated with object returns true even when the
+object is a future. Instead, explicitly test for the `Future` case:
+
+{:.good-style}
+<?code-excerpt "misc/lib/effective_dart/usage_good.dart (test-future-or)"?>
+{% prettify dart %}
+Future<T> logValue<T>(FutureOr<T> value) async {
```
chalin: To help keep some focus in this example, how about having `logValue()` only log rather than return a value? I.e., let the return type be `Future<void>`, and dropping the return statements?

src/_guides/language/effective-dart/usage.md
```diff
@@ -1156,6 +1233,54 @@ Future<bool> fileContainsBear(String path) async {
...
+  if (value is T) {
+    print(value);
+    return value;
+  } else {
+    var result = await value;
+    print(result);
+    return result;
+  }
+}
+{% endprettify %}
```
chalin: How about eliding the if-then-else bodies? We don't need them to make the point that testing against `value is T` is a bad idea.

src/_guides/language/effective-dart/usage.md
```diff
@@ -1156,6 +1233,54 @@ Future<bool> fileContainsBear(String path) async {
...
+### DO test for a future when disambiguating a `FutureOr<T>` whose value type could be `Object`.
+
+Before you can do anything useful with a `FutureOr<T>`, you typically need to do
+an `is` check to see if you have a `Future<T>` or a bare `T`. If the type
+argument is some specific type like `FutureOr<int>`, it doesn't matter which
+test you use, `is int` or `is Future<int>`. Either works because those two types
+are disjoint.
+
+However, if the value type is `Object` or a type parameter that could possibly
+be instantiated with `Object` then the two types overlap. `Future<Object>`
```
kwalrath: then -> , then

src/_guides/language/effective-dart/usage.md
```diff
@@ -1156,6 +1233,54 @@ Future<bool> fileContainsBear(String path) async {
...
+itself implements `Object`, so `is Object` or `is T` where `T` is some type
+parameter that could be instantiated with object returns true even when the
```
kwalrath: with object -> with `Object`

src/_guides/language/effective-dart/design.md

src/_guides/language/effective-dart/design.md
munificent: I meant "sentinel", but it's not a perfect match. It's the best word I could come up with: https://en.wikipedia.org/wiki/Sentinel_value. "Nonce" isn't right.

src/_guides/language/effective-dart/usage.md
munificent: > 1. I would add the three above excerpts to the tests file,

`whereType()` isn't actually fully supported yet (it doesn't work in Dart 1 mode). Is it OK if I don't worry about tests for this now. (For what it's worth, I find tests to be overkill for these tiny excerpts in Effective Dart. It takes much longer for me to make changes now and the tests and analysis have caught few issues for me.)

> 2. Consider renaming objects to iterable (which was the name that you used in other examples involving iterables).

I used the more specific names here because there are two iterables with different types in the excerpt.

> 3. Consider linking to the API entry for whereType():

Done.

src/_guides/language/effective-dart/usage.md

src/_guides/language/effective-dart/usage.md

src/_guides/language/effective-dart/usage.md

src/_guides/language/effective-dart/usage.md
munificent: The replaced code is Dart 2 compliant. But the analyze script has "no implicit dynamic" on, I think.

src/_guides/language/effective-dart/usage.md
munificent: See above.

src/_guides/language/effective-dart/usage.md
```diff
@@ -1156,6 +1233,54 @@ Future<bool> fileContainsBear(String path) async {
 }
 {% endprettify %}
 
+
+### DO test for a future when disambiguating a `FutureOr<T>` whose value type could be `Object`.
```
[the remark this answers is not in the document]
munificent: That works, but it might be overly specific. If you have a `FutureOr<int>`, it's perfectly fine to test for `is int` and might give you cleaner code do to that.

src/_guides/language/effective-dart/usage.md
```diff
@@ -1156,6 +1233,54 @@ Future<bool> fileContainsBear(String path) async {
...
+
+Before you can do anything useful with a `FutureOr<T>`, you typically need to do
+an `is` check to see if you have a `Future<T>` or a bare `T`. If the type
+argument is some specific type like `FutureOr<int>`, it doesn't matter which
+test you use, `is int` or `is Future<int>`. Either works because those two types
+are disjoint.
```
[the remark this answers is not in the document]
munificent: Reworded.

> Isn't it sufficient to test against Future instead of Future<T>?

Yes, but I think it gives you worse type promotion.

src/_guides/language/effective-dart/usage.md
```diff
@@ -1156,6 +1233,54 @@ Future<bool> fileContainsBear(String path) async {
...
+
+However, if the value type is `Object` or a type parameter that could possibly
+be instantiated with `Object` then the two types overlap. `Future<Object>`
+itself implements `Object`, so `is Object` or `is T` where `T` is some type
+parameter that could be instantiated with object returns true even when the
+object is a future. Instead, explicitly test for the `Future` case:
```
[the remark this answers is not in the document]
munificent: I'm not crazy about how wordy this paragraph is either. (I'm on the fence about whether it's worth even having this guideline), but I think your suggestion is too brief. The challenging case — which is really why this guideline is useful, because the case is non-obvious — is:

```
foo<T>(FutureOr<T> f) {
  if (f is T) {
    // Assume f is not a future...
  } else {
    // Assume f is a future...
  }
}
```

This code looks correct, doesn't mention `Object`, and does work correctly for many type arguments. But if you happen to call it with `Object` as the generic method type argument, it falls over.

The prose and example try to show this.

src/_guides/language/effective-dart/usage.md
```diff
@@ -1156,6 +1233,54 @@ Future<bool> fileContainsBear(String path) async {
...
+
+{:.good-style}
+<?code-excerpt "misc/lib/effective_dart/usage_good.dart (test-future-or)"?>
+{% prettify dart %}
+Future<T> logValue<T>(FutureOr<T> value) async {
+  if (value is Future<T>) {
```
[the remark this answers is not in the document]
munificent: See above. I *want* this to be a generic method, because that's the case where `FutureOr<T>` is easy to mis-use.

src/_guides/language/effective-dart/usage.md
munificent: See above.

src/_guides/language/effective-dart/usage.md
```diff
@@ -352,6 +352,83 @@ var ints = new List<int>.from(numbers);
 But if your goal is just to copy the iterable and preserve its original type, or
 you don't care about the type, then use `toList()`.
 
+
+### DO use `whereType()` to filter a collection by type.
+
+Let's say you have a list containing a mixture of objects and you want to get just the integers out of it. You could do:
```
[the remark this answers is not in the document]

src/_guides/language/effective-dart/usage.md
```diff
@@ -352,6 +352,83 @@ var ints = new List<int>.from(numbers);
...
+{% prettify dart %}
+var objects = [1, "a", 2, "b", 3];
+var ints = objects.where((e) => e is int);
+{% endprettify %}
+
+This is verbose, but, worse, it returns an Iterable whose type probably isn't
+what you want. In the example here, it returns an `Iterable<Object>` even though
+you likely want an `Iterable<int>` since that's the type you're filtering it to.
+
+Sometimes you see code that "corrects" the above error by doing:
```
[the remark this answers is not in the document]

src/_guides/language/effective-dart/usage.md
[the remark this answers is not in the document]
munificent: I like "concise" too. If programmers were put off by the negative connotations of "terse", they probably wouldn't be so terse themselves... 😉

src/_guides/language/effective-dart/usage.md
```diff
@@ -352,6 +352,83 @@ var ints = new List<int>.from(numbers);
...
+
+{:.bad-style}
+<?code-excerpt "misc/lib/effective_dart/usage_bad.dart (where-type-2)"?>
+{% prettify dart %}
+var objects = [1, "a", 2, "b", 3];
+var ints = objects.where((e) => e is int).cast<int>();
+{% endprettify %}
+
+That's verbose, and causes two wrappers to be created, with two layers of
```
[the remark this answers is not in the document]

src/_guides/language/effective-dart/usage.md
```diff
@@ -352,6 +352,83 @@ var ints = new List<int>.from(numbers);
...
+{% prettify dart %}
+var objects = [1, "a", 2, "b", 3];
+var ints = objects.whereType<int>();
+{% endprettify %}
+
+This is terse, produces an Iterable of the desired type, and has no unnecessary
+levels of wrapping.
+
+
+### DON'T use `cast()` or `retype()` if you are already performing an operation that can change the type.
```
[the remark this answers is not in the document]
munificent: I try to avoid "unnecessarily" in guidelines because it's sort of a tautology that relies on the prose to actually define it:

**DON'T do things that are forbidden by this guideline.**

😕

Came up with a better way to reword this.

src/_guides/language/effective-dart/usage.md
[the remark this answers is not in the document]
munificent: Done^3.

src/_guides/language/effective-dart/usage.md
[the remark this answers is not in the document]
munificent: D'ne.

src/_guides/language/effective-dart/usage.md
```diff
@@ -352,6 +352,83 @@ var ints = new List<int>.from(numbers);
...
+{% endprettify %}
+
+{:.bad-style}
+<?code-excerpt "misc/lib/effective_dart/usage_bad.dart (cast-list)"?>
+{% prettify dart %}
+var stuff = <dynamic>[1, 2];
+var ints = stuff.toList().cast<int>();
+{% endprettify %}
+
+If you are calling `map()`, give it an explicit type argument so that it produces an iterable of the desired type. Type inference will often pick the correct type for you based on the function you pass to `map()`, but sometimes you need to be explicit.
```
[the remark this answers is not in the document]
munificent: Does.

src/_guides/language/effective-dart/usage.md

src/_guides/language/effective-dart/usage.md

munificent:
Thanks!

kwalrath:
Taking another look... This is staged at https://kw-www-dartlang-1.firebaseapp.com/guides/language/effective-dart.

src/_guides/language/effective-dart/design.md
chalin: From Wikipedia (emphasis mine):

> **sentinel value** is a special value in the context of an algorithm which uses its presence as a _condition of termination_, typically in a loop or recursive algorithm

The value isn't being used "as a condition of termination".

Maybe the title needs to be reworked:

### AVOID implementing optional parameters as mandatory parameters

The opening paragraph then clearly explains what is meant by this.

@kwalrath - any thoughts on this too?

src/_guides/language/effective-dart/usage.md
chalin: > Is it OK if I don't worry about tests for this now

Surely.

src/_guides/language/effective-dart/usage.md
chalin: Hmm, really. Ok, I'll check that later.

src/_guides/language/effective-dart/design.md
munificent: I think that wording is a little too clever and likely to confuse. Changed it to:

> AVOID mandatory parameters that accept a special "no argument" value.

PR #811, opened 2018-04-22
Update AngularDart link and description

Meronat:
The old link was to the old repo location, so I changed it to the webdev link commonly used elsewhere. I also modified the description to the description presented on webdev as AngularDart has become quite distinct from the JS Angular and it seems other mentions have moved away from the idea of it being a "port" anyway.


kwalrath:
Much better. Thanks for fixing this!

PR #818, opened 2018-04-26
Update windows instructions (Chocolatey)

kwalrath:


kwalrath:
@munificent one of your samples changed when dartfmt ran, so I updated it. If you don't like the new formatting, we can fix it another way (later).

munificent:
Oops, I must have forgotten to format. Too many scripts to remember to run. Thanks for fixing it!

PR #826, opened 2018-05-01
chore: env-setup from shared; fix links

chalin:
- Use env-setup from the refreshed site-shared.
- Add link checks to Travis using shared link-check script. Travis is set to fail if there are invalid links.
- Fix invalid anchors.


chalin:
[This Travis job is failing](https://travis-ci.org/dart-lang/site-www/jobs/373482078), as it should, showing that link checking using @filiph's link checker is working:
```nocode
http://localhost:4000/guides/language/language-tour
- (1400:0) 'type ann..' => http://localhost:4000/guides/language/effective-dart/design#do-type-annotate-public-apis (HTTP 200 but missing anchor)
```

PR #828, opened 2018-05-01
Add guidelines for optional new and const.

munificent:


chalin:
Nice! LGTM, but I'll let Kathy do the official review.

src/_guides/language/effective-dart/usage.md
```diff
@@ -1013,6 +1013,37 @@ View(Style style, List children)
...
+actually return a new object.
+
+The language still permits `new` in order to make migration less painful, but
+consider it deprecated and remove it from your code.
+
+
+### DON'T use `const` when not needed.
+
+In contexts where only constant expressions are allowed, the `const` keyword is
+implicit, doesn't need to be written, and shouldn't. Those contexts are:
```
kwalrath: This little in-sentence list isn't parallel and stopped me cold (shouldn't what? I knew after a second, but it still stopped me). Also, the intro to the bulleted list should be a sentence. How about:

...is implicit. Don't use `const` in the following constant contexts:

src/_guides/language/effective-dart/usage.md
```diff
@@ -1013,6 +1013,37 @@ View(Style style, List children)
...
+
+* An element expression in a const collection literal.
+* An argument expression in a const constructor call or metadata annotation.
+* The initializer for a const variable declaration.
+* A switch case expression&mdash;the part right after `case` before the `:`, not
+  the body of the case.
+
+(Default values are *not* included in this list because future versions of Dart
+may support non-const default values. So you still need to explicit write
```
kwalrath: explicit write -> explicitly use

(This usage of "write" sounds wrong to me.)

src/_guides/language/effective-dart/usage.md
```diff
@@ -1013,6 +1013,37 @@ View(Style style, List children)
...
+`const` on the outermost expression there so that it doesn't become a breaking
```
kwalrath: The second sentence is a little hard to read. Maybe:

Explicitly using `const` on the outermost expressions of default values helps you avoid possible breaking changes later.

(Not sure that's easier to read, but... more precise, at least.)

src/_guides/language/effective-dart/usage.md
```diff
@@ -1013,6 +1013,37 @@ View(Style style, List children)
...
+change to later support non-const expressions in that context.)
+
+Basically, any place where it would be a static error to write `new` instead of
```
kwalrath: I'm not sure what "static error" means. Do you mean where the analyzer would tell you it's an error?

src/_guides/language/effective-dart/usage.md

src/_guides/language/effective-dart/usage.md
munificent: Removed the second sentence entirely because it's probably not that relevant. (What it's supposed to mean is that requiring *you* to write const on default values means that *the language team* can change Dart to support non-const default values without that becoming a breaking language change.)

src/_guides/language/effective-dart/usage.md
munificent: Yes. I use that term in a couple of other places. "Compile error" would sort of work too, but I feel that's too focused on a specific action (compilation) when really what we mean is "errors that can be detected without having to run the code".

munificent:
@kwalrath, friendly ping.

Issue #651, opened 2018-03-13
Update Dart icon, website color scheme, and design 


anders-sandholm:
@filiph, once you've had a chance to land the changes needed on flutter.io, it would great if you would help us get dartlang up to snuff as well on the master branch.

---

Edited by @chalin:

- [x] Logos used in site pages.
- [x] Site favicon.ico
- [x] Site social media shared images: Use new logo as og_image #832 
- [x] Examples: any use of logos, including favicon.
- [x] Delete unused `src/twitter` folder and the images it contains? Cf. https://github.com/dart-lang/site-www/issues/651#issuecomment-386271055

filiph:
Okay. We're not ready for the Flutter change yet, but it will land sometime between now and I/O. Hope that's okay.

JekCharlsonYu:
Can we confirm that this includes the following sites


https://pub.dartlang.org/
https://www.dartlang.org/
https://webdev.dartlang.org/

chalin:
@filiph - if some of the resources are ready, then I can put them in the site-shared repo as a first step.

kwalrath:
Another possibility is to put stand-ins for the final resources in the site-shared repo, so we can migrate to shared assets before the final ones are ready.

JekCharlsonYu:
Flutter Logo change github issue:  https://github.com/flutter/website/pull/953

chalin:
You can see below an example of what I have so far. The following changes have been made

- Font-family base matches Flutter site.
- Slight change in colors to match Flutter palette
- This uses the new logos (including the Angular logo that I reworked/colorized).

### New

<img width="1155" alt="screen shot 2018-04-25 at 17 57 24" src="https://user-images.githubusercontent.com/4140793/39275143-2dce7bf8-48b2-11e8-9fd2-d99328ba5449.png">

### Old

<img width="1157" alt="screen shot 2018-04-25 at 17 58 19" src="https://user-images.githubusercontent.com/4140793/39275167-4ad560c2-48b2-11e8-997b-4a6cd9b3e9a3.png">

chalin:
@kwalrath, @Sfshaza - AFAICT, the images under https://github.com/dart-lang/site-www/tree/master/twitter aren't used by site-www. The logo in that folder is obsolete, so I'd like to remove the entire folder, is that ok with you?

Sfshaza:
Sounds good to me.

kwalrath:
I don't remember why we ever had those images.

[...]

Issue #637, opened 2018-03-07
[Meta] Site changes for Dart 2


chalin:
We're using the [2-beta2 milestone](https://github.com/dart-lang/site-www/milestone/5) to track the most urgent Dart 2 issues, but here's a summary of the pages on the site and what might need to change. (Less urgent changes are under the [2-beta3 milestone](https://github.com/dart-lang/site-www/milestone/6).)

**Note:** The 2-dev version of the website is staged at https://dartlang-org-dev.firebaseapp.com. The current (1.x) version of the website will live on (and is hand-staged) at https://v1-dartlang-org.firebaseapp.com.

- [x] [Homepage](https://www.dartlang.org) - #653, #693 @anders-sandholm with support from @kwalrath (PR: #665, #750)
- [x] Language
  - [x] [Overview](https://www.dartlang.org/guides/language) - @kwalrath (minor)
  - [x] [Tour](https://www.dartlang.org/guides/language/language-tour) - #507 @kwalrath (2-beta2 changes done; more to come for beta3)
  - [x] [Effective Dart](https://www.dartlang.org/guides/language/effective-dart) and subpages - #632 @munificent (for now, @kwalrath has added a "Not Dart 2 compatible yet" banner: #688)
  - [x] [Sample Code](https://www.dartlang.org/samples) - #655 @kwalrath (dropped the cookbook in #687, should maybe change samples, but that could wait for beta3)
  - [x] "Strong Mode Dart" pages all need updates to change "strong mode" back to "sound Dart" (see [this diff](https://github.com/dart-lang/site-www/commit/6789bdf0725cddb7eb766b3637a7c3bc3ab1dc3c#diff-db463bf6e99ff3398977e0289f2026c7)), reflect the latest tool status, fix [known issues](https://github.com/dart-lang/site-www/issues?q=is%3Aissue+is%3Aopen+strong+mode+label%3ASoundDart) - @kwalrath (PR: #694)
    https://www.dartlang.org/guides/language/sound-dart
    https://www.dartlang.org/guides/language/sound-faq
    https://www.dartlang.org/guides/language/sound-problems
  - [x] https://www.dartlang.org/guides/language/analysis-options - #689 @kwalrath (get rid of "Specifying strong mode", 1.x, ...) (PR: #694)
  - [x] https://www.dartlang.org/guides/language/spec - OK as is?
- [x] Libraries
  - [x] [Overview](https://www.dartlang.org/guides/libraries) - OK as is.
[...]

PR #830, opened 2018-05-02
Refreshed look with new logos

chalin:
Updated:
- Logos, based on the new official Dart and Flutter logos.
- Some color adjustments
- Switched to using Roboto font, which is what the current Flutter site uses.

This PR doesn't include any frontpage redesign. That will be addressed separately (if at all for this site).

Contributes to #651 and #637

cc @anders-sandholm @JekCharlsonYu 

---

- [x] Logos used in site pages.
- [x] favicon.ico
- [x] Social media shared images: Use new logo as og_image #832 
- [x] Ensure submodule branch was merged and that this PR points to the updated master.
- [x] Examples: any use of logos, including favicon.
- [x] Delete unused `src/twitter` folder and the images it contains? Cf. https://github.com/dart-lang/site-www/issues/651#issuecomment-386271055:
  - Except for (currently blocked) followup item: #833.


chalin:
I used ImageMagick to create the favicon.ico, e.g.:

    convert angular/logo/square.png -define icon:auto-resize=128,64,48,32,16 angular/favicon.ico

chalin:
Staging server refreshed (to include the updated favicon.ico).

chalin:
Staging server refreshed with new social media sharing image, which (according to the Facebook Sharing Debugger), will appear like this:

screen shot 2018-05-02 at 17 51 36

anders-sandholm:
I really like what I see. Thanks Patrice for getting all these beautiful changes lined up.
Unless there are obvious parts missing, let's get this landed.
Then we can always iterate further in the coming days.

chalin:
The final submodule sync commit is in.
@kwalrath - awaiting for your official thumbs up.

chalin:
I've scanned the entire site for any images that might contain old logos. All that remains is to update DartPad screenshots; see #833.

src/_assets/stylesheets/_variables.scss
```diff
@@ -16,14 +16,14 @@ $blue-light:                #e2f4fd;
 $brand-primary:             #1155cc;
 
 // Blue color taken from gmail's stylesheet.
-$default-link:              #1155cc;
-$default-link-visited:      #1155cc;
+$default-link:              #0175C2;
```
kwalrath: The comment above says this blue color is taken from gmail's stylesheet. Is that really true?

src/_assets/stylesheets/_variables.scss
```diff
@@ -16,14 +16,14 @@ $blue-light:                #e2f4fd;
...
+$default-link-visited:      #0175C2;
 
 // Typeography
-$font-family-base:          'Source Sans Pro', sans-serif;
-$font-family-condensed:     'Trade Gothic W01', sans-serif;
-$font-family-extended:      'Trade Gothic W01', sans-serif;
-$font-family-serif:         'Trade Gothic W01', sans-serif;
+$font-family-base:          Roboto, sans-serif;
```
kwalrath: While you're in here... Typeography -> Typography
:)

src/_includes/page-footer.html
```diff
@@ -4,7 +4,7 @@
       <div class="col-sm-12 col-md-3">
         <div class="content">
           <div class="brand">
-            <img src="{% asset_path 'logo-white.svg' %}" alt="Dart logo" class="brand"/>
+            <img src="{% asset_path 'shared/dart/logo+text/horizontal/white.svg' %}" alt="Dart logo" class="brand"/>
```
kwalrath: This logo seems less visible against the black footer than the old one, but I think we can fix this later (if at all).

kwalrath:
I just noticed that the font is heavier than what I see on flutter.io. E.g. look at https://dartlang-org-staging-0.firebaseapp.com/guides/get-started vs. https://flutter.io/setup-macos/. It looks different from https://developers.google.com/style/, too (header font is different, and it looks like d.g.c uses an equally heavy but grayer font, which is pretty but questionable for accessibility reasons).

I'm OK with these font diffs — just pointing them out, in case anyone cares.

src/_assets/stylesheets/_variables.scss
chalin: Good catch. Removed.

chalin:
Yes, the main font is scaled on flutter.io, but both sites use 16pt as the font base. I haven't found where in the flutter.io styling that they do the scaling. As you mention, we can address that later as necessary.

src/_assets/stylesheets/_variables.scss
chalin: Oops, I just noticed that I forgot to change `$brand-primary` to also be #0175C2. I've done that now. I've refreshed the staging server. PTAL (the color is very similar).

chalin: (You'll likely have to empty the browser cache for the new CSS to load.)

PR #835, opened 2018-05-03
Drop favicon.ico in favor of a single icon.png

### LINKED ISSUE #834: Drop favicon.ico in favor of a single icon.png
This is what is done on flutter.io: https://github.com/flutter/website/blob/master/_includes/head.html#L6

```html
<link rel="shortcut icon" href="{{ "/images/favicon.png" | prepend: site.baseurl  }}">
```

chalin:
Fixes #834 

cc @devoncarew

### DIFF
--- a/src/_includes/head.html
+++ b/src/_includes/head.html
@@ -9,7 +9,7 @@
   <title>{% if page.short-title %}{{ page.short-title }}{% else %}{{ page.title }}{% endif %} | {{ site.title }}</title>
 
   <!-- Favicon / Touch Icons -->
-  <link href="/favicon.ico" rel="shortcut icon">
+  <link rel="icon" sizes="64x64" href="/assets/shared/dart/icon/64.png">
   <link href="{% asset_path 'touch-icon-iphone.png' %}" rel="apple-touch-icon">
   <link href="{% asset_path 'touch-icon-ipad.png' %}" rel="apple-touch-icon" sizes="76x76">
   <link href="{% asset_path 'touch-icon-iphone-retina.png' %}" rel="apple-touch-icon" sizes="120x120">


chalin:
The build is broken, but because of issues with Effective Dart. These will be addressed separately.

src/_includes/head.html
```diff
@@ -9,7 +9,7 @@
   <title>{% if page.short-title %}{{ page.short-title }}{% else %}{{ page.title }}{% endif %} | {{ site.title }}</title>
 
   <!-- Favicon / Touch Icons -->
-  <link href="/favicon.ico" rel="shortcut icon">
+  <link rel="icon" sizes="64x64" href="/assets/shared/dart/icon/64.png">
```
devoncarew: Shoulld this not still be `rel="shortcut icon"`?

chalin: Thanks for the feedback. From my readings, "shortcut" shouldn't be used anymore.

From [MDN Link types](https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types):

> The `shortcut` link type is often seen before `icon`, but this link type is **non-conforming, ignored and web authors must not use it anymore**.

From the [HTML Living Standard — Last Updated 2 May 2018 >> Link types >> `rel=icon`](https://html.spec.whatwg.org/multipage/links.html#rel-icon):

> For historical reasons, the `icon` keyword _may_ be preceded by the keyword "shortcut". ...

If for some reason you know that "shortcut" should be present, please let me know and I'll add it back.

devoncarew: > If for some reason you know that "shortcut" should be present

I don't! I just saw the diff and wanted to make sure it was intentional.

PR #844, opened 2018-05-04
Remove all uses of "new" and "const" from "Effective Dart".

munificent:
It's happening! \o/

I'm not sure if now is the right time to land this, but I wanted to
send it out before I forgot.

### DIFF
--- a/examples/misc/lib/effective_dart/design_bad.dart
+++ b/examples/misc/lib/effective_dart/design_bad.dart
@@ -36,7 +36,7 @@ void miscDeclAnalyzedButNotTested() {
 
   {
     // #docregion cascades
-    var buffer = new StringBuffer0() //!<br>
+    var buffer = StringBuffer0() //!<br>
         .write('one')
         .write('two')
         .write('three');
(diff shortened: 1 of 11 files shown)


chalin:
The easiest solution is to wait for an SDK that supports this.
The second easiest solution is to disable analysis and testing on Travis :).

chalin:
Looking at the current Travis log I see that only one test file is failing to load due to:

    var ints = List<int>.from(numbers);

I'd rather leave the analysis and tests running on Travis. We could reinsert the new in the code but use `replace="..."` over the code-excerpt in the docs (with a note to remove it). That way we can have our cake and eat it too (w/o having to wait). WDYT?

munificent:
> We could reinsert the new in the code but use replace="..." over the code-excerpt in the docs (with a note to remove it). That way we can have our cake and eat it too (w/o having to wait).

That sounds like a hassle, especially for excerpts that are already using replace. How about we just sit on this until it's safe?

chalin:
> How about we just sit on this until it's safe?

SGTM

chalin:
> How about we just sit on this until it's safe?

I think that the time is right. (Note that rebasing _might_ result in a lot of conflicts, in which case it might be easier to just recreate this PR.)

munificent:
I have some other stuff to worry about today and I want to make sure the Angular compiler issue around inferred `const` in metadata annotations is fixed before "Effective Dart" starts pushing people to omit it. Let me get on top of those first and then I'll circle back to this.

munificent:
Closing this. I'll re-apply and send out a new PR.

PR #855, opened 2018-05-09
Add a guideline about using Future<void>.

munificent:


src/_guides/language/effective-dart/usage.md
```diff
@@ -1307,6 +1307,19 @@ Future<bool> fileContainsBear(String path) async {
 {% endprettify %}
 
 
+### DO use `Future<void>` as the return type of asynchronous members that do not produce values.
+
+Often you have an operation that is asynchronous but doesn't actually return a
+useful value when it's done. If it were a synchronous function, you'd use `void`
+as the return type to indicate that.
+
+The asynchronous equivalent of that is `Future<void>`. You may see code that
```
chalin: I'm unsure about the "often" qualifier in the first paragraph. How about something like:

> You use `void` as the return type of a synchronous function that doesn't return a value. 
> The asynchronous equivalent is `Future<void>`.
>
> You may see code that ...

munificent: Done. Also, I moved this to the "Design" page because it's about the signature of a member, not its implementation.

PR #860, opened 2018-05-10
Compress all PNGS

kevmoo:
Before: 3,523,937
 After: 2,470,759

Used find . -name '*.png' -exec pngquant --ext .png --force 256 --skip-if-larger {} \;


chalin:
FYI, I'll handle image compression on webdev and site-shared.

kevmoo:
Cool!

PR #863, opened 2018-05-11
More revisions to "Effective Dart" based on feedback.

munificent:
Staged: https://sz-www-1.firebaseapp.com/guides/language/effective-dart

- Move the "positive bool" guideline because it got stuck between two
  guidelines that refer to each other.

- Show that the new typedef syntax still supports parameter names.

- Clarify library doc comment guideline.

- Link to formatting rules on dartfmt wiki.

- Remove guideline on super() placement since the language mandates it
  now.

- Clarify guideline on importing from "src".

- Remove references to deprecated retype() method.

- Include examples for "AVOID using cast()".

- Add example for "DON'T use new".

- Other copy-edits.

### DIFF
--- a/examples/misc/lib/effective_dart/usage_good.dart
+++ b/examples/misc/lib/effective_dart/usage_good.dart
@@ -79,23 +79,19 @@ void miscDeclAnalyzedButNotTested() {
   }
 
   // #docregion cast-at-create
-  List<int> makeList([int a, int b]) {
+  List<int> singletonList(int value) {
     var list = <int>[];
-    if (a != null) list.add(a);
-    if (b != null) list.add(b);
+    list.add(value);
     return list;
   }
   // #enddocregion cast-at-create
 
   // #docregion cast-iterate
-  int sum(List<Object> objects) {
-    var result = 0;
+  void printEvens(List<Object> objects) {
     // We happen to know the list only contains ints.
-    for (int n in objects) {
-      result += n;
+    for (var n in objects) {
+      if ((n as int).isEven) print(n);
     }
-
-    return result;
   }
   // #enddocregion cast-iterate
 
@@ -222,6 +218,20 @@ void miscDeclAnalyzedButNotTested() {
     }
     // #enddocregion avoid-completer-alt
   }
+
+  // TODO(rnystrom): Uncomment this when the analyzer run on these excerpts
+  // supports optional const.
+  {
+    /*
+    // #docregion no-const
+    const primaryColors = [
+      Color("red", [255, 0, 0]),
+      Color("green", [0, 255, 0]),
+      Color("blue", [0, 0, 255]),
+    ];
+    // #enddocregion no-const
+    */
+  }
 }
 
 //----------------------------------------------------------------------------
@@ -235,6 +245,10 @@ class Animal {
 
 class Person {
   int zip;
+}
+
+class Color {
+  const Color(String name, List<int> channels);
 }
 
 //----------------------------------------------------------------------------
@@ -426,15 +440,32 @@ class Point2 {
 }
 // #enddocregion semicolon-for-empty-body
 
+class Widget {}
+
+class BuildContext {}
+
+class Row extends Widget {
+  Row({children});
+}
+
+class RaisedButton {
+  RaisedButton({child});
+}
+
+class Text {
+  Text(String text);
+}
+
+// TODO(rnystrom): Remove "new" here instead of stripping it out later once
+// we support optional new for excerpts.
 // #docregion no-new
 Widget build(BuildContext context) {
   return new Row(
     children: [
       new RaisedButton(
-        onPressed: increment,
         child: new Text('Increment'),
       ),
-      new Text('Count: $counter'),
+      new Text('Click!'),
     ],
   );
 }
--- a/examples/misc/lib/effective_dart/usage_good.dart
+++ b/examples/misc/lib/effective_dart/usage_good.dart
@@ -78,6 +78,36 @@ void miscDeclAnalyzedButNotTested() {
     // #enddocregion cast-map
   }
 
+  // #docregion cast-at-create
+  List<int> makeList([int a, int b]) {
+    var list = <int>[];
+    if (a != null) list.add(a);
+    if (b != null) list.add(b);
+    return list;
+  }
+  // #enddocregion cast-at-create
+
+  // #docregion cast-iterate
+  int sum(List<Object> objects) {
+    var result = 0;
+    // We happen to know the list only contains ints.
+    for (int n in objects) {
+      result += n;
+    }
+
+    return result;
+  }
+  // #enddocregion cast-iterate
+
+  // #docregion cast-from
+  int median(List<Object> objects) {
+    // We happen to know the list only contains ints.
+    var ints = objects.cast<int>();
+    ints.sort();
+    return ints[ints.length ~/ 2];
+  }
+  // #enddocregion cast-from
+
   (Iterable<Animal> animals) {
     // #docregion use-higher-order-func
     var aquaticNames = animals
@@ -395,6 +425,20 @@ class Point2 {
   Point2(this.x, this.y);
 }
 // #enddocregion semicolon-for-empty-body
+
+// #docregion no-new
+Widget build(BuildContext context) {
+  return new Row(
+    children: [
+      new RaisedButton(
+        onPressed: increment,
+        child: new Text('Increment'),
+      ),
+      new Text('Count: $counter'),
+    ],
+  );
+}
+// #enddocregion no-new
 
 //----------------------------------------------------------------------------
 
--- a/src/_guides/language/effective-dart/design.md
+++ b/src/_guides/language/effective-dart/design.md
@@ -220,6 +220,20 @@ setters are invoked in templates, not from other Dart code.
 </aside>
 
 
+### CONSIDER omitting the verb for a named boolean *parameter*.
+
+This refines the previous rule. For named parameters that are boolean, the name
+is often just as clear without the verb and it reads better at the callsite.
+
+{:.good-style}
+<?code-excerpt "misc/lib/effective_dart/design_good.dart (omit-verb-for-bool-param)"?>
+{% prettify dart %}
+Isolate.spawn(entryPoint, message, paused: false);
+var copy = new List.from(elements, growable: true);
+var regExp = new RegExp(pattern, caseSensitive: false);
+{% endprettify %}
+
+
 ### PREFER the "positive" name for a boolean property or variable.
 
 Most boolean names have conceptually "positive" and "negative" forms where the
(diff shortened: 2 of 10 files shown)


examples/misc/lib/effective_dart/usage_good.dart
```diff
@@ -396,6 +426,20 @@ class Point2 {
...
+    children: [
+      new RaisedButton(
+        onPressed: increment,
+        child: new Text('Increment'),
+      ),
+      new Text('Count: $counter'),
+    ],
+  );
+}
+// #enddocregion no-new
```
chalin: Consider adding at least a FIXME note to actually remove the `new` keyword instances once it is supported (or place this example in the `..._bad.dart` file :)

kwalrath:
Looking now...

src/_guides/language/effective-dart/usage.md
```diff
@@ -441,33 +444,110 @@ var reciprocals = stuff.map((n) => 1 / n).cast<double>();
...
 This is the softer generalization of the previous rule. Sometimes there is no
 nearby operation you can use to fix the type of some object. Even then, when
-possible avoid using `cast()` or `retype()` to "change" a collection's type.
+possible avoid using `cast()` to "change" a collection's type.
 
 Prefer any of these options instead:
 
-*   Change the code where the collection is first created so that it has the
-    right type.
+#### Create it with the right type
```
kwalrath: Although I appreciate the sample code, I miss seeing all the options in one windowful. How about keeping the list, but making each item a link to the subheading. Like this:

```
...instead:

* [Create it with the right type](#create-it-with-the-right-type)
* [Cast the elements on access](#cast-the-elements-on-access)
* ...

#### Create it with the right type
```

src/_guides/language/effective-dart/usage.md
```diff
@@ -441,33 +444,110 @@ var reciprocals = stuff.map((n) => 1 / n).cast<double>();
...
+  return ints[ints.length ~/ 2];
+}
+{% endprettify %}
+
+The `cast()` method returns a lazy collection that checks the element type on
+*every operation*. If you perform only a few operations on only a few elements,
+that laziness can be good. But in many cases, the overhead of lazy validation
+and of wrapping outweighs the benefits.
+
+These alternatives don't always work, of course, and sometimes `cast()` is the
```
kwalrath: Move this paragraph up, to just after the bulleted list.

src/_guides/language/effective-dart/usage.md
```diff
@@ -1053,6 +1098,22 @@ actually return a new object.
...
+      RaisedButton(
+        onPressed: increment,
+        child: Text('Increment'),
+      ),
+      Text('Count: $counter'),
+    ],
+  );
+}
+{% endprettify %}
+
```
kwalrath: AngularDart doesn't yet support using `new`. Maybe add a note about this, with a link to https://github.com/dart-lang/angular/issues/1313?

src/_guides/language/effective-dart/usage.md
```diff
@@ -1053,6 +1098,22 @@ actually return a new object.
...
 
 ### DON'T use `const` when not needed.
```
kwalrath: No example for this one? :)

examples/misc/lib/effective_dart/usage_good.dart
munificent: Done.

src/_guides/language/effective-dart/design.md
```diff
@@ -220,6 +220,20 @@ setters are invoked in templates, not from other Dart code.
 </aside>
 
 
+### CONSIDER omitting the verb for a named boolean *parameter*.
+
+This refines the previous rule. For named parameters that are boolean, the name
+is often just as clear without the verb and it reads better at the callsite.
```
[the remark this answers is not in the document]
munificent: Tried a slightly different wording.

src/_guides/language/effective-dart/usage.md
munificent: Done. I liked seeing everything in a list too. I wish the site supported code snippets inside bullet lists, and I wish the code snippets were a little more compact. Maybe something to consider for the next redesign.

src/_guides/language/effective-dart/usage.md
munificent: With the examples split out from the list, I like this paragraph at the end now because I feel like it summarizes the whole thing.

src/_guides/language/effective-dart/usage.md
munificent: I revised the guideline to make it clear that it only applies if you don't need to support tools that don't implement it. I'm hesitant to mention Angular specifically because I don't want the doc to get out of date.

src/_guides/language/effective-dart/usage.md
munificent: Done. Got lazy. :)

munificent:
OK, should be more betterer now.

src/_guides/language/effective-dart/usage.md
kwalrath: It does, and yet no one might see it. With the headings, it looks like it only applies to the last section, and the scope of "These alternatives" is unclear.

src/_guides/language/effective-dart/usage.md
```diff
@@ -1105,17 +1100,16 @@ Widget build(BuildContext context) {
...
-      Text('Count: $counter'),
+      Text('Click!'),
     ],
   );
 }
 {% endprettify %}
 
 
-### DON'T use `const` when not needed.
+### DON'T use `const` unnecesarily in code that doesn't need to support Dart 1.
```
kwalrath: unnecesarily -> unnecessarily

src/_guides/language/effective-dart/usage.md
```diff
@@ -452,71 +452,70 @@ possible avoid using `cast()` to "change" a collection's type.
...
+    first created so that it has the right type.
 
-Change the code where the collection is first created so that it has the right
-type:
+*   **Cast the elements on access.** If you immediately iterate over the
+    collection, cast each element inside the iteration.
+
+*   **Eagerly cast using `List.from()`.** If you'll eventually access most of
+    the elements in the collection, and you don't need the object to be backed
+    by the original live object, convert it using `List.from()`:
```
kwalrath: : -> .

src/_guides/language/effective-dart/usage.md
```diff
@@ -452,71 +452,70 @@ possible avoid using `cast()` to "change" a collection's type.
...
+
+    The `cast()` method returns a lazy collection that checks the element type
+    on *every operation*. If you perform only a few operations on only a few
+    elements, that laziness can be good. But in many cases, the overhead of lazy
+    validation and of wrapping outweighs the benefits.
+
+Here are examples of each approach:
```
kwalrath: I avoid using two colons in a row. Actually, let's delete this sentence and convert the headings into sentence-style intros. (Is there an important reason to have headings here?) The result would be something like this:

----------------------

...outweighs the benefits.

Here is example code for the first option, **create it with the right type**:
...
Here are examples for **cast each element on access**:
...
Here are examples for **cast eagerly using `List.from()`**:
...

These alternatives don’t always work....

----------------------

kwalrath: @chalin can you think of a way to save vertical space here? These samples could be side-by-side, like diffs (except I suspect that diffs here wouldn't save space and might not be clear).

src/_guides/language/effective-dart/usage.md
kwalrath: Could you file an issue about that? If you could point to an existing example of formatting that you really like, that would be helpful.

src/_guides/language/effective-dart/usage.md
kwalrath: But if we remove the headings, it can stay!! (My soon-to-be-sent review has a suggestion for that.)

src/_guides/language/effective-dart/usage.md
chalin: (Do you have a preview somewhere by any chance? If not, I'll render and take a look tomorrow.)

src/_guides/language/effective-dart/usage.md
```diff
@@ -1128,11 +1122,33 @@ expression inside:
...
   the body of the case.
 
-(Default values are *not* included in this list because future versions of Dart
+(Default values are not included in this list because future versions of Dart
 may support non-const default values.)
 
-Basically, any place where it would be a static error to write `new` instead of
-`const`, you can and should omit the `const`.
+Basically, any place where it would be an error to write `new` instead of
+`const`, Dart 2 allows you to omit the `const`. If your code does not need to
```
kwalrath: does not -> doesn't

https://developers.google.com/style/contractions

kwalrath:
Just staged it (and put the URL in the first comment, so it's easy to find):

https://sz-www-1.firebaseapp.com/guides/language/effective-dart

src/_guides/language/effective-dart/usage.md
chalin: I'd vote to leave it as is for now. If we decide to tackle putting good/bad code excerpts side-by-side (for wide screens), we should do that throughout the Effective Dart pages. Do you want to create a separate issue to track this?

src/_guides/language/effective-dart/usage.md
munificent: You're right. I was worried about "Effective Dart" prohibiting people from writing code that worked in Dart 1, but I think that ship has sailed. For example, another guideline tells you to use `whereType()`, and that method doesn't work at all in Dart 1 (nor is it even defined in older versions of the SDK).

src/_guides/language/effective-dart/usage.md

src/_guides/language/effective-dart/usage.md
munificent: I like your suggestion, Kathy.

src/_guides/language/effective-dart/usage.md
munificent: Removed the sentence.

src/_guides/language/effective-dart/usage.md
```diff
@@ -1083,15 +1081,14 @@ class Point {
...
+### DON'T use `new`.
 
 Dart 2 makes the `new` keyword optional. Even in Dart 1, its meaning was never
 clear because factory constructors mean a `new` invocation may still not
 actually return a new object.
 
 The language still permits `new` in order to make migration less painful, but
-consider it deprecated and remove it from your code as soon as you no longer
-need to run on older Dart tools.
+consider it deprecated and remove it from your code.
```
kwalrath: IDEs still don't have good support for generating code that doesn't have `new`. Perhaps wait on this guideline? Make it CONSIDER?

/cc @wmleler @filiph @kevmoo

src/_guides/language/effective-dart/usage.md
```diff
@@ -1126,9 +1123,7 @@ expression inside:
 may support non-const default values.)
 
 Basically, any place where it would be an error to write `new` instead of
-`const`, Dart 2 allows you to omit the `const`. If your code does not need to
-run on older versions of Dart, you should take advantage of that and drop the
-`const`.
+`const`, Dart 2 allows you to omit the `const`.
```
kwalrath: @wmleler @filiph @kevmoo do the tools do a good job with `const` yet?

src/_guides/language/effective-dart/usage.md
kevmoo: I'm not an expert on where this is...

src/_guides/language/effective-dart/usage.md
kevmoo: 🤷‍♂️

munificent:
It looks like optional `const` isn't supported by AngularDart in metadata annotations yet. So I commented out those guidelines for now. I'll re-enable them once they're safe to use.

kwalrath:
Thanks for the new/const adjustment. We usually comment out using {% comment %}...{% endcomment %}, so I'll just make sure that it renders OK...

PR #864, opened 2018-05-12
Update README.md

kwalrath:


kwalrath:
I canceled the build in case someone else needs Travis for something more substantive.

Issue #888, opened 2018-05-23
language-tour page's description of booleans is confusing


jamesderlin:
https://www.dartlang.org/guides/language/language-tour#if-and-else states:

> Remember, unlike JavaScript, Dart treats all values other than true as false. See Booleans for more information.

The Booleans section that links to states:

> To represent boolean values, Dart has a type named bool. Only two objects have type bool: the boolean literals true and false, which are both compile-time constants.
> 
> Dart’s type safety means that you can’t use code like if (nonbooleanValue) or assert (nonbooleanValue). Instead, explicitly check for values, like this:

This seems inconsistent (or at least, confusing).  How can non-true values be treated as false if conditions require boolean values (which must be either true or false)? 

Is the cited paragraph in the if-and-else section a vestige from Dart 1?

jamesderlin:
I suppose that this is fallout from 9b0c79b47038e4d89208ee2f5bfa8282e1a2002e for #781.

kwalrath:
Yes, it is. You've found something we missed. Thanks for the report! (Want to submit a PR?)

chalin:
Closed by #891, right?

jamesderlin:
Right.

PR #891, opened 2018-05-25
Remove mention of non-booleans as booleans (#888)

jamesderlin:
With Dart 2's type-safety, only true and false are allowed where booleans are expected, so remove the language-tour's mention of treatment of non-booleans as booleans.


jamesderlin:
I signed it!

(Well, not actually, but I don't need to.)

kwalrath:
Thanks for the fix! (And sorry for the delay in merging it.)

PR #892, opened 2018-05-25
update google privacy policy url

bcko:


kwalrath:
I'm going to commit this and then remove the ?hl=en separately.

PR #893, opened 2018-05-25
Show example alternate verbs in "avoid get" guideline.

munificent:


src/_guides/language/effective-dart/design.md
```diff
@@ -355,10 +355,18 @@ named `breakfastOrder`.
...
 Even if the member does need to be a method because it takes arguments or
 otherwise isn't a good fit for a getter, you should still avoid `get`. Like the
-previous guidelines state, you can use a noun phrase name like
-`breakfastOrder()` or choose a verb that more usefully describes how the method
-produces its result. The word `get` conveys nothing to the caller that they
-don't already know.
+previous guidelines state, either:
+
+* Simply drop `get` and [use a noun phrase name][noun] like `breakfastOrder()`
+  if the caller mostly cares about the value the method returns.
```
yjbanov: I think it could be confusing to use noun phrases for getters and methods. It's not clear when you should add parens. IDEs might help, but not sure if we want to rely on IDEs too much.

munificent: > I think it could be confusing to use noun phrases for getters and methods.

That's been a guideline basically forever. In practice, it hasn't seemed to cause too many problems: `indexOf()`, `substring()`, `abs()`, `ceil()`, `firstWhere()`, etc.

> IDEs might help, but not sure if we want to rely on IDEs too much.

Unprefixed imports and implicit calls on `this` have basically forced that hand for us. In practice, even if you don't use an IDE, if you omit the parens when you shouldn't, you'll get a compile error pretty quickly.

PR #902, opened 2018-05-31
Update example to comply with = for default parameters.

johnsonmh:
Previously stated rule: https://www.dartlang.org/guides/language/effective-dart/usage#do-use--to-separate-a-named-parameter-from-its-default-value

Named parameters should use = to separate parameters from its default value.

### DIFF
--- a/src/_guides/language/effective-dart/design.md
+++ b/src/_guides/language/effective-dart/design.md
@@ -1682,12 +1682,12 @@ DateTime(int year,
     int microsecond = 0]);
 
 Duration(
-    {int days: 0,
-    int hours: 0,
-    int minutes: 0,
-    int seconds: 0,
-    int milliseconds: 0,
-    int microseconds: 0});
+    {int days = 0,
+    int hours = 0,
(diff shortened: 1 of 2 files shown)


johnsonmh:
I tried to sign it but I am Googler.

src/_guides/language/effective-dart/design.md
```diff
@@ -1682,12 +1682,12 @@ DateTime(int year,
...
-    int minutes: 0,
-    int seconds: 0,
-    int milliseconds: 0,
-    int microseconds: 0});
+    {int days = 0,
+    int hours = 0,
+    int minutes = 0,
+    int seconds = 0,
+    int milliseconds = 0,
+    int microseconds = 0});
```
chalin: Thanks for spotting this. Could you also update the source from which this excerpt is taken (that'll get the build to pass on Travis). It is in `misc/lib/effective_dart/design_good.dart`; search for the region named `omit-optional-positional`.

chalin:
Thanks for the PR!

Issue #668, opened 2018-03-14
Re-enable frontpage main example tests and "Open in Dartpad" button 


chalin:
We've dropped the use of `new` from the main example. While we can still run the analyzer over this code (w/ `--preview-dart-2`), we can no longer run tests or launch Dartpad.

- [x] Run tests, since these require a web compiler that supports optional new/const - #903
- [x] Dartpad isn't Dart 2 compatible yet.

kwalrath:
Why does this test require a web compiler? I'd have thought we'd use the Dart VM wherever possible on site-www. (Although since this is intended to work in DartPad, I guess it makes sense to test this particular example with a web compiler.)

chalin:
> since this is intended to work in DartPad, I guess it makes sense to test this particular example with a web compiler

Right, it is tested in DartPad as a web app.

kwalrath:
DartPad now supports Dart 2.

chalin:
Closed by #750, #903

Issue #870, opened 2018-05-14
Remove new & const (most) everywhere


kwalrath:
- [x] Update language tour to avoid using new/const except where necessary. #507 
- [x] Add optional new/const to /dart-2. #945 
- [x] Update all **tested** code samples to NOT use new/const unless absolutely necessary. #966
- [x] Remove all uses of "new" and "const" from "Effective Dart". #844

kwalrath:
We've updated all tested code samples, which means all of the most used/updated pages. @chalin, do you know whether everything else that might contain new and const is marked as obsolete?

chalin:
> do you know whether everything else that might contain new and const is marked as obsolete?

No, I don't know.

chalin:
I'm checking now to ensure that pages that need to be are marked as (potentially) out-of-date (using `obsolete: true`).

chalin:
A few extra pages needed to be marked as obsolete, see #973.

chalin:
The analyzer is checking for `unnecessary_const` and `unnecessary_new` in all examples (#1026). No issues are reported, so I'd say that we're done!

PR #905, opened 2018-06-04
Re-enable homepage "Open in DartPad" button

kwalrath:
Contributes to #668 

Staged at https://kw-www-dartlang-1.firebaseapp.com.

@chalin, do you think we need testing in place before merging this? This PR does not remove the 2 instances of `new`.


chalin:
> do you think we need testing in place before merging this?

No, we can do that separately.

> This PR does not remove the 2 instances of new.

Do we have an issue open for removing those? #870? (I thought we had another more specific issue but I can't find it. Maybe I'm mistaking it for the PR we applied, and reverted.)

PR #912, opened 2018-06-09
chore: use new code excerpter

chalin:
Switch to using the new code excerpter. It will buy us more flexibility moving forward.

### DIFF
--- a/examples/misc/lib/language_tour/classes/employee.dart
+++ b/examples/misc/lib/language_tour/classes/employee.dart
@@ -14,9 +14,8 @@ class Person {
 
 // #docregion method-then-constructor
 class Employee extends Person {
-  // #enddocregion ,
+  // #enddocregion ''
   Employee() : super.fromJson(getDefaultData());
-  // ···
   // #enddocregion method-then-constructor
   // #docregion
   // Person does not have a default constructor;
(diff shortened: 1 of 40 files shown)


src/_guides/language/language-tour.md
```diff
@@ -3139,7 +3135,7 @@ constants:
 <?code-excerpt "misc/lib/language_tour/classes/misc.dart (static-field)"?>
 {% prettify dart %}
 class Queue {
-  static const int initialCapacity = 16;
+  static const initialCapacity = 16;
```
chalin: Dropped unnecessary type annotation. /cc @kwalrath

chalin:
Other than dropping an unnecessary `int` type annotation on a `const`, and a few minor plaster/ellipses tweaks, this PR introduced no doc changes. Travis is green. Merging.

examples/misc/lib/language_tour/classes/employee.dart
```diff
@@ -14,9 +13,8 @@ class Person {
 
 // #docregion method-then-constructor
 class Employee extends Person {
-  // #enddocregion ,
+  // #enddocregion ''
   Employee() : super.fromJson(getDefaultData());
-  // ···
```
chalin: This is an example of a manually inserted plaster. This sort of thing was necessary because plasters used to be turned on/off for an entire file. This is no longer the case.

PR #915, opened 2018-06-11
Add api.dartlang.org link to every page

### LINKED ISSUE #913: api.dartlang.org is not linked from www.dartlang.org
At least not on the homepage, nav bar or footer.

kwalrath:
Fixes #913.

I considered linking to `{{site.dart_api}}/{{site.data.pkg-vers.SDK.channel}}`, but I think it's better to just link to the API homepage, which has the big note about versions.

### DIFF
--- a/src/_includes/page-footer.html
+++ b/src/_includes/page-footer.html
@@ -50,6 +50,7 @@
         <div class="content">
           <h4>Resources</h4>
           <ul>
+            <li><a href="{{site.dart_api}}">API reference</a></li>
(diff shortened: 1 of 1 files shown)


chalin:
IMHO, we should like to `{{site.dart_api}}/{{site.data.pkg-vers.SDK.channel}}`. That notice about versions is easy to miss.

kwalrath:
True, people often gloss over notes. I'll do the other link.

PR #918, opened 2018-06-11
Fix typedef examples.

munificent:


chalin:
> The original syntax looks like:
> ```dart
> typedef int Comparison<T>(T a, T b);
> ```
> That syntax has a couple of problems:
> - There is no way [with the original typedef syntax] to assign a name to a generic function type.

This PR changes the typedef definition in the new syntax to the following, which is the equivalent of what is given above in the original syntax:

```dart
typedef Comparison<T> = int Function(T, T);
```

It might be worth illustrating the "way to assign a name to a generic function type", which is what was illustrated before the fix introduced in this PR:

```dart
typedef Comparison = int Function<T>(T, T);
```

That might give folks a concrete example of the increased expressiveness. WDYT? /cc @kwalrath

munificent:
The generic function type use case is vanishingly rare. (I don't recall ever seeing one in the wild, nor do I know of any real-world use cases that motivated adding it to the language.) I don't think it's worth putting in "Effective Dart".

PR #922, opened 2018-06-13
HTTP tutorial: remove "not Dart-2" warning since it was updated a while ago

chalin:
The code was "migrated" to Dart 2 via #463, and its completion status was marked as done in #664.


chalin:
@kwalrath - since this was already marked as done, I'm going to merge :).

PR #923, opened 2018-06-13
Use new typedef syntax everywhere

chalin:
Use new typedef syntax everywhere, except places where we are illustrating the old syntax :)

### DIFF
--- a/examples/strong/analyzer-2-results.txt
+++ b/examples/strong/analyzer-2-results.txt
@@ -10,8 +10,8 @@ Analyzing lib, test...
   error • Invalid override. The type of 'MyAdder.add' ('(int, int) → int') isn't a subtype of 'NumberAdder.add' ('(num, num) → num') at lib/common_problems_analysis.dart:61:3 • strong_mode_invalid_method_override
   error • Invalid override. The type of 'Subclass.method' ('(int) → void') isn't a subtype of 'Superclass<dynamic>.method' ('(dynamic) → void') at lib/common_problems_analysis.dart:74:3 • strong_mode_invalid_method_override
   error • super call must be last in an initializer list (see https://goo.gl/EY6hDP): 'super(food)' at lib/common_problems_analysis.dart:92:9 • strong_mode_invalid_super_invocation
-  error • A value of type '(String) → bool' can't be assigned to a variable of type '(dynamic) → bool' at lib/common_problems_analysis.dart:103:17 • invalid_assignment
-  error • The function expression type '(String) → bool' isn't of type '(dynamic) → bool'. This means its parameter or return type does not match what is expected. Consider changing parameter type(s) or the returned type(s) at lib/common_problems_analysis.dart:103:17 • strong_mode_invalid_cast_function_expr
+  error • The function expression type '(String) → bool' isn't of type '(dynamic) → bool'. This means its parameter or return type does not match what is expected. Consider changing parameter type(s) or the returned type(s) at lib/common_problems_analysis.dart:102:17 • strong_mode_invalid_cast_function_expr
+  error • A value of type '(String) → bool' can't be assigned to a variable of type '(dynamic) → bool' at lib/common_problems_analysis.dart:102:17 • invalid_assignment
   error • The argument type 'List' can't be assigned to the parameter type 'List<int>' at lib/strong_analysis.dart:27:17 • argument_type_not_assignable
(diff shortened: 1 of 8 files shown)


examples/strong/analyzer-2-results.txt
```diff
@@ -10,8 +10,8 @@ Analyzing lib, test...
...
   error • super call must be last in an initializer list (see https://goo.gl/EY6hDP): 'super(food)' at lib/common_problems_analysis.dart:92:9 • strong_mode_invalid_super_invocation
-  error • A value of type '(String) → bool' can't be assigned to a variable of type '(dynamic) → bool' at lib/common_problems_analysis.dart:103:17 • invalid_assignment
-  error • The function expression type '(String) → bool' isn't of type '(dynamic) → bool'. This means its parameter or return type does not match what is expected. Consider changing parameter type(s) or the returned type(s) at lib/common_problems_analysis.dart:103:17 • strong_mode_invalid_cast_function_expr
+  error • The function expression type '(String) → bool' isn't of type '(dynamic) → bool'. This means its parameter or return type does not match what is expected. Consider changing parameter type(s) or the returned type(s) at lib/common_problems_analysis.dart:102:17 • strong_mode_invalid_cast_function_expr
+  error • A value of type '(String) → bool' can't be assigned to a variable of type '(dynamic) → bool' at lib/common_problems_analysis.dart:102:17 • invalid_assignment
```
chalin: Line number change (L103 -> L102), and the order of the two errors is reversed.

src/_guides/language/sound-problems.md
```diff
@@ -371,25 +371,25 @@ in a compile-time error.
...
 {:.console-output}
 <?code-excerpt "strong/analyzer-2-results.txt" retain="/type '\(String\) → bool'.*common_problems/"?>
 ```nocode
-error • A value of type '(String) → bool' can't be assigned to a variable of type '(dynamic) → bool' • invalid_assignment
 error • The function expression type '(String) → bool' isn't of type '(dynamic) → bool'. This means its parameter or return type does not match what is expected. Consider changing parameter type(s) or the returned type(s) • strong_mode_invalid_cast_function_expr
+error • A value of type '(String) → bool' can't be assigned to a variable of type '(dynamic) → bool' • invalid_assignment
 ```
 
-#### Fix: Add generic type parameters _or_ cast from dynamic explicitly
+#### Fix: Add type parameters _or_ cast from dynamic explicitly
```
chalin: @munificent - are you ok with this change? I dropped "generic", since, under the new syntax

```dart
typedef Filter<T> = bool Function(T any);
```

the`<T>` in `Filter<T>` type isn't a _generic type_ (although it is a type parameter).

examples/strong/lib/common_fixes_analysis.dart
```diff
@@ -82,13 +82,13 @@ class HoneyBadger extends Animal {
...
-typedef bool Filter<T>(T any);
+typedef Filter<T> = bool Function(T any);
 Filter<String> filter = (String x) => x.contains('Hello');
 // #enddocregion func-T
 
 //-----------------------------------------------
 
 // #docregion func-cast
-typedef bool Filter1(dynamic any);
+typedef Filter1<T> = bool Function(T any);
```
munificent: This should be:

```dart
typedef Filter1 = bool Function(dynamic any);
```

Shouldn't it?

src/_guides/language/sound-problems.md
munificent: 👍

src/_guides/language/sound-problems.md
```diff
@@ -398,7 +398,7 @@ Otherwise use casting:
 {:.passes-sa}
 <?code-excerpt "strong/lib/common_fixes_analysis.dart (func-cast)" replace="/([Ff]ilter)1/$1/g; /as \w+/[!$&!]/g"?>
 {% prettify dart %}
-typedef bool Filter(dynamic any);
+typedef Filter<T> = bool Function(T any);
```
munificent:
```dart
typedef Filter = bool Function(dynamic any);
```

examples/strong/lib/common_fixes_analysis.dart
chalin: Yes (they should be equivalent, but I agree that the original explicit use of `dynamic` there is better).

Fixed.

src/_guides/language/sound-problems.md
chalin: Fixed.

PR #925, opened 2018-06-13
chore: prep to globally set `DART_VM_OPTIONS=--preview-dart-2`

chalin:


chalin:
This has turned into a preparatory PR, awaiting a fix to https://github.com/dart-lang/sdk/issues/33430 before we can globally set `DART_VM_OPTIONS`. It is still worth committing this because it has a change in the submodule version, and it uses the updated `linkcheck`.

chalin:
@kwalrath:

- As usual, you'll need to run `git submodule update --init --remote`
- You'll also need to run `pub upgrade` (from the project root).

PR #928, opened 2018-06-14
Sidenav: allow top-level entries to be links

### LINKED ISSUE #746: Sidenav: allow top-level entries to be links, like on webdev
The sidenav currently looks like this:

> sidenav

Sidenav real estate is precious on mobile devices. On **webdev we've allowed top-level entries to be links** as well. Doing this here would allow use to eliminate all three "Overview" subitems as well as the "Dart Testing" subitem:

> screen shot 2018-04-05 at 10 45 42

chalin:
Fixes #746 

Staged: see, e.g., https://dartlang-org-staging-0.firebaseapp.com/guides/language, which looks like this:

screen shot 2018-06-14 at 09 46 12


src/_data/side-nav.yml
```diff
@@ -37,11 +37,12 @@
...
 
 - title: Testing
-  children:
-  - title: Dart Testing
-    permalink: /guides/testing
+  permalink: /guides/testing
+  urlExactMatch: true
 
 - title: Resources
+  permalink: /resources
```
kwalrath: This page is pretty useless. Would it be a problem to remove the permalink? Another option is to make the page actually, you know, useful. At the very least, it should include links to everything listed under it in the sidenav.

chalin:
> What do you think of making it not a link and filing an issue to fix it and link to it?

Done, and done (#929).

Issue #879, opened 2018-05-18
Improve language tour's typedef section


kwalrath:
https://www.dartlang.org/guides/language/language-tour#typedefs needs a rewrite to remove or deemphasize the old syntax and explain more clearly how to use the new syntax. I suspect we can cut the paragraph and code that exist solely to justify typedefs, and just talk about use cases and gotchas.

More info: dart-lang/sdk#33165

/cc @leafpetersen

chalin:
> https://www.dartlang.org/guides/language/language-tour#typedefs needs a rewrite to remove or deemphasize the old syntax and explain more clearly how to use the new syntax.

- [x] This has been done. The Language Tour contains examples of the new syntax only.

---

The [Tour's Typedef section](https://www.dartlang.org/guides/language/language-tour#typedefs) ends with the following note:

> **New function type syntax:** Dart 1.24 introduced a new form of function types, the [generic function type alias.](https://github.com/dart-lang/sdk/blob/master/docs/language/informal/generic-function-type-alias.md) ...

- [ ] I suspect that soon, more readers will be learning/familiar with Dart 2 as opposed to Dart 1. I'd **suggest dropping this note**, esp. given that Effective Dart has [DON’T use the legacy typedef syntax](https://www.dartlang.org/guides/language/effective-dart/design#dont-use-the-legacy-typedef-syntax).

---

> I suspect we can cut the paragraph and code that exist solely to justify typedefs, and just talk about use cases and gotchas.

For a Language Tour, the current content seems reasonable.

---

@kwalrath WDYT?

kwalrath:
Agreed we can drop that note.

kwalrath:
Is there anything else we need to do to close this issue?

chalin:
I don't think that there is anything else to be done.

PR #931, opened 2018-06-14
Language tour: drop note about "new" typedef syntax

chalin:
Contributes to #879


chalin:
Approved via https://github.com/dart-lang/site-www/issues/879#issuecomment-397392085. Merging.

PR #932, opened 2018-06-14
Example tests: ensure assertion checking is enabled

chalin:


chalin:
Ah, found the previous sanity test! This mainly moves it to the top level of the misc project test folder.

Issue #422, opened 2017-11-09
Remove embedded DartPads


chalin:
Known affected pages:
- https://www.dartlang.org/guides/get-started
- https://www.dartlang.org/guides/language/language-tour
- https://www.dartlang.org/tutorials/language/futures
- https://www.dartlang.org/tutorials/language/streams

There may be more.

Chrome warnings as seen in the developer console:
- [Deprecation] Styling master document from stylesheets defined in HTML Imports is deprecated, and is planned to be removed in M65, around March 2018. Please refer to https://goo.gl/EGXzpw for possible migration paths.
- [Deprecation] /deep/ combinator is no longer supported in CSS dynamic profile. It is now effectively no-op, acting as if it were a descendant combinator. You should consider to remove it. See https://www.chromestatus.com/features/4964279606312960 for more details.
- Blink deferred a task in order to make scrolling smoother. Your timer and network tasks should take less than 50ms to run to avoid this. Please see https://developers.google.com/web/tools/chrome-devtools/profile/evaluate-performance/rail and https://crbug.com/574343#c40 for more information.

kwalrath:
Is this caused by DartPad?

chalin:
Ah yes, that could very well be it!

kwalrath:
I don't see the Blink warning when I view those pages in Chrome Canary: Version 64.0.3261.0 (Official Build) canary (64-bit).

kwalrath:
DartPad's version of Polymer has been updated, but there are now **4** warnings in in https://www.dartlang.org/guides/get-started:

* [Deprecation] Styling master document from stylesheets defined in HTML Imports is deprecated, and is planned to be removed in M67, around May 2018. Please refer to https://goo.gl/EGXzpw for possible migration paths.
* imports.html:11890 This file is deprecated. Please use `iron-flex-layout/iron-flex-layout-classes.html`, and one of the specific dom-modules instead
* imports.html:12193 This file is deprecated. Please use `iron-flex-layout/iron-flex-layout-classes.html`, and one of the specific dom-modules instead
* [Deprecation] /deep/ combinator is no longer supported in CSS dynamic profile.It is now effectively no-op, acting as if it were a descendant combinator. /deep/ combinator will be removed, and will be invalid at M65. You should remove it. See https://www.chromestatus.com/features/4964279606312960 for more details.

For more info, see https://github.com/dart-lang/dart-pad/issues/791.

kwalrath:
[...]

PR #940, opened 2018-06-15
Embrace optional new/const in the language tour

kwalrath:
Contributes to #870, #507.

Staged: https://kw-www-dartlang-1.firebaseapp.com/guides/language/language-tour#using-constructors


chalin:
Hmm, if you open the staged page's Console, there are lots of errors there.

chalin:
Rebasing (to get https://github.com/dart-lang/site-www/pull/941), will help with the Travis errors.

examples/misc/test/language_tour/classes_test.dart
```diff
@@ -79,6 +84,28 @@ void main() {
...
+    // #docregion const-context-noconst
+    // Only one const, which establishes the constant context.
+    const pointAndLine2 = {
+      'point': [ImmutablePoint(0, 0)],
+      'line': [ImmutablePoint(1, 10), ImmutablePoint(-2, 11)],
+    };
+    // #enddocregion const-context-noconst
+
+    // Same instances!
+    assert(identical(pointAndLine1, pointAndLine2));
```
chalin: Suggestions:
- Use `expect()`
- `const` values are canonicalized, so you can use `==`

E.g.,

```dart
expect(pointAndLine1 == pointAndLine2, isTrue);
...
```

chalin:
This is what I see in the JS console (maybe these are from the embedded DartPad? cc @jcollins-g):
screen shot 2018-06-15 at 16 36 20

kwalrath:
I see what looks like the same console errors in https://www.dartlang.org/guides/language/language-tour. I think we need to remove all the embedded DartPads (in another PR) to fix this (#422).

examples/README.md
```diff
@@ -21,17 +21,46 @@ The original sources for the (large) samples were copied from the
...
+
+To update an analyzer results file:
+
+1. Run `./scripts/analyze-and-test-examples.sh  --save-logs`.
+2. Look at the diffs for the results files.
+3. If the diffs look good but some comments are missing,
+   add back the comments that are still relevant.
+4. Run `./scripts/analyze-and-test-examples.sh` to confirm that
+   your changes are good.
+
```
chalin: Thanks for adding these helpful instructions.

src/_guides/language/language-tour.md
```diff
@@ -2271,33 +2247,43 @@ https://gist.github.com/0cb25997742ed5382e4a
...
+Some classes provide constant constructors. To create a compile-time
+constant using a constant constructor, put the `const` keyword
+before the constructor name:
+
+<?code-excerpt "misc/test/language_tour/classes_test.dart (const)"?>
+{% prettify dart %}
+final p = const ImmutablePoint(2, 2);
+{% endprettify %}
+
 Constructing two identical compile-time constants results in a single,
```
chalin: I know that this is from before, but maybe we should link to [identical()](#) here?

The definition is already given at the end of this file:

    [identical()]: {{site.dart_api}}/{{site.data.pkg-vers.SDK.channel}}/dart-core/identical.html

src/_guides/language/language-tour.md
```diff
@@ -2309,6 +2295,37 @@ var b = const ImmutablePoint(1, 1);
...
+// Only one const, which establishes the constant context.
+const pointAndLine = {
+  'point': [ImmutablePoint(0, 0)],
+  'line': [ImmutablePoint(1, 10), ImmutablePoint(-2, 11)],
+};
+{% endprettify %}
+
+<aside class="alert alert-info" markdown="1">
+**Version note:** The `const` keyword became optional
+(within a constant context) in Dart 2.
```
chalin: Maybe drop the parentheses?

examples/misc/test/language_tour/classes_test.dart
```diff
@@ -61,7 +66,7 @@ void main() {
   test('const, identical, runtimeType', () {
     _test() {
       // #docregion const
-      var p = const ImmutablePoint(2, 2);
+      final p = const ImmutablePoint(2, 2);
```
chalin: Why this change from `var` to `final`?

kwalrath: I was wondering if you'd catch that. :) I don't remember why I changed it. I think it was a side effect of the DartPad changes I was making. I'll change it back.

src/_guides/language/language-tour.md
kwalrath: I'm going to leave this out, because I'd rather not add more words.

examples/misc/test/language_tour/classes_test.dart
chalin: :)
👍

jcollins-g:
@kwalrath Removing all the embedded dartpads seems like an unfortunate consequence. 

The errors look like they might be a consequence of having an embedded Dartpad, but exactly why isn't clear.  Filing an issue.

PR #942, opened 2018-06-18
Fix lint name unnecessary_brace_in_string_interps

a14n:


chalin:
Thanks for the PR Alexandre!

PR #945, opened 2018-06-19
Add new and const to /dart-2

kwalrath:
Staged: https://kw-staging-dartlang-2.firebaseapp.com/dart-2

Contributes to #870.

/cc @kevmoo


src/dart-2.md
```diff
@@ -20,6 +20,10 @@ The Dart language, libraries, build system, and web development tools have chang
 * [Dart's type system][sound Dart] is now sound.
   * [Fixing common type problems][Fixing Common Type Problems]
   * [Flutter announcement: Breaking Change: `--preview-dart-2` turned on by default][Leaf's email]
+* Some keywords used for object creation are now optional,
```
chalin: It would be a bit more technically accurate to write:

> Instance creation keywords are now generally optional,

PR #961, opened 2018-06-25
Update platforms.md

anders-sandholm:
Renaming the third platform from "Dart VM" to "Server". As discussed in #245.

### DIFF
--- a/src/_guides/platforms.md
+++ b/src/_guides/platforms.md
@@ -25,16 +25,16 @@ or server, there's a Dart solution for that.
 
   Write an app that runs in any modern web browser.
 
-- {:.card.narrow}[![Dart logo][]{: width="100"}][Dart VM]{:.no-automatic-external}
+- {:.card.narrow}[![Dart logo][]{: width="100"}][Server]{:.no-automatic-external}
   {:.logo}
 
-  ### [Dart VM][]
+  ### [Server][]
 
-  Write a command-line app or server.
+  Write a command-line app or server for the Dart VM.
 
 
 [Dart logo]: {% asset_path 'shared/dart/logo/default.svg' %}
-[Dart VM]: {{site.dart_vm}}
+[Server]: {{site.dart_vm}}
 [Flutter]: {{site.flutter}}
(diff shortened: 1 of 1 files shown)


src/_guides/platforms.md
```diff
@@ -25,16 +25,16 @@ or server, there's a Dart solution for that.
...
 
-- {:.card.narrow}[![Dart logo][]{: width="100"}][Dart VM]{:.no-automatic-external}
+- {:.card.narrow}[![Dart logo][]{: width="100"}][Server]{:.no-automatic-external}
   {:.logo}
 
-  ### [Dart VM][]
+  ### [Server][]
 
-  Write a command-line app or server.
+  Write a command-line app or server for the Dart VM.
```
chalin: Two comments about this sentence:

1. For the first part, how about:

    > Write a command-line or server-side app.

2. I'd be inclined to leave this sentence w/o the addition of "for the Dart VM".

    I _think_ that your intent was to somehow mention "Dart VM", but I'm unsure about the use of "for". A variant might be "... or server that runs on the Dart VM", but I still end up with the feeling that mentioning the Dart VM isn't that relevant.

@kwalrath WDYT?

src/_guides/platforms.md
```diff
@@ -25,16 +25,16 @@ or server, there's a Dart solution for that.
 
   Write an app that runs in any modern web browser.
 
-- {:.card.narrow}[![Dart logo][]{: width="100"}][Dart VM]{:.no-automatic-external}
+- {:.card.narrow}[![Dart logo][]{: width="100"}][Server]{:.no-automatic-external}
```
chalin: The second sentence of the opening paragraph further above reads:

> Whether you're creating a mobile app, web app, command-line script, or server, there's a Dart solution for that.

Questions:

- Do we still want to mention "command-line", or should it be removed so as to match the platform list?
- How about replacing "server" by "server-side app"?

kwalrath:
I'm going to merge this and quick-like do a matching PR for /install.

PR #966, opened 2018-06-26
run dartfmt --fix on examples

chalin:
- Run dartfmt --fix on examples, ensuring to leave  excerpts illustrate bad/old syntax untouched
- In Effective Dart Usage: highlight unnecessary new/const in "bad" examples

Staged at https://dartlang-org-staging-0.firebaseapp.com. In particular see:

- https://dartlang-org-staging-0.firebaseapp.com/effective-dart/usage#dont-use-new
- https://dartlang-org-staging-0.firebaseapp.com/effective-dart/usage#dont-use-const-redundantly
- https://dartlang-org-staging-0.firebaseapp.com/guides/language/language-tour#using-constructors (mentions optional new, and const contexts).

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -286,7 +286,7 @@ Any variable can have a constant value.
 // const [] creates an empty, immutable list (EIL).
 var foo = const []; // foo is currently an EIL.
 final bar = const []; // bar will always be an EIL.
-const baz = const []; // baz is a compile-time constant EIL.
+const baz = []; // baz is a compile-time constant EIL.
 
 // You can change the value of a non-final, non-const variable,
 // even if it used to have a const value.
@@ -516,7 +516,7 @@ const aConstString = 'a constant string';
 var aNum = 0;
 var aBool = true;
 var aString = 'a string';
-const aConstList = const [1, 2, 3];
+const aConstList = [1, 2, 3];
 
 const validConstString = '$aConstNum $aConstBool $aConstString';
 // const invalidConstString = '$aNum $aBool $aString $aConstList';
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -305,12 +305,12 @@ even if it used to have a const value:
 foo = [1, 2, 3]; // Was const []
 {% endprettify %}
 
-Being compile-time constants, you can't assign to a const variable:
+You can't change the value of a const variable:
 
 {:.fails-sa}
 <?code-excerpt "misc/lib/language_tour/variables.dart (cant-assign-to-const)"?>
 {% prettify dart %}
-baz = []; // Error: Constant variables can't be assigned a value.
+baz = [42]; // Error: Constant variables can't be assigned a value.
 {% endprettify %}
 
 For more information on using `const` to create constant values, see
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -259,11 +259,18 @@ Here's an example of creating and setting a final variable:
 <?code-excerpt "misc/lib/language_tour/variables.dart (final)"?>
 {% prettify dart %}
 final name = 'Bob'; // Without a type annotation
-// name = 'Alice';  // Uncommenting this causes an error
 final String nickname = 'Bobby';
 {% endprettify %}
 
-Use `const` for variables that you want to be compile-time constants. If
+You can't change the value of a final variable:
+
+{:.fails-sa}
+<?code-excerpt "misc/lib/language_tour/variables.dart (cant-assign-to-final)"?>
+{% prettify dart %}
+name = 'Alice'; // Error: a final variable can only be set once.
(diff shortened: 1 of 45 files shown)


src/_guides/language/language-tour.md
```diff
@@ -286,7 +286,7 @@ Any variable can have a constant value.
 // const [] creates an empty, immutable list (EIL).
 var foo = const []; // foo is currently an EIL.
 final bar = const []; // bar will always be an EIL.
-const baz = const []; // baz is a compile-time constant EIL.
+const baz = []; // baz is a compile-time constant EIL.
```
kwalrath: We need to have a note here about being able to omit the `const`.

src/_guides/language/language-tour.md
```diff
@@ -516,7 +516,7 @@ const aConstString = 'a constant string';
 var aNum = 0;
 var aBool = true;
 var aString = 'a string';
-const aConstList = const [1, 2, 3];
+const aConstList = [1, 2, 3];
```
kwalrath: Note here, too?

src/samples/index.md
```diff
@@ -209,10 +208,10 @@ You might use the `Spacecraft` class like this:
...
 <?code-excerpt "misc/test/samples_test.dart (use class)"?>
 {% prettify dart %}
-var voyager = new Spacecraft('Voyager I', new DateTime(1977, 9, 5));
+var voyager = Spacecraft('Voyager I', DateTime(1977, 9, 5));
 voyager.describe();
 
-var voyager3 = new Spacecraft.unlaunched('Voyager III');
+var voyager3 = Spacecraft.unlaunched('Voyager III');
 voyager3.describe();
 {% endprettify %}
```
kwalrath: In the [Read more] that follows, we should add "optional `new` and `const`" and delete "constant constructors".

chalin:
> We need to adjust the words in some places. In particular, https://dartlang-org-staging-0.firebaseapp.com/guides/language/effective-dart/usage has a bunch of occurrences of new in the text.

I had looked for those in the markdown files, I must have missed `usage.md`. Fixing that now.

src/_guides/language/language-tour.md
chalin: I wouldn't add it here. Most folks naturally write `const aConstList = [1, 2, 3]` (forgetting the `const` on the list), only to be reminded (in Dart 1) by the IDE/analyzer about it missing.

src/_guides/language/language-tour.md
chalin: I've reworked the examples and prose in the area of the tour. PTAL

chalin:
Staging server refreshed, PTAL at https://dartlang-org-staging-0.firebaseapp.com/guides/language/language-tour#final-and-const

src/_guides/language/language-tour.md
```diff
@@ -282,19 +289,28 @@ Any variable can have a constant value.
...
+{% prettify dart %}
+foo = [1, 2, 3]; // Was const []
+{% endprettify %}
+
+Being compile-time constants, you can't assign to a const variable:
+
+{:.fails-sa}
+<?code-excerpt "misc/lib/language_tour/variables.dart (cant-assign-to-const)"?>
+{% prettify dart %}
+baz = []; // Error: Constant variables can't be assigned a value.
```
kwalrath: Can we avoid using "assigned" in this comment, or is that in the error message?

chalin: That is the actual error message.

src/_guides/language/language-tour.md
```diff
@@ -282,19 +289,28 @@ Any variable can have a constant value.
...
 
-// You can't change the value of a final or const variable.
-// bar = []; // Unhandled exception.
-// baz = []; // Unhandled exception.
+<?code-excerpt "misc/lib/language_tour/variables.dart (reassign-to-non-final)"?>
+{% prettify dart %}
+foo = [1, 2, 3]; // Was const []
+{% endprettify %}
+
+Being compile-time constants, you can't assign to a const variable:
```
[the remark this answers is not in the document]
chalin: How about:

> While a const variable must be initialized, you can't assign to it at runtime:

src/_guides/language/language-tour.md
kwalrath: Ah, darn it all.

chalin: (FWIW, I think that "assigned" is the proper term here. As you pointed out earlier, the other occurrence of "=" is actually part of an initializing declaration.)

src/_guides/language/language-tour.md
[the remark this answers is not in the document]
kwalrath: "While" might be confusing. It'd be clearer to say "you can't change the value of a const value", which is true but the implementation goes further (you can't assign it the same value, either).

But maybe we should change the example to so that it _does_ use a different value... That's how most people would mess up, isn't it?

PR #969, opened 2018-06-27
Change the recommendation for doc comments

TedSander:
DartDoc will show setter comments as a property. It is only when there is a comment on the setter and a getter that the comment is ambiguous, and the getter is favored. Change the recommendation so that teams can comment the setter if it makes more sense for the readability of the code.

### DIFF
--- a/src/_guides/language/effective-dart/documentation.md
+++ b/src/_guides/language/effective-dart/documentation.md
@@ -244,8 +244,8 @@ int weekday;
 int get checkedCount => ...
 {% endprettify %}
 
-If there's both a setter and a getter, comment only the getter. That way,
-dartdoc will treat it like a variable.
+Avoid having a doc comment on the setter and the getter, as DartDoc will only show
+one (the one on the getter.)
 
 ### PREFER starting library or type comments with noun phrases.
 


munificent:
Is there a reason to let people be inconsistent here? Why not just always put it on the getter?

TedSander:
So there are a couple of reasons:

- When reading code/reviewing there are times when the comment has more context on the setter rather than the getter. The setter itself has more logic and is the one that contains details about what is happening with the value. Most of the time the getter is just a simple pass through. We read way more code than dart docs especially since they aren't available to us internally.
- There are times when we have a setter and no getter for angular components. (I know there is arguments that some don't like this, but the API of a component is a lot different than a Dart API.) So we would start the comment on the setter and then if a getter is eventually added we would need to move the comment.
-  For angular we put an annotation on the setter itself because it is the value that is getting set. The DartDoc is actually more documentation for that annotation.

```dart
  // This getter is actually inconsequential here.
  // In fact I would probably want to mark it as visible to the template only.
  String get slide => ...;

  // This is the important part
  /// Sets the slide direction.
  @Input()
  set slide(String value) {
  }
```

So I wonder why be prescriptive here? Especially if the original reason (DartDoc made it not into a property) is no longer valid.

PR #976, opened 2018-06-29
Language tour: operator-override example improvements

### LINKED ISSUE #311: Overridable operators example is confusing
Part of the example states:
`// v == (2, 3)`
`assert(v.x == 2 && v.y == 3);`

This, using the rest of the example as a basis, implies that the "==" operator has been overridden and a new Vector object can be created without the "new " keyword. Neither of which (to my limited knowledge) is true.

chalin:
Fixes #311 

Staged at https://dartlang-org-staging-0.firebaseapp.com/guides/language/language-tour#overridable-operators

cc @srawlins 

---

The original (problematic) sample code looked like this:

```dart
  final v = Vector(2, 3);
  final w = Vector(2, 2);
  ...
  // v + w == (4, 5)
  assert((v + w).x == 4 && (v + w).y == 5);

  // v - w == (0, 1)
  assert((v - w).x == 0 && (v - w).y == 1);
```

It is now:

```dart
  final v = Vector(2, 3);
  final w = Vector(2, 2);

  assert(v + w == Vector(4, 5));
  assert(v - w == Vector(0, 1));
```

The `Vector` class defines `==` and `hashCode` but the implementations are elided (since the section that contains this code is about operators). Instead, the interested reader is referred to another section covering `==` and `hashCode`.


srawlins:
Woohoo! Seems clear to me. Thanks @chalin !

src/_guides/language/language-tour.md
```diff
@@ -2816,62 +2816,6 @@ class EffectiveDoer extends Doer {
 Calling an abstract method results in a runtime error.
 
 
-#### Overridable operators
```
chalin: Moved this section to be immediately after "Overriding members"

PR #990, opened 2018-07-06
Minor fixes in docs

illright:
# Remove a letter that was probably mistyped

The section tells that we should use lowerCamelCase for constant names. Then there are some examples in other styles, one of which is `kDefaultTimeout`, which technically is lower camel case. Perhaps `DefaultTimeout` was actually meant to demonstrate UpperCamelCase.

# Correct lowerCamelCase in noSuchMethod

One of the links in that section refers to "nosuchMethod forwarding...", which is probably mistakenly capitalized. Changing it to "noSuchMethod", as everywhere else.

### DIFF
--- a/src/_guides/language/effective-dart/style.md
+++ b/src/_guides/language/effective-dart/style.md
@@ -174,7 +174,7 @@ class Dice {
 <?code-excerpt "misc/lib/effective_dart/style_bad.dart (const-names)"?>
 {% prettify dart %}
 const PI = 3.14;
-const kDefaultTimeout = 1000;
+const DefaultTimeout = 1000;
 final URL_SCHEME = RegExp('^([a-z]+):');
 
 class Dice {
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -3011,7 +3011,7 @@ and the dynamic type of the receiver has an implemention of `noSuchMethod()`
 that's different from the one in class `Object`.
 
 For more information, see the informal
-[nosuchMethod forwarding specification.](https://github.com/dart-lang/sdk/blob/master/docs/language/informal/nosuchmethod-forwarding.md)
+[noSuchMethod forwarding specification.](https://github.com/dart-lang/sdk/blob/master/docs/language/informal/nosuchmethod-forwarding.md)
 
 
 <a id="enums"></a>


illright:
I signed it!

kwalrath:
Thanks for the PR, @illright! @munificent, are you good with the Effective Dart fix? (Is the `k` in the BAD example because Effective Dart used to say something about prefixes?)

The build fail is valid: the code in the test doesn't match the code in the snippet any more. Once we're sure what the code should say, we (could be me) need to update the code in the /examples directory, as described in the [CONTRIBUTING file](https://github.com/dart-lang/site-www/blob/master/CONTRIBUTING.md#updating-code-samples).

src/_guides/language/effective-dart/style.md
```diff
@@ -174,7 +174,7 @@ class Dice {
 <?code-excerpt "misc/lib/effective_dart/style_bad.dart (const-names)"?>
 {% prettify dart %}
 const PI = 3.14;
-const kDefaultTimeout = 1000;
```
chalin: This is meant to be a _bad_ example.

chalin: > Is the k in the BAD example because Effective Dart used to say something about prefixes?

Oops, missed this comment: yes I believe so, even if we don't mention it in the prose. But I'll let Bob confirm.

munificent: The intent was to show that you should not use any sort of prefix to name a constant. How about leaving this example and also adding a separate guideline at the top:

```
### DON'T use Hungarian notation or any other prefixed letters on identifiers to identify their type or scope.

The type system is more than happy to tell you what the scope, mutability, and type of any identifier is, so there's no need to decorate its name yourself with that information.
```

chalin: SGTM, though it might make sense to end the **Identifiers** section with that guideline (so we don't have to include good and bad examples. @kwalrath WDYT?

illright: @munificent @chalin if the intent was to show that a prefix is not necessary, it definitely doesn't belong to that section. Also, it's pretty unclear what that prefix means, so as a reader, I can say that this intent may not be understood. Also I don't know what Hungarian notation is, so better add an example if you want to include a separate guideline :)
And finally, I still feel like the `lowerCamelCase` section should include `UpperCamelCase` as a bad example, therefore `DefaultTimeout` would make sense as to why it's bad.

munificent: > Also, it's pretty unclear what that prefix means

The Chrome team has a style (for their C++ code) of using a "k" prefix for constants, and that has bled into some Dart code primarily from Dart developers who used to write C++ in Chrome.

> Also I don't know what Hungarian notation is

https://en.wikipedia.org/wiki/Hungarian_notation

> I still feel like the lowerCamelCase section should include UpperCamelCase as a bad example, therefore DefaultTimeout would make sense as to why it's bad.

+1.

kwalrath:
I'm going to accept this PR and then create two followup PRs that fix the build and add the new section for prefixes.

munificent:
Thanks!

PR #991, opened 2018-07-06
Add note about whereType implementation status

kwalrath:
Staged: https://kw-www-dartlang-1.firebaseapp.com/guides/language/effective-dart/usage#do-use-wheretype-to-filter-a-collection-by-type

/cc @munificent @lrhn 

Contributes to dart-lang/sdk#32463.

### DIFF
--- a/src/_guides/language/effective-dart/usage.md
+++ b/src/_guides/language/effective-dart/usage.md
@@ -357,6 +357,9 @@ you don't care about the type, then use `toList()`.
 
 ### DO use `whereType()` to filter a collection by type.
 
+{% comment %}
+update-for-dart-2
+{% endcomment %}
 <aside class="alert alert-warning" markdown="1">
   **Before using `whereType()`, make sure it's implemented.**
   We expect `whereType()` to be added late in Dart&nbsp;2 development.
--- a/src/_guides/language/effective-dart/usage.md
+++ b/src/_guides/language/effective-dart/usage.md
@@ -357,6 +357,13 @@ you don't care about the type, then use `toList()`.
 
 ### DO use `whereType()` to filter a collection by type.
 
+<aside class="alert alert-warning" markdown="1">
+**Before using whereType, make sure it's implemented.**
+We expect `whereType()` to be added late in Dart&nbsp;2 development.
+For details, see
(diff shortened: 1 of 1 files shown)


src/_guides/language/effective-dart/usage.md
```diff
@@ -357,6 +357,13 @@ you don't care about the type, then use `toList()`.
 
 ### DO use `whereType()` to filter a collection by type.
 
+<aside class="alert alert-warning" markdown="1">
+**Before using whereType, make sure it's implemented.**
+We expect `whereType()` to be added late in Dart&nbsp;2 development.
+For details, see
+[SDK issue #32463.](https://github.com/dart-lang/sdk/issues/32463#issuecomment-402975456)
+</aside>
```
chalin: Nit: indent `<aside>` content.

src/_guides/language/effective-dart/usage.md
```diff
@@ -357,6 +357,16 @@ you don't care about the type, then use `toList()`.
...
 ### DO use `whereType()` to filter a collection by type.
 
+{% comment %}
+update-for-dart-2
+{% endcomment %}
+<aside class="alert alert-warning" markdown="1">
+  **Before using `whereType()`, make sure it's implemented.**
+  We expect `whereType()` to be added late in Dart&nbsp;2 development.
+  For details, see
+  [SDK issue #32463.](https://github.com/dart-lang/sdk/issues/32463#issuecomment-402975456)
```
munificent: That issue is closed already. Is it time to take this back out? :)

kwalrath: Once it's out in the real world, then yes! It looks like it was only added a few hours ago, though, and I can't tell the implementation status from the API docs.

@lrhn do you expect this to go out in dev.69? @Hixie, is there a milestone I can follow to see when Flutter has this?

lrhn: It should be available in the next release, whichever one that will be. So, yes.

Hixie: Should be in Flutter Beta 7.

kwalrath: OK, I'll set a reminder to check about this in the beta 7 timeframe.

PR #1000, opened 2018-07-12
Update Dart 2 constraint guidance for packages

mit-mit:
Our current guidance suggests an upper constraint of `<2.0.0`, which will cause resolution failures once Dart 2 ships on the stable channel.

### DIFF
--- a/src/dart-2.md
+++ b/src/dart-2.md
@@ -91,7 +91,7 @@ As a package owner, you need to do the following:
 
 * Follow the migration tips for the platforms that your package supports
   (see above).
-* Make sure your package passes Dart 2 analysis (see 'Run the analyzer' above)
+* Make sure your package passes Dart 2 analysis (see **Run the analyzer** above)
 * Make sure your package's users know how to report issues.
 * Respond quickly to issue reports.
 * If code changes aren't backward compatible,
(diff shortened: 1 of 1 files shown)


kwalrath:
Not sure why the build broke; looks like it was a flake. I've restarted it.

src/dart-2.md
```diff
@@ -107,41 +107,29 @@ or (if an API has been removed) to use an alternative 1.x API.
...
-# Backward incompatible change requires at least 2.0.0-dev.1.2:
-sdk: '>=2.0.0-dev.1.2 <2.0.0'
+# Works in Dart 2 only, starting with Dart 2 dev build 61
+sdk: '>=2.0.0-dev.61.0 <3.0.0'
 ```
 
-Eventually, you'll need to publish new versions of your packages to
-declare Dart 2 compatibility, most likely using a `<3.0.0` SDK constraint.
-Because incompatible changes might occur in any Dart 2 pre-release,
-don't declare Dart 2 compatibility until we announce that it's safe to do so.
```
kwalrath: How are we making this announcement?

src/dart-2.md
```diff
@@ -107,41 +107,29 @@ or (if an API has been removed) to use an alternative 1.x API.
...
-
-When you update an existing package or publish a new one,
-specify an upper constraint of `<2.0.0` for the SDK version. Examples:
+Once your package passes Dart 2 analysis, please update the upper constraint
+to declare that the package is compatible with Dart 2:
 
 ```yaml
-# Works in 1.20.1+; might work in 2.0.0-dev:
-sdk: '>=1.20.1 <2.0.0'
+# Works in Dart 1, build 20.1+ and forward, and works in Dart 2
```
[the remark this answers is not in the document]
kwalrath: Or if you want to clarify that Dart 1.20.1+ doesn't imply that it works in Dart 2, how about something like this:

```
# Works in Dart 1.x releases (starting with 1.20.1) and in Dart 2.
```

src/dart-2.md
mit-mit: On https://medium.com/dartlang and https://news.dartlang.org/. I shared the draft with you for review in email.

mit-mit:
I'm going to merge this, as the new Pub site we are deploying right now links to it. If more fixes are needed, let me know and I'll create a new PR for those immediately

PR #1006, opened 2018-07-16
Add prefix item to Effective Dart

kwalrath:
Followup to #990

I changed the wording, so please make sure it says when you meant to say. Should we add more examples? 

Staged:
https://kw-www-dartlang-1.firebaseapp.com/guides/language/effective-dart/style#dont-use-prefix-letters-to-indicate-type-mutability-or-scope
https://kw-www-dartlang-1.firebaseapp.com/guides/language/effective-dart#style

### DIFF
--- a/src/_guides/language/effective-dart/style.md
+++ b/src/_guides/language/effective-dart/style.md
@@ -228,6 +228,24 @@ Db
 {% endprettify %}
 
 
+### DON’T use prefix letters to indicate type, mutability, or scope.
+
+The type system can tell you all about these for any identifier,
+so there's no need to use
+[Hungarian notation](https://en.wikipedia.org/wiki/Hungarian_notation)
+or any other prefix scheme to encode this information.
+
+{:.good-style}
+{% prettify dart %}
+defaultTimeout
+{% endprettify %}
+
+{:.bad-style}
+{% prettify dart %}
+kDefaultTimeout
+{% endprettify %}
+
+
 ## Ordering
 
 To keep the preamble of your file tidy, we have a prescribed order that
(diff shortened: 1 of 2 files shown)


src/_guides/language/effective-dart/style.md
```diff
@@ -228,6 +228,24 @@ Db
 {% endprettify %}
 
 
+### DON’T use prefix letters to indicate type, mutability, or scope.
```
munificent: I worry this is too specific and readers may assume it is OK to use prefixes for other things. How about:

```
### DON'T decorate identifiers with letters that describe the kind of thing being named.
```

?

src/_guides/language/effective-dart/style.md
```diff
@@ -228,6 +228,24 @@ Db
...
+
+The type system can tell you all about these for any identifier,
+so there's no need to use
+[Hungarian notation](https://en.wikipedia.org/wiki/Hungarian_notation)
+or any other prefix scheme to encode this information.
```
munificent: It's not *quite* the type system that tracks things like scope and mutability. Maybe:

```
[Hungarian notation](https://en.wikipedia.org/wiki/Hungarian_notation) and other
schemes arose in the time of BCPL when the compiler didn't do much to help you
understand your code. In Dart, the language will happily tell you the type,
scope, mutability, and other properties of your declarations, so there is no
reason to encode it in the name.
```

src/_guides/language/effective-dart/style.md
kwalrath: I'd like to be a little less general and use fewer words, if possible. Let's brainstorm offline...

kwalrath:
@munificent ptal, since I didn't use your exact wording. (I made it more boring because [style](https://developers.google.com/style/tone).)

munificent:
👍

PR #1014, opened 2018-07-20
Single quotes on raw string example in lang tour

RedBrogdon:
This is a tiny nit, but I just ran into trouble trying to figure out what this statement from `built_value` was doing:

      StandardJsonPlugin({this.discriminator = r'$', this.valueKey = ''});

I'd never seen the "leading r" syntax for a raw string before. I Cmd-F'ed the site for "r'" and came up empty because the example is using double quotes instead. If we generally encourage single quotes as the default (which is implied by the other examples), could we change this example to match?

### DIFF
--- a/examples/misc/lib/language_tour/built_in_types.dart
+++ b/examples/misc/lib/language_tour/built_in_types.dart
@@ -34,7 +34,7 @@ void miscDeclAnalyzedButNotTested() {
 
   {
     // #docregion raw-strings
-    var s = r"In a raw string, even \n isn't special.";
+    var s = r'In a raw string, not even \n gets special treatment.';
     // #enddocregion raw-strings
   }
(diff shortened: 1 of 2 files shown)


kwalrath:
Good change! Can you update the corresponding example ([misc/lib/language_tour/built_in_types.dart](https://github.com/dart-lang/site-www/blob/master/examples/misc/lib/language_tour/built_in_types.dart#L37)) as well, so the test will pass? (If it's not convenient, I can do it.)

RedBrogdon:
That hurts. Not only did I fail to read the contributor's guide section on updating the code for the tests, but my two-character change was in fact broken.

I've altered the text of the example slightly in order to remove the offending apostrophe and updated the test code to match.

PR #1022, opened 2018-07-24
Make article obsolescence warning less scary

kwalrath:
Contributes to dart-lang/sdk#33960

Staged at URLs like https://kw-staging-dartlang-2.firebaseapp.com/articles/libraries/zones

### DIFF
--- a/src/_includes/page-header.html
+++ b/src/_includes/page-header.html
@@ -1,8 +1,8 @@
 <header id="page-header">
   {% include navigation-main.html %}
   {% if page.obsolete == true %}
-  <div class="alert alert-danger">
-    <h4 class="text-center">Content of this page might be obsolete.</h4>
+  <div class="alert alert-warning">
+    <h4 class="text-center">Some of the content of this page might be obsolete.</h4>
   </div>
   {% endif %}
 </header>
--- a/src/_includes/page-header.html
+++ b/src/_includes/page-header.html
@@ -2,7 +2,9 @@
   {% include navigation-main.html %}
   {% if page.obsolete == true %}
   <div class="alert alert-warning">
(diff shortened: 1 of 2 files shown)


src/_includes/page-header.html
```diff
@@ -1,8 +1,8 @@
 <header id="page-header">
   {% include navigation-main.html %}
   {% if page.obsolete == true %}
-  <div class="alert alert-danger">
-    <h4 class="text-center">Content of this page might be obsolete.</h4>
+  <div class="alert alert-warning">
+    <h4 class="text-center">Some of the content of this page might be obsolete.</h4>
```
sjindel-google: Can we say "outdated" instead of "obsolete"?

chalin: How about "... page might be outdated or obsolete."

chalin:
https://github.com/dart-lang/sdk/issues/33960 mentions that the obsolete/outdated pages will be triaged. That is good. We might find, after this exercise, that we need distinct banners: one for outdated content and another for obsolete (archived) content. Although, maybe archived content will simply be removed (since the pages are found under the 1.x version of the site).

src/_includes/page-header.html
chalin: Some reformatting would be needed here, but we can do that later.
Kathy, consider adding a Material alert [icon](https://material.io/tools/icons/?style=baseline).

kwalrath: _Outdated_ and _obsolete_ are synonyms; I'm fine with using _outdated_ or maybe _out of date_.

kwalrath:
I changed the wording to "out of date". I didn't use an icon, since there were alignment problems and I didn't feel like spending time on something that'll be replaced soon.

PR #1026, opened 2018-07-25
Analysis options: add lints for unnecessary cont and new, replace symlinks with `include`

chalin:


chalin:
Tests failing because of https://github.com/dart-lang/test/issues/915. Will wait for a fix.

chalin:
Travis is green once again. Merging.

PR #1040, opened 2018-08-02
Library tour: drop sentence mentioning import prefix

### LINKED ISSUE #1039: Library tour:"The following examples use the prefix math to make clear which top-level functions and constants are from the Math library:"
The description says that there is a ‘math’ prefix at function.
but new version examples does not have prefix.
it makes  confused.

chalin:
This is a followup to https://github.com/dart-lang/site-www/pull/480#discussion_r157355550, where we dropped the import prefix.

Fixes #1039 

cc @kwalrath


Sfshaza:
LGTM

PR #1042, opened 2018-08-03
Dartpad prose updates

chalin:
(Dart 2.0 prep)

As of at least a couple of months ago:

- Assertion checking is enabled by default in DartPad (https://github.com/dart-lang/dart-pad/issues/310); drop warning to the contrary.
- SDK version is shown directly in DartPad editor footer

cc @kevmoo @kwalrath @Sfshaza


chalin:
Travis is green and I need to pile on a few more minor PRs, so I'll go ahead and merge now.

PR #1056, opened 2018-08-08
Tweak banner, add link to announcement.

kwalrath:
Staged: https://kw-staging-dartlang-2.firebaseapp.com

@chalin, how do you stage changes to the banner? I temporarily moved the `{% if site.url == site.www %}` to the bottom of banner.html, deployed, then moved it back.

### DIFF
--- a/src/_includes/banner.html
+++ b/src/_includes/banner.html
@@ -2,8 +2,9 @@
 {% comment %}Drop the div .alert classes to revert to the default banner styling.{% endcomment -%}
 <div class="banner alert alert-info">
   <p class="banner__text">
-    <a href="/dart-2"><b>Dart 2.0</b></a> is now available!
-    <a href="/install">Install Dart</a>.
+    <a href="https://medium.com/dartlang/dart-2-stable-and-the-dart-web-platform-3775d5f8eac7" target="_blank">Dart 2.0 is now stable!</a>
+    <a href="/dart-2#migration">Migrate your code</a> or
+    <a href="/install">install Dart</a>.
     {% comment %}
     Still using Dart 1.x? See the
     <a href="{{site.prev-url}}" class="no-automatic-external" target="_blank" rel="noopener">archived site</a>


chalin:
> how do you stage changes to the banner?

It wasn't an issue the last time. We can change that condition if it no longer suites our purpose, or even drop the conditional altogether.

src/_includes/banner.html
```diff
@@ -2,8 +2,9 @@
 {% comment %}Drop the div .alert classes to revert to the default banner styling.{% endcomment -%}
 <div class="banner alert alert-info">
   <p class="banner__text">
-    <a href="/dart-2"><b>Dart 2.0</b></a> is now available!
-    <a href="/install">Install Dart</a>.
+    <a href="https://medium.com/dartlang/dart-2-stable-and-the-dart-web-platform-3775d5f8eac7" target="_blank">Dart 2.0 is now stable!</a>
```
chalin: - Consider adding `class="no-automatic-external"` and `rel="noopener"`?
- Also, how about putting **Dart 2** and/or **stable** in bold?

(We really should make this a markdown file, ... but in another PR.)

src/_includes/banner.html
```diff
@@ -2,8 +2,9 @@
...
+    <a href="/dart-2#migration">Migrate your code</a> or
```
chalin: Good addition.

kwalrath:
Maybe it's time to drop the conditional.

src/_includes/banner.html
```diff
@@ -2,8 +2,9 @@
...
+    <a href="/install">install Dart</a>.
     {% comment %}
```
chalin: We can probably drop this comment now that are mentioning migration once again in the banner.

chalin:
Oh, you merged already. Oh well, I guess that the other issues can be addressed separately (maybe along with converting the banner include to a markdown file).

kwalrath:
Yeah, I wanted the banner updated tout de suite! I'll make those other changes today or tomorrow.

src/_includes/banner.html
kwalrath: I rather like having the external link, since it's otherwise hard to tell that the text has any links at all. `rel="noopener"` seems like a reasonable thing, though. (Do we ever automatically add that?)

chalin: I'd like to generate those icons statically. At the same time adding `target="_blank" rel="noopener"` seems like an excellent idea.

chalin: #1058.

PR #1057, opened 2018-08-08
followup on banner work

kwalrath:
Staged at https://kw-www-dartlang-1.firebaseapp.com. The only visible change is that "Dart 2" is now bold, and I dropped the ".0".

### DIFF
--- a/src/_includes/banner.html
+++ b/src/_includes/banner.html
@@ -1,15 +1,8 @@
-{% if site.url == site.www %}
 {% comment %}Drop the div .alert classes to revert to the default banner styling.{% endcomment -%}
 <div class="banner alert alert-info">
   <p class="banner__text">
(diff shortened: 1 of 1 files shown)


src/_includes/banner.html
```diff
@@ -1,15 +1,8 @@
-{% if site.url == site.www %}
 {% comment %}Drop the div .alert classes to revert to the default banner styling.{% endcomment -%}
 <div class="banner alert alert-info">
   <p class="banner__text">
-    <a href="https://medium.com/dartlang/dart-2-stable-and-the-dart-web-platform-3775d5f8eac7" target="_blank">Dart 2.0 is now stable!</a>
+    <a href="https://medium.com/dartlang/dart-2-stable-and-the-dart-web-platform-3775d5f8eac7" rel="noopener"><b>Dart 2</b> is now stable!</a>
```
chalin: You'll need both: `target="_blank" rel="noopener"`

PR #1060, opened 2018-08-09
Examples: drop remaining unnecessary const

### LINKED ISSUE #972: Library tour: const or no const for Durations?
The following code appears in the [Dates and times](https://www.dartlang.org/guides/libraries/library-tour#dates-and-times) section of the Library Tour:

https://github.com/dart-lang/site-www/blob/36ea96c7d2a8857d7046df6c2b42062c52f9f716/examples/misc/test/library_tour/core_test.dart#L606-L612

Notice the two `const` `Duration`s. I'd be tempted to drop them, but chose not to do so during my last sweep.

@munificent @kwalrath @TedSander - any opinions about this?

More broadly: should we generally be promoting (via our example code) use of `const` for (immutable) types like Duration? Maybe we need a separate guideline on when to use `const` (esp. in light of issues like https://github.com/dart-lang/sdk/issues/33430, that highlight that const canonicalization is lazy, so IMHO it isn't that clear cut what the runtime cost/benefits are of using `const`).

chalin:
Fixes #972 

cc @kwalrath

### DIFF
--- a/examples/misc/lib/tutorial/cat_no_hash.dart
+++ b/examples/misc/lib/tutorial/cat_no_hash.dart
@@ -7,7 +7,7 @@ Future<void> main(List<String> args) async {
   var lines = file
       .openRead()
       .transform(utf8.decoder) //!<br>
-      .transform(const LineSplitter());
+      .transform(LineSplitter());
   await for (var line in lines) {
     if (!line.startsWith('#')) print(line);
   }
--- a/examples/httpserver/bin/number_guesser.dart
+++ b/examples/httpserver/bin/number_guesser.dart
@@ -9,7 +9,7 @@ import 'dart:convert';
 import 'dart:io';
 import 'dart:math';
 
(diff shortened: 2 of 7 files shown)


examples/misc/lib/tutorial/cat_no_hash.dart
```diff
@@ -7,7 +7,7 @@ Future<void> main(List<String> args) async {
   var lines = file
       .openRead()
       .transform(utf8.decoder) //!<br>
-      .transform(const LineSplitter());
+      .transform(LineSplitter());
```
kwalrath: Is this really optional here?

chalin: It isn't optional in the sense that the context isn't a const context, but the const qualifier isn't functionally necessary.

kwalrath:
(I saw #972 after I saw this PR. Carry on.)

PR #1062, opened 2018-08-09
Added entry to who uses dart

jimmyff:
There appears to be no order to this list and it's rather large. It may be worth categorising these in to Tooling, Products & services? You could feature the high-profile projects like flutter & adwords and then order the smaller projects alphabetically? Shall I implement and send a PR?

### DIFF
--- a/src/community/who-uses-dart.md
+++ b/src/community/who-uses-dart.md
@@ -63,6 +63,9 @@ Google internal sales tool
 [drone.io](http://drone.io)
 : Continuous integration with support for testing Dart apps.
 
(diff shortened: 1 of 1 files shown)


jimmyff:
I signed it!

kwalrath:
There's a Flutter showcase page (https://flutter.io/showcase/) that this page should point to. It has a few apps but also points to itsallwidgets.com.

Maybe this page should just point to pages on other sites, since they seem to be doing a good job (at least for Flutter). We might want to have similar pages for web apps and other non-Flutter applications, in that case.

PR #1063, opened 2018-08-09
Remove (beta) suffix from Dart 2.0 in version menu

mit-mit:


src/_includes/_version-menu-items.html
```diff
@@ -1,5 +1,5 @@
 {% assign dart1status = 'archive' %}
-{% assign dart2status = 'beta' %}
+{% assign dart2status = '' %}
```
chalin: Ah yes, missed that, thanks! Beta should be changed to `stable`.

mit-mit: done

PR #1064, opened 2018-08-09
Upgrade to Node 10

chalin:
Temporarily test out site-shared via a branch.


chalin:
@kwalrath - FYI, all is good here, but I want to finish migrating webdev before merging for both sites.

PR #1072, opened 2018-08-11
Fix async tutorial, and give brief explanations of Future<void>

### LINKED ISSUE #1070: Async articles are out of date / wrong
_From @MichaelRFairhurst on August 10, 2018 17:15_

This gives a lot of no longer accurate advice:

https://www.dartlang.org/tutorials/language/futures

See a confused user here: https://www.reddit.com/r/dartlang/comments/965b8u/async_in_dart/?utm_content=full_comments&utm_medium=message&utm_source=reddit&utm_name=frontpage

Not sure who to ping: @kwalrath @chalin @munificent or if ithis is the right repo for this issue

_Copied from original issue: dart-lang/sdk#34123_

MichaelRFairhurst:
Fix #1070

### DIFF
--- a/examples/misc/lib/tutorial/daily_news.dart
+++ b/examples/misc/lib/tutorial/daily_news.dart
@@ -16,23 +16,23 @@ import 'dart:async';
 
 // #docregion sync
 // Synchronous code
-_printDailyNewsDigestSync() {
+void _printDailyNewsDigestSync() {
   String news = _gatherNewsReportsSync(); // Can take a while.
   print(news);
 }
 // #enddocregion sync
 
 // #docregion main-async
-Future _printDailyNewsDigestAsync() async {
+Future<void> _printDailyNewsDigestAsync() async {
   String news = await _gatherNewsReportsAsync();
   print(news);
 }
 // #enddocregion main-async
 
 // #docregion main-future-api
-_printDailyNewsDigestAsyncUsingFutureAPI() {
+Future<void> _printDailyNewsDigestAsyncUsingFutureAPI() {
   final future = _gatherNewsReportsAsync();
-  future.then((news) => print(news));
+  return future.then((news) => print(news));
 }
 // #enddocregion main-future-api
 
@@ -93,7 +93,7 @@ Future<String> _gatherNewsReportsAsync() => newsStream.first;
 //
 // import 'dart:html';
 //
-// Future _gatherNewsReportsFromServer() => HttpRequest.getString(
+// Future<String> _gatherNewsReportsFromServer() => HttpRequest.getString(
 //      'https://www.dartlang.org/f/dailyNewsDigest.txt',
 //    );
 // #enddocregion main-async, main-future-api
@@ -102,7 +102,7 @@ Future<String> _gatherNewsReportsAsync() => newsStream.first;
 // Extra function variants below here
 
 // #docregion try-catch
-_printDailyNewsDigestAsyncWithTryCatch() async {
+Future<void> _printDailyNewsDigestAsyncWithTryCatch() async {
   try {
     String news = await _gatherNewsReportsAsync();
     print(news);
@@ -113,9 +113,9 @@ _printDailyNewsDigestAsyncWithTryCatch() async {
 // #enddocregion try-catch
 
 // #docregion main-future-api-using-braces
-_printDailyNewsDigestAsyncUsingFutureApiAndBraces() {
+Future<void> _printDailyNewsDigestAsyncUsingFutureApiAndBraces() {
   final future = _gatherNewsReportsAsync();
-  future.then((news) {
+  return future.then((news) {
     print(news);
     // Do something else...
   });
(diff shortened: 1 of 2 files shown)


kwalrath:
Thanks, Mike! It looks like some of the source examples need to be updated. (Travis says you need to scripts/refresh-code-excerpts.sh, but that would just reverse your changes.)

I think the affected files might be (under examples):
misc/lib/tutorial/daily_news.dart
misc/test/tutorial/futures_test.dart
misc/lib/tutorial/misc.dart

MichaelRFairhurst:
Thanks Kathy, was able to get it working. Nice check, that caught places that the tutorial was inconsistent once I consistently fixed the examples! :)

kwalrath:
I've staged the changes at https://kw-www-dartlang-1.firebaseapp.com/tutorials/language/futures. Looking at the text now...

kwalrath:
This is going to take a little while, but I've started working on updating the image. @MichaelRFairhurst what do you think of this? (I'd remove 4 from your list or combine it with 3 because it isn't something we can label.)

image

chalin:
@kwalrath - shall I wait for you to be done with your review?

src/_tutorials/language/futures.md
```diff
@@ -178,30 +179,41 @@ daily news digest is printed.
...
    forecast is printed.
-1. The body of the `printDailyNewsDigest()` function starts executing.
-
-1. After reaching the await expression (`await gatherNewsReports()`) and
-   calling `gatherNewsReports()`, the program pauses, waiting for the Future
-   returned by `gatherNewsReports()` to complete.
-1. Once that Future completes, execution of `printDailyNewsDigest()`
-   continues, printing the news.
+1. Once the Future returned by `gatherNewsReports()` completes (in this case, it
+   will complete immediately), execution of `printDailyNewsDigest()` continues,
```
kwalrath: will complete -> completes

src/_tutorials/language/futures.md
```diff
@@ -178,30 +179,41 @@ daily news digest is printed.
...
+   printing the news.
 1. When the `printDailyNewsDigest()` function body has completed executing,
    the Future that it originally returned completes, and the app exits.
 
 <aside class="alert alert-info" markdown="1">
-  **Note:** If an async function doesn't explicitly return a value, it returns
-  a Future wrapped around a null value.
+Note the return types carefully. Instead of `gatherNewsReports()` returning a
```
kwalrath: Good stuff, but I have a bit of rewriting (because I'm me, and to avoid using method names as verbs). How about:

Note the return types. The return type of `gatherNewsReports()` is `Future<String>`, which means that it returns a Future that completes with a String value. The `printDailyNewsDigest()` function, which doesn't explicitly return a value, has the return type `Future<void>`.

src/_tutorials/language/futures.md
```diff
@@ -367,17 +381,21 @@ written in a couple different ways.
 
   <?code-excerpt "misc/lib/tutorial/daily_news.dart (main-future-api-pass-print)"?>
   {% prettify dart %}
-  printDailyNewsDigest() =>
+  Future<void> printDailyNewsDigest() =>
       gatherNewsReports().then(print);
   {% endprettify %}
 
+If you use `.then()` on a `Future<void>`, you still need to specify a parameter
```
kwalrath: Let's rewrite this slightly. I don't think we need to sell them on the feature, but it'd be good to _show_ them how to use it. Maybe something like this:

----------------

If you use `then()` on a `Future<void>` object, you still need to specify a parameter to `then()`'s callback:

```
final future = printDailyNewsDigest();
return future.then((_)) { // Unused parameter (named `_`, by convention)
  // Code that doesn't use the `_` parameter...
});
```

----------------

It'd be best if we could add this example to the examples/ directory, so we can check the syntax.

examples/misc/lib/tutorial/daily_news.dart
```diff
@@ -102,9 +104,9 @@ Future<String> _gatherNewsReportsAsync() => newsStream.first;
 // Extra function variants below here
 
 // #docregion try-catch
-_printDailyNewsDigestAsyncWithTryCatch() async {
+Future<void> _printDailyNewsDigestAsyncWithTryCatch() async {
```
kwalrath: I'll need to update the gists to use the new code.

MichaelRFairhurst:
I like it @kwalrath !

Yeah, 3 & 4 are tricky.

I almost imagine it like a line like this:

(2) -> `gatherNewsReports()` -> `newsStream.first` -> `await` -> `printWinningLotteryNumbers` -> (6) `return` -> `String news =` -> (8)

It also looks like the staged site shows old code still. Do I need to update dartpad separately than this PR? One such change that didn't show up is "String news" to "var news." If I can get that in DartPad, that should be part of the graphic too I think

kwalrath:
Our comments crossed!

> It also looks like the staged site shows old code still. Do I need to update dartpad separately than this PR? One such change that didn't show up is "String news" to "var news." If I can get that in DartPad, that should be part of the graphic too I think

The DartPads are driven by gists, which we currently update (or recreate) by hand. E.g. if the src="{{site.custom.dartpad.embed-dart-prefix}}?id=a345206b09cc09eee1753fab11ec006b&horizontalRatio=99&verticalRatio=73", then the gist is http://gist.github.com/a345206b09cc09eee1753fab11ec006b, which is owned by @chalin.

Patrice or I can easily update our gists. For other gists, I usually just copy/fork the existing gist, edit it, and update the page to refer to the new gist number.

kwalrath:
@MichaelRFairhurst and I just talked offline, and we've agreed that I'll take over this PR (or rather, a PR based on it), now that we're getting into the wordsmithing stage.

PR #1074, opened 2018-08-13
Recommend SDK lower constraint of 2.0.0.

sortie:
The pub website is now punishing packages for not having an SDK constraint, make it recommended to have such a constraint. Additionally, now that Dart 2 stable is out, recommend packages use the SDK constraint `>=2.0.0 <3.0.0`. Add such a constraint to the example.

Not sure if the 'dependency_overrides` should be in the example pubspec, it seems like something that could confuse newcomers while having small value for most users.

Why is carret syntax not recommended for the SDK constraint? Wouldn't `^2.0.0` be equal to `>=2.0.0 <3.0.0`?

### DIFF
--- a/src/tools/pub/pubspec.md
+++ b/src/tools/pub/pubspec.md
@@ -51,7 +51,7 @@ A pubspec can have the following fields:
   [_Learn more._](#dependencies)
 
 `environment`
-: Optional. Can be used to require a specific version of the Dart SDK.
+: Optional. It should be used to require a compatible version of the Dart SDK.
   [_Learn more._](#sdk-constraints)
 
 `executables`
@@ -86,6 +86,8 @@ description: >-
 author: Natalie Weizenbaum <nweiz@google.com>
 homepage: https://newtify.dartlang.org
 documentation: https://docs.newtify.com
+environment:
+  sdk: '>=2.0.0 <3.0.0'
 dependencies:
   efts: ^2.0.4
   transmogrify: ^0.4.0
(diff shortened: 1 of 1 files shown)


kwalrath:
Thanks very much, @sortie. Before reviewing the text, I'll wait for @kevmoo or @mit-mit to weigh in on caret syntax. (I think you're right about dependency_overrides.)

kevmoo:
Caret syntax was not recommended historically because it was added ~ Dart 1.8 (?) so there was a chance that an old SDK would barf on the SDK constraint.

I wonder if we should care at all about that anymore. CC @nex3 who has all of the context here

nex3:
We can't allow the caret syntax in SDK constraints in particular because that will *crash* older versions of pub. They use the SDK constraint to determine whether they should even parse the rest of the pubspec, which means that the format of that constraint is the one thing that can't ever be modified.

sortie:
Thanks for the explanation Natalie, that makes sense.

Kevin, the pub website doesn't currently require the SDK constraint to mark it as a Dart 2 package. If it is blank, it is still considered a Dart 2 package if it passes analysis. For instance <https://pub.dartlang.org/packages/font_awesome_flutter#-analysis-tab-> is an interesting case where the package contains resources and minimal code. It is something that we could consider doing (there is a package score penalty for not allowing Dart 2 already, but we don't hard fail).

If we require the SDK constraint, we should change "Optional. It should be used to require a compatible version of the Dart SDK." to "Required. It must be used to require a compatible version of the Dart SDK. Other uses are optional.".

kevmoo:
> If we require the SDK constraint, we should change "Optional. It should be used to require a compatible version of the Dart SDK." to "Required. It must be used to require a compatible version of the Dart SDK. Other uses are optional.".

YES!

PR #1079, opened 2018-08-15
Update the Futures tutorial

kwalrath:
This PR incorporates and adds to PR #1072 by @MichaelRFairhurst.

A couple of changes affect other pages:
* the workaround for a DartPad issue (dart-lang/dart-pad#847)
* the version note about async-await

Staged: https://kw-staging-dartlang-2.firebaseapp.com/tutorials/language/futures


src/_tutorials/language/futures.md
```diff
@@ -367,17 +382,31 @@ written in a couple different ways.
...
   {% endprettify %}
 
+You need to provide an argument to `then()`'s callback,
+even if the Future is of type `Future<void>`:
+
+<?code-excerpt "misc/lib/tutorial/daily_news.dart (main-future-api-then-no-arg)"?>
+```dart
+Future<void> future = printDailyNewsDigest();
+
+return future.then((_) {
```
chalin: Consider dropping the `return` keyword.

src/_tutorials/language/futures.md
```diff
@@ -367,17 +382,31 @@ written in a couple different ways.
...
+  // Code that doesn't use the `_` parameter...
```
chalin: You you instead want to mention the convention behind use of `_` (or link to a section that does in the prose before this code excerpt)?

src/_tutorials/language/futures.md
```diff
@@ -175,45 +174,59 @@ program prints.  If `gatherNewsReports()` takes a little while to complete its
...
-  **Note:** If an async function doesn't explicitly return a value, it returns
-  a Future wrapped around a null value.
-</aside>
+Note that an async function starts executing right away (synchronously).
+The function pauses execution and returns an uncompleted Future
+when it reaches the first occurrence of any of the following:
+
+* The function's first `await` expression (after the function
+  gets the uncompleted Future from that expression).
+* The function's `return` statement.
```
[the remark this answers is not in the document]
kwalrath: What if the function calls another (synchronous) function? Does that cause execution to pause?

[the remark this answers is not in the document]
chalin: That's a good question. I don't know.

The point I was trying to make is that a function might have more than one `return` statement.

kwalrath:
@MichaelRFairhurst could you take a look at Patrice's comments? I'm not sure what to do...

src/_tutorials/language/futures.md
```diff
@@ -269,9 +282,11 @@ https://gist.github.com/chalin/5e7eb16cd725f0a51926b85e3caa64b5
...
-printDailyNewsDigest() {
+Future<void> printDailyNewsDigest() {
   final future = gatherNewsReports();
-  future.then((news) => print(news));
+  return future.then((news) => print(news));
+  // You don't *have to* return the future here. But if you don't, callers can't
+  // await it.
 }
 
 main() {
```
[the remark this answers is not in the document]
kwalrath: @MichaelRFairhurst what's the usual style?

src/_tutorials/language/futures.md
```diff
@@ -269,9 +282,11 @@ https://gist.github.com/chalin/5e7eb16cd725f0a51926b85e3caa64b5
 
 import 'dart:async';
 
-printDailyNewsDigest() {
+Future<void> printDailyNewsDigest() {
   final future = gatherNewsReports();
-  future.then((news) => print(news));
+  return future.then((news) => print(news));
+  // You don't *have to* return the future here. But if you don't, callers can't
+  // await it.
```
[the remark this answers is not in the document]
kwalrath: Except that the way you can call them will be different...

src/_tutorials/language/futures.md
```diff
@@ -70,15 +70,17 @@ Such functions return their value using a [Future.][Future]
 
 ## What is a Future? {#what-is-a-future}
 
-A Future represents a means for getting a value sometime in the future. When a
-function that returns a Future is invoked, two things happen:
+A Future represents an asynchronous operation that, optionally,
+returns a value. When a function that returns a Future is invoked,
```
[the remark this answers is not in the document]
kwalrath: _Expression_ sounds too lightweight to me — like it's a small Dart unit of code and you can't even have a Future execute a whole statement.

src/_tutorials/language/futures.md
chalin: No :-/ ... I remember search for it a while back and having a hard time finding it.

kwalrath: I just grepped the site-www source code and can't find it there. A web search for [[unused dart parameter underscore](https://www.google.com/search?q=unused+dart+parameter+underscore)] came up with dart-lang/sdk#23050 and https://stackoverflow.com/questions/33146619. So it's a longstanding convention but, as far as I can tell, an undocumented one.

@munificent did you consider documenting `_` as a convention for unused parameter names?

src/_tutorials/language/futures.md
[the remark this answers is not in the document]
MichaelRFairhurst: While its technically true that a Future always returns a value, a void value is one that you aren't supposed to use. The fact that there's something there is more of an implementation detail, which is fully specified so that you can grab it in an emergency, and so that we don't have to jump through some technical hoops completely guaranteeing its vacuousness.

Also agree that expression is too lightweight for this.

You could say "optionally returns a usable value," or "optionally passing along a value" (since a `Future<void>` indicates that the value is not supposed to be read, its not really being "passed along"?) or "optionally including a usable result" ?

src/_includes/async-await-2.0.md
```diff
@@ -1,6 +1,6 @@
 <aside class="alert alert-info" markdown="1">
-  **Dart 2 note:** In Dart 2, instead of immediately suspending, async functions
-  execute synchronously until the first await. In most cases, you won't notice a
-  difference in behavior relative to Dart 1&mdash;as long as you await each time you
-  call an async method, it won't be an issue.
+  **Version note:** In Dart 1.x, async functions immediately suspended
+  execution. In Dart 2, instead of immediately suspending, async functions
+  execute synchronously until the first `await`. If you use `await` each time
```
MichaelRFairhurst: oh man, I only just saw this.

Looks like my revisions here were maybe mostly unwarranted :(

I had this wrong:

```
Future<void> f() async {
  print('before await');
  await null;
  print('after await');
}

Future<void> wihoutAwait() async {
  print('before f()');
  f();
  print('after f()');
}

Future<void> withAwait() async {
  print('before await f()');
  await f();
  print('after await f()');
}
```

I thought Dart 2 changed the order of both `withAwait()` and `withoutAwait()`! But its only `withoutAwait()` that changed........and I changed code describing code that's more like `withAwait()`!

src/_tutorials/language/futures.md
```diff
@@ -175,45 +174,59 @@ program prints.  If `gatherNewsReports()` takes a little while to complete its
...
 <img src="../images/async-await.png"
-     alt="diagram showing flow of control through the main() and printDailyNewsDigest functions" />
+     alt="diagram showing flow of control through the main() and printDailyNewsDigest functions">
 
 1. The app begins executing.
-1. The main function calls `printDailyNewsDigest()`, which (because it's
-   marked `async`), immediately returns a Future, _before any code is
-   executed_.
+1. The `main()` function calls the async function `printDailyNewsDigest()`,
+   which begins executing synchronously.
```
MichaelRFairhurst: so technically yes, this part is true, because printDailyNewsDigest() is not awaited. But if that's important to cover, this now needs to be made clear!

src/_tutorials/language/futures.md
```diff
@@ -175,45 +174,59 @@ program prints.  If `gatherNewsReports()` takes a little while to complete its
...
+1. `printDailyNewsDigest()` uses `await` to call the async function
+   `gatherNewsReports()`, which begins executing synchronously.
```
MichaelRFairhurst: here I was wrong! Since there's an await, it pauses and only starts gatherNewsReports() after main() finishes!

src/_tutorials/language/futures.md
```diff
@@ -175,45 +174,59 @@ program prints.  If `gatherNewsReports()` takes a little while to complete its
...
+   execution. First, the Future returned by `gatherNewsReports()` completes.
+   Then `printDailyNewsDigest()` continues executing, printing the news.
+1. When the `printDailyNewsDigest()` function body finishes executing,
    the Future that it originally returned completes, and the app exits.
 
-<aside class="alert alert-info" markdown="1">
-  **Note:** If an async function doesn't explicitly return a value, it returns
-  a Future wrapped around a null value.
-</aside>
+Note that an async function starts executing right away (synchronously).
```
MichaelRFairhurst: unless called with await :(

src/_tutorials/language/futures.md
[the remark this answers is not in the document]
MichaelRFairhurst: When not awaited, yes, and I tested to confirm:

```
uture f1() async {                                                              
  print('5');                                                                    
  f2();                                                                          
  await null;                                                                    
  print('6');                                                                    
}                                                                                
                                                                                 
Future f2() async {                                                              
  print('7');                                                                    
  f3();                                                                          
  await null;                                                                    
  print('8');                                                                    
}                                                                                
                                                                                 
void f3() {                                                                      
  print('9');                                                                    
}
```

calling `f()` without `await`, it prints 5, 7, 9, before 6, 8.

src/_tutorials/language/futures.md
```diff
@@ -175,45 +174,59 @@ program prints.  If `gatherNewsReports()` takes a little while to complete its
...
+  gets the uncompleted Future from that expression).
+* The function's `return` statement.
+* The end of the function body.
+
 
 ### Handling errors {#handling-errors-async}
 
 If a Future-returning function completes with an error, you probably want to
-capture that error.  Async functions can use try-catch to capture the error.
+capture that error. Async functions can use try-catch to capture the error.
```
MichaelRFairhurst: I know you only added a space here, but could this use reduced redundancy? "use try-catch to do so" ?

src/_tutorials/language/futures.md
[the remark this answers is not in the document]
MichaelRFairhurst: Hmm, I thought the way you call them will stay the same?

It will maintain similarity behind how they look, but not in how they run. If the goal is to show what async/await translates to in terms of the Future API, its best to include the fact that it returns IMO. It might be less to throw at the user to elide it, but I'm not sure it will help to show them less in this case?

src/_tutorials/language/futures.md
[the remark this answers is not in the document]
MichaelRFairhurst: yeah, adding void is probably best.

I see a bit of an identity crisis in the examples here though, maybe. Do we want our code examples to have omitted types everywhere, look super clean and rely on dynamic programming? Or do we want to have precise type definitions so users can learn how to type their code properly during these examples?

If we're following effective-dart, I believe that would recommend putting void in here...

src/_tutorials/language/futures.md
```diff
@@ -367,17 +382,31 @@ written in a couple different ways.
 
   <?code-excerpt "misc/lib/tutorial/daily_news.dart (main-future-api-pass-print)"?>
   {% prettify dart %}
-  printDailyNewsDigest() =>
+  Future<void> printDailyNewsDigest() =>
       gatherNewsReports().then(print);
   {% endprettify %}
 
+You need to provide an argument to `then()`'s callback,
```
MichaelRFairhurst: I like that you're covering this because its counter-intuitive. (or, wait, did I write this? anyways! :))

I wonder if there's a way to present this as more of a simple mechanism, rather than as cautionary advice.

"the method `then()` on `Future` takes a callback with a single parameter. That parameter's type will then be inferred, so its a `String` on a `Future<String>`. The same mechanism occurs for `Future<void>`, so you get a parameter of type `void`, which means the compiler won't let you accidentally use it."

OK not nearly as concise, and, probably more confusing.

I wonder if we can eventually support things like:

```
  StreamController<void>().add(); // simplification of .add(null); since the parameter is void
  Future<void>().then(() { ... }); // don't require a parameter if its void
```

That would be slick.

src/_tutorials/language/futures.md
MichaelRFairhurst: Is there a way to maybe also state that using _ here would be an error?

src/_tutorials/language/futures.md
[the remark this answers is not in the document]
kwalrath: Maybe "returns a non-void value"?

src/_tutorials/language/futures.md
kwalrath: Even if it's awaited, it begins executing synchronously! (See https://dartpad.dartlang.org/b62c036c308fd018ff37da5d77670140.)

src/_tutorials/language/futures.md
kwalrath: That's not my understanding. It *does* begin executing synchronously.

src/_tutorials/language/futures.md
kwalrath: I don't think so. (See https://dartpad.dartlang.org/b62c036c308fd018ff37da5d77670140.)

src/_tutorials/language/futures.md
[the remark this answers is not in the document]
kwalrath: I think we want to show Future code that's equivalent to the async-await version of the code, in terms of what gets executed when.

src/_tutorials/language/futures.md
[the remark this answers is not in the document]
kwalrath: I think that, generally, we should follow Effective Dart. We should have a good reason for each exception to that rule (e.g. non-Dart programmers are likely to see this page, and we don't want to scare them off by introducing too many concepts at one time.)

src/_tutorials/language/futures.md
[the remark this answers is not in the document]
MichaelRFairhurst: That works for me!

src/_tutorials/language/futures.md
MichaelRFairhurst: Oh man, I am confused

src/_includes/async-await-2.0.md
MichaelRFairhurst: Hmm, looks like maybe this section is wrong, or poorly worded?

"if your methods all begin with an await" its the same dart 2/ dart 1?

src/_tutorials/language/futures.md
MichaelRFairhurst: Sorry, I misinterpreted the results of my test >_<

src/_tutorials/language/futures.md
MichaelRFairhurst: Redacted, sorry!

src/_tutorials/language/futures.md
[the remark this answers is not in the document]
kwalrath: I changed this to "Any `return` statement in the function."

src/_tutorials/language/futures.md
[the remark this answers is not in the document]
kwalrath: Actually, I decided to use "returns a value (which can be `void`)." That simplifies some language that appears a few lines later: "Later... completes with a value or with an error."

kwalrath:
I've addressed some (but not all) of your comments and updated the staging server: https://kw-staging-dartlang-2.firebaseapp.com/tutorials/language/futures.

Before I address everything, I need to make sure I'm understanding the async-await semantics correctly. I thought that what you originally said was correct, Mike, so I was confused when you retracted your statements. :) I tested my understanding [in this dartpad](https://dartpad.dartlang.org/b62c036c308fd018ff37da5d77670140) and was able to predict the behavior. But maybe it isn't a complete enough test?

src/_includes/async-await-2.0.md
kwalrath: I think that if all async methods begin with await, then the dart 1 & dart 2 behavior is the same. But since you might not control the implementation of the async methods you call, I think this note is concentrating on what you can do as a caller: don't count on the function returning immediately.

kwalrath:
I think we might need an async-await expert to review this. @leafpetersen can you recommend someone?

leafpetersen:
@munificent or @lrhn might be good people to have a look?

src/_includes/async-await-2.0.md
MichaelRFairhurst: Ah, so this is suggesting:

```
await null; // suspend
foo(); // async function
```

This could be clearer, and also seems unreasonable advise to follow. But its not wrong, so that's good.

chalin:
> Travis is happy

FYI, rebasing should make Travis happier. After rebasing you might need to run `bundle update`.

src/_tutorials/language/futures.md
leafpetersen: Yes, this was the change for Dart 2: async functions begin synchronously.

kwalrath:
Great, thanks Leaf!

PR #1082, opened 2018-08-16
Add `environment` field to sdk constraint section

jakemac53:
In the sdk constraint section of the Dart 2 page it previously showed only the `sdk` field, which must be nested under the `environment` field to work. It also didn't have the proper indentation so it looked like a top level key.

This adds the `environment` key with proper indentation for the `sdk` field, which should help clarify how to set up sdk constraints.

This caused some confusion in the gitter channel today for a user trying to get their sdk constraints updated to allow Dart 2.

### DIFF
--- a/src/dart-2.md
+++ b/src/dart-2.md
@@ -119,11 +119,12 @@ Once your package passes Dart 2 analysis, update the upper constraint
 to declare that the package is compatible with Dart 2:
 
 ```yaml
-# Works in Dart 1 (starting with 1.20.1), and works in Dart 2
-sdk: '>=1.20.1 <3.0.0'
+environment:
+  # Works in Dart 1 (starting with 1.20.1), and works in Dart 2
+  sdk: '>=1.20.1 <3.0.0'
 
-# Works in Dart 2 only, starting with Dart 2 dev build 61
-sdk: '>=2.0.0-dev.61.0 <3.0.0'
+  # Works in Dart 2 only, starting with Dart 2 dev build 61
+  sdk: '>=2.0.0-dev.61.0 <3.0.0'
 ```
 
 <aside class="alert alert-warning" markdown="1">


chalin:
IMHO, if we add context for the `sdk` entry, then we need to split the example, something like this:

```yaml
environment:
  # Works in Dart 1 (starting with 1.20.1), and works in Dart 2
  sdk: '>=1.20.1 <3.0.0'
```

```yaml
environment:
  # Works in Dart 2 only, starting with Dart 2 dev build 61
  sdk: '>=2.0.0-dev.61.0 <3.0.0'
```

jakemac53:
Sure, I like that even more @chalin I was just trying to make the most minimal change to be the least controversial ;)

chalin:
> was just trying to make the most minimal change 

Yep, that is a good strategy! :)

> Happy to change it if we all agree that is the way it should look.

@kwalrath?

kwalrath:
We should have a bit of text between the examples. Also, don't we expect supporting Dart 2 stable to be the common case now? How about something like this:

-----------

Once your package passes Dart 2 analysis, update the SDK constraint to declare that the package is compatible with Dart 2:

```yaml
environment:
  # Works in Dart 2 only.
  sdk: '>=2.0.0 <3.0.0'
```

If you want to maintain compatibility with older versions of Dart, adjust the lower SDK constraint accordingly:

```yaml
environment:
  # Works in Dart 1 (starting with 1.20.1), and works in Dart 2.
  sdk: '>=1.20.1 <3.0.0'
```

-----------

kwalrath:
(thanks for the PR, @jakemac53!)

jakemac53:
Updated, I also noticed the next paragraph is out of date:

```
Packages that declare an upper constraint of `<2.0.0` can still be consumed by
  Dart 2 dev builds, as those have lax upper constraint checking. However,
  once Dart 2 ships on the stable channel, those constraints will no longer
  resolve, and **only packages that declare an upper constraint of `<3.0.0` will
  work with Dart 2**.
```

I could include new wording in this pr if you all want or we could do that separately :).

PR #1084, opened 2018-08-16
Update paragraph about constraints

kwalrath:
Followup to #1082.

Staged: https://sz-www-2.firebaseapp.com/dart-2#upper-constraints-on-the-sdk-version

### DIFF
--- a/src/dart-2.md
+++ b/src/dart-2.md
@@ -118,9 +118,6 @@ If a backward-compatible change isn't possible,
 Once your package passes Dart 2 analysis, update the upper constraint
 to declare that the package is compatible with Dart 2:
 
-Once your package passes Dart 2 analysis, update the SDK constraint to declare
-that the package is compatible with Dart 2:
-
 ```yaml
 environment:
   # Works in Dart 2 only.
--- a/src/dart-2.md
+++ b/src/dart-2.md
@@ -137,11 +137,11 @@ environment:
 ```
 
 <aside class="alert alert-warning" markdown="1">
-  Packages that declare an upper constraint of `<2.0.0` can still be consumed by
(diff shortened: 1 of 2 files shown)


src/dart-2.md
```diff
@@ -137,11 +137,11 @@ environment:
...
 
 <aside class="alert alert-warning" markdown="1">
-  Packages that declare an upper constraint of `<2.0.0` can still be consumed by
-  Dart 2 dev builds, as those have lax upper constraint checking. However,
-  once Dart 2 ships on the stable channel, those constraints will no longer
-  resolve, and **only packages that declare an upper constraint of `<3.0.0` will
-  work with Dart 2**.
+  **Packages must have an upper constraint of `<3.0.0`** to work in
+  Dart 2 stable and subsequent Dart 2 releases.
+  Dart 2 dev builds _before_ the stable release have
```
chalin: Is it worth mentioning how things were in Dart 2-dev?

src/dart-2.md
```diff
@@ -137,11 +137,11 @@ environment:
 ```
```
chalin: There is a repeated paragraph in lines 118 and 121 above:

> Once your package passes Dart 2 analysis, update the upper constraint to declare that the package is compatible with Dart 2:
>
> Once your package passes Dart 2 analysis, update the SDK constraint to declare that the package is compatible with Dart 2:

That needs to be fixed.

kwalrath: Oops!

src/dart-2.md
kwalrath: You mean, I might consider deleting 142-144? I thought developers might want to know when things changed, especially since Flutter is still on a pre-stable dev build. @kevmoo, what do you think?

PR #1094, opened 2018-08-20
Changed color for neutral  part 

mayank2424:
Hello, I tried to change color for neutral part under Wheretype() heading to original neutral color from light orange which was looking similar to bad part ,Someone tell me if I had approach this thing is a right way ??


mayank2424:
I signed it!

kwalrath:
@mayank2424 thanks very much for trying to fix this! I'm closing, as explained in #1095, but we really appreciate your willingness to contribute!

Issue #527, opened 2018-01-22
Upgrade to bootstrap 4


chalin:
[Bootstrap 4 (stable)](https://getbootstrap.com) is out! See the [official migration guide](https://getbootstrap.com/docs/4.0/migration).

---

### Config

- [ ] Tooltips and popovers: now provided by third-party `popper.js` which must be separately loaded.

### [Grid system](https://getbootstrap.com/docs/4.0/migration/#grid-system)

- [ ] Drop `$screen-*` variable overrides (they don't exist anymore)
- [ ] Replace:
  - `/@media\s\(min-width: \$screen-(\w*)-min\)/` with `@include media-breakpoint-up($1)`
  - `/@media\s*\(max-width: \$screen-(\w*)-min\)/` with `@include media-breakpoint-down($1)`
- [ ] A new tier below `sm` was added. Every tier has been bumped up one level:
  `xs` -> can go to nothing or `sm`, `sm` -> `md` -> `lg` -> `xl`.
  E.g., `.col-xs-6` -> `.col-6`, `col-md-6` -> `.col-lg-6`. This affects size-related class names in markdown, and media queries in CSS.

  The new `$grid-breakpoints` are:
   - `xs: 0`
   - `sm: 576px`
   - `md: 768px` (formerly `sm`)
   - `lg: 992px` (formerly `md`)
   - `xl: 1200px` (formerly `lg`)

### [Buttons](https://getbootstrap.com/docs/4.0/migration/#buttons)

- [ ] Renamed `.btn-default` to `.btn-secondary`.

### [Dropdowns](https://getbootstrap.com/docs/4.0/migration/#dropdowns)

- [ ] Use appropriate classes, such as the required `.dropdown-item`.
- [ ] Drop unnecessary custom caret.

### [Navs](https://getbootstrap.com/docs/4.0/migration/#navs)

- [ ] Use appropriate classes, such as the required `.nav`, `.nav-item`, etc.

### [Navbars](https://getbootstrap.com/docs/4.0/migration/#navbar)

- [ ] ...

### Dynamic elements

- [ ] Tooltips and popovers: slightly different config.

### Utils

- [ ] `.pull-right` -> `.float-right`

bcko:
I am interested. May I get your guidance?

chalin:
This is a major undertaking that needs to be coordinated across site-www and site-webdev. It also requires a good understanding of how styles are currently used across both sites.

[...]

PR #1106, opened 2018-08-27
Upgrade to bootstrap 4, Step 1: essentials

chalin:
@kwalrath - here is step 1.

Contributes to #527 

Things that are known to be off (but will be fixed in step 2) include:

- Drop-down menu alignment in the main nav is off, but there's no quick fix for that.
- In narrow displays, the sidenav pop-out animation doesn't seem to always work.

In both case, there is no quick fix, so these will be addressed in step 2 of the migration.

Staged at https://sz-www-1.firebaseapp.com

cc @Sfshaza @filiph


kwalrath:
LGTM once Travis is happy and we (temporarily) remove 3rd level headings from the auto-generated TOCs.

chalin:
> LGTM once Travis is happy

Travis is green.

> and we (temporarily) remove 3rd level headings from the auto-generated TOCs.

Done.

PR #1110, opened 2018-08-29
Update the dart:io article

kwalrath:
Updated the dart:io article to auto-include code and to reflect Dart 2 practices. Many of the content and code changes were suggested by @mkustermann, who reviewed the article and provided new code.

Staged:
https://kw-www-dartlang-1.firebaseapp.com/articles/libraries/dart-io
https://kw-www-dartlang-1.firebaseapp.com/articles/libraries
https://kw-www-dartlang-1.firebaseapp.com/articles#libraries-and-apis

/cc @chalin, to make sure I included/tested the code correctly. (It's not really tested, but it _is_ analyzed.)


kwalrath:
Hmmm... the generated page has disappeared, so the staging URL isn't working right now. (This is something that's happened intermittently in the last few days; I don't know why yet.)

Unrelated: I should probably move this article out of the server-side article subgroup (https://www.dartlang.org/articles/dart-vm) and into the main group of articles (https://www.dartlang.org/articles). So I'll make that change while I'm trying to figure out what's wrong with the page generation.

kwalrath:
OK, I've moved the page and it's now staged:

https://kw-www-dartlang-1.firebaseapp.com/articles/libraries/dart-io
https://kw-www-dartlang-1.firebaseapp.com/articles/libraries
https://kw-www-dartlang-1.firebaseapp.com/articles#libraries-and-apis

I still need to do redirects and URL adjustments.

src/_articles/libraries/dart-io.md
```diff
@@ -0,0 +1,386 @@
...
+
+Running this example at the command line, we get:
+
+```terminal
+$ dart timer.dart
+end of main
+timer
+```
+
+Had we made the timer repeating by using the Timer.periodic() constructor,
```
mkustermann: Maybe quote the `Timer.periodic()` and/or link to the documentation?

src/_articles/libraries/dart-io.md
```diff
@@ -0,0 +1,386 @@
...
+  print('exit code: ${await process.exitCode}');
+}
+{% endprettify %}
+
+Notice that `exitCode` can complete before all of the lines of output
+have been processed. Also note
+that we do not explicitly close the process. In order to
+not leak resources we have to drain both the stderr and the stdout
+streams. To do that we set a listener (using `await for`)
+to drain the stderr stream.
```
mkustermann: For `stderr` this code example does not use `await for`. Maybe rephrase this to something like this:

.. we have to listen to both streams, stderr and stdout. We use `await for` for listening to the stdout of the subprocess whereas we use `stderr.drain()`, which will listen to a stream and discard it.

src/_articles/libraries/dart-io.md
```diff
@@ -0,0 +1,386 @@
...
+import 'dart:io';
+
+Future<void> main() async {
+  var output = File('output.txt').openWrite();
+  Process process = await Process.start('ls', ['-l']);
+
+  // Wait for the process to finish; get the exit code.
+  final exitCode = (await Future.wait([
+    process.stdout.pipe(output), // Send stdout to file
+    process.stderr.drain(),
```
mkustermann: You could add a comment here `// drains stderr and discards it`

src/_articles/libraries/dart-io.md
```diff
@@ -0,0 +1,386 @@
...
+Writing HTTP clients is very similar to using the
+[HttpClient]({{site.dart_api}}/{{site.data.pkg-vers.SDK.channel}}/dart-io/HttpClient-class.html)
+class.
+
+
+## Feature requests welcome
+
+The dart:io library is already capable of performing a lot of tasks.
+As an example of our own use of dart:io,
+we have rewritten the Dart testing scripts from Python to Dart.
```
mkustermann: I think that our testing scripts were originally written in Python might not be very interesting. Maybe we could change this to:

The dart:io library is capable of performing a lot of tasks. As an example the `pub.dartlang.org` site is written using it.

src/_articles/libraries/dart-io.md
```diff
@@ -0,0 +1,386 @@
...
+import 'dart:convert';
+import 'dart:io';
+
+Future<void> main() async {
+  var file = File(Platform.script.toFilePath());
+  print("${await (file.readAsString(encoding: ascii))}");
+}
+{% endprettify %}
+
+Notice that the readAsString() method is asynchronous;
```
mkustermann: Maybe quote `readAsString()` ?

src/_articles/libraries/dart-io.md
```diff
@@ -0,0 +1,386 @@
...
+Dart is a single-threaded programming language.
+If an operation blocks the Dart thread,
+the application makes no progress before that operation completes.
+For scalability it is therefore crucial that no I/O operations block.
+Instead of blocking on I/O operations,
+dart:io uses an asynchronous programming model inspired by
+[node.js,](http://nodejs.org)
+[EventMachine,](https://github.com/eventmachine/eventmachine/wiki) and
+[Twisted.](http://twistedmatrix.com/trac/)
+
```
mkustermann: Thinking about this again, maybe we should de-emphasize it a little bit that most of these can be used in Flutter as well, because I think they have more layers on top of the normal `dart:io` and some things should probably not be used (subprocesses).

We could add a section here saying:

## Usage in flutter

Flutter has specific APIs to access assets of an application, which are more high-level than dart:io. Please refer to https://flutter.io/assets-and-images/ for more information.

examples/misc/lib/articles/io/io_http_server_file_test.dart
```diff
@@ -0,0 +1,44 @@
+// #docregion
+import 'dart:async' show Future;
+import 'dart:io';
+
+Future<void> runServer(String basePath) async {
+  HttpServer server = await HttpServer.bind('127.0.0.1', 8082);
```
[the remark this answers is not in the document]
kwalrath: Yep: https://www.dartlang.org/guides/language/effective-dart/design#avoid-type-annotating-initialized-local-variables

So it looks like I should use final or var here. I think we decided to err on the side of using `var`, since the performance implications of `final` aren't clear yet (varies by platform).

src/_articles/libraries/dart-io.md
kwalrath: We used to have a policy on dartlang.org of avoiding code font when possible, but that contradicts the Google developer docs style guide (https://developers.google.com/style/code-in-text) and what many developers expect. So we're updating as we go. I'll add code font to this article.

kwalrath:
@Sfshaza Is this a good note for the Flutter caveat:

image

Is there anything else I should say? You can see the whole article at https://kw-www-dartlang-1.firebaseapp.com/articles/libraries/dart-io.

Sfshaza:
I like the Flutter note! I can't think of anything else to add.

Sfshaza:
Unless you want to point out the URL for the Flutter API reference.//

PR #1112, opened 2018-08-29
Add alt text to images that don't already have it

### LINKED ISSUE #1104: Add `alt` attributes to all images
Related: #1103

kwalrath:
Fixes #1104.

Staged at URLs under https://kw-staging-dartlang-2.firebaseapp.com.

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -758,7 +758,7 @@ units. Use the `runes` property to get the runes of a string.
 
 The following example illustrates the relationship between runes,
 16-bit code units, and 32-bit code points.
-Click the run button ( {% asset red-run.png %} )
+Click the run button ( {% asset red-run.png alt="image of run button" %} )
 to see runes in action.
 
 {% comment %}
@@ -1026,7 +1026,7 @@ that specifies a default list for the `list`
 parameter and a default map for the `gifts` parameter.
 {% comment %}
 The function is called three times with different values.
-Click the run button ( {% asset red-run.png %} )
+Click the run button ( {% asset red-run.png alt="image of run button" %} )
 to see list and map default values in action.
 {% endcomment %}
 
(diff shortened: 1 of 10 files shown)


src/_guides/language/language-tour.md
```diff
@@ -758,7 +758,7 @@ units. Use the `runes` property to get the runes of a string.
 
 The following example illustrates the relationship between runes,
 16-bit code units, and 32-bit code points.
-Click the run button ( {% asset red-run.png %} )
+Click the run button {% asset red-run.png alt="" %}
```
chalin: Maybe `alt="(>)"`, here and elsewhere?

chalin: (Btw, my plan was to eventually replace these images with the appropriate Material Icon.)

kwalrath: The image doesn't add anything for non-sighted users, so per the accessibility guidelines I've found (e.g. https://webaim.org/techniques/alttext/), I set it to an empty string.

kwalrath: (These are screenshots, so I'm not sure the Material icon is appropriate here.)

chalin: We don't actually want a screenshot here, we want to show a button (which we can color red if that is deemed necessary):

    Click the run button <i class="material-icons">play_circle_filled</i>.

As you can see from this DartPad screenshot:

> screen shot 2018-08-31 at 05 57 41

The DartPad run button is essentially a standard Material play button, so it seems appropriate IMHO:

> screen shot 2018-08-31 at 06 00 04

Let me know if you agree. If so, I'll create a followup issue.

kwalrath: Sure, that sounds good to me. I didn't realize the DartPad button conformed to Material Design. :)

chalin: Done: #1118

PR #1119, opened 2018-09-01
Make clear that const constructors don't always create constant objects

kwalrath:
Staged:
https://kw-staging-dartlang-2.firebaseapp.com/guides/language/language-tour#using-constructors
https://kw-staging-dartlang-2.firebaseapp.com/guides/language/language-tour#constant-constructors

I'm not sure if the second one (in the constructor implementation section) is really necessary. Let me know what you think.


kwalrath:
Improved, I hope. I've updated the staged version.

src/_guides/language/language-tour.md
```diff
@@ -2345,9 +2345,21 @@ const pointAndLine = {
 };
 {% endprettify %}
 
+If a constant constructor is outside of a constant context
```
munificent: Do you define "constant context" anywhere? If not, it would be good to show an example where you don't need `const` but it *does* create a const object, like:

```dart
const foo = Foo(); // "const Foo()" not needed here.
```

kwalrath: We don't explicitly define it, but we show it. E.g. in the example before, line 2341 says "only one const, which establishes the constant context".

PR #1125, opened 2018-09-04
Improve links in the library tour

### LINKED ISSUE #1120: Exception section links to itself
https://www.dartlang.org/guides/libraries/library-tour#exceptions  

The sentence, "For more information, see Exceptions and the Exception API docs".  

Exceptions is a link to https://www.dartlang.org/guides/libraries/library-tour#exceptions  

Not sure if there really is a more informative section on Exceptions. If so, the link should be updated. If not, the link should be removed.

kwalrath:
Exceptions section used to link to itself.
Also changed "docs" to "reference", per https://developers.google.com/style/word-list#documentation.

Fixes #1120.

### DIFF
--- a/src/_guides/libraries/library-tour.md
+++ b/src/_guides/libraries/library-tour.md
@@ -243,7 +243,7 @@ assert('  '.isNotEmpty);
 #### Replacing part of a string
 
 Strings are immutable objects, which means you can create them but you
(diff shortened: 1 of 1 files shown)


Sfshaza:
Muuuch better. LGTM.

PR #1129, opened 2018-09-05
chore: use `site.g` as generated site folder

chalin:
(This will be the folder name used across all three sites. cc @kwalrath @Sfshaza @filiph)


chalin:
There was an issue with link checking, or maybe an issue with superstatic. I'll merge anyways so that @kwalrath can pickup the new site-shared submodule.

PR #1136, opened 2018-09-06
Address "newby" feedback on the language tour

kwalrath:
Staged:
https://kw-www-dartlang-1.firebaseapp.com/guides/language/language-tour#important-concepts
https://kw-www-dartlang-1.firebaseapp.com/guides/language/language-tour#keywords
https://kw-www-dartlang-1.firebaseapp.com/guides/language/language-tour#final-and-const
https://kw-www-dartlang-1.firebaseapp.com/guides/language/language-tour#strings

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -125,10 +125,15 @@ mind:
 -   *Identifiers* can start with a letter or underscore (\_), followed by any
     combination of those characters plus digits.
 
--   Sometimes it matters whether something is an *expression* or a
-    *statement*, so it helps to be precise about those two words.
+-   Dart has both *expressions* (which have runtime values) and
+    *statements* (which don't).
+    For example, the [conditional expression](#conditional-expressions)
+    `condition ? expr1 : expr2` has a value of `expr1` or `expr2`.
+    Compare that to an [if-else statement](#if-and-else), which has no value.
+    A statement often contains one or more expressions,
+    but an expression can't contain a statement.
 
--   Dart tools can report two kinds of problems: warnings and errors.
+-   Dart tools can report two kinds of problems: _warnings_ and _errors_.
     Warnings are just indications that your code might not work, but
     they don’t prevent your program from executing. Errors can be either
     compile-time or run-time. A compile-time error prevents the code
@@ -139,6 +144,7 @@ mind:
 ## Keywords
 
 The following table lists the words that the Dart language treats specially.
+**Avoid using these words as identifiers.**
 
 {% assign bii = '&nbsp;<sup title="built-in-identifier" alt="built-in-identifier">1</sup>' %}
 {% assign lrw = '&nbsp;<sup title="limited reserved word" alt="limited reserved word">2</sup>' %}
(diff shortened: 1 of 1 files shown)


src/_guides/language/language-tour.md
```diff
@@ -125,10 +125,15 @@ mind:
...
-    *statement*, so it helps to be precise about those two words.
-
--   Dart tools can report two kinds of problems: warnings and errors.
+-   Dart has both *expressions* (which have runtime values) and
+    *statements* (which don't).
+    For example, the [conditional expression](#conditional-expressions)
+    `condition ? expr1 : expr2` has a value of `expr1` or `expr2`.
+    Compare that to an [if-else statement](#if-and-else), which has no value.
+    A statement often contains one or more expressions,
+    but an expression can't contain a statement.
```
lrhn: ... can't directly contain ...

(otherwise pedants will say that `() { print(42); }` is an expression containing a statement).

src/_guides/language/language-tour.md
```diff
@@ -160,22 +166,23 @@ The following table lists the words that the Dart language treats specially.
...
-an identifier in any function body marked with `async`, `async*`, or `sync*`.
-For more information, see
-[Asynchrony support](#asynchrony-support).
+When necessary, the words marked with superscripts can sometimes be used as
+identifiers:
+
+* Words with the superscript **1** are **built-in identifiers**.
+  To simplify the task of porting JavaScript code to Dart,
+  built-in identifiers are valid in most places,
+  but they can't be used as class or type names.
```
lrhn: Maybe say, "they can't, e.g., be used as ..." because there is one more restriction (can't be used as an import prefix), or add that to the list.

src/_guides/language/language-tour.md
```diff
@@ -160,22 +166,23 @@ The following table lists the words that the Dart language treats specially.
...
 
-All other words in the keyword table are **reserved words**.
-You can't use reserved words as identifiers.
+* Words with the superscript **2** are newer, limited reserved words related to
+  the **asynchrony** support that was added after Dart's 1.0 release.
+  You can't use `async`, `await`, or `yield` as an identifier
```
lrhn: `async` is no longer in this group, it's only `await` and `yield`.

src/_guides/language/language-tour.md
```diff
@@ -160,22 +166,23 @@ The following table lists the words that the Dart language treats specially.
...
-
-<sup>2</sup> Words with the superscript **2**
-are newer, limited reserved words related to **asynchrony** support
-added after Dart's 1.0 release.
-You can't use `async`, `await`, or `yield` as
-an identifier in any function body marked with `async`, `async*`, or `sync*`.
-For more information, see
-[Asynchrony support](#asynchrony-support).
+When necessary, the words marked with superscripts can sometimes be used as
+identifiers:
```
lrhn: Not sure that `sync*`, `async*` and `yield*` fits in the table above since they are not identifiers.

Even the plain `sync` or `async` are not reserved in any way.
They are no more reserved than `on`, `show` or `hide`, which are plain contextual keywords (means something only when they occur in a specific place, but are completely valid identifiers in every way otherwise).

lrhn: Cant see `Function` in the table above. It's also a built-in identifier in Dart 2.

src/_guides/language/language-tour.md
```diff
@@ -252,6 +259,10 @@ the first time it's used.
 <div class="alert alert-info" markdown="1">
 **Note:**
 Instance variables can be `final` but not `const`.
+Final instance variables must be initialized before
+the constructor body starts —
+at the variable declaration, in a constructor argument,
```
lrhn: "in a constructor argument" reads weirdly to me (but it just might be me).
Maybe "by a constructor argument"? (Or, better, "by a constructor parameter" - we try to use "parameter" about the function's declared parameters, and "argument" about the value being passed at the call site).

src/_guides/language/language-tour.md
```diff
@@ -441,7 +452,8 @@ const msUntilRetry = secondsUntilRetry * msPerSecond;
 
 ### Strings
 
-A Dart string is a sequence of UTF-16 code units. You can use either
+A Dart string is a sequence of UTF-16 code units. (When Dart is compiled to
+JavaScript, the string representation remains the same.) You can use either
```
lrhn: Not sure what the addition gives us here. 
It's basically saying "A is true. Even when B, A is still true." If I didn't have the expectation that B would change anything, now I'm just confused :(

Is this because we have previously said that the same isn't true for `int`?
Otherwise, I'd just drop the sentence, or change it to something like:
> , the same as a JavaScript, Java or C# string.
(except that doesn't flow very well, so maybe something like that as written by a professional writer)

src/_guides/language/language-tour.md
kwalrath: They're not reserved, but don't you think it'd be confusing to see them used as identifiers?

kwalrath: Maybe I should have `async`, `on`, `show`, and `hide` in the table in a new category (contextual keywords). (Are there any others?)

This table is all about which words are are meaningful to Dart. I think any meaningful word might be confusing if used as an identifier, even if it's allowed, but I can soften the "avoid" language if you think that's important.

I'll remove `sync*`, `async*`, and `yield*`.

src/_guides/language/language-tour.md
kwalrath: I think I'll just drop it. In other places we talk about JavaScript comparisons, but we should probably move away from those.

kwalrath:
@lrhn I think my new version is better than the old one, so I'm going to commit these changes. Please take another look at the keyword table, if you can.

lrhn:
Looks great!

PR #1138, opened 2018-09-08
Remove banners

### LINKED ISSUE #1137: Remove Dart 2 banner
We think it's been long enough since Dart 2 launch that existing developers will have heard the news, and the banner is of no value to new developers, so can we please go ahead and remove the blue 'Dart 2 is now stable! Migrate your code or install Dart.' banner?

kwalrath:
Fixes #1137.

Staged at URLs like:
https://kw-www-dartlang-1.firebaseapp.com
https://kw-www-dartlang-1.firebaseapp.com/guides/language/effective-dart

### DIFF
--- a/src/_guides/language/effective-dart/design.md
+++ b/src/_guides/language/effective-dart/design.md
@@ -8,8 +8,6 @@ prevpage:
 <?code-excerpt replace="/([A-Z]\w*)\d\b/$1/g"?>
 <?code-excerpt plaster="none"?>
 
-{% include effective-dart-banner.html %}
-
 Here are some guidelines for writing consistent, usable APIs for libraries.
 
 * TOC
--- a/src/_layouts/default.html
+++ b/src/_layouts/default.html
@@ -8,7 +8,7 @@
       {% include navigation-toc.html %}
       <article>
         <div class="content">
-          {% include banner.html %}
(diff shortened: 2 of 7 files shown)


src/_guides/language/effective-dart/design.md
```diff
@@ -8,8 +8,6 @@ prevpage:
 <?code-excerpt replace="/([A-Z]\w*)\d\b/$1/g"?>
 <?code-excerpt plaster="none"?>
 
-{% include effective-dart-banner.html %}
```
chalin: `effective-dart-banner.html` should probably be deleted since no more uses remain.

src/_layouts/default.html
```diff
@@ -8,7 +8,7 @@
       {% include navigation-toc.html %}
       <article>
         <div class="content">
-          {% include banner.html %}
+          {% comment %} {% include banner.html %} {% endcomment %}
```
chalin: Adding / removing the banner is common enough. We could control its inclusion via a site config field (something like:

```yaml
show_banner: true # or false or simply excluded
```

Then in this file:

```
{% if site.show_banner %} {% include banner.html %} {% endif %}
```

(with a similar change in `src/index.hmlt`.

kwalrath:
I'm going to merge this, then apply Patrice's improvements.

PR #1139, opened 2018-09-08
Update references to /dart-2; clean up includes

kwalrath:
Don't commit until #1138 is committed.

Related to dart-lang/site-shared#43.

### DIFF
--- a/src/_includes/banner.html
+++ b/src/_includes/banner.html
@@ -1,8 +1,6 @@
 {% comment %}Drop the div .alert classes to revert to the default banner styling.{% endcomment -%}
 <div class="banner alert alert-info">
   <p class="banner__text">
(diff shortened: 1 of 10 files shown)


src/_includes/banner.html
```diff
@@ -1,8 +1,6 @@
 {% comment %}Drop the div .alert classes to revert to the default banner styling.{% endcomment -%}
 <div class="banner alert alert-info">
   <p class="banner__text">
-    <a href="https://medium.com/dartlang/dart-2-stable-and-the-dart-web-platform-3775d5f8eac7" target="_blank" rel="noopener"><b>Dart 2</b> is now stable!</a>
-    <a href="/dart-2#migration">Migrate your code</a> or
-    <a href="/install">install Dart</a>.
+    TEXT GOES HERE
```
chalin: Maybe put this message in a comment, in case it ever is displayed by mistake :)

kwalrath: I did that and also applied your suggestions from #1138.

PR #1157, opened 2018-09-14
Apply Bob's feedback to futures tutorial

### LINKED ISSUE #1143: Incorporate Bob's review comments for Futures tutorial
As a followup to #1079, @munificent sent me the following comments by email.

> for example, by waiting for a time-consuming operation or blocking on I/O

"Waiting for" also sounds like it could mean an OS timer operation. Maybe "performing a long-running calculation"?

> Asynchronous operations let your program run without getting blocked.

Maybe, "let your program complete other work while waiting for an operation to complete"?

> since gatherNewsReports() blocks, the remaining code runs only when gatherNewsReports() returns with the contents of the file, however long that takes.

"when" -> "after" to make the synchrony clearer.

> If you use await each time you call an async function, the behavior is the same in Dart 1.x and Dart 2.

I don't think this sentence is precisely true and I worry that even informally it mostly serves to confuse. How about just omitting it?

```
const news = '<gathered news goes here>';
const oneSecond = Duration(seconds: 1);

final newsStream = new Stream.periodic(oneSecond, (_) => news);

// Imagine that this function is more complex and slow. :)
Future<String> gatherNewsReports() => newsStream.first;
```

Why use a Stream? How about:

```
Future<String> gatherNewsReports() =>
    Future.delayed(
        Duration(seconds: 1),
        () => '<gathered news goes here>');
```

> The following diagram shows the flow of execution through the code.

The 7 line might be a little confusing because it reads like reaching the end of the program is the thing that causes the future to complete. Maybe have a floating "one second later..." and have the line start from that?

> The functions get triggered in quick succession; when all of them complete with a value, Future.wait() returns a new Future. This Future completes with a list containing the values produced by each function.

The "functions get triggered" part is a little confusing because it reads like Future.wait() only works with functions that return futures, not simply futures. It's not Future.wait() that is triggering those functions. They are called when evaluating the list literal before passing the result to Future.wait().

How about:

When you pass Future.wait() a list of uncompleted futures, it immediately returns you a Future. That future doesn't complete until all of the given futures are done. When that happens, it completes with a list containing the values produced by each future in the original list.

kwalrath:
Fixes #1143.

Staged: https://kw-staging-dartlang-2.firebaseapp.com/tutorials/language/futures


kwalrath:
I should do a final review before publishing to make sure the code is consistent everywhere (in the figure and the DartPads), but I _think_ it's OK and ready for review by @munificent.

examples/misc/lib/tutorial/daily_news.dart
```diff
@@ -17,22 +17,22 @@ import 'dart:async';
...
+  var newsDigest = await _gatherNewsReportsAsync();
+  print(newsDigest);
 }
 // #enddocregion main-async
 
 // #docregion main-future-api
 Future<void> _printDailyNewsDigestAsyncUsingFutureAPI() {
   final future = _gatherNewsReportsAsync();
-  return future.then((news) => print(news));
+  return future.then((newsDigest) => print(newsDigest));
```
munificent: Effective Dart [would tell you](https://www.dartlang.org/guides/language/effective-dart/usage#dont-create-a-lambda-when-a-tear-off-will-do) to just do:

```dart
return future.then(print);
```

Doing that is a little confusing in a tutorial, though. Maybe there's a way to do a little extra work in there to justify the lambda so it doesn't break the style guide?

src/_tutorials/language/futures.md
```diff
@@ -13,19 +13,19 @@ nextpage:
 
   * Dart is single-threaded.
   * Synchronous code can make your program freeze.
-  * Use Futures to perform asynchronous operations.
-  * Use `await` in an async function to pause execution until a Future completes.
-  * Or use Future's `then()` method.
+  * Use `Future` objects (_futures_) to perform asynchronous operations.
+  * Use `await` in an async function to pause execution until a future completes.
```
munificent: How about "suspend" instead of "pause"?

src/_tutorials/language/futures.md
```diff
@@ -59,47 +59,49 @@ Baseball score: Red Sox 10, Yankees 0
...
 asynchronous model when defining functions that do potentially expensive work.
-Such functions return their value using a [Future.][Future]
+Such functions return their value using a [`Future`.][Future]
 
-## What is a Future? {#what-is-a-future}
+## What is a future? {#what-is-a-future}
 
-A Future represents an asynchronous operation that returns a value (which
-can be `void`). When a function that returns a Future is invoked,
+A `Future` object represents an asynchronous operation that returns a value (which
```
munificent: Maybe "returns" -> "produces"?

src/_tutorials/language/futures.md
```diff
@@ -59,47 +59,49 @@ Baseball score: Red Sox 10, Yankees 0
...
+can be `void`). When a function that returns a future is invoked,
```
munificent: [Ackshually](https://imgur.com/gallery/TKyAy), `void` isn't a value, it's a type. A `Future<void>` can theoretically complete to any value, you're just not allowed to use it.

How about:

> represents an asynchronous operation that may produce a value

src/_tutorials/language/futures.md
```diff
@@ -456,9 +470,11 @@ Nested callbacks also work, but they're harder to read and not as Dart-y.
 If the order of execution of the functions is not important, you can use
 `Future.wait()`.
 
-The functions get triggered in quick succession; when all of them complete
-with a value, `Future.wait()` returns a new Future.  This Future completes
-with a list containing the values produced by each function.
+When you pass `Future.wait()` a list of uncompleted futures,
```
munificent: Technically, you can give it completed futures too. I would just say "a list of futures".

src/_tutorials/language/futures.md
kwalrath: And remove the "(which can be void)"? This little sentence has been contentious!

An unmerged PR (#1079) [changed the text](https://github.com/dart-lang/site-www/pull/1072/files#diff-197c74d48dc0011b29d95d2e3a42eba4) from "A Future represents a means for getting a value sometime in the future" to "...tracking a calculation...".

I thought that was too limiting, so when I created PR #1079, I [initially changed it](https://github.com/dart-lang/site-www/pull/1079/commits/51c8774b337803294f42c6af616495db0044a13e#diff-197c74d48dc0011b29d95d2e3a42eba4) to "A Future represents an asynchronous operation that, optionally, returns a value." We ultimately settled on the wording "... returns a value (which can be `void`)".

So... how about this:

> A future is a `Future<T>` object, which represents an asynchronous operation that produces a result of type `T`. If the result isn't a usable value, then the future's type is `Future<void>`.

munificent: Works for me.

kwalrath:
I plan to merge this tomorrow morning.

PR #1160, opened 2018-09-19
chore(tool): use updated site-shared deploy script to manage robots.txt

chalin:


src/robots.txt
```diff
@@ -1 +1,15 @@
...
+---
+
+{% comment %}
+
+This is the default robots file used when serving locally or on staging servers.
+Instructions for deploying to the site's official server are in the README.md.
+
+{% endcomment -%}
+
+User-agent: *
```
chalin: @kwalrath: as agreed, the default robots.txt disallows all agents.

src/robots.txt
```diff
@@ -1 +1,15 @@
...
+Disallow: /
+User-agent: linkcheck
+Allow: /
```
chalin: Note that linkcheck hasn't yet been updated to recognize this entry (cc @filiph).

PR #1166, opened 2018-09-20
Update language-tour.md to add a note for overriding operators.

mzaman07:
This is to address issue #1081 by providing a note to the documentation indicating that `!=` is not an overriding operator but a syntactic sugar shortcut for evaluating an expression like `!(e1 == e2)`.

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -2990,9 +2990,9 @@ Vector class, you might define a `+` method to add two vectors.
 {:.table}
 
 <div class="alert alert-info" markdown="1">
-**Note** You may have noticed that `!=` is not an overridable operator. 
-This is because `!=` is not an operator in its own right, the expression `e1 != e2` 
-is just syntactic sugar for `!(e1 == e2)`.
+  **Note:** You may have noticed that `!=` is not an overridable operator.
+  This is because `!=` is not an operator in its own right; the expression `e1 != e2`
+  is just syntactic sugar for `!(e1 == e2)`.
 </div>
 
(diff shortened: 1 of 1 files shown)


src/_guides/language/language-tour.md
```diff
@@ -2989,6 +2989,12 @@ Vector class, you might define a `+` method to add two vectors.
 `–`  | `%`  | `>>`
 {:.table}
 
+<div class="alert alert-info" markdown="1">
+**Note** You may have noticed that `!=` is not an overridable operator.
```
kwalrath: * Note -> Note:
* , the -> ; the
* Indent (two spaces) everything between the div tags.

src/_guides/language/language-tour.md
```diff
@@ -2989,6 +2989,12 @@ Vector class, you might define a `+` method to add two vectors.
 `–`  | `%`  | `>>`
 {:.table}
 
+<div class="alert alert-info" markdown="1">
+  **Note:** You may have noticed that `!=` is not an overridable operator.
+  This is because `!=` is not an operator in its own right; the expression `e1 != e2`
+  is just syntactic sugar for `!(e1 == e2)`.
```
chalin: Great note! I'd suggest that we avoid stating that `!=` isn't "an operator in its own right". Maybe simplify this sentence to the following? @kwalrath, WDYT?

> This is because `!=` is just syntactic sugar for `!(e1 == e2)`.

kwalrath: I was thinking about something that, but I didn't want to have too much back and forth. I'll make this change later.

PR #1174, opened 2018-09-24
Update article: Emulating Functions

### LINKED ISSUE #600: Update or remove Emulating Functions article
The [Emulating Functions article](https://www.dartlang.org/articles/language/emulating-functions) hasn't been updated since 2014. Callable classes will be changing in Dart 2, as described [this announcement](https://groups.google.com/a/dartlang.org/d/msg/misc/hvt6jpiYxsE/ZQsrT67ZAQAJ). 

/cc @eernstg

kwalrath:
Staged at https://kw-staging-dartlang-2.firebaseapp.com/articles/language/emulating-functions

Fixes #600.


src/_articles/language/emulating-functions.md
```diff
@@ -45,14 +47,23 @@ suitably defined object.
...
-
-Otherwise, `noSuchMethod()` gets invoked. The default implementation of
-noSuchMethod() checks to see whether it was invoked due to an attempt to use call(),
-and if so issues a helpful error message suggesting you might have wanted to use
-a closure.
+When <code>x(a<sub>1</sub>, .., a<sub>n</sub>)</code> is evaluated,
+if it is a normal
+function, it gets called in the normal way. If it isn't we just invoke `call()`
+on it. If `x` supports a `call()` method with suitable arguments it gets called.
+Otherwise, `noSuchMethod()` gets invoked.
```
eernstg: Amazing, this is true!

We actually just invoke `call` when the static type of `x` is `dynamic`, or it is a class that has a method named `call`, but the other cases are compile-time errors. 

There are other situations where an instance that has a `call` method won't work so smoothly as a function, but I'm really happy to conclude that it still admits such a nice description when we talk about invocations.

So: No problems here whatsoever! ;-)

PR #1180, opened 2018-09-27
Update Creating Streams article

### LINKED ISSUE #1172: Update "Creating Streams" article
Re: https://www.dartlang.org/articles/libraries/creating-streams

The very least we could do (which I might do in the very short term) would be to put a note at the top, saying:

* The article doesn't cover async*-yield, which is the recommended way to work with streams today.
* The code isn't continuously tested, and there's a chance it doesn't run.
* The code doesn't follow the latest guidelines (e.g., it uses `new`).

Better (and my medium-term plan) would be to update the article. @lrhn started work on a replacement article that uses async*-yield, and he's updated it to be more Dart 2-y.

In the long term, we should have real, consolidated docs for learning about streams. #512

kwalrath:
Ready for review. Note that the contents of someFile.txt could be much more interesting, and maybe go under dartlang.org/f. I plan to make a follow-up PR that removes the code files under src and links to the new files under examples/ on GitHub.

Staged: https://kw-www-dartlang-1.firebaseapp.com/articles/libraries/creating-streams

Fixes #1172.


lrhn:
LGTM, thanks.

PR #1187, opened 2018-10-10
Remove sentence discouraging => with void

natebosch:
In practice we see a lot of usage of `=>` with a void return. The
practice is more closely aligned with using arrow functions for nearly
any single expression method, regardless of whether it returns a value.

We don't need to explicitly encourage using arrow functions this way,
but we also want to allow it for the authors who are already doing it.


chalin:
👍

PR #1202, opened 2018-10-22
Explicitly allow returing `void` from async

### LINKED ISSUE #1198: Clarify that it's OK to return `void` from an `async` function
https://www.dartlang.org/guides/language/effective-dart/design#do-use-futurevoid-as-the-return-type-of-asynchronous-members-that-do-not-produce-values

Seems to imply that `Future<void>` is preferred over `void` - they both have their use cases.

natebosch:
Closes #1198

### DIFF
--- a/src/_guides/language/effective-dart/design.md
+++ b/src/_guides/language/effective-dart/design.md
@@ -1564,8 +1564,9 @@ should use it. Doing so more directly matches how you'd type a similar
 synchronous function, and gives you better error-checking for callers and in the
 body of the function.
 
-For asynchronous functions that do not return a useful value and do not need to
-have their asynchronous work awaited, use a return type of `void`.
+For asynchronous functions that do not return a useful value and where no
+callers need to await the asynchronous work or handle an asynchronous failure,
+use a return type of `void`.
 
 
 ### AVOID using `FutureOr<T>` as a return type.
--- a/src/_guides/language/effective-dart/design.md
+++ b/src/_guides/language/effective-dart/design.md
@@ -1564,7 +1564,7 @@ should use it. Doing so more directly matches how you'd type a similar
 synchronous function, and gives you better error-checking for callers and in the
 body of the function.
 
-For functions which are "fire and forget", which will neither return a useful
+For functions that are "fire and forget", which neither return a useful
 value nor need to have their asynchronous work finished before the caller
 containues, use a return type of `void`.
 
--- a/src/_guides/language/effective-dart/design.md
+++ b/src/_guides/language/effective-dart/design.md
@@ -1555,7 +1555,7 @@ bool convertToBool(dynamic arg) {
 ### DO use `Future<void>` as the return type of asynchronous members that do not produce values.
 
 When you have a synchronous function that doesn't return a value, you use `void`
-as the return type. The asynchronous equivalent for a method which does not
+as the return type. The asynchronous equivalent for a method that doesn't
 produce a value, but should have the work awaited, is `Future<void>`.
 
 You may see code that uses `Future` or `Future<Null>` instead because older
--- a/src/_guides/language/effective-dart/design.md
+++ b/src/_guides/language/effective-dart/design.md
@@ -1556,7 +1556,7 @@ bool convertToBool(dynamic arg) {
 
 When you have a synchronous function that doesn't return a value, you use `void`
 as the return type. The asynchronous equivalent for a method that doesn't
-produce a value, but should have the work awaited, is `Future<void>`.
+produce a value, but that the caller might need to await, is `Future<void>`.
 
 You may see code that uses `Future` or `Future<Null>` instead because older
 versions of Dart didn't allow `void` as a type argument. Now that it does, you
--- a/src/_guides/language/effective-dart/design.md
+++ b/src/_guides/language/effective-dart/design.md
@@ -1555,13 +1555,18 @@ bool convertToBool(dynamic arg) {
 ### DO use `Future<void>` as the return type of asynchronous members that do not produce values.
 
 When you have a synchronous function that doesn't return a value, you use `void`
-as the return type. The asynchronous equivalent is `Future<void>`.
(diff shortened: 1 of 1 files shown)


src/_guides/language/effective-dart/design.md
```diff
@@ -1555,14 +1555,19 @@ bool convertToBool(dynamic arg) {
 ### DO use `Future<void>` as the return type of asynchronous members that do not produce values.
 
 When you have a synchronous function that doesn't return a value, you use `void`
-as the return type. The asynchronous equivalent is `Future<void>`.
+as the return type. The asynchronous equivalent for a method which does not
```
kwalrath:
```suggestion
as the return type. The asynchronous equivalent for a method that doesn't
```

src/_guides/language/effective-dart/design.md
```diff
@@ -1555,14 +1555,19 @@ bool convertToBool(dynamic arg) {
...
+produce a value, but should have the work awaited, is `Future<void>`.
 
 You may see code that uses `Future` or `Future<Null>` instead because older
 versions of Dart didn't allow `void` as a type argument. Now that it does, you
 should use it. Doing so more directly matches how you'd type a similar
 synchronous function, and gives you better error-checking for callers and in the
 body of the function.
 
+For functions which are "fire and forget", which will neither return a useful
```
kwalrath:
```suggestion
For functions that are "fire and forget", which neither return a useful
```

src/_guides/language/effective-dart/design.md
```diff
@@ -1555,14 +1555,19 @@ bool convertToBool(dynamic arg) {
...
+value nor need to have their asynchronous work finished before the caller
+containues, use a return type of `void`.
```
kwalrath:
```suggestion
continues, use a return type of `void`.
```

src/_guides/language/effective-dart/design.md
```diff
@@ -1555,14 +1555,19 @@ bool convertToBool(dynamic arg) {
 ### DO use `Future<void>` as the return type of asynchronous members that do not produce values.
 
 When you have a synchronous function that doesn't return a value, you use `void`
-as the return type. The asynchronous equivalent is `Future<void>`.
+as the return type. The asynchronous equivalent for a method which does not
+produce a value, but should have the work awaited, is `Future<void>`.
```
kwalrath: Is "should" required here? Mightn't it be optional? If so, maybe:
...but that the caller might need to await, ...

natebosch: Yeah it is optional. Will take your suggested wording.

src/_guides/language/effective-dart/design.md
```diff
@@ -1555,14 +1555,19 @@ bool convertToBool(dynamic arg) {
 ### DO use `Future<void>` as the return type of asynchronous members that do not produce values.
 
 When you have a synchronous function that doesn't return a value, you use `void`
-as the return type. The asynchronous equivalent is `Future<void>`.
+as the return type. The asynchronous equivalent for a method that doesn't
+produce a value, but that the caller might need to await, is `Future<void>`.
```
chalin: Maybe "await on"? @kwalrath WDYT?

src/_guides/language/effective-dart/design.md
```diff
@@ -1555,14 +1555,19 @@ bool convertToBool(dynamic arg) {
...
 
 You may see code that uses `Future` or `Future<Null>` instead because older
 versions of Dart didn't allow `void` as a type argument. Now that it does, you
 should use it. Doing so more directly matches how you'd type a similar
 synchronous function, and gives you better error-checking for callers and in the
 body of the function.
 
+For functions that are "fire and forget", which neither return a useful
+value nor need to have their asynchronous work finished before the caller
+continues, use a return type of `void`.
```
chalin: I'd drop the "fire and forget" (I don't think that we are trying to define/re-enforce the definition of such a term), and make it clear that we're talking about `async` functions (even if we are in a section discussing asynchronous members). How about:

> For async functions that do not return a useful value and do not need to have their asynchronous work awaited on, use a return type of `void`.

kwalrath: I thought functions marked `async` automatically had Future return values. Plus maybe we don't want to imply that this applies only to those kinds of functions. Instead of "async", should we say "asynchronous"?

chalin: Oh yes, sorry, I meant "asynchronous".

src/_guides/language/effective-dart/design.md
kwalrath: I had to look this up to be sure, but "await" as a verb already has an implied preposition—it's the equivalent of "wait for". So we wouldn't need the "on".

src/_guides/language/effective-dart/design.md
natebosch: Done

src/_guides/language/effective-dart/design.md
```diff
@@ -1555,14 +1555,18 @@ bool convertToBool(dynamic arg) {
...
+as the return type. The asynchronous equivalent for a method that doesn't
+produce a value, but that the caller might need to await, is `Future<void>`.
 
 You may see code that uses `Future` or `Future<Null>` instead because older
 versions of Dart didn't allow `void` as a type argument. Now that it does, you
 should use it. Doing so more directly matches how you'd type a similar
 synchronous function, and gives you better error-checking for callers and in the
 body of the function.
 
+For asynchronous functions that do not return a useful value and do not need to
```
munificent: "do not need" implies the wrong default. Async functions should almost always return a future. The only time they shouldn't is if they are OK with *no* caller ever being able to await the work.

How about "For asynchronous functions that do not return a useful value and where no callers need to await the asynchronous work or handle an asynchronous failure..."

?

natebosch: Good suggestion. Done.

PR #1212, opened 2018-10-27
Add build_runner page

kwalrath:
Staged: https://kw-www-dartlang-1.firebaseapp.com/tools/build_runner

### DIFF
--- a/src/tools/build_runner.md
+++ b/src/tools/build_runner.md
@@ -4,7 +4,8 @@ description: A tool for building, testing, and running Dart code.
 ---
 
 The [build_runner][] package provides general-purpose commands for generating files,
-and for optionally serving or testing the generated files.
+and for optionally testing the generated files
+or serving both source and generated files.
 Read this page for an overview of using build_runner, with links to
 where you can find more information.
 For details of using build_runner with a specific package,
@@ -71,7 +72,7 @@ serve
 : Runs a development server.
   Instead of directly using this command,
   you can use [`webdev serve`,][webdev serve]
-  which has better default behavior.
+  which has convenient default behavior.
 
 test
 : Runs [tests.][tests]


kwalrath:
Ready for review, @natebosch & @jakemac53!

One question I had was whether I should use **any** in the dev dependencies. It's technically OK, since (1) this code isn't auto-included (and thus not auto-updated) so it might bitrot unless we do add those tests, and (2) this is a dev dependency, so presumably you don't pay the performance penalty that you would for a dependency whose dependencies are followed. I'm just a little leery because we haven't used **any** anywhere else; I'd probably need to add an explanatory note here.

src/tools/build_runner.md
```diff
@@ -0,0 +1,111 @@
...
+The build_runner package includes the following commands:
+
+build
+: Performs a one-time build.
+
+serve
+: Runs a development server.
+  Instead of directly using this command,
+  you can use [`webdev serve`,][webdev serve]
+  which has better default behavior.
```
jakemac53: "better" is subjective haha, I would maybe say instead "which is more convenient to use"

jakemac53: although I guess that is subjective as well haha

kwalrath: Maybe "which has convenient default behavior". @kevmoo?

src/tools/build_runner.md
```diff
@@ -0,0 +1,111 @@
+---
+title: build_runner
+description: A tool for building, testing, and running Dart code.
+---
+
+The [build_runner][] package provides general-purpose commands for generating files,
+and for optionally serving or testing the generated files.
```
natebosch: Replace "the" with "with" - we serve both sources and generated files

src/tools/build_runner.md
```diff
@@ -0,0 +1,111 @@
...
+## More information
+
+If you're working on web-specific code,
+see the [webdev and build_runner page.][webdev]
+
+For details on using build_runner, see the following:
+
+- Documentation for packages that require you to use build_runner.
+  These packages generally have a dependency
+  [on build][] or [on build_runner.][]
```
natebosch: Packages that would require you to use `build_runner` would only depend on `build` or maybe `build_config` - but shouldn't ever have a non-dev dependency on `build_runner`...

src/tools/build_runner.md
```diff
@@ -0,0 +1,111 @@
...
+
+To use build_runner, add a [dev dependency][] on **build_runner**
+to your app's pubspec:
+
+```
+  dev_dependencies:
+    # ···
+    build_runner: ^1.0.0
+    build_test: ^0.10.3
+```
```
chalin: Kathy: We should be able to get this via a code excerpt, no?

src/tools/build_runner.md
kwalrath: How about:
and for optionally testing the generated files or serving both source and generated files
?

src/tools/build_runner.md
kwalrath: Is it important to distinguish between the two types of dependencies here?

kwalrath:
I've updated the source code for the changes that seemed most necessary. PTAL, @jakemac53 or @natebosch.

src/tools/build_runner.md
chalin: > It wasn't convenient ...

Would you like for me to look into it in a followup PR?

> ... note about that being OK because this is a dev dependency

@kevmoo are you ok with such a note?

src/tools/build_runner.md
jakemac53: For non-published packages it doesn't matter but we don't want people to accidentally publish with real build_runner dependencies.

src/tools/build_runner.md
kwalrath: A followup PR would be great. Thanks, @chalin!

src/tools/build_runner.md
kwalrath: I'm not instructing people to depend on those packages... I'm only providing a link to packages that *do* have those dependencies (as reported by pub.dartlang.org). Those links were handy for me, but maybe they're TMI. I could just delete or comment out this sentence.

src/tools/build_runner.md
chalin: #1217

kwalrath:
@jakemac53 & @natebosch, are we good to go?

PR #1222, opened 2018-11-07
Clarify "single-threaded" in Futures tutorial

### LINKED ISSUE #1218: Futures tutorial text is confusing: "dart is single-threaded"
https://github.com/dart-lang/site-www/blob/master/src/_tutorials/language/futures.md

The article states:
"Dart is single-threaded."
And "Dart is a single-threaded programming language."

This statement has confused at least one person - see this discussion:
https://www.reddit.com/r/dartlang/comments/9rs5d3/my_experiences_so_far/

Dart doesn't have shared-state multi-threading or locks, which I assume is the point which this article is trying to make. But isolates are a form of multi-threading.

Perhaps this would be better expressed as something like: "Futures are used to express concurrency within a single thread. (See isolates for running code on multiple cores)". 

I'm sure someone on the team can come up with something better than this - but I hope you get the gist ;)

kwalrath:
Staged: https://kw-www-dartlang-1.firebaseapp.com/tutorials/language/futures

Fixes #1218

### DIFF
--- a/src/_tutorials/language/futures.md
+++ b/src/_tutorials/language/futures.md
@@ -11,44 +11,49 @@ nextpage:
 <div class="mini-toc" markdown="1">
   <h4>What's the point?</h4>
 
-  * Dart code runs straight through, in a single "thread" of execution.
+  * Dart code runs in a single "thread" of execution.
   * Code that blocks the thread of execution can make your program freeze.
-  * `Future` objects (_futures_) represent asynchronous operations —
-    processing or I/O to be completed later.
+  * `Future` objects (_futures_) represent the results of
+    _asynchronous operations_ — processing or I/O to be completed later.
   * To suspend execution until a future completes,
     use `await` in an async function (or use `then()`).
   * To catch errors, use try-catch expressions in async functions
     (or use `catchError()`).
   * To run code concurrently, create an _isolate_
-    (or for web apps, a _worker_).
+    (or for a web app, a _worker_).
 </div>
 
-Dart code runs in a predictable order, in a single "thread" of execution.
-Any code that blocks that thread of execution
-(for example, by performing a long-running calculation or blocking on I/O)
-can freeze a Dart program. Asynchronous operations let
-your program complete other work while waiting for an operation to complete.
-Dart uses `Future` objects to represent asynchronous operations.
+Dart code runs in a single "thread" of execution.
+If Dart code blocks — for example, by performing a
+long-running calculation or waiting for I/O — the entire program freezes.
+
+Asynchronous operations let your program complete other work while
+waiting for an operation to finish.
+Dart uses `Future` objects (futures) to represent the results of
+asynchronous operations.
+To work with futures,
+you can use either `async` and `await` or the `Future` API.
 
 <aside class="alert alert-info" markdown="1">
   **Note:**
-  All Dart code runs in the context of an **isolate** that
+  All Dart code runs in the context of an _isolate_ that
   owns all of the memory that the Dart code uses.
   While Dart code is executing, no other code in the same isolate can run.
-  This page describes how to pause execution and
-  give control to another part of the program (in the same isolate)
-  using `async` and `await` or the `Future` API.
 
-  <br>If you want multiple parts of the program to run concurrently,
-  you can run them in separate isolates (or for web apps, workers).
+  <br>
+  If you want multiple parts of Dart code to run concurrently,
+  you can run them in separate isolates.
+  (Web apps use workers instead of isolates.)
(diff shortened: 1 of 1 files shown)


src/_tutorials/language/futures.md
```diff
@@ -11,22 +11,41 @@ nextpage:
...
+  * To catch errors, use try-catch expressions in async functions
+    (or use `catchError()`).
+  * To run code concurrently, create an _isolate_
+    (or for web apps, a _worker_).
 </div>
 
-Dart is a single-threaded programming language. If any code blocks the thread
-of execution (for example, by performing a long-running calculation or
-blocking on I/O) the program effectively freezes. Asynchronous operations let
+Dart code runs in a predictable order, in a single "thread" of execution.
```
lrhn: Repeats the phrase used earlier, but with a different first half. Is that significant?

src/_tutorials/language/futures.md
```diff
@@ -11,22 +11,41 @@ nextpage:
...
 
-  * Dart is single-threaded.
-  * Synchronous code can make your program freeze.
-  * Use `Future` objects (_futures_) to perform asynchronous operations.
-  * Use `await` in an async function to suspend execution until a future completes.
-  * Or use the `then()` method.
-  * Use try-catch expressions in async functions to catch errors.
-  * Or use the `catchError()` method.
-  * You can chain futures to run asynchronous functions in order.
+  * Dart code runs straight through, in a single "thread" of execution.
```
lrhn: I find that description confusing, because it seems to ignore "async" completely.

I think it's mostly the "straight through" phrase which doesn't match the way I think of an event loop.

I would rather say that:

> The Dart *language* is single-threaded. It has no concept of *concurrency*. A Dart program only executes one thing at a time, and code which blocks will block the entire program.

... then move onto isolates, which allow running multiple independent Dart programs concurrently, and async, which allows interleaving code inside a single isolate.

(Still not great, I admit).

src/_tutorials/language/futures.md
lrhn: I also oppose "predictable order" since it's ... not. Not really. We try to make futures complete in a *deterministic* order, but I wouldn't claim that the order is *predictable* to users. So, maybe, "deterministic".

Example:
```dart
var f1 = Future.value(42);
var f2 = f1.then((x) { print("f2"); return x + 1; });
var f3 = f1.then((x) { print("f3"); return x + 2; });
var f4 = f2.then((x) { print("f4"); return x + 3; });
```
I'm not sure how many will predict that this prints: "f2", "f4", "f3", in that order.

src/_tutorials/language/futures.md
```diff
@@ -11,22 +11,41 @@ nextpage:
...
+can freeze a Dart program. Asynchronous operations let
 your program complete other work while waiting for an operation to complete.
 Dart uses `Future` objects to represent asynchronous operations.
 
+<aside class="alert alert-info" markdown="1">
+  **Note:**
+  All Dart code runs in the context of an **isolate** that
+  owns all of the memory that the Dart code uses.
+  While Dart code is executing, no other code in the same isolate can run.
+  This page describes how to pause execution and
```
lrhn: "pause" in relation to isolates usually means something else.
Maybe "delay" (since "yield" also means something else, all the good names are taken!)

src/_tutorials/language/futures.md
```diff
@@ -11,22 +11,41 @@ nextpage:
...
-  * Synchronous code can make your program freeze.
-  * Use `Future` objects (_futures_) to perform asynchronous operations.
-  * Use `await` in an async function to suspend execution until a future completes.
-  * Or use the `then()` method.
-  * Use try-catch expressions in async functions to catch errors.
-  * Or use the `catchError()` method.
-  * You can chain futures to run asynchronous functions in order.
+  * Dart code runs straight through, in a single "thread" of execution.
+  * Code that blocks the thread of execution can make your program freeze.
+  * `Future` objects (_futures_) represent asynchronous operations —
```
lrhn: ... represents the results of asynchronous operations

(Pet peeve, sorry, but I do think it's an important distinction).

src/_tutorials/language/futures.md
kwalrath: Thanks for the comments, @lrhn. I'm worried about equating isolates with programs, but I'm working on some wording that might work...

kwalrath:
I've updated the text... ptal.

kwalrath:
Thanks, @lrhn! In the previous layout of this page, the repetition wouldn't have been as noticeable. I still need to adjust to the new, less subtle layout.

PR #1232, opened 2018-11-16
Update version SDK version from 2.0.0 to 2.1.0.

### LINKED ISSUE #1231: Dart version dropdown list in menu has 2.0.0 instead of 2.1.0
Reported via gitter https://gitter.im/dart-lang/TALK-general?at=5beeadd61665691e0fb1db5f

kwalrath:
Fixes #1231

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -4,7 +4,7 @@ author: Dart Team <misc@dartlang.org>
 homepage: https://webdev.dartlang.org
 
 environment:
-  sdk: '>=2.0.0 <3.0.0'
+  sdk: '>=2.1.0 <3.0.0'
 
 dev_dependencies:
   build_runner: ^0.9.1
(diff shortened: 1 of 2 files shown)


pubspec.yaml
```diff
@@ -4,7 +4,7 @@ author: Dart Team <misc@dartlang.org>
 homepage: https://webdev.dartlang.org
 
 environment:
-  sdk: '>=2.0.0 <3.0.0'
```
chalin: This pubspec is for the site's tooling. Leaving it at 2.0.0 should be ok.

chalin: Unless, of course, you know that some of the packages _require_ 2.1.0.

kwalrath: Ah, I wondered about that. Thought I'd have Travis test this. :)

chalin: Travis won't complain about having a tighter bound. It'll just prevent us from running the local tools on an SDK below 2.1.0.

kwalrath:
I think some dev tests might fail, but I'll let Travis be the judge of that (and then temporarily disable those tests).

PR #1251, opened 2018-12-03
Add a bit more about the library directive.

### LINKED ISSUE #1249: The library keyword is mentioned but not explained
The `library` keyword is mentioned on the Library section, almost at the end, but it isn't explained, nor there's a reference to read more about it.

kwalrath:
Fixes #1249.

Also update site-shared.


kwalrath:
Nope, this is all. Thanks, @chalin. :)

PR #1276, opened 2018-12-24
add responsive table in content block

Arhell:
Hello.
on mobile devices, the table is sticking out
example 
https://www.dartlang.org/guides/get-started

### DIFF
--- a/src/_assets/css/main.scss
+++ b/src/_assets/css/main.scss
@@ -1279,6 +1279,16 @@ body.obsolete {
       }
     }
   }
+
+  .content {
+    table {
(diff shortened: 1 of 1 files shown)


chalin:
Thanks for the PR, though we might handle the issue slightly differently. I'll followup as soon as I can in the next couple of days.

chalin:
Thanks @Arhell for pointing out the issue. #1278 addresses the Platforms table in the Get Started page. We'll handle other tables on a case-by-case basis. Feel free to file an issue if you find another mis-formatted table or page.

PR #1283, opened 2018-12-27
add responsive table (tools)

Arhell:
Hello.
Fix table on https://www.dartlang.org/tools
after 
http://joxi.ru/DrlKxVbCvlY7GA
Thank for #1278


kwalrath:
I see no visible difference between the two. Can you tell me why we should make this change? Thanks, @Arhell!

Arhell:
Hi, @kwalrath 
if you go to the page and put a resolution of 320, then the table will pop out beyond the boundaries.

PR #1288, opened 2018-12-29
Fix the stagehand url link

netroby:


netroby:
see #1289

chalin:
Fixes #1289

PR #1291, opened 2019-01-02
Document `repository` and `issue_tracker` properties

jonasfj:
As added in: https://github.com/dart-lang/pubspec_parse/pull/30

### DIFF
--- a/src/tools/pub/pubspec.md
+++ b/src/tools/pub/pubspec.md
@@ -31,8 +31,16 @@ A pubspec can have the following fields:
   [_Learn more._](#authorauthors)
 
 `homepage`
-: Optional.
+: Optional. URL pointing to the package's homepage (or source code repository).
   [_Learn more._](#homepage)
+
+`repository`
+: Optional. URL pointing to the package's source code repository.
+  [_Learn more._](#repository)
+
+`issue_tracker`
+: Optional. URL pointing to an issue tracker for the package.
+  [_Learn more._](#issue-tracker)
 
 `documentation`
 : Optional. Can be used to automatically create documentation.
@@ -176,10 +184,26 @@ public.
 This should be a URL pointing to the website for your package.
 For [hosted packages](/tools/pub/dependencies#hosted-packages),
 this URL is linked from the package's page.
-While this is technically optional *please do* provide one. It
-helps users understand where your package is coming from. If nothing else, you
-can always use the URL where you host the source code, such as
-[GitHub](https://github.com).
+While providing a `homepage` is optional, *please provide* it or `repository`
+(or both). It helps users understand where your package is coming from.
+
+### Repository
+
+The optional `repository` field should contain the URL for your package's source
+code repository — for example, `https://github.com/<user>/<repository>`.
+If you publish your package to pub.dartlang.org, then your package's page
+displays the repository URL.
+While providing a `repositoryu` is optional, *please provide* it or `homepage`
+(or both). It helps users understand where your package is coming from.
+
+### Issue tracker
+
+The optional `issue_tracker` field should contain a URL for the package's
+issue tracker, where existing bugs can be viewed and new bugs can be filed.
+The pub.dartlang.org site attempts to display a link to each package's issue
+tracker, using the value of this field. If `issue_tracker` is missing but
+`repository` is present and points to GitHub, then pub.dartlang.org uses the
+default issue tracker (`https://github.com/<user>/<repository>/issues`).
 
 ### Documentation
 


src/tools/pub/pubspec.md
```diff
@@ -181,6 +189,24 @@ helps users understand where your package is coming from. If nothing else, you
 can always use the URL where you host the source code, such as
 [GitHub](https://github.com).
 
+### Repository
+
+The `repository` property should be a URL pointing to the repository for the
```
kwalrath: Let's tighten this up a bit and refer to the homepage field. How about:

> This optional field should contain the URL for your package's source code — for example, `https://github.com/<user>/<repository>`. If you publish your package to pub.dartlang.org, then your package's page displays the repository URL. Also see the [homepage field](#homepage).

We should adjust the **Homepage** section as well to reflect how it interacts with `repository`.

@kevmoo, what should we say in the `homepage` section? Maybe change the last sentence from "If nothing else, you can always use the URL where you host the source code, such as GitHub." to something like this: "If nothing else, omit the [`repository` field](#repository) and set `homepage` to the URL where you host the source code."

src/tools/pub/pubspec.md
```diff
@@ -181,6 +189,24 @@ helps users understand where your package is coming from. If nothing else, you
...
+package. This is often the URL for a github repository:
+`https://github.com/<user>/<repository>`.
+For [hosted packages](/tools/pub/dependencies#hosted-packages),
+this URL will be displayed on package's page.
+
+### Issue Tracker
```
kwalrath: Tracker -> tracker
(we use sentence case for section heads)

src/tools/pub/pubspec.md
```diff
@@ -181,6 +189,24 @@ helps users understand where your package is coming from. If nothing else, you
...
+
+The `issue_tracker` property should be a URL pointing to the issue tracker for
```
kwalrath: How about:

Help your package's users find and file bugs in the right place by putting a URL in the `issue_tracker` field. The pub.dartlang.org site attempts to display a link to each package's issue tracker, using the value of this field. If `issue_tracker` is missing but [`repository`](#repository) is present and points to GitHub, then pub.dartlang.org uses the default issue tracker (`https://github.com/<user>/<repository>/issues`).

src/tools/pub/pubspec.md
```diff
@@ -34,6 +34,14 @@ A pubspec can have the following fields:
 : Optional.
   [_Learn more._](#homepage)
 
+`repository`
+: Optional. URL pointing to the repository.
```
kwalrath: repository -> the package's source code

(or something else that doesn't just repeat "repository". we could say only "Optional. Learn more." like some other entries, but I like your idea of giving _some_ info here.)

src/tools/pub/pubspec.md
jonasfj: Actually, I don't think we care if people provide both `homepage` and `repository`, or just one of them... either way is probably fine :)

So let's add our plea to both places:
> While this is technically optional *please do* provide either a `homepage`
or a `repository`. It helps users understand where your package is coming from.

src/tools/pub/pubspec.md
```diff
@@ -176,10 +184,26 @@ public.
 This should be a URL pointing to the website for your package.
 For [hosted packages](/tools/pub/dependencies#hosted-packages),
 this URL is linked from the package's page.
-While this is technically optional *please do* provide one. It
-helps users understand where your package is coming from. If nothing else, you
-can always use the URL where you host the source code, such as
-[GitHub](https://github.com).
+While this is technically optional *please do* provide either a `homepage`
+or a `repository`. It helps users understand where your package is coming from.
+
```
kwalrath: Both is OK, too, right? "either...or" makes it sound like you can't do both. How about:

While providing a `homepage` is optional, please provide it or `repository` (or both).
These fields help users understand...

src/tools/pub/pubspec.md
```diff
@@ -176,10 +184,26 @@ public.
...
+### Repository
+
+The optional `repository` field should contain the URL for your package's source
+code repository — for example, `https://github.com/<user>/<repository>`.
+If you publish your package to pub.dartlang.org, then your package's page
+displays the repository URL.
+While this is technically optional *please do* provide either a `homepage`
```
kwalrath: Same as above, except switch `homepage` and `repository`.

src/tools/pub/pubspec.md
```diff
@@ -176,10 +184,26 @@ public.
...
+or a `repository`. It helps users understand where your package is coming from.
+
+### Issue tracker
+
+The optional `issue_tracker` field should contain a URL for the package's
+issue tracker, where existing bugs can be viewed and new bugs can be filed.
+The pub.dartlang.org site attempts to display a link to each package's issue
+tracker, using the value of this field. If `issue_tracker` is missing, but
+repository is present and points to GitHub, then pub.dartlang.org uses the
```
kwalrath: , but repository -> but `repository`
(no comma, add code font)

src/tools/pub/pubspec.md
jonasfj: good point :)

kwalrath:
LGTM. There's a tiny typo, but I'll fix that after committing. Thanks for the docs, @jonasfj!

PR #1302, opened 2019-01-04
remove advice to prefer constructors to static methods

### LINKED ISSUE #1194: "PREFER defining constructors ..." presumes mandatory "new"
Now that "new" is optional and usually omitted, the guidance around when to prefer a constructor over a static method is different. We should remove or re-write that guideline.

kwalrath:
Fixes #1194

### DIFF
--- a/examples/misc/lib/effective_dart/design_bad.dart
+++ b/examples/misc/lib/effective_dart/design_bad.dart
@@ -1,7 +1,6 @@
 // ignore_for_file: type_annotate_public_apis, unused_element, unused_local_variable, avoid_types_as_parameter_names
 
 import 'dart:async';
-import 'dart:math';
 
 import 'package:dartlang_examples_util/ellipsis.dart';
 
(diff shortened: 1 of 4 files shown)


examples/misc/lib/effective_dart/design_bad.dart
```diff
@@ -1,7 +1,6 @@
 // ignore_for_file: type_annotate_public_apis, unused_element, unused_local_variable, avoid_types_as_parameter_names
 
 import 'dart:async';
```
chalin: Remove the `import 'dart:async'`?

examples/misc/lib/effective_dart/design_good.dart
```diff
@@ -3,7 +3,6 @@
 import 'dart:async';
```
chalin: Remove the `import 'dart:async'`?

kwalrath: Good idea! I just tried and it won't work, though — both files use `FutureOr`.

PR #1328, opened 2019-01-29
Update analysis options for pedantic

### LINKED ISSUE #1322: Show pedantic in analysis-options
Soon all the stagehand templates will use pedantic (dart-lang/stagehand#594). We should show that and maybe talk about how to tie to a specific version of pedantic.

Along the way, we'll have to document include (#971).

kwalrath:
...and more changes to fix issues I had with the page. WIP because the page has a few questions (please look for "PENDING").

Staged: https://kw-www-dartlang-1.firebaseapp.com/guides/language/analysis-options

Related: dart-lang/stagehand#594

Fixes #1322.

### DIFF
--- a/src/_guides/language/analysis-options.md
+++ b/src/_guides/language/analysis-options.md
@@ -6,7 +6,9 @@ description: Customize static analysis using an analysis options file.
 Static analysis allows you to find problems before
 executing a single line of code. It's a powerful tool
 used to prevent bugs and ensure that code conforms to style
-guidelines. With the help of the analyzer, you can find
+guidelines.
+
+With the help of the analyzer, you can find
 simple typos. For example, perhaps an accidental semicolon
 made its way into an `if` statement:
 
@@ -45,7 +47,7 @@ add static analysis to your tool, see the
 <aside class="alert alert-info" markdown="1">
 **Note:**
 The analyzer error codes are listed in the [Dart SDK
-repo.](https://github.com/dart-lang/sdk/blob/master/pkg/analyzer/lib/error/error.dart)
+repo.][analyzer error codes]
 </aside>
 
 ## The analysis options file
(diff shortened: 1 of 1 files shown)


src/_guides/language/analysis-options.md
```diff
@@ -67,30 +69,34 @@ at the root of the package, in the same directory as the pubspec file.
...
-and use 2 spaces to denote each level of indentation.
+The <code>include: <em>url</em></code> entry
+brings in YAML code from the specified URL — in this case,
+from a file in the `pedantic` package.
+The `linter:` entry [enables linter rules](#enabling-linter-rules).
+You can use the `analyzer:` entry to customize static analysis —
+[enabling stricter type checks](#enabling-additional-type-checks), 
+[excluding files](#excluding-files),
+suppressing specific rules [for a file](#suppressing-rules-for-a-file) or
+[for a line of code](#suppressing-rules-for-a-line-of-code),
```
bwilkerson: Users can't actually suppress a rule for either a file or a single line from within the analysis options file. Doing so requires adding 'ignore' comments to a source file.

src/_guides/language/analysis-options.md
```diff
@@ -121,15 +127,10 @@ analyzer:
 
 You can use the flags together or separately;
 both default to `true`.
-The presence of either flag, regardless of value, enables
-the Dart 2 type system.
 
-{% comment %}
-**PENDING:
-Will these flags still appear under strong-mode in Dart 2.0?
+**[PENDING: Do these flags still appear in Dart 2.0/2.1?
```
bwilkerson: If by "these flags" you mean 'implicit-casts' and 'implicit-dynamic', then yes.

Possible future: Leaf has a proposal to split those up into three flags, but that hasn't been decided on.

src/_guides/language/analysis-options.md
```diff
@@ -121,15 +127,10 @@ analyzer:
...
 Should we mention related command-line flags
-(--no-implicit-casts, --no-implicit-dynamic)?**
-{% endcomment %}
+(--no-implicit-casts, --no-implicit-dynamic)?]**
```
bwilkerson: The flags are still available, but in general I think users are better off using the analysis options file because they'll get the same behavior from both the command-line and any server-based tooling they use, and everyone who works on the package will get the same behavior as well.

src/_guides/language/analysis-options.md
```diff
@@ -212,77 +258,105 @@ analyzer:
...
 {% endprettify %}
 
-This analysis options file instructs the analysis tools to ignore
-the TODO rule.
-
-Alternatively, as of Dart 1.24 you can ignore a specific rule for a
-specific file using an `ignore_for_file` comment:
+**[PENDING: The wording implies that `ignore` works only for analyzer error codes,
+not for linter rules.
+Could you also use `ignore` with an included bunch of linter rules (e.g. pedantic)?]**
```
bwilkerson: Yes, it works with lint rules too.

Possible future: There's discussion that we might stop allowing users to ignore errors, but no decision yet.

src/_guides/language/analysis-options.md
pq: Good catch.  This would still be great to include but clarifying that this is done in source and not options is a good idea.

src/_guides/language/analysis-options.md
```diff
@@ -67,30 +69,34 @@ at the root of the package, in the same directory as the pubspec file.
...
+[ignoring specific rules](#ignoring-rules), or
+[changing the severity of rules](#changing-the-severity-of-rules).
+Another tag you might see is `language:`,
+which is used for experimental language features.
 
 <aside class="alert alert-info" markdown="1">
-**Note**: You might come across a `language:` tag in an analysis options file.
-This tag is used for testing experimental features. You can ignore it.
+  **YAML is sensitive to whitespace.** Don't use tabs in a YAML file,
+  and use 2 spaces to denote each level of indentation.
```
pq: Nice 👍

src/_guides/language/analysis-options.md
```diff
@@ -155,14 +156,44 @@ TODO: Clarify that description, and insert an example here.
...
+```
+
+<aside class="alert alert-warning" markdown="1">
+  **Note:**
+  When a **new version of `pedantic`** is published,
+  code that previously passed analysis might **start failing analysis.**
+  We recommend updating your code to work with the new rules.
+  Other options are to [use a specific version][]
+  of `pedantic` or to explicitly enable individual linter rules.
+</aside>
```
pq: Oh, this is just great.  Thank you!

src/_guides/language/analysis-options.md
```diff
@@ -212,77 +258,105 @@ analyzer:
...
+
+`info`
+: An informational message that doesn't cause analysis to fail.
+  Example: [`todo`][todo]
+
+`warning`
+: A warning that doesn't cause analysis to fail unless
+  the analyzer is configured to treat warnings as errors.
+  Example: Linter rules, such as [`avoid_as`][avoid_as]
+  **[PENDING: are all linter rules warnings by default?]**
```
pq: @bwilkerson could you speak to this?

src/_guides/language/analysis-options.md
kwalrath: OK, I'll clarify that. Good catch!

src/_guides/language/analysis-options.md
kwalrath: Makes sense. Thanks, @bwilkerson!

src/_guides/language/analysis-options.md
kwalrath: Turns out they're all "info".

Issue #1332, opened 2019-01-30
'Effective Dart: Usage' page issue


patlillis:
Page URL: https://www.dartlang.org/guides/language/effective-dart/usage
Page source: https://github.com/dart-lang/site-www/tree/master/src/_guides/language/effective-dart/usage.md

Description: "DON’T use `this.` when not needed to avoid shadowing." needs an exception for redirecting to a named constructor.

That guideline states "The only time you need to use `this.` is when a local variable with the same name shadows the member you want to access.". This is not true, you also need it when redirecting to a named constructor.

E.g. following does not compile:
```dart
class Thingy {
  Thingy._black();

  Thingy.blue() : _black();
}
```

Instead it needs to be:
```dart
class Thingy {
  Thingy._black();

  Thingy.blue() : this._black();
}
```

munificent:
Fixed by #1334.

PR #1323, opened 2019-01-23
Effective Dart: underscores in directory names

kwalrath:
Also fixed the package issues in the TOC autogenerator, but now it's not
so automatic any more. (@filiph, any ideas?)

Closes #1215.

PR #1334, opened 2019-01-31
Updating uses of 'this.' in Effective Dart.

RedBrogdon:
Tweaks the text of the "this." section in Effective Dart's usage file. Also adds code examples to the relevant files.

### DIFF
--- a/src/_guides/language/effective-dart/usage.md
+++ b/src/_guides/language/effective-dart/usage.md
@@ -938,7 +938,7 @@ set x(num value) => center = Point(value, center.y);
 {% endprettify %}
 
 
-### DON'T use `this.` when not needed to avoid shadowing.
+### DON'T use `this.` except to avoid shadowing or redirect a named constructor.
 
 {% include linter-rule.html rule="unnecessary_this" %}
 
@@ -946,8 +946,8 @@ JavaScript requires an explicit `this.` to refer to members on the object whose
 method is currently being executed, but Dart&mdash;like C++, Java, and
 C#&mdash;doesn't have that limitation.
 
-The only time you need to use `this.` is when a local variable with the same
-name shadows the member you want to access.
+There are only two times you need to use `this.`. One is when a local variable
+with the same name shadows the member you want to access:
 
 {:.bad-style}
 <?code-excerpt "misc/lib/effective_dart/usage_bad.dart (this-dot)"?>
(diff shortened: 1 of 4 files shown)


RedBrogdon:
Fixes #1332.

patlillis:
Minor nit: you need `this.` when redirecting __to__ a named constructor. It's possible to redirect a named constructor to the default constructor using only "this" and no period:

```dart
class ShadeOfGray {
  final int brightness;

  ShadeOfGray(int val) : brightness = val;

  // Do not need to use "this." here, only "this" since it's
  // redirecting to the default constructor.
  ShadeOfGray.black() : this(0);

  // Do need to use "this." here, since it's redirecting to
  // a named constructor.
  ShadeOfGray.reallyBlack() : this.black();
}
```

kwalrath:
@RedBrogdon are you going to be able to fix this PR, or should I close it?

RedBrogdon:
I still plan to fix it, just got bogged down with some issues on DartPad today. I haven't forgotten this, though, and want to close it out with a merge.

mit-mit:
Any news @RedBrogdon ?

RedBrogdon:
PTAL!

src/_guides/language/effective-dart/usage.md
```diff
@@ -938,16 +938,16 @@ set x(num value) => center = Point(value, center.y);
 {% endprettify %}
 
 
-### DON'T use `this.` when not needed to avoid shadowing.
+### DON'T use `this.` except to avoid shadowing or redirect to a named constructor.
```
kwalrath: I'm tempted to say or redirect -> or to redirect (since I first read it as "to avoid redirect", but then we have so many "to"s in a row...

src/_guides/language/effective-dart/usage.md
```diff
@@ -938,16 +938,16 @@ set x(num value) => center = Point(value, center.y);
...
 
 {% include linter-rule.html rule="unnecessary_this" %}
 
 JavaScript requires an explicit `this.` to refer to members on the object whose
 method is currently being executed, but Dart&mdash;like C++, Java, and
 C#&mdash;doesn't have that limitation.
 
-The only time you need to use `this.` is when a local variable with the same
-name shadows the member you want to access.
+There are only two times you need to use `this.`. One is when a local variable
```
kwalrath: We generally avoid empty words like "There are". But I'm blanking on how to fix it, so ... let it be.

src/_guides/language/effective-dart/usage.md
kwalrath: We need to update the big list o' links in https://www.dartlang.org/guides/language/effective-dart#usage. To keep old links working, we should also specify an anchor that's the same as the old link.

An example of doing this is in https://github.com/dart-lang/site-www/pull/1323.

kwalrath:
I've made the anchor fix. OK by you, @RedBrogdon?

kwalrath:
Staged:
https://kw-www-dartlang-1.firebaseapp.com/guides/language/effective-dart/usage#dont-use-this-when-not-needed-to-avoid-shadowing
https://kw-www-dartlang-1.firebaseapp.com/guides/language/effective-dart#usage

RedBrogdon:
LGTM.

Thanks for the help. I spent several hours yesterday trying to get jekyll up and running and kept running into a dependency, nokogiri, that wouldn't build on my machine.

kwalrath:
> I spent several hours yesterday trying to get jekyll up and running and kept running into a dependency, nokogiri, that wouldn't build on my machine.

:( I'm sorry that was so difficult.

RedBrogdon:
It's most likely some weird config problem on my machine rather than anything to do with your stack or instructions. 😄

PR #1342, opened 2019-02-04
Update text on "pub cache repair"

jonasfj:
* Avoid talking about the `packages/` folder as something that contains symlinks.
 * Mention `.packages` and how your editor might follow references in this file.

### DIFF
--- a/src/tools/pub/cmd/pub-cache.md
+++ b/src/tools/pub/cmd/pub-cache.md
@@ -45,8 +45,8 @@ versions.</dd>
 
 <dt><code>repair</code></dt>
 <dd>It's possible for packages in your pub cache to change or break.
-For example, if you follow a symlink in the <code>packages</code> directory
-and modify a package, this affects the canonical copy of that package in
+For example, if your editor follows the reference in the <code>.packages</code> file
+and you modify a file in a package, this affects the canonical copy of that package in
 the system cache. The <code>pub cache repair</code> command performs a clean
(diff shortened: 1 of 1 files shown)


jonasfj:
It's was rather random that I noticed this... I suppose it's possible that all the docs on `pub` could use a quick review. The example above on how to install `barback` probably isn't ideal either. We could change the example to install the `test` package maybe... anyways, when we have a rainy day maybe.

kwalrath:
Good catch! I wonder if we need to get into such detail. Maybe we could just say:

<dd>It's possible for packages in your pub cache to change or break.
For example, you might accidentally edit or delete a file in the pub cache.
The <code>pub cache repair</code> command performs a clean reinstall of all
hosted and git packages in the system cache.</dd>

kwalrath:
My IDE gives me warnings if I try to do that. :) (Although it does give me the option of saving the file anyway.) Your IDE doesn't warn you?

jonasfj:
I don't think vscode gave me a warning... If so I happily ignored it :)

Anyways, I'm perfectly okay with either solution... Feel free to merge this
or close it and write a different text..

Main point of this PR was to ensure we cleaned up the worst legacy..

kwalrath:
Sorry for the delay on this. I aim to commit this PR by EOW.

kwalrath:
I'm proposing some text that doesn't mention `.packages`.

PR #1349, opened 2019-02-13
Update return types in the design section of the dart style guide

jcabreur:


jcabreur:
I am Googler, I  do not need to sign a CLA.

kwalrath:
Thanks for the PR! Could you also update the code files that this pulls from?
https://github.com/dart-lang/site-www/blob/master/examples/misc/lib/effective_dart/design_good.dart
https://github.com/dart-lang/site-www/blob/master/examples/misc/lib/effective_dart/design_bad.dart

If that's not convenient, I can commit this PR and update those files to unbreak the build.

/cc @munificent

munificent:
Technically, Dart will infer a `bool` return type here, so the annotation isn't necessary. But, in practice, most Dart users do seem to prefer writing them. 👍

jcabreur:
I updated the PR with the changes to design_good.dart and design_bad.dart files.

kwalrath:
Perfect, thanks!

PR #1361, opened 2019-02-21
Add sets to language tour

### LINKED ISSUE #1257: Cover set literals in the language tour
On the page https://www.dartlang.org/guides/language/language-tour#built-in-types is not Set a built-in type?

kwalrath:
Fixes #1257.

Staged: https://kw-www-dartlang-1.firebaseapp.com/guides/language/language-tour#sets

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -740,14 +740,14 @@ var halogens = {'fluorine', 'chlorine', 'bromine', 'iodine', 'astatine'};
 
 <aside class="alert alert-info" markdown="1">
   **Note:**
-  The analyzer infers that `halogens` has the type
+  It is inferred that `halogens` has the type
   `Set<String>`. If you try to add the wrong type of value
   to the set, the analyzer or runtime raises an error.
   For more information, read about
   [type inference.](/guides/language/sound-dart#type-inference)
 </aside>
 
-To create an empty set, use `{}` preceded by a generic type annotation,
+To create an empty set, use `{}` preceded by a type argument,
 or assign `{}` to a variable of type `Set`:
 
 <?code-excerpt "misc/lib/language_tour/built_in_types.dart (set-vs-map)"?>
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -406,6 +406,7 @@ The Dart language has special support for the following types:
 - strings
 - booleans
 - lists (also known as *arrays*)
+- sets
 - maps
 - runes (for expressing Unicode characters in a string)
 - symbols
@@ -557,12 +558,9 @@ object’s `toString()` method.
 var s = 'string interpolation';
 
 assert('Dart has $s, which is very handy.' ==
-    'Dart has string interpolation, ' +
-        'which is very handy.');
-assert('That deserves all caps. ' +
-        '${s.toUpperCase()} is very handy!' ==
-    'That deserves all caps. ' +
-        'STRING INTERPOLATION is very handy!');
+    'Dart has string interpolation, ' + 'which is very handy.');
+assert('That deserves all caps. ' + '${s.toUpperCase()} is very handy!' ==
+    'That deserves all caps. ' + 'STRING INTERPOLATION is very handy!');
 {% endprettify %}
 
 <div class="alert alert-info" markdown="1">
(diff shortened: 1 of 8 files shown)


mit-mit:
cc @lrhn can you give this a review?

src/_guides/language/language-tour.md
```diff
@@ -720,6 +718,64 @@ information about lists, see [Generics](#generics) and
 [Collections](/guides/libraries/library-tour#collections).
 
 
+### Sets
```
mit-mit: We should add something about how to create a set literal for an empty set (that needs disambiguation with the map literal, so needs a type on either the left or right side, see: https://github.com/dart-lang/language/blob/master/accepted/future-releases/set-literals/feature-specification.md#examples)

kwalrath: Done. PTAL.

kwalrath:
Tests are failing because https://storage.googleapis.com/dart-archive/channels/dev/release/latest/sdk/dartsdk-linux-x64-release.zip is 2.1.2-dev.0.0. It should be 2.2.0-dev.2.1. We could push manually if necessary.

kwalrath:
Tests are passing! @mit-mit, feel free to merge this if you're OK with the wording.

src/_guides/language/language-tour.md
```diff
@@ -720,6 +721,83 @@ information about lists, see [Generics](#generics) and
...
+Here is a simple Dart set, created using a set literal:
+
+<?code-excerpt "misc/lib/language_tour/built_in_types.dart (set-literal)"?>
+{% prettify dart %}
+var halogens = {'fluorine', 'chlorine', 'bromine', 'iodine', 'astatine'};
+{% endprettify %}
+
+<aside class="alert alert-info" markdown="1">
+  **Note:**
+  The analyzer infers that `halogens` has the type
```
lrhn: Why only "the analyzer"?
Maybe "The Dart language " ...

src/_guides/language/language-tour.md
```diff
@@ -720,6 +721,83 @@ information about lists, see [Generics](#generics) and
...
+  `Set<String>`. If you try to add the wrong type of value
+  to the set, the analyzer or runtime raises an error.
+  For more information, read about
+  [type inference.](/guides/language/sound-dart#type-inference)
+</aside>
+
+To create an empty set, use `{}` preceded by a generic type annotation,
```
lrhn: "a generic type annotation" -> "a type argument"
(if you mean to write, e.g., `<int>{}`, but maybe just write that example instead of only describing it in prose).

Ob-pedantry: Something is generic if it works for multiple types, usually meaning that it is parameterized by type. The `List` class is generic, but `List<int>` is not. A "generic type argument" is not really correct, but semantically close enough that the meaning is clear - a type argument for something generic, even if the type argument is not *itself* generic).
Also, `<int>` is a type argument, not a type annotation. Type annotations are types like `int x;` in a variable declaration.

src/_guides/language/language-tour.md
```diff
@@ -720,6 +721,83 @@ information about lists, see [Generics](#generics) and
...
+<?code-excerpt "misc/test/language_tour/built_in_types_test.dart (set-length)"?>
+{% prettify dart %}
+var elements = <String>{};
+elements.add('fluorine');
+elements.addAll(halogens);
+assert(elements.length == 5);
+{% endprettify %}
+
+To create a set that's a compile-time constant,
+add `const` before the set literal:
```
lrhn: ... or just use it in a constant context

....

const colorGroups = { "warm": {"red", "yellow", "orange"}, "cold": {"green", "blue", "purple"}};

src/_guides/language/language-tour.md
mit-mit: Changed to the more neutral 'It is inferred that...'

src/_guides/language/language-tour.md
mit-mit: Done

src/_guides/language/language-tour.md
mit-mit: Didn't get this done

kwalrath:
Thanks for the edits, @mit-mit. I can do a follow-up PR later.

src/_guides/language/language-tour.md
kwalrath: To avoid the passive voice, I've changed this to "Dart infers" here. I've also updated the other two places where we had "The analyzer infers".

If we want to be more precise, we could say "Dart tools infer that..." but I'm not sure whether that's helpful.

src/_guides/language/language-tour.md
lrhn: This creates a compile-time constant without writing `const` before the literal (well, not *just* before the literal).

PR #1385, opened 2019-03-06
Call out `List.of` in the tip for `List.from`

natebosch:
When copying is the main goal `List.of` better expresses the intent
than `toList()` because its a constructor call rather than something
that looks like a transformation of the original.

### DIFF
--- a/src/_guides/language/effective-dart/usage.md
+++ b/src/_guides/language/effective-dart/usage.md
@@ -376,9 +376,9 @@ numbers.removeAt(1); // Now it only contains integers.
 var ints = List<int>.from(numbers);
 {% endprettify %}
 
-If your goal is to _copy_ the iterable and preserve its original element type,
-then use `List.of()`. If you don't care about the element type, or the intent is to
-get a `List` from an `Iterable` or another collection type, then use `toList()`.
+If your goal is to copy the iterable and preserve its runtime element type, or
+you don't care about the type, then use `toList()`. If your goal is to copy the
+iterable and preseve it's static element ype, use `List.of()`.
 
 
 ### DO use `whereType()` to filter a collection by type.
--- a/src/_guides/language/effective-dart/usage.md
+++ b/src/_guides/language/effective-dart/usage.md
@@ -376,8 +376,8 @@ numbers.removeAt(1); // Now it only contains integers.
 var ints = List<int>.from(numbers);
 {% endprettify %}
 
-If your goal is to _copy_ the iterable and preserve its original element type
-use `List.of()`. If you don't care about the element type, or the intent is to
+If your goal is to _copy_ the iterable and preserve its original element type,
+then use `List.of()`. If you don't care about the element type, or the intent is to
 get a `List` from an `Iterable` or another collection type, then use `toList()`.
 
 
--- a/src/_guides/language/effective-dart/usage.md
+++ b/src/_guides/language/effective-dart/usage.md
@@ -378,7 +378,7 @@ var ints = List<int>.from(numbers);
 
 If your goal is to copy the iterable and preserve its runtime element type, or
 you don't care about the type, then use `toList()`. If your goal is to copy the
-iterable and preseve it's static element ype, use `List.of()`.
+iterable and preserve its static element type, use `List.of()`.
 
 
 ### DO use `whereType()` to filter a collection by type.
--- a/src/_guides/language/effective-dart/usage.md
+++ b/src/_guides/language/effective-dart/usage.md
@@ -376,8 +376,9 @@ numbers.removeAt(1); // Now it only contains integers.
 var ints = List<int>.from(numbers);
 {% endprettify %}
 
-But if your goal is just to copy the iterable and preserve its original type, or
-you don't care about the type, then use `toList()`.
+If your goal is to _copy_ the iterable and preserve its original element type
+use `List.of()`. If you don't care about the element type, or the intent is to
+get a `List` from an `Iterable` or another collection type, then use `toList()`.
 
 
 ### DO use `whereType()` to filter a collection by type.


munificent:
I think this goes against the established idiom. In Dart today, when people want to copy a list, they typically use `toList()`. "Effective Dart" usually tries to be descriptive and not prescriptive and this is a case where I think it makes sense to follow what users do. I don't think `List.of()` adds much value over `toList()`. The latter definitely reads better at the end of a long method chain.

natebosch:
> In Dart today, when people want to copy a list, they typically use `toList()`

That's because `List.of` didn't exist for a long time. This discussion came up when we added `List.of` - the purpose was because it better expressed intent over `.toList()` which could be used for other reasons beyond copying.

> "Effective Dart" usually tries to be descriptive and not prescriptive and this is a case where I think it makes sense to follow what users do.

IMO it does not pay off to describe code that was written before new options are available. We're effectively being prescriptive that you should never use `List.of`.

> I don't think `List.of()` adds much value over `toList()`.

We added it for a reason... I don't like that our best practices guide would steer folks away from new features just because they are new.

lrhn:
I disagree completely with this change. 
If your goal is to copy the elements and preserve the element type, you *should* use `toList`. 
Only use `List.from` or `List.of` if you need to change the type. `List.from` allows any type, but doesn't work with inference, `List.of` only allows a super-type, but can infer the type from the context.

src/_guides/language/effective-dart/usage.md
```diff
@@ -376,8 +376,9 @@ numbers.removeAt(1); // Now it only contains integers.
 var ints = List<int>.from(numbers);
 {% endprettify %}
 
-But if your goal is just to copy the iterable and preserve its original type, or
-you don't care about the type, then use `toList()`.
+If your goal is to _copy_ the iterable and preserve its original element type,
```
munificent: `List.of()` doesn't preserve the original element type. It uses the static type, which may be different. Sometimes you want to preserve the runtime type and sometimes you want to *not* preserve the runtime type and use the static type.

I think we have basically two options:

* Provide minimal guidance (basically avoid `List.from()` with an uninferred type) and let people sort it out.

* Provide detailed guidance for all of the places where each option is best.

natebosch:
> If your goal is to copy the elements and preserve the element type, you _should_ use `toList`.

How often do we think the goal is to preserve the runtime element type instead of the static element type? If elements will be added `List.of` is safer - it will always work. If elements aren't added either should be fine - the only time that `.toList()` is necessary to use over `List.of` is if the list is going to be cast back to the original runtime element type later on - which I'd say is a non-idiomatic pattern.

> It is idiomatic for creating a new list with the same type.

This is idiomatic because it was the only option for a long time.


I'd argue that `List.of` is more understandable because it's behavior is based on what I can see - not on runtime information I may not have in front of me. It looks like I have a `List<num>` and I do, I won't accidentally end up with a `List<int>`. It also matches the intent - "Create a _new_ list with the same elements as this old one" rather than "Turn this list into a list".


What are the risks of having more users writing `List.of`?

munificent:
> How often do we think the goal is to preserve the runtime element type instead of the static element type?

I don't actually know. Even after the Dart 2.0 transition, I don't see much use of `List.of()`. In Code Search, I see 14,985 `toList()` calls and 102 `List.of()`. To me, that implies keeping the runtime type is at least not wrong if not actually superior to using the static type.

My hunch is that most lists are actually invariant, so the difference doesn't matter.

lrhn:
I also believe that in most situations,, there is no static context type to worry about.

People use `.toList()` in many situations, but only two general use-cases matter:
* After doing iterable-operations, and wanting to get the results as a list (or maybe just to force eager evaluation),
* When copying a list because they (or somebody else) may need to modify only one of the lists.

In both cases, you are usually happy to keep the type of the list because the iterable already has the expected type from, e.g., your `map` operations, or because you really want a *copy* of the list.

If you have a type expectation different from the list type, that's where you should use `cast`.

Creating a new list with a different type is reserved for situations where you have both a type discrepancy *and* need to modify the list, or where you know you will be reading the list repeatedly, and the extra overhead of the `cast` wrapper will eventually outweigh the cost of copying and type checking just once. That's not an evaluation I expect most people to be able to make (they should just use `cast` and avoid premature optimization).

The use-cases for `List.of`/`List.from` are few and far between. They are *not* a way to make a copy and preserve type. They are a way to make a copy and *change* the type.

natebosch:
What do you think of the new wording?

- It mentions `List.of` so that readers won't be confused and think that `List.of` should never be used.
- It calls out the runtime vs static type preservation.

I avoided going in to too much detail. I also left the `.toList()` callout first so that most users will tend to use it.

src/_guides/language/effective-dart/usage.md
```diff
@@ -376,8 +376,9 @@ numbers.removeAt(1); // Now it only contains integers.
 var ints = List<int>.from(numbers);
 {% endprettify %}
 
-But if your goal is just to copy the iterable and preserve its original type, or
-you don't care about the type, then use `toList()`.
+If your goal is to copy the iterable and preserve its runtime element type, or
+you don't care about the type, then use `toList()`. If your goal is to copy the
+iterable and preserve its static element type, use `List.of()`.
```
munificent: This is better, but now that I look at the whole guideline, the "two obvious ways" part is clearly out of date too since now we also have `List.of()`. This really needs more work.

lrhn: Also "preserve its static element type" doesn't mean what I think it's intended to say. I read it as making another list with the same *static* type (which both `.toList()` and `List.of` will do). To say what I think this intends to say, I'd use "... and reify its static element type, ...".
Most likely users won't understand that.

On top of that, what you reify is not always the *static type*, but rather the *static context type* if there is one. If you do:
```dart
var x = Iterable<int>.generate(...);
List<num> l = List.of(x);
```
then `List.of` will infer `num` as its type argument. 
The case described here is closer to:
```dart
Iterable<num> x = Iterable<int>.generate(...);
var l = List.of(x);
```
In this case, with no context type, `List.of` infers its type argument from the static type of its argument.

`List.of` can only convert an iterable of T to a list of a super-type of T. The only reason to do that, rather than `toList`, is because you want to *add* elements to (or modify elements of) the created list. So, the text should mention that.

Maybe:

> If you need to create a list that will be modified (added or written to), you need it to have a specific element type, not just any subtype of that. For that, you can usually use the `List.of` constructor, which will infer the runtime element type from the context or the static type its argument.

kwalrath:
Where'd we end up? Should I close this PR?

natebosch:
I think we can close for now. If someone has an idea of a way to phrase this that can be sufficient without being overwhelming we can revisit.

PR #1386, opened 2019-03-06
Add >>> to operator table; point to the spec

### LINKED ISSUE #1372: 'Operator table' omits `>>>` and should be described as non-normative
The language tour [operator table](https://www.dartlang.org/guides/language/language-tour#operators) matches the appendix of the language specification with one exception: The operator `>>>` has been added (it is in the language specification, but is not yet implemented), and it does not occur in the table. Similarly, the corresponding compound assignment operator `>>>=` is missing.

We just had an issue (https://github.com/dart-lang/sdk/issues/36055) where the language tour operator table was use as a source of truth for a question about parsing.

So I think it would be useful to mention in the section that contains the operator table (as it is in the language specification appendix) that this table is not normative, and parsing questions should be resolved by consulting the grammar rules. I guess it's enough to say the latter, if the word 'normative' does not fit in with the preferred style in the language tour. I've added labels 'bug' for the missing operator and 'enhancement' for the 'not normative' part. Hope that works. ;-)

kwalrath:
Fixes #1372 

Staged: https://kw-www-dartlang-1.firebaseapp.com/guides/language/language-tour#operators

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -1507,7 +1507,7 @@ You can override many of these operators, as described in
 
 <aside class="alert alert-info" markdown="1">
   **Note:**
-  For the latest, definitive list of operators, consult the
+  The definitive list of Dart operators is in the
   [Dart language specification][].
 </aside>
 
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -1510,9 +1510,10 @@ You can override many of these operators, as described in
 | assignment               | `=`    `*=`    `/=`    `+=`    `-=`    `&=`    `^=`    <em>etc.</em> |
 {:.table .table-striped}
 
-<aside class="alert alert-info" markdown="1">
-  **Note:**
-  The definitive list of Dart operators is in the
+<aside class="alert alert-warning" markdown="1">
+  **Warning:**
+  Operator precedence is an approximation of the behavior of a Dart parser.
+  For definitive answers, consult the grammar in the
   [Dart language specification][].
 </aside>
 
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -12,7 +12,7 @@ that you already know how to program in another language.
 To learn more about Dart's core libraries, see
 [A Tour of the Dart Libraries](/guides/libraries/library-tour).
 Whenever you want more details about a language feature,
-consult the [Dart Language Specification](/guides/language/spec).
+consult the [Dart language specification][].
 
 <div class="alert alert-info" markdown="1">
 **Tip:**
@@ -1491,7 +1491,7 @@ You can override many of these operators, as described in
 | unary prefix             | <code>-<em>expr</em></code>    <code>!<em>expr</em></code>    <code>~<em>expr</em></code>    <code>++<em>expr</em></code>    <code>--<em>expr</em></code>   |
 | multiplicative           | `*`    `/`    `%`    `~/`                      |
 | additive                 | `+`    `-`                                     |
-| shift                    | `<<`    `>>`                                   |
+| shift                    | `<<`    `>>`    `>>>`                          |
 | bitwise AND              | `&`                                            |
 | bitwise XOR              | `^`                                            |
 | bitwise OR               | `|`                                            |
@@ -1502,8 +1502,14 @@ You can override many of these operators, as described in
 | if null                  | `??`                                           |
 | conditional              | <code><em>expr1</em> ? <em>expr2</em> : <em>expr3</em></code> |
 | cascade                  | `..`                                           |
-| assignment               | `=`    `*=`    `/=`    `~/=`    `%=`    `+=`    `-=`    `<<=`    `>>=`    `&=`    `^=`    `|=`    `??=` |
+| assignment               | `=`    `*=`    `/=`    `+=`    `-=`    `&=`    `^=`    <em>etc.</em> |
 {:.table .table-striped}
+
+<aside class="alert alert-info" markdown="1">
+  **Note:**
+  For the latest, definitive list of operators, consult the
+  [Dart language specification][].
+</aside>
 
 When you use operators, you create expressions. Here are some examples
 of operator expressions:
@@ -3281,8 +3287,7 @@ Enumerated types have the following limits:
 * You can't subclass, mix in, or implement an enum.
 * You can't explicitly instantiate an enum.
 
-For more information, see the
-[Dart Language Specification](/guides/language/spec).
+For more information, see the [Dart language specification][].
 
 
 ### Adding features to a class: mixins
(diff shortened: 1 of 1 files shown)


src/_guides/language/language-tour.md
```diff
@@ -1502,9 +1502,15 @@ You can override many of these operators, as described in
...
 | cascade                  | `..`                                           |
-| assignment               | `=`    `*=`    `/=`    `~/=`    `%=`    `+=`    `-=`    `<<=`    `>>=`    `&=`    `^=`    `|=`    `??=` |
+| assignment               | `=`    `*=`    `/=`    `+=`    `-=`    `&=`    `^=`    <em>etc.</em> |
 {:.table .table-striped}
 
+<aside class="alert alert-info" markdown="1">
+  **Note:**
+  For the latest, definitive list of operators, consult the
+  [Dart language specification][].
+</aside>
```
eernstg: This is probably not so helpful: A missing operator is a simple issue (you'll just get a syntax error when using it, and it's easy to understand that this operator just can't be used in programs, yet). Moreover, `>>>` is currently in the spec, but implementation was postponed because other things were deemed more critical. So at this point checking the spec won't actually help, in the case of `>>>`; conversely, if developers encounter this kind of problem, it should be easy to clarify the situation (e.g., by pointing to a github issue which requests the implementation).

In contrast, operator precedence is much more subtle: You can have a given expression _e_ and consider what it means. Different precedence rules (say, "`|` has higher precedence than `||`; or maybe it's the other way around?") could then give rise to completely different answers, and the whole point of operator precedence is that the structure which produces those different answers is _invisible_. So the developer who is confused about precedence when writing or reading an expression needs to have a much more complex structure in mind than the developer who's confused about whether or not a given operator exists.

This means that it is not a very good idea to give developers rules about precedence that aren't known to hold. The operator precedence table in the spec is explicitly marked as 'non-normative', and we should be similarly cautious here. To illustrate that the rules aren't just trivially true, consider the following example:

```dart
A()..i = 42..j = 43;
```

If you look in the table you will see that `=` has the lowest precedence and `..` is the second lowest. If you look in the table in the specification you'll see that `=` is right-associative. So we should have the following structure on the above example:

```dart
(A()..i) = ((42..j) = 43);
```

However, that's not the structure you will get from the parser (and it certainly won't do the same thing as the actual program above, it's actually wrong in a whole lot of ways). Here is the actual structure:

```dart
(A()..i = 42)..j = 43;
```

We can't put the parentheses to show that `i = 42` and `j = 43` are subterms of the `..` term, but that's indeed how it is, and if we are to create an abstract syntax tree for `A()..i = 42..j = 43` then the abstract syntax tree for `i = 42` would be a child of one of the `..` nodes, and `j = 43` a child of the other `..` node.

The reason why the reference to the precedence table gave us a completely misleading result is that the precedence table does not completely and correctly reflect the actual grammar of Dart. In particular, `=` occurs as an operator directly in the grammar rule for `<expression>`, but it also occurs in a complex beast that's called `<cascadeSection>`.

A precedence table which is capable of correctly representing this situation _does not exist_. So there is no way we can claim that "here's the precedence table for Dart", because that statement can never be true, no matter how much we'd try to fiddle with that table.

I suppose this whole precedence thing came up (probably many years ago) because somebody asked for it, assuming that the concept would be meaningful for Dart, and then the table got added to the spec.

I'd admit that the table works in a lot of situations, so it may indeed 'be helpful' as the spec says, but in the end it is the grammar that counts.

I cannot add a review comment on a line that has no changes, so I'll put it here: At line 1543, we could add something like the following:

```
<div class="alert alert-warning" markdown="1">
**Warning:**
Operator precedence is an approximation of the behavior of a Dart
parser. For a definitive answer, please consult the grammar.
</div>
```

Finally, it again looks like `[]` should be deleted, in line 1511.

src/_guides/language/language-tour.md
```diff
@@ -3281,8 +3287,7 @@ Enumerated types have the following limits:
 * You can't subclass, mix in, or implement an enum.
 * You can't explicitly instantiate an enum.
 
-For more information, see the
-[Dart Language Specification](/guides/language/spec).
+For more information, see the [Dart language specification][].
```
eernstg: Delete `[]`?

src/_guides/language/language-tour.md
```diff
@@ -3989,7 +3994,7 @@ Also see the articles
 [Dart Language Asynchrony Support: Phase 1](/articles/language/await-async)
 and
 [Dart Language Asynchrony Support: Phase 2](/articles/language/beyond-async),
-and the [Dart language specification](/guides/language/spec).
+and the [Dart language specification][].
```
eernstg: Delete `[]`?

src/_guides/language/language-tour.md
```diff
@@ -4328,8 +4333,7 @@ your comments, see
 
 This page summarized the commonly used features in the Dart language.
 More features are being implemented, but we expect that they won’t break
-existing code. For more information, see the [Dart Language
-Specification](/guides/language/spec) and
+existing code. For more information, see the [Dart language specification][] and
```
eernstg: Delete `[]`?

eernstg:
I don't usually use the 'request changes' button, but with this particular topic (operator precedence) I think it's really important to add a sentence that explicitly informs the reader about the fact that the whole concept is incorrect.

When they have taken that in, it's OK if they use this concept, just because it's easier, to get approximately correct answers. ;-)

kwalrath:
> We must add a sentence where we mention that operator precedence is a flawed concept.

OK, I'll add something. I just don't want to get too technical/scary in here, since this is the first language doc many people see, and most people don't think of it as a spec.

> Also, unless this is a non-standard kind of markdown, I think there are several spurious occurrences of `[]`.

It's actually standard (as much as anything is in markdown). You can find it if you look for "implicit link name" in https://daringfireball.net/projects/markdown/syntax. We like it better than `[...](...)` because it saves space and lets you reuse the link. We like it better than plain `[...]` for reasons that @chalin explained to me, which made sense but I've since forgotten.

kwalrath:
Oh, and... you can verify for yourself that the syntax works by viewing the staged version (as mentioned above):

https://kw-www-dartlang-1.firebaseapp.com/guides/language/language-tour#operators

eernstg:
> OK, I'll add something.

Sounds good! I hope something like the following is concise enough to be non-scary:

```
<div class="alert alert-warning" markdown="1">
**Warning:**
Operator precedence is an approximation of the behavior of a Dart
parser. For a definitive answer, please consult the grammar.
</div>
```

> It's actually standard (as much as anything is in markdown)

OK, thanks!

eernstg:
> Oh, and... you can verify for yourself ..

Thanks, it looks fine! I actually saw those `[]` rendered when I tried it, but I didn't know how to test it in a way that would faithfully show how it's all rendered.

kwalrath:
> Sounds good! I hope something like the following is concise enough to be non-scary:
> 
> ```
> <div class="alert alert-warning" markdown="1">
> **Warning:**
> Operator precedence is an approximation of the behavior of a Dart
> parser. For a definitive answer, please consult the grammar.
> </div>
> ```

Thanks, that helps!

PR #1392, opened 2019-03-11
Fixing 3-column layout and TOC header.

isoos:
At first I wanted to copy the structure and styles from Flutter, but (a) it is really complex, and was hard to figure out the subset of layout styles needed, and (b) it is sub-optimal there too, a similar change could improve that site too.

Assets:
- `src/_assets/css/_site_overrides.scss` is from the Flutter site
- `src/_assets/css/_toc.scss` is from the Flutter site

Sidebar updates:
- using `position: fixed` all the time
- right-side border

Table of contents updates:
- using `position: fixed` and positioned on the right when the screen is large
- **becomes part of the content** when on smaller screen (same as Flutter)
- header text and button follows what we have on Flutter

### DIFF
--- a/src/_assets/css/_site_overrides.scss
+++ b/src/_assets/css/_site_overrides.scss
@@ -1,0 +1,5 @@
+// Flutter-site specific style overrides
+
+.site-toc--button__page-top { display: none; }
+
+.highlight .gp { user-select: none; }
--- a/src/_assets/css/_site_overrides.scss
+++ b/src/_assets/css/_site_overrides.scss
@@ -1,4 +1,4 @@
-// Flutter-site specific style overrides
+// Dart-site specific style overrides
 
 .site-toc--button__page-top { display: none; }
 
--- a/src/_assets/css/_toc.scss
+++ b/src/_assets/css/_toc.scss
@@ -1,0 +1,65 @@
+#site-toc--side {
+  $padding: map-get($spacers, 2);
+
+  display: none;
+  right: 0;
+
+  .nav { display: block; }
+  .nav-link {
+    padding-bottom: $padding;
+
+    color: theme-color('primary');
+    &.active {
+      color: theme-color('secondary'); // $gray-dark;
+      font-weight: 500;
+    }
+  }
+
+  > ul > li {
+    padding-bottom: $padding; // Extra space for top-level entries
+    li { padding-left: $padding; } // Indent subentries
+  }
+}
+
+.site-toc {
+  $line-height: normal;
+
+  @include media-breakpoint-up(xl) {
+    padding-bottom: $site-content-top-padding;
+    padding-top: $site-content-top-padding;
+
+    @at-root {
+      #site-toc--inline { display: none; }
+      #site-toc--side { display: block; }
+    }
+  }
+
+  .nav-item {
+    font-size: $font-size-sm;
+
+    .nav-link {
+      color: $site-color-body-light;
+      line-height: $line-height;
+      padding: 0;
+    }
+  }
+
+  &__title {
+    font-family: $site-font-family-alt;
+    font-size: $font-size-lg;
+    margin-bottom: bs-spacer(2);
+  }
+
+  &--button__page-top {
+    @include float-right;
+    font-size: $font-size-sm;
+    line-height: $line-height;
+    padding-top: 0;
+    padding-right: 0;
+
+    &::after {
+      content: 'vertical_align_top'; // 'keyboard_arrow_up';
+      font: $site-font-icon;
+    }
+  }
+}
--- a/src/_assets/css/main.scss
+++ b/src/_assets/css/main.scss
@@ -19,6 +19,8 @@
 @import 'books';
 @import 'os-tabs';
 @import 'sidebar';
+@import 'toc';
+@import 'site_overrides';
 
 // Extra variables (defined in terms of bootstrap and other defaults)
 $font-color: $body-color;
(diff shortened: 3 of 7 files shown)


kwalrath:
I've staged this: https://kw-staging-dartlang-2.firebaseapp.com

isoos:
> "Dart language overview" is highlighted in the sidenav. Also, the sidenav collapses back to just 2 levels.

Great find! It is unrelated to the current PR, I'll address it separately.

> Is there anything in particular you'd like me to look for?

In this PR the focus could be the mobile and table view, and if the side menu, TOC and the general layout behaves nicely in all cases.

isoos:
> Looking at a few pages, the only other sidenav issue I've seen is that when you scroll to the bottom, the footer overlaps the sidenav, so that you can't see anything below Testing.

Oh, that is something I've overlooked, and need to check if I can resolve it easily (and still removing the JS code that changes the sidebar height). The smaller footer will probably help to "get away" with some simpler fix to it..

mit-mit:
@isoos if it's less work for you, can we easily move to the Flutter.dev style "narrow" bottom right away. Just let me know, and I'll get you the content for what links we would want in it.

isoos:
> if it's less work for you, can we easily move to the Flutter.dev style "narrow" bottom right away. Just let me know, and I'll get you the content for what links we would want in it.

Unfortunately, unless the footer is strictly below the content part, we do need to keep the JS updates. I've tried to use `position: sticky` to come around that, but it introduced another visual issue: it left a placement gap (because `position: sticky` does leave out the component's space in the layout), and it become similarly complex to get rid of that gap.

In the end I've decided to keep most of the current JS update code, and now I'm preparing to update the PR with the appropriate changes.

isoos:
@kwalrath: PTAL. I've kept most of the JS code for now. Applied the Flutter's double-TOC structure.

src/_assets/css/_site_overrides.scss
```diff
@@ -0,0 +1,5 @@
+// Flutter-site specific style overrides
```
filiph: Shouldn't this say "Dart-site" ? Or is it overrides "from" the Flutter site?

src/_assets/css/_toc.scss
```diff
@@ -0,0 +1,65 @@
...
+}
+
+.site-toc {
+  $line-height: normal;
+
+  @include media-breakpoint-up(xl) {
+    padding-bottom: $site-content-top-padding;
+    padding-top: $site-content-top-padding;
+
+    @at-root {
```
filiph: (I'm not familiar with this macro but I trust you know what you're doing, especially if it's used elsewhere.)

src/_assets/css/_toc.scss
```diff
@@ -0,0 +1,65 @@
...
+
+  &--button__page-top {
+    @include float-right;
+    font-size: $font-size-sm;
+    line-height: $line-height;
+    padding-top: 0;
+    padding-right: 0;
+
+    &::after {
+      content: 'vertical_align_top'; // 'keyboard_arrow_up';
```
filiph: The end-of-the-line comments bother me a little. Are they documentation or deletions?

If they're documentation, I'd put them on a separate line and remove the `;`.

If they're deletions, I'd remove them completely.

No need to do this in the current PR, though.

src/_assets/css/main.scss
```diff
@@ -394,10 +395,12 @@ img {
 // #footernav {}
 
 #sidenav {
-  position: relative;
+  position: fixed;
```
filiph: I always thought `position: fixed` is dangerous and/or unsupported. But maybe that's an outdated notion. 

Please make sure, for example, that one the sidenav is `position: fixed`, you can actually scroll the contents of the sidebar when the window is too short.

src/_assets/css/main.scss
```diff
@@ -410,27 +413,18 @@ img {
...
-  &::-webkit-scrollbar {
-    height: 4px;
-    width: 4px;
-  }
-  &::-webkit-scrollbar-thumb {
-    background: rgba(0,0,0,0.26);
-  }
 }
-#toc {
+#site-toc--side {
```
filiph: A documentation comment would be helpful here — I don't know what `site-toc--side` is. Alternatively, another name.

src/_assets/css/main.scss
```diff
@@ -1074,18 +1068,11 @@ body.hide_toc {
       width: calc(100vw - 300px);
     }
   }
-  #toc {
+  #site-toc--side {
```
filiph: This looks like you're nesting two `#site-toc--side` rules, one in another. Is that the idea?

src/_assets/css/_toc.scss
isoos: I'm not familiar with it either, but seems to be a standard SCSS macro:
https://sass-lang.com/documentation/file.SASS_REFERENCE.html#at-root_without__and_at-root_with_

(The file is a direct copy from Flutter and will be moved to shared).

src/_assets/css/_toc.scss
isoos: I'd defer this, as the file will be moved to shared.

src/_assets/css/main.scss
isoos: Flutter uses a similar `position:fixed`-style sidebar, that's the reason I've moved this. I've checked scrolling (with Chrome on Mac).

src/_assets/css/main.scss
isoos: Adding doc, because the name is from the `_toc.scss` (from Flutter).

src/_assets/css/main.scss
isoos: The `#toc` id is replaced by the `#site-toc--side` id, there is no nesting here. There are `li.toc-entry` entries later in the rules, but those are for the internal structure of the table of contents.

PR #1394, opened 2019-03-12
Fix: sidebar URL match.

isoos:
Fixes the "Dart language overview" issue described in https://github.com/dart-lang/site-www/pull/1392#pullrequestreview-212969974.


kwalrath:
Oh, that was an easy fix. :) I'll wait to see whether the restarted build works before approving this. (I think the breakage was flakage.)

PR #1396, opened 2019-03-13
Expanding search box in the top navigation.

isoos:
This is to replicate the search box behaviour on the Flutter site:
- search box is small by default
- clicking on it (focusing) expands it
- unfocus makes it small again

`src/_assets/css/_header.scss` is copy from the Flutter site, but it needed a few overrides, as there are a few differences in the top navigation menu. 

A few changes that will be addressed in subsequent PRs:
- top nav differences
- the copied variables need to be merged with the current ones
- sidebar's search box


src/_assets/css/_header.scss
```diff
@@ -0,0 +1,217 @@
+.site-header {
```
filiph: I suggest starting this file with something like 

```css
// Taken from http://github.com/path/to/original/file.scss. Some dartlang-specific overrides applies.
```

Another way would be to put this into `site-shared` and do the overrides, but maybe that's over-engineering things.

isoos: I'll move this (and the _toc.scss) to the the `site-shared` repo, but first I wanted to make sure we have a working version here. As the file is an exact copy of the one in Flutter, I'd rather keep it this way (the overrides happen in main.scss).

PR #1398, opened 2019-03-14
Full-sized search box in sidenav's mobile view.

isoos:
- Moving `#sidenav` outside of `#page-content`, because otherwise clicking on the search field automatically closed it.
- Made the `#sidenav` wider, same as Flutter site, but the trigger when it is enabled for mobile view still differs (TBD in a subsequent PR).
- Removed branding logo from sidenav's mobile view (Flutter has none either).
- Removed search page link from sidebar, instead using the same full-sized search field as Flutter site.
- Added extra styles to remove the search field's outline.
- Added styles to compensate the differences of the to-be-shared `_header.scss`.

### DIFF
--- a/src/_assets/css/_overwrites.scss
+++ b/src/_assets/css/_overwrites.scss
@@ -20,17 +20,6 @@ a code {
 	color: #fff;
 }
 
-// SIDE NAV
-
-.sidenav__search {
-	margin-bottom: 20px;
-	display: block;
-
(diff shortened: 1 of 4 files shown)


isoos:
> Is removing the search field from the mobile view's top nav something for a future PR? (I know there's other top nav work to be done, so that would make sense.)

Yes, I wanted to do that separately, as it may involve changes in the top menu markup too.

PR #1401, opened 2019-03-15
Move top navigation closer to the structure Flutter has.

isoos:
`page_url_path` is slightly different than in the Flutter templates: instead of `doing url == 'x' or url contains 'x/'`, I'm prepending and appending the url string, making the checks simpler (and more reliable).

I haven't replicated the exact structure as the Flutter site has, because we still have dropdowns, and it would be easier to replicate the remaining structure and styles after those are removed.

### DIFF
--- a/src/_assets/css/main.scss
+++ b/src/_assets/css/main.scss
@@ -280,7 +280,6 @@ img {
 }
 #mainnav {
   display: flex;
-  color: $gray-base;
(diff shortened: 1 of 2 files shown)


kwalrath:
This build failed for the same reason the other PR's build failed. I don't think we have to fix it to commit either change, but we should fix it soon. Let me know if you want me to do that.

PR #1402, opened 2019-03-15
Fix sidebar.

isoos:
https://github.com/dart-lang/site-www/pull/1398 broke it, and it took me a while to figure out what was missing. Without the z-index the main content gets priority on scroll and click events.

PR #1403, opened 2019-03-15
Fix dev build by updating constant set formatting

kwalrath:
I wonder if this will break the stable build. One way to tell...

Also updated site-shared.


kwalrath:
A dartfmt change caused our dev builds to break (#1401, #1402, #1403). I think it might be due to https://github.com/dart-lang/sdk/commit/ab132a8aa36594b01b16083b012861dd62ec11fc.

kwalrath:
@isoos this change fixes the build, woohoo!

PR #1409, opened 2019-03-19
Fix obsolete banner

isoos:
The change moves the `.site-header` style to the top element of the header, and it makes it `position: sticky`. With that, the `fixed` style can be removed, and it becomes more consistent with the Flutter site.

With the fixed banner for obsolete page, the JS needs to get updated to calculate the top position, and it also needs to fix the sidebar in mobile view.

### DIFF
--- a/src/_assets/css/main.scss
+++ b/src/_assets/css/main.scss
@@ -184,19 +184,14 @@ img {
 // Structure
 
 #page-header {
-  position: fixed;
-  width: 100%;
-  height: $site-header-height;
-  top: 0;
-  left: 0;
-  right: 0;
-  background-color: $gray-light;
   box-shadow: 0 3px 5px rgba(0,0,0,0.1);
-  z-index: 1030;
+
+  // override of shared/_header.scss
+  .site-header {
+    box-shadow: none;
+  }
 }
 #page-content {
-  position: relative;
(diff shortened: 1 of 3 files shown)


kwalrath:
Checking now... I've staged the site at https://kw-staging-dartlang-2.firebaseapp.com.

isoos:
One example for obsolete banner:
https://kw-staging-dartlang-2.firebaseapp.com/articles/libraries/zones

kwalrath:
Not new to this PR, but there's an issue in the transition from sidenav to no-sidenav. To reproduce:

1. Go to any page, e.g. https://kw-staging-dartlang-2.firebaseapp.com/guides/get-started (or https://dartlang-org-staging-0.firebaseapp.com/guides/get-started). The window can be any height.

2. Make the window wider if necessary, so that it shows the sidenav on the left. No issues.

3. Make the window narrower. The sidenav gets really big, just before animating down to nothing.

This issue isn't present on dartlang.org.

Should I file a bug?

isoos:
I can do it as next the next task, no need for filing a bug.

isoos:
On a second thought, maybe a tracking issue would be worth it, because the effect is the same on the Flutter site, and if we want to get rid of it, it may require more work.

PR #1416, opened 2019-03-21
Adjust global font sizes.

isoos:
Because the base font size is decreasing, I'm changing some of the derived sizes to be the same as before:
- `$font-size-base * 0.8` -> `$font-size-base * 0.91`
- `$font-size-base * 0.7` -> `$font-size-base * 0.8`

### DIFF
--- a/src/_assets/css/_overwrites.scss
+++ b/src/_assets/css/_overwrites.scss
@@ -55,12 +55,12 @@ body.homepage #code-sample {
 body.homepage {
 	#code-sample-section {
 		pre {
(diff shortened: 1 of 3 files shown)


kwalrath:
Even though I'm running `git pr` from the `dash` branch, I don't see the latest dash changes in there, so I can't judge the appearance of this PR. It seems good in theory...

Can you incorporate the latest dash changes?

isoos:
> Can you incorporate the latest dash changes?

Done. PTAL.

PR #1417, opened 2019-03-21
Update right-side TOC styles.

isoos:
The styles were removed (and added) with a lot of exploration, and it was also with #1416 applied (but submitted separately). It reverts the highlight color to the same one Flutter uses, and applies the font settings from the pdf spec.


isoos:
> Looks good, except we need to add some padding to 2nd-level TOC entries.

Ah, I missed those. Updated the PR with extra padding, PTAL.

PR #1420, opened 2019-03-22
Apply text colors

isoos:
- `$brand-primary-dark` is no longer used.
- Added `Roboto Mono` to `$font-family-monospace`.
- Applied font colors. I think on my monitor the `code`'s `006666` is a bit different color than it was on the screenshot, but going with it, as it was specified in the pdf.

### DIFF
--- a/src/_assets/css/_overwrites.scss
+++ b/src/_assets/css/_overwrites.scss
@@ -5,11 +5,6 @@ dd {
 	margin-bottom: 15px;
 }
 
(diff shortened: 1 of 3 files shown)


kwalrath:
I agree, #006666 is nowhere near the color in the screenshot. (As a side note... Chrome's tools for playing around with text color — with the extra detail for contrast ratio — are pretty great!) Unfortunately, I couldn't find anything near the color that was both distinguishable from the black and had a good contrast ratio.

So let's stick with these colors, for now.

PR #1421, opened 2019-03-22
Fix wide content without ToC

isoos:
The width calculation of `width: calc(100vw - 300px);` is not required, and I moved the hiding logic to a more natural place.

### DIFF
--- a/src/_assets/css/main.scss
+++ b/src/_assets/css/main.scss
@@ -431,6 +431,10 @@ img {
     }
   }
   ul.section-nav.nav > li > .nav-link:before { margin-left: -0.88rem; }
+
+  body.hide_toc & {
+    display: none;
+  }
 }
 
 #site-search-results {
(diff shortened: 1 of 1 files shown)


kwalrath:
I'm confused... Don't we want the content to be as wide as possible? Compare these three:

* https://www.dartlang.org/guides/get-started (so wide. but that's what the reader asked for, right?)
* https://dartlang-org-staging-0.firebaseapp.com/guides/get-started (dash before these changes; also wide)
* https://dartlang-org-staging-1.firebaseapp.com/guides/get-started (dash WITH these changes; seems to reserve space for the TOC)

isoos:
@mit-mit linked to https://flutter.dev/docs/development/ui/navigation
the difference is that the github icons (and the content) do not go to the right of the page.

mit-mit:
@kwalrath going "as wide as possible" causes a very jarring layout transition when moving form a page that has a right-nav and one that doesn't. Try clicking between 'Dart language overview' and 'Tour' in each of the two staging sites you have, and you should see it.

PR #1426, opened 2019-03-28
fix some mismatch example

xpwmaosldk:
DO capitalize acronyms and abbreviations longer than two letters like words example mismatch


xpwmaosldk:
Contributor License Agreement signed

mit-mit:
cc @munificent

PR #1430, opened 2019-04-03
Increase z-index of search icon.

isoos:


src/_assets/css/main.scss
```diff
@@ -491,7 +491,7 @@ img {
 .site-header__search {
   &::before {
     color: $gray;
-    z-index: 1;
+    z-index: $site-z-header + 1;
```
kwalrath: This doesn't fix the issue, but when I use the inspector to unset the "absolute" position, _then_ it shows up!

kwalrath: Same thing happens on flutter.dev. So it isn't a z-index issue, it's a positioning issue.

kwalrath:
The search icon is back, so I'm going to close this.

PR #1435, opened 2019-04-05
Full-width banner.

isoos:
- needed to move
- remove extra classnames on the HTML element
- updated CSS styles to follow design

### DIFF
--- a/src/_assets/css/main.scss
+++ b/src/_assets/css/main.scss
@@ -262,13 +262,14 @@ img {
 
 // Banner
 .banner {
-  background-color: $blue-light;
+  background-color: #e7f8ff;
+  padding: 20px 0;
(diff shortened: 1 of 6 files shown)


isoos:
PTAL, I've updated both the shadow and the homepage.

isoos:
@kwalrath: I've found a significant difference between this solution and the Flutter page: there, the banner scrolls out of the view, while here the banner stays in the view. Should I re-do this to scroll it out of the view?

isoos:
@kwalrath: Don't mind my [previous](https://github.com/dart-lang/site-www/pull/1435#issuecomment-480799507) comment, I've updated the PR with a scrolling banner. It requires an update to the JS, which is now more similar to how Flutter's site works.

PR #1440, opened 2019-04-08
Update card bg, color, border and font.

isoos:


isoos:
btw. the design on gallery had a font color of `#f8f9fa`, while in the doc it is mentioned as `white`. Which should I use?

kwalrath:
Taking a look now... It looks like the build failed due to a dartfmt change in the dev branch. We can ignore that, for now.

src/_assets/css/main.scss
```diff
@@ -751,15 +751,17 @@ main .content {
   margin: 15px;
   overflow: hidden;
   box-sizing: border-box;
-  background-color: #29b6f6;
+  background-color: #294058;
+  border-radius: 0;
   box-shadow: 0px 1px 3px 0px rgba(0, 0, 0, 0.21);
-  color: $site-color-white;
+  color: #f8f9fa;
```
kwalrath: Why not use a variable here?

src/_assets/css/main.scss
```diff
@@ -751,15 +751,17 @@ main .content {
...
 
   h1, h2, h3, h4, h5, h6 {
     font-family: $site-font-family-gsans;
     margin-top: 0;
     a {
-      color: $site-color-white;
+      color: #f8f9fa;
```
kwalrath: and here...

isoos:
> I'm not sure why the values are literals instead of variables

Can't really decide when to introduce a new variable and when to use the inline literal. Updated the code to use a variable.

PR #1448, opened 2019-04-09
Collapsible table of contents.

isoos:
I wanted to have the max-height transitioned and the content being unrolled, but it seems it can't be done with sanity. We could probably do a JS-based animated expand and collapse, but I would postpone that for now.

I also added some extra styles (opacity and background) in order to mark the area different from the body content. I'm not sure what looked better, nor about the icons, so let's iterate on it :)

### DIFF
--- a/src/_assets/css/main.scss
+++ b/src/_assets/css/main.scss
@@ -431,6 +431,46 @@ img {
   }
 }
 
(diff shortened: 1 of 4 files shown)


kwalrath:
Looking now...

PR #1454, opened 2019-04-11
Initial sidebar positions.

### LINKED ISSUE #1453: Shifting menu
In the `dash` branch, I see a weird shifting of the leftnav when I click menu entries:

This reproes for me on MacOS using Chrome 73. I doesn't seem to repro when using Safari.

glitch

isoos:
Fixes #1453.

I've tried to stop the jumps via JavaScript, e.g. to start the fix of top position earlier in the page load, but couldn't make it work. Instead, I'm using these pre-calculated values and it seems like this fixes the issue for now.

### DIFF
--- a/src/_assets/css/_variables.scss
+++ b/src/_assets/css/_variables.scss
@@ -96,6 +96,9 @@ $site-font-icon: 24px/1 $site-font-family-icon;
 
 // Layout
 $site-header-height: 50px;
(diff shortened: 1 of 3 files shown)


kwalrath:
Staged: https://kw-staging-dartlang-2.firebaseapp.com/guides/language/language-tour

PR #1455, opened 2019-04-12
Fix some of the external link icons

isoos:
Addresses the [initially large icons](https://github.com/dart-lang/site-www/issues/1453#issuecomment-482305030) on the sidebar, and also the shifting position of the banner's text. Unfortunately I couldn't find a good way to replace the JS code that extends links with the `external` style, but luckily, updating the banner template solves the most prominent of them.


kwalrath:
Looking now...

PR #1464, opened 2019-04-17
Create top-level page style

kwalrath:
~~Issue: The TOC disappears if the page is very wide. I'd like it to always be inline and visible unless the page has `toc: false` in the YAML at the top.~~

Unless the page has `toc: false` in the YAML at the top, the TOC is always in the page, rather than on the right. We might want to take steps to reduce the max page width.


kwalrath:
Staged:
https://kw-www-dartlang-1.firebaseapp.com/guides/platforms
https://kw-www-dartlang-1.firebaseapp.com/community

legalcodes:
LGTM

PR #1478, opened 2019-04-24
Add /tools/dart-devtools

kwalrath:
Also removed the placeholder for an overview page that doesn't seem necessary.

### DIFF
--- a/src/tools/dart-devtools.md
+++ b/src/tools/dart-devtools.md
@@ -5,11 +5,11 @@ toc: false
 ---
 
 Dart DevTools is a suite of performance tools for Dart and Flutter.
-DevTools supports source-level debugging of Dart command-line apps,
+In addition to Flutter-specific features,
+DevTools supports source-level debugging of Dart apps,
 as well as viewing general log and diagnostics information about a running app.
-
-We expect developers to use DevTools as part of their existing
-IDE or command-line based development workflow.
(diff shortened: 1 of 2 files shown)


src/tools/dart-devtools.md
```diff
@@ -0,0 +1,16 @@
+---
+title: Dart DevTools
+description: A suite of performance tools for Dart and Flutter.
+toc: false
+---
+
+Dart DevTools is a suite of performance tools for Dart and Flutter.
+DevTools supports source-level debugging of Dart command-line apps,
```
mit-mit: There is nothing command-line specific about DevTools. You can use it for Flutter apps, command-line apps, server apps, and web apps. I suggest we just say 'Dart apps'.

src/tools/dart-devtools.md
```diff
@@ -0,0 +1,16 @@
...
+as well as viewing general log and diagnostics information about a running app.
+
+We expect developers to use DevTools as part of their existing
```
mit-mit: It's not quite clear to me what this sentence adds. Should I read is as saying that DevTools works for both IDE and command-line based developer workflows?

kwalrath: I took it from the ddt page, and understood it as meaning "this isn't an IDE". But your wording is better.

kwalrath:
ptal

PR #1479, opened 2019-04-24
Add content for /web

kwalrath:
Staged: https://kw-staging-dartlang-2.firebaseapp.com/web

It's minimalistic, but at least it has a pretty picture.

Can we update the text about how Google uses Dart for web apps?


src/web/index.md
```diff
@@ -4,4 +4,14 @@ description: Resources for developing Dart web apps.
 toc: false
 ---
 
-[TBD: Write this page.]
+
+Google builds many critical web apps using the Dart language, often with
+[AngularDart.]({{site.angulardart}})
+The Google Ads front end is built in Dart.
```
mit-mit: Consider changing 'Ads front end' to 'Google AdWords' and link to https://news.dartlang.org/2016/03/the-new-adwords-ui-uses-dart-we-asked.html ?

kwalrath: I think they changed the name from Google AdWords to Google Ads. (https://en.wikipedia.org/wiki/Google_Ads says it happened in 2018.) I'll try to write some text that links to it in a clear way.

PR #1480, opened 2019-04-24
Update samples heading

mit-mit:
Update heading to be consistent with the destination it links to


mit-mit:
OK, I don't feel strongly about it

PR #1482, opened 2019-04-24
Layout updates: content, banner and sidebar padding, positions

isoos:
- Applied the new 10% padding design.
- Increased sidebar and TOC width and paddings.
- Decreased sidebar and TOC first-level heading font size.
- Decreased banner top and bottom padding.
- Decreased code block left and right padding.
- Fixed the toggle button position on the leftnav (it was not always on the top).


kwalrath:
I notice a layout issue with http://localhost:4000/codelabs/dart-cheatsheet#code-example that isn't there in the staging-0 version of the site. Could you look into this, @isoos?

I can stage this version somewhere, if that would be helpful.

/cc @RedBrogdon

isoos:
> I notice a layout issue with http://localhost:4000/codelabs/dart-cheatsheet#code-example that isn't there in the staging-0 version of the site. Could you look into this, @isoos?

@kwalrath what's the issue you see?

Screen Shot 2019-04-24 at 21 39 53

It took a few seconds for the content to load, but otherwise the layout seems to be ok.

kwalrath:
The right border (line) of the dartpad is visible initially, but once the UI fills in, the right line goes away.

image

kwalrath:
Staged to https://kw-www-dartlang-1.firebaseapp.com/codelabs/dart-cheatsheet#code-example.
This issue is visible only at a particular width, which seems to be just before the TOC appears on the right. If you drag to change the window width, you might see the issue.

image

isoos:
As I resize the window, I can see it disappearing, and then reappearing. Are you using Chrome Canary? Maybe the reappearing is broken in Canary?

Dartpad is embedded through an iframe, which has a fix 1px border. I don't see anything on our side we could add to that to make sure there is always a border...

kwalrath:
I saw it in both Canary and Chrome stable.

legalcodes:
I'm seeing this as well -- the right right side border blinks in and out as I change the window size. If I stop resizing the window at just the right point, the right side border vanishes and remains that way until I resize the window again. 

Screen Shot 2019-04-24 at 3 01 21 PM

kwalrath:
On second thought, I think we still need to use more of the horizontal room. Compare:

* [this PR's layout](https://kw-www-dartlang-1.firebaseapp.com/codelabs/dart-cheatsheet#code-example-3) (line 7 has to wrap unless the window's very wide; seems like too much space to the left of the TOC)
* [layout before this PR](https://dartlang-org-staging-0.firebaseapp.com/codelabs/dart-cheatsheet#code-example-3) (not much different)
* [current site's layout](https://www.dartlang.org/codelabs/dart-cheatsheet#code-example-3)

isoos:
I'm sorry, I've screwed up a rebase, and pushed here... apparently my git-fu is not enough to recover from it. I'll re-do the PR.

kwalrath:
Glad I'm not the only one this has happened to... :)

PR #1484, opened 2019-04-24
Update inline table-of-contents.

isoos:


isoos:
The design did not specify it entirely, so I've re-styled the multi-level contents to go without list item dots, and I've applied the grey background for both the expanded and the too-short case.

kwalrath:
Staged: https://kw-www-dartlang-1.firebaseapp.com/guides/language/language-tour

Looks great to me! We might want to tweak the amount of text visible by default, and maybe add ..., but that's something we can do later.

PR #1485, opened 2019-04-24
Underline the active top header item.

isoos:


kwalrath:
When I click Platforms, Docs is selected as well, maybe because both paths start with /guides. I've been thinking about moving Docs to /docs (to match flutter), so I think I'll accept this change and see if that fixes the problem.

isoos:
That part is handled via https://github.com/dart-lang/site-www/blob/dash/src/_includes/navigation-main.html#L11-L15

`/*/guides/platforms/` and `/*/guides/` share the common prefix, so that needs to be fixed, either way (the template change shouldn't be too much in either case).

PR #1491, opened 2019-04-25
Layout updates: content, banner and sidebar padding, positions

isoos:
- re-done #1482 with the correct rebase
- Applied the new padding design.
- Increased sidebar and TOC width and paddings.
- Decreased sidebar and TOC first-level heading font size.
- Decreased banner top and bottom padding.
- Decreased code block left and right padding.
- Fixed the toggle button position on the leftnav (it was not always on the top).

### DIFF
--- a/src/_assets/css/main.scss
+++ b/src/_assets/css/main.scss
@@ -245,16 +245,16 @@ img {
 
   > article {
     > .content {
-      padding-left: 8%;
+      padding-left: 5%;
       padding-right: 8%;
     }
 
     @media(min-width: 1024px) {
-      margin-left: 246px;
+      margin-left: 236px;
     }
 
     @media(min-width: 1200px) {
-      margin-right: 246px;
+      margin-right: 236px;
     }
   }
 }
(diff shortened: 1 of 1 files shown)


isoos:
@kwalrath @legalcodes: PTAL

I've reduced the padding to 80% of the design (e.g. 50px -> 40px, 10%->8%). I've tried 60%, it felt a bit more packed, but could still work if needed.

kwalrath:
Staged at https://kw-www-dartlang-1.firebaseapp.com. I still see a single column in https://kw-www-dartlang-1.firebaseapp.com/guides sometimes (might be specific to its layout).

isoos:
PTAL.

kwalrath:
There's still at least one width (hard to catch!) when the boxes are in a single column:
image

Can you fix that?

isoos:
> Can you fix that?

PTAL. I've removed another 10-10px from the sides. I can't reproduce the issue, but if it still persist, we can reduce the right-side 8% to 7% and that should probably be enough.

kwalrath:
I can't reproduce the issue in your newest version, woohoo! Giving it a more thorough look now...

PR #1460, opened 2019-04-15
Platforms

mit-mit:
* Add Platforms node
* Move Tools to top-level
* Remove Development node

PR #1492, opened 2019-04-25
Add web deployment page

mit-mit:
Doesn't look like you copied over this from webdev?

I had this in https://github.com/dart-lang/site-www/pull/1460; this ports it separately here.

### DIFF
--- a/src/web/deployment.md
+++ b/src/web/deployment.md
@@ -1,11 +1,10 @@
 ---
-title: "Web Production Deployment"
-short-title: "Production deployment"
-description: "Learn how to build your Dart Web app for production deployment"
+title: "Web deployment"
+description: "Learn how to build your Dart web app for production deployment"
 toc: true
 ---
 
-Deploying an Dart Web app is similar to deploying any other web app:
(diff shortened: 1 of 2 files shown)


kwalrath:
Yikes, that's scary that I missed this page. I wonder how...

src/web/deployment.md
```diff
@@ -0,0 +1,174 @@
+---
+title: "Web Production Deployment"
```
kwalrath: This is very different from the name in the sidenav (and also we use sentence case for titles now). How about we call it "Building for web deployment". If we need a short-title (which is probably good, tab titles being so limited), it could be "Web deployment".

Globally, Web -> web. (https://developers.google.com/style/word-list#letter-W)

src/web/deployment.md
```diff
@@ -0,0 +1,174 @@
...
+short-title: "Production deployment"
+description: "Learn how to build your Dart Web app for production deployment"
+toc: true
+---
+
+Deploying an Dart Web app is similar to deploying any other web app:
```
kwalrath: an Dart Web app -> a Dart web app

PR #1493, opened 2019-04-25
Create separate JS interop page

mit-mit:
Porting this change from https://github.com/dart-lang/site-www/pull/1460


src/web/js-interop.md
```diff
@@ -0,0 +1,23 @@
+---
+title: "JS/TS interop"
+short-title: JS-interop
```
kwalrath: How about we do something like this:

```
title: Interoperating with JavaScript and TypeScript code
short-title: "JS/TS interop"

File: src/web/js-interop.md [hunk 2]
```diff
@@ -0,0 +1,23 @@
...
+description: "The Dart Web platform supports JavaScript/TypeScrupt integration via package:js"
```
kwalrath: Ah, I see why you were capitalizing "Web" before. I'm still leery of it, since it makes us look like we wrote this in 1995. How about:

description: "Use package:js to integrate JavaScript or TypeScript code into your Dart web app."

src/_data/side-nav.yml
```diff
@@ -96,6 +96,8 @@
           permalink: /web
         - title: Get started
           permalink: /web/get-started
+        - title: JS/TS interop
```
kwalrath: I'm a little leery of using an abbreviation like "interop" in the sidenav. (Web results mostly link to a particular conference.) Let's do "interoperation" unless you have a better word.

kwalrath:
We should coordinate our terminology with whatever flutter.dev uses.

src/_data/side-nav.yml
mit-mit: Isn't that the term most use? First many hits here are exactly about calling JS from language-x: https://www.google.com/search?q=js+interop

kwalrath:
Looking at flutter.dev, I see https://flutter.dev/docs/development/platform-integration/platform-channels doesn't use the string "interop" _anywhere_. It does mention _integration_ (barely, in the nav) but mostly uses _platform-specific API_.

@sfshaza2, do you have any guidance for us?

mit-mit:
Actually, what Flutter has is very different: Interop in Dart is a language-to-language calling mechanism. Flutter has a send-event-to-other-language mechanism. That's why we use a different term there.

But we could also consider using 'Calling JS/TS from Dart', similar to https://kotlinlang.org/docs/reference/js-interop.html

[note though how they still have js-interop in their page name!]

kwalrath:
> Actually, what Flutter has is very different: Interop in Dart is a language-to-language calling mechanism. Flutter has a send-event-to-other-language mechanism. That's why we use a different term there.

OK, that's defensible. :)

> But we could also consider using 'Calling JS/TS from Dart', similar to https://kotlinlang.org/docs/reference/js-interop.html

Do we cover the other direction — calling Dart from JS/TS?

mit-mit:
> Do we cover the other direction — calling Dart from JS/TS?

Not sure; @kevmoo ?

kevmoo:
> > Do we cover the other direction — calling Dart from JS/TS?
> 
> Not sure; @kevmoo ?

You can call Dart from JS, but it's a lot more involved. Not sure if we have a solid story here. @vsmenon @jmesserly @sigmundch

sfshaza2:
I can see I missed the excitement but, yeah, we don't do interop, exactly.

src/_data/side-nav.yml
kwalrath: Interestingly, there's https://developer.android.com/kotlin/interop, which has **interop** in the title but **interoperability** in the page. Let's follow their lead.

mit-mit:
@kwalrath anything else here?

PR #1494, opened 2019-04-25
Add mobile apps

mit-mit:
Now that we don't have https://www.dartlang.org/guides/platforms anymore, and to make sure Development represents all types of apps you can create with Dart


mit-mit:
This probably needs some copy-editing; feel free to push those directly to the PR.

PR #1495, opened 2019-04-25
Top-menu nav items in the mobile view of the side-menu.

isoos:
The font size follows Flutter in that these menu items are slightly larger than the otherwise top-level items (they are 20px, and the top-level items will be 16px). However, the active styling is up to discussion: should we lighten the non-active ones? Should we make the active one blue?

### DIFF
--- a/src/_assets/css/main.scss
+++ b/src/_assets/css/main.scss
@@ -498,6 +498,25 @@ img {
         font-weight: 400;
       }
     }
+
+    .navbar-nav {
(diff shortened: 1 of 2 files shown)


kwalrath:
I agree that these "section" menu items should be larger than the top-level Docs menu items (more like Flutter, in other words). It looks like Flutter bolds the active section item, which is subtle but fine by me. The current page is blue on both, which lgtm.

PR #1500, opened 2019-04-25
Platforms overview

mit-mit:


mit-mit:
This is failing in travis, as it assumes that https://github.com/dart-lang/site-www/pull/1492 has landed

```
http://localhost:4000/platforms
- (228:10) 'deployme..' => http://localhost:4000/web/deployment (HTTP 404)
```

mit-mit:
@kwalrath I added the link to Flutter as requested

kwalrath:
> @kwalrath I added the link to Flutter as requested

Cool, thanks. I've updated the staging server. I believe @messi01 or I had some other comments regarding "platform" and Dart as a noun. Are you going to address those, or do you want me to give it a shot?

mit-mit:
I don't believe I've seen that feedback; there don't seem to be any outstanding review comments here?

kwalrath:
> I don't believe I've seen that feedback; there don't seem to be any outstanding review comments here?

I believe it was in email...

kwalrath:
I think I'll just do a copy edit, and then we can discuss the content more, if need be.

kwalrath:
@mit-mit & @messi01 ptal. I've updated the staged version: http://go/dart.dev2/platforms

mit-mit:
> I think I'll just do a copy edit, and then we can discuss the content more, if need be.


Made just a few tweaks to that in https://github.com/dart-lang/site-www/pull/1500/commits/86d3a2ecb4b6daaa1e9aab42a91262e59ae652ec

src/platforms.md
```diff
@@ -1,41 +1,41 @@
 ---
 title: Platforms
-description: "You can use the Dart language to write mobile apps, web apps, standalone apps, servers, and more."
+description: "You can use the Dart SDK to write mobile apps, web apps, standalone apps, servers, and more."
```
kwalrath: You use the Flutter SDK, not the Dart SDK, to write mobile apps.

src/platforms.md
```diff
@@ -1,41 +1,41 @@
...
 toc: true
 ---
 
-Using the Dart programming language and libraries, you can write simple scripts or
-full-featured apps. Whether you're creating a mobile app, web app, command-line
-script, or server-side app, there's a Dart solution for that.
+You can use Dart to write simple scripts or full-featured apps. Whether you're
```
kwalrath: I like these words, but we're not supposed to use _Dart_ as a noun. :/

src/platforms.md
mit-mit: Yeah, good point.

I don't like 'Dart language' -- that's to narrow (it's all the surrounding stuff that enables the mobile, web, server, ..., part.

Can we just say 'You can use Dart to write...' ?

src/platforms.md
mit-mit: Oh, same comment above. What can we use then that captures "all the things" we provide, as opposed to just the language?

src/platforms.md
kwalrath: Still a noun. :(

mit-mit: Changing to just Dart as per verbal discussion

src/platforms.md
mit-mit: Resolving as per verbal discussion

mit-mit:
@kwalrath can we merge this now?

PR #1503, opened 2019-04-26
Resize h3, h4, h5

isoos:
As stated in the doc comments: we rarely use h4, and never use h5, h6.


isoos:
Updated the h4, h5, h6 to 15px, as discussed in the document.

PR #1505, opened 2019-04-26
Update web entries in core lib table + reformat

mit-mit:


kwalrath:
Staged: https://sz-www-1.firebaseapp.com/guides/libraries

mit-mit:
The link to 'Dart platforms' just above the table was meant to serve as the legend.

kwalrath:
Trying to come up with descriptions for the annotations (which I was trying to make as short as possible), I realized I was confused about Flutter & JIT. Here's what I have, so far:

* **JIT:** The Dart VM in just-in-time compilation mode. Includes Flutter mobile, command-line, and server apps, unless otherwise noted.
* **AOT:** Dart code compiled to machine code. Includes AOT compiled command-line and server apps, as well as deployable Flutter mobile apps.
* **Web:** Dart code compiled to JavaScript. Supported by both compilers — dartdevc (development time) and dart2js (deployment) — unless otherwise noted.
* **All:** All of the above.

mit-mit:
I think those descriptions should be on the /platforms page; i.e. into https://github.com/dart-lang/site-www/pull/1500

kwalrath:
> I think those descriptions should be on the /platforms page; i.e. into #1500

It's really inconvenient for people to have to flip back and forth between pages. I think this page needs to be self-contained, although of course it can refer to the platform page for people who want more information.

mit-mit:
We're going to have to refer to these terms from a bunch of pages, and I don't think it a good idea to duplicate the definitions across all of those.

kwalrath:
> We're going to have to refer to these terms from a bunch of pages, and I don't think it a good idea to duplicate the definitions across all of those.

Then we should use includes!

mit-mit:
Sure; I'd like to have the discussion of the phrasing for this over in the /platforms PR https://github.com/dart-lang/site-www/pull/1500 though, as that is the main page that describes out terminology here. We can then come back and update the present page later.

For now I'd like to land this PR that is a factual update to capture which libraries work where.

kwalrath:
I've created #1509 so you can see what I have in mind (without mashing my stuff on top of your PR).

kwalrath:
I applied my changes from #1509, minus the legend.

I also changed the title to just "Core libraries", since I don't think we need "Overview" in the title if the page title matches the category. (E.g. tools/ would be Tools, even if it's Tools > Overview in the sidenav.)

PR #1507, opened 2019-04-26
Fix jitter by adjusting the pre-calculated banner height

isoos:
The layout was updated by removing 8-8px from the top and bottom of the banner, decreasing the pre-calculated height by 16px.


kwalrath:
What should I do to see what changed?

isoos:
> What should I do to see what changed?

On frequent reload the left-side sidebar don't jumps right at the page load time.

kwalrath:
Hmmm... that's what I tried. Tried emptying the cache, using incognito mode, etc. But if you say it works, I trust you.

PR #1509, opened 2019-04-26
CE, clarify platforms, alphabetize libraries

kwalrath:
Suggested changes to #1505.

### DIFF
--- a/src/_guides/libraries/index.md
+++ b/src/_guides/libraries/index.md
@@ -5,37 +5,41 @@ short-title: "Core libraries"
 toc: false
 ---
 
-Dart has a rich set of core libraries that provides essentials for many everyday
+Dart has a rich set of core libraries that provide essentials for many everyday
 programming tasks such as working on collections of objects,
 (`dart:collection`), making calculations (`dart:math`), and encoding/decoding
-data (`dart:convert`).
+data (`dart:convert`). Additional APIs are available in
+[community contributed packages](/guides/libraries/useful-libraries).
 
-The following table shows an inventory of all Dart core libraries. Aadditional
-APIs are available in community contributed packages located [in shared
-packages](/guides/libraries/useful-libraries). Also note that some core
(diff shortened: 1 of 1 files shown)


src/_guides/libraries/index.md
```diff
@@ -5,37 +5,41 @@ short-title: "Core libraries"
...
 (`dart:collection`), making calculations (`dart:math`), and encoding/decoding
-data (`dart:convert`).
+data (`dart:convert`). Additional APIs are available in
+[community contributed packages](/guides/libraries/useful-libraries).
 
-The following table shows an inventory of all Dart core libraries. Aadditional
-APIs are available in community contributed packages located [in shared
-packages](/guides/libraries/useful-libraries). Also note that some core
-libraries are only available on some [Dart platforms](/platforms).
+The following table lists all of the Dart core libraries.
```
mit-mit: Please add back the link to packages

src/_guides/libraries/index.md
```diff
@@ -5,37 +5,41 @@ short-title: "Core libraries"
...
+Each library works on at least one [platform](/platforms):
+
+* **JIT:** The Dart VM in just-in-time compilation mode. Includes Flutter mobile, command-line, and server apps, unless otherwise noted.
```
mit-mit: Please let's discuss these over in the platforms PR; I don't like the way these are phrased. So can we please remove them for now, work on them in the platforms PR, and then add back here (possibly via an include) when they are done.

src/_guides/libraries/index.md
kwalrath: I moved it up above.

src/_guides/libraries/index.md
kwalrath: OK, will do. I'll edit #1505 to be a less invasive edit, perhaps with a TBD for the legend.

PR #1510, opened 2019-04-27
Rewrote package overview, combined two pub/package pages

kwalrath:
rewrote /tools/pub/get-started to introduce packages more thoroughly
moved /tools/pub/get-started to /guides/packages
added the content formerly in /tools/pub and /guides/packages
redirected /tools/pub to /tools/pub/cmd


src/_guides/packages.md
```diff
@@ -0,0 +1,227 @@
...
+such as git repositories.
+Wherever your packages come from, pub manages version dependencies,
+helping you get package versions that work with each other and
+with your SDK version.
+
+Most [Dart-savvy IDEs][] offer support for using pub that
+includes creating, installing, updating, and publishing packages.
+Or you can use [`pub` on the command line](/tools/pub/cmd).
+
+{% include flutter-packages.md %}
```
mit-mit: I think this Flutter "warning" stands out way too much. I would either totally remove it (those that do Flutter, are already using Flutter.dev), or move it single sentence like "If you are using Flutter, use `flutter pub <sub-command>` instead of `flutter <sub-command>`" up in the previous paragraph

src/_guides/packages.md
```diff
@@ -0,0 +1,227 @@
...
+
+At a minimum,
+a Dart package is a directory containing a [pubspec file](/tools/pub/pubspec).
+The pubspec contains some metadata about the package. Additionally,
+a package can contain dependencies (listed in the pubspec),
+Dart libraries, apps, resources, tests, images, and examples.
+
+<aside class="alert alert-info" markdown="1">
```
mit-mit: Not sure we need this alert. I would just remove it.

src/_guides/packages.md
```diff
@@ -0,0 +1,227 @@
...
+  **Note:** If your Dart app depends on one or more packages,
+  then your app itself must be in a package.
+</aside>
+
+To use a package, do the following:
+
+* Create a pubspec (a file that lists package dependencies and includes
```
mit-mit: s/pubspec/`pubspec.yaml`/

src/_guides/packages.md
```diff
@@ -0,0 +1,227 @@
...
+  other metadata, such as a name for your package).
+* Use pub to get your package's dependencies.
+* If your Dart code depends on a library in the package, import the library.
+
+## Creating a pubspec
+
+The pubspec is a file named <code class="literal">pubspec.yaml</code>
+that's in the top directory of your application.
+The simplest possible pubspec lists only the package name:
```
mit-mit: Maybe expand to '...only the package name of the app'?

src/_guides/packages.md
```diff
@@ -0,0 +1,227 @@
...
+
+{% comment %} TODO: make 3-backtick-yaml the same as prettify yaml {% endcomment %}
+{% prettify yaml %}
+name: my_app
+{% endprettify %}
+
+Here is an example of a pubspec that declares dependencies on
+two packages that are hosted on the Pub site:
```
mit-mit: Consider s/two packages/two packages `js` & `intl`/

src/_guides/packages.md
```diff
@@ -0,0 +1,227 @@
...
+dependencies:
+  js: ^0.6.0
+  intl: ^0.15.8
+{% endprettify %}
+
+For details on creating a pubspec,
+see the [pubspec documentation](/tools/pub/pubspec)
+and the documentation for the packages that you want to use.
+
+## Installing packages
```
mit-mit: Installing sounds like it's more than just downloading? Maybe 'Retrieving packages'?

src/_guides/packages.md
```diff
@@ -0,0 +1,227 @@
...
+$ pub get
+```
+
+This process is called _getting the dependencies_.
+
+The `pub get` command determines which packages your app depends on,
+and puts them in a central [system cache](/tools/pub/glossary#system-cache).
+If your app depends on a published package, pub downloads that package from the
+[Pub site.]({{site.pub}})
+For a git dependency, pub clones the git repository.
```
mit-mit: Consider linking 'git dependency' to some place that had details for that

src/_guides/packages.md
```diff
@@ -0,0 +1,227 @@
...
+pub downloads the latest version of it that's compatible with
+your other dependencies.
+It then locks your package to *always* use that version by
+creating a **lockfile**.
+This is a file named `pubspec.lock` that pub creates and stores next to your
+pubspec. It lists the specific versions of each dependency (immediate and
+transitive) that your package uses.
+
+If your package is an application package,
+you should check this file into source control.
```
mit-mit: Maybe link 'source control' to /guides/libraries/private-files ?

src/_guides/packages.md
kwalrath: This would be a pretty invasive change, and we have a policy of not referring to files by name. (Witness the packages dir/file fun and the name change of the analysis options file.) But I will clarify the name here and keep an eye out for other places.

src/_guides/packages.md
kwalrath: The context is broader in this sentence, and I'd have to make it a lot more specific if I add those words.

src/_guides/packages.md
kwalrath: It gets a little more complicated, since you don't download packages that are already in the system cache.

mit-mit: I really dislike 'installing'. Lots of developers dislike tools that 'install' stuff; installation typically means mocking with systems directories, system configuration, etc.

I think 'Getting' would work great. I don't see how duplicating the command name is a bad thing. That seems like a good thing.

kwalrath: Thinking about it, I agree that Getting works here.

kwalrath:
I think we've resolved all issues. I'm going to merge this, so I can merge one of my other PRs...

PR #1518, opened 2019-04-30
Update example path to dart sdk w/ homebrew path

legalcodes:


src/tools/jetbrains-plugin.md
```diff
@@ -89,8 +89,8 @@ Here's one way to configure Dart support:
   </p>
 
   <p>
-    For example, the SDK path might be
-    <code><em>&lt;dart installation directory></em>/dart/dart-sdk</code>.
+    For example, the SDK path might be (if you installed via homebrew)
```
kwalrath: I don't think the majority of users are using macOS. (Analytics should be able to tell us.) Maybe we should show two or three common paths?

mit-mit: Actually, the majority (~45%) are on Windows. Then macOS, and then a small share on Linux.

kwalrath: Do we know the path that the majority of Windows users are likely to have?

mit-mit: Sorry, no, I think that varies a lot depending on how they install it.

kwalrath: @legalcodes maybe you could ask on Gitter?

legalcodes: Great suggestion re: showing multiple common paths. Will find out the most common path for Windows setup. Thanks!

kwalrath:
@legalcodes were you able to get any Windows details?

legalcodes:
I was able to get a Windows machine -- will go through the install flow and update this issue.

mit-mit:
Is this still being worked on?

legalcodes:
Yes, though windows setup is taking longer than expected. Will keep this issue updated.

kwalrath:
@legalcodes any progress, or should I close this?

PR #1523, opened 2019-04-30
Fix a few Uri's

mit-mit:


kwalrath:
I didn't know pub.dev was working already! I think we mention pub.dartlang.org in text, but we can fix that in a followup PR.

PR #1527, opened 2019-04-30
align navigation arrows with text

johnpryan:
this aligns the arrows and text towards the vertical center:

Screen Shot 2019-04-30 at 2 25 19 PM

### DIFF
--- a/src/_assets/css/main.scss
+++ b/src/_assets/css/main.scss
@@ -480,11 +480,16 @@ img {
       font-size: 13px;
       line-height: 45px;
 
+      // center text and arrow button
+      display: flex;
+      flex-direction: row;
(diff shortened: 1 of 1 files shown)


src/_assets/css/main.scss
```diff
@@ -480,11 +480,16 @@ img {
...
 
+      // center text and arrow button
+      display: flex;
+      flex-direction: row;
+      align-items: center;
+
       // copy of shared/_sidebar.scss
       // @mixin collapsable() is not reachable from here, this is a copy of that
       &.collapsable {
         display: flex;
```
isoos: Nit: if you add the styles on the parent rule, they can be removed from here.

PR #1539, opened 2019-05-02
Remove /articles/server

kwalrath:
And move /articles/server/native-extensions to /server/c-interop-old.

I'm not sure I got the facts right, so I'll need a sanity check from mit-mit@.


src/server/c-interop-old.md
```diff
@@ -1,15 +1,15 @@
...
-category: dart-vm
+description: The original way for command-line Dart apps to call C/C++ functions.
 ---
 
 _Written by William Hesse <br>
 May 2012_
 
 <aside class="alert alert-warning" markdown="1">
-  The example code on this page has some
+  The extension mechanism discussed in this page is being replaced by an
```
mit-mit: I don't know that we have decided to do that, actually. Let's be a bit more general, e.g.:

The extension mechanism discussed in this page is for deep integration of the VM. If you just need to call existing code written in C or C++, see [C & C++ interop using FFI]()

src/server/c-interop.md
```diff
@@ -12,4 +12,7 @@ to call native C and C++ APIs.
 This mechanism is currently [in active
 development](https://github.com/dart-lang/sdk/issues/34452) and is not yet
 complete. For a small preview example, see the [dart:ffi sqllite
-sample](https://github.com/dart-lang/sdk/blob/master/samples/ffi/sqlite/README.md).
+sample.](https://github.com/dart-lang/sdk/blob/master/samples/ffi/sqlite/README.md)
+
+For information on the old native interface, see
```
mit-mit: I don't think we need to link in this direction

src/server/c-interop-old.md
```diff
@@ -1,15 +1,15 @@
 ---
 title: Native extensions for the standalone Dart VM
-description: Learn how to enable command-line Dart apps to call C/C++ functions.
-date: 2012-05-01
-category: dart-vm
+description: The original way for command-line Dart apps to call C/C++ functions.
```
mit-mit: I think I would use something like 'Integrating Dart and native code using native extensions'

Also, I would prefer not to rename the file. Old carries connotations. Native-extensions.md is neutral.

src/_data/side-nav.yml
```diff
@@ -84,8 +84,6 @@
           permalink: /server/libraries
         - title: C & C++ interop
           permalink: /server/c-interop
-        - title: Native extensions
```
mit-mit: I think we may want to keep (an updated version) of this page for the specialized use case of embedding Dart into existing native apps. Can we log a bug to remind ourselves to discuss that later?

src/server/c-interop-old.md
kwalrath: I'd like it to start with the same letters as the other page, so the other page will be selected in the nav when this one is current. Maybe... c-interop-extensions.

src/_data/side-nav.yml
kwalrath: Done: #1548

kwalrath:
ptal

src/server/c-interop-old.md
mit-mit: `c-interop-native-extensions` would be ok

PR #1540, opened 2019-05-02
Update landing page with latest text and images

mit-mit:


mit-mit:
@kwalrath can you check that I moved the content over correctly?
@FaisalAbid can you check that I didn't mess with the implementation?

FaisalAbid:
Just tested and works good

FaisalAbid:
Ah cool, i'll merge my branch with this when its in dash

mit-mit:
@kwalrath I'm just going to merge this so Faisal can work off it. If you think we need text updates, just do a new PR!

PR #1543, opened 2019-05-02
Proper link styles for external links (per mock spec)

FaisalAbid:


kwalrath:
Sorry it's taken me so long to get back to you... I'm having trouble seeing the difference. What should I be looking for?

PR #1556, opened 2019-05-03
use a 1330px max-width for header

johnpryan:
This is for users with large screens... we don't want to make the video too big 

before:

Screen Shot 2019-05-03 at 12 40 57 PM

after:

Screen Shot 2019-05-03 at 12 40 38 PM


FaisalAbid:
Nice I love it

sent from my phone

PR #1562, opened 2019-05-04
Remove 32-bit support based on feedback

mit-mit:


mit-mit:
Good catch! Forgot to git add it first time around

PR #1578, opened 2019-05-06
image when on mobile

FaisalAbid:
@mit-mit


mit-mit:
Maybe I'm missing something, but will this not cause both the video and the image to show when on mobile?

FaisalAbid:
the `'d-sm-none d-md-block'` means on narrow dissapear and md and above show up

and vice versa for `d-sm-block d-md-none`

kwalrath:
> Maybe I'm missing something, but will this not cause both the video and the image to show when on mobile?

It doesn't look like it does: https://dartlang-org-staging-1.firebaseapp.com

mit-mit:
Got it; did a bit of testing and seems to work great

PR #1596, opened 2019-05-08
focus outline and tab navigation

FaisalAbid:
Makes the website navigate-able via tab including the galleries


kwalrath:
Staged: https://dartlang-org-staging-0.firebaseapp.com

FaisalAbid:
@kwalrath is it okay that the outline shows up if you click on the <li> element also? 

I can look into making it such that it doesn't when you click but only when you tab

FaisalAbid:
Actually hold off on merging this, something is up with my hover state

PR #1629, opened 2019-05-20
Removed [spread operator proposal] link and line.

Suyash-Purwar:
[spread operator proposal link] was broken. I checked the source code and I found that the URL was broken due to the absence of folder which contains the file to which it was made to be directed. Deleted the whole sentence because it was meaningless without the link.


Suyash-Purwar:
I have added back the links and updated the URLs.

Thanks for the support!

PR #1631, opened 2019-05-21
Update deps

mit-mit:
Both `rack` and `bootstrap` are stuck on old versions flagged for vulnerabilities


kwalrath:
Thanks, @mit-mit!

Issue #1447, opened 2019-04-09
'Pub Package Layout Conventions' page issue: LICENSE is REQUIRED


kevmoo:
See https://github.com/dart-lang/pub/blob/master/lib/src/validator/license.dart

Not sure how we want to document this.

Must be one of `LICENSE`, `COPYING`, `UNLICENSE` with any extension and any case, it appears.

The docs make it seems like this is optional for publishing...tis not...

jonasfj:
I suggest that we:
1. State that `LICENSE` is required, and,
2. Do not mention that the file could be named `COPYING` or `UNLICENSE`.

Some day we should probably make a specification for the packaging format. This specification should ideally not leave room for ambiguity, but it might have to feature an appendix documenting that legacy packages may contain `COPYING` or `UNLICENSE` instead of `LICENSE`.

kwalrath:
Fixed by #1640.

Issue #1344, opened 2019-02-06
'Pub Versioning Philosophy' page issue


shrike69:
Page URL: https://www.dartlang.org/tools/pub/versioning
Page source: https://github.com/dart-lang/site-www/tree/master/src/tools/pub/versioning.md

Description: In the paragraph that reads

When you add a dependency to your package, you’ll sometimes want to specify a range of versions to allow. How do you know what range to pick? You need to forward compatible, so ideally the range encompasses future versions that haven’t been released yet. But how do you know your package is going to work with some new version that doesn’t even exist yet?

the third sentence should read

You need to be forward compatible...

kwalrath:
I'll happily fix this typo, but in the future, it might be quicker for you to just edit the page.

shrike69:
Sorry, I thought I was following the correct procedure. I'm happy to fix typos myself. Can you private message me with the info on how to do that?

kwalrath:
You did nothing wrong! We're always happy to accept either issues or PRs for little things like typos. For bigger issues, we prefer some advance notice so we can coordinate. I should update the [CONTRIBUTING](https://github.com/dart-lang/site-www/blob/master/CONTRIBUTING.md) file to be clearer about this.

kwalrath:
Fixed by #1640.

Issue #1321, opened 2019-01-22
Add README.md info to package layout page


kwalrath:
See dart-lang/pub-dartlang-dart#1973 for details.

kwalrath:
@isoos or @jonasfj, can you point me to a good example of a package that has an example/README.md file? I looked in the sdk repo but didn't find one there.

isoos:
@kwalrath: Not sure what qualifies as good, but this package has many things in the example directory, and uses a readme.md:  https://github.com/flutter/plugins/tree/master/packages/in_app_purchase

isoos:
A few others:
- https://github.com/jacobaraujo7/bloc-pattern
- https://github.com/DanTup/tiler
- https://github.com/RafaelBarbosatec/achievement_view_flutter
- https://github.com/ganeshsp1/wallpaperplugin

kwalrath:
Fixed by #1640.

Issue #1317, opened 2019-01-16
Pubspec page should warn about generic custom top-level fields


jonasfj:
Currently the [pubspec page](https://github.com/dart-lang/site-www/blob/1f27503b89a57d3bc50c96cd625d4e742f5c1435/src/tools/pub/pubspec.md#supported-fields) says that additional top-level fields are ignored.

We should document that:
 * custom top-level fields should have unique names (maybe suggest using a registered package name).
 * generically named top-level fields should be avoided as `pub` may choose to define a format for such fields in the future, which would break your package.
 * We could go further to suggest that custom top-level fields be prefixed with `_` or `$`, such that there is a clear convention to follow.

Examples of generic field names includes:
 * `bugs`
 * `license`
 * `scripts`
 * `resources`
 * `links`
 * `config`
 * `build_steps`

/cc @munificent

kwalrath:
@jonasfj what would be an example of a registered package name? `my_pkg_resources`?

I'm trying to think of how to phrase this. Maybe something like this:

> ## Supported fields
> ...
> 
> Pub ignores all other fields.
> 
> If you add custom fields to the pubspec, avoid generic field names, which might clash with future additions to the supported fields. We recommend a prefix of `$`, `_`, or your package name. For example, instead of `resources`, you might add `$resources`, `_resources`, or `my_pkg_resources`.

/cc @kevmoo

jonasfj:
It might be fine to just discourage generically named properties.

Most people probably don't want to read a lot of about custom top-level properties anyways :)

(note, I doubt is high priority, it's mostly just a nit)

kwalrath:
Fixed by #1640.

PR #1640, opened 2019-05-24
Misc. small pub page improvements

### LINKED ISSUE #1623: 'pub global' page issue
Page URL: https://www.dartlang.org/tools/pub/cmd/pub-global.html
Page source: https://github.com/dart-lang/site-www/tree/master/src/tools/pub/cmd/pub-global.md

Description of issue:

#activating-a-package does not explain the term "activate".  After reading the doc, I understand _how_ to use "pub global activate" but have no idea _what_ the command does.

kwalrath:
Fixes #1623, #1447, #1344, #1321, #1317, #1305, #1182, #801, #757

Staged:
https://dartlang-org-staging-0.firebaseapp.com/tools/pub/obsolete
https://dartlang-org-staging-0.firebaseapp.com/tools/pub/package-layout
https://dartlang-org-staging-0.firebaseapp.com/tools/pub/pubspec
https://dartlang-org-staging-0.firebaseapp.com/tools/pub/versioning
https://dartlang-org-staging-0.firebaseapp.com/tools/pub/cmd/pub-global

### DIFF
--- a/src/tools/pub/package-layout.md
+++ b/src/tools/pub/package-layout.md
@@ -101,9 +101,11 @@ enchilada/
   LICENSE
 {% endprettify %}
 
-If you're publishing your package, you need a license file.
-The [Pub site]({{site.pub}}) accepts files named `LICENSE`,
-`COPYING`, or `UNLICENSE`, with any file extension. 
+If you're publishing your package, include a license file named `LICENSE`,
+optionally with a file extension such as `.md`. 
+We recommend using an [OSI-approved license](https://opensource.org/licenses)
+such as [BSD-3-Clause,](https://opensource.org/licenses/BSD-3-Clause)
+so that others can reuse your work.
 
 ## README
 
@@ -118,7 +120,7 @@ to the [Pub site,]({{site.pub}}) your README is shown on
 the page for your package. This is the perfect place to introduce people to
 your code.
 
-If your README ends in `.md`, `.markdown`, or `.mdown`, it's parsed as
+If your README ends in `.md`, it's parsed as
 [Markdown.][markdown]
 
 [markdown]: http://daringfireball.net/projects/markdown/
@@ -136,7 +138,7 @@ release. When you upload your package to the
 [Pub site,]({{site.pub}}) your package's changelog file (if any)
 appears in the changelog tab.
 
-If your CHANGELOG ends in `.md`, `.markdown`, or `.mdown`, it's parsed as
+If your CHANGELOG ends in `.md`, it's parsed as
 [Markdown.][markdown]
 
 ## Public directories
--- a/src/tools/pub/package-layout.md
+++ b/src/tools/pub/package-layout.md
@@ -14,7 +14,7 @@ package, and how to name things.
   **Note:**
   Flutter apps can use custom directories for their assets.
   For details, see
-  [Adding Assets and Images in Flutter]({{site.flutter}}/assets-and-images/)
+  [Adding assets and images]({{site.flutter}}/assets-and-images/)
   on the [Flutter website.]({{site.flutter}})
   </div>
 </div></aside>
@@ -29,9 +29,9 @@ enchilada/
   .packages *
   pubspec.yaml
   pubspec.lock **
+  LICENSE
   README.md
   CHANGELOG.md
-  LICENSE
   benchmark/
     make_lunch.dart
   bin/
@@ -67,13 +67,13 @@ enchilada/
     [application package](/tools/pub/glossary#application-package).
 
 \*** The `doc/api` directory exists locally after you've run
-      [dartdoc](https://github.com/dart-lang/dartdoc#dartdoc).
+      [dartdoc.](https://github.com/dart-lang/dartdoc#dartdoc)
       Don't check the `api` directory into source control.
 
 {% include packages-dir.html %}
 
 
-## The basics
+## The pubspec
 
 {% prettify none %}
 enchilada/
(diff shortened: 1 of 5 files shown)


src/tools/pub/package-layout.md
```diff
@@ -67,13 +67,13 @@ enchilada/
     [application package](/tools/pub/glossary#application-package).
 
 \*** The `doc/api` directory exists locally after you've run
-      [dartdoc](https://github.com/dart-lang/dartdoc#dartdoc).
+      [dartdoc.](https://github.com/dart-lang/dartdoc#dartdoc)
```
kevmoo: Really? I'd think you want the dot outside the link?

src/tools/pub/package-layout.md
```diff
@@ -115,12 +114,12 @@ enchilada/
 
 One file that's very common in open source is a README file that
 describes the project. This is especially important in pub. When you upload
-to the [Pub site]({{site.pub}}), your README is shown on
+to the [Pub site,]({{site.pub}}) your README is shown on
```
kevmoo: ditto

src/tools/pub/pubspec.md
```diff
@@ -74,14 +74,19 @@ A pubspec can have the following fields:
 : Optional. Specify where to publish a package.
   [_Learn more._](#publish_to)
 
-Pub ignores all other fields.
+Pub ignores all other fields,
```
kevmoo: Really?

src/tools/pub/package-layout.md
```diff
@@ -85,26 +85,25 @@ Every package has a [_pubspec_](/tools/pub/pubspec), a file named
...
-Don't check this into source control.
-
-The open source community has a few other files that commonly appear at
-the top level of a project: `LICENSE`, `AUTHORS`, etc. If you use any
-of those, they can go in the top level of the package too.
-
-For more information, see the [pubspec page](/tools/pub/pubspec).
+If you're publishing your package, you need a license file.
+The [Pub site]({{site.pub}}) accepts files named `LICENSE`,
+`COPYING`, or `UNLICENSE`, with any file extension.
```
jonasfj: I would prefer if we only documented `LICENSE`, I'm not convinced we should encourage users to use different names for license files...

Maybe in some future we want to formalize the packaging format, and then I could imagine that we would require it be called `LICENSE`. It's unfortunate that we accept multiple filenames right now, as it leaves room for ambiguity.

jonasfj: Side note: If you want to make a recommendation, you can recommend using an OSI approved license as this allows other to reuse your work.
Maybe even point out that all packages by the Dart team is released under "BSD-3-Clause" (https://spdx.org/licenses/BSD-3-Clause.html).

src/tools/pub/package-layout.md
```diff
@@ -115,12 +114,12 @@ enchilada/
...
 One file that's very common in open source is a README file that
 describes the project. This is especially important in pub. When you upload
-to the [Pub site]({{site.pub}}), your README is shown on
+to the [Pub site,]({{site.pub}}) your README is shown on
 the page for your package. This is the perfect place to introduce people to
 your code.
 
-If your README ends in `.md`, `.markdown`, or `.mdown`, it is parsed as
-[Markdown][].
+If your README ends in `.md`, `.markdown`, or `.mdown`, it's parsed as
```
jonasfj: Can we remove `.markdown` and `.mdown`.

I think `.md` has won on GitHub, and there is no reason for us to suggest that users use anything other than `README.md`. Even if we also support other filenames at the moment.

src/tools/pub/package-layout.md
```diff
@@ -134,11 +133,11 @@ enchilada/
 To show users the latest changes to your package, you can include a changelog
 file where you can write a short note about the changes in your latest
 release. When you upload your package to the
-[Pub site]({{site.pub}}), your package's changelog file (if any)
+[Pub site,]({{site.pub}}) your package's changelog file (if any)
 appears in the changelog tab.
 
-If your CHANGELOG ends in `.md`, `.markdown`, or `.mdown`, it is parsed as
-[Markdown][].
+If your CHANGELOG ends in `.md`, `.markdown`, or `.mdown`, it's parsed as
```
jonasfj: Same here, can we limit the suggestion to `CHANGELOG.md`?

src/tools/pub/package-layout.md
kwalrath: We have a little external-link indicator that looks terrible when it's inside the punctuation.

https://github.com/dart-lang/site-angulardart/wiki/Why-does-external-link-text-contain-punctuation%3F

kevmoo: Oh... 🤷‍♂

src/tools/pub/pubspec.md
kwalrath: @jonasfj or @isoos can you confirm?

isoos: Yes, at the moment pub site ignores other fields. We had discussions to reject pubspecs that have extra fields, but at the time it was decided to continue ignoring instead.

kwalrath: Hmmm... That doesn't sound as ironclad as I thought it was. Do you think there's a chance we'll change our minds?

jonasfj: > Do you think there's a chance we'll change our minds?

There is always a small chance/risk we'll change it, I would like to, but that's my opinion and I don't see us having the bandwidth anytime soon... On the other hand, do we need to document that other fields are ignored?

ie. if we don't write anything about it, the "feature" is less likely to be abused. IMO, if we were to write anything, we should encourage users introducing custom properties to avoid using generic names (ie. use unique names, such as their package name, or something like that).

PR #1645, opened 2019-05-28
Update Dart SDK version in pkg-vers.json to 2.3.1

mit-mit:
Hi Kathy, I think we should consider rolling the Dart SDK on the website everytime we ship a stable SDK?


kwalrath:
I'm working on a PR that'll update all the examples to 2.3.1 and package:pedantic.

Issue #1651, opened 2019-05-30
Update lints to use the latest language features (when appropriate)


kwalrath:
#1652 introduced the pedantic package. Now we need to add some lints, such as prefer_collection_literals, and perhaps enable the one pedantic lint we turned off (avoid_shadowing_type_parameters).

Related: #1365

kwalrath:
I don't think this is urgent. (I'm not even sure it's still necessary.)

Issue #1469, opened 2019-04-19
Re-enable stable check-code


kwalrath:
Related: #1446.

PR #1652, opened 2019-05-31
 Use 2.3.1 SDK and pedantic package

### LINKED ISSUE #1365: Add set literals to docs where appropriate
Library tour, Effective Dart (#1410).

Related: #1257.

kwalrath:
Contributes to #1651.

In a subsequent PR, we should update the code samples to work with avoid_shadowing_type_parameters (if possible) and add lints such as prefer_collection_literals, which will help us fix #1365.

### DIFF
--- a/examples/analysis_options.yaml
+++ b/examples/analysis_options.yaml
@@ -10,6 +10,7 @@ analyzer:
 
 linter:
   rules:
+    avoid_shadowing_type_parameters: false # TODO: reinstate?
     - annotate_overrides
     - await_only_futures
     - camel_case_types
--- a/examples/analysis_options.yaml
+++ b/examples/analysis_options.yaml
@@ -1,6 +1,8 @@
 # Supported lint rules and documentation:
 #   http://dart-lang.github.io/linter/lints/
 
+include: package:pedantic/analysis_options.yaml
+
 analyzer:
   exclude: [build/**]
   strong-mode:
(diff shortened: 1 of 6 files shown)


examples/httpserver/pubspec.yaml
```diff
@@ -3,10 +3,11 @@ description: Sample HTTP clients and servers
 homepage: https://www.dartlang.org/docs/tutorials/httpserver/
 
 environment:
-  sdk: '>=2.1.0 <3.0.0'
+  sdk: '>=2.3.1 <3.0.0'
```
johnpryan: I think travis is unhappy with this change:

```
Because httpserver requires SDK version >=2.3.1 <3.0.0, version solving failed.
```

Can we set it to 2.3.0 or do we need to upgrade the SDK?

kwalrath: I think we should use 2.3.1, since that's the latest stable version. Let me see what I can do to make that stick...

kwalrath: Turns out I filed a bug to remind myself to do this (or something similar, anyway): #1469.

examples/analysis_options.yaml
```diff
@@ -1,64 +1,42 @@
...
     - package_api_docs
     - package_names
     - package_prefixed_library_names
-    - prefer_contains
-    - prefer_equal_for_default_values
-    - prefer_is_not_empty
-    #- public_member_api_docs
-    - recursive_getters
-    - slash_for_doc_comments
     #- sort_constructors_first # TODO: reinstate after https://github.com/dart-lang/site-www/issues/1196 is resolved?
```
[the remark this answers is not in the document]
kwalrath: I'm reinstating this (assuming the Travis build passes, like my local one did).

johnpryan:
👍

PR #1653, opened 2019-05-31
Clarify coverage of updating/disabling pedantic rules

kwalrath:
While updating site-www examples to use package:pedantic, I noticed some places where we could improve the docs.

### DIFF
--- a/src/_guides/language/analysis-options.md
+++ b/src/_guides/language/analysis-options.md
@@ -181,8 +181,9 @@ include: package:pedantic/analysis_options.yaml
   When a **new version of `pedantic`** is published,
   code that previously passed analysis might **start failing analysis.**
   We recommend updating your code to work with the new rules.
-  Other options are to use a [specific version of pedantic,][]
-  include a specific version of the [pedantic analysis options file,][]
+  Other options are to
+  include a specific version of the pedantic analysis options file
+  (as described [in the pedantic README][]),
   explicitly enable individual linter rules,
   or [disable individual rules][].
 </aside>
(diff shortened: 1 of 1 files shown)


src/_guides/language/analysis-options.md
```diff
@@ -181,8 +181,9 @@ include: package:pedantic/analysis_options.yaml
   When a **new version of `pedantic`** is published,
   code that previously passed analysis might **start failing analysis.**
   We recommend updating your code to work with the new rules.
-  Other options are to [use a specific version][]
-  of `pedantic`, explicitly enable individual linter rules,
+  Other options are to use a [specific version of pedantic,][]
+  include a specific version of the [pedantic analysis options file,][]
```
davidmorgan: We want to give one suggestion here, not two.

The original idea was that they would depend on a specific version of `pedantic`, but we realized this would be horribly broken as soon as you want to depend on two packages that are using different versions of `pedantic`--it would be a version conflict. So actually we need everyone on the same (or at least, compatible) version of `pedantic`.

The new form of words is better though, we want people to use a specific version of the options, not of the package. So e..

Other options are to include a specific version of the [pedantic analysis options file,][], explicitly enable ...

src/_guides/language/analysis-options.md
```diff
@@ -402,4 +410,5 @@ Use the following resources to learn more about static analysis in Dart:
 [linter rules]: http://dart-lang.github.io/linter/lints/
 [sound-dart]: /guides/language/sound-dart
 [todo]: {{site.pub-api}}/analyzer/latest/analyzer/TodoCode/TODO-constant.html
-[use a specific version]: https://github.com/dart-lang/pedantic/blob/master/README.md#using-the-lints
+[specific version of pedantic,]: https://github.com/dart-lang/pedantic/blob/master/README.md#using-the-lints
+[pedantic analysis options file,]: https://github.com/dart-lang/pedantic/tree/master/lib
```
davidmorgan: I think the original link is better--at the end of the section we show how to depend on a specific version of the yaml file.

https://github.com/dart-lang/pedantic/blob/master/README.md#using-the-lints

They do still need to add the dependency. `^1.7.0` means 'use anything compatible with 1.7.0' so it gets around the version incompatibility problem I described.

kwalrath: I realized that it's probably best to link to the pub.dev version of the README, instead of the github blob, so now I'm linking to https://pub.dev/packages/pedantic#using-the-lints.

davidmorgan:
Looks good, thanks!

kwalrath:
Thanks for the review, @davidmorgan!

PR #1654, opened 2019-05-31
[DO NOT MERGE] Mockup of diagnostics page

kwalrath:
Staged: https://kw-www-dartlang-1.firebaseapp.com/tools/diagnostics

Here's my first go at formatting this page. I've noticed a few issues:

* Names in the TOC are cut off (when the page is wide enough for the TOC to appear on the right).
* It looks awkward to me (and is forbidden by the [style guide](https://developers.google.com/style/headings)) to have two headings with no text between them. For example, `### AMBIGUOUS_SET_OR_MAP_LITERAL_BOTH` is immediately followed by `#### Message`. Maybe we could just drop the **Message** headings?
* The code highlighting (e.g. in the Example part of https://kw-www-dartlang-1.firebaseapp.com/tools/diagnostics#default_value_on_required_parameter) could be more noticeable.

Do you see anything else? Have any suggestions? Feel free to grab this PR and play with it.


src/tools/diagnostics.md
```diff
@@ -0,0 +1,2436 @@
...
+
+### AMBIGUOUS_SET_OR_MAP_LITERAL_BOTH
+
+#### Message
+
+This literal contains both 'Map' and 'Iterable' spreads, which makes it impossible to determine whether the literal is a map or a set.
+
+#### Description
+
+Because map and set literals use the same delimiters (`{` and `}`), the analyzer looks at the type arguments and the elements to determine which kind of literal you meant. When there are no type arguments and all of the elements are spread elements (which are allowed in both kinds of literals), then the analyzer uses the types of the expressions that are being spread. If all of the expressions have the type `Iterable`, then it's a set literal; if they all have the type `Map`, then it's a map literal.
```
legalcodes: Our styles aren't changing the background color for monospaced codeblocks, so what should look acceptable here e.g. :

the same delimiters (`{` and `}`), the analyzer...

ends up looking like this:
Screen Shot 2019-05-31 at 3 29 17 PM

(The above is a little jarring to my eye, but disregard this comment if it's palatable for everyone else)

One workaround is to find a way to craft the sentence without using the parens, e.g. :

Because map and set literals use the same delimiters `{` and `}` the analyzer looks...

legalcodes:
This is purely an enhancement, not a bug:

It'd be nice to replace our highlighted code style with a squiggly-underline code style, for at least two reasons:
1. The audience for the docs will likely be users who have encountered an analyzer error in their editor, and most editors will be squiggly-underlining the offending code. Using the same style to highlight offending code in these docs will provide continuity in user experience (sort of like welcoming the user with a big sign that says "Hi! You're here because of syntax issues and we're here to help...)
2. Squiggly-underlining offending code in these docs might  help to minimize the cost of context-switching on the user's cognitive load and ultimately reduce the amount of time it takes the user to fix the issue.

src/tools/diagnostics.md
```diff
@@ -0,0 +1,2436 @@
...
+
+The analyzer produces this diagnostic when some of the expressions being spread have the type `Iterable` and others have the type `Map`, making it impossible for the analyzer to determine whether you are writing a map literal or a set literal.
+
+#### Example
+
+The following code produces this diagnostic:
+
+{% prettify dart %}
+union(Map<String, String> a, List<String> b, Map<String, String> c) => [!{...a, ...b, ...c}!];
+{% endprettify %}
```
legalcodes: Is there a reason we weren't using the \`\`\` dart for syntax highlighting this codeblock? We're losing syntax highlighting only for this codeblock and it's a little confusing to read:

Screen Shot 2019-05-31 at 3 49 04 PM

kwalrath: IIRC, highlights didn't work if I used the ```dart syntax. We should fix this.

kwalrath:
I forgot to add that the line length is really long. I'd want 80-character lines — or possibly much shorter. (I've always liked breaking on phrases, to reduce diff noise.)

src/tools/diagnostics.md
```diff
@@ -0,0 +1,2436 @@
...
+
+### Potentially non-nullable
+
+A type is _potentially non-nullable_ if it's either explicitly non-nullable or
+if it's a type parameter.
+The latter case is included because
+the actual runtime type might be non-nullable.
+
+
+## Compile-time errors
```
bwilkerson: I seem to remember that we had decided not to group diagnostics by their severity. I had those headers in the doc because it made it easier for me to keep track of which diagnostics already had documentation, not because I thought they belonged in the output.

The strongest argument I can remember against grouping diagnostics by severity is that users can change the severities in the analysis options file, so the published severity won't necessary match what they see in the IDE and command line output.

Also, given the way I expect users will interact with this page, I don't believe that grouping the diagnostics (in any way) will be useful.

I am, of course, happy to discuss the question. But if you agree that we don't want to group by severity, then I need to know whether we want a second-level heading before the diagnostics (to align with "Glossary") or whether each diagnostic name will be a second level heading.

src/tools/diagnostics.md
```diff
@@ -0,0 +1,2436 @@
...
+
+The analyzer produces the following diagnostics when [PENDING...].
+
+### AMBIGUOUS_SET_OR_MAP_LITERAL_BOTH
+
+#### Message
```
bwilkerson: I'm fine with dropping the "Message" heading. I don't think it adds much, and there's nothing that I can think of to put between it and the name of the diagnostic.

src/tools/diagnostics.md
bwilkerson: It looks odd to me to drop the parens, but I'm not a professional writer, so I'll go with whatever the two of you decide.

src/tools/diagnostics.md
```diff
@@ -0,0 +1,2436 @@
...
+  sdk: '>=2.2.0 <2.4.0'
+```
+
+If you do need to support older versions of the SDK, then replace the set literal with code that creates the set without the use of a literal:
+
+```dart
+var s = new Set<int>();
+```
+
+{% comment %}
```
bwilkerson: I'm not planning to make the generator generate any of the content from this line down. Let me know if there's some value to doing so.

bwilkerson:
> It'd be nice to replace our highlighted code style with a squiggly-underline code style ...

+1

> I forgot to add that the line length is really long.

The line length in the doc, or the line length in the PR?

In either case, when I copy the documentation into the analyzer code base it will get re-formatted to fit within the 80 character line length, so that issue should go away.

bwilkerson:
This is very cool to see! Thank you!

Were there any content changes other than increasing the depths of the headers and adding 'dart' to the code blocks? I had been planning on grabbing the documentation comment from the doc when I move it into the analyzer code base, but want to know if I need to grab it from this PR instead.

src/tools/diagnostics.md
```diff
@@ -0,0 +1,2436 @@
...
+---
+
+This page lists all diagnostic messages produced by the Dart analyzer,
+with details about what those messages mean
+and how you can fix your code.
+For more information about the analyzer, see
+[Customizing static analysis](/guides/language/analysis-options).
+
+
+## Glossary
```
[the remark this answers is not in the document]
kwalrath: I have a slight preference for the top of the page, because some people walk up the URL hierarchy, and those that do are the thorough types who might actually read the glossary.

src/tools/diagnostics.md
kwalrath: I'm fine with not grouping by severity. They should probably be alphabetical. I'd keep them as third-level headings, so we'd have two second-level headings:

* Glossary
* Diagnostics

kwalrath:
> Were there any content changes other than increasing the depths of the headers and adding 'dart' to the code blocks? I had been planning on grabbing the documentation comment from the doc when I move it into the analyzer code base, but want to know if I need to grab it from this PR instead.

I don't remember, unfortunately... I don't think I did anything except make formatting changes, but there were probably a bunch of those (e.g. code font within paragraphs).

src/tools/diagnostics.md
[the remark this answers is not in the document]
bwilkerson: Will do.

src/tools/diagnostics.md
bwilkerson: Will do.

bwilkerson:
No problem. I'll compare the two and update the doc accordingly.

kwalrath:
> No problem. I'll compare the two and update the doc accordingly.

@bwilkerson do you want me to update this mockup, or are you good to go?

src/tools/diagnostics.md
```diff
@@ -0,0 +1,2436 @@
...
+Because map and set literals use the same delimiters (`{` and `}`), the analyzer looks at the type arguments and the elements to determine which kind of literal you meant. When there are no type arguments and all of the elements are spread elements (which are allowed in both kinds of literals), then the analyzer uses the types of the expressions that are being spread. If all of the expressions have the type `Iterable`, then it's a set literal; if they all have the type `Map`, then it's a map literal.
+
+The analyzer produces this diagnostic when some of the expressions being spread have the type `Iterable` and others have the type `Map`, making it impossible for the analyzer to determine whether you are writing a map literal or a set literal.
+
+#### Example
+
+The following code produces this diagnostic:
+
+{% prettify dart %}
+union(Map<String, String> a, List<String> b, Map<String, String> c) => [!{...a, ...b, ...c}!];
```
bwilkerson: I just noticed (while verifying that there are no non-formatting related changes) that you replaced all of the span tags with exclamation points.

Is there a different markdown that we could employ? The exclamation point is a prefix operator `!running`, used in type tests `x is! Foo`, part of a comparison operator `x != y`, and a postfix operator in nnbd `x!`.

If not, will we need to remember to escape it? Do we use a backslash?

bwilkerson:
> ... do you want me to update this mockup ...

I think I have enough information to get at least a first cut at the generator working. I'm sure we'll want to iterate (for example, about the markdown for highlighting), but it's probably easier to work from something concrete.

src/tools/diagnostics.md
kwalrath: We can use our old convention: `[[highlight]]...[[/highlight]]`. I think we dropped it because it took up a lot of space. I believe that all of the examples on flutter.dev & dart.dev & angulardart.dev use the new syntax, and it hasn't caused any problems. But the old syntax has the advantage of being very clear!

bwilkerson: Ok. I might be borrowing trouble. Sounds like it's easier on your end if we go with the new syntax. I assume I can use a backslash when necessary.

kwalrath:
I wanted to test the highlighting, so I've updated the draft to use the new highlighting. I've also deleted the unneeded headings (changing the alphabetization as needed), and deleted the commented out text.

kwalrath:
I've updated https://kw-www-dartlang-1.firebaseapp.com/tools/diagnostics. I'm not crazy about the formatting of the message. I'm going to play with that a bit...

kwalrath:
I tried a few different ways to format the message. I think I like italics (`<em>...</em>`) the best, but bold is OK too. I think using an HTML tag also has the advantage of not tempting the markdown parser into formatting the message specially.

Code font:
image
"Message:" intro:
image
Italics (`<em>...</em>`):
image
Bold (`<b>...</b>`):
image

bwilkerson:
You probably have a better sense of style than I do so I'm happy to go with whatever you prefer, but I would rank them (from best to worst) as italics, bold, message intro, code font.

> I think using an HTML tag also has the advantage of not tempting the markdown parser into formatting the message specially.

Is that actually a problem? I guess I would expect any markdown parser to be able to handle that.

Speaking of messages, many of the messages are templates used to build messages by substitution. We use a `{<integer>}` placeholder. But that looks a bit odd to me in the documentation. Do we want to leave it that way, or should we change it?

src/tools/diagnostics.md
```diff
@@ -0,0 +1,377 @@
...
+
+A type is _potentially non-nullable_ if it's either explicitly non-nullable or
+if it's a type parameter.
+The latter case is included because
+the actual runtime type might be non-nullable.
+
+
+## List of diagnostics
+
+The analyzer produces the following diagnostics when [PENDING...].
```
bwilkerson: The best I've come up with so far is

> The analyzer produces the following diagnostics when the code being analyzed
> does not conform to the language specification or contains other errors that
> will cause it to work in unexpected ways.

but it might be better to just write

> The analyzer produces the following diagnostics.

and leave it at that.

kwalrath: I like the explanation in the first one. I'd tweak it a bit (of course, I'm me :), so I'd say something more like this:

> The analyzer produces the following diagnostics for code that
> doesn't conform to the language specification or that
> contains other errors that might cause it to work in unexpected ways.

Is "errors" correct? Also, we don't know what they expect, so "might" seemed warranted. I was tempted to say "...errors that might cause it to misbehave." But that seemed a little too anthropomorphic.

kwalrath:
> You probably have a better sense of style than I do so I'm happy to go with whatever you prefer, but I would rank them (from best to worst) as italics, bold, message intro, code font.

That's the same as my ranking, but I too don't trust my sense of style.

> > I think using an HTML tag also has the advantage of not tempting the markdown parser into formatting the message specially.
> 
> Is that actually a problem? I guess I would expect any markdown parser to be able to handle that.

I haven't seen it (either Markdown or HTML interpretation) be a problem with the messages, so far. But it'd be good to try them all and see.
 
> Speaking of messages, many of the messages are templates used to build messages by substitution. We use a `{<integer>}` placeholder. But that looks a bit odd to me in the documentation. Do we want to leave it that way, or should we change it?

I actually like having the substitution in there. We could use something in italics instead (e.g. _integer_), which would be more what you might expect in docs. But that might be more trouble than it's worth.

src/tools/diagnostics.md
bwilkerson: > Is "errors" correct?

In the "Diagnostic Design Guidelines" I chose "problem", but I'm not sure what will read best to our users. The word "issue" is another possibility. Any synonym is probably fine.

> ... and we avoid "will" ...

Is it OK in contexts where it's accurate, such as when we're describing the runtime semantics of the language ("doing X will cause Y to happen")? I'll try to curb my use of it elsewhere.

bwilkerson:
> We could use something in italics instead (e.g. integer), which would be more what you might expect in docs. But that might be more trouble than it's worth.

I agree, but I thought I should ask. :-)

I was planning to have the generator automatically insert the message so that it could never get out of sync with what will actually be produced, but the generator can't figure out the semantics of what will be substituted. So having something like

> _The parameter ‘parameterName’ can’t have a value of ‘null’ because ..._

would require us to write that explicitly, making it easier for the docs to get out of sync.

kwalrath:
> would require us to write that explicitly, making it easier for the docs to get out of sync.

Yeah, let's not do that. In sync is better than perfect.

src/tools/diagnostics.md
bwilkerson: How about if we avoid the question altogether by saying

> The analyzer produces the following diagnostics for code that doesn't conform to the language specification or that might work in unexpected ways.

kwalrath: > > ... and we avoid "will" ...
> 
> Is it OK in contexts where it's accurate, such as when we're describing the runtime semantics of the language ("doing X will cause Y to happen")? I'll try to curb my use of it elsewhere.

In that case, I'd still use "causes" instead of "will cause".

kwalrath:
Closing now that we can generate the real file: https://github.com/dart-lang/sdk/blob/master/pkg/analyzer/tool/diagnostics/diagnostics.md.

PR #1655, opened 2019-06-01
Clarify description of assertions

### LINKED ISSUE #1606: Language Tour should explicitly mention asserts do not run the statement
https://dart.dev/guides/language/language-tour#assert

Currently it says "Assert statements have no effect in production code", however because of their function-like usage it is not immediately obvious that means the statement inside the parentheses is not run, e.g. `assert(x++ == 1)` will increment `x` when debugging but won't in production code.

It would be good to explicitly mention this to clarify the behaviour.

kwalrath:
Fixes #1606

Staged:
https://kw-staging-dartlang-2.firebaseapp.com/guides/language/language-tour#default-value
https://kw-staging-dartlang-2.firebaseapp.com/guides/language/language-tour#assert

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -2323,7 +2323,7 @@ That depends on the tools and framework you're using:
 
 * Flutter enables assertions in [debug mode.][Flutter debug mode]
 * Development-only tools such as [dartdevc][]
-  typically support assertions by default.
+  typically enable assertions by default.
 * Some tools, such as [dart][] and [dart2js,][dart2js]
   support assertions through a command-line flag: `--enable-asserts`.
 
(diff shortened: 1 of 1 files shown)


src/_guides/language/language-tour.md
```diff
@@ -2326,6 +2318,18 @@ is true, the assertion succeeds and execution
 continues. If it's false, the assertion fails and an exception (an
 [AssertionError][]) is thrown.
 
+When exactly do assertions work?
+That depends on the tools and framework you're using:
+
+* Flutter enables assertions in [debug mode.][Flutter debug mode]
+* Development-only tools such as [dartdevc][]
+  typically support assertions by default.
```
johnpryan: I can't seem to find what `webdev` uses when it runs the dev compiler, but  I think changing to "typically enable assertions by default" might be more accurate here

PR #1664, opened 2019-06-04
[WIP] Reenable dev builds

### LINKED ISSUE #1663: Fix dev build
More info: #1662

kwalrath:
DO NOT MERGE. We can't fix the dev build without breaking the stable build.

Fixes #1663.

* Fixed issues that broke the old build.
* Updated expected error messages, which broke the build once I fixed the above.
* Updated .travis.yml
* Updated sound-problems

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,9 +23,6 @@ matrix:
   # allow_failures:
     # - env: TASK="./tool/analyze-and-test-examples.sh"
       # dart: stable
-  allow_failures:
-    - env: TASK="./tool/analyze-and-test-examples.sh"
-      dart: dev
   exclude:
     - env: TASK="bundle exec jekyll build"
       dart: dev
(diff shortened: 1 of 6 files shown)


kwalrath:
It turns out that one of our pages actually referred to some of the error messages that changed. I've staged it, and it turns out that one of the old messages isn't showing up at all any more(?!). Maybe you can help me figure it out:

https://kw-staging-dartlang-2.firebaseapp.com/guides/language/sound-problems#a-function-of-type--cant-be-assigned

kwalrath:
Ugh, this might be one of those lovely cases where we're just going to have to let the dev build fail until we're ready to switch over. We'll still need to figure out why the sound-problems page has a blank.

kwalrath:
I think we need to keep allowing the dev builds to fail, since we’re making a fair number of analysis and language changes. #1717 has the rest of the changes from this PR (and then some).

PR #1681, opened 2019-06-14
[WIP] Add doc for analyzer diagnostics

legalcodes:
Staged:
https://jt-dartdev-1.firebaseapp.com/guides/diagnostics

Taken from [the sdk doc](https://github.com/dart-lang/sdk/blob/master/pkg/analyzer/tool/diagnostics/diagnostics.md).


domesticmouse:
Just to quickly check, is this a direct copy from the SDK repo, or do we need to make changes to the generator script?

legalcodes:
I copied this content directly from the sdk repo: https://github.com/dart-lang/sdk/blob/master/pkg/analyzer/tool/diagnostics/diagnostics.md

> do we need to make changes to the generator script?


I'll defer to Brian on that one.

src/_guides/diagnostics.md
```diff
@@ -0,0 +1,298 @@
...
+
+#### Example
+
+The following code generates this diagnostic:
+
+```dart
+var map = <String, int>{'a': 0, 'b': 1, !'c'!};
+```
+
+#### Common fixes
```
sfshaza2: -> Common fix  (make it singular, since there's only one fix)

src/_guides/diagnostics.md
```diff
@@ -0,0 +1,298 @@
...
+
+If the expression is intended to compute either a key or a value in an
+entry, fix the issue by completing the code:
```
sfshaza2: by completing the code:  =>  by replacing the expression with the key or the value. For example:

src/_guides/diagnostics.md
```diff
@@ -0,0 +1,298 @@
...
+#### Example
+
+The following code generates this diagnostic:
+
+```dart
+var m = <String, int>{'a': 0, 'b': 1};
+var s = <String>{...m};
+```
+
+#### Common fixes
```
sfshaza2: Common fixes => Common fix

src/_guides/diagnostics.md
```diff
@@ -0,0 +1,298 @@
...
+#### Description
+
+The analyzer produces this diagnostic when a set literal is found in code
+that has an SDK constraint whose lower bound is less than 2.2. Set literals
+were not supported in earlier versions, so this code won't be able to run
+against earlier versions of the SDK.
+
+#### Example
+
+In a package that defines SDK constraints in the `pubspec.yaml` file that
```
sfshaza2: defines SDK constraints in the `pubspec.yaml` file that have => 
defines the SDK constraint (in the `pubspec.yaml` file), with a lower bound of less than 2.2.

mit-mit:
I'm confused about what the checks in https://github.com/dart-lang/sdk/blob/master/pkg/analyzer/tool/diagnostics/diagnostics.md are, and how they are different from other analyzer messages, and lints. @bwilkerson can you clarify?

src/_guides/diagnostics.md
```diff
@@ -0,0 +1,298 @@
...
+```
+
+The list `b` can only be spread into a set, and the maps `a` and `c` can
+only be spread into a map, and the literal can't be both.
+
+#### Common fixes
+
+There are two common ways to fix this problem. The first is to remove all
+of the spread elements of one kind or the other, so that the elements are
+consistent. In this case, that likely means removing the list (and
```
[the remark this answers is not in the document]
bwilkerson: Done

src/_guides/diagnostics.md
```diff
@@ -0,0 +1,298 @@
...
+
+```dart
+union(Map<String, String> a, List<String> b, Map<String, String> c) =>
+    {...a, for (String s in b) s: s, ...c};
+```
+
+### ambiguous\_set\_or\_map\_literal\_either
+
+_This literal must be either a map or a set, but the elements don't have enough
+type information for type inference to work._
```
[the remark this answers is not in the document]
bwilkerson: Done

src/_guides/diagnostics.md
```diff
@@ -0,0 +1,298 @@
...
+
+#### Description
+
+Because map and set literals use the same delimiters (`‘{` and `}`), the
```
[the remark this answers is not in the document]
bwilkerson: Done

src/_guides/diagnostics.md
```diff
@@ -0,0 +1,298 @@
...
+analyzer looks at the type arguments and the elements to determine which
+kind of literal you meant. When there are no type arguments and all of the
+elements are spread elements (which are allowed in both kinds of literals)
+then the analyzer uses the types of the expressions that are being spread
+to decide. If all of the expressions have the type `Iterable`, then it's a
```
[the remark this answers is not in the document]
bwilkerson: Done

src/_guides/diagnostics.md
```diff
@@ -0,0 +1,298 @@
...
+set literal, if they all have the type `Map`, then it's a map literal.
```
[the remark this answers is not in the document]
bwilkerson: Done

src/_guides/diagnostics.md
```diff
@@ -0,0 +1,298 @@
...
+
+This diagnostic is produced when none of the expressions being spread has a
```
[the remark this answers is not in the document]
bwilkerson: Done

src/_guides/diagnostics.md
```diff
@@ -0,0 +1,298 @@
...
+```dart
+void f(C c) {
+  c.!m!();
+}
+```
+
+#### Common fixes
+
+The documentation for declarations that are annotated with `@deprecated`
+should have documentation to indicate what code to use in place of the
```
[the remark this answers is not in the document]
bwilkerson: Done

src/_guides/diagnostics.md
bwilkerson: Done

src/_guides/diagnostics.md
bwilkerson: Done

src/_guides/diagnostics.md
bwilkerson: Done

src/_guides/diagnostics.md
bwilkerson: Done

bwilkerson:
> I'm confused about what the checks in https://github.com/dart-lang/sdk/blob/master/pkg/analyzer/tool/diagnostics/diagnostics.md are, and how they are different from other analyzer messages, and lints. @bwilkerson can you clarify?

The file `diagnostics.md` doesn't define any diagnostics, it contains documentation intended to help users understand the existing diagnostics (errors, warnings, hints and lints) that analyzer already produces. As you know, sometimes users need more information than a single line of text can provide, and this is intended to be one way to get that information to our users.

The analysis server is now sending a URL as part of each diagnostic sent to the client (when a URL is available). In IntelliJ that URL is now being displayed as a clickable link that will open a browser on the docs. I haven't checked to see whether the same is being done in VSCode, but Danny is aware of this work and has plans to add support there as well (if it isn't already being done). I believe we intend to update DartPad, and we should update the command-line tool to display these URLs as well.

So far we have URLs for all of the lints that point to the pre-existing docs for them. (We might want to integrate the lint docs into this generated file, but that's lower priority in my mind than adding docs for diagnostics that don't yet have any.) This is the first set of non-lints that are being documented. After this file is published I can go back and add the URLs to analyzer so that they will also be sent to clients.

bwilkerson:
The CL to apply the requested changes is here: https://dart-review.googlesource.com/c/sdk/+/106120.

legalcodes:
@bwilkerson I've removed diagnostics from the side-nav

legalcodes:
@bwilkerson I have updated this doc w/ the latest changes from the sdk doc

bwilkerson:
This LGTM, but I'd be interested in hearing what @sfshaza2 thinks.

Still curious: Is there anything that needs to be done in order to allow us to gather analytics on the number of times each URL is followed to see documentation? That would be valuable information to help us understand when the messages need to be improved.

sfshaza2:
I haven't looked yet, but I will.

We have google analytics enabled for both flutter.dev and api.flutter.dev.

src/_guides/diagnostics.md
```diff
@@ -0,0 +1,298 @@
...
+{% endcomment -%}
+
+This page lists diagnostic messages produced by the Dart analyzer,
+with details about what those messages mean and how you can fix your code.
+For more information about the analyzer, see
+[Customizing static analysis](/guides/language/analysis-options).
+
+## Diagnostics
+
+The analyzer produces the following diagnostics for code that
```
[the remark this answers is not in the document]
sfshaza2: LOL, @bwilkerson.

src/_guides/diagnostics.md
```diff
@@ -0,0 +1,298 @@
...
+doesn't conform to the language specification or
+that might work in unexpected ways.
+
+### ambiguous\_set\_or\_map\_literal\_both
```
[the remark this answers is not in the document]
sfshaza2: Jon thought they were needed but they aren't. Could you remove them? (Makes searching harder, for example.)

src/_guides/diagnostics.md
```diff
@@ -0,0 +1,298 @@
...
+
+_This literal contains both 'Map' and 'Iterable' spreads, which makes it
```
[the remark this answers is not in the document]
sfshaza2: Oh! OK. I withdraw my objection.  :)

src/_guides/diagnostics.md
```diff
@@ -0,0 +1,298 @@
...
+    {...a, ...b, ...c};
+```
+
+The list `b` can only be spread into a set, and the maps `a` and `c` can
+only be spread into a map, and the literal can't be both.
+
+#### Common fixes
+
+There are two common ways to fix this problem. The first is to remove all
+of the spread elements of one kind or the other, so that the elements are
```
sfshaza2: or the other => or another

src/_guides/diagnostics.md
```diff
@@ -0,0 +1,298 @@
...
+
+The analyzer produces this diagnostic when a set literal is found in code
+that has an SDK constraint whose lower bound is less than 2.2. Set literals
+were not supported in earlier versions, so this code won't be able to run
+against earlier versions of the SDK.
+
+#### Example
+
+In a package that defines the SDK constraint (in the pubspec.yaml file),
+with a lower bound of less than 2.2.:
```
sfshaza2: This one's probably on me, but:

of less than 2.2.: => of less than 2.2. For example:

legalcodes: Done

src/_guides/diagnostics.md
legalcodes: Done

legalcodes:
I've removed the escape characters @sfshaza2

bwilkerson:
I have sent out a CL with the changes to the analyzer source and a re-generated file (https://dart-review.googlesource.com/c/sdk/+/106200). We should check to make sure that the generated file contains all the changes you made.

domesticmouse:
There was one missed fix in the PR. I've added it as a late review to your above CL. Thanks!

mit-mit:
> , it contains documentation intended to help users understand the existing diagnostics (errors, warnings, hints and lints) that analyzer already produces

That sounds very useful, thanks. But it wasn't super clear from the intro paragraph at the top of the page.

Also, would like to see this cross linked from the other Analyzer pages.

Perhaps @kwalrath can fine tune this a bit?

bwilkerson:
> But it wasn't super clear from the intro paragraph at the top of the page.

I'm happy to see it improved. Do you have suggestions for how to improve the introductory paragraph?

Also, my expectation is that users will arrive at the specific documentation they need by clicking on a link from their IDE or the command-line, and that almost nobody will want to just read the page. Hence, I'm not sure that (a) it will be confusing to users because of the context by which they arrived at the page or (b) that anyone will see the introductory paragraph. Of course, neither of those are intended to imply that I think it shouldn't be improved.

> Also, would like to see this cross linked from the other Analyzer pages.

I'd like to hold off on that until we have figured out a page structure that allows us to gather the analytics we need.

We might also want to wait until there's enough content here to make it more useful; we currently have documentation for about 1% of the diagnostics.

mit-mit:
> my expectation is that users will arrive at the specific documentation they need by clicking on a link from their IDE or the command-line

Ah. If so, I wonder if we should 1) keep the page as-is, but 2) remove the leftnav entry for it? cc @kwalrath

legalcodes:
We removed the leftnav entry for diagnostics before landing the PR.

mit-mit:
👍, everything is great, I'll be quiet now :-)

sfshaza2:
Also, another comment from the peanut gallery. This doc doesn't belong in
_guides. It's not a guide, it's a reference. Could you move the file?

legalcodes:
Shams and I just sat down at the dart.dev site together to come up with the
best place for this page, and it's hard to find a better spot than in Tools
> Static Analysis >. Given that it won't be a page that we expect people to
be looking for while browsing on dart.dev, we're fine with the current URL
for now. I agree that /guides/diagnostics feels just a little off, so if
anyone can recommend a better place, we're all for it. I just haven't found
a better spot yet.

Jon

legalcodes:
And if it wasn't clear from my last email, it just so happens that Tools >
Static Analysis > maps to /guides/ -- hence the /guides/diagnostics URL
we're currently using.

kwalrath:
About the language, I’m starting to agree with Shams that _diagnostics_ isn’t as clear as _diagnostic messages_. I think the meaning of _diagnostic(s)_ is broader than just messages, so it’s good to clarify.

A [Google search for diagnostics](https://www.google.com/search?q=diagnostics) comes up with _diagnostic_, which has the expanded definition of “a program or routine that helps a user to identify errors.”

Google docs don’t typically use “diagnostic(s)” as a noun, except in API names, where they often represent a collection of diagnostic information. The docs usually use “diagnostic(s)” as an adjective modifying “data”, “information”, “output”, “resources”, or “exception”. (See [diagnostics](https://developer.android.com/s/results/?q=diagnostics) in Android docs and [diagnostics](https://developers.google.com/s/results/?q=diagnostics&p=%2F)/[diagnostic](https://developers.google.com/s/results/?q=diagnostic&p=%2F) in other Google docs.

Finally, the [wikipedia page for error message](https://en.wikipedia.org/wiki/Error_message) doesn’t have the string “diagnos” at all.

bwilkerson:
> I’m starting to agree with Shams that _diagnostics_ isn’t as clear as _diagnostic messages_.

And yet, as I pointed out before https://www.dictionary.com/browse/diagnostic defines "diagnostic" exactly the way I'm using it. Also, see https://microsoft.github.io/language-server-protocol/specification which also uses "diagnostic" in the same way. I think it's the way the vocabulary is going, not necessarily where it's been in the past, which might explain why search isn't finding many uses of this definition.

> I think the meaning of _diagnostic(s)_ is broader than just messages ...

I agree. Each diagnostic has a message associated with it, but a diagnostic is bigger than the message.

I suppose it's kind of a fine line, but I think of what we're doing as explaining the diagnostics (why they are produced and what can be done in response to them) rather than what the messages mean. We happen to be including the messages in order to make it easier for users to know that they're looking at the right docs, but that doesn't mean that it's the message that's being documented.

kwalrath:
> > I’m starting to agree with Shams that _diagnostics_ isn’t as clear as _diagnostic messages_.
> 
> …Also, see https://microsoft.github.io/language-server-protocol/specification which also uses "diagnostic" in the same way.

You could argue that that’s in the context of the Diagnostic interface, so it fits the API name case I mentioned. In general, that page seems to use the jargon that tool authors use, not that ordinary programmers use. Our page is aimed at the latter.

> I think it's the way the vocabulary is going, not necessarily where it's been in the past, which might explain why search isn't finding many uses of this definition.

Dictionaries aside, I conducted my usage search in Google tech docs, which _do_ set the precedent that we should follow, if possible.

> > I think the meaning of _diagnostic(s)_ is broader than just messages ...
> 
> I agree. Each diagnostic has a message associated with it, but a diagnostic is bigger than the message.
> 
> I suppose it's kind of a fine line, but I think of what we're doing as explaining the diagnostics (why they are produced and what can be done in response to them) rather than what the messages mean. We happen to be including the messages in order to make it easier for users to know that they're looking at the right docs, but that doesn't mean that it's the message that's being documented.

I still feel like “Diagnostics” as a page title is not as helpful as it could be. When we add it to the sidenav, people won’t know what it really means — diagnostic tools? tips for diagnosing issues? a complete list of diagnostic messages? A more helpful title (though probably too long) might be something like ”Diagnosing and fixing analysis issues”.

Within the page we should still err on the side of clarity, since people might never see the top of the page or the glossary.

bwilkerson:
> I still feel like “Diagnostics” as a page title is not as helpful as it could be. When we add it to the sidenav, people won’t know what it really means ...

That's true; I wasn't thinking about the title in isolation. I would be good to scope the page to the analyzer (the CFE doesn't produce the same diagnostics). Perhaps something as short as "Analyzer Diagnostics" or the slightly longer "Analyzer Diagnostic Messages".

> Within the page we should still err on the side of clarity, since people might never see the top of the page or the glossary.

I agree. At the end of the day what matters is whether these docs bring value to users.

I would have expected that use of "diagnostic" would have been fairly clear from context ("The analyzer produces this diagnostic when ..."), but you know better than I.

bwilkerson:
> When we add it to the sidenav, people won’t know what it really means ...

Are we planning on adding this to the sidenav? My working hypothesis was that users would only want to find the page as a result of traversing one of the URLs that we return. This really doesn't feel like the kind of page anyone would want to read unless they were looking for information about a specific diagnostic.

PR #1683, opened 2019-06-14
Run Dart 2.3.2 dartfmt

mit-mit:


sfshaza2:
Of course, this PR primarily looks fine. The only possible issue, is that we often run dartfmt with a maximum line length of 70 (or so) to avoid having the example code wrap when displayed on the site. Are we sure that this was not going on here?

Otherwise, LGTM.  :)

mit-mit:
I ran it with the default line length. Is that not what we usual do?

sfshaza2:
You'd have to check with Kathy for recent info, but we used to run it with
a default line length of 70, otherwise the code often wrapped in those code
blocks. But things may have changed.

johnpryan:
Is this still necessary now that we've upgraded to 2.4?

mit-mit:
Yes, see: https://travis-ci.org/dart-lang/site-www/jobs/551280797

But I think we can wait until @kwalrath is back to answer the line length question

sfshaza2:
Agreed. This isn't urgent..

kwalrath:
We don’t run dartfmt directly… instead, we use `tool/dartfmt.sh`. When I run that, I see only one diff. I’m looking into it now…

kwalrath:
Only one file needs to change due to dartfmt. I’ve fixed that in #1717.

PR #1685, opened 2019-06-15
Fix links

mit-mit:


mit-mit:
TBR @kwalrath 

I'm curious why the build passed on these? https://travis-ci.org/dart-lang/site-www/builds/546092678

kwalrath:
We don’t fail the build on external links (unlike flutter.dev, which fails all the time when an external page goes bad).

What we should do is have a test that checks external links but is allowed to fail. That way we’ll catch these kinds of things without making our site more vulnerable to external factors.

PR #1686, opened 2019-06-15
Fix another core lib link

mit-mit:


mit-mit:
TBR @kwalrath 

This was also broken. Is the link checked actually working?

PR #1688, opened 2019-06-16
Fix content-container on mobile breakpoint (479px)

Arhell:
#1689


johnpryan:
Ditto - I've included this change in #1692 - Could you go there and sign the CLA? Thanks for filing an issue and providing a fix!

PR #1690, opened 2019-06-17
Updating banner for 2.4.0

sfshaza2:
Staged:  https://sz-www-1.firebaseapp.com/guides

PR #1693, opened 2019-06-17
Revert "Updating banner for 2.4.0"

mit-mit:
Reverts dart-lang/site-www#1690


mit-mit:
Reverting, 2.4 is not out yet

PR #1701, opened 2019-06-25
Update sound-dart.md

MountainHill:


johnpryan:
Thanks for the pull request - the previous section defines *soundness* and this section is meant to discuss the benefits.

Changing to "sound" wouldn't be grammatically correct in this case.

PR #1703, opened 2019-06-26
use 960px max width for all content

johnpryan:
Adding a limit to the width of content is more readable for people with larger screen resolutions. Here's a before and after:

Screen Shot 2019-06-26 at 12 04 02 PM

Screen Shot 2019-06-26 at 12 03 34 PM

### DIFF
--- a/src/_assets/css/main.scss
+++ b/src/_assets/css/main.scss
@@ -250,8 +250,9 @@ img {
 
   > article {
     > .content {
-      padding-left: 5%;
-      padding-right: 8%;
+      max-width: 960px;
+      margin-left: auto;
+      margin-right: auto;
       padding-top: $top-content-padding;
     }
 


src/_assets/css/main.scss
```diff
@@ -250,8 +250,9 @@ img {
 
   > article {
     > .content {
-      padding-left: 5%;
-      padding-right: 8%;
+      max-width: 960px;
```
sfshaza2: My concern about this approach, John, is what about mobile?

johnpryan:
Deployed this to firebase here: https://ryjohn-test.firebaseapp.com/guides/libraries/library-tour

johnpryan:
here's some screenshots from Chrome's mobile device emulator:

Screen Shot 2019-07-02 at 1 56 36 PM
Screen Shot 2019-07-02 at 1 56 57 PM

PR #1704, opened 2019-06-27
Roll Dart to 2.4

mit-mit:


mit-mit:
Merging; travis issues are related to formatting checks, which has a different PR for when Kathy is back

PR #1706, opened 2019-06-27
(do not merge) use dart stable branch for checking examples

johnpryan:
This is an attempt to pin down the dart version to fix [failures like these](https://travis-ci.org/dart-lang/site-www/jobs/551282105)


johnpryan:
Closing - build failures will be fixed at a later point https://github.com/dart-lang/site-www/pull/1704#issuecomment-506501506

PR #1715, opened 2019-07-01
fix syntax in lint excludes

pq:
/cc @kwalrath @sfshaza2 

fyi @danrubel

### DIFF
--- a/src/_guides/language/analysis-options.md
+++ b/src/_guides/language/analysis-options.md
@@ -240,14 +240,16 @@ include: package:pedantic/analysis_options.yaml
 linter:
   rules:
     avoid_shadowing_type_parameters: false
-    - await_only_futures
+    await_only_futures: true
 {% endprettify %}
 
 <aside class="alert alert-info" markdown="1">
-  The reason you don't include an initial dash (`- `)
-  when you specify a value (`: false`) is that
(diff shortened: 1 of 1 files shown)


kwalrath:
It sounds like this should be a warning, not a note, and we should make the warning harder to miss. How about something like this:

```
<aside class="alert alert-warning” markdown="1">
  If you disable any rules, then use key-value syntax to enable rules.
  For example, the preceding example can’t use `- await_only_futures`;
  instead, it must use `await_only_futures: true`.
  The reason it can’t use the dash shortcut (`-`) to enable `await_only_futures`
  is that YAML doesn't support mixing list and key-value syntaxes.
  </aside>
```

pq:
Thanks @kwalrath.  Updates made!

kwalrath:
Thanks, @pq!

pq:
Cheers @kwalrath.  Thanks for the help word-smithing. 😄   Happy to do it.

Should I ignore the (unrelated) build breakage?

kwalrath:
> Should I ignore the (unrelated) build breakage?

Yes, ignore it. I’m following up on that.

kwalrath:
The build is fixed, woohoo!

PR #1718, opened 2019-07-02
matching versions in text and code example

joriki:
I don't see a reason for using different versions in the text and in the following code example -- if you know your package works with version `2.5.7`, then that's what you'd write in the dependencies.


joriki:
I signed it.

PR #1719, opened 2019-07-02
Update .travis.yml to run linker checker on all builds (failures allowed)

mit-mit:


.travis.yml
```diff
@@ -28,9 +29,13 @@ matrix:
       dart: dev
     - env: TASK="./tool/check-code.sh"
       dart: dev
+    - env: TASK="bundle exec jekyll build;./tool/shared/check-links.sh"
```
kwalrath: Should this be `check-links.sh -e`?

mit-mit: That's for checking external paths. What I'm doing here is ensuring that inside this matrix job the site is actually generated first. But I got the syntax wrong.

mit-mit:
After (too) much experimentation, this appears to be working:
https://travis-ci.org/dart-lang/site-www/jobs/553694124

The failure in other jobs in the matrix is unrelated and is also happening in head.

I'd like to get the link checker to fail the check job on warnings, but need a change in shared for that:
https://github.com/dart-lang/site-shared/pull/101

mit-mit:
@kevmoo can you help with reviewing this?

.travis.yml
```diff
@@ -14,23 +13,28 @@ env:
...
+    - env: TASK="bundle exec jekyll build"
+      dart: stable
+    - env: TASK="./tool/check-code.sh"
+      dart: stable
+    - env: TASK="./tool/check-code.sh"
+      dart: dev
+    - env: TASK="./tool/analyze-and-test-examples.sh"
+      dart: stable
+    - env: TASK="./tool/analyze-and-test-examples.sh"
+      dart: dev
```
johnpryan: should this be `stable`?

kwalrath:
Thanks, @johnpryan!

PR #1726, opened 2019-07-03
Update get-dart.md

mit-mit:
The 'This site uses Dart xxx' message seems relevant on https://dart.dev/tools/sdk, but I find it confusing on the download page (especially when we are in a period when get-dart has a more recent build than the site has). Would you be OK removing it?


kwalrath:
fyi, I think the new pedantic version is what’s causing the build failure.

kwalrath:
> The 'This site uses Dart xxx' message seems relevant on https://dart.dev/tools/sdk, but I find it confusing on the download page (especially when we are in a period when get-dart has a more recent build than the site has). Would you be OK removing it?

I’m afraid that if we remove it, people will download the dev version without realizing that it’s ahead of what the site covers. Perhaps we could be more explicit about the site covering the stable version, so if they get the dev version, our docs don’t cover the latest features. We could also point to the release notes…

mit-mit:
I see. I think that is a very valid concern, but I also think that the info box mentioning what version the site happens to use is a suboptimal way of conveying that point.

I tried to update this PR by moving the section about which release type we have up. WDYT?

kwalrath:
Revisiting this... I think it's fine to just remove that note. So commit #1 👍. Commit #2 👎. (I really don't want to add a section above the installation instructions.)

PR #1727, opened 2019-07-03
Fix the build

kwalrath:
The new version of pedantic includes prefer_iterable_wheretype, so it catches an intentional occurrence of that bad pattern in usage_bad.dart.


kwalrath:
It works! I’m going to submit this TBR.

PR #1728, opened 2019-07-04
Fix or skip links

kwalrath:
These changes should get rid of all the warnings and errors that reliably occur.


tool/config/linkcheck-skip-list.txt
```diff
@@ -3,3 +3,11 @@
 # FIXME(Temporary): linkcheck seems to be intermittently failing when accessing main.css
 # https://github.com/dart-lang/site-www/issues/1107
 /assets/main(-[0-9a-f]+)?.css
+https://localhost:4040
```
mit-mit: Do we want to use a wildcard to skip all localhost links?

kwalrath: I don’t think so, because the link checker uses localhost when it checks links.

PR #1729, opened 2019-07-04
Improve banner legibility

kwalrath:
It’s a small change that makes all pages easier to read, so I’m committing TBR.

### DIFF
--- a/src/_assets/css/_dash.scss
+++ b/src/_assets/css/_dash.scss
@@ -198,6 +198,9 @@ $dash-header-bg: #17212f;
   .banner__text {
     color: white;
   }
+  a {
+    color: #40c4ff;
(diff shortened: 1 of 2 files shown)


src/_assets/css/_dash.scss
```diff
@@ -198,6 +198,9 @@ $dash-header-bg: #17212f;
   .banner__text {
     color: white;
   }
+  a {
+    color: #40c4ff;
```
johnpryan: (nitpick) I suspect there's a variable with color defined somewhere

kwalrath: I looked for this color and couldn’t find it, but yeah, it should be a variable.

PR #1731, opened 2019-07-04
Configure linkchecker to fail on warnings

mit-mit:


mit-mit:
Looks like it's working: https://travis-ci.org/dart-lang/site-www/builds/554148257

PR #1737, opened 2019-07-07
Update side-nav.yml

devoncarew:
- reorder the `Editors & debuggers` section of the sitenav

### DIFF
--- a/src/_data/side-nav.yml
+++ b/src/_data/side-nav.yml
@@ -148,14 +148,14 @@
     permalink: /tools
   - title: Editors & debuggers
     children:
-      - title: Dart DevTools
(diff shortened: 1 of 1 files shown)


devoncarew:
Thanks! Looks like I don't have permission to merge myself.

johnpryan:
I'm not sure why that is - FYI @kwalrath

kwalrath:
> Thanks! Looks like I don't have permission to merge myself.

That’s weird. Everyone in Dart Core Eng has Write permission, which I’d think would be sufficient.

devoncarew:
There may be an additional check in the admin page? It was a 'can't merge' message that I hadn't seen before.

PR #1739, opened 2019-07-08
use the same variable for selected navbar link and banner link

johnpryan:
This variable is currently used for links in cards and has the value that we need (#40C4FF)


kwalrath:
Thanks!

PR #1743, opened 2019-07-09
Update highlight in design.md

bencalabrese:
Updates the highlight in the "good" example given for "PREFER signatures in function type annotations." Previously, the good example did not highlight the return type in the function signature which is an important part of the change from the "bad" example.

### DIFF
--- a/src/_guides/language/effective-dart/design.md
+++ b/src/_guides/language/effective-dart/design.md
@@ -1322,7 +1322,7 @@ function.
 {:.good-style}
 <?code-excerpt "misc/lib/effective_dart/design_good.dart (avoid-Function)" replace="/(void )?Function(\(.*?\))?/[!$&!]/g"?>
 {% prettify dart %}
-bool isValid(String value, bool [!Function(String)!] test) => ...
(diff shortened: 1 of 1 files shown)


bencalabrese:
Apologies, it's not clear to me why this test is failing. I assume this issue is with the code example not being updated as described here: https://github.com/dart-lang/site-www/blob/master/CONTRIBUTING.md#updating-code-samples

Looking at the offending line though, I'm not sure what I would need to adjust in order to accommodate this change: https://github.com/dart-lang/site-www/blob/master/examples/misc/lib/effective_dart/design_good.dart#L215

I suspect I'm missing something though :-)

kwalrath:
Thanks for the PR! I think the regex above the sample needs to change to include the “bool “ before “Function”. (I don’t know why the “void” is in the regex.) Once the regex is correct, we might not need to change the example source code.

If you don’t have a full build of site-www, I can work on this change for you.

bencalabrese:
Thanks, kwalrath@! I don't have a full build, so your help would be much appreciated. Thanks for looking into it!

Issue #1744, opened 2019-07-10
Support code highlighting in markdown code blocks


kwalrath:
I’d love to be able to use only triple-tick dart, never `{% prettify …%}`. But at least one use case (highlighting: `[!…!]`) doesn’t work with triple-tick dart. Having to remember to use prettify causes unnecessary complexity for site contributors. Also, I think prettify styles are different from ```dart styles.

Second, it’d be good to improve the formatting of code within strings: #1108.

chalin:
It is possible to use triple-tick markdown and get highlighting to work -- that what we did in `site-webdev`. AFAIR, most of the necessary infrastructure has already been moved into `site-shared`.

chalin:
(Maybe the title of this issue should be updated to something like: _Support code highlighting in markdown code blocks_.)

parlough:
I want to standardize on the markdown code blocks, so I'll look into this soon.

parlough:
I'm going to close this in favor of tracking the work in https://github.com/dart-lang/site-www/issues/3846.

PR #1745, opened 2019-07-10
[DO NOT COMMIT] Fix highlights in the diagnostics page

kwalrath:
The good news: I got the highlights working, and you can see them at https://kw-www-dartlang-1.firebaseapp.com/guides/diagnostics.

The bad news: to get the highlights working, I [had to change](https://github.com/dart-lang/site-www/commit/6afcc3891a415556d4c1cc1aa7f9e77cf5144d01) ```dart to `{% prettify dart %}`. Yuck. This is a known problem, and I’ve created #1744 to track it.

Other issues to ponder:
* We might want to use a different URL.
* The right-hand TOC is too narrow, so the entries for ambiguous_set_or_map_literal_both &
ambiguous_set_or_map_literal_either look the same.

/cc @domesticmouse @legalcodes @bwilkerson @johnpryan

### DIFF
--- a/src/_guides/diagnostics.md
+++ b/src/_guides/diagnostics.md
@@ -44,7 +44,7 @@ The following code produces this diagnostic:
 
 ```dart
 union(Map<String, String> a, List<String> b, Map<String, String> c) =>
-    {...a, ...b, ...c};
+    [!{...a, ...b, ...c}!];
 ```
 
 The list `b` can only be spread into a set, and the maps `a` and `c` can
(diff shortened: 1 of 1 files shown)


bwilkerson:
> The good news: I got the highlights working ...

Yeah!

> The bad news: to get the highlights working ...

Not as bad as it might sound. The markdown file is generated, so we could automatically convert from one form to the other if need be.

> We might want to use a different URL.

I'll hold off on sending the URLs from server (or printing them from the command-line) until we've decided. After we start producing URLs we'll need to have a process in place to change them; otherwise users using older versions of the SDK will get obsolete URLs, and that wouldn't be a good UX.

> The right-hand TOC is too narrow ...

Yep. We noticed that last time too.

kwalrath:
Closing in favor of #1790.

PR #1747, opened 2019-07-10
set font weight to 400 for code blocks and in-line code

johnpryan:
before and after: 

Screen Shot 2019-07-10 at 9 23 16 AM
Screen Shot 2019-07-10 at 9 22 47 AM


kwalrath:
I really like this change for the code snippets. I’m not as sure about the inline code. Did we get feedback that that, too, was hard to read?

johnpryan:
no, I'll take that change out

PR #1765, opened 2019-07-15
Note that only hosted dependencies are permitted.

jonasfj:
This was announced through breaking change process in:
https://github.com/dart-lang/sdk/issues/36765

It serves to ensure that dependencies of a package cannot
become unavailable in the future. Thus, if you can `pub get`
something once, then you can `pub get` the exact same thing
if you retain the `pubspec.lock`.

Existing packages with git dependencies remain on `pub.dev`,
but new packages containing git dependencies will not be
permitted.

### DIFF
--- a/src/tools/pub/publishing.md
+++ b/src/tools/pub/publishing.md
@@ -36,10 +36,10 @@ are a few additional requirements for uploading a package:
   it's too large, consider splitting it into multiple packages, or cutting down
   on the number of included resources or examples.
 
-* Your package should only have hosted dependencies. Git dependencies are
-  allowed but strongly discouraged; not everyone using Dart has Git installed,
-  and Git dependencies don't support version resolution as well as hosted
-  dependencies do.
+* Your package should depend only on hosted dependencies (from the default pub
+  package server) and SDK dependencies (`sdk: flutter`). These restrictions
+  ensure that dependencies of your packages cannot become unavailable in the
+  future.
 
 Be aware that the email address associated with your Google account is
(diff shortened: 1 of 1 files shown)


src/tools/pub/publishing.md
```diff
@@ -36,10 +36,9 @@ are a few additional requirements for uploading a package:
   it's too large, consider splitting it into multiple packages, or cutting down
   on the number of included resources or examples.
 
-* Your package should only have hosted dependencies. Git dependencies are
-  allowed but strongly discouraged; not everyone using Dart has Git installed,
-  and Git dependencies don't support version resolution as well as hosted
-  dependencies do.
+* Your package should only have hosted dependencies from the default pub package
+  repository, and SDK dependencies (ie. `sdk: flutter`). This ensures that
```
kwalrath: We avoid i.e. in docs, since people often mix up i.e. and e.g. (https://developers.google.com/style/word-list#letter-I). Also, we avoid unadorned “This”, since it can be unclear (https://developers.google.com/style/word-list#letter-T). How about:

Your package should depend only on hosted dependencies (from the default pub package server) and SDK dependencies (`sdk: flutter`). These restrictions ensure that…

kwalrath:
Thanks, @jonasfj! I missed this change.

src/tools/pub/publishing.md
jonasfj: Much better! :)

I'll update the PR tomorrow, thanks.

kwalrath: Cool, thanks!

PR #1769, opened 2019-07-16
Fix bad link

### LINKED ISSUE #1768: /resources/books leads to 404
On the https://dart.dev/resources/books page the link to the "Dart: Scalable Application Development" book leads to a 404 page.

Not sure but looks like the correct link is https://www.oreilly.com/library/view/dart-scalable-application/9781787288027/

.. at least that is what I got when I googled the title.

kwalrath:
Fixes #1768.


kwalrath:
Committing TBR.

PR #1771, opened 2019-07-18
Remove outdated comment about initializing formals.

lrhn:
You *can* refer to the name of an initializing formal in the initializer list.


munificent:
Bad timing! I just sent out #1772 which addresses this (and other issues).

PR #1772, opened 2019-07-18
Fix a number of "Effective Dart" issues.

### LINKED ISSUE #947: Add "DON'T use a leading underscore for local variables or parameters."
Privacy only applies to top-level declarations and members, so it's misleading to see a "private" name for a local.

### LINKED ISSUE #967: Add exception for rule "AVOID lines longer than 80 characters" on multilines strings
@munificent

### LINKED ISSUE #1393: Make uses of "exception" bold in Effective Dart?
I occasionally refer back to articles in Effective Dart to check if there's an exception for a given case. This would make them easier to find, particularly when preceded by other paragraphs explaining the reasoning.

### LINKED ISSUE #1476: Recommend "?? false" and "?? true"
There are a few places in the language where "Boolean conversion" happens — where you use a value to determine whether or not to execute some code: if conditions, while conditions, operands to `&&` and `||`, etc. That conversion throws an exception if the value is null. Unlike other scripting languages, we don't treat null as "falsey".

That means when users do have an expression that could evaluate to null, they need to wrap it in something that turns it into true or false, depending on what they want it to represent. There are a couple of common idioms:

```dart
if (nullableCondition ?? false) // null is falsey.
if (nullableCondition ?? true) // null is truthy.

if (nullableCondition == true) // null is falsey.
if (nullableCondition == false) // null is truthy.

if (nullableCondition != false) // null is falsey.
if (nullableCondition != true) // null is truthy.
```

All are somewhat common. I think "Effective Dart" should encourage the first two. My arguments are:

* The `??` operator makes it clear that there is something to do with *null* going on. The others could be comparing objects of other arbitrary types.

* It's actually pretty confusing to mentally work through the Boolean logic to figure out what the other examples even mean. The fact that `== true` mentions "true" but means "treat null as *false*" is particularly opaque. But the fact that `!= true` means that you have to do the negation in your head too is also confusing.

### LINKED ISSUE #1502: Update lint association for type annotations on APIs advice.
Page URL: https://www.dartlang.org/guides/language/effective-dart/design
Page source: https://github.com/dart-lang/site-www/tree/master/src/_guides/language/effective-dart/design.md

 **PREFER type annotating public fields and top-level variables if the type isn’t obvious.** links to `prefer_typing_uninitialized_variables` when it should probably better point to: 

http://dart-lang.github.io/linter/lints/type_annotate_public_apis.html 

See:

image

/cc @kwalrath

### LINKED ISSUE #1575: 'Effective Dart: Style' page need to be updated for Unified Collections
Page URL: https://www.dartlang.org/guides/language/effective-dart/style
Page source: https://github.com/dart-lang/site-www/tree/master/src/_guides/language/effective-dart/style.md

Description of issue:

[Unified Collections](https://github.com/dart-lang/language/blob/master/accepted/future-releases/unified-collections/feature-specification.md) adds Control Flow Collections to the language. The rule "DO use curly braces for all flow control structures" needs to be updated to mention this syntax, since it also does not use curly braces.

### LINKED ISSUE #1597: Recommend mixin keyword after 2.1?
https://dart.dev/guides/language/effective-dart/design#do-document-if-your-class-supports-being-used-as-a-mixin

Should recommend mixin keyword after 2.1?

### LINKED ISSUE #1725: 'Effective Dart: Usage' page issue
Page URL: https://www.dartlang.org/guides/language/effective-dart/usage
Page source: https://github.com/dart-lang/site-www/tree/master/src/_guides/language/effective-dart/usage.md

Description of issue:

This part of the "DO use initializing formals when possible" style rule:

"You can't always take advantage of it. In particular, using it means
the parameter is not visible in the initialization list. But, when you can, you
should." 

is not valid since [Dec 2016].(https://github.com/dart-lang/sdk/blob/master/CHANGELOG.md#1210---2016-12-07) when this changed was applied:

"Support access to initializing formals, e.g., the use of x to initialize y in class C { var x, y; C(this.x): y = x; }."

Perhaps it should be removed?

munificent:
This doesn't address all of them, but it covers most of the easy and
non-contentious ones.

Fix #947. Fix #967. Fix #1393. Fix #1476. Fix #1502. Fix #1575.
Fix #1597. Fix #1725.

### DIFF
--- a/src/_guides/language/effective-dart/design.md
+++ b/src/_guides/language/effective-dart/design.md
@@ -206,13 +206,11 @@ closingWindow // Returns a bool or a window?
 showPopup     // Sounds like it shows the popup.
 {% endprettify %}
 
-<aside class="alert alert-info" markdown="1">
-There is one exception to this rule. Input properties in [Angular][]
-components sometimes use imperative verbs for boolean setters because these
-setters are invoked in templates, not from other Dart code.
+**Exception:** Input properties in [Angular][] components sometimes use
+imperative verbs for boolean setters because these setters are invoked in
+templates, not from other Dart code.
 
 [angular]: {{site.angulardart}}
-</aside>
 
 
 ### CONSIDER omitting the verb for a named boolean *parameter*.
@@ -261,15 +259,15 @@ if (!socket.isDisconnected && !database.isEmpty) {
 }
 {% endprettify %}
 
-An exception to this rule is properties where the negative form is what users
-overwhelmingly need to use. Choosing the positive case would force them to
-negate the property with `!` everywhere. Instead, it may be better to use the
-negative case for that property.
-
 For some properties, there is no obvious positive form. Is a document that has
 been flushed to disk "saved" or "*un*-changed"? Is a document that *hasn't* been
 flushed "*un*-saved" or "changed"? In ambiguous cases, lean towards the choice
 that is less likely to be negated by users or has the shorter name.
+
+**Exception:** With some properties, the negative form is what users
+overwhelmingly need to use. Choosing the positive case would force them to
+negate the property with `!` everywhere. Instead, it may be better to use the
+negative case for that property.
 
 
 ### PREFER an imperative verb phrase for a function or method whose main purpose is a side effect.
@@ -548,7 +546,7 @@ a huge monolithic library, just that you are allowed to place more than one
 class in a single library.
 
 
-## Classes
+## Classes and Mixins
 
 Dart is a "pure" object-oriented language in that all objects are instances of
 classes. But Dart does not require all code to be defined inside a
@@ -694,25 +692,52 @@ If your class can be used as an interface, mention that in the class's doc
 comment.
 
 
-### AVOID mixing in a class that isn't intended to be a mixin.
+### DO use `mixin` to define a mixin type.
 
-If a constructor is added to a class that previously did not define any, that
-breaks any other classes that are mixing it in. This is a seemingly innocuous
-change in the class, and the restrictions around mixins aren't widely known.
-It's likely an author may add a constructor without realizing it will break your
-class that's mixing it in.
+Dart originally did not have a separate syntax for declaring a class intended to
+be mixed in to other classes. Instead, any class that met certain restrictions
+(no non-default constructor, no superclass, etc.) could be used as a mixin. This
+was confusing because the author of the class may not have intended it to be
+mixed in.
 
-Like with subclassing, this means a class needs to be deliberate about whether
-or not it wants to allow being used as a mixin. If the class doesn't have a doc
-comment or an obvious name like `IterableMixin`, you should assume you cannot
-mix in the class.
+Dart 2.1.0 added a `mixin` keyword for explicitly declaring a mixin. Types
+created using that can *only* be used as mixins, and the language also ensures
+your mixin stays within the restrictions. When defining a new type that you
+intend to be used as a mixin, use this syntax.
+
+{:.good-style}
+<?code-excerpt "misc/lib/effective_dart/design_good.dart (mixin)"?>
+{% prettify dart %}
+mixin ClickableMixin implements Control {
+  bool _isDown = false;
+
+  void click();
+
+  void mouseDown() {
+    _isDown = true;
+  }
+
+  void mouseUp() {
+    if (_isDown) click();
+    _isDown = false;
+  }
+}
+{% endprettify %}
+
+You may still encounter older code using `class` to define mixins, but the new
+syntax is preferred.
 
 
-### DO document if your class supports being used as a mixin.
+### AVOID mixing in a type that isn't intended to be a mixin.
 
-Mention in the class's doc comment whether the class can or must be used as a
-mixin. If your class is designed for use only as a mixin, then consider adding
-`Mixin` to the end of the class name.
+For compatibility, Dart still allows you to mix in classes not defined using
+`mixin`. However, that's risky. If the author of the class does not intend the
+class to be used as a mixin, they may change the class in a way that breaks the
+mixin restrictions. For example, if they add a constructor, it will break your
+class that mixes it in.
+
+If the class doesn't have a doc comment or an obvious name like `IterableMixin`,
+assume you cannot mix in the class if it isn't declared using `mixin`.
 
 
 ## Constructors
@@ -895,15 +920,13 @@ you want to add. Object's shouldn't generally expose more state than they need
 to. If you have some piece of an object's state that can be modified but not
 exposed in the same way, use a method instead.
 
-<aside class="alert alert-info" markdown="1">
-There is one exception to this rule. An [Angular][] component class may expose
-setters that are invoked from a template to initialize the component. Often,
-these setters are not intended to be invoked from Dart code and don't need a
-corresponding getter. (If they are used from Dart code, they *should* have a
-getter.)
+**Exception:** An [Angular][] component class may expose setters that are
+invoked from a template to initialize the component. Often, these setters are
+not intended to be invoked from Dart code and don't need a corresponding getter.
+(If they are used from Dart code, they *should* have a getter.)
 
 [angular]: {{site.angulardart}}
-</aside>
+
 
 ### AVOID returning `null` from members whose return type is `bool`, `double`, `int`, or `num`.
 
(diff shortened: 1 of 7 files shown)


kwalrath:
Thanks, @munificent! I'll take a look.

kwalrath:
Looking at the build breakage now...

kwalrath:
The build is broken due to a hint (easily ignored) and an error (not so much) in `usage_bad.dart` for the following code:

```
    if (optionalThing?.isEnabled) {
```

The code is used here:

https://sz-www-2.firebaseapp.com/guides/language/effective-dart/usage#do-use--to-convert-null-to-a-boolean-value

Here's the error message:

```
  error • Conditions must have a static type of 'bool' at lib/effective_dart/usage_bad.dart:12:9 • non_bool_condition
```

We can make the test pass by adding code like the following:

```
class EnableableThing {
  bool isEnabled;
  EnableableThing(this.isEnabled);
}

void miscDeclAnalyzedButNotTested() {
  {
    var optionalThing = EnableableThing(true);
```

Is that fix OK by you, @munificent?

src/_guides/language/effective-dart/index.md
```diff
@@ -108,6 +108,10 @@ Each guideline starts with one of these words:
 * **CONSIDER** guidelines are practices that you might or might not want to
   follow, depending on circumstances, precedents, and your own preference.
 
+Some guidelines may describe an **exception** where the rule does *not* apply.
```
kwalrath: delete "may"

src/_guides/language/effective-dart/index.md
```diff
@@ -108,6 +108,10 @@ Each guideline starts with one of these words:
...
+When listed, the exceptions may not be exhaustive&mdash;you may still need to
```
kwalrath: may -> might

src/_guides/language/effective-dart/style.md
```diff
@@ -420,13 +437,15 @@ Note that dartfmt does 99% of this for you, but the last 1% is you. It does not
...
-We make an exception for URIs and file paths. When those occur in comments or
-strings (usually in imports and exports), they may remain on a single line even
-if they go over the line limit. This makes it easier to search source files for
-a given path.
+**Exception:** When a URI or file path occurs in a comment or string (usually in
+an import or export), it may remain whole even if it causes the line to go over
+80 characters. This makes it easier to search source files for a path.
 
+**Exception:** Multi-line strings can contain lines longer than 80 characters
+since newlines are significant inside the string and splitting the lines into
```
kwalrath: since -> because
(some people — especially non-native English speakers — have trouble distinguishing between the time sense of "since" and the causality sense.)

src/_guides/language/effective-dart/style.md
```diff
@@ -444,9 +463,8 @@ if (isWeekDay) {
 }
 {% endprettify %}
 
-There is one exception to this: an `if` statement with no `else` clause where
-the entire `if` statement and the then body all fit in one line. In that case,
-you may leave off the braces if you prefer:
+**Exception:** An `if` statement with no `else` clause that fits entirely on one
```
kwalrath: All the other exceptions start with sentences, so let's do that here, too:

When you have an `if` statement with no `else` clause and the whole `if` statement fits on one line,
you can omit the braces if you prefer:

src/_guides/language/effective-dart/usage.md
```diff
@@ -119,6 +119,60 @@ library. Follow these two rules:
 * An import path should never contain `/lib/`.
 * A library under `lib` should never use `../` to escape the `lib` directory.
 
+## Booleans
+
+### DO use `??` to convert `null` to a Boolean value.
```
kwalrath: Boolean value -> boolean value
(https://developers.google.com/style/word-list#letter-B, and it's what we do elsewhere in this doc.)

src/_guides/language/effective-dart/usage.md
```diff
@@ -119,6 +119,60 @@ library. Follow these two rules:
...
+
+{:.bad-style}
+<?code-excerpt "misc/lib/effective_dart/usage_bad.dart (null-aware-condition)"?>
+{% prettify dart %}
+if (optionalThing?.isEnabled) {
+  print("Have enabled thing.");
+}
+{% endprettify %}
+
+This code will throw an exception if `optionalThing` is `null`. To fix this, you
```
kwalrath: will throw -> throws

src/_guides/language/effective-dart/usage.md
```diff
@@ -119,6 +119,60 @@ library. Follow these two rules:
...
+need to "convert" the `null` value either to `true` or `false`. There are two
```
kwalrath: either to -> to either

src/_guides/language/effective-dart/usage.md
```diff
@@ -119,6 +119,60 @@ library. Follow these two rules:
...
+{% endprettify %}
+
+Both operations produce the same result and do the right thing, but `??` is
+preferred for three main reasons:
+
+*   The `??` operator clearly signals that the code has something to do with a
+    `null` value.
+
+*   The `== true` looks like a common new programmer mistake where the equality
+    operator is redundant and can be removed. That's true when the Boolean
```
kwalrath: GLOBAL: Boolean -> boolean (except when it's the first word in a section head, of course)

src/_guides/language/effective-dart/usage.md
```diff
@@ -119,6 +119,60 @@ library. Follow these two rules:
...
+    expression on the left will not produce `null`, but not when it can.
+
+*   The `?? false` and `?? true` clearly show what value will be used when the
+    expression is `null`. With `== true`, you have to think through the Boolean
+    logic to realize that means that a `null` gets converted to *false*.
```
kwalrath: delete "that means"

src/_guides/language/effective-dart/usage.md
```diff
@@ -119,6 +119,60 @@ library. Follow these two rules:
 * An import path should never contain `/lib/`.
 * A library under `lib` should never use `../` to escape the `lib` directory.
 
+## Booleans
+
+### DO use `??` to convert `null` to a Boolean value.
+
+You sometimes have an expression that can evaluate to `true`, `false`, or `null`
```
kwalrath: This sounds presumptuous. How about:

This rule applies when an expression can evaluate ..., and you need to pass ....

src/_guides/language/effective-dart/usage.md
```diff
@@ -119,6 +119,60 @@ library. Follow these two rules:
...
+<?code-excerpt "misc/lib/effective_dart/usage_bad.dart (null-aware-condition)"?>
+{% prettify dart %}
+if (optionalThing?.isEnabled) {
+  print("Have enabled thing.");
+}
+{% endprettify %}
+
+This code will throw an exception if `optionalThing` is `null`. To fix this, you
+need to "convert" the `null` value either to `true` or `false`. There are two
+ways to do this, using `??` or `==`:
```
kwalrath: If you read this without going further down, it sounds like either one is OK. Let's be more judgy here. Maybe:

Although you could do this using `==`, we recommend using `??`:

src/_guides/language/effective-dart/usage.md
```diff
@@ -312,20 +366,17 @@ people.forEach((person) {
 });
 {% endprettify %}
 
-The exception is if all you want to do is invoke some already existing function
-with each element as the argument. In that case, `forEach()` is handy.
+Note that this guideline specifically says "function *literal*". If want to
```
kwalrath: want -> you want

src/_guides/language/effective-dart/usage.md
```diff
@@ -1096,10 +1147,9 @@ class Point {
 {% endprettify %}
 
 This `this.` syntax before a constructor parameter is called an "initializing
-formal". You can't always take advantage of it. In particular, using it means
-the parameter is not visible in the initialization list. But, when you can, you
-should.
-
+formal". You can't always take advantage of it. Sometimes you want to have a
+named parameter whose name doesn't match the name of the field you are
+initializing. But, when you can use initialing formals, you should.
```
kwalrath: But, -> But
maybe, for emphasis and ease of reading: can -> _can_   (and should -> _should_)
initialing -> initializing

src/_guides/language/effective-dart/design.md
```diff
@@ -548,7 +546,7 @@ a huge monolithic library, just that you are allowed to place more than one
 class in a single library.
 
 
-## Classes
+## Classes and Mixins
```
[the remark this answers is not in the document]

munificent:
Thank you!

PR #1780, opened 2019-07-22
Updated home screen images

FaisalAbid:


kwalrath:
Thanks for the PR, @FaisalAbid. Can you tell me what's changed with these files, and what we should be looking for as reviewers?

kwalrath:
A few files were new (maybe due to a typo in the filename). Is that a problem?

 create mode 100644 src/assets/dash/1x/logo_flutter packages.png
 create mode 100644 src/assets/dash/1x/logo_flutter packages.svg
 create mode 100644 src/assets/dash/2x/logo_flutter packages@2x.png
 create mode 100644 src/assets/dash/2x/responsicve design comparisons@2x-80.jpg

johnpryan:
- the flutter packages images are supposed to be new, but I'm not sure where in the new design they should be used.
- `responsicve design comparisons@2x-80.jpg` is an image from the designer to indicate requested changes to the site, so I think it should be taken out.

johnpryan:
I've removed that image here: https://github.com/dart-lang/site-www/pull/1788

FaisalAbid:
Sorry! I dont know how I missed these notifications.

Yes I don't know why the new images are there, they were part of the asset
package sent to me, I guessed it was for a future thing coming up and hence
why its in the commit

kwalrath:
I'm going to file a bug to make sure that we really want to include those flutter image files.

PR #1785, opened 2019-07-22
Documenting how to make a pre-release

jonasfj:
I'm sure I need some help making this more readable :)

### DIFF
--- a/src/tools/pub/publishing.md
+++ b/src/tools/pub/publishing.md
@@ -140,6 +140,35 @@ the first and only person authorized to upload additional versions of the packag
 To allow or disallow other people to upload versions,
 use the [`pub uploader`](cmd/pub-uploader) command.
 
+## Publishing pre-releases
+
+A package can be published as a pre-release while under development. This is
+often useful when:
+ * The next major of version of the package is under active development,
+ * The package depends on unstable Dart or Flutter SDK, or,
+ * Beta testing of the next release candidate is desired.
+
+As outlined in [semantic versioning][semver], a pre-release of version `2.0.0`
+is made by appending a suffix to the version (`2.0.0-dev.1`). When version
+`2.0.0` is released it'll take precedence over all `2.0.0-XXX` the pre-releases.
+`pub` will prefer stable releases when available, thus, developers who wish to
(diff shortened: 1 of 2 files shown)


kwalrath:
Thanks for the PR!

I made some edits, but I'm not sure about some of them (constraints, examples). Could you please take a careful look at my changes?

src/tools/pub/publishing.md
```diff
@@ -140,6 +140,38 @@ the first and only person authorized to upload additional versions of the packag
...
+
+Because `pub` prefers stable releases when available, users of a pre-release package
+might need to change their dependency constraints.
+For example, if a user wants to test pre-releases of version 2.1, then
+instead of `^2.0.0` or `^2.1.0` they might specify `^2.1.0-dev.1`.
+
+<aside class="alert alert-info" markdown="1">
+  **Note:**
+  Sometimes `pub` might choose a pre-release instead of a stable release;
+  this can occur when dependencies are nested.
```
jonasfj: I wish we could say something more precise here :)

I did a bit of digging and filed some tests in: https://github.com/dart-lang/pub/pull/2175

I think the sane and simple thing we can say something along the lines of:
> `pub` may choose a pre-release instead of a stable release, if there exists a stable package in the dependency graph depends on a pre-release.

(I'm sure it can be phrased better)

@jakemac53, isn't that a reasonably fair characterization of the cases where pre-releases may be chosen, despite stable releases being available?

jonasfj:
Apart from the `<aside>`, I think this is really good! See comment, I've done some digging to find a better characterization of the cases where this might surprise users.

kwalrath:
@jakemac53 could you please check my new words? They're even stronger than Jonas's.

src/tools/pub/publishing.md
jakemac53: Yes, that matches my understanding pretty well.

jakemac53: Well, actually it is more complicated than that but I don't know that it matters? There doesn't need to be any explicit dependencies on a pre-release. It can happen if a pre-release is the only version that allows the newest version of some other stable dependency.

Issue #1792, opened 2019-07-25
Advertise the code of conduct


kwalrath:
We should point to the Dart code of conduct from more places.

DONE: ~~Also, it should probably match (be the same as?) the Flutter code of conduct: https://github.com/flutter/flutter/blob/master/CODE_OF_CONDUCT.md~~

/cc @filiph @timsneath

kwalrath:
Also see https://github.com/dart-lang/sdk/issues/37663.

kwalrath:
In general, it seems like a good idea to link from CONTRIBUTING.md files to the code of conduct. https://github.com/dart-lang/.github/blob/master/CODE_OF_CONDUCT.md links to it and covers all GitHub dart-lang projects, but it's easy to miss.

kwalrath:
I've updated this repo's CONTRIBUTING.md file to link to the code of conduct.

kwalrath:
I think we're good for now. Open a new issue if you think of any other places that should link to the code of conduct.

PR #1793, opened 2019-07-26
Update Dart's code of conduct

kwalrath:
Contributes to #1792, making this closer to the Flutter code of conduct. 

Staged: https://kw-staging-dartlang-2.firebaseapp.com/code-of-conduct

Sorry, @kevmoo, I didn't add the cartoon... short and sweet, that's how I like my pages!

### DIFF
--- a/src/code-of-conduct.md
+++ b/src/code-of-conduct.md
@@ -4,21 +4,21 @@ description: Be excellent to each other.
 toc: false
 ---
 
-As contributors and maintainers of the Dart project,
-we promise to extend courtesy and respect to everyone involved in this project
-regardless of gender, gender identity, sexual orientation, disability, age,
-race, ethnicity, religion, or level of experience.
-We expect anyone contributing to the Dart project to do the same.
+We expect Dart contributors to act professionally and respectfully, and
+we expect our social spaces to be safe and dignified environments.
 
-Communication through any of Dart's channels
-(GitHub, Gitter, mailing lists, Google+, Twitter, etc.)
-must be constructive and never resort to personal attacks, trolling,
-public or private harassment, insults, or other unprofessional conduct.
+Specifically:
+
+* Respect people, their identities, their culture, and their work.
+* Be kind. Be courteous. Be welcoming.
+* Listen. Consider and acknowledge people's points before responding.
 
 If any member of the community violates this code of conduct,
-the maintainers of the Dart project may take action, removing issues,
-comments, and PRs or blocking accounts as deemed appropriate.
+the maintainers of the Dart project may take action, removing issues, comments, and PRs or blocking accounts as deemed appropriate.
+We will not tolerate harassment from anyone in the Dart community,
+even outside of Dart's public communication channels.
 
 If you are subject to or witness unacceptable behavior,
 or have any other concerns,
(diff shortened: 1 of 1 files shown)


src/code-of-conduct.md
```diff
@@ -4,21 +4,21 @@ description: Be excellent to each other.
 toc: false
 ---
 
-As contributors and maintainers of the Dart project,
-we promise to extend courtesy and respect to everyone involved in this project
-regardless of gender, gender identity, sexual orientation, disability, age,
```
timsneath: I personally think this is actually quite important. We should call out various examples forms of harassment so that folk can't claim that their comment is 'out of scope' somehow. 

Another example of this is here: https://gist.github.com/Integralist/4e84af31c483d0f5a182, which I also like. I think upon occasion people appreciate having the specifics spelled out.

src/code-of-conduct.md
```diff
@@ -4,21 +4,21 @@ description: Be excellent to each other.
...
-race, ethnicity, religion, or level of experience.
-We expect anyone contributing to the Dart project to do the same.
+We expect Dart contributors to act professionally and respectfully, and
```
timsneath: I wonder if the right phrase here is "community members" rather than "contributors"? Could this be taken as implying that this is _our_ responsibility, rather than a responsibility we expect the entire community to uphold, perhaps?

src/code-of-conduct.md
```diff
@@ -4,21 +4,21 @@ description: Be excellent to each other.
...
 If any member of the community violates this code of conduct,
-the maintainers of the Dart project may take action, removing issues,
-comments, and PRs or blocking accounts as deemed appropriate.
+the maintainers of the Dart project may take action, removing issues, comments, and PRs or blocking accounts as deemed appropriate.
+We will not tolerate harassment from anyone in the Dart community,
+even outside of Dart's public communication channels.
 
 If you are subject to or witness unacceptable behavior,
 or have any other concerns,
 please [email us](mailto:conduct@dartlang.org).
```
timsneath: Are we now conduct@dart.dev? (I confess that I haven't tried this yet myself, but I know that we're gradually moving towards dart.dev as the canonical domain for all of this.)

timsneath:
When we're done, we should also add this to the GitHub repo, per https://help.github.com/en/articles/adding-a-code-of-conduct-to-your-project

src/code-of-conduct.md
kwalrath: I like being less specific, because you can never list all the categories. Someone wanted political leanings added, which I totally get, but I hesitated to pile on so many things that the details obscure the overall meaning.

I modeled this after the Flutter code of conduct, which is nicely general (to my eyes).

src/code-of-conduct.md
kwalrath: I'm fine with community members. Again, I was mimicking the Flutter code of conduct.

src/code-of-conduct.md
kwalrath: afaik, we haven't moved any of the dartlang.org groups yet.

kwalrath:
@timsneath ptal

PR #1797, opened 2019-07-29
Replace v2.2 spec pdf in order to get correct index references

eernstg:
The existing DartLangSpec-v2.2.pdf has incorrect references from the index to the main text. For instance, p206 shows that 'static variable' is defined on page 11, but the actual page is 15. This updates provides a version of DartLangSpec-v2.2.pdf where these references are correct.


mit-mit:
rslgtm

PR #1802, opened 2019-07-30
Update usage.md

shumingch:
Fix typo


kwalrath:
Nice job of tracking down the reason for the failing test, @shumingch! Now we just need @munificent to review the code change.

munificent:
That fix is exactly right. Thanks!

PR #1805, opened 2019-08-01
Add DartPad to the front page

johnpryan:
adds DartPad to the front page with a dropdown to select from a set of samples

Screen Shot 2019-08-08 at 11 10 41 AM

@kwalrath I will update the samples soon, making this PR to get initial feedback on the code

this also exposed an issue with the reset button - fix here: https://github.com/dart-lang/dart-pad/pull/1193

staged: https://jr-dartlang-1.firebaseapp.com


mit-mit:
Top screenshot looks good! What does it look like on a mobile browser?

johnpryan:
@mit-mit I almost forgot... Added 9195004 to hide DartPad on mobile browsers. Here's a screenshot:

Screen Shot 2019-08-06 at 1 11 49 PM

An alternative would be to use a CSS media query to hide on small screens, but I'm not sure if a large tablet should still show DartPad.

src/index.html
```diff
@@ -249,4 +252,12 @@
                 src="assets/dash/2x/3-1 - aot compile@2x-80.jpg"/>
         </div>
     </section>
+    <section class="row dash-row">
+        <div class="dash-dartpad">
+            <h2>Try Dart in your browser!</h2>
+            <select id="dartpad-select"></select>
+            <div id="dartpad-host"></div>
+            <h3>Another call to action</h3>
```
johnpryan: @kwalrath do we have the words ready yet?

kwalrath: Sorry, I didn't realize you were waiting on me. I'll work on the words.

In the meantime, could you modify the "Try Dart" link(s?) to go to the embedded dartpad?

kwalrath: How about something like this:

Want more practice? <a href="/codelabs">Try a codelab.</a> Or <a href="/get-dart">download Dart.</a>

kwalrath: /cc @messi01, @galeyang for the text/approach.

@johnpryan if we use an h3, we'll need to fix the link color.

johnpryan:
staged at: https://jr-dartlang-1.firebaseapp.com

src/index.html
```diff
@@ -252,12 +252,13 @@
...
-    <section class="row dash-row">
+    <section class="row">
         <div class="dash-dartpad">
-            <h2>Try Dart in your browser!</h2>
+            <a name="try-dart"></a>
+            <h2>Try Dart in your browser</h2>
             <select id="dartpad-select"></select>
             <div id="dartpad-host"></div>
-            <h3>Another call to action</h3>
+            <h3>Want more practice? <a href="/codelabs">Try a codelab.</a> Or <a href="/get-dart">download Dart</a>.</h3>
```
kwalrath: </a>. -> .</a>

(For consistency. I don't usually like having the punctuation within the link text, but it looks better to me here.)

src/_includes/navigation-main.html
```diff
@@ -19,8 +19,8 @@
         {%- if page_url_path contains '/*/community/' %} active {%- endif -%}">Community</a>
     </li>
     <li>
-      <a href="https://dartpad.dev" target="_blank" class="nav-link no-automatic-external 
-        {%- if page_url_path contains '/*/#dartpad' %} active {%- endif -%}">Try Dart</a>
+      <a href="/#try-dart" class="nav-link no-automatic-external
```
kwalrath: We don't need the `no-automatic-external` here, do we?

johnpryan: good catch, removed

johnpryan:
@kwalrath The `page.url` doesn't seem to contain the hash that we are using to navigate to the "Try Dart" section, so I took that piece out.

This means that we can't show an underline if you navigate to /#try-dart

kwalrath:
> @kwalrath The `page.url` doesn't seem to contain the hash that we are using to navigate to the "Try Dart" section, so I took that piece out.
> 
> This means that we can't show an underline if you navigate to /#try-dart

Ah, I think that's because the page.url is something that's set to the page path by jekyll. So that makes sense. If we wanted to highlight it, we might want to use code like that in the TOCs... but I don't think we need to worry about it too much.

galeyang:
Could we add a border to the iframe? `border: 1px solid #293542`

Or maybe we should include a border in DartPad for both themes by default. Currently, we have to manually ensure the border is added at the iframe's parent level.

johnpryan:
@galeyang that border looks much nicer, thanks!

johnpryan:
This will be ready soon - I'm going to update the samples based on some initial feedback.

johnpryan:
The samples have been updated and are staged here: https://jr-dartlang-1.firebaseapp.com/

PR #1823, opened 2019-08-08
Change factory constructor sample

whesse:
This change rewrites the factory constructor example to use putIfAbsent to implement a cache.

It can well be that we want to keep the existing code, instead of rewriting the cache example.
The existing code is easy to understand for beginners.
The Map.putIfAbsent is a more complex concept than the explicit code (doing the same thing) that
is used now in the code.  But putIfAbsent is the more idiomatic way of writing it.

### DIFF
--- a/examples/misc/lib/language_tour/classes/logger.dart
+++ b/examples/misc/lib/language_tour/classes/logger.dart
@@ -10,13 +10,8 @@ class Logger {
       <String, Logger>{};
 
   factory Logger(String name) {
-    if (_cache.containsKey(name)) {
-      return _cache[name];
-    } else {
-      final logger = Logger._internal(name);
-      _cache[name] = logger;
-      return logger;
-    }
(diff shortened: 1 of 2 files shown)


kwalrath:
Your code seems much better, @whesse. Thanks!

To get the build to work, we'll need to modify [/examples/misc/lib/language_tour/classes/logger.dart](https://github.com/dart-lang/site-www/blob/master/examples/misc/lib/language_tour/classes/logger.dart) to reflect the new code.

Can you do that, or would you like me to?

whesse:
I will update the required code tomorrow.  I wasn't sure that introducing the putIfAbsent idiom would be confusing to readers or not, glad you think it is better.

kwalrath:
@whesse if it's too inconvenient to update the code, I can do it.

whesse:
OK, feel free to do it. I've just gotten busy with all sorts of other things.

PR #1846, opened 2019-08-21
Advise UpperCamelCase extensions

pq:
/cc @munificent @kwalrath


pq:
Not quite sure how to fix the breakage.  When I follow the advise from travis I hit this snag:

```
[~/src/repos/site-www] (master) $ tool/refresh-code-excerpts.sh
/Users/pq/src/repos/site-www/tool/env-set.sh: line 3: tool/shared/env-set.sh: No such file or directory
```

(and indeed I don't see `env-set.sh`...)

kwalrath:
I'll take a look...

pq:
Thanks!

kwalrath:
There are a couple of things going on. The first is that there's no corresponding change in examples/misc/lib/effective_dart/style_good.dart. Once you add it, I think you run into the issue that the stable channel doesn't support `extension` yet.

kwalrath:
Let's let this sit a little bit... @munificent feel free to review the words and code. When we're ready to publish 2.5-related stuff (whether on dev or in stable), then I'll make this work.

pq:
Sounds good.  I added the 2.5 milestone.  Let me know if I should do anything else.

Thanks!

kwalrath:
It looks like even the dev channel (2.5.0-dev.2.1) doesn't support extension yet.

pq:
Yeah.  At current, you need to enable the experiment (either w/ a command line flag or analysis option).

kwalrath:
To make the build pass, we need to (1) support extension methods (so... not yet) and (2) update the examples directory to contain the code we import into the page. I've done the latter. We'll have to wait on this PR until we have an SDK version that supports the former.

kwalrath:
The dev build is happier now. Analyzer messages have changed, so the dev build is still failing, but that's a known issue. (#1894)

kwalrath:
I think we're ready to publish. Agreed, @pq?

Staged: https://kw-www-dartlang-1.firebaseapp.com/guides/language/effective-dart/style#do-name-extensions-using-uppercamelcase

pq:
Agreed.  Thanks!

PR #1849, opened 2019-08-22
dart:ffi docs

kwalrath:
@mit-mit what do you think of the direction of these docs?

Staged: https://kw-staging-dartlang-2.firebaseapp.com/guides/libraries/c-interop

/cc @mjohnsullivan


src/_guides/libraries/c-interop.md
```diff
@@ -125,10 +125,18 @@ Once you understand the hello_world example, you should be ready to look at the
 [other dart:ffi examples](#examples).
 
 
+## Packaging and loading C libraries
+
+How you get C libraries, distribute them, and load them into your code
```
mit-mit: "get C libraries" is a bit confusing here. I would just say "How you package C libraries, distribute them, and load them..."

mit-mit: Those are the right steps (for 3 I've seen both the terms packaging and distributing used depending on the methods used.

Rephrasing my feedback in terms of these, my point was that the docs really don't cover 1. We're assuming that the developer already has some C code (wether source or compiled) and then are trying to do steps 2-4.

src/_guides/libraries/c-interop.md
```diff
@@ -0,0 +1,142 @@
...
+6. Call the C function.
+```dart
+  hello();
+```
+
+Once you understand the hello_world example, you should be ready to look at the
+[other dart:ffi examples](#examples).
+
+
+## Loading and packaging C libraries
```
mit-mit: Can we swap these? Logically you need to package the C lib first so that you can load it later

src/_guides/libraries/c-interop.md
```diff
@@ -0,0 +1,142 @@
...
+
+How you load C libraries and bundle them with your package or app
```
mit-mit: Same comment about swapping. Also, any reason to not be consistent here and use package again (or use bundle in the heading)?

kwalrath: It's fine, but I'll want to use some synonyms somewhere so people can find this info, even if they don't know our chosen terminology.

distributing, bundling, packaging... are there any other synonyms we should consider using?

mit-mit: Got it. But would still prefer to swap the order.

src/_guides/libraries/c-interop.md
```diff
@@ -0,0 +1,142 @@
...
+The dart:ffi library is [in active development][ffi issue]
+and isn't complete yet.
+Although its API reference documentation hasn't been published yet,
+<!-- https://github.com/dart-lang/sdk/issues/37946 -->
+you can find its source code
+[in the SDK repo.](https://github.com/dart-lang/sdk/tree/master/sdk/lib/ffi)
+
+## Examples
+
+The following examples show how to use the dart:ffi library:
```
mit-mit: @kwalrath, are these going to stay in Matt's repo, or are we moving them to this repo to make it more self contained?

src/_guides/libraries/c-interop.md
```diff
@@ -0,0 +1,142 @@
...
+| c/hello.o | An intermediate object file. |
+{:.table .table-striped }
+
+### Building and running
+
+Here's an example of building the dynamic library and executing the Dart app:
+
+```terminal
+$ cd c
+$ make dylib
```
mit-mit: Is `make dylib` a command?

mit-mit: (but we could perhaps explain what it does)

src/_guides/libraries/c-interop.md
kwalrath: They're moving to the new samples repo: https://github.com/dart-lang/samples/pull/1

src/_guides/libraries/c-interop.md
kwalrath: @mjohnsullivan what would you like to do?

kwalrath: Taking a closer look, I see that the gcc line is just the output of the make command (i.e. what the make command is actually doing). I think it makes sense in the rendered output:

image

kwalrath:
@mjohnsullivan not urgent, but when you get a chance could you tell me why setup.sh is needed? (Can it be removed?)

mit-mit:
@kwalrath can we merge this today, Monday?

kwalrath:
> @kwalrath can we merge this today, Monday?

Yes, we can. It'll break the build, but once we ship stable, the build should be fine. (And we can always push the site manually.)

kwalrath:
I'd like to update the example links as soon as dart-lang/samples#1 is merged, but that doesn't need to block the publication of this page.

mit-mit:
Samples just landed!

kwalrath:
OK, this page is ready when you are.

PR #1862, opened 2019-09-03
Minor grammar fix: Adding a missing word

rowild:


rowild:
@googlebot I signed it!

PR #1864, opened 2019-09-03
Fix formatting of a warning

### LINKED ISSUE #1863: 'Customizing static analysis' page issue
Page URL: https://dart.dev/guides/language/analysis-options
Page source: https://github.com/dart-lang/site-www/tree/master/src/_guides/language/analysis-options.md

Description of issue:
At the end of this subsection

[https://dart.dev/guides/language/analysis-options#disabling-individual-rules](https://dart.dev/guides/language/analysis-options#disabling-individual-rules)

there is an `<aside>` warning block, which does not render correctly, even though the markdown seems to be totally correct.

I, personally, have no idea, why the block is not rendered correctly, but I hope you do, and if so, maybe you could mention it here, so that in future cases I know what to do.

Thank you!

kwalrath:
Fixes #1863

### DIFF
--- a/src/_guides/language/analysis-options.md
+++ b/src/_guides/language/analysis-options.md
@@ -243,8 +243,8 @@ linter:
     await_only_futures: true
 {% endprettify %}
 
-<aside class="alert alert-warning” markdown="1">
-  If you disable any rules, then use key-value syntax to enable rules.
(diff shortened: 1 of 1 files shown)


legalcodes:
Nice catch, LGTM

PR #1870, opened 2019-09-05
[WIP] Add info about 2.5 const expansion

kwalrath:
Staged: https://kw-www-dartlang-1.firebaseapp.com/guides/language/language-tour#final-and-const (scroll down to the end of the section)

### DIFF
--- a/examples/strong/analyzer-2-results.txt
+++ b/examples/strong/analyzer-2-results.txt
@@ -1,6 +1,6 @@
 Analyzing lib, test...
-  error • 'HoneyBadger.parent' ('Root Function()') isn't a valid override of 'Animal.parent' ('Animal Function()') at lib/animal_bad.dart:15:3 • invalid_override
-  error • 'Cat.chase' ('void Function(Mouse)') isn't a valid override of 'Animal.chase' ('void Function(Animal)') at lib/animal_bad.dart:26:3 • invalid_override
+  error • 'HoneyBadger.parent' ('Root Function()') isn't a valid override of 'Animal.parent' ('parent') at lib/animal_bad.dart:15:3 • invalid_override
+  error • 'Cat.chase' ('void Function(Mouse)') isn't a valid override of 'Animal.chase' ('chase') at lib/animal_bad.dart:26:3 • invalid_override
   error • A value of type 'List' can't be assigned to a variable of type 'List<Cat>' at lib/animal_bad.dart:49:19 • invalid_assignment
   error • The method 'add' isn't defined for the class 'Iterable' at lib/bounded/instantiate_to_bound.dart:7:5 • undefined_method
   error • A value of type 'Element' can't be assigned to a variable of type 'CanvasElement' at lib/common_fixes_analysis.dart:21:28 • invalid_assignment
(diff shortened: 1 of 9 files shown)


kwalrath:
@bwilkerson could you please take a look at the diffs to the [analyzer results](https://github.com/dart-lang/site-www/pull/1870/files#diff-d1944da8407ea8703c30a41e24c3fe5a)?

There are some changes that look weird — e.g. what used to be a type in the message changed to an identifier.

For example, in the following example, `Animal Function()` changes to `parent`.

Before: 
```
error • 'HoneyBadger.parent' ('Root Function()') isn't a valid override of 'Animal.parent' ('Animal Function()') at lib/animal_bad.dart:15:3 • invalid_override
```

After:
```
error • 'HoneyBadger.parent' ('Root Function()') isn't a valid override of 'Animal.parent' ('parent') at lib/animal_bad.dart:15:3 • invalid_override
```

bwilkerson:
> could you please take a look at the diffs to the analyzer results?

That looks like an unintentional regression in analyzer's behavior. We should definitely try to fix that on the analyzer side, but I don't know when that will happen, so you probably don't want to wait for that to happen before committing.

kwalrath:
Whew, finally fixed the dev build with that last putback. (The stable build is red due to having to increase the SDK dependency to avoid analyzer warnings about using new const features.)

kwalrath:
> > could you please take a look at the diffs to the analyzer results?
> 
> That looks like an unintentional regression in analyzer's behavior. We should definitely try to fix that on the analyzer side, but I don't know when that will happen, so you probably don't want to wait for that to happen before committing.

OK, sounds reasonable. Thanks or the quick response. Would you like me to file a bug, or do you have it handled?

bwilkerson:
Issue created: https://github.com/dart-lang/sdk/issues/38238

mit-mit:
Travis seems unhappy still

kwalrath:
> Travis seems unhappy still

It'll stay unhappy until there's a stable release of 2.5.

mit-mit:
Passing now! Merging

examples/strong/analyzer-2-results.txt
```diff
@@ -1,18 +1,18 @@
 Analyzing lib, test...
-  error • 'HoneyBadger.parent' ('Root Function()') isn't a valid override of 'Animal.parent' ('Animal Function()') at lib/animal_bad.dart:15:3 • invalid_override
-  error • 'Cat.chase' ('void Function(Mouse)') isn't a valid override of 'Animal.chase' ('void Function(Animal)') at lib/animal_bad.dart:26:3 • invalid_override
+  error • 'HoneyBadger.parent' ('Root Function()') isn't a valid override of 'Animal.parent' ('parent') at lib/animal_bad.dart:15:3 • invalid_override
+  error • 'Cat.chase' ('void Function(Mouse)') isn't a valid override of 'Animal.chase' ('chase') at lib/animal_bad.dart:26:3 • invalid_override
```
chalin: Hmm, the analyzer used to report two `Function()` types and now instead of the second function's type it repeats the second function's name? Is that expected behavior?

kwalrath: No, it's a bug: https://github.com/dart-lang/sdk/issues/38238

chalin: Thanks for the context!

PR #1886, opened 2019-09-10
Enable manual testing of the httpserver example

kwalrath:
I couldn't run `webdev` on the httpserver example (which I did to test the output for #1878) until I added these dev dependencies. I sloppily didn't fill in the version... not sure that it matters here.


kwalrath:
Weirdly, the Travis build status wasn't updating. I believe it passed, but it's running again, just to be sure.

Issue #1803, opened 2019-07-31
Test all code excerpts


kwalrath:
This bug replaces #407, which is the equivalent for the pre-dart.dev version of this site. We'll need to be able to test all samples as [null safety](https://github.com/dart-lang/language/tree/master/accepted/future-releases/nnbd) is released.

---

New pages that need tests for code excerpts:

- [ ] [Dart cheatsheet codelab](https://dart.dev/codelabs/dart-cheatsheet)
- [ ] [Dart language cheatsheet](https://dart.dev/guides/language/cheatsheet)
- [ ] [Dart async/await codelab](https://dart.dev/codelabs/async-await)

Old pages that should have tests for code excerpts, including DartPad code (which in most cases is contained in a Gist):

- [x] [src/_tutorials/web/fetch-data.md](https://github.com/dart-lang/site-www/blob/master/src/_tutorials/web/fetch-data.md)
  - Gists are out of date. E.g., the "Get portmanteaux" example fails. We should put the code in this repo instead of in gists. See [src/codelabs/async-await.md](https://raw.githubusercontent.com/dart-lang/site-www/master/src/codelabs/async-await.md) (in particular, the code in the `else` statements for `useIframe`) for an example of how to do this.
  - Consider dropping or editing the "Implementation note"; is it really relevant to the reader's experience and/or understanding?
- [x] examples/fetch_data: write tests or check web app build for problems #2113
- [ ] [src/_tutorials/web/low-level-html/connect-dart-html.md](https://github.com/dart-lang/site-www/blob/master/src/_tutorials/web/low-level-html/connect-dart-html.md)
  - Noticed in the course of creating a DartPad-related PR (#1956).
- [ ] [src/_tutorials/web/low-level-html/add-elements](https://github.com/dart-lang/site-www/blob/master/src/_tutorials/web/low-level-html/add-elements.md)
  - Has "new" and other signs of old code.

Lower priority code in new pages:
- [ ] Some pages migrated from the old webdev site:
  - [ ] [webdev](https://dart.dev/tools/webdev) tool page (pubspec entry).
  - [x] [build_runner](https://dart.dev/tools/build_runner) page (pubspec entry).

Other:

- [ ] New excerpts in language tour, including the one fixed in #2184.
- [ ] #1959 Link/include mini, mini_with_style?
- [ ] [Zones article](https://dart.dev/articles/archive/zones) code refresh (to the extent possible)? (no.67, 0.25%). Related #1812.

Tested to some extent:
[...]

PR #1888, opened 2019-09-11
cmdline: Hello World example cleanup and test

chalin:
Contributes to #1803.

(@kwalrath: Since I haven't submitted a PR in a while, I'm making this first one deliberately small :))

### DIFF
--- a/src/_tutorials/server/cmdline.md
+++ b/src/_tutorials/server/cmdline.md
@@ -20,17 +20,16 @@ prevpage:
   * Most input and output requires the use of streams.
 </div>
 
-<aside class="alert alert-info" markdown="1">
-<strong>Prerequisite:</strong>
-This tutorial uses the `async` and `await` language features, which rely on the
-<a href="{{site.dart_api}}/{{site.data.pkg-vers.SDK.channel}}/dart-async/Future-class.html" target="_blank">Future</a>
-and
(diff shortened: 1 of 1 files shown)


src/_tutorials/server/cmdline.md
```diff
@@ -20,17 +20,16 @@ prevpage:
...
-This tutorial uses the `async` and `await` language features, which rely on the
-<a href="{{site.dart_api}}/{{site.data.pkg-vers.SDK.channel}}/dart-async/Future-class.html" target="_blank">Future</a>
-and
-<a href="{{site.dart_api}}/{{site.data.pkg-vers.SDK.channel}}/dart-async/Stream-class.html"
-   target="_blank">Stream</a> classes for asynchronous support.
-To learn more about these features, see the
-[asynchronous programming codelab](/codelabs/async-await) and the
-[streams tutorial](/tutorials/language/streams).
-</aside>
+{{site.alert.note}}
```
chalin: @kwalrath This aside used to have a _Prerequisite_ inlined heading, but the note text isn't asking the reader to do anything. I've changed this to have a standard **Note** heading (with the usual icon). Maybe this should be an **Important** note instead? If you'd like to keep _Prerequisite_ as a heading, then maybe the text should be tweaked to say something like "ensure that your `PATH` is appropriately setup", but possibly just a **Note** is enough?

PR #1896, opened 2019-09-13
Update search-all.html

kwalrath:
Someone had updated our main search engine to search more than dart.dev. I've changed it back (because getting flutter.dev results when I wanted dart.dev was really annoying) and added more sites to this search.


mit-mit:
LGTM

PR #1898, opened 2019-09-13
Update vers to 2.5

kwalrath:
Visible in https://dart.dev/tools/sdk.
Process documented in https://github.com/dart-lang/site-www/wiki/Updating-to-new-SDK-releases.

### DIFF
--- a/src/_data/pkg-vers.json
+++ b/src/_data/pkg-vers.json
@@ -3,6 +3,6 @@
     "doc-path": "install",
     "channel": "stable",
     "prev-vers": "1.24.3",
-    "vers": "2.4.0"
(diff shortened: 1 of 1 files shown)


chalin:
Might we also want to update the min SDK version in the example pubspecs?

```yaml
environment:
  sdk: '>=2.5.0-dev.2.1 <3.0.0'
```

(It isn't strictly necessary, since the examples run fine, but it would make for a cleaner min SDK version :)).

kwalrath:
> Might we also want to update the min SDK version in the example pubspecs?

Good catch. I'll do that (and update the wiki).

PR #1901, opened 2019-09-15
Tool rename

chalin:
Prep work for #1803. cc @kwalrath


chalin:
Tooling-only PR. Travis is green. Merging. cc @kwalrath

PR #1904, opened 2019-09-16
_config.yml: add a generic warning alert w/o a label

chalin:


chalin:
Trivial config addition. Travis is green. Merging. cc @kwalrath

PR #1908, opened 2019-09-16
Code excerpts: distribute site-global replace pattern to individual files

chalin:
No site content changes.

Prep work so that we can actually show code excerpts using analyzer ignore directives.
Contributes to #1803


chalin:
Infrastructure change only. Travis is green. Merging. cc @kwalrath

PR #1911, opened 2019-09-17
Upgrade firebase-tools

chalin:


chalin:
Infra-only PR. Travis is green. Merging. cc @kwalrath

chalin:
@kwalrath - This PR updates `package.json`. Although I think that you run a globally installed version of firebase-tools, you might still want to run `npm install` under the site-www folder (so as to get the latest npm packages).

Issue #1919, opened 2019-09-19
Support loading DartPad sources from site vs Gist


chalin:
The [dartpad_picker](https://github.com/dart-lang/site-www/tree/master/tool/dartpad_picker) illustrates how Dartpad sources can be loaded from somewhere other than a Gist. The tool is currently used to build a dropdown list that is shown on the homepage, but the functionality to load Dartpad sources could be factored out and reused elsewhere on the site.

Such an approach would address this issue that we opened over dartpad: https://github.com/dart-lang/dart-pad/issues/788 Load content from any GitHub repo path.

Related: https://github.com/dart-lang/dart-pad/issues/1096 Load exercises from a repo in addition to gists. In particular this comment: https://github.com/dart-lang/dart-pad/issues/1096#issuecomment-517842744.

RedBrogdon:
I've been working on this for a bit, and should have a PR to allow for loading from a repo ready for review in the next few days.

chalin:
As I mentioned out-of-band, the [GitHub rate limit](https://developer.github.com/v3/#rate-limiting) (as was mentioned in [this comment](https://github.com/dart-lang/site-www/issues/1972#issuecomment-542870466)), might be another reason to explore the solution proposed here.

PR #1921, opened 2019-09-19
Get-started server tutorial: analyze/test code excerpts

chalin:
- Resolves 3rd item of #1803, for the `tutorials/server/get-started` page.
  - Note that I've chosen not to do anything with the opening Dartpad hello-world example. It doesn't really make sense to do anything with that example (since the Gist is created manually) until #1919 is addressed.
- I've also consistently applied the recommendation of using `$` as the default command-line prompt (see "Command prompt" in [Documenting command-line syntax](https://developers.google.com/style/code-syntax)). The copyediting for this is in a separate commit.

Staged at https://dart-dev-staging-1.firebaseapp.com/tutorials/server/get-started
Original https://dart.dev/tutorials/server/get-started

Possible followup issue: #1919.

---

I've checked external links: 

```console
$ blc -fgi http://localhost:4000/tutorials/server/get-started
Getting links from: http://localhost:4000/tutorials/server/get-started
├───OK─── https://dartpad.dev/
├───OK─── https://chocolatey.org/
├───OK─── http://www.google.com/intl/en/policies/privacy/
├───OK─── https://pub.dev/
├───OK─── http://creativecommons.org/licenses/by/3.0/
├───OK─── http://brew.sh/
├───OK─── https://dartpad.dev/27e044ec9e2957d9c5c7062871ce8bf3
├───OK─── https://flutter.dev/
├───OK─── https://pub.dev/packages/stagehand
├───OK─── https://dartpad.dev/experimental/embed-new-inline.html?id=27e044ec9e2957d9c5c7062871ce8bf3
├───OK─── https://api.dart.dev/
├───OK─── https://github.com/dart-lang/site-www/tree/master/src/_tutorials/server/get-started.md
├───OK─── https://github.com/dart-lang/site-www/issues/new?...
├───OK─── https://medium.com/dartlang/announcing-dart-2-5-super-charged-development-328822024970
├───OK─── https://medium.com/dartlang
├───OK─── https://github.com/dart-lang/site-www
```

### DIFF
--- a/src/_includes/get-sdk.md
+++ b/src/_includes/get-sdk.md
@@ -1,9 +1,6 @@
 Once you’re ready to move beyond DartPad and develop real apps,
 you need the Dart SDK.
 
-As you install, **note the path to the SDK.**
-You'll need it in the next step.
-
 <ul class="tabs__top-bar">
   <li class="tab-link current" data-tab="tab-sdk-install-windows">Windows</li>
   <li class="tab-link" data-tab="tab-sdk-install-linux">Linux</li>
@@ -23,15 +20,15 @@ You'll need it in the next step.
 
    1. Perform the following one-time setup:
       ```terminal
-      > sudo apt-get update
-      > sudo apt-get install apt-transport-https
-      > sudo sh -c 'curl https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add -'
-      > sudo sh -c 'curl https://storage.googleapis.com/download.dartlang.org/linux/debian/dart_stable.list > /etc/apt/sources.list.d/dart_stable.list'
+      $ sudo apt-get update
+      $ sudo apt-get install apt-transport-https
+      $ sudo sh -c 'curl https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add -'
+      $ sudo sh -c 'curl https://storage.googleapis.com/download.dartlang.org/linux/debian/dart_stable.list > /etc/apt/sources.list.d/dart_stable.list'
       ```
    2. Install the Dart SDK:
       ```terminal
-      > sudo apt-get update
-      > sudo apt-get install dart
+      $ sudo apt-get update
+      $ sudo apt-get install dart
       ```
 </div>
 
@@ -40,9 +37,12 @@ You'll need it in the next step.
   installing Dart is easy.
 
   ```terminal
-  > brew tap dart-lang/dart
-  > brew install dart
+  $ brew tap dart-lang/dart
+  $ brew install dart
   ```
 </div>
 
-More information: [Get the Dart SDK](/get-dart)
+{{site.alert.important}}
+  For more information, including how to **adjust your `PATH`**, see
+  [Get the Dart SDK](/get-dart).
+{{site.alert.end}}
--- a/src/_includes/mac-path.md
+++ b/src/_includes/mac-path.md
@@ -1,0 +1,5 @@
+{{site.alert.important}}
+  Ensure that the **Homebrew `bin` directory is in your `PATH`** if you'd like
+  direct command-line access to Dart commands such as the Dart VM (`dart`). For
+  details, consult the [Homebrew FAQ](https://docs.brew.sh/FAQ).
+{{site.alert.end}}
--- a/src/_includes/mac-path.md
+++ b/src/_includes/mac-path.md
@@ -1,5 +1,0 @@
-{{site.alert.important}}
-  Ensure that the **Homebrew `bin` directory is in your `PATH`** if you'd like
-  direct command-line access to Dart commands such as the Dart VM (`dart`). For
-  details, consult the [Homebrew FAQ](https://docs.brew.sh/FAQ).
-{{site.alert.end}}
--- a/src/_tutorials/server/get-started.md
+++ b/src/_tutorials/server/get-started.md
@@ -105,7 +105,7 @@ Hello world: 42!
 Let's customize the app you just created.
 
  1. Edit `lib/cli.dart` to calculate a different result. For example, divide the
-    previous value by two (see [Arithmetic operators][] for operator details):
+    previous value by two (operator details, see [Arithmetic operators][]):
 
     <?code-excerpt "misc/test/tutorial/get_started.dart (calculate)" replace="/~\/ 2/[!$&!]/g"?>
     {% prettify dart %}
--- a/src/_tutorials/server/get-started.md
+++ b/src/_tutorials/server/get-started.md
@@ -44,23 +44,17 @@ More information:
 
 {% include get-sdk.md %}
 
-<!-- PENDING: the following instructions assume you have set the PATH.
-     We should update the included instructions to refer to that. -->
-
 ## 3. Get more command-line developer tools
 
 Install [`stagehand`,][stagehand] which gives you templates for creating Dart apps:
 
 ```terminal
-> pub global activate stagehand
+$ pub global activate stagehand
 ```
-
 
 Note that although these instructions feature the command line,
 many IDEs support Dart development.
 Those IDEs use Stagehand behind the scenes when you create new Dart projects.
-
-<!-- PENDING: the following instructions assume you have the bin directory for the system cache in your path. -->
 
 More information:
 
@@ -72,9 +66,9 @@ More information:
 Create a command-line app:
 
 ```terminal
-> mkdir cli
-> cd cli
-> stagehand console-full
+$ mkdir cli
+$ cd cli
+$ stagehand console-full
 ```
 
 These commands create a small Dart app that has the following:
@@ -93,7 +87,7 @@ Use the [`pub`](/tools/pub/cmd) command to get the packages
 that the app depends on:
 
 ```terminal
-> pub get
+$ pub get
 ```
 
 ## 6. Run the app
@@ -102,7 +96,7 @@ To run the app from the command line, use the Dart VM by running the
 [`dart`](/tools/dart-vm) command:
 
 ```terminal
-> dart bin/main.dart
+$ dart bin/main.dart
 Hello world: 42!
 ```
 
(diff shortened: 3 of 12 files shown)


src/_includes/get-sdk.md
```diff
@@ -1,9 +1,6 @@
 Once you’re ready to move beyond DartPad and develop real apps,
 you need the Dart SDK.
 
-As you install, **note the path to the SDK.**
-You'll need it in the next step.
```
chalin: I've removed this sentence because, AFAICT, nowhere where this file is included, is there a "next step" that explicitly asks for the "path to the SDK".

src/_includes/get-sdk.md
```diff
@@ -40,9 +37,12 @@ You'll need it in the next step.
...
+  $ brew tap dart-lang/dart
+  $ brew install dart
   ```
 </div>
 
-More information: [Get the Dart SDK](/get-dart)
+{{site.alert.important}}
+  For more information, including how to **adjust your `PATH`**, see
+  [Get the Dart SDK](/get-dart).
+{{site.alert.end}}
```
chalin: I added this alert to address the following need that used to be documented in src/_tutorials/server/get-started.md: "PENDING: the following instructions assume you have set the PATH. We should update the included instructions to refer to that."

src/_includes/mac-path.md
```diff
@@ -0,0 +1,5 @@
+{{site.alert.important}}
+  Ensure that the **Homebrew `bin` directory is in your `PATH`** if you'd like
+  direct command-line access to Dart commands such as the Dart VM (`dart`). For
+  details, consult the [Homebrew FAQ](https://docs.brew.sh/FAQ).
+{{site.alert.end}}
```
chalin: Note: While one my be tempted to instruct the read to add the Dart SDK `bin` folder to their path, that's not the usual way that things are done under Homebrew. Instead on configures `PATH` once to refer to the Homebrew `bin` directory. Following that, all of the standard packages installed with Homebrew will have their commands accessible.

src/_tutorials/server/get-started.md
```diff
@@ -44,24 +44,18 @@ More information:
 
 {% include get-sdk.md %}
 
-<!-- PENDING: the following instructions assume you have set the PATH.
-     We should update the included instructions to refer to that. -->
```
chalin: I believe that I've address the needs expressed here, so I removed the comment.

src/_tutorials/server/get-started.md
```diff
@@ -44,24 +44,18 @@ More information:
...
-> pub global activate stagehand
+$ pub global activate stagehand
 ```
 
-
 Note that although these instructions feature the command line,
 many IDEs support Dart development.
 Those IDEs use Stagehand behind the scenes when you create new Dart projects.
 
-<!-- PENDING: the following instructions assume you have the bin directory for the system cache in your path. -->
```
chalin: Adding the system cache bin to one's path is usually a onetime setup. The "More information" bullet [Running a script from your path](http://localhost:4000/tools/pub/cmd/pub-global#running-a-script-from-your-path) explains how to adjust `PATH`.

Is that good enough? Or should we convert the "More info" bullet into a warning?

src/_tutorials/server/get-started.md
```diff
@@ -102,28 +96,31 @@ To run the app from the command line, use the Dart VM by running the
...
- 1. Edit `lib/cli.dart` to return a different result:
+ 1. Edit `lib/cli.dart` to calculate a different result. For example, divide the
+    previous value by two (see [Arithmetic operators][] for operator details):
 
-    ```dart
+    <?code-excerpt "misc/test/tutorial/get_started.dart (calculate)" replace="/~\/ 2/[!$&!]/g"?>
+    {% prettify dart %}
     int calculate() {
-      return -1;
+      return 6 * 7 [!~/ 2!];
```
chalin: Following Filip's principle of making examples interesting rather than trivial, I changed the return expression (which in this case exposes the reader to the `~/` operator, which they might not know exists). Let me know if this works for you!

src/_tutorials/server/get-started.md
```diff
@@ -139,21 +136,25 @@ Dart code to optimized native machine code.
...
 
 Notice how the compiled program starts instantly, completing quickly:
 
 ```terminal
-> time dartaotruntime bin/main.dart.aot
+$ time dartaotruntime bin/main.dart.aot
+Hello world: 21!
 
-real	0m0.032s
+real	0m0.016s
```
kwalrath: Wow, your computer's a lot faster than mine! Right now I'm getting this (with 2.5.0 stable, but 2.6.0-dev.1.0 is about the same):

```
Hello world: 21!

real	0m0.064s
user	0m0.024s
sys	0m0.038s
```

I wonder why there's such a difference between your and my computer. I hesitate to use your numbers if mine are more representative. (And vice versa...)

src/_includes/mac-path.md
```diff
@@ -0,0 +1,5 @@
+{{site.alert.important}}
+  Ensure that the **Homebrew `bin` directory is in your `PATH`** if you'd like
+  direct command-line access to Dart commands such as the Dart VM (`dart`). For
+  details, consult the [Homebrew FAQ](https://docs.brew.sh/FAQ).
```
kwalrath: FAQ](...). -> FAQ.](...)

src/_includes/mac-path.md
kwalrath: Where else is this file included? (If there's only one include, let's not make it a separate file, for simplicity's sake.)

src/_tutorials/server/get-started.md
kwalrath: I think we're good.

src/_includes/mac-path.md
```diff
@@ -0,0 +1,5 @@
+{{site.alert.important}}
+  Ensure that the **Homebrew `bin` directory is in your `PATH`** if you'd like
```
kwalrath: Hmmm... our style is to have the IF part first, but that seems to bury the important part. How about something like this:

**Make sure the Homebrew...PATH**. Setting up the path correctly makes it easier to use Dart SDK commands such as `dart` and `dartfmt`.

src/tools/pub/troubleshoot.md
```diff
@@ -8,28 +8,30 @@ description: "Common gotchas you might run into when using pub."
 
 You receive the following error when running `pub publish`:
 
-{% prettify none %}
+```terminal
+$ pub publish ...
```
kwalrath: I don't like adding the commands to these, because the emphasis should be on the error, not the command. (also, the preceding text says "the following error", so it seems wrong to include the command)

src/_tutorials/server/get-started.md
```diff
@@ -139,21 +136,25 @@ Dart code to optimized native machine code.
...
-> dart2aot bin/main.dart bin/main.dart.aot
+$ dart2aot bin/main.dart bin/main.dart.aot
 ```
 
 To run the compiled program, use the Dart runtime (`dartaotruntime`):
 
 ```terminal
-> dartaotruntime bin/main.dart.aot
+$ dartaotruntime bin/main.dart.aot
+Hello world: 21!
```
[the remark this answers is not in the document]
chalin: > we didn't include it in the previous command

Just to be clear, the previous command (`dart2aot bin/main.dart bin/main.dart.aot`) doesn't output anything (which is why nothing is shown).

> Let's not include the output here. It's a little distracting, 

I find it reassuring to have _something_ to indicate that output _is_ generated, even if the details are elided. Maybe `...` might be less distracting?

```console
$  dartaotruntime bin/main.dart.aot
...
```

But, if you prefer nothing at all, I'll go with that.

src/_includes/mac-path.md
chalin: Oh, I thought that I had at least two files. I'll double check and merge the file if there is only one point of use.

src/tools/pub/troubleshoot.md
chalin: Ok, I'll take them out.

src/_tutorials/server/get-started.md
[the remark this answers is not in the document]
kwalrath: We have other places, like step 3 & step 4, that don't list the output.

src/_includes/get-sdk.md
```diff
@@ -23,15 +20,15 @@ You'll need it in the next step.
 
    1. Perform the following one-time setup:
       ```terminal
-      > sudo apt-get update
-      > sudo apt-get install apt-transport-https
-      > sudo sh -c 'curl https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add -'
-      > sudo sh -c 'curl https://storage.googleapis.com/download.dartlang.org/linux/debian/dart_stable.list > /etc/apt/sources.list.d/dart_stable.list'
+      $ sudo apt-get update
+      $ sudo apt-get install apt-transport-https
+      $ sudo sh -c 'curl https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add -'
```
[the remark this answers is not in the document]
chalin: I've adjusted the string and name styles, and ensured that the contrast is good enough:

>

src/_tutorials/server/get-started.md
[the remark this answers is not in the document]
chalin: Output removed.

src/_tutorials/server/get-started.md
chalin: Keeping the new numbers, as agreed out of band.

src/_includes/mac-path.md
chalin: I think that I had originally anticipated including this file directly in `tutorials/server/get-started` (as well as `get-dart`). But having it only in `get-dart`, like we have it now in this PR, seems sufficient. Merging the file.

src/_includes/mac-path.md
chalin: How about this:
>   Make sure the **Homebrew `bin` directory is in your `PATH`**. Setting up the path correctly makes it easier to use Dart SDK commands such as `dart` and `dartfmt`. For help setting up your path, consult the [Homebrew FAQ.](https://docs.brew.sh/FAQ)

src/tools/pub/troubleshoot.md
chalin: I've removed the commands, reverting to what was there on the page before my edits. But I've managed to find a way to keep the "terminal output" styling (which is what I was looking for before). E.g.:

> image

as compared to what we had before:

> image

WDYT?

chalin:
Changes committed. I left two questions related to the latest round of changes in the inlined comments. Staging server refreshed.

src/tools/pub/troubleshoot.md
kwalrath: LGTM

Issue #1922, opened 2019-09-19
'Asynchronous programming: streams' page issue


iChicago:
Page URL: https://dart.dev/tutorials/language/streams
Page source: https://github.com/dart-lang/site-www/tree/master/src/_tutorials/language/streams.md

Found a typo? You can fix it yourself by going to the page source and clicking the pencil icon. Or finish creating this issue.

Description of issue:

In the section *"Methods that modify a stream"* you added the method 
```dart Stream<R> retype<R>();```
I think this method was removed Stream<T> Class.
https://api.dart.dev/stable/2.5.0/dart-async/Stream-class.html

kwalrath:
Fixed by #1923. Thank you, @iChicago!

PR #1923, opened 2019-09-19
removed retype() method from Stream class

iChicago:
Hello, 
This is a fix for this issue 
https://github.com/dart-lang/site-www/issues/1922

### DIFF
--- a/src/_tutorials/language/streams.md
+++ b/src/_tutorials/language/streams.md
@@ -285,7 +285,6 @@ listening on the original.
 Stream<R> cast<R>();
 Stream<S> expand<S>(Iterable<S> Function(T element) convert);
 Stream<S> map<S>(S Function(T event) convert);
-Stream<R> retype<R>();
 Stream<T> skip(int count);
 Stream<T> skipWhile(bool Function(T element) test);
 Stream<T> take(int count);


kwalrath:
Good catch, thanks for the fix!

I'm going to investigate why the build failed before I commit this, but I don't think your change was the culprit.

chalin:
> I'm going to investigate why the build failed ...

As you probably noticed already, the change needed to be done in the originating source file: `misc/lib/tutorial/stream_interface.dart`.

kwalrath:
> > I'm going to investigate why the build failed ...
> 
> As you probably noticed already, the change needed to be done in the originating source file: `misc/lib/tutorial/stream_interface.dart`.

Nope, I did NOT notice. I thought it was just a plain old list. I'll fix the build. Thanks, @chalin!

chalin:
That source file can probably be tweaked to ensure that the analyzer catches errors like the one fixed by this PR (a method removed from an API). I'm not sure if it is worth it though.

kwalrath:
> That source file can probably be tweaked to ensure that the analyzer catches errors like the one fixed by this PR (a method removed from an API). I'm not sure if it is worth it though.

Were you thinking of `@override` annotations? I thought about that but then figured that, like you said, it wasn't worth the trouble for an API that's been so stable (with this one exception).

chalin:
> Were you thinking of `@override` annotations?

Yes.

> ... [not] worth the trouble for an API that's been so stable

Agreed.

Issue #1894, opened 2019-09-11
Setup analyzer-results files based on Dart channel rather than Dart version


chalin:
Just before Dart 2 was released, the site build was adapted so that it could run under Dart 1 and Dart 2. In particular, this implied supporting the analysis and testing of the site examples under both versions of Dart. Part of the challenge was to support:

- Analyzer options that differed across versions
- Analyzer output that differed across versions

This issue concerns the adaptation of the build scripts to support analyzer-results variants that differ across Dart channels (or some similar criteria), rather than Dart major version.

This will allow us to make the dev channel example tests to pass once again.

Remember to drop the allow-failure of dev channel examples tests after this issue is fixed.

kwalrath:
I'd love to have this working!

PR #1925, opened 2019-09-19
Fix the build

kwalrath:
Follow up to #1923


kwalrath:
The dev build is still broken, for some reason, but fixing that isn't urgent.

chalin:
We need to address #1894 in order to fix the `dev` build.

PR #1933, opened 2019-09-20
Drop reference to Jekyll conf var that doesn't exist

chalin:
The `site.dartlang` Jekyll conf variable doesn't exist. This is a vestige from the page migration of #1764.

The site generated before and after this PR are identical.


chalin:
cc @kwalrath

kwalrath:
Thanks!

PR #1902, opened 2019-09-15
Analyze/test cmdline tutorial code excerpts

chalin:
Resolves first item of #1803. (Builds on #1900, #1901.)

Staged at https://dart-dev-staging-0.firebaseapp.com/tutorials/server/cmdline
Original https://dart.dev/tutorials/server/cmdline

---

- Added `dcat.dart` example, which comes from https://github.com/dart-lang/dart-tutorials-samples/tree/master/cmdline/bin (we should consider removing it from that repo).
- Created other examples sources (not found under dart-tutorials-samples, etc.) as necessary.
- Updated cmdline.md page links to ensure that they all reference the proper API version (based on the channel). Also ensured that all links to external sites consistently open in a new tab. (We aren't doing that elsewhere, but at least the links in this page will be consistent in their behavior).

**Note**: `dcat`, and the rest of the newly added code, is analyzed but NOT tested; we don't test other command line apps in this repo -- doing so would require some refactoring or additional recripting, which probably isn't worth it at this point.

I've ensured that all **external links** in the page are valid:
```nocode
~ > blc -fgi http://localhost:4000/tutorials/server/cmdline
Getting links from: http://localhost:4000/tutorials/server/cmdline
├───OK─── http://www.google.com/intl/en/policies/privacy/
├───OK─── https://dartpad.dev/
├───OK─── https://pub.dev/
├───OK─── https://dart-lang.github.io/server
├───OK─── https://flutter.dev/
├───OK─── https://pub.dev/packages/args
├───OK─── http://creativecommons.org/licenses/by/3.0/
├───OK─── https://api.dart.dev/
├───OK─── https://github.com/dart-lang/site-www/tree/master/src/_tutorials/server/cmdline.md
├───OK─── https://pub.dev/documentation/args/latest/args/ArgParser-class.html
├───OK─── https://medium.com/dartlang/announcing-dart-2-5-super-charged-development-328822024970
├───OK─── https://github.com/dart-lang/site-www/issues/new?...
├───OK─── https://api.dart.dev/stable/dart-async/Future-class.html
├───OK─── https://medium.com/dartlang
├───OK─── https://pub.dev/documentation/args/latest/args/ArgResults-class.html
├───OK─── https://github.com/dart-lang/site-www
├───OK─── https://pub.dev/documentation/args/latest/
├───OK─── https://api.dart.dev/stable/dart-async/Stream-class.html
├───OK─── https://github.com/dart-lang/site-www/issues
[...]

PR #1945, opened 2019-09-23
Add a basic test for each of the three "cat" commands

### LINKED ISSUE #1927: Examples: move cat_no_hash.dart to misc/bin
Consider moving `misc/lib/tutorial/cat_no_hash.dart` to `misc/bin` (and update the path in the streams.md file) so that it is with the newly added `dcat` apps.

chalin:
Contributes to #1803.
Followup to #1902.
Also Fixes #1927.


chalin:
Thanks @johnpryan!

PR #1953, opened 2019-09-26
Update example pubspecs to 2.5 sdk

### LINKED ISSUE #1906: Update wiki & pubspec (2.5)
For details, see #1898.

kwalrath:
Fixes #1906


kwalrath:
Closing because, after talking with @chalin, I don't think we need to update the sdk lower bounds with every release. I'll update the wiki, too.

PR #1954, opened 2019-09-26
Update FFI page to reflect new flutter.dev FFI page

kwalrath:
Staged: https://kw-www-dartlang-1.firebaseapp.com/guides/libraries/c-interop

/cc @sfshaza2

### DIFF
--- a/src/_guides/libraries/c-interop.md
+++ b/src/_guides/libraries/c-interop.md
@@ -8,29 +8,33 @@ sqllite: "https://github.com/dart-lang/sdk/tree/master/samples/ffi/sqlite"
 
 Dart mobile, command-line, and server apps running on the [Dart Native
 platform](/platforms/) can use the dart:ffi library to call native C APIs.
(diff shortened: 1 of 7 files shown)


src/_guides/libraries/c-interop.md
```diff
@@ -139,10 +143,12 @@ Once you understand the hello_world example, you should be ready to look at the
 How you bundle (or _package_ or _distribute_)
 a C library with your package or app and then load that library
 depends on your platform and the type of library.
-For details, see [Binding to native code via FFI.][binding]
+For details, see the following:
```
mit-mit: I suggest we add something about how to bundle in pure Dart apps, specifically how to do it for an app compiled with `dart2native`

kwalrath: Would you use dart2aot before 2.6?

kwalrath:
I'm going to merge now. We can add something about dart2aot/dart2native later.

PR #1964, opened 2019-10-01
Update analysis-options.md with info about using 2 YAML formats

### LINKED ISSUE #1913: Customizing static analysis > Disabling linter rules: using 2 YAML formats
Page URL: https://dart.dev/guides/language/analysis-options
Page source: https://github.com/dart-lang/site-www/tree/master/src/_guides/language/analysis-options.md

[Disabling individual rules](https://dart.dev/guides/language/analysis-options#disabling-individual-rules) has the following aside:

> **If you disable any rules, then use key-value syntax to enable rules**. For example, the preceding example can’t use - await_only_futures; instead, it must use await_only_futures: true. The reason it can’t use the dash shortcut (-) to enable await_only_futures is that YAML doesn’t support mixing list and key-value syntaxes.

This is true if all of the linter rules are in the same options file, but that doesn't need to be the case. The YAML list style can be used in the main options file with the key-value format used in an included file. An example of this is included in #1912.

Adjust the note and/or text.

KNawm:
Fixes #1913

Should I put examples in this or looks clear enough?

### DIFF
--- a/src/_guides/language/analysis-options.md
+++ b/src/_guides/language/analysis-options.md
@@ -294,7 +294,7 @@ linter:
 ```
 
 {{site.alert.note}}
-  Due to YAML restrictions, you can't mix list and key-value syntax in the same rules entry.
+  Due to YAML restrictions, **you can't mix list and key-value syntax in the same rules entry.**
   You can, however, use the other syntax for rules in an included file.
 {{site.alert.end}}
 
--- a/src/_guides/language/analysis-options.md
+++ b/src/_guides/language/analysis-options.md
@@ -295,8 +295,8 @@ linter:
 
 {{site.alert.important}}
   **Don't mix list and key-value syntaxes in the same options file.**
-  YAML doesn't support mixing both syntaxes,
-  however you can use the other syntax in another YAML file
+  YAML doesn't support mixing both syntaxes.
+  However, you can use the other syntax in another YAML file
   and add it to your main options file using `include:`.
 {{site.alert.end}}
 
--- a/src/_guides/language/analysis-options.md
+++ b/src/_guides/language/analysis-options.md
@@ -293,12 +293,11 @@ linter:
     await_only_futures: true
 ```
 
-{{site.alert.note}}
-  **If you disable any rules, then use key-value syntax to enable rules.**
-  For example, the preceding example can’t use `- await_only_futures`;
-  instead, it must use `await_only_futures: true`.
-  The reason it can’t use the dash shortcut (`-`) to enable `await_only_futures`
-  is that YAML doesn't support mixing list and key-value syntaxes.
+{{site.alert.important}}
+  **Don't mix list and key-value syntaxes in the same options file.**
+  YAML doesn't support mixing both syntaxes,
+  however you can use the other syntax in another YAML file
+  and add it to your main options file using `include:`.
 {{site.alert.end}}
 
 ## Excluding code from analysis
--- a/src/_guides/language/analysis-options.md
+++ b/src/_guides/language/analysis-options.md
@@ -294,8 +294,8 @@ linter:
 ```
 
 {{site.alert.important}}
-  **Don't mix list and key-value syntaxes in the same options file.**
-  YAML doesn't support mixing both syntaxes.
+  **Don't mix list and key-value syntaxes for the rules in the same options file.**
+  YAML doesn't support mixing both syntaxes in the same collection.
   However, you can use the other syntax in another YAML file
   and add it to your main options file using `include:`.
 {{site.alert.end}}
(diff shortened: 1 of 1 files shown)


src/_guides/language/analysis-options.md
```diff
@@ -293,12 +293,11 @@ linter:
...
 ```
 
-{{site.alert.note}}
-  **If you disable any rules, then use key-value syntax to enable rules.**
-  For example, the preceding example can’t use `- await_only_futures`;
-  instead, it must use `await_only_futures: true`.
-  The reason it can’t use the dash shortcut (`-`) to enable `await_only_futures`
-  is that YAML doesn't support mixing list and key-value syntaxes.
+{{site.alert.important}}
+  **Don't mix list and key-value syntaxes in the same options file.**
```
kwalrath: in -> for linter rules in
?
I think you can mix the syntaxes, as long as they aren't modifying the same tag.

KNawm: You can't even if they're different tags as it would give a parsing error

kwalrath: Oh, that's so counterintuitive! @pq can you confirm?

pq: Hmmm. IIRC @KNawm is right. 😬

chalin: I'd be really surprised if such a constraint existed. In fact, the Dart analyzer (2.5.1) is perfectly happy with the following YAML, which contains both syntaxes:

```yaml
include: package:pedantic/analysis_options.yaml

linter:
  rules:
    camel_case_types: false

analyzer:
  exclude:
    - path/to/excluded/files/**
    - some_other_path/file.*
```

```terminal
$ dartanalyzer analysis_options.yaml .
Analyzing analysis_options.yaml, ...
No issues found!
```

Can you give an example of the kind of error that you are getting?

KNawm: Oh, I was talking about the rules collection specifically, so maybe it isn't clear enough.
I'm talking about this:
```yaml
linter:
  rules:
    - annotate_overrides
    - await_only_futures
    - camel_case_types
    avoid_shadowing_type_parameters: false
    await_only_futures: true
```

src/_guides/language/analysis-options.md
```diff
@@ -293,12 +293,11 @@ linter:
...
 
-{{site.alert.note}}
-  **If you disable any rules, then use key-value syntax to enable rules.**
-  For example, the preceding example can’t use `- await_only_futures`;
-  instead, it must use `await_only_futures: true`.
-  The reason it can’t use the dash shortcut (`-`) to enable `await_only_futures`
-  is that YAML doesn't support mixing list and key-value syntaxes.
+{{site.alert.important}}
+  **Don't mix list and key-value syntaxes for the rules in the same options file.**
+  YAML doesn't support mixing both syntaxes in the same collection.
```
KNawm: Thoughts on this phrasing?

chalin: Hi. Thanks for working on this issue. As for feedback ...

The inlined heading isn't accurate. Working from what you propose, how about something like this:

> **YAML doesn't support mixing both list and key-value syntax in the same rules entry.** 
> However, your are free to use any syntax for `rules` appearing in an included options file.

@kwalrath WDYT? Also do you agree with the use of "syntax" in "singular" form there?

Finally, if we all agree with the updated note text, then I'd suggested switching back to `alert.note`.

kwalrath:
How about this:

{{site.alert.note}}
Due to YAML restrictions, **you can't mix list and key-value syntax in the same `rules` entry.**
You can, however, use the other syntax for rules in an included file.

src/_guides/language/analysis-options.md
chalin: Kathy's suggestion LGTM

src/_guides/language/analysis-options.md
```diff
@@ -294,11 +294,8 @@ linter:
 ```
 
 {{site.alert.note}}
-  **If you disable any rules, then use key-value syntax to enable rules.**
-  For example, the preceding example can’t use `- await_only_futures`;
-  instead, it must use `await_only_futures: true`.
-  The reason it can’t use the dash shortcut (`-`) to enable `await_only_futures`
-  is that YAML doesn't support mixing list and key-value syntaxes.
+  Due to YAML restrictions, **you can't mix list and key-value syntax in the same rules entry.**
```
chalin: Nit: ... same rules entry -> ... same `rules` entry 
(to match Kathy's proposal).

chalin: Otherwise, LGTM

chalin:
LGTM. Since you've implemented @kwalrath's suggestion, I'll assume that she's still ok with the change. Merging.

Thanks for the PR @KNawm!

Issue #1807, opened 2019-08-02
Todo app missing in "Add elements to the DOM"


kwalrath:
The first example app (Todo) in https://dart.dev/tutorials/web/low-level-html/add-elements is missing. You should see a DartPad, but instead, you see "Example goes here..."

The example disappeared when we moved the web tutorials to webdev.dartlang.org. (The basic web tutorials have since moved back to dart.dev.)

You can see what it originally looked like here, back when it ran in the page:

http://web.archive.org/web/20130318084500/http://www.dartlang.org:80/docs/tutorials/add-elements

The latest www.dartlang.org version (which uses DartPad) is here:

http://web.archive.org/web/20160702202609/https://www.dartlang.org/docs/tutorials/add-elements/

The first webdev.dartlang.org version is here:

http://web.archive.org/web/20160926070243/https://webdev.dartlang.org/tutorials/low-level-html/add-elements

kwalrath:
The good news is that the first app in https://dart.dev/tutorials/web/low-level-html/remove-elements is working — and that app is based on the one in add-elements. So it's just a matter of locating the source code, creating a new gist if necessary, and adding the DartPad to add-elements.

smolck:
@kwalrath I’d like to help with fixing this. The only difference between the app in https://dart.dev/tutorials/web/low-level-html/remove-elements and the missing one is the addition of a delete button, right? If that’s the case, I presume it could be fixed by copying the code in the `remove-elements` tutorial into a gist and then removing the code for the delete button?

kwalrath:
I think this might be it:

https://github.com/dart-lang/dart-tutorials-samples/tree/master/todo

smolck:
> I think this might be it:
> 
> https://github.com/dart-lang/dart-tutorials-samples/tree/master/todo

Great, thanks!

I'm not sure how I need to go about creating the gist (I've never done this before) for the add-elements page. This is the part I'm not sure about, in the [`dart-tutorials-samples`](https://github.com/dart-lang/dart-tutorials-samples) README:

> Run "gist generate". See the readme for information on available options. If you are updating, or creating, gist files rather than generating a test gist, or performing a dry run of the gist generator, you'll need to provide a Gist token. (See the project lead for info on obtaining an existing Gist token.)

Do I need a Gist token in order to generate the gist to add to the add-elements page?

kwalrath:
[...]

PR #1967, opened 2019-10-02
Add missing example in add-elements

smolck:
See #1807.

### DIFF
--- a/src/_tutorials/web/low-level-html/add-elements.md
+++ b/src/_tutorials/web/low-level-html/add-elements.md
@@ -61,10 +61,8 @@ Then type in the app's text field, and press return.
 The app adds an item to the list.
 Enter a few items into the input field.
 
-Example goes here...
-
-<!-- need to fix this section, not working properly -->
-<!-- https://gist.github.com/Sfshaza/65c90ff6b078c332d4b6
+{% comment %}
+https://gist.github.com/kwalrath/aab6ad8c04d161458d32161ea6e87a97
 
 main.dart:
 // Copyright (c) 2012, the Dart project authors.
--- a/src/_tutorials/web/low-level-html/add-elements.md
+++ b/src/_tutorials/web/low-level-html/add-elements.md
@@ -61,40 +61,12 @@ Then type in the app's text field, and press return.
 The app adds an item to the list.
 Enter a few items into the input field.
 
-{% comment %}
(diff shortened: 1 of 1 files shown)


kwalrath:
Did you mean to upload this? It looks like this change just comments out the rest of the file. We need to embed a DartPad that references the new gist. Once you add that, I'll stage the page.

smolck:
> Did you mean to upload this? It looks like this change just comments out the rest of the file. We need to embed a DartPad that references the new gist. Once you add that, I'll stage the page.

Oops! Sorry about that! I misunderstood how the DartPad was embedded...fixed it.

However, when I tested it through the browser's dev tools (inserted the new id in an already-embedded DartPad), it didn't work. That's probably why the example was removed in the first place (seems that way from the comments left behind).

smolck:
> You can see it staged here: https://kw-staging-dartlang-2.firebaseapp.com/tutorials/web/low-level-html/add-elements

@kwalrath Does the code show up in the DartPad for you? When I visit the staged page it is just an empty DartPad with the text "UI Output" in it.

kwalrath:
Yes, it works for me, although I have to run it to get the output. Here's what I see:

Before clicking **Run**:
image

After clicking **Run**:
image

kwalrath:
Can you see the code at https://gist.github.com/kwalrath/aab6ad8c04d161458d32161ea6e87a97? I believe it's a public gist. Is it possible you've loaded gists so much in the past hour that you've been temporarily shut out?

Also... It looks like we need to do a couple of things that I could/should file as issues:

* Add an include for the DartPad troubleshooting page high up in this page (probably deleting the include that's lower in this page).
* Add info to the DartPad troubleshooting page about what happens if you load too many DartPads and run into GitHub rate-limiting. (Rare unless you're repeatedly reloading a page full of DartPads.)

smolck:
> Can you see the code at https://gist.github.com/kwalrath/aab6ad8c04d161458d32161ea6e87a97? I believe it's a public gist. Is it possible you've loaded gists so much in the past hour that you've been temporarily shut out?

It appears to be a problem with Brave or DartPad itself. Everything works fine in FireFox Dev Edition and IOS Safari. I may need to look into why things don't show with Brave.

kwalrath:
Ah, that's a known problem. Brave disables cookies, which disables localStorage, which DartPad depends on.

Issue #1955, opened 2019-09-26
Clarify C/C++/... interop


kwalrath:
The Dart & Flutter FFI pages should be clearer about what languages you can interface to. E.g., currently the [Dart page](https://dart.dev/guides/libraries/c-interop) says "C interop", but the [Flutter page](https://flutter.dev/docs/development/platform-integration/c-interop) says "C/C++ interop".

According to @sjindel-google (from [this comment](https://github.com/flutter/website/issues/2941#issuecomment-535693884)):

> FFI can bind to a C _interface_, regardless of whether that interface is implemented by C, C++, Swift, Objective-C, Rust or even Go.

/cc @mjohnsullivan, @sfshaza2, @mit-mit

mit-mit:
I suggest we just use 'C interop' everywhere

dhanushuUzumaki:
Hello @kwalrath can I take this one? Should I update C/C++ interop to C interop in flutter docs?
https://github.com/flutter/website/blob/1fbbdc481550b9d8bc3b5ef29ca695575a839f4b/src/_data/sidenav.yml#L128

kwalrath:
@dhanushuUzumaki I'm in the midst of figuring this one out, but you updating C/C++ interop to C interop in the Flutter docs SGTM (sounds good to me) if it's OK by @sfshaza2.

But first, let's wait until my PR (incoming) for this page is approved. That'll give you a model to follow.

dhanushuUzumaki:
@kwalrath Sure.

kwalrath:
Moving to flutter/website, since I believe that https://flutter.dev/docs/development/platform-integration/c-interop should use "C interop", to match what you see at https://dart.dev/guides/libraries/c-interop

PR #1973, opened 2019-10-03
[WIP] Add info about non-C languages you might use with FFI.

kwalrath:
~~Staged: https://kw-staging-dartlang-2.firebaseapp.com/guides/libraries/c-interop~~

Contributes to #1955

### DIFF
--- a/src/_guides/libraries/c-interop.md
+++ b/src/_guides/libraries/c-interop.md
@@ -12,6 +12,13 @@ _FFI_ stands for [_foreign function interface._][FFI]
 Other terms for similar functionality include _native interface_
 and _language bindings._
 
+{{ site.alert.info }}
+Although dart:ffi requires a C interface,
+you aren't restricted to using only C libraries.
+Many languages — including C++, Swift, Objective-C, Rust, and Go — 
+can implement C interfaces.
+{{ site.alert.end }}
+
 {{ site.alert.warning }}
 The dart:ffi library is [in active development][ffi issue]
 and isn't complete yet.


src/_guides/libraries/c-interop.md
```diff
@@ -12,6 +12,13 @@ _FFI_ stands for [_foreign function interface._][FFI]
 Other terms for similar functionality include _native interface_
 and _language bindings._
 
+{{ site.alert.info }}
+Although dart:ffi requires a C interface,
+you aren't restricted to using only C libraries.
+Many languages — including C++, Swift, Objective-C, Rust, and Go — 
+can implement C interfaces.
```
mit-mit: I'm not sure 'implement' is the right word here? Maybe 'expose'?

src/_guides/libraries/c-interop.md
```diff
@@ -12,6 +12,13 @@ _FFI_ stands for [_foreign function interface._][FFI]
 Other terms for similar functionality include _native interface_
 and _language bindings._
 
+{{ site.alert.info }}
```
mit-mit: Overall I think this may cause more questions than offer answers. Is there something we can link to regarding how to expose a C interface from these various languages?

kwalrath: @mjohnsullivan any suggestions?

kwalrath: I see some relevant C++ topics in Stackoverflow, such as https://stackoverflow.com/questions/2045774/developing-c-wrapper-api-for-object-oriented-c-code.

I haven't found anything similar for Swift.

https://en.wikipedia.org/wiki/Foreign_function_interface#By_language seems like it might be a good resource.

mit-mit: @sjindel-google @dcharkes are you aware of any resources we can link to?

sjindel-google: For Go: https://golang.org/cmd/cgo/
For Swift: https://stackoverflow.com/questions/35106118/what-is-the-best-way-to-call-into-swift-from-c, not officially supported

kwalrath:
I plan to open a new PR with these changes after 2.6. Until then, I'll keep this PR open in case anyone has more suggestions.

kwalrath:
Closing because this has languished too long.

PR #1974, opened 2019-10-04
Slightly subdue the analyzer error formatting

kwalrath:
I found myself expecting a mouseover or click to do something.
This change removes the yellow background and moves up the
dashed line a bit to compensate.

Before:
image

After:
image


chalin:
I did consider added a tooltip popup there, but that's for another day maybe :).
I kinda liked the standard yellow highlight, but your changes do make it look a bit cleaner.

PR #1979, opened 2019-10-04
Fix issue in shared-pkgs tutorial

wang-ories:
Update Pub site reference.


wang-ories:
@googlebot I signed it!

Issue #1966, opened 2019-10-01
More information in web tutorials (Connect dart with html)


wang-ories:
When editing the HTML source code in this tutorial, https://dart.dev/tutorials/web/low-level-html/connect-dart-html#create-html, the checkbox **Show web content** is not checked by default. The **HTML**, at the upper left of DartPad is not visible. 
**This can be an issue for new dev in Dartpad environment only.**
A little precision in this section can be more helpful.

dhanushuUzumaki:
Hello @kwalrath Can I take this up? Will it be good to have **show web content** enabled by default? https://github.com/dart-lang/dart-pad/blob/0b4156655fc07d18b0ee4f45bfe2e7005ba80e92/web/index.html#L133

Regards,
Dhanushu A P

wang-ories:
Someone already fixing this? I think we can just notify the user to verify that **Show web content** is checked. For those who are familiar with dartpad it is not an issue.

kwalrath:
Sorry for the delay. Since @dhanushuUzumaki asked to handle this, I'll give him first crack.

I think we need to add a new step in the "Create a new Dart app" that explicitly tells the reader to make sure that **Show web content** is selected.

If there's a way to select it automatically, all the better, but I think it's good to make sure that they know that that checkbox needs to be selected.

dhanushuUzumaki:
@kwalrath So I have to Add an explicit message to let the users know they have to enable show web content (https://github.com/dart-lang/site-www/blob/master/src/_tutorials/web/low-level-html/connect-dart-html.md#create-a-new-dart-app-create-dart-app) and also enable it by default. Please let me know if I am misinterpreting this.

Regards,
Dhanushu A P

kwalrath:
That sounds right to me, @dhanushuUzumaki.

PR #1980, opened 2019-10-05
Patch #1966

dhanushuUzumaki:


kwalrath:
@wang-ories or @dhanushuUzumaki, did you have a way to select the checkbox by default?

wang-ories:
@kwalrath  I already saw a similar way where the  the checkbox is selected  by default.

dhanushuUzumaki:
@kwalrath The change needs to be done in dart-pad. I raise a PR today.

dhanushuUzumaki:
@kwalrath I made the change to show it by default. But there is already a check which shows the web content by default if the gist has any web content. Do we want to override this behavior and show web content no matter the gist contains one or not?

kwalrath:
> @kwalrath I made the change to show it by default. But there is already a check which shows the web content by default if the gist has any web content. Do we want to override this behavior and show web content no matter the gist contains one or not?

Nope, by default we shouldn't show web content unless there's an HTML file.

dhanushuUzumaki:
> > @kwalrath I made the change to show it by default. But there is already a check which shows the web content by default if the gist has any web content. Do we want to override this behavior and show web content no matter the gist contains one or not?
> 
> Nope, by default we shouldn't show web content unless there's an HTML file.

This functionality is already in. So for the scope of #1966 I think adding this message will be enough.

PR #1987, opened 2019-10-08
Fix homepage layout on very wide screens

### LINKED ISSUE #1969: homepage looks bad when window is very wide
URL: https://dart.dev

@mit-mit noticed that when you have a large monitor and maximize the window, you end up with the homepage flowing to the full width, and the "proof points" looking very enlarged.

The designer says that we should limit the max-height to 644px, and allow for empty space on each side rather than scaling the images to the width of the container. Here's a mockup:

image

@johnpryan would this be easy to fix?

johnpryan:
also disable text selection in the bulleted info sections

staged: https://jr-dartlang-1.firebaseapp.com

closes #1969

### DIFF
--- a/src/_assets/css/_dash.scss
+++ b/src/_assets/css/_dash.scss
@@ -1,3 +1,11 @@
+// Homepage styles
+
+// Sizes
+$dash-row-height: 640px;
+$dash-info-size: 512px;
+$dash-info-padding: 24px;
+
+// Colors
 $dash-callout: #121a26;
 $dash-sub-grey: #8d9399;
 $dash-dark-grey: #82878c;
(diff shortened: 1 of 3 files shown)


johnpryan:
Just found an issue where the images don't change. Investigating...

src/_assets/css/_dash.scss
```diff
@@ -179,11 +198,25 @@ $dash-dartpad-border: #293542;
 
   .bullet-text{
     margin-left: 20px;
+
+    // Disable text selection to avoid blue focus outline on macOS
```
kwalrath: Does this reduce accessibility? (Also, selfishly, I often copy text, so anything that makes that impossible worries me.)

johnpryan:
No problem, fixed. This is ready for another look

PR #1989, opened 2019-10-09
Add missing command line parameter

wang-ories:
Tell people to download quote.txt file and add the link to this file.

### DIFF
--- a/src/_tutorials/server/cmdline.md
+++ b/src/_tutorials/server/cmdline.md
@@ -91,20 +91,15 @@ brief description of key app features, click the highlighted code below.
 
 {% include_relative _dcat-example.html %}
 
-Here's an example of running the app from the command line:
+You can run the app from the command line over the quote.txt file (which you can download from this [repo] ({{site.repo.this}}/blob/master/examples/misc/test_data/quote.txt?raw=1){:target="_blank"}) or any other text file, as shown here:
 
- 1. Download the file called [quote.txt]({{site.repo.this}}/blob/master/examples/misc/test_data/quote.txt){:target="_blank"} and save it.
- 
- 2. In the directory that contains the file quote.txt and dcat.dart, run the program:
- 
-    ```terminal
-     $ dart dcat.dart -n quote.txt
-     1 Be yourself. Everyone else is taken. -Oscar Wilde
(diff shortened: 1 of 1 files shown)


src/_tutorials/server/cmdline.md
```diff
@@ -93,13 +93,17 @@ brief description of key app features, click the highlighted code below.
...
+ 
+ 2. In the directory that contains the file quote.txt and dcat.dart, run the program:
+ 
+    ```terminal
+     $ dart dcat.dart -n quote.txt
+     1 Be yourself. Everyone else is taken. -Oscar Wilde
+     2 Don't cry because it's over, smile because it happened. -Dr. Seuss
+     3 You only live once, but if you do it right, once is enough. -Mae West
+     ...
+     ```
```
chalin: Hi. Thanks for the PR!

Making the download of `quote.txt` a _required_ step feels a little to heavy and unnecessarily constraining. How about simply replacing

> Here's an example of running the app from the command line:

with something like this:

> You can run the app from the command line over the `quote.txt` file (which you can [download from this repo](#put-the-right-link-here)) or any other text file, as shown here:

Good job on the encoding of the link. I'd suggest appending `?raw=1` to the end of that link so that the reader is presented with the raw content immediately.

WDYT? @kwalrath?

src/_tutorials/server/cmdline.md
```diff
@@ -91,7 +91,7 @@ brief description of key app features, click the highlighted code below.
 
 {% include_relative _dcat-example.html %}
 
-Here's an example of running the app from the command line:
+You can run the app from the command line over the quote.txt file (which you can download from this [repo] ({{site.repo.this}}/blob/master/examples/misc/test_data/quote.txt?raw=1){:target="_blank"}) or any other text file, as shown here:
```
kwalrath: We try to keep lines 80 characters or less. (Not possible here, perhaps, but still, the lines could be shorter.

Also, how about we make the link more clearly to the quote.txt file? Also, make sure there's no space between the `]` and the `(`. Maybe:

```
You can run the app from the command line over `quote.txt`
([downloadable file]({{site.repo.this}}/blob/master/examples/misc/test_data/quote.txt?raw=1){:target="_blank"})
or any other text file:
```

PR #1986, opened 2019-10-08
Verified publishers docs

mit-mit:
Initial publisher docs for pub.dev Verified Publishers.

TODO:
  - [x] Complete steps for creating
  - [x] Link to a sample publisher
  - [x] Add docs for consumers of packages

Staged:
https://kw-www-dartlang-1.firebaseapp.com/tools/pub/publishing
https://kw-www-dartlang-1.firebaseapp.com/tools/pub/verified-publishers

PR #1990, opened 2019-10-09
Update landing page native visualization

mit-mit:
This should not be merged until D26 launches


kwalrath:
Staged: https://kw-www-dartlang-1.firebaseapp.com/tools/pub/publishing

kwalrath:
I did a pretty thorough hatchet job. :) Along the way, I had some questions/thoughts about terminology:

* Can we avoid saying "[verified] publisher **account**"? It sounds like Google Account, but it's just a pub.dev thing, right? Does "account" show up in the pub.dev UI anywhere?
* What's the **user menu**? ("In the user menu in the top-right corner, select **Create Publisher**.") Is there a better name for it? An icon? Alt text? (For formatting the description, I'd follow the precedent in https://developers.google.com/style/ui-elements#term-menus.)
* At one point the text refers to **adding DNS records** ("When adding DNS records, it may take a few hours before the Search Console reflects the changes.") Can we be more general?

mit-mit:
I think these comments are for https://github.com/dart-lang/site-www/pull/1986 ?

PR #1995, opened 2019-10-12
Update testing.md

ThinkDigitalSoftware:
Fix broken link to mockito iss tests

### DIFF
--- a/src/_guides/testing.md
+++ b/src/_guides/testing.md
@@ -76,7 +76,7 @@ following packages are useful across Dart platforms:
   expected ways.
   For an example that uses both package:test and package:mockito,
   see the [International Space Station API library and its unit
-  tests](https://github.com/dart-lang/mockito/tree/master/test/example/iss)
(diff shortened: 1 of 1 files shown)


kwalrath:
Thanks for pointing this out! We can't auto-test links to GitHub, so it's very helpful when people tell us about these problems.

Question for @srawlins: Should the link actually be https://github.com/dart-lang/mockito/tree/master/test?

ThinkDigitalSoftware:
@kwalrath that doesn't contain tests for iss though. It seems like the word tests was accidentally added to the path in the wrong place as I found the iss tests in the folder I linked to

srawlins:
Hi @kwalrath , no I don't think so. The text,

> see the International Space Station API library and its unit tests

means we should link to https://github.com/dart-lang/mockito/tree/master/example/iss, which points to both the lib and tests.

kwalrath:
Thanks, @ThinkDigitalSoftware and @srawlins!

Issue #1929, opened 2019-09-20
Change default <code> color from RED to something else


chalin:
In general, the color of `<code>` elements under the new site design is the default `pretty-print` red. Red should be used parsimoniously.

> <img width="501" alt="Screen Shot 2019-09-20 at 09 12 52" src="https://user-images.githubusercontent.com/4140793/65329797-30f5e500-db87-11e9-93f6-74477e3be00d.png">

How about matching the flutter.dev color (##008f83, which happens to be the same color used on angulardart.dev and, I believe, was the same color that we used to have as part of the previous site design), or one that is just a bit darker in order to have [sufficient contrast](https://webaim.org/resources/contrastchecker) (#006555):

> ![image](https://user-images.githubusercontent.com/4140793/65330706-2ccac700-db89-11e9-8da3-a67fecc05b4f.png)

chalin:
@kwalrath WDYT?

kwalrath:
@mit-mit looking at the designs, the code font was supposed to be teal. I don't remember when/how it turned red, do you?

chalin:
Under `flutter/website` the code color is defined on this line: https://github.com/flutter/website/blob/master/src/_assets/css/_bootstrap_adjust.scss#L102

```scss
code {
  color: #008f83;
}
```

It seems that that definition wasn't migrated over to site-www's `_bootstrap_adjust.scss` when the site redesign happened.

As a consequence, the code color defaults to what is defined at https://github.com/dart-lang/site-www/blob/master/src/_assets/css/_variables.scss#L141, which is the old color setting from over 3 years ago:

```scss
//== Code

$code-color:                  #c7254e;
$code-bg:                     #f9f9f9;
```

If we're going to use the same code color for site-www and flutter/website, then we might as well set it in https://github.com/dart-lang/site-shared/blob/master/src/_assets/css/_code_pre.scss.

chalin:
Related issue: #1942

dhanushuUzumaki:
Can I take this up?

kwalrath:
No, I think the sites will each have their own colors.

dhanushuUzumaki:
@kwalrath So, the new color for code would be #006555?

PR #2000, opened 2019-10-17
Update code color

dhanushuUzumaki:
Contributes to #1929

### DIFF
--- a/src/_assets/css/_variables.scss
+++ b/src/_assets/css/_variables.scss
@@ -138,7 +138,7 @@ $card-border-width: 0;
 
 //== Code
 
-$code-color:                  #c7254e;
+$code-color:                  #247776;
 $code-bg:                     #f9f9f9;
 
 $kbd-color:                   #fff;
--- a/src/_assets/css/main.scss
+++ b/src/_assets/css/main.scss
@@ -27,6 +27,7 @@
 $font-color: $body-color;
 $font-weight-light-bold: ($font-weight-normal + $font-weight-bold)/2;
 $code-weight: 400;
+$code-color: #247776; // Temporary fix
 
 /* -----------------------------------------
   Global
@@ -211,7 +212,7 @@ pre, pre.prettyprint {
 }
 code {
   background-color: transparent;
-  color: #bc0056;
+  color: $code-color;
   font-size: 100%;
   padding: 0;
 }


chalin:
Hi @dhanushuUzumaki. Thanks for the fix!

It is customary to put the issue number in the PR comment rather than in the title, for example, something like this: Closes #1929. In this way GitHub will automatically close the corresponding issue when the PR is merged.

chalin:
Have you generated the site locally to ensure that the color update works?

dhanushuUzumaki:
> Hi @dhanushuUzumaki. Thanks for the fix!
> 
> It is customary to put the issue number in the PR comment rather than in the title, for example, something like this: Closes #1929. In this way GitHub will automatically close the corresponding issue when the PR is merged.

Sure @chalin Will use this format from next time.

dhanushuUzumaki:
> But the code font is still red; for example the "bin" in "Homebrew bin directory...".

Sorry, My bad. Have trouble with colors. But found the issue.

https://github.com/dart-lang/site-www/blob/e61fa47052b910aed9eadcf86795f7444a8613de/src/_assets/css/main.scss#L214 needs to be changed. Changing this seems to do the trick. However, if I use $color-code variable, it gets the value of #4A4A4A due to this line https://github.com/dart-lang/site-shared/blob/8f783bfa10019afb023d4917f9cd8e614227bd3d/src/_assets/css/_bootstrap_config.scss#L7

Should I use the hex value instead of the variable?

dhanushuUzumaki:
Using $color-code variable

image

Using hex value directly

image

chalin:
There is something wrong with the SCSS overrides that I'll fix in a followup PR.

In the meantime, I suggest that you:
- Use the `$code-color` variable in the definition of `code` in `main.scss`.
- Still in `main.scss`, right after the declaration of `$code-weight: 400;`, repeat the declaration of `$code-color` (duplicate it from `_variables`), and add a comment to say that this is a temporary fix: `// Temporary fix`.

Regenerate the site (delete all site-generated files), and ensure that the color is updated as it should be. Thanks.

dhanushuUzumaki:
@chalin Made the changes and I can see the new color.

image

chalin:
Great, thanks!

PR #2004, opened 2019-10-17
Markdown cleanup: no need for spaces in Jekyll directives

chalin:
Broad search-and-replace. No changes to the generated site.


chalin:
(This is prep for a link-fix PR which will follow shortly.)

Issue #1247, opened 2018-11-29
CI: check external links via cron build


chalin:
- [x] Setup cron job.
- [ ] Adapt scripts so that external link checking is done only in a cron job.

chalin:
It might be possible to fix #1107 at the same time.

kwalrath:
We should add a test that looks for bad external links, but (unlike flutter.dev’s current behavior) I don’t think we should fail the build just because an external URL no longer works.

Related: #1685, https://github.com/flutter/website/issues/2734

chalin:
Since this issue has been opened, linkcheck has been updated, and no reports command failures on connection failures. That makes external checking much more robust.

Note that most of the machinery is in place for this, and link checking (which is currently only for internal links), is allowed to fail w/o causing Travis to be alarmed. So, if we don't mind doing external link check for every PR and master build, then all we need to do is use the `--external` flag of `check-links`.

Otherwise, a little bit of config could ensure that we only do external link checking for, say, daily cron job builds.

WDYT?

chalin:
For the Flutter website it was suggested that we only check external links every X days, see https://github.com/flutter/website/issues/3079. I assume that we'll want something similar here too?

chalin:
Cron job scheduled to run daily:
> ![image](https://user-images.githubusercontent.com/4140793/67124547-bdb0b480-f1c0-11e9-9e94-9ee9c74de0e8.png)

First cron job run successfully (and none of the jobs trigger a site deploy, per #2011 -- also related is https://github.com/flutter/website/issues/3127):
> ![image](https://user-images.githubusercontent.com/4140793/67124927-9c9c9380-f1c1-11e9-8472-dc551f0f1831.png)

cc @johnpryan @legalcodes

antfitch:
Closing this one. It's quite old and I don't see movement here since 2019. If this is still an issue, go ahead and open a new ticket with any uncompleted requests.

PR #2017, opened 2019-10-18
CI: deploy site only once per build

### LINKED ISSUE #2011: CI: site is deployed twice by two separate jobs in same build
Changes introduced by #1719 cause the site to be deployed twice, now that all jobs are passing. For example, consider build [#3839](https://travis-ci.org/dart-lang/site-www/builds/599545208); the site gets deployed for these two jobs:
- [3839.1](https://travis-ci.org/dart-lang/site-www/jobs/599545209):
  > Deploying application ...
  > Deploying to Firebase project: default
- [3839.6](https://travis-ci.org/dart-lang/site-www/jobs/599545214):
  > Deploying application ...
  > Deploying to Firebase project: default

cc @johnpryan @legalcodes @kwalrath

chalin:
Closes #2011
Followup to #1894

cc @legalcodes @kwalrath


chalin:
Travis is green, but we won't know if this actually works until it is merged into master. 🤷‍♂

.travis.yml
```diff
@@ -62,7 +41,7 @@ deploy:
     on:
       repo: dart-lang/site-www
       branch: master
-      condition: $TASK == *build*
+      condition: $TASK == *build* && $TRAVIS_DART_VERSION == stable
```
johnpryan: We might consider making the wildcard more exact. It's possible that more scripts with the world "build" might be introduced. Maybe *tool/build.sh* would be less prone to breakage?

.travis.yml
```diff
@@ -13,28 +14,15 @@ env:
...
-    - env: TASK="./tool/analyze-and-test-examples.sh"
-      dart: dev
-    - env: TASK="./tool/check-code.sh"
-      dart: dev
-    - env: TASK="./tool/build-and-check-links.sh"
-      dart: stable
+# matrix:
+#   allow_failures:
+#     - env: TASK="./tool/analyze-and-test-examples.sh"
+#       dart: dev
```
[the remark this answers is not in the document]
chalin: Done. (I usually like to keep an example of the allow_failures section because I tend to forget the syntax so uncommenting is quicker than having to do a look up :).

.travis.yml
chalin: Good idea. Done

.travis.yml
```diff
@@ -62,7 +36,7 @@ deploy:
     on:
       repo: dart-lang/site-www
       branch: master
-      condition: $TASK == *build*
+      condition: $TASK == *tool/build.sh* && $TRAVIS_DART_VERSION == stable && $TRAVIS_EVENT_TYPE == push
```
chalin: I've started looking into CRON jobs builds for this site (for #1247). I've added an extra condition to ensure that cron jobs don't trigger a site deploy.

chalin:
Changes done. Good to go @johnpryan?

PR #2023, opened 2019-10-22
Add $HOME/.pub-cache/bin to suggested PATH

domesticmouse:


kwalrath:
We should adjust the description to match the extended path, and also provide Windows & Mac equivalents, no?

domesticmouse:
I thought about this patch. Apart from the need to find a PC so I can try and remember how to appropriately patch the search path of a Windows user, I also realised I needed to introduce what `pub global` was all about to motivate adding that additional path component.

And about there I decided it was just fine the way it is. =)

PR #2037, opened 2019-10-24
Style: fix positioning of static analysis code block labels

chalin:
Closes #2035

Staged, e.g., see:
- https://dart-dev-staging-0.firebaseapp.com/guides/language/analysis-options#enabling-additional-type-checks
- https://dart-dev-staging-0.firebaseapp.com/guides/language/language-tour#final-and-const

PR #2038, opened 2019-10-25
Reclaim 'good' and 'bad' style names

chalin:
No net change in the site appearance.

This is cleanup following #2037, which dropped old and unused style names "good" and "bad".
This completes the TODO item in `main.scss`: reclaim and use the class names "good" and "bad", rather than "good-style" and "bad-style". (The "-style" suffix was only added while we were transitioning away from the old "good" and "bad".)


chalin:
I also checked that the generated site didn't have any lingering uses of `good-style` or `bad-style`.

PR #2048, opened 2019-10-28
dart2native & dartaotruntime docs

### LINKED ISSUE #1948: dart2aot -> dart2native
The `dart2aot` command will be replaced by [`dart2native`](https://github.com/dart-lang/sdk/tree/master/pkg/dart2native).

We'll need to do the following:

* Write a new page for dart2native.
* Remove https://dart.dev/tools/dart2aot and redirect it to dart2native.
* Be sure to cover the aot & exe formats, and when you'd want each one.

/cc @mit-mit @cskau-g

kwalrath:
In 2.6, dart2aot will be superseded by dart2native. This PR adds pages for dart2native & dartaotruntime, and it redirects dart2aot to dart2native. It also cleans up the /tools page a bit.

Fixes #1948

Staged:
https://sz-www-1.firebaseapp.com/tools
https://sz-www-1.firebaseapp.com/tools/dart2aot
https://sz-www-1.firebaseapp.com/tools/dart2native
https://sz-www-1.firebaseapp.com/tools/dartaotruntime


src/tools/dart2native.md
```diff
@@ -1,7 +1,105 @@
...
+$ my_app
+```
+
+
+## Creating AOT snapshots
+
+To create an AOT snapshot, add `-o aot` to the command:
+
+```terminal
+$ dart2native bin/main.dart -o aot
```
domesticmouse: I'm confused about this command. Above we use the same flag to nominate the output file, yet here we are using it to define the type of compilation. I'd understand if the file extension defines the compilation style, e.g.

```bash
$ dart2native bin/main.dart -o bin/main.aot
```

kwalrath: Oops, I overcorrected. It should be `-k`. Let me fix this.

kwalrath: OK, fixed. ptal.

(Also, I filed https://github.com/dart-lang/sdk/issues/39154 and fixed https://github.com/dart-lang/site-www/issues/1948#issuecomment-535382470.)

src/tools/dart2native.md
```diff
@@ -1,7 +1,105 @@
...
-toc: false
 ---
 
-Coming soon.
+Use the `dart2native` command to AOT (ahead-of-time) compile
+a Dart program to native x64 machine code.
+The output of `dart2native` is either
+a **standalone executable** (the default)
+or an **AOT snapshot** that you can run with the [`dartaotruntime` command][].
+The `dart2native` command is supported on Windows, macOS, and Linux.
```
domesticmouse: Is it worth explaining what situations that an AOT snapshot would be useful over a standalone executable? Currently, with what little I know, it feels like AOT snapshot is the equivalent of a `.o` file. Is the AOT snapshot version independent?

src/tools/dart2native.md
domesticmouse: lgtm, thanks =)

mit-mit:
Looks like there are a few more pages mentioning `dart2aot`, e.g. https://dart.dev/tools/sdk. Did you run a search?

src/tools/dartaotruntime.md
```diff
@@ -0,0 +1,23 @@
...
+Hello from Dart.
+```
+
+For information on command-line options, use the --help flag:
+
+```terminal
+$ dartaotruntime --help
+```
+
+For more information, see the
```
mit-mit: I would move this up right after the example, and make it a bit more explicit, e.g. `To compile Dart programs into AUT snapshots compatible with this tool, see the ...`

src/tools/dart2native.md
mit-mit: We could, for example something like "AOT snapshots can be helpful when you are limited on disk space, as they allow for running multiple AOT snapshots without each incurring the full AOT runtime disk overhead."

kwalrath:
> Looks like there are a few more pages mentioning `dart2aot`, e.g. https://dart.dev/tools/sdk. Did you run a search?

Oops. No. Fixed.

src/tools/dart2native.md
kwalrath: I had this down below, but it's too easy to miss. I've removed that section and put all that content in the intro.

kwalrath:
ptal

src/_tutorials/server/get-started.md
```diff
@@ -134,25 +134,18 @@ More information:
...
-Dart code to optimized native machine code.
+during development. Now that your small app is done,
+it's time to AOT compile your Dart code to optimized native machine code.
 
-Use the `dart2aot` tool to AOT compile the program to machine code:
+Use the `dart2native` tool to AOT compile the program to machine code:
 
 ```terminal
-$ dart2aot bin/main.dart bin/main.dart.aot
+$ dart2native bin/main.dart -o bin/my_app
```
mit-mit: I find it odd to use the `-o` argument to do nothing but remove the extension. To make this a bit more meaningful, can we change the output location, e.g. `-o out/my_app` ?

src/tools/dart2native.md
```diff
@@ -1,7 +1,101 @@
...
+```terminal
+$ dart2native bin/main.dart -o bin/my_app
+```
+
+You can distribute and run that executable like you would
+any other executable file:
+
+```terminal
+$ cp bin/my_app ~/bin
+$ my_app
```
mit-mit: I believe you need `./my_app` (current dir is not in path on anything but windows)

kwalrath: Ah, but I copied it to ~/bin, which is in MY path. :)

But you're probably right that this is too subtle.

src/_tutorials/server/get-started.md
kwalrath: It removes the extension AND changes the name, which would otherwise be bin/main.aot, right? That said, I don't mind changing the output directory if that's a good practice. (Is it, @domesticmouse?)

domesticmouse: Leaving the generated executable in the bin directory along side the source.dart file makes sense to me. It’s the technique I’m using in my Docker images.

src/tools/dart2native.md
kwalrath: I've changed it to be /usr/local/bin instead of ~/bin. OK, @mit-mit?

domesticmouse: Most Linux users don’t have write permissions on /usr/local/bin. Please revert.

kwalrath: OK... So @mit-mit are you OK with the previous text? It works. (In my setup, at least.)

mit-mit: Yes, `/usr/local/bin` is not a good choice; Windows users are going to have no understanding of that that is.

I suggest you simply:

```terminal
cp bin/my_app .
./my_app
```

kwalrath:
@mit-mit ptal

mit-mit:
LGTM, merging as 2.6 is out

Issue #1899, opened 2019-09-13
Help people debug and improve performance of command-line apps


kwalrath:
We have dart.dev/tools/dart-devtools, but we need to add some resources (videos, docs, ...) to help people debug their apps and improve their apps' performance.

See #2055 for more discussion.

kwalrath:
Let me explain the weird effort labels...It'll take minutes to do the basic getting started instructions, but doing more (providing a walkthrough, documenting how to use DevTools, etc.) is likely to take hours or days.

devoncarew:
The pages which will work for CLI apps are the debugger, memory usage, the profiler page, and logging.

parlough:
Closing in favor of https://github.com/dart-lang/site-www/issues/4352 which discusses documenting various DevTools views.

PR #2055, opened 2019-10-31
Add Dart DevTools docs

mit-mit:
Fixes https://github.com/dart-lang/site-www/issues/1899

### DIFF
--- a/src/tools/dart-devtools.md
+++ b/src/tools/dart-devtools.md
@@ -1,16 +1,73 @@
 ---
 title: Dart DevTools
-description: A suite of performance tools for Dart and Flutter.
+description: A suite of debugging and performance tools.
 toc: false
 ---
 
-Dart DevTools is a suite of performance tools for Dart and Flutter.
-In addition to Flutter-specific features,
-DevTools supports source-level debugging of Dart apps,
-as well as viewing general log and diagnostics information about a running app.
-No matter whether you use an IDE or the command line,
-DevTools works with your existing development workflow.
+Dart DevTools is a suite of debugging and performance tools for Dart and
+Flutter. In addition to [Flutter-specific features][Flutter DevTools documentation.],
+DevTools supports source-level debugging of Dart apps, as well as viewing
+general log and diagnostics information about a running app.
 
-For more information, see the [Dart DevTools documentation.][]
+## Using DevTools with a command-line app
 
-[Dart DevTools documentation.]: https://flutter.github.io/devtools/
+### Install DevTools
+
+If you have `pub` on your path, you can run:
+
+```console
+$ pub global activate devtools
+```
+
+That command installs (or updates) DevTools on your machine.
+
+### Launch the DevTools application server
+
+Next, run the local web server, which serves the DevTools application itself. To
+do that, run the following command. You should see output that looks something
+like this:
+
+```console
+$ pub global run devtools
+
+Serving DevTools at http://127.0.0.1.9100
+```
+
+### Start an application to debug
+
+Next, start an app to connect to. Once the app starts, you'll see a message in
+your terminal that looks like:
+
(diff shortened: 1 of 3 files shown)


src/tools/dart-devtools.md
```diff
@@ -1,16 +1,73 @@
...
 
-For more information, see the [Dart DevTools documentation.][]
+## Using DevTools with a command-line app
 
-[Dart DevTools documentation.]: https://flutter.github.io/devtools/
+### Install DevTools
+
+If you have `pub` on your path, you can run:
+
+```console
```
chalin:
```terminal
    ```

Tends to give better results than `console` in Jekyll .md files (as opposed to .md files rendered by GitHub, like READMEs).

Comment (mit-mit) [reply src/tools/dart-devtools.md]: Ah, good tip, that does look slightly better (it formats the `$` prefix with a different color than the succeeding command)

**Comment by kwalrath** [comment]
It looks like the devtools-connect.png figure is missing.

**Comment by mit-mit** [comment]
Sorry, added missing image

**Comment by kwalrath** [comment]
Once you get the DevTools running, how do you use them with a command-line app? Do we have docs anywhere? A video? Are the DevTools useful for one-and-done command-line apps, or are they only useful for server-type apps?

**Comment by sfshaza2** [comment]
You can use with a command line app. When you run an app from the terminal, it says what port is running DevTools. Open that URL in a browser.

I do like your idea of expanding the server tutorial to use these!

**Comment by kwalrath** [comment]
> You can use with a command line app. When you run an app from the terminal, it says what port is running DevTools. Open that URL in a browser.

I got it running. I just don't know what to do after that.

**Comment by sfshaza2** [comment]
Oooh! Yeah, we *really* need more videos on this.

**Comment by kwalrath** [comment]
I've edited dart-devtools.md (and have some other, smaller changes I'd like to make), but I'd really like to be able to point to more info. We get people started, but then what do they do?

**Comment by kwalrath** [comment]
I've staged my changes at https://kw-staging-dartlang-2.firebaseapp.com/tools/dart-devtools.

@mit-mit are they OK by you?

File: src/tools/index.md [hunk 2]
```diff
@@ -130,6 +130,10 @@ Here are the general-purpose tools you might want to install:
 [dart_style]: {{site.pub-pkg}}/dart_style
 [dartfix]: {{site.pub-pkg}}/dartfix
 
+### Debugging and diagnostics
```
kwalrath: Should this be under **Tools for developing command-line apps and servers**?

src/tools/index.md
```diff
@@ -130,6 +130,10 @@ Here are the general-purpose tools you might want to install:
...
+
+[Dart DevTools](/tools/dart-devtools)
+: A suite of suite of debugging and performance tools.
```
kwalrath: suite of ~~suite of~~

kwalrath: We should probably say that they're useful for both Flutter and command-line apps.

mit-mit:
> Are the DevTools useful for one-and-done command-line apps,

Yes, I personally used them last week for setting some breakpoints and stepping through a command-line app I was trying to fix.

src/tools/index.md
mit-mit: Cal you elaborate? We already say "for Dart and Flutter", and have subsections below for Flutter and console...

src/tools/index.md
mit-mit: I think debugging is one of the main gaps this addresses for console apps.

src/tools/index.md
mit-mit: I see, you added those subsections in your edits. That looks good; resolving this comment.

mit-mit:
@kwalrath your edits LGTM!

PR #2057, opened 2019-10-31
Always recommend using constraints in pubspec.yaml

mklim:
Recommend using explicit dependency versions even for apps.

I have seen issues in the ecosystem based on the latest versions of multiple packages conflicting with each other, or the latest version of a package not quite being compatible with an arbitrary version of Flutter. I think even for apps it's better to recommend using versions in `pubspec.yaml` even if a `pubspec.lock` committed to source would make this redundant in the general case.

### DIFF
--- a/src/tools/pub/dependencies.md
+++ b/src/tools/pub/dependencies.md
@@ -235,20 +235,32 @@ Instead, the typical workflow is:
 
 ## Version constraints
 
-If your package is an application, you don't usually need to specify [version
-constraints](/tools/pub/glossary#version-constraint) for your dependencies. You
-typically want to use the latest versions of the dependencies when you first
-create your app. Then you'll create and check in a
-[lockfile](/tools/pub/glossary#lockfile) that pins your dependencies to those specific
-versions. Specifying version constraints in your pubspec then is usually
-redundant (though you can do it if you want).
+It's always important to have some kind of [version
+constraints](/tools/pub/glossary#version-constraint) set for your
+dependencies. Your own project can immediately and randomly break if any of
+its dependencies are updated with a breaking change.
 
-For a [library package](/tools/pub/glossary#library-package) that you want users to
-reuse, though, it is important to specify version constraints. That lets people
-using your package know which versions of its dependencies they can rely on to
-be compatible with your library. Your goal is to allow a range of versions as
-wide as possible to give your users flexibility. But it should be narrow enough
-to exclude versions that you know don't work or haven't been tested.
+Specifying version constraints in your pubspec.yaml always gives you the most
+control. You technically don't need to do this if your package is an
+application. You can instead create and check in a
+[lockfile](/tools/pub/glossary#lockfile) that pins your dependencies to those
+specific versions. Specifying version constraints in your pubspec after that
+is generally redundant. However the only way to upgrade your dependencies
+using just a lockfile is with [pub upgrade](/tools/pub/cmd/pub-upgrade),
+which will limit you to the latest version and by default will try to upgrade
+all your packages at once. This can be a problem if the latest version of the
+package isn't compatible with some other latest version in your dependency
+graph. If you want more control over your versions you should still specify
+them in the pubspec.yaml so that you can pick and choose specific versions if
+you need to.
+
+For a [library package](/tools/pub/glossary#library-package) that you want
+users to reuse, though, it is necessary to always specify version
+constraints. That lets people using your package know which versions of its
+dependencies they can rely on to be compatible with your library. Your goal
+is to allow a range of versions as wide as possible to give your users
+flexibility. But it should be narrow enough to exclude versions that you know
+don't work or haven't been tested.
 
 The Dart community uses [semantic versioning][]<sup id="fnref:semver"><a
 href="#fn:semver">1</a></sup>, which helps you know which versions should work.


mklim:
/cc @sfshaza2. Not sure who else should be added to review this, feel free to add them. :)

@blasten and I were discussing this offline, we've both seen a lot of issues that appear to be coming from people never locking their dependencies. To some extent that implies people are ignoring the part about this that says to use `pubspec.lock`, but I also think it would be helpful to change the official recommendation here to use `pubspec.yaml` all of the time.

kwalrath:
@munificent would you be the right person to review this recommendation? I can do a final editing pass once we're sure we have the right content.

src/tools/pub/dependencies.md
```diff
@@ -235,20 +235,32 @@ Instead, the typical workflow is:
...
-For a [library package](/tools/pub/glossary#library-package) that you want users to
-reuse, though, it is important to specify version constraints. That lets people
-using your package know which versions of its dependencies they can rely on to
-be compatible with your library. Your goal is to allow a range of versions as
-wide as possible to give your users flexibility. But it should be narrow enough
-to exclude versions that you know don't work or haven't been tested.
+It's always important to have some kind of [version
+constraints](/tools/pub/glossary#version-constraint) set for your
+dependencies. Your own project can immediately and randomly break if any of
+its dependencies are updated with a breaking change.
```
munificent: No, there's nothing "immediate" or "random" about it. It only happens when you run "pub upgrade". If you have ever successfully run "pub get" *once* then you have a lock file and none of your dependencies will change under you until you choose to do something that affects that lockfile.

src/tools/pub/dependencies.md
```diff
@@ -235,20 +235,32 @@ Instead, the typical workflow is:
...
+
+Specifying version constraints in your pubspec.yaml always gives you the most
+control. You technically don't need to do this if your package is an
+application. You can instead create and check in a
+[lockfile](/tools/pub/glossary#lockfile) that pins your dependencies to those
+specific versions. Specifying version constraints in your pubspec after that
+is generally redundant. However the only way to upgrade your dependencies
+using just a lockfile is with [pub upgrade](/tools/pub/cmd/pub-upgrade),
```
munificent: "pub upgrade" is the primary way to update your dependencies regardless of whether or not you specify version constraints. Yes, if you change a package's constraint to exclude the currently locked version, then "pub get" is *forced* to implicitly upgrade because the existing lockfile is no longer valid. But that's sort of an edge case and not the normal way to upgrade things. If you want to upgrade... you run "pub upgrade". That's why it's called "upgrade".

src/tools/pub/dependencies.md
```diff
@@ -235,20 +235,32 @@ Instead, the typical workflow is:
...
+which will limit you to the latest version and by default will try to upgrade
```
munificent: > limit you to the latest version

No, it doesn't. It gets the most recent version that satisfies all of the transitive constraints on it.

If you specifically *don't* want the latest version for some reason, yes, you can then pin it in your pubspec by adding a constraint. But for the most part users do want to upgrade their app the latest versions of dependencies when they choose to upgrade.

src/tools/pub/dependencies.md
```diff
@@ -235,20 +235,32 @@ Instead, the typical workflow is:
...
+all your packages at once. This can be a problem if the latest version of the
```
munificent: > all your packages at once

If you only want to unlock a single package, you can just run "pub upgrade [foo]".

src/tools/pub/dependencies.md
```diff
@@ -235,20 +235,32 @@ Instead, the typical workflow is:
...
+package isn't compatible with some other latest version in your dependency
+graph. If you want more control over your versions you should still specify
```
munificent: That's not a problem. Pub will pick the latest version that *is* compatible. That's its main reason for being.

src/tools/pub/dependencies.md
```diff
@@ -235,20 +235,32 @@ Instead, the typical workflow is:
...
+them in the pubspec.yaml so that you can pick and choose specific versions if
+you need to.
+
+For a [library package](/tools/pub/glossary#library-package) that you want
+users to reuse, though, it is necessary to always specify version
+constraints. That lets people using your package know which versions of its
+dependencies they can rely on to be compatible with your library. Your goal
```
munificent: This sentence is pretty confusing.

But, basically, the reason a library package puts in constraints is because it doesn't know which context it will be used in. It needs a constraint so that it can ensure it gets compatible dependencies when used in the context of many different package graphs.

An application is only ever used in the context of a single package graph — its own. Therefore, it has no real need to tell the system what *range* of versions it's compatible with. At any point in time it will only ever use exactly one specific version of each dependency. The easiest way to pin a dependency to a single version is to let the lockfile do it for you.

munificent:
> The problem for Flutter is that pub doesn't actually meaningfully understand "transitive constraints" because Flutter packages can have *platform level* transitive constraints that pub is completely unaware of.

That is a Very Hard Problem and not one any package manager or simple human can likely solve gracefully. Even if we take pub out of the loop and make it the user's problem, how do we expect *them* to solve it? How do they determine that two or more packages that they may never have heard of deep in their transitive dependency graph are fighting over some non-pub shared dependency that they also have never heard of and that, somehow, they are supposed to figure out which version of one of them will work?

> I think the main point of contention is that you're coming at it from the understanding that pub will automatically handle version solving in all cases

I'm not claiming it handles it well in all cases, but it does seem to do so in most. This PR is recommending changing the *default* user behavior, and that only makes sense to me if the current behavior is wrong more often than it's right. Otherwise, it is a good default.

Most packages users rely on do not have external platform dependencies. Most of those packages can be easily upgraded and pub will handle pinning them gracefully using the lockfile. Telling the user they should write down the *same* version number in their pubspec is redundant busywork. That number is already perfectly well encoded in the lockfile. That's *why pub has lockfiles.*

I think it's reasonable for users to pin a package in their application's pubspec if they need to. The tool supports that for a reason. But I don't see much evidence that that should be the *default* behavior. For most packages the current policy is both easy and seems to work well.

If we do need to change the policy to handle platform dependencies better, that's OK, but we should explain *that.* Right now, this PR just says "It's always important to have some kind of version constraints set for your dependencies. Your own project can immediately and randomly break if any of its dependencies are updated with a breaking change." which is both wrong and confusing. The fundamental architectural choice of pub is that your project *cannot* immediately and randomly break.

It might be worth adding a paragraph saying something like: "Some pub packages for Flutter plug-ins have dependencies on external resources like Maven packages that pub cannot see or control. In those cases, you may want to pin your dependency on that package to a specific version if the versions it selects lead to conflicts between multiple pub packages that all use that resource."

jonasfj:
Coming from the node.js eco-system I'll admit I was surprised that Dart didn't recommend writing version constraints in the `pubspec.yaml`. But I can see how it might make sense for small projects, and it makes it easy to get started.

But personally, I would use `pubspec.yaml` to express:
 * What my code depends on, and,
 * What version constraints my code requires, typically as a `^x.y.z` constraint.

In this school of though, it is always wrong to:
 * import something not specified as a dependency in `pubspec.yaml`,
 * Not specifying a version constraint on a dependency in `pubspec.yaml` (as I cannot possibly know if my code is compatible with the next breaking version of a given package).

This ensures that `pub upgrade` should work without requiring any changes to my code. **If** my dependencies honor semantic versioning. In practice, however, we all know that semantic versioning is hard, and whether changing some behavior is a breaking change can be subject to interpretation. External dependencies in Flutter is just another source of semantic versioning violations that can be subject to interpretation.

Hence, we rely on `pubspec.lock` to freeze the entire dependency tree. Notably, this mechanism ensures consistent behavior across workspaces, CI, and production. Most importantly it ensures we don't ship code we didn't test.

-----
I suspect some of this is just personal preference.

munificent:
> Coming from the node.js eco-system I'll admit I was surprised that Dart didn't recommend writing version constraints in the `pubspec.yaml`. But I can see how it might make sense for small projects, and it makes it easy to get started.

npm is very architecturally different from pub because npm generally doesn't do shared dependencies. That in turn means that version resolution is much simpler because there aren't many constraints to solve. You just give each package what it wants and each gets it own thing. You may end up with fifteen different versions of package foo floating around in your app but 🤷‍♂. Since there no few shared constraints, you can make your dependencies as precise as you want. So the convention is to pin stuff in the package manifest and there's no real need for a lockfile.

That design doesn't work in a nominally, statically-typed language like Dart where a Foo object from one version of foo is not the same as a Foo from a different version. So pub does shared dependencies, which in turn means constraint solving. That in turn means you want wide constraints to ensure solutions can be found. But that sacrifices predictability. You get that back by adding in lockfiles.

Pub's design is based on Bundler. In Bundler, specifying a version on a dependency [is optional](https://bundler.io/man/gemfile.5.html#GEMS) and users are [encouraged to check in their lock file](https://bundler.io/v2.0/guides/using_bundler_in_applications.html) to have deterministic version solving.

In Cargo, which was created by the author of Bundler, you do specify version ranges in your manifest but you [still rely on the lock file to precisely pin them](https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html). I believe they pick this compromise position because Cargo does a mixture of shared and unshared dependencies. It will share dependencies within the same compatible semver range, but allow multiple different major versions of the same dependency, I think.

Pub doesn't do the last (in large part because the language itself gives pub no way to make that actually work) part. Dependencies are fully shared in pub: "foo" is only ever a single version of the package in the context of an application.

mit-mit:
I'm also *not* in favor of this change. I can provide a longer response later when I have a bit more time.

jonasfj:
> npm is very architecturally different from pub because npm generally doesn't do shared dependencies.

Afaik, npm does [share dependencies](https://docs.npmjs.com/cli/dedupe.html) when possible.
But yes, it's a different eco-system.

In any event, whether you specify a version range in `pubspec.yaml` or not, is also a personal preference. We nothing stops users from doing this, if they don't want `pub upgrade` to give breaking version upgrades.

In either case, users who develop applications should always check in `pubspec.lock` and rely on `pubspec.lock` for locking all dependencies. This should be our key recommendation.

----
Regarding this issue, I think current recommendations are fine. We do allow specification of version constraints in `pubspec.yaml` and hint at this with the sentence `(though you can do it if you want)`.

I suppose we could extend that to say something like `(though you can do it if you want, or if the latest version of a package is causing problems)`. But this seems a little obvious to me.

kwalrath:
Do we want to close this PR or upload a smaller set of changes?

munificent:
I think the right answer is to close it. I appreciate you starting this discussion @mklim, but I don't think we want to make a large change to the recommended practices.

Issue #2072, opened 2019-11-06
Change Windows installer link to be https


jonasfj:
The [community supported Windows installer](http://www.gekorm.com/dart-windows) referenced in:

https://github.com/dart-lang/site-www/blob/a4b74d94bf0e94fafbaeb4af8e55f69ff4abf39b/src/tools/sdk/_windows.md#L50-L55

Only has an HTTP-only website (ie. without HTTPS support).

https://github.com/dart-lang/site-www/blob/a4b74d94bf0e94fafbaeb4af8e55f69ff4abf39b/src/tools/sdk/_windows.md#L61

@GeKorm, is this still maintained? If so could we consider upgrading the website to HTTPS, or moving it to some static web host with HTTPS support.

Or maybe we should consider not advertising this installation method.

jonasfj:
I just saw the github repository [github.com/GeKorm/dart-windows](https://github.com/GeKorm/dart-windows), which has been updated fairly recent, so I assume this is still maintained.

@GeKorm, maybe it would be safer to host the website for the installer on github pages which has HTTPS?

GeKorm:
Hi, it is indeed hosted on github pages. I'll look into enabling HTTPS, thanks

jonasfj:
If it's on GitHub pages it should just be an option to enforce HTTPS:
https://github.blog/2018-05-01-github-pages-custom-domains-https/

Thanks for looking into this :)

GeKorm:
Yep those are the changes. Github used a different set of IPs for A records before that blog post

GeKorm:
Should be good now, with http -> https redirection

kwalrath:
Thanks, @GeKorm!

PR #2073, opened 2019-11-06
Use HTTPS in links when possible

jonasfj:
Just a lot of nits upgrading HTTP to HTTPS links whenever possible.

I found one or two links that didn't work or which seemed broken, so I left those alone.

### DIFF
--- a/src/_articles/archive/numeric-computation.md
+++ b/src/_articles/archive/numeric-computation.md
@@ -105,7 +105,7 @@ graduates from smi to mint to bigint:
 ### Floating-point numbers
 
 The VM has a single representation for floating-point numbers, the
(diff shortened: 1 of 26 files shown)


kwalrath:
The problem is that examples/misc/lib/effective_dart/docs_good.dart needs to be updated to use https instead of http. I can't do that through the GitHub interface (afaik) since you didn't edit that file.

If you can fix that file before EOD, great. If not, I'll merge this PR and then fix the build.

kwalrath:
It looks like @jonasfj won't get to this, so I'll merge this and then commit a PR to fix the build and #2072.

jonasfj:
Sorry, I was a bit behind on my inbox.. thanks fixing it.

kwalrath:
No worries, it was easy.

PR #2079, opened 2019-11-08
Document conditional imports

### LINKED ISSUE #1569: Conditional imports is not covered in the language tour
Page URL: https://www.dartlang.org/guides/language/language-tour
Page source: [src/_guides/language/language-tour.md](https://github.com/dart-lang/site-www/tree/master/src/_guides/language/language-tour.md).

Conditional imports is not covered by the language tour. Conditional imports are usually done like:

```dart
import 'src/impl_stub.dart'  // stub implementation
    if (dart.library.io) 'src/impl_io.dart' // implementation using dart:io
    if (dart.library.html) 'src/impl_web.dart'; // implementation using dart:html|js
```

It would be nice to cover this in the language tour as the specification is hard to read. A few good examples of what you can make conditions on would also be useful, as this is not obvious.

@lrn, please correct me if this is not a stable language feature?

kwalrath:
I put most of the content in the "Creating library packages" page, not directly in the language tour.

Staged:
https://kw-staging-dartlang-2.firebaseapp.com/guides/language/language-tour#implementing-libraries
https://kw-staging-dartlang-2.firebaseapp.com/guides/libraries/create-library-packages#conditionally-importing-and-exporting-library-files

I'd love some feedback on the most common use cases (e.g. which platforms you'd support, and if there's something else I should show/cover).

Fixes #1569


mit-mit:
cc @lrhn can you give this a review, please?

examples/create_libraries/lib/src/hw.dart
```diff
@@ -0,0 +1 @@
+String get message => 'Hello World!';
```
jonasfj: This is the stub implementation, typically it's called `hw_stub.dart` and often it might be useful to throw an `Error`.
I would suggest [UnimplementedError](https://api.dartlang.org/stable/2.6.0/dart-core/UnimplementedError-class.html) because this is not implemented, and this file is not intended to be imported into any actual production code.

I suspect @lrhn would suggest throwing [UnsupportedError](https://api.dartlang.org/stable/2.6.0/dart-core/UnsupportedError-class.html)? :)

But throwing an `Error` from the `_stub.dart` implementation is probably a good pattern to give users. Sometimes a stub implementation can implement the full feature -- but often it's a bug if the stub implementation is imported.

src/_guides/libraries/create-library-packages.md
```diff
@@ -153,6 +154,63 @@ it could instead use the `package:` directive
...
+    if (dart.library.html) 'src/hw_html.dart'; // JS|dart:html implementation
+```
+
+Here's how the conditional export code works:
+
+* If this code is running in the Dart VM (JIT or AOT),
+  then export `hw_io.dart`.
+* If this code is running in a browser,
+  then export `hw_html.dart`.
+* Otherwise, export `hw.dart`. **[PENDING: when would this happen?]**
```
jonasfj: I think the analyzer will import this during development, so that might be where code completion comes from (check if this is true).

Also if one day a platform that supports neither `dart:html` or `dart:io` then I guess it would be imported there. Though no such platform currently exist.

examples/create_libraries/test/hw_mp_browser_test.dart
```diff
@@ -0,0 +1,14 @@
+// To run this test:
+// cd create_libraries
+// pub run test -p chrome
+@Tags(['browser'])
```
jonasfj: Why tag it `'browser'`? asking because I haven't done that in my code, does it mean anything?

src/_guides/libraries/create-library-packages.md
```diff
@@ -153,6 +154,63 @@ it could instead use the `package:` directive
...
+## Conditionally importing and exporting library files
+
+If your library supports multiple platforms,
+then you might need to conditionally import or export library files.
+A common use case is a library that supports both web and native platforms.
+
+To conditionally import or export,
+you must check for the presence of `dart:*` libraries.
+Here's an example of conditional export code that
+checks for the presence of `dart:io` and `dart:html`:
```
jonasfj: nit: I would perhaps consider testing on `dart:js` instead of `dart:html` because I would imagine `dart:js` is more comon in Flutter for web packages (just a guess).

examples/create_libraries/test/hw_mp_browser_test.dart
kwalrath: It's required on site-www to get browser tests detected and run.

src/_guides/libraries/create-library-packages.md
kwalrath: I've seen dart:html in the wild, but that's probably pre-flutter web, and I'd love to get a definitive answer on this. @johnpryan?

examples/create_libraries/lib/src/hw.dart
lrhn: I'd say `UnsupportedError`, yes. If you plan to later implement, but you haven't gotten to it yet, you use `UnimplementedError`. It's really only something you should use during development. The `UnsupportedError` is for permanent lack of support, like writing to a read-only list.

Otherwise I agree. I would probably use `hw_none.dart` as name, but `stub` is fine too.

src/_guides/libraries/create-library-packages.md
lrhn: It depends on what platform library you actually depend on.
This is *library* testing, not *platform* testing&mdash;which is why the example is actually not very good. It's not detecting "VM" or "JavaScript", it's detecting the presence of the `dart:io` library or the `dart:html` library.

I know of at least one experiment with embedding the VM and a browser engine in the same program, and that platform hae *both* `dart:io` and `dart:html` (not sure if it had `dart:js` though, since you used Dart instead).

So, if you depend on `dart:js` then test for `dart.library.js`, and if you depend on `dart:html` then test for `dart.library.html`. That makes your code maximally supported. 

In most cases, the feature you need is a DOM feature, not a plain JS feature, so I'd test for `dart.library.html`.

examples/create_libraries/lib/src/hw_html.dart
```diff
@@ -0,0 +1 @@
+String get message => 'Hello World from JavaScript!';
```
lrhn: I'd add something here which is actually `dart:html` specific.
For example:
```dart
import "dart:html";

void alarm([String text]) {
  window.alert(text ?? message);
}

String get message => 'Hello World from JavaScript!';
```

examples/create_libraries/lib/src/hw_io.dart
```diff
@@ -0,0 +1 @@
+String get message => 'Hello World from the VM!';
```
lrhn: And:
```dart
import "dart:io";

void alarm([String text]) {
  stderr.writeln(text ?? message);
}

String get message => 'Hello World from the VM!';
```

examples/create_libraries/example/hw_example.dart
```diff
@@ -0,0 +1,5 @@
+import 'package:create_libraries/hw_mp.dart';
+
+main() {
+  print('greeting: ${message}');
```
lrhn: Or (see below):
```dart
main(List<String> arguments) {
  if (arguments.isEmpty) {
     alarm();
  } else {
     alarm(arguments.first);  
  }
}
```

src/_guides/libraries/create-library-packages.md
kwalrath: What happens if both dart:io & dart:html are supported? Would only the first `if` take effect, or might you import multiple libraries? (I'd hope it's the former, not the latter.)

examples/create_libraries/example/hw_example.dart
kwalrath: I ended up not using this code, to keep the focus on the import.

kwalrath:
ptal

kwalrath:
@lrhn still lgty?

kwalrath:
I'm taking that as a _yes_. :)

examples/create_libraries/example/hw_example.dart
```diff
@@ -0,0 +1,5 @@
+import 'package:create_libraries/hw_mp.dart';
+
+main() {
```
chalin: @kwalrath: `main()` -> `void main()` (to avoid an analyzer issue)?

chalin: FYI, I added `void` in a recent PR, matching how we declared the `alarm()` function. If we've decided that to promote the declaration of `main()` w/o it's type, I'll revert the change.

Issue #2029, opened 2019-10-23
Deprecating the `authors` field in `pubspec.yaml`


jonasfj:
Maybe wait until https://github.com/dart-lang/pub/issues/2234 is done, as the `authors` field is currently required when publishing a package.

In the future the `authors` field will not be displayed on `pub.dev`. The `pub publish` command will warn that it has been deprecated, but beyond that there is no harm in including the field.

However, once https://github.com/dart-lang/pub/issues/2234 has landed in `pub` and released in a stable Dart/Flutter release, we should be sure to update the documentation site.

jonasfj:
/CC @sigurdm, who is working to make this land in Dart 2.7.

kwalrath:
This is apparently visible in at least some releases, per #2084, so I should update the docs.

sigurdm:
Yes - I think that would be fine.

pouloghost:
We have deployed a pub server inside our company, and author field is crucial for looking for the developer of the package/plugin.
All packages are published using one CI machine, hence we can't use uploader.
And I think this is more or less the standard routine in big companies, so
What is the purpose of deprecating author? What's the alternative?

sigurdm:
> What is the purpose of deprecating author? What's the alternative?

The problem with the authors field is that there is no validation of it.
You could claim you are A else misusing the end-users trust in A.

Therefore we have stopped displaying it on pub.dev.
We find it misleading to support a tag that we are never going to use.

Though it is deprecated you can still use it internally on your own server - I don't think we can or will ever forbid you from doing that.

Alternatively you can invent your own alternative field for this information. The [pubspec docs](https://dart.dev/tools/pub/pubspec) explains that you can add custom fields. For example: `<company_name>_authors` or some such.

I don't know how you are using the author information, but for CI error reporting you should probably look up the person making the commit that triggered the CI run instead. For Travis you can maybe use
https://stackoverflow.com/questions/39681509/travis-ci-how-to-get-committer-email-author-name-within-after-script-command

kwalrath:
Fixed in #2085.

PR #2085, opened 2019-11-13
Updates related to verified publishers, author/authors

kwalrath:
Staged:
https://sz-www-1.firebaseapp.com/tools/pub/publishing
https://sz-www-1.firebaseapp.com/tools/pub/glossary#uploader
https://sz-www-1.firebaseapp.com/tools/pub/cmd/pub-uploader
https://sz-www-1.firebaseapp.com/tools/pub/pubspec

Contributes to #2029


src/tools/pub/glossary.md
```diff
@@ -146,16 +146,34 @@ and C are transitive ones.
 
 ## Uploader
```
mit-mit: Maybe indent this subsection and the verified publisher subsection below under a new overall section?

```
## Managing who has rights to publish

### Uploader

### Verified publisher
```

src/tools/pub/glossary.md
```diff
@@ -146,16 +146,34 @@ and C are transitive ones.
...
+then all members of the publisher can upload the package.
 
 Anyone uploading a new package automatically becomes an uploader for
-that package. Otherwise, to become an uploader, you need to contact an
+that package. Otherwise, to become an uploader of a package that
+doesn't have a verified publisher, you need to contact an
 existing uploader and ask them to add you as another uploader.
 
+
+## Verified publisher
```
mit-mit: Here, rather than repeating all the details of what a verified publisher is, consider focusing on what members of a publisher have the rights to do (like in the previous Uploaders section). That is:

  - Publish new packages under the publisher
  - Publish new versions of existing packages
  - Edit the publisher metadata

src/tools/pub/pubspec.md
```diff
@@ -172,22 +167,22 @@ The description is plain text: no markdown or HTML.
...
+(`Natalie Weizenbaum <nweiz@google.com>`).
+However, these values weren't verified,
+and pub.dev no longer displays them.
 
-{% prettify yaml %}
-authors:
-- Natalie Weizenbaum <nweiz@google.com>
-- Bob Nystrom <rnystrom@google.com>
-{% endprettify %}
+As of Dart 2.7, the `pub publish` command
```
mit-mit: maybe add here: ", and the pub.dev site no longer displays the `author(s)` on a package page.".

src/tools/pub/glossary.md
kwalrath: This is a glossary, not a task oriented doc, so alpha order is the rule.

src/tools/pub/glossary.md
kwalrath: I just stole this definition from pub.dev, since it seemed worthwhile to have the definition on this site too. I didn't want to get into functionality here, since it's easy to miss updating the glossary. Instead I just linked to the verified publishers page. (We don't have any similar page for uploaders, so I guess it makes sense to have the info here.)

(I just checked, and people actually do come to the glossary, yay.)

src/tools/pub/pubspec.md
kwalrath: I say that up above.

kwalrath:
Thanks for the review, @mit-mit! After my explanations, do you still think I need to make changes?

src/tools/pub/glossary.md
mit-mit: Even if this is the glossary, why the asymmetry between this section (which talks about publishers from a consumers PoV) and the Uploader section (which talks about uploaders from a publisher PoV)?

src/tools/pub/pubspec.md
mit-mit: Sorry, missed that. I think it would be easier to not miss if we did something like this:

```
You might see an `author` or `authors` section in old pubspecs.
These optional fields were a way to describe
the author(s) of your package and to provide contact information.
Each author could be either a single name
(`Natalie Weizenbaum`) or a name and an email address
(`Natalie Weizenbaum <nweiz@google.com>`).

As of Dart 2.7, pub.dev no longer displays them and the `pub publish` command
displays a warning if your pubspec has an `author` or `authors` section.
```

mit-mit:
See comments

src/tools/pub/pubspec.md
kwalrath: The thing is, pub.dev no longer displaying the fields is independent of 2.7. But I'll try to reword it to remove that dependency.

src/tools/pub/glossary.md
kwalrath: Funny, I don't see the audiences of the entries as being that different; it's only the degree of completeness that's different, because there's no page dedicated to the uploader concept. The only link to glossary#uploader is from /tools/pub/cmd/pub-uploader, which is intended for publishers; there are no links to glossary#verified-publisher because there's a verified publisher page.

I'm going to cut down the verified publisher entry, since it's only here for completeness.

kwalrath: I've made the glossary changes (and some others while I'm in there).

If we want to say anything more about publishers, we should say it in one of the other docs, since it's unlikely people will come here to find out about verified publishers.

We might also want to revamp the verified publishers page to be clearer about which sections are for which audiences.

src/tools/pub/glossary.md
```diff
@@ -146,16 +146,34 @@ and C are transitive ones.
...
-same as the [author](/tools/pub/pubspec#authorauthors) of a package.
+Someone who has administrative permissions for a package.
+A package uploader can upload new versions of the package,
+and they can also [add and remove other uploaders](/tools/pub/cmd/pub-uploader)
+for that package. 
+
+If a package has a verified publisher,
+then all members of the publisher can upload the package.
 
 Anyone uploading a new package automatically becomes an uploader for
```
mit-mit: I would consider removing this paragraph. It repeats the last part of the first paragraph in this section.

PR #2092, opened 2019-11-14
Server > Get Started: use tested sources for DartPad Hello World

chalin:
Contributes to #1803 by replacing a Gist-based DartPad with a DartPad initialized from an analyzed and tested code excerpt from our examples folder.

**Notes:**
- This PR uses the **_experimental_** DartPad [inject_embed feature](https://github.com/dart-lang/dart-pad/wiki/Embedding-Guide#converting-code-blocks-to-dartpad).
- The DartPad is initialized with the base Hello World excerpt that was already a part of our examples, rather than create a new Hello World variant that illustrates use of `const`.

Followup to #1921
Related to #1919

### Staged
https://dart-dev-staging-1.firebaseapp.com/tutorials/server/get-started

### Screen shot

> Screen Shot 2019-11-14 at 12 12 35

cc @johnpryan @legalcodes @RedBrogdon

PR #2094, opened 2019-11-14
Server > Get Started: fix front matter

chalin:
Followup to #2092.

It seems that the Jekyll YAML parser under MacOS is different from (more lenient than) the Linux one. Under Linux, a YAML syntax error was reported, but not under MacOS.

cc @kwalrath @johnpryan @legalcodes


chalin:
Btw, I wasn't able to reproduce the problem locally, under MacOS.

chalin:
Travis is green. Merging since this fixes a page that is currently broken.

Issue #1256, opened 2018-12-07
Improve/update pub pages


kwalrath:
I found some non-urgent issues when doing a survey of all the pages under https://www.dartlang.org/tools/pub/cmd. (I was comparing them to the output of `pub help <cmd>`, for pub version 2.1.0). Here are my notes, in case anyone has a yen to improve these pages:

* `pub help` and `pub version` don't have pages. (I don't think this is a problem, since they're kind of obvious.)
* I filed dart-lang/pub#2001 about small issues with the pub command's references to doc pages.
* No pages talk about the `no-` modifier to `--` options. (I don't think this is a problem, since it's in the help.)
* Options section contents and formatting aren't consistent.
* `pub deps` page is missing `--dev` and `--executables`.
* `pub downgrade` page is missing `-n, --dry-run`. (Why doesn't `pub downgrade` have `--precompile`?)
* `pub get` and `pub upgrade` pages are missing `--precompile` and `-n, --dry-run`.

There was a more important issue in `pub run`, which I'm fixing in #1255.

kwalrath:
Also see #1342, which mentions that all the pub pages could probably use a look-through.

sil3nthill:
hello i would like to work for this and make a PR , what should i exactly do ?

kwalrath:
@sil3nthill thank you for the offer! We're in the middle of a bunch of site changes, so we're avoiding content changes until after I/O (early May).

Would you be willing to look at this in May?

gryclmn:
I would like to help with this issue as well for a first time PR. How can I help?

Purukitto:
Is this issue fixed to satisfaction or can I contribute to it?

Purukitto:
@kwalrath Is this issue fixed to satisfaction or can I contribute to it?

kwalrath:
Sorry, @sil3nthill, @gryclmn, and @Purukitto for not getting back to you sooner!

We're ready to accept changes on these pages now. What I'd love is for each of you to claim one page (here in a comment) and commit a PR for that page fixing whatever issues I noted above for the page. I'm open to any other fixes you deem necessary, as well.

Once we have a single-page PR from you, then we can figure out how to organize the rest of the work.

Does that sound OK?

Purukitto:
@kwalrath 
> Does that sound OK?

Sure does
<br><br>
> `pub help` and `pub version` don't have pages. (I don't think this is a problem, since they're kind of obvious.)

[...]

PR #2099, opened 2019-11-15
Update pub-deps.md

jtxmp:
Update description for pub deps options.

Contributes to #1256

### DIFF
--- a/src/tools/pub/cmd/pub-deps.md
+++ b/src/tools/pub/cmd/pub-deps.md
@@ -7,7 +7,7 @@ toc: false
 _Deps_ is one of the commands of the [pub tool](/tools/pub/cmd).
 
 {% prettify none %}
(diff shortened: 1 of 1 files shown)


jtxmp:
@googlebot I signed it!

chalin:
> I'm not sure how to treat default options (e.g. `--dev`) and negatable options (`dev`/`--no-dev`) ...

I'd suggest that we use the same style the args library promotes, e.g., for the `pub deps` command we get `--[no-]dev`:

```console
> pub deps -h
Print package dependencies.

Usage: pub deps
-h, --help           Print this usage information.
-s, --style          How output should be displayed.
                     [compact, tree (default), list]

    --[no-]dev       Whether to include dev dependencies.
                     (defaults to on)

    --executables    List all available executables.

Run "pub help" to see global options.
```

PR #2101, opened 2019-11-16
Add directions for getting the args package

### LINKED ISSUE #432: Not clear how to add dependencies to a command-line dart app from the command-line tutorial
https://dart.dev/tutorials/server/cmdline

It's not clear from this tutorial that you need a pubspec or some other definition to resolve the dependency on package:args.

kwalrath:
Fixes #432

Staged: https://kw-staging-dartlang-2.firebaseapp.com/tutorials/server/cmdline#getting-dependencies

I tried to make this as lightweight as possible, to keep the focus on the code rather than on how to get and modify the template. But that means it breaks some conventions (as I admit in the text).

### DIFF
--- a/src/_tutorials/server/cmdline.md
+++ b/src/_tutorials/server/cmdline.md
@@ -91,9 +91,55 @@ brief description of key app features, click the highlighted code below.
 
 {% include_relative _dcat-example.html %}
 
-You can run the app from the command line over `quote.txt`
-([downloadable file]({{site.repo.this}}/blob/master/examples/misc/test_data/quote.txt?raw=1){:target="_blank"})
-or any other text file:
+
+### Getting dependencies
+
+You might notice that dcat depends on a package named **args**.
+To get the args package, use the
+[pub package manager](/guides/packages).
+
+A real app has tests, license files, and so on, in a file hierarchy
+such as the one in the [Dart command-line app template.][]
+But for this first app, let's do the minimum necessary to get the code to run:
+
+[Dart command-line app template.]: https://github.com/dart-lang/stagehand/tree/master/templates/console-full
+
+1. Create a directory named `dcat`, and change to that directory.
+2. Inside `dcat`, create a file named `dcat.dart`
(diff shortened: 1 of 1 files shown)


src/_tutorials/server/cmdline.md
```diff
@@ -91,9 +91,55 @@ brief description of key app features, click the highlighted code below.
...
+   $ pub get
+   Resolving dependencies... 
+   + args 1.5.2
+   Changed 1 dependency!
+   ```
+
+{{site.alert.note}}
+  To learn more about using packages and organizing your code, see the
+  [package documentation](/guides/packages) and
+  [layout conventions](/tools/pub/package-layout).
```
johnpryan: nit: Could there also be a link to ["the pub tool"](https://dart.dev/tools/pub/cmd)?

src/_tutorials/server/cmdline.md
```diff
@@ -91,9 +91,55 @@ brief description of key app features, click the highlighted code below.
...
+   with the following code:
+
+   ```yaml
+   name: dcat
+
+   environment:
+     sdk: '>=2.6.0 <3.0.0'
+
+   dependencies:
+     args: ^1.5.0
```
chalin: Do you want me to setup a code-excerpt instruction for this code block afterwards?

kwalrath: That would be great. I didn't because it wasn't obvious how to do so. :)

src/_tutorials/server/cmdline.md
kwalrath: Do you think that readers will think something's missing if I leave out that link? The package docs introduce the pub tool and link to all the pub command docs, so the pub tool link seems unnecessary and possibly confusing.

kwalrath:
I'm going to merge... we can always update later.

src/_tutorials/server/cmdline.md
johnpryan: ok, that makes sense.

PR #2110, opened 2019-11-17
Update pub-downgrade.md

jtxmp:
Update description for pub downgrade options.

Note: There is no option for `--precompile` in `pub downgrade`: https://github.com/dart-lang/pub/blob/master/lib/src/command/downgrade.dart

Contributes to #1256.

### DIFF
--- a/src/tools/pub/cmd/pub-downgrade.md
+++ b/src/tools/pub/cmd/pub-downgrade.md
@@ -6,7 +6,7 @@ description: Use pub downgrade to get the lowest versions of all dependencies us
 _Downgrade_ is one of the commands of the [pub tool](/tools/pub/cmd).
 
 {% prettify sh %}
-$ pub downgrade [args] [dependencies] [--offline] [--no-offline] [--dry-run]
+$ pub downgrade [--[no]-offline] [-n|--dry-run] [dependencies...] 
 {% endprettify %}
 
 Without any additional arguments, `pub downgrade` gets the lowest versions of
@@ -105,15 +105,12 @@ available.
 For options that apply to all pub commands, see
 [Global options](/tools/pub/cmd#global-options).
 
-`--offline`
-: Use locally cached packages instead of accessing the network.
-
-`--no-offline`
-: Connect to the network to check for hosted dependencies. This is the 
-default setting.
+`--[no]-offline`
+: By default, pub connects to the network to check for hosted 
+dependencies (`--no-offline`). To use cached packages instead, use `--offline`.
 
 `--dry-run` or `-n`
-: Report what dependencies would change but would not change any.
+: Report what dependencies would change but don't change any.
 
 
 <aside class="alert alert-info" markdown="1">
--- a/src/tools/pub/cmd/pub-downgrade.md
+++ b/src/tools/pub/cmd/pub-downgrade.md
@@ -99,13 +99,22 @@ trying to find a set of versions that work with your package from what's already
 available.
 
 
-## Options {#options}
 
-The `pub downgrade` command supports the
-[`pub get` options](/tools/pub/cmd/pub-get#options).
+## Options
 
 For options that apply to all pub commands, see
 [Global options](/tools/pub/cmd#global-options).
+
+`--offline`
+: Uses locally cached packages instead of accessing the network.
+
+`--no-offline`
+: Connects to the network to check for hosted dependencies. This is the 
+default setting.
+
+`--dry-run` or `-n`
+: Reports what dependencies would change but would not change any.
+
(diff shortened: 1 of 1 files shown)


src/tools/pub/cmd/pub-downgrade.md
```diff
@@ -99,14 +99,23 @@ trying to find a set of versions that work with your package from what's already
...
 
 For options that apply to all pub commands, see
 [Global options](/tools/pub/cmd#global-options).
 
+`--offline`
+: Use locally cached packages instead of accessing the network.
+
+`--no-offline`
+: Connect to the network to check for hosted dependencies. This is the 
+default setting.
```
chalin: I think that we can be more terse when it comes to expressing what the default setting is. How about: _Connect to the network to check for hosted dependencies (default)._

This is often what tools do. @kwalrath, thoughts?

src/tools/pub/cmd/pub-downgrade.md
```diff
@@ -6,7 +6,7 @@ description: Use pub downgrade to get the lowest versions of all dependencies us
 _Downgrade_ is one of the commands of the [pub tool](/tools/pub/cmd).
 
 {% prettify sh %}
-$ pub downgrade [args] [dependencies]
+$ pub downgrade [args] [dependencies] [--offline] [--no-offline] [--dry-run]
```
chalin: Hi. Thanks for the PR.

As I mentioned in https://github.com/dart-lang/site-www/pull/2099#issuecomment-554756767, I'd suggest using the same syntax as the command help, namely `--[no-]offline`, etc.

- The `args` were meant to represent the options. If you're explicitly listing the options, they need to appear before the dependencies (and so drop `args`)
- dependencies -> dependencies... 
- Maybe include the shorthand too: `--dry-run` -> `-n|--dry-run`? @kwalrath WDYT?

---

```console
> pub downgrade -h
Downgrade the current package's dependencies to oldest versions.

This doesn't modify the lockfile, so it can be reset with "pub get".

Usage: pub downgrade [dependencies...]
-h, --help            Print this usage information.
    --[no-]offline    Use cached packages instead of accessing the network.
-n, --dry-run         Report what dependencies would change but don't change any.

Run "pub help" to see global options.
See https://dart.dev/tools/pub/cmd/pub-downgrade for detailed documentation.
```

src/tools/pub/cmd/pub-downgrade.md
```diff
@@ -99,14 +99,23 @@ trying to find a set of versions that work with your package from what's already
...
 
+`--offline`
+: Use locally cached packages instead of accessing the network.
+
+`--no-offline`
+: Connect to the network to check for hosted dependencies. This is the 
+default setting.
+
+`--dry-run` or `-n`
+: Report what dependencies would change but would not change any.
```
chalin: You can stick to the command help text: _Report what dependencies would change but don't change any._

src/tools/pub/cmd/pub-downgrade.md
```diff
@@ -99,14 +99,23 @@ trying to find a set of versions that work with your package from what's already
 available.
 
 
-## Options {#options}
```
chalin: Good move dropping the unnecessary `{#options}` syntax.

src/tools/pub/cmd/pub-downgrade.md
jtxmp: Hi @chalin,

Thanks for your guidance! Pending @kwalrath 's input, I will amend the line to read:

`
$ pub downgrade [--[no-]offline] [-n|--dry-run] [dependencies...] 
`

src/tools/pub/cmd/pub-downgrade.md
kwalrath: It might be clearer to have the "default" at the beginning of the explanation, or after the period. Maybe:

```
:  Default. Connect ...
```

But if we combine negatable flags, that won't work. Maybe, in that case:

```
`--[no-]offline`
: By default, pub connects to the network... (`--offline`). To use cached packages instead, use `--no-offline`.
```

kwalrath: (It took me a couple of times to get the sense of the flag right... I think.)

src/tools/pub/cmd/pub-downgrade.md
kwalrath: @jtxmp I'll be happy to review again once you've made the changes @chalin suggested.

kwalrath:
Oops, it looks like I dropped the ball here. I'll try to take a look today.

PR #2111, opened 2019-11-17
Fix note_client.dart bug; add missing server tests; test cleanup

chalin:
- Contributes to #1803
- Fix _examples/httpserver_ `note_client.dar`t: misformatted JSON
- Add test for` examples/httpserver/bin/note_server.dart`
- Add test for `examples/httpserver/bin/static_file_server.dart`
- Use `dart_test.yaml` include feature rather than links.


examples/create_libraries/example/hw_example.dart
```diff
@@ -1,5 +1,5 @@
 import 'package:create_libraries/hw_mp.dart';
 
-main() {
+void main() {
```
chalin: The analyzer was complaining about a missing return type, and we include the `void` return type for the `alarm()` method, so I added it here too.

examples/create_libraries/pubspec.yaml
```diff
@@ -1,16 +1,10 @@
...
 environment:
   sdk: '>=2.6.0 <3.0.0'
 
-#dependencies:
-#  path: ^1.6.0
-
 dev_dependencies:
-  build_runner: any
-  build_test: any
-  build_web_compilers: any
```
chalin: There's nothing to launch in the package so I removed these dependencies.

examples/create_libraries/test/hw_mp_browser_test.dart
```diff
@@ -1,6 +1,3 @@
-// To run this test:
-// cd create_libraries
-// pub run test -p chrome
```
chalin: Instructions have been added to the examples/README.md file.

chalin:
TBR. Travis is green, merging. I'll address comments, if any, in a followup PR.

examples/README.md
```diff
@@ -18,17 +18,28 @@ $ dartanalyzer analysis_options.yaml .
...
 ```
 ./tool/analyze-and-test-examples.sh
 ```
 
-If you get a warning about test failures or analysis errors,
-you might need to update one or more analyzer results files.
-For example, update `examples/misc/analyzer-results.txt`, if
-the change applies to both stable and dev Dart releases. Otherwise,
-update only the release specific file: either
+If you get a warning about test failures or analysis errors, you might need to
```
kwalrath: Just FYI, we've finally codified the convention of using semantic line breaks to avoid unnecessary diffs.

https://github.com/dart-lang/site-shared/wiki/Writing-for-Dart-and-Flutter-websites#semantic-line-breaks

chalin: Got it. Thanks for the links!

src/_tutorials/server/httpserver.md
```diff
@@ -856,22 +856,18 @@ also uses the http_server package.
 This server serves any file from the server's directory
 or subdirectory.
 
-Run `static_file_server.dart`,
-and test it with the URL localhost:4048/file.txt.
-Change `file.txt` to other filenames within the directory.
+Run `static_file_server.dart`, and test it with the URL
+[localhost:4048](http://localhost:4048).
```
kwalrath: @johnpryan Do you think it's OK to have links to localhost? (I think we've been inconsistent, and I don't know what the considerations are.)

chalin: FYI, we have links to localhost on all three sites (Dart, Flutter and NgDart). But of course, I'm open to any other considerations that there might be.

johnpryan: It looks like the link checker is working, so from a tooling perspective this seems fine. AFAICT this works on all platforms, and isn't a security issue.

PR #2114, opened 2019-11-18
Async-await codelab: a code to examples folder; use DartPad inject

chalin:
**Edit**: PR updated so that it keeps the original embed DPs via iframes using Gists. The iframe code is contained within a Liquid if-then-else. Once we're clear to use inject-DP, simply set `useIframe` to false at the top of the file (and/or clean out the iframe elements).

---

Original comment:

This PR switches from Gist-based DartPads to use of the inject DartPad feature for all DartPads except the 3 exercises (since those require two files rather than one).

(The exercise DartPads will be addressed in a followup PR.)

- Contributes to #1803
- Code taken from Gists, applied dartfmt, and added `ignore` analyzer directives to eliminate warnings.

Staged at https://dart-dev-staging-1.firebaseapp.com/codelabs/async-await

### DIFF
--- a/examples/async_await/bin/futures_intro.dart
+++ b/examples/async_await/bin/futures_intro.dart
@@ -1,0 +1,19 @@
+// #docregion ''
+Future<void> fetchUserOrder() {
+  // Imagine that this function is fetching user info from another service or database.
+  return Future.delayed(Duration(seconds: 2), () => print('Large Latte'));
+}
+// #enddocregion ''
+
+// #docregion error
+Future<void> fetchUserOrderError() {
+// Imagine that this function is fetching user info but encounters a bug
+  return Future.delayed(Duration(seconds: 2),
+      () => throw Exception('Logout failed: user ID is invalid'));
+}
+// #docregion ''
+
+void main() {
+  fetchUserOrder();
+  print('Fetching user order...');
+}
--- a/src/codelabs/async-await.md
+++ b/src/codelabs/async-await.md
@@ -1,9 +1,26 @@
 ---
 title: "Asynchronous programming: futures, async, await"
 description: Learn about and practice writing asynchronous code in DartPad!
+js: [{url: 'https://dartpad.dev/experimental/inject_embed.dart.js', defer: true}]
 ---
+{% assign useIframe = true -%}
+<?code-excerpt replace="/ *\/\/\s+ignore_for_file:[^\n]+\n//g; /(^|\n) *\/\/\s+ignore:[^\n]+\n/$1/g; /(\n[^\n]+) *\/\/\s+ignore:[^\n]+\n/$1\n/g"?>
+<?code-excerpt plaster="none"?>
+<style>
+{% comment %}
+TODO(chalin): move this into one of our SCSS files
+{% endcomment -%}
+iframe[src^="https://dartpad"] {
+  border: 1px solid #ccc;
+  margin-bottom: 1rem;
+  min-height: 150px;
+  resize: vertical;
+  width: 100%;
+}
+</style>
+
 This codelab teaches you how to write asynchronous code using
-futures and the `async` and `await` keywords. Using the embedded DartPad
+futures and the `async` and `await` keywords. Using embedded DartPad
 editors, you can test your knowledge by running example code and completing
 exercises.
 
@@ -44,6 +61,7 @@ The following example shows the wrong way to use an asynchronous function
 Before running this example, try to spot the issue -- what do you think the
 output will be?
 
+{% if useIframe -%}
 <iframe
   src="{{site.dartpad-embed}}?id=5c8c7716b6b4284842f15fe079f61e47"
   style="border: 1px solid lightgrey; margin-top: 10px; margin-bottom: 25px"
@@ -51,6 +69,29 @@ output will be?
   height="420"
   width="100%" >
 </iframe>
+{% else -%}
+
+<?code-excerpt "async_await/bin/get_order_sync_bad.dart" remove="Fetching"?>
+```dart:run-dartpad:height-380px
+// This example shows how *not* to write asynchronous Dart code.
+
+String createOrderMessage() {
+  var order = fetchUserOrder();
+  return 'Your order is: $order';
+}
+
+Future<String> fetchUserOrder() =>
+    // Imagine that this function is more complex and slow.
+    Future.delayed(
+      Duration(seconds: 2),
+      () => 'Large Latte',
+    );
+
+void main() {
+  print(createOrderMessage());
+}
+```
+{% endif -%}
 
 Here's why the example fails to print the value that `fetchUserOrder()` eventually
 produces:
@@ -126,6 +167,7 @@ printing to the console. Because it doesn't return a usable value,
 `fetchUserOrder()` has the type `Future<void>`. Before you run the example,
 try to predict which will print first: "Large Latte" or "Fetching user order...".
 
+{% if useIframe -%}
 <iframe
   src="{{site.dartpad-embed}}?id=57e6085344cbd1719ed42b32f8ad1bce"
   style="border: 1px solid lightgrey; margin-top: 10px; margin-bottom: 25px"
@@ -133,6 +175,21 @@ try to predict which will print first: "Large Latte" or "Fetching user order..."
   height="300"
   width="100%" >
 </iframe>
+{% else -%}
+
+<?code-excerpt "async_await/bin/futures_intro.dart"?>
+```dart:run-dartpad:height-300px
+Future<void> fetchUserOrder() {
+  // Imagine that this function is fetching user info from another service or database.
+  return Future.delayed(Duration(seconds: 2), () => print('Large Latte'));
+}
+
+void main() {
+  fetchUserOrder();
+  print('Fetching user order...');
+}
+```
+{% endif -%}
 
 In the preceding example, even though `fetchUserOrder()` executes before
 the `print()` call on line 8, the console shows the output from line 8
@@ -144,6 +201,7 @@ This is because `fetchUserOrder()` delays before it prints "Large Latte".
 Run the following example to see how a future completes with an error.
 A bit later you'll learn how to handle the error.
 
+{% if useIframe -%}
 <iframe
   src="{{site.dartpad-embed}}?id=d843061bbd9388b837c57613dc6d5125"
   style="border: 1px solid lightgrey; margin-top: 10px; margin-bottom: 25px"
@@ -151,6 +209,22 @@ A bit later you'll learn how to handle the error.
   height="275"
   width="100%" >
 </iframe>
+{% else -%}
+
+<?code-excerpt "async_await/bin/futures_intro.dart (error)" replace="/Error//g"?>
+```dart:run-dartpad:height-300px
+Future<void> fetchUserOrder() {
+// Imagine that this function is fetching user info but encounters a bug
+  return Future.delayed(Duration(seconds: 2),
+      () => throw Exception('Logout failed: user ID is invalid'));
+}
+
+void main() {
+  fetchUserOrder();
+  print('Fetching user order...');
+}
+```
+{% endif -%}
 
 In this example, `fetchUserOrder()` completes with an error indicating that the
 user ID is invalid.
@@ -191,8 +265,9 @@ function.
 
 First, add the `async` keyword before the function body:
 
+<?code-excerpt "async_await/bin/get_order_sync_bad.dart (main-sig)" replace="/main\(\)/$& async/g; /async/[!$&!]/g; /$/ ··· }/g"?>
 {% prettify dart %}
-void main() [!async!] {
+void main() [!async!] { ··· }
 {% endprettify %}
 
 If the function has a declared return type, then update the type to be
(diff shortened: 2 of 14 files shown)


examples/async_await/bin/futures_intro.dart
```diff
@@ -0,0 +1,27 @@
+// #docregion ''
+Future<void> getUserOrder() =>
+    // Imagine that this function fetches user info from a data source
+    Future.delayed(
```
kwalrath: I was surprised to see => used for > one line of code.
(I looked it up in Effective Dart, and I don't see anything saying
you should stick to one line for =>... but still, it surprised me.)
[Old code](https://dart.dev/codelabs/async-await#example-incorrectly-using-an-asynchronous-function)
[New code](https://dart-dev-staging-1.firebaseapp.com/codelabs/async-await#example-incorrectly-using-an-asynchronous-function)

/cc @munificent

src/codelabs/async-await.md
```diff
@@ -44,13 +59,26 @@ The following example shows the wrong way to use an asynchronous function
...
 output will be?
 
-<iframe
-  src="{{site.dartpad-embed}}?id=5c8c7716b6b4284842f15fe079f61e47"
-  style="border: 1px solid lightgrey; margin-top: 10px; margin-bottom: 25px"
-  frameborder="no"
-  height="420"
-  width="100%" >
-</iframe>
+<?code-excerpt "async_await/bin/get_order_sync_bad.dart" remove="Fetching"?>
```
kwalrath: Switching to a non-gist source loses the ability to open this example in its own window... but I'm OK with that.

In fact, although I like the ability to copy the code, I find it really annoying that the icon for that covers the code.

Before:  image    After: image

/cc @RedBrogdon @johnpryan

examples/async_await/bin/futures_intro.dart
```diff
@@ -0,0 +1,27 @@
+// #docregion ''
+Future<void> getUserOrder() =>
+    // Imagine that this function fetches user info from a data source
```
kwalrath: We should use the same comments as the gist, no?
(Except that comments should end in periods!)

src/codelabs/async-await.md
```diff
@@ -144,13 +179,20 @@ This is because `getUserOrder()` delays before it prints "Large Latte".
...
-  src="{{site.dartpad-embed}}?id=d843061bbd9388b837c57613dc6d5125"
-  style="border: 1px solid lightgrey; margin-top: 10px; margin-bottom: 25px"
-  frameborder="no"
-  height="275"
-  width="100%" >
-</iframe>
+<?code-excerpt "async_await/bin/futures_intro.dart (error)" replace="/Error//g"?>
+```dart:run-dartpad:height-300px
+Future<void> getUserOrder() =>
+    // Imagine that this function fetches user info from a data source
```
kwalrath: Comment should say something about the error (like the gist does).

src/codelabs/async-await.md
```diff
@@ -144,13 +179,20 @@ This is because `getUserOrder()` delays before it prints "Large Latte".
...
+    Future.delayed(
+      Duration(seconds: 2),
```
kwalrath: @legalcodes are you OK with the durations changing?

src/codelabs/async-await.md
```diff
@@ -44,13 +59,26 @@ The following example shows the wrong way to use an asynchronous function
...
+}
+
+Future<String> getUserOrder() =>
+    // Imagine that this function is more complex and slow
+    Future.delayed(
+      Duration(seconds: 2),
+      () => 'Large Latte',
+    );
+
+void main() {
```
kwalrath: Using `void main()` here means that there's one more thing that changes in the asynchronous example later. (We intentionally left the return type off of `main()` to simplify the example... although we definitely considered other approaches.)

src/codelabs/async-await.md
```diff
@@ -231,58 +275,62 @@ your window is wide enough — is to the right of the synchronous example.
...
+    Future.delayed(
+      Duration(seconds: 2),
+      () => 'Large Latte',
+    );
 
-// Asynchronous
-main() [!async!] {
-print('Fetching user order...');
-print([!await!] createOrderMessage());
+[!Future<void>!]  main() [!async!] {
```
kwalrath: This change needs to be reflected in the "three ways" list that follows... if we keep this change.

src/codelabs/async-await.md
```diff
@@ -321,16 +369,39 @@ synchronous code before the first `await` keyword executes immediately.
...
-<iframe
-  src="{{site.dartpad-embed}}?id=d7abfdea1ae5596e96c7c0203d975dba"
-  style="border: 1px solid lightgrey; margin-top: 10px; margin-bottom: 40px"
-  frameborder="no"
-  height="600"
-  width="100%">
-</iframe>
+<?code-excerpt "async_await/bin/get_order.dart (with-counter)"?>
+```dart:run-dartpad:height-530px
+Future<String> createOrderMessage() async {
```
kwalrath: This code is different from the previous code. @legalcodes are you OK with that?
[Old code](https://dart.dev/codelabs/async-await#example-execution-within-async-functions)
[New code](https://dart-dev-staging-1.firebaseapp.com/codelabs/async-await#example-execution-within-async-functions)

legalcodes:
I'll try to review this PR tomorrow, thanks for the heads up!

src/codelabs/async-await.md
johnpryan: This is good feedback - I've filed an issue: https://github.com/dart-lang/dart-pad/issues/1381

src/codelabs/async-await.md
legalcodes: Let's keep the function signatures as they were -- @galeyang recommended these based on user feedback (i.e. these were carefully chosen for pedagogical reasons). The rest of the changes look fine (reformatting, removing unneeded import).

src/codelabs/async-await.md
```diff
@@ -1,9 +1,24 @@
 ---
 title: "Asynchronous programming: futures, async, await"
 description: Learn about and practice writing asynchronous code in DartPad!
+js: [{url: 'https://dartpad.dev/experimental/inject_embed.dart.js', defer: true}]
 ---
+<?code-excerpt plaster="none"?>
+<style>
+{% comment %}
+TODO(chalin): move this into one of our SCSS files
```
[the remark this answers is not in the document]
legalcodes: See https://github.com/flutter/website/issues/3270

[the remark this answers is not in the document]
kwalrath: And https://github.com/dart-lang/site-www/issues/2119. I don't necessarily want to lose the TODOs from the source code, but I do want them tracked as issues.

examples/async_await/bin/futures_intro.dart
munificent: There are no hard and fast rules for how long the body of an `=>` member can be. I think once it gets to be more than a few lines, then a regular braced body is easier to read (and has less indentation), but it's mostly a judgement call.

kwalrath: For me, the `=>` made this sample harder to read. It's easy to miss the `=>` and wonder where the `}` went.

galeyang:
Thanks for raising this PR! I'd like to bring up the concern of current analytics. As far as what I'm aware of, UXR team currently won't be able to track the usage of individual embed. Currently for Gist-based DartPads, we have gist id as unique identifier which is not available in injected code. The issue is tracked at https://github.com/dart-lang/dart-pad/issues/1352.

Can we make this change/PR after https://github.com/dart-lang/dart-pad/issues/1352 is fixed so that we won't lose track of individual embed?

cc: @johnpryan @RedBrogdon 
Also cc @jayoung-lee as she has ongoing research on this specific codelab. This change may be related her current progress.

jayoung-lee:
To add to Gale's comment, I am using Gist IDs to identify which action (e.g., run, test-failure, etc.) happened on which embed (e.g., the 3rd example in async-await codelab). 

If we are going to replace Gist-based DP with DP inject, can we assign unique IDs (like Gist IDs) to each injected DP and log that into Analytics? I'd suggest using a [custom dimension](https://support.google.com/analytics/answer/2709828?hl=en).

johnpryan:
Adding an analytics ID is on the DartPad roadmap, but isn't supported yet: https://github.com/dart-lang/dart-pad/issues/1352

jayoung-lee:
If the DartPads are replaced without Analytics update, we'll loose the current granularity of analysis until some IDs are added. 

What about reviewing the data we've collected so far and see if it is enough first? If the data we've collect is enough to give feedback to the codelab, we can just use that. If not, we might want to collect more data, before replacing the gist-based DartPads.

src/codelabs/async-await.md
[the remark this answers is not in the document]
chalin: Makes sense. Most often I create issues, but this is an interim PR and I was planning on moving the style definition before I'm done (although that will be in a followup PR). Do you still want me to create an issue for this?

src/codelabs/async-await.md
chalin: Oops, I missed that the comment was different. Fixed.

src/codelabs/async-await.md
chalin: To fill you in: I brought it down to 2 secs where the wait was unnecessary, but kept it 4 sec for the example that relies on the 4 sec counter.

src/codelabs/async-await.md
chalin: Here's part of the old code that is relevant to this discussion:

```
}

}

main() async {
  countSeconds(4);
  await createOrderMessage();
}
```

    >
    > ```dart
    > Future<void> main() async {
    > ```

    So, ...

Thoughts?

examples/async_await/bin/futures_intro.dart
chalin: I've reverted to using `{...}`.

examples/async_await/bin/futures_intro.dart
chalin: Reverted the text and added a period.

src/codelabs/async-await.md
chalin: Reverted; i.e., removed `void`.

src/codelabs/async-await.md
chalin: Reverted.

src/codelabs/async-await.md
chalin: I'm reverting to the original code (modulo dartfmt changes, since the original code was manually formatted, and a change in duration for the delay). I'll move the comment above to an issue if you think that it is worth pursuing.

chalin: Btw, when `createOrderMessage()` is declared as void, we get the following issue reported by the analyzer:

```nocode
```

chalin: I've added an `ignore` directive for now.
The code has been reverted.

chalin: I created #2128 to track this.

galeyang: Thanks for moving it to a new issue. We have some relevant unresolved discussion earlier. I will take a look and continue the discussion here. Thanks!

chalin:
PR now uses the DartPad inject feature for all DartPads except the 3 exercises (since those require two files rather than one). @legalcodes: to you guys already have support for injecting multiple files?

The exercise DartPads can be handled in a followup PR.

chalin:
Blocked on https://github.com/dart-lang/dart-pad/issues/1352

legalcodes:
Re: https://github.com/dart-lang/site-www/pull/2114#issuecomment-557226845 @chalin Do you mean support for injecting `main.dart`, `test.dart`, and `solution.dart` as in https://github.com/dart-lang/dart-pad/blob/master/web/experimental/inject_demo.html#L116-L126 ?

chalin:
> Do you mean support for injecting main.dart, test.dart ...

Yes thanks!

chalin:
FYI, just created https://github.com/dart-lang/dart-pad/issues/1388, which might help in this case.

chalin:
Just to be clear, this PR now uses the same code as the Gists, modulo running dartfmt.

src/codelabs/async-await.md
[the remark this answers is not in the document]
chalin: Just a heads up that I won't have the time to move these iframe styles into one of the SCSS files. So I guess that this TODO, and the one in src/_tutorials/server/get-started.md will be addressed via #2119.

chalin:
@kwalrath @legalcodes: I'd really like to get this PR wrapped up, and hopefully merged so that I can move on to other projects. In support of that, I've updated this PR so that it keeps the original embed DPs via iframes using Gists. The iframe code is contained within a Liquid if-then-else. Once we're clear to use inject-DP, simply set `useIframe` to false at the top of the file (and/or clean out the iframe elements).

PTAL and let me know if there is anything else I can do to help bring this closer to being merged.

kwalrath:
@legalcodes @johnpryan do you approve?

chalin:
Btw, I've updated the staging server: https://dart-dev-staging-1.firebaseapp.com/codelabs/async-await

legalcodes:
I might be able to review this later this afternoon, if that's alright! Otherwise @johnpryan feel free to weigh in.

chalin:
Thanks John and Jon!

PR #2115, opened 2019-11-18
Add video element to page

johnpryan:
Deployed: https://jr-dartlang-1.firebaseapp.com/

### DIFF
--- a/tool/dartpad_picker/compile_dev.sh
+++ b/tool/dartpad_picker/compile_dev.sh
@@ -1,0 +1,10 @@
+#!/usr/bin/env bash
+
+: ${TMP:=tmp}
+
+set -x
+
+pub get
+pub run build_runner build --output $TMP --define "build_web_compilers|entrypoint=compiler=dart2js"
+cp $TMP/web/* ../../src/assets/dash/js
+rm -rf $TMP
--- a/tool/dartpad_picker/lib/dartpad_picker.dart
+++ b/tool/dartpad_picker/lib/dartpad_picker.dart
@@ -51,7 +51,12 @@ class DartPadPicker {
       ..src = iFrameSrc(theme: 'dark', mode: 'dart');
     iFrameHost.children.add(_iFrameElement);
     window.addEventListener('message', (dynamic e) {
-      if (e.data['type'] == 'ready') {
+      // Avoid handling
+      if (e.data != null &&
+          e.data is Map &&
+          e.data.containsKey('type') &&
+          e.data['type'] is String &&
+          e.data['type'] == 'ready') {
         _sendSourceCode();
       }
     });
--- a/tool/dartpad_picker/lib/dartpad_picker.dart
+++ b/tool/dartpad_picker/lib/dartpad_picker.dart
@@ -51,7 +51,7 @@ class DartPadPicker {
       ..src = iFrameSrc(theme: 'dark', mode: 'dart');
     iFrameHost.children.add(_iFrameElement);
     window.addEventListener('message', (dynamic e) {
-      // Avoid handling
+      // Don't handle events from other iframe elements
       if (e.data != null &&
           e.data is Map &&
           e.data.containsKey('type') &&
--- a/src/_assets/css/_dash.scss
+++ b/src/_assets/css/_dash.scss
@@ -35,6 +35,27 @@ $dash-dartpad-border: #293542;
   text-align: center;
   padding: 64px 0 64px 0 !important;
 
+  .callout-button {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    margin-top: 18px;
+    > button {
+      display: flex;
+      flex-direction: row;
+      text-decoration: none;
+      align-items: center;
+      color: $site-color-card-link;
+      &:hover {
+        text-decoration: none;
+        color: darken($site-color-card-link, 20%);
+      }
+      > i {
+        margin-right: 8px;
+      }
+    }
+  }
+
   .callout-title {
     font-size: 28px;
     color: white;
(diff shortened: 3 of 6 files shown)


johnpryan:
This also fixes the console warnings similar to https://github.com/flutter/website/issues/3213

johnpryan:
What's the new error? I'm only seeing cookie-related warnings

kwalrath:
> What's the new error? I'm only seeing cookie-related warnings

Maybe it was a cache issue? I just opened in an incognito window, and I'm not seeing the error any more — and far fewer warnings. These are the warnings I'm seeing:

```
[DOM] Found 2 elements with non-unique id #q: (More info: https://goo.gl/9p2vKq) <input class=​"site-header__searchfield form-control" type=​"search" name=​"q" id=​"q" autocomplete=​"off" placeholder=​"Search" aria-label=​"Search">​ <input class=​"site-header__searchfield form-control" type=​"search" name=​"q" id=​"q" autocomplete=​"off" placeholder=​"Search" aria-label=​"Search">​
(index):1 A cookie associated with a cross-site resource at http://dartpad.dev/ was set without the `SameSite` attribute. A future release of Chrome will only deliver cookies with cross-site requests if they are set with `SameSite=None` and `Secure`. You can review cookies in developer tools under Application>Storage>Cookies and see more details at https://www.chromestatus.com/feature/5088147346030592 and https://www.chromestatus.com/feature/5633521622188032.
(index):1 A cookie associated with a cross-site resource at http://youtube.com/ was set without the `SameSite` attribute. A future release of Chrome will only deliver cookies with cross-site requests if they are set with `SameSite=None` and `Secure`. You can review cookies in developer tools under Application>Storage>Cookies and see more details at https://www.chromestatus.com/feature/5088147346030592 and https://www.chromestatus.com/feature/5633521622188032.
(index):1 A cookie associated with a cross-site resource at http://doubleclick.net/ was set without the `SameSite` attribute. A future release of Chrome will only deliver cookies with cross-site requests if they are set with `SameSite=None` and `Secure`. You can review cookies in developer tools under Application>Storage>Cookies and see more details at https://www.chromestatus.com/feature/5088147346030592 and https://www.chromestatus.com/feature/5633521622188032.
DevTools failed to parse SourceMap: https://jr-dartlang-1.firebaseapp.com/assets/dash/js/dartpad_picker_main.dart.js.map
DevTools failed to parse SourceMap: https://dartpad.dev/experimental/new_embed_dart.dart.js.map
DevTools failed to parse SourceMap: https://dartpad.dev/split.min.js.map
DevTools failed to parse SourceMap: https://dartpad.dev/packages/primer_css/primer.css.map
DevTools failed to parse SourceMap: https://dartpad.dev/primer.css.map```

**Comment by johnpryan** [comment]
we should probably heed those warnings, but right now I think they won't be an issue (and are unrelated to this change I think)

File: tool/dartpad_picker/lib/dartpad_picker.dart [hunk 1]
```diff
@@ -51,7 +51,12 @@ class DartPadPicker {
       ..src = iFrameSrc(theme: 'dark', mode: 'dart');
     iFrameHost.children.add(_iFrameElement);
     window.addEventListener('message', (dynamic e) {
-      if (e.data['type'] == 'ready') {
+      // Don't handle events from other iframe elements
+      if (e.data != null &&
+          e.data is Map &&
+          e.data.containsKey('type') &&
+          e.data['type'] is String &&
+          e.data['type'] == 'ready') {
```
chalin: Hmm, are you saying that there are situations under which `e.data['type'] == 'ready'` is true and yet the new if condition is false? Or is it that the original condition just missing a `?` (as in `e?.data['type'] == 'ready'`), because otherwise it was throwing an exception?

tool/dartpad_picker/compile_dev.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+: ${TMP:=tmp}
+
+set -x
+
+pub get
+pub run build_runner build --output $TMP --define "build_web_compilers|entrypoint=compiler=dart2js"
+cp $TMP/web/* ../../src/assets/dash/js
+rm -rf $TMP
```
chalin: Just curious, when do you use this script (and why compile under dart2js and not the ddc)?

chalin:
I'm still seeing the SameSite error:
> image

tool/dartpad_picker/lib/dartpad_picker.dart
johnpryan: The issue was that the `type` attribute isn't set on every event. The YouTube iframe element sends events to the parent window, which don't have the 'type' attribute This adds a few more checks to make sure we are handling the right event.

It might be better to send a 'dartpad-ready' event instead so that we are more certain that the event is coming from the DartPad embed.

tool/dartpad_picker/compile_dev.sh
johnpryan: Using DDC to debug was tricker than using dart2js in non-release mode. DDC modules expect package source code to be hosted at the root of the project (`localhost:4000/packages/blah`). I think I was seeing an issue where Jekyll wasn't serving the `src/packages/` directory correctly.

johnpryan:
@chalin @kwalrath maybe we could create a separate issue for the cookie warnings?

kwalrath:
I'll create that issue unless someone else is already doing it.

chalin:
You were quicker than I was! :) https://github.com/dart-lang/dart-pad/issues/1378 covers this issue (although is assuming that the problem originates from use of DartPads).

tool/dartpad_picker/lib/dartpad_picker.dart
chalin: If the `type` attribute is missing then the value of `e.data['type']` should be null, according to the Dart [Map [] operator semantics](https://api.dartlang.org/stable/2.6.1/dart-core/Map/operator_get.html). And if the value is null, it can't equal `'ready'`.

Is the issue that, in the compiled JS code, `e.data` is a "native" JS value that isn't then adhering to Dart semantics? Thanks for helping me understand.

(Btw, the `e.data != null` term is unnecessary since `e.data is Map` implies that `e.data` isn't null in Dart.)

PR #2120, opened 2019-11-19
Cleanup linkcheck-skip-list.txt localhost rules

chalin:
Followup to #1728

### DIFF
--- a/tool/config/linkcheck-skip-list.txt
+++ b/tool/config/linkcheck-skip-list.txt
@@ -1,6 +1,6 @@
 # Enter regexp pattern, one per line.
 
-(https?://)?localhost\b
+^(https?://)?localhost\b
 
 # FIXME(Temporary): linkcheck seems to be intermittently failing when accessing main.css
 # https://github.com/dart-lang/site-www/issues/1107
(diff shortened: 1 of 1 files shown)


chalin:
Travis is green. Merging.

kwalrath:
So... just have to check... You're sure that bad in-site URLs will still be detected, right?

chalin:
Do you mean in-site URLs like `/samples`? Then the answer is yes. 
Or are you referring to URLs containing `localhost`? Then the answer is that all URLs referring to localhost will be ignored. Is that ok, or did you prefer having specific URLs for each distinct case of a URL involving localhost?

kwalrath:
I just meant in-site URLs. I thought they were staged using localhost, so I worried about ignoring localhost. :)

If we detect bad in-site links (which I assume you tested, but I can't stop myself from asking), then I'm good with using the new code.

chalin:
Thanks for asking! No I hadn't tested the updated rule locally, I had the update as a note in my local To-Do list and made the change via GH. And ... **you're right that does exclude local in-site links**. Hmm, we used to have a catch all rule for localhost. I think that the difference is that our previous rule didn't include a protocol spec. 

I can't fix this tonight, but I'll do it first thing tomorrow.

PR #2124, opened 2019-11-20
Ensure Try-Dart example sources are analyzed and tested

### LINKED ISSUE #1919: Support loading DartPad sources from site vs Gist

chalin:
- Closes #1919
- **Note:** I've not changed the source code of the examples except where the analyzer reported an issue. I'll signal those via comments.

Staged, see:
- Homepage https://dart-dev-staging-0.firebaseapp.com/#try-dart
- ~Try Dart page: https://dart-dev-staging-0.firebaseapp.com/try-dart~

cc @mit-mit


examples/misc/bin/try_dart/classes.dart
```diff
@@ -0,0 +1,41 @@
...
+
+class Sword implements Item {
+  int get damage => 5;
+
+  use() => print("$this dealt $damage damage.");
+}
+
+// Classes can extend other classes.
+class DiamondSword extends Sword {
+  int get damage => 50;
```
chalin: The original source declared Sword.damage and DiamondSword.damage as a field. The analyzer (rightly) complained that we shouldn't override fields: `Don't override fields`. So I change the fields to getters. Ok?

examples/misc/bin/try_dart/functions.dart
```diff
@@ -0,0 +1,22 @@
...
+  return x * 2;
+}
+
+// Arrow syntax is shorthand for `{ return expr; }`.
+int timesFour(int x) => timesTwo(timesTwo(x));
+
+// Functions are objects.
+int runTwice(int x, int Function(int) f) {
+  for (var i = 0; i < 2; i++) {
+    x = f(x);
```
chalin: **Difference relative to the original source:** The analyzer wouldn't allow an untyped `f` to be used to assigned to the typed `x`, so I declared `f` as a function from int to int.

examples/misc/lib/pi_monte_carlo.dart
```diff
@@ -20,6 +22,7 @@ int numIterations = 500; //!web-only
 // after 500 iterations). //!web-only
 //!web-only
 //!tip("main()")
+// #docregion try-dart
 main() async {
```
chalin: Rather than create another "Compute Pi" that was console-only, I'm reusing the HTML version, and stripping out the HTML-specific code. (This version of the code actually had updated syntax as compared to the original try-dart code.)

examples/misc/bin/try_dart/classes.dart
mit-mit: Can we move the updating of the examples to a separate PR, please?

mit-mit:
I'd really prefer to not introduce a second page with DartPad on it. That would leave us with three places: 1) landing page, 2) try-dart page, 3) dartpad.dev. 

My idea was to add some more white space around the dartpad embed on the landing page so it seems more focused. Checkout this demo:
https://mit-staging.firebaseapp.com/#try-dart

Note that it currently is broken in the mobile/narrow view, I left a TODO, should be easy to fix (?) for someone who unlike me actually knows html & css.

examples/misc/bin/try_dart/classes.dart
chalin: Yes, will do.

chalin: Code change undone.

examples/misc/lib/pi_monte_carlo.dart
chalin: Code change undone.

examples/misc/bin/try_dart/functions.dart
chalin: Code change undone.

examples/misc/bin/try_dart/classes.dart
chalin: I've created #2126 to track this, and offer a place for discussion.

chalin:
> I'd really prefer to not introduce a second page with DartPad on it. ...

I've removed the trial /try-dart page.

> My idea was to add some more white space around the dartpad embed ...

I've created #2127 to track this and offer a place for discussion in case @galeyang et all might want to offer some advice.

Staging server refreshed with most recent changes.

chalin:
Btw, the current approach always loads the try-dart example sources (because the sources are embedded in the homepage). If you believe that that is too inefficient for mobile (where we don't show the embedded DartPad), then I can move the try-dart example sources to another page that is only conditionally fetched. Of course there are tradeoffs to having the sources on another page, since that other page might fail to load -- not a situation that we need to deal with in the current approach.

examples/misc/bin/try_dart/classes.dart
```diff
@@ -0,0 +1,41 @@
...
+
+class Sword implements Item {
+  int damage = 5;
+
+  use() => print("$this dealt $damage damage.");
+}
+
+// Classes can extend other classes.
+class DiamondSword extends Sword {
+  int damage = 50; // ignore: overridden_fields
```
johnpryan: Let's instead change the `damage` field to a getter to make this warning go away

examples/misc/bin/try_dart/functions.dart
```diff
@@ -0,0 +1,22 @@
...
+  return x * 2;
+}
+
+// Arrow syntax is shorthand for `{ return expr; }`.
+int timesFour(int x) => timesTwo(timesTwo(x));
+
+// Functions are objects.
+int runTwice(int x, Function f) {
+  for (var i = 0; i < 2; i++) {
+    x = f(x); // ignore: invalid_assignment
```
johnpryan: would using a local variable make this warning go away?

examples/misc/bin/try_dart/functions.dart
```diff
@@ -0,0 +1,22 @@
...
+  }
+  return x;
+}
+
+// ignore: type_annotate_public_apis
```
johnpryan: let's make the return type `void` instead.

examples/misc/bin/try_dart/control_flow.dart
```diff
@@ -0,0 +1,26 @@
...
+  // A for-in loop.
+  for (var i in numbers) {
+    // A single-line if statement.
+    if (isEven(i)) evenNumbers.add(i);
+  }
+
+  return evenNumbers;
+}
+
+// ignore: type_annotate_public_apis
```
johnpryan: same here, let's make the return type `void`.

examples/misc/bin/try_dart/collection_literals.dart
```diff
@@ -0,0 +1,23 @@
...
+};
+
+// A set literal.
+var frogs = {
+  'Tree',
+  'Poison dart',
+  'Glass',
+};
+
+main() {
```
johnpryan: does this need a `void` return type to pass analysis?

examples/misc/bin/try_dart/functions.dart
```diff
@@ -0,0 +1,22 @@
+// A function declaration.
+int timesTwo(int x) {
+  return x * 2;
+}
+
+// Arrow syntax is shorthand for `{ return expr; }`.
+int timesFour(int x) => timesTwo(timesTwo(x));
+
+// Functions are objects.
+int runTwice(int x, Function f) {
```
johnpryan: Why did this change from `int Function(int) f`?

chalin: It currently is `Function f` on dart.dev. I changed it to `int Function(int) f` in the first commit to this PR, and then reverted the change so that we could separately discuss and address any required code changes in another PR.

examples/misc/bin/try_dart/collection_literals.dart
chalin: If we want to drop the ignore directive in line 1, then yes.

examples/misc/bin/try_dart/functions.dart
johnpryan: ok, no problem.

examples/misc/bin/try_dart/functions.dart
chalin: No, but giving `f` the type of `int Function(int)` works.

chalin:
@johnpryan - aside from the examples sources, any feedback on the changes to the other files -- including `tool/dartpad_picker/lib/dartpad_picker.dart`?

tool/dartpad_picker/web/dartpad_picker_main.dart
```diff
@@ -13,224 +15,41 @@ void main() {
...
 
+Snippet _createSnippet(HtmlElement root, String name) => Snippet(
+      name,
+      _getSrc(
+        root,
+        '#try-dart-${name.toLowerCase().replaceAll(' ', '-')}',
+      ),
+    );
+
+String /*?*/ _getSrc(HtmlElement root, String id, [String fallback]) =>
```
johnpryan: What does the `/*?*/` do?

chalin: Oh, thanks for catching that. The function originally returned a nullable String, but it doesn't anymore. (I use `/*?*/` to mark types as nullable, until we can use the `?` syntax for real :)).

chalin: Removed comment.

chalin:
Travis is green, merging. Followup changes will be tracked via #2126.

PR #2134, opened 2019-11-25
Fix cases of method named with get

natebosch:
https://dart.dev/guides/language/effective-dart/design#avoid-starting-a-method-name-with-get

PR #2137, opened 2019-11-25
Closes #2128 : Change get* function names to fetch* function names.

### LINKED ISSUE #2128: Asynchronous programming codelab: consider updates to example code
Moving the discussion from https://github.com/dart-lang/site-www/pull/2114#discussion_r348243358 to this issue so that PR #2114 can contain only the original code. Copying the relevant parts of the discussion below.

Also consider that the **analyzer** (with the package using the `pedantic` analyzer options), **raises issues** with quite a few samples of the code, including: `await_only_futures`, `argument_type_not_assignable`, `invalid_assignment`, `type_annotate_public_apis` (which is the most common, but probably the least problematic). This should probably be addressed. In #2114, we used `ignore` directives to silence the analyzer.

---

Here's part of the old code that is relevant to this discussion:

```
}

}

main() async {
  countSeconds(4);
  await createOrderMessage();
}
```

    >
    > ```dart
    > Future<void> main() async {
    > ```

    So, ...

Thoughts?

---

Btw, when `createOrderMessage()` is declared as `void`, we get the following issue reported by the analyzer:

```nocode
```

legalcodes:
Original issue:
https://github.com/dart-lang/site-www/issues/2128

This PR incorporates changes from https://github.com/dart-lang/site-www/pull/2134/commits/072dcc6bbc2ae27d3359348430989aebd4e5d254 that were ultimately removed from https://github.com/dart-lang/site-www/pull/2134.

Staged:
https://jt-dartdev-1.firebaseapp.com/codelabs/async-await

### DIFF
--- a/src/codelabs/async-await.md
+++ b/src/codelabs/async-await.md
@@ -192,7 +192,7 @@ function.
 First, add the `async` keyword before the function body:
 
 {% prettify dart %}
-main() [!async!] {
+void main() [!async!] {
 {% endprettify %}
 
 {{site.alert.note}}
--- a/src/codelabs/async-await.md
+++ b/src/codelabs/async-await.md
@@ -40,7 +40,7 @@ and the `async` and `await` keywords.
 
 ### Example: Incorrectly using an asynchronous function
 The following example shows the wrong way to use an asynchronous function
-(`getUserOrder()`). Later you'll fix the example using `async` and `await`.
+(`fetchUserOrder()`). Later you'll fix the example using `async` and `await`.
 Before running this example, try to spot the issue -- what do you think the
 output will be?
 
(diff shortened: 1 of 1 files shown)


src/codelabs/async-await.md
```diff
@@ -217,7 +217,7 @@ Now that you have an `async` function, you can use the `await` keyword to wait
 for a future to complete:
 
 ```dart
-print(await createOrderMessage());
+print(await printOrderMessage());
```
chalin: This instance of `createOrderMessage()` should not have been renamed. It doesn't make sense to have `print(await printFoo())`. As @natebosch pointed out, the rename should have been applied only to those definitions of `createOrderMessage()` that printed the order message in their body -- @natebosch [wrote](https://github.com/dart-lang/site-www/issues/2128#issuecomment-558307422) (emphasis mine):

> +1 for changing the name between `create` and `print`
> **depending on the behavior, return vs print**.

As I mentioned in point (2) of https://github.com/dart-lang/site-www/issues/2128#issue-526758837, the function name `createOrderMessage` was being used for *two different purposes*:
1. In one case to name a function that _returns an order message_, and 
2. In the other case to name a function that _prints an order message_.

My suggestion (and as I understand it, @natebosch's suggestion as well) was to rename to `printOrderMessage` _only_ those occurrences of `createOrderMessage()` that printed the message and returned nothing.

src/codelabs/async-await.md
```diff
@@ -233,12 +233,12 @@ your window is wide enough — is to the right of the synchronous example.
 
 ```dart
 // Synchronous
-String createOrderMessage() {
-  var order = getUserOrder();
+String printOrderMessage() {
+  var order = fetchUserOrder();
   return 'Your order is: $order';
 }
```
chalin: This is the function definition of `createOrderMessage()` that should remain named `createOrderMessage()` and that I was referring to my previous comment.

chalin: Btw, the _other_ definition of `createOrderMessage()` is currently only found in the following Gist: https://gist.github.com/d7abfdea1ae5596e96c7c0203d975dba

src/codelabs/async-await.md
natebosch: ah, yes. Sorry I missed this during my review.

legalcodes: @chalin Would you be amenable to leaving `printOrderMessage()` in, and removing the `print()` method that's wrapping it?

src/codelabs/async-await.md
```diff
@@ -192,7 +192,7 @@ function.
 First, add the `async` keyword before the function body:
 
 {% prettify dart %}
-main() [!async!] {
+void main() [!async!] {
 {% endprettify %}
 
 {{site.alert.note}}
```
chalin: This note should be dropped now that `main` has a declared return type. (This codelab focuses on async programming, IMHO, we don't need to mention optional types.)

src/codelabs/async-await.md
chalin: Sure, although I thought that part of the goal was to show a function with a non-`void` return type evolve from a(n incorrect) use of `T` to `Future<T>`.

In the end, it's your call, but if you do opt to making the change that you propose, then the codelab's text should be carefully reviewed to ensure that it still makes sense with the updated code (e.g. to avoid things like https://github.com/dart-lang/site-www/pull/2137/files#r351422017).

IMHO, it would be easier to restore the original `createOrderMessage()` function name (for those places where the body actually returns a string).

legalcodes: I'm curious to get your thoughts on [the following example](https://gist.github.com/25ade03f0632878a9169209e3cd7bef2): 
Screen Shot 2019-11-27 at 9 47 17 AM

The function both prints and "creates" an order message (uses the result from fetching the user order) but doesn't explicitly return. Do you have an opinion about the name we should use?

chalin: Would go with naming the first function in that example as `printOrderMessage()`.

PR #2141, opened 2019-11-27
Closes #2128 Reinstate `createOrderMessage()` in cases lacking print statements

### LINKED ISSUE #2128: Asynchronous programming codelab: consider updates to example code

legalcodes:


src/codelabs/async-await.md
```diff
@@ -247,9 +238,9 @@ Future<String> fetchUserOrder() {
 }
 
 // Synchronous
-main() {
+void main() {
```
chalin: Ah good, this was the missing `void` I had noted from your previous commit.

src/codelabs/async-await.md
```diff
@@ -195,15 +195,6 @@ First, add the `async` keyword before the function body:
 void main() [!async!] {
 {% endprettify %}
 
-{{site.alert.note}}
```
chalin: 👍

PR #2155, opened 2019-12-06
Extension methods page

### LINKED ISSUE #2086: Document extension methods
[Design doc](https://github.com/dart-lang/language/blob/master/accepted/2.6/static-extension-members/feature-specification.md)
[Article](https://medium.com/dartlang/extension-methods-2d466cd8b308)

kwalrath:
Staged: https://kw-www-dartlang-1.firebaseapp.com/guides/language/extension-methods

Presentable, but I still need to upload testable code excerpts ~~and add something to the language tour~~. Also, I should expand the generic section.

Fixes #2086

### DIFF
--- a/src/_data/side-nav.yml
+++ b/src/_data/side-nav.yml
@@ -22,12 +22,6 @@
 - title: Language
   expanded: false
   children:
-    - title: Tour
-      permalink: /guides/language/language-tour
-    - title: Type system
-      permalink: /guides/language/sound-dart
-    - title: Specification
-      permalink: /guides/language/spec
     - title: Effective Dart
       expanded: false
       children:
@@ -43,6 +37,14 @@
           permalink: /guides/language/effective-dart/usage
         - title: Design
           permalink: /guides/language/effective-dart/design
+    - title: Extension methods
+      permalink: /guides/language/extension-methods
+    - title: Specification
+      permalink: /guides/language/spec
+    - title: Tour
+      permalink: /guides/language/language-tour
+    - title: Type system
+      permalink: /guides/language/sound-dart
 
 - title: Core libraries
   expanded: false
--- a/src/_guides/language/extension-methods.md
+++ b/src/_guides/language/extension-methods.md
@@ -1,0 +1,246 @@
+---
+title: Extension methods
+description: Learn how to add to existing APIs.
+---
+<?code-excerpt replace="/ *\/\/\s+ignore_for_file:[^\n]+\n//g; /([A-Z]\w*)\d\b/$1/g; /\b(main)\d\b/$1/g"?>
+
+Extension methods, introduced in Dart 2.7,
+are a way to add functionality to existing libraries.
+You might not ever need to _use_ extension methods, much less _create_ them.
+But if you generate code or find yourself writing lots of wrappers,
+consider using extension methods to reduce
+the amount of boilerplate code that you need to write.
+
+You might use extension methods without even knowing it.
+For example, when you use code completion in an IDE,
+it suggests extension methods alongside regular methods.
+
+{% comment %}
+[PENDING: embed video here.]
+Watch this video for an introduction to extension methods.
+{% endcomment %}
+
+## Overview
+
+When you’re using someone else’s API or
+when you implement a library that’s widely used,
+it’s often impractical or impossible to change the API.
+But you might still want to add some functionality.
+
+For example, consider the following code that parses a string into an integer:
+
+```dart
+int.parse('42')
+```
+
+It might be nice — shorter and easier to use with tools — to
+have that functionality be in the string instead:
+
+```dart
+'42'.toInt()
+```
+
+To enable that code,
+you can import a library that contains an extension of the String class:
+
+```dart
+import string_apis.dart';
+...
+print('42'.toInt());
+```
+
+Extensions can define not just methods and getters,
+but also other members such as setters and operators.
+Also, extensions have names, which can be helpful if an API conflict arises.
+Here's how you might implement the extension method `toInt()`,
+using an extension (named `ParseNumbers`) that operates on strings:
+
+```dart
+extension ParseNumbers on String {
+  int toInt() {
+    return int.parse(this);
+  }
+}
+```
+
+The next section describes how to _use_ extension methods.
+After that are sections about _implementing_ extension methods.
+
+
+## Using extension methods
+
+Like all Dart code, extension methods are in libraries.
+You’ve already seen how to use an extension method —
+just import the library it’s in, and use it like an ordinary method:
+
+```dart
+// Import a library that contains an extension on String. 
+import string_apis.dart';
+...
+print('42'.padLeft(5)); // Use a String method.
+print('42'.toInt()); // Use an extension method.
+```
+
+That’s all you usually need to know to use extension methods.
+As you write your code, you might also need to know
+how extension methods depend on static types (as opposed to `dynamic`) and
+how to resolve [API conflicts](#api-conflicts).
+
+### Static types and dynamic
+
+You can’t invoke extension methods on variables of type `dynamic`.
+For example, the following code results in a runtime exception:
+
+```dart
+dynamic d = '2';
+print(d.toInt()); // Runtime exception: NoSuchMethodError
+```
+
+Extension methods _do_ work with Dart’s type inference.
+The following code is fine because
+the variable `v` is inferred to have type `String`:
+
+```dart
+var v = '2';
+print(v.toInt()); // Output: 2
+```
+
+The reason that `dynamic` doesn’t work is that
+extension methods are resolved against the static type of the receiver.
+For more information about static types and `dynamic`, see
+[The Dart type system](/guides/language/sound-dart).
+
+### API conflicts
+
+If an extension member conflicts with
+an interface or with another extension member,
+then you have a few options.
+
+One option is changing how you import the conflicting extension.
+You can use `show` or `hide`, or you can remove the import:
+
+```dart
+// Both libraries define the String extension method toInt().
+// The extensions might have the same name or different names.
+import 'string_apis.dart';
+import 'string_apis_2.dart' hide ParseNumbers2;
+...
+// Use the toInt() defined in 'string_apis.dart'.
+print('42'.toInt());
+```
+
+Another option is applying the extension explicitly, which results in code that looks as if the extension is a wrapper class:
+
+```dart
+// Both libraries define extensions on String that contain toInt(),
+// and the extensions have different names.
+import 'string_apis.dart'; // Contains ParseNumbers extension.
+import 'string_apis_2.dart'; // Contains ParseNumbers2 extension.
+...
+//  print('42'.toInt()); // Doesn't work.
+print(ParseNumbers('42').toInt());
+print(ParseNumbers2('42').toInt());
+```
+
+If both extensions have the same name,
+then you might need to import using a prefix:
+
+```dart
+// Both libraries define extensions named ParseNumbers
+// that contain the extension method toInt(). One ParseNumbers
+// extension (in 'string_apis_3.dart') also defines toNum().
+import 'string_apis.dart';
+import 'string_apis_3.dart' as rad;
+...
+//  print('42'.parseInt()); // Doesn't work.
+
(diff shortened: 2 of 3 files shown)


src/_data/side-nav.yml
```diff
@@ -43,6 +37,14 @@
           permalink: /guides/language/effective-dart/usage
         - title: Design
           permalink: /guides/language/effective-dart/design
+    - title: Extension methods
```
mit-mit: I think this very specific topic should move below the much more general spec, tour, and type system topics

src/_guides/language/extension-methods.md
```diff
@@ -0,0 +1,246 @@
+---
+title: Extension methods
+description: Learn how to add to existing APIs.
+---
+<?code-excerpt replace="/ *\/\/\s+ignore_for_file:[^\n]+\n//g; /([A-Z]\w*)\d\b/$1/g; /\b(main)\d\b/$1/g"?>
+
+Extension methods, introduced in Dart 2.7,
+are a way to add functionality to existing libraries.
+You might not ever need to _use_ extension methods, much less _create_ them.
```
mit-mit: I think this is an overstatement. I'd be very surprised if the majority of Dart developers were using at least one extension methods just a few months from now. We're even discussing adding a few extensions directly to the Flutter framework (as that would be non-breaking as opposed to changing core classes)

src/_guides/language/extension-methods.md
```diff
@@ -0,0 +1,246 @@
...
+But if you generate code or find yourself writing lots of wrappers,
+consider using extension methods to reduce
+the amount of boilerplate code that you need to write.
```
mit-mit: To me the main value add is not less boilerplate. It's offering APIs that as as accessible as-if they had been added to the original class.

src/_guides/language/extension-methods.md
```diff
@@ -0,0 +1,246 @@
...
+
+You might use extension methods without even knowing it.
+For example, when you use code completion in an IDE,
+it suggests extension methods alongside regular methods.
```
mit-mit: I wonder if we rank extension methods differently from other methods. Maybe check with the analyzer team?

src/_guides/language/extension-methods.md
```diff
@@ -0,0 +1,246 @@
...
+
+When you’re using someone else’s API or
+when you implement a library that’s widely used,
+it’s often impractical or impossible to change the API.
+But you might still want to add some functionality.
+
+For example, consider the following code that parses a string into an integer:
+
+```dart
+int.parse('42')
```
mit-mit: Missing terminating semicolon

src/_guides/language/extension-methods.md
```diff
@@ -0,0 +1,246 @@
...
+```
+
+It might be nice — shorter and easier to use with tools — to
+have that functionality be in the string instead:
+
+```dart
+'42'.toInt()
```
mit-mit: Missing terminating semicolon

src/_guides/language/extension-methods.md
```diff
@@ -0,0 +1,246 @@
...
+To enable that code,
+you can import a library that contains an extension of the String class:
+
+```dart
+import string_apis.dart';
+...
+print('42'.toInt());
+```
+
+Extensions can define not just methods and getters,
```
mit-mit: The grouping of methods + getters vs setters + operators in other is a bit surprising. Maybe change to "Extensions can define not just methods and getters,
 

 
The grouping of methods + getters vs setters + operators in other is a bit surprising. Maybe change to "Extensions can define not just methods, but also other members such as getters, setters, and operators." ?

src/_guides/language/extension-methods.md
```diff
@@ -0,0 +1,246 @@
...
+But you might still want to add some functionality.
+
+For example, consider the following code that parses a string into an integer:
+
+```dart
+int.parse('42')
+```
+
+It might be nice — shorter and easier to use with tools — to
+have that functionality be in the string instead:
```
mit-mit: s/in the string/on String/

src/_guides/language/extension-methods.md
```diff
@@ -0,0 +1,246 @@
...
+}
+```
+
+The next section describes how to _use_ extension methods.
+After that are sections about _implementing_ extension methods.
+
+
+## Using extension methods
+
+Like all Dart code, extension methods are in libraries.
```
mit-mit: General comment, outside of this feature: I wonder how often developers deal with libraries? Is it not much more common to deal with (import etc.) packages these days?

src/_guides/language/extension-methods.md
```diff
@@ -0,0 +1,246 @@
...
+The following code is fine because
+the variable `v` is inferred to have type `String`:
+
+```dart
+var v = '2';
+print(v.toInt()); // Output: 2
+```
+
+The reason that `dynamic` doesn’t work is that
+extension methods are resolved against the static type of the receiver.
```
mit-mit: Maybe mention the perf gain this offers?

src/_guides/language/extension-methods.md
```diff
@@ -0,0 +1,246 @@
...
+[The Dart type system](/guides/language/sound-dart).
+
+### API conflicts
+
+If an extension member conflicts with
+an interface or with another extension member,
+then you have a few options.
+
+One option is changing how you import the conflicting extension.
+You can use `show` or `hide`, or you can remove the import:
```
mit-mit: I would remove the "remove the import" part. That seems a bit given...

src/_guides/language/extension-methods.md
```diff
@@ -0,0 +1,246 @@
...
+
+```dart
+// Both libraries define the String extension method toInt().
```
mit-mit: I think "the extension method toInt() on String" might be clearer?

src/_guides/language/extension-methods.md
```diff
@@ -0,0 +1,246 @@
...
+// The extensions might have the same name or different names.
```
mit-mit: I'd just leave this out

src/_guides/language/extension-methods.md
```diff
@@ -0,0 +1,246 @@
...
+import 'string_apis.dart';
+import 'string_apis_2.dart' hide ParseNumbers2;
+...
+// Use the toInt() defined in 'string_apis.dart'.
```
mit-mit: I think I'd move the comment up to the import statement with the hide in it; that's the part that resolves the conflict. This line with the call just invokes it *after* the conflict has been resolved

src/_guides/language/extension-methods.md
```diff
@@ -0,0 +1,246 @@
...
+print('42'.toInt());
+```
+
+Another option is applying the extension explicitly, which results in code that looks as if the extension is a wrapper class:
```
mit-mit: long line

src/_guides/language/extension-methods.md
```diff
@@ -0,0 +1,246 @@
...
+
+```dart
+// Both libraries define extensions on String that contain toInt(),
+// and the extensions have different names.
+import 'string_apis.dart'; // Contains ParseNumbers extension.
+import 'string_apis_2.dart'; // Contains ParseNumbers2 extension.
```
mit-mit: Maybe NumberParsing to make it a bit more realistic?

src/_guides/language/extension-methods.md
```diff
@@ -0,0 +1,246 @@
...
+either omit the extension name or give it a name
+that starts with an underscore (`_`).
+
+The members of the extension can be methods, getters, setters, operators.
+Extensions can also have static fields and static helper methods.
+
+## Implementing generic extensions
+
+Extensions can have generic type parameters.
+For example, here’s some code that extends the built-in `List<T>` type:
```
mit-mit: maybe add "with a getter and two methods"?

src/_guides/language/language-tour.md
```diff
@@ -3281,6 +3285,32 @@ For more information, see the informal
 [noSuchMethod forwarding specification.](https://github.com/dart-lang/sdk/blob/master/docs/language/informal/nosuchmethod-forwarding.md)
 
 
+### Extension methods
+
+Extension methods, introduced in Dart 2.7,
+are a way to add functionality to existing libraries.
+You might not ever need to _use_ extension methods, much less _create_ them.
```
mit-mit: Same comment as above. Especially not needed here in the tour, I think.

src/_guides/language/language-tour.md
```diff
@@ -3281,6 +3285,32 @@ For more information, see the informal
...
+But if you generate code or find yourself writing lots of wrappers,
+consider using extension methods to reduce
+the amount of boilerplate code that you need to write.
+
+You might use extension methods without even knowing it.
```
mit-mit: Here in the tour, I'd move this below the example.

src/_data/side-nav.yml
kwalrath: The problem is that after the list reaches a certain length, it's hard to find items, so I tend to go to alpha order so it's predictable. The tour is the most visited page by far in this list, but I suspect that most visitors don't get to it from this sidenav.

mit-mit: Sure, but we don't need extension methods to be easily locatable in the ToC. I don't think it's something we need users to stumble into organically. If they need it, they will find in in the tour, or search for it.

On the other hand we do want users to stumble into the tour and type system pages.

src/_guides/language/extension-methods.md
kwalrath: It's just a code fragment, not a statement. If we added a semicolon, it'd be a useless statement... That said, I can do it. It'd be easier to extract & test.

src/_guides/language/extension-methods.md
kwalrath: I'm going to change these back to parseInt, FYI (per Bob's statement that "to*" shouldn't be subject to failure).

src/_guides/language/extension-methods.md
kwalrath: I'm not sure what you mean... You deal with packages in the pubspec, but in Dart code you import and deal with library files (which are often inside of packages), no? And when I go to the API reference, I see libraries (though the imports aren't usually in the API docs, sadly).

src/_guides/language/extension-methods.md
kwalrath: It's a getter, an operator, and a method, right?

src/_guides/language/extension-methods.md
kwalrath: actually, I think I'll say it in both places:

```
// Defines the String extension method parseInt().
import 'string_apis.dart';

// Also defines parseInt(), but hiding NumberParsing2 
// hides that extension method. 
import 'string_apis_2.dart' hide NumberParsing2;
...
// Uses the parseInt() defined in 'string_apis.dart'.
print('42'.parseInt());
```

src/_guides/language/extension-methods.md
kwalrath: I hesitate to get too specific about exactly where the extension methods are displayed.

src/_guides/language/extension-methods.md
kwalrath: Yep. This is a leftover from old code that used a getter, not an extension method.

src/_guides/language/extension-methods.md
kwalrath: I've rewritten it slightly.

kwalrath: Proposed:

> One option is changing how you import the conflicting extension,
using `show` or `hide` to limit the exposed API:

src/_guides/language/extension-methods.md
kwalrath: Here's what I'm thinking of:

```
// Defines the String extension method parseInt().
import 'string_apis.dart';

// Also defines parseInt(), but hiding NumberParsing2 
// hides that extension method. 
import 'string_apis_2.dart' hide NumberParsing2;
...
// Uses the parseInt() defined in 'string_apis.dart'.
print('42'.parseInt());
```

src/_guides/language/language-tour.md
kwalrath: I rearranged it a bit.

PR #2156, opened 2019-12-06
Runes -> Characters

kwalrath:
Staged:
https://kw-staging-dartlang-2.firebaseapp.com/guides/language/language-tour#characters
https://kw-staging-dartlang-2.firebaseapp.com/guides/libraries/useful-libraries

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -986,10 +986,9 @@ For more information about maps, see
 <a id="characters"></a>
 ### Runes and grapheme clusters
 
-In Dart, [runes][] are the Unicode code points of a string.
-As of Dart 2.6, instead of using runes use the
-[characters package,][characters package]
-which has better support for user-perceived characters,
+In Dart, [runes][] expose the Unicode code points of a string.
+As of Dart 2.6, use the [characters package][]
+to view or manipulate user-perceived characters,
 also known as
 [Unicode (extended) grapheme clusters.][grapheme clusters]
 
@@ -1013,7 +1012,6 @@ a sequence of grapheme clusters.
 Here's an example of using the characters API:
 
 {% comment %}
-TODO: add link to extension methods section?
 TODO: add test code
 {% endcomment %}
 
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -610,8 +610,8 @@ You can create a “raw” string by prefixing it with `r`:
 var s = r'In a raw string, not even \n gets special treatment.';
 ```
 
-See [Runes](#runes) for details on how to express Unicode
-characters in a string.
+See [Runes and grapheme clusters](#runes) for details on how
+to express Unicode characters in a string.
 
 Literal strings are compile-time constants,
 as long as any interpolated expression is a compile-time constant
@@ -983,16 +983,21 @@ For more information about maps, see
 [Generics](#generics) and
 [Maps](/guides/libraries/library-tour#maps).
 
-### Runes
+<a id="runes"></a>
+### Runes and grapheme clusters
 
 In Dart, runes are the UTF-32 code points of a string.
+As of Dart 2.6, instead of using runes use the
+[characters package,][characters package]
+which has better support for user-perceived characters,
+also known as
+[Unicode (extended) grapheme clusters.][grapheme clusters]
 
 Unicode defines a unique numeric value for each letter, digit,
 and symbol used in all of the world's writing systems.
 Because a Dart string is a sequence of UTF-16 code units,
 expressing 32-bit Unicode values within a string requires
 special syntax.
-
 The usual way to express a Unicode code point is
 `\uXXXX`, where XXXX is a 4-digit hexadecimal value.
 For example, the heart character (♥) is `\u2665`.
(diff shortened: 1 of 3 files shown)


src/_guides/language/language-tour.md
```diff
@@ -983,66 +983,36 @@ For more information about maps, see
 [Generics](#generics) and
 [Maps](/guides/libraries/library-tour#maps).
 
-### Runes
+<a id="runes"></a>
+### Runes and grapheme clusters
 
 In Dart, runes are the UTF-32 code points of a string.
+As of Dart 2.6, instead of using runes use the
```
mit-mit: Link `runes` to https://api.dartlang.org/stable/2.7.0/dart-core/Runes-class.html ?

src/_guides/language/language-tour.md
```diff
@@ -610,8 +610,8 @@ You can create a “raw” string by prefixing it with `r`:
 var s = r'In a raw string, not even \n gets special treatment.';
 ```
 
-See [Runes](#runes) for details on how to express Unicode
-characters in a string.
+See [Runes and grapheme clusters](#runes) for details on how
```
mit-mit: I suggest we make the anchor link `#characters` given that's the recommendation

src/_guides/language/language-tour.md
```diff
@@ -983,66 +983,36 @@ For more information about maps, see
...
 [Maps](/guides/libraries/library-tour#maps).
 
-### Runes
+<a id="runes"></a>
+### Runes and grapheme clusters
 
 In Dart, runes are the UTF-32 code points of a string.
+As of Dart 2.6, instead of using runes use the
+[characters package,][characters package]
+which has better support for user-perceived characters,
```
mit-mit: @lrhn are there any use cases where `runes` are preferred?

src/_guides/language/language-tour.md
```diff
@@ -983,66 +983,36 @@ For more information about maps, see
...
-
-<iframe
-src="{{site.dartpad-embed-inline}}?id=589bc5c95318696cefe5"
-    width="100%"
-    height="400px"
-    style="border: 1px solid #ccc;">
-</iframe>
+If you need to read or write individual Unicode characters,
+use the `Characters` class from the characters package.
+A `Characters` object is a string viewed as a sequence of grapheme clusters.
```
mit-mit: Can we show some small example here, please? I think you can just grab part of the `example.dart` from the package.

src/_guides/language/language-tour.md
kwalrath: I was trying to preserve the existing anchor name. Probably not a big issue, though.

src/_guides/language/language-tour.md
kwalrath: ptal

mit-mit: Nice, thanks!

mit-mit:
lgtm once we can clarification from Lasse on the Runes question above

src/_guides/language/language-tour.md
lrhn: When you want to work at the code-point level.

Reasons for doing that could be: 
* Converting the string to UTF-8 or UTF-32.
* Representing code points, like an escaped string,`"\u{10feed}"`, or Unicode code-point notation for debugging, `U+10FEED`.
* Implementing regular expressions (which do not understand grapheme clusters anyway).
* Writing the `characters` package :)

In practice, all of those currently work directly on the UTF-16 code-units to avoid the extra abstraction layer, but they are examples of operations that do work on code points, not grapheme clusters.

Not something most users should worry about.

src/_guides/language/language-tour.md
```diff
@@ -983,66 +983,62 @@ For more information about maps, see
 [Generics](#generics) and
 [Maps](/guides/libraries/library-tour#maps).
 
-### Runes
+<a id="characters"></a>
+### Runes and grapheme clusters
 
-In Dart, runes are the UTF-32 code points of a string.
+In Dart, [runes][] are the UTF-32 code points of a string.
```
lrhn: Nitpick: Use "Unicode code points" or "UTF-32 code units" (preferably the former). 

Code points are abstract integers with no size. UTF-32 is an encoding of those into machine numbers, like UTF-8 and UTF-16, except that the *code unit* is a 32-bit integer, which again means that each code point can be represented as one code unit.

src/_guides/language/language-tour.md
```diff
@@ -983,66 +983,62 @@ For more information about maps, see
...
+As of Dart 2.6, instead of using runes use the
+[characters package,][characters package]
+which has better support for user-perceived characters,
+also known as
+[Unicode (extended) grapheme clusters.][grapheme clusters]
 
 Unicode defines a unique numeric value for each letter, digit,
 and symbol used in all of the world's writing systems.
 Because a Dart string is a sequence of UTF-16 code units,
 expressing 32-bit Unicode values within a string requires
```
lrhn: "32-bit Unicode values" is vague and probably misleading. Just use "Unicode code points".

Code points are (so far) only 21-bit numbers.

kwalrath:
Good to go? These work with 2.6, so no reason to wait, right?

src/_guides/language/language-tour.md
mit-mit: @kwalrath I think it's worthvile explaining the code point part of runes from a positioning view. Maybe something like change:

> In Dart, [runes][] are the Unicode code points of a string. As of Dart 2.6, instead of using runes use the [characters package,][characters package] which has better support for user-perceived characters,

to:

> In Dart, [runes][] expose the Unicode code points of a string. As of Dart 2.6, use the [characters package,][characters package] to view or manipulate user-perceived characters.

kwalrath: Done. Ready to go? (I still need to add a test for the code, but that can be in a separate PR.)

PR #2170, opened 2019-12-11
Analyzable code excerpts for extension-methods

kwalrath:
Staged: https://kw-www-dartlang-1.firebaseapp.com/guides/language/extension-methods


mit-mit:
I find some of the indentation to select lines of the code a bit confusion. Was that intentional? E.g.:
Screenshot 2019-12-12 at 11 06 29

kwalrath:
Yeah, I couldn't figure out how to make the indentation look better. :(

@chalin, do you have any advice?

chalin:
The code-excerpt tool respects the original code's indentation, which is why we're getting that output. In normal code, we wouldn't have a `print()` statement at the same level as an import.

So one solution is to include a little context, like this (possibly adding highlights on the point of call of the extension method):

```dart
import 'foo';

void main() {
  print(...);
}
```

If you find that too distracting, then trim the `print()` statement's leading spaces using the `replace="..."` code-excerpt-instruction argument.

kwalrath:
@mit-mit I think I've fixed the indentation. PTAL

kwalrath:
I'm going to submit this TBR, since it fixes a few buglets (in comments, iirc) in the samples.

PR #2172, opened 2019-12-11
Fix missing quote characters

rokob:
In two examples there are missing single quotes before the path to files. This results in the syntax highlighting being wonky.


mit-mit:
Thanks for the fix!

kwalrath:
Yes, thanks, @rokob! These issues were fixed by #2170... but that PR hasn't landed as fast as I'd hoped, so it's good to get these fixed now.

PR #2177, opened 2019-12-17
Improve collapsed contents on smaller screens

joshvocal:
Staged: https://jv-dart-dev-2.firebaseapp.com/guides/language/sound-dart

after_contents

Original: https://dart.dev/guides/language/sound-dart

before_contents

Try resizing the screen on desktop.

When collapsed, some of the contents would be cut off with > 10 items.
Instead, don't show any of the contents until expanded.

### DIFF
--- a/src/_assets/css/main.scss
+++ b/src/_assets/css/main.scss
@@ -544,7 +544,7 @@ img {
 
     &.toc-collapsed {
       .section-nav {
-        max-height: 72px;
+        max-height: 0px;
         overflow: hidden;
       }
 
@@ -554,6 +554,10 @@ img {
 
       .toc-toggle-down {
         display: inline-block;
+      }
+
+      .site-toc__title {
+        margin: 0px;
       }
     }
   }


kwalrath:
Thanks for the PR and for staging it!

I'm concerned that people won't notice **Contents** unless we show at least a few of the entries. (People don't read!) Granted, it's ugly/weird that https://dart.dev/guides/language/sound-dart shows ~3.2 lines (a "..." would be better).

While we're here... Do you know why https://jv-dart-dev-2.firebaseapp.com/guides/language/effective-dart and https://dart.dev/guides/language/effective-dart both show the whole TOC? I thought they'd be 0 items and ~3.2 items, respectively.

joshvocal:
You're right, people might pass/gloss over the contents without a little preview. I just thought it looked better visually.

In `navigation-toc.html` there's `{% if include.collapsible and itemCount > 10 %}` which collapses the contents if there's more than 10 of them.

kwalrath:
@joshvocal can you think of a solution that looks better but still shows some of the contents?

kwalrath:
I'm going to close, but please feel free to suggest other possible fixes.

PR #2178, opened 2019-12-17
Weird hack to fix the build

kwalrath:
I have no idea why this fix works, but it does (locally, at least).


kwalrath:
I'm going to commit this so the build unbreaks, but... I'd love a review/consultation later.

Issue #1058, opened 2018-08-08
Consider generating the external link icon statically.


chalin:
- [ ] Generating the external link icon statically.
- [ ] Add `target="_blank" rel="noopener"`.
- [ ] Handle punctuation correctly.

parlough:
For now, this isn't a big enough deal. If we do switch tooling, we can consider it then.

PR #2210, opened 2020-01-11
fixes a typo

ryanmcginnis:
it seems like the semicolon should live outside of the Markdown link brackets.

### DIFF
--- a/src/codelabs/iterables.md
+++ b/src/codelabs/iterables.md
@@ -242,7 +242,7 @@ Your goal is to implement the predicate for `singleWhere()` that satisfies the f
 * The element contains the character `'a'`.
 * The element starts with the character `'M'`.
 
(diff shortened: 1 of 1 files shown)


kwalrath:
Actually... we should document this more clearly, but when it's an external link, we put the punctuation inside the link.

The reason is that it looks weird to have a gap between the link text and the punctuation, which is what happens when the external link symbol comes before the punctuation. We're not 100% consistent about implementing this — for example, the "Reporting vulnerabilities" paragraph in https://dart.dev/security has both placements of punctuation. Oops. But it does illustrate the issue:

image

It'd be better to fix the CSS so this wouldn't be necessary. I thought we had an issue open for this, but all I can find is #1058. I'll file an issue.

kwalrath (approved):
With #2679, we can now move punctuation to be outside the link!

Issue #2214, opened 2020-01-14
fix dev build


kwalrath:
The [dev build has broken](https://travis-ci.org/dart-lang/site-www/jobs/637095921#L1329) in a weird way. I'm going to allow that failure for now, in hopes that the next dev release will fix the problem. If it doesn't, I'll take a look. (Or someone else is welcome to take a look... just let me know.)

From the [raw build log](https://api.travis-ci.org/v3/job/637095921/log.txt):

```
Analyzing analysis_options.yaml, bin, lib, test...
  error â€¢ Undefined class 'extension'. â€¢ lib/effective_dart/style_good.dart:108:1 â€¢ undefined_class
  error â€¢ Functions must have an explicit list of parameters. â€¢ lib/effective_dart/style_good.dart:108:11 â€¢ missing_function_parameters
  error â€¢ Undefined class 'on'. â€¢ lib/effective_dart/style_good.dart:108:26 â€¢ undefined_class
  error â€¢ A function body must be provided. â€¢ lib/effective_dart/style_good.dart:108:26 â€¢ missing_function_body
  error â€¢ Functions must have an explicit list of parameters. â€¢ lib/effective_dart/style_good.dart:108:29 â€¢ missing_function_parameters
  error â€¢ Undefined class 'extension'. â€¢ lib/effective_dart/style_good.dart:110:1 â€¢ undefined_class
  error â€¢ Functions must have an explicit list of parameters. â€¢ lib/effective_dart/style_good.dart:110:11 â€¢ missing_function_parameters
  error â€¢ Undefined class 'on'. â€¢ lib/effective_dart/style_good.dart:110:28 â€¢ undefined_class
  error â€¢ A function body must be provided. â€¢ lib/effective_dart/style_good.dart:110:28 â€¢ missing_function_body
  error â€¢ Functions must have an explicit list of parameters. â€¢ lib/effective_dart/style_good.dart:110:31 â€¢ missing_function_parameters
  lint â€¢ Name non-constant identifiers using lowerCamelCase. â€¢ lib/effective_dart/style_good.dart:108:11 â€¢ non_constant_identifier_names
  lint â€¢ Name non-constant identifiers using lowerCamelCase. â€¢ lib/effective_dart/style_good.dart:108:29 â€¢ non_constant_identifier_names
  lint â€¢ Name non-constant identifiers using lowerCamelCase. â€¢ lib/effective_dart/style_good.dart:110:11 â€¢ non_constant_identifier_names
[...]

PR #2215, opened 2020-01-14
Temporarily allow a dev build failure

kwalrath:
Contributes to #2214


kwalrath:
Btw, when I undo this change, I'm going to leave the code in, but commented out. It took far too long to find the right code to use.

We might want to consider using `jobs` instead of `matrix`, since they appear to be equivalent, and all the Travis docs use `jobs`.

PR #2229, opened 2020-01-16
Change double quotes to single quotes

Haugen:
Suggestion to change the double quotes for the example code from `basic_writer_server.dart` to single quotes to comply with the `prefer_single_quotes` rule. This is on the documentation page for writing HTTP servers and clients. https://dart.dev/tutorials/server/httpserver

I also created a PR to apply the same changes in the actual examples files. https://github.com/dart-lang/dart-tutorials-samples/pull/57

### DIFF
--- a/examples/httpserver/bin/basic_writer_server.dart
+++ b/examples/httpserver/bin/basic_writer_server.dart
@@ -33,12 +33,12 @@ Future main() async {
       } catch (e) {
         response
           ..statusCode = HttpStatus.internalServerError
-          ..write("Exception during file I/O: $e.");
+          ..write('Exception during file I/O: $e.');
       }
     } else {
       response
         ..statusCode = HttpStatus.methodNotAllowed
-        ..write("Unsupported request: ${req.method}.");
+        ..write('Unsupported request: ${req.method}.');
     }
     await response.close();
   }
--- a/src/_tutorials/server/httpserver.md
+++ b/src/_tutorials/server/httpserver.md
@@ -671,12 +671,12 @@ Future main() async {
       } catch (e) {
         response
           ..statusCode = HttpStatus.internalServerError
-          ..write("Exception during file I/O: $e.");
+          ..write('Exception during file I/O: $e.');
       }
     } else {
       response
         ..statusCode = HttpStatus.methodNotAllowed
-        ..write("Unsupported request: ${req.method}.");
+        ..write('Unsupported request: ${req.method}.');
     }
     await response.close();
   }


Haugen:
I checked the failing Travis tasks, but am not sure exactly what they mean or what I need to adjust.

kwalrath:
Thank you! I believe the travis errors are because the repo file that this excerpt comes from ([examples/httpserver/bin/basic_writer_server.dart](https://github.com/dart-lang/site-www/blob/master/examples/httpserver/test/httpserver_test.dart)) needs to be updated to match the displayed code.

Do you know of an easy way to add that file to this PR? If not, are you willing to make an accompanying PR that fixes the example file?

Haugen:
I'd love to help out with this, but I'm still not sure exactly what needs to be done.

You linked to the file [httpserver_test.dart](https://github.com/dart-lang/site-www/blob/master/examples/httpserver/test/httpserver_test.dart). Can you elaborate a little bit on what you think might be the issue here?

kwalrath:
The short answer: I think you need to edit basic_writer_server.dart, changing the double quotes to single quotes.

The long answer:
We really need to improve the docs for this repo, but here's what they should tell you (and perhaps they do, though I can't find the right page at the moment).

In the file you edited, at the top of the code excerpt, there's this:
```
<?code-excerpt "httpserver/bin/basic_writer_server.dart" replace="/\/\*\d\*\/|contentType.*? == \S+|(content|data) = [^;]*|req\.uri[^ ]*/[!$&!]/g"?>
```

That means that the code excerpt is automatically updated from a file under the `examples` directory: [examples/httpserver/bin/basic_writer_server.dart](https://github.com/dart-lang/site-www/blob/master/examples/httpserver/test/httpserver_test.dart).

Often, included files have `// #docregion ... // #enddocregion` comments, indicating which lines should be included.  (E.g. see https://github.com/dart-lang/site-www/blob/master/examples/extension_methods/lib/fancylist.dart, which is included by https://dart.dev/guides/language/extension-methods.)

In this case, basic_writer_server.dart doesn't have docregion comments, so I believe that the `replace` regex (in the `code-excerpt` line above) finds the lines we need and maybe does some manipulation of those lines. (Obviously, I didn't write that code.)

Haugen:
Thank you for the answer Kathy! It makes sense now.

I've updated `examples/httpserver/bin/basic_writer_server.dart` to reflect the same changes I made in the documentation.

kwalrath:
Thanks!

PR #2232, opened 2020-01-17
Improve sizing of embedded DartPad

mit-mit:
Goals:

- Clicking Try Dart in the top-nav should bring the viewport to show just DartPad with no disturbing content around it
- The embedded DartPad should size itself to better use the space in larger browser windows

Staging link: https://mit-staging.firebaseapp.com/#try-dart


legalcodes:
LGTM

PR #2250, opened 2020-02-01
Fix the dev build

kwalrath:
The analyzer notices an unused field that it didn't used to


kwalrath:
Dev build is passing now, woohoo!

PR #2255, opened 2020-02-04
Cross-link to pub.dev policy

mit-mit:
* Clarify that 'publishing is forever'
* Cross-link to pub.dev policy

Note: Not ready for merging until the next pub.dev deploy


mit-mit:
The destination link is now live; publishing PR

PR #2256, opened 2020-02-04
Reorder example file priority order

mit-mit:


mit-mit:
The code will be changing; updating title to reflect this should not be merged yet.

mit-mit:
The pub.dev changes have landed; as an example [shared_preferences](https://pub.dev/packages/shared_preferences#-example-tab-) now has a nice code sample rather than before where it just linked to [an empty readme](https://github.com/flutter/plugins/tree/master/packages/shared_preferences/shared_preferences/example).

PR #2258, opened 2020-02-05
Tweak string example

### LINKED ISSUE #2257: var pi
var pi = 3.14;

so "var"  is variable? maybe const?

kwalrath:
Fixes #2257

Staged: https://kw-staging-dartlang-2.firebaseapp.com/#try-dart
(choose **Strings** from the drop down)


johnpryan:
It looks like  `tool/refresh-code-excerpts.sh` needs to be locally?

PR #2273, opened 2020-02-12
Fix a typo, add a link

kwalrath:
It seems like the **Debugging and diagnostics** section should refer to the diagnostics page. Plus I saw a typo.


kwalrath:
Good points. Maybe I'll just change the title of the section instead.

PR #2279, opened 2020-02-14
add rel="noopener" to _blank hrefs

### LINKED ISSUE #176: Fix hrefs that use target="_blank"
For performance and security reasons:
https://jakearchibald.com/2016/performance-benefits-of-rel-noopener/

We need a macro for this site and for webdev.

I found 86 instances on www.

kwalrath:
...or sometimes just delete the _blank. We aren't consistent in how we use _blank. :/ Maybe we should never use it?

Staged: https://kw-staging-dartlang-2.firebaseapp.com

Fixes #176.

### DIFF
--- a/src/404.html
+++ b/src/404.html
@@ -29,6 +29,7 @@ sitemap: false
     Click your browser's Back button or
     return to the <a href="/">homepage</a>. <br />
     If you think the URL should work, please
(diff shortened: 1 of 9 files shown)


johnpryan:
My understanding from reading Mathias' [original article](https://mathiasbynens.github.io/rel-noopener/) is that this is only for user-generated content?

kwalrath:
How can we get a definitive answer on this?

johnpryan:
Yes we got an answer, this is not just for user-generated content, so LGTM.

PR #2281, opened 2020-02-18
fix: remove reference to 404 page

jon-simpkins:
https://dart.academy/build-a-real-time-chat-web-app-with-dart-angular-2-and-firebase-3/ currently resolves to https://fail.ghost.org/, it looks like dart.academy is gone


mit-mit:
This LGTM but I defer to @kwalrath for sign-off

PR #2293, opened 2020-03-03
add google analytics ID to DartPad snippets

johnpryan:
Adds a unique ID to snippets independent of the Gist ID or whether
or not the sample uses the inject script

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -1290,7 +1290,7 @@ https://gist.github.com/d988cfce0a54c6853799
 {{site.dartpad}}/d988cfce0a54c6853799
 (The gist needs updating: see https://github.com/dart-lang/site-www/issues/189)
 <iframe
-src="{{site.dartpad-embed-inline}}?id=d988cfce0a54c6853799"
+src="{{site.dartpad-embed-inline}}?id=d988cfce0a54c6853799&ga_id=default_parameter_values"
     width="100%"
     height="450px"
     style="border: 1px solid #ccc;">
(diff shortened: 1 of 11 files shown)


kwalrath:
Hmmm... where is run-dartpad defined? I can't find it, and it doesn't seem to be working. E.g. https://dart.dev/tutorials/server/get-started#1-play-with-dart-code-in-dartpad says it uses an embedded DartPad, but it's just code.

kwalrath:
The change that introduced run-dartpad appears to be #2092. I don't know when it stopped working.

johnpryan:
https://github.com/dart-lang/site-www/commit/0f735bf62fba670c8b7a17fadb1371cc07530f79 should fix the issue

johnpryan:
Good eye, fixed.

I think the language tour needs updating too...

johnpryan:
@kwalrath there were a few more places to update but this should be ready for another review.

src/_guides/language/language-tour.md
```diff
@@ -4125,7 +4125,7 @@ main() => print(out);
 {% endcomment %}
 
 <iframe
-src="{{site.dartpad-embed-inline}}?id=405379bacf30335f3aed"
+src="{{site.dartpad-embed-inline}}?id=405379bacf30335f3aed&ga_id=callable_classes"
```
kwalrath:
```suggestion
src="{{site.dartpad-embed-inline}}?id=3723fcf3915ca935d13393b8a9f86fd5&ga_id=callable_classes"
```

kwalrath:
Should we have a list of the ga_ids somewhere so we don't accidentally duplicate any and it'll be easy to look up which ids correspond to which pages?

If so, how can we make this list easy to find and keep up-to-date?

johnpryan:
@kwalrath I would prefer to keep one source of truth, but I agree that this approach has an extra maintenance cost. If we're not interested in analytics on certain pages, we can leave it out.

kwalrath:
I think we always want the ability to have analytics. I suppose we can search for ga_id and make sure the IDs are unique before we add new ones. (I assume you were careful about not duplicating the IDs?)

johnpryan:
Yes the `ga_id` tag needs to be unique for that page. (It's okay to have the same id if they are on different pages)

kwalrath:
Do you want to add the final ga_id to src/_tutorials/language/streams.md before we merge this?

johnpryan:
I think streams.md has two iframes with two ga_id tags now.

kwalrath:
Looks good. Merged!

PR #2304, opened 2020-03-13
Add localhost troubleshooting tip for pub

### LINKED ISSUE #1822: 'Troubleshooting pub' explain workaround for "localhost not working in browser"
Page URL: https://dart.dev/tools/pub/troubleshoot
Page source: https://github.com/dart-lang/site-www/tree/master/src/tools/pub/troubleshoot.md

Description of issue:

If working on Chromebook, working over SSH, or in a container, it can be the case that the `localhost` where `pub` is running is different from the `lcoalhost` in the browser the user is using.
This makes it impossible to login. The user will be able open the sign-in link and sign-in, but they will then be redirected to `http://localhost:<port>?code=...` which won't work. The browser will say something like:
> "This site can’t be reached localhost refused to connect."

In this case, the user can copy the URL and do `curl http://localhost:<port>?code=...` in a terminal next to the pending `pub publish` command.

I've seen this issue twice now: https://github.com/dart-lang/pub/issues/2187

So maybe we should add an entry to the troubleshooting section. Something along the lines of:

------

## Localhost unreachable after sign-in
If running `pub publish` in container or over an SSH session it might be that the `localhost` which `pub` is listening to is different from the `localhost` accessible in your browser. If this is the case, it is possible to sign-in, after which the browser will complain that `http://localhost:<port>?code=...` is not reachable.

A possible workaround for this is to copy the URL from the browser into a terminal session next to the `pub publish` awaiting for sign-in and run `curl http://localhost:<port>?code=...`.

------

Does this make any sense, I doubt this is super frequent, if it turns out to be super common, then perhaps we should revise the sign-in flow and avoid listening on localhost. Or offer an alternative sign-in flow.

kwalrath:
Fixes #1822 

Staged:
https://kw-www-dartlang-1.firebaseapp.com/tools/pub/troubleshoot#localhost-unreachable-after-sign-in

I tried to make it as clear as possible, but since I haven't done this myself, I might have gotten something wrong.

@aardappel could you check the wording? Thanks!

### DIFF
--- a/src/tools/pub/troubleshoot.md
+++ b/src/tools/pub/troubleshoot.md
@@ -116,9 +116,10 @@ Try this workaround, which uses the command line to complete sign-in:
    (`http://localhost:<port>?code=...`)
    but complains that the URL isn't reachable.
 3. Copy the _new localhost URL_  from the browser.
-4. In another terminal window, use the `curl` command to
+4. In another terminal window in the same container or on the same host
+   as the one where `pub publish` was called, use the `curl` command to
    complete sign-in using the _new localhost URL_:
 
    ```terminal
-$ curl http://localhost:<port>?code=...
+$ curl 'http://localhost:<port>?code=...'
 ```
--- a/src/tools/pub/troubleshoot.md
+++ b/src/tools/pub/troubleshoot.md
@@ -99,4 +99,26 @@ On Windows PowerShell:
 ```terminal
 $ $Env:https_proxy="username:password@hostname:port"
 ```
+## Localhost unreachable after sign-in
 
+When you run `pub publish` in a container or over an SSH session,
+the `localhost` that `pub` is listening to might be different from
+the `localhost` that's accessible in your browser.
+Although you can sign in using the browser,
+the browser then complains that `http://localhost:<port>?code=...`
+is not reachable.
+
+Try this workaround, which uses the command line to complete sign-in:
+
+1. In a terminal window, run `pub publish`.
+2. In the browser window that comes up, sign in. <br>
+   The browser is redirected to a _new localhost URL_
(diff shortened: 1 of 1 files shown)


aardappel:
LGTM

src/tools/pub/troubleshoot.md
```diff
@@ -99,4 +99,26 @@ On Windows PowerShell:
...
+
+Try this workaround, which uses the command line to complete sign-in:
+
+1. In a terminal window, run `pub publish`.
+2. In the browser window that comes up, sign in. <br>
+   The browser is redirected to a _new localhost URL_
+   (`http://localhost:<port>?code=...`)
+   but complains that the URL isn't reachable.
+3. Copy the _new localhost URL_  from the browser.
+4. In another terminal window, use the `curl` command to
```
jonasfj: Another terminal window in the same container or on the same host as `pub publish` was called.

I'm not sure it is necessary to say this... It ought to be obvious. Just noting it.

Probably it's obvious that it has to be on the same host/container as `pub publish`.

src/tools/pub/troubleshoot.md
```diff
@@ -99,4 +99,26 @@ On Windows PowerShell:
...
+   complete sign-in using the _new localhost URL_:
+
+   ```terminal
+$ curl http://localhost:<port>?code=...
```
jonasfj: Put single quotes around the URL...

Otherwise, bash is going to interpret `&` as meaning the command should run in the background.

(The URL may contain `&` if there is multiple query string values)

kwalrath:
Thanks for the reviews! Sorry for the delay in getting this content up.

Issue #2268, opened 2020-02-10
Remove extra space at top of cmd-line example


kwalrath:
Followup to #2266; see https://github.com/dart-lang/site-www/pull/2266#pullrequestreview-355494669 for details.

Nishikoh:
hi , @kwalrath .
If I will work on this issue, should I fix like next images?
  
**current space**
![image](https://user-images.githubusercontent.com/55723667/77228082-25cc5200-6bc8-11ea-9c80-bf1196629ccd.png)

**fix space**
![image](https://user-images.githubusercontent.com/55723667/77228293-a8094600-6bc9-11ea-8fa1-c7eced31d0cc.png)

kwalrath:
Actually, no. This bug is about a single sample, visible at https://dart.dev/tutorials/server/cmdline#overview-of-the-dcat-app-code. It seems to have an extra blank line at the top (above `import 'dart:convert';`), and I don't know why.

Nishikoh:
@kwalrath 
I understand.
I'll try to fix this issue.

micpap25:
Looks like this issue was fixed with this merged, we can close this.

kwalrath:
True. Thanks, @micpap25!

PR #2315, opened 2020-03-25
Remove extra space at top of cmd-line example 

Nishikoh:
Remove extra space at top of cmd-line example (#2268).

**before image**
スクリーンショット 2020-03-25 7 49 44

**after fix image**
スクリーンショット 2020-03-25 7 50 12


kwalrath:
Hmmm... that fix does work, although the comment above ("Do NOT EDIT this file directly. It is autogenerated by tool/create_code_with_tooltips.dart") gives me pause.

Still, we haven't autogenerated that file in forever, so for now I think I'll just file a bug about that and accept your fix. Thank you, @Nishikoh!

kwalrath:
Fixes #2268

PR #2328, opened 2020-04-02
fixing equality operator typo

mshwf:


mshwf:
@googlebot I signed it!
________________________________

kwalrath:
Actually, that `.` was intentional, as a way of emphasizing that `==` (like all operators) is actually a method invoked on the first operand.

However, since that code doesn't actually work (and I don't know if it ever did), maybe we should put it another way.

@munificent do you have any suggestions? The relevant text is in https://dart.dev/guides/language/language-tour#equality-and-relational-operators.

mshwf:
No way to think that, especially it's in a code styled. It's frustrating more than clarifying. However, it actually helps seeing this operator from another angle. May be put that in a note or something.

munificent:
Yeah, I agree with Kathy, the syntax here is deliberate and I think it does clarify things for most readers.

kwalrath:
I'm going to close this PR but open an issue describing the problem with the text, as it's written. Thank you, @mshwf, for bringing this to our attention!

PR #2332, opened 2020-04-07
Document pub outdated

### LINKED ISSUE #2290: Document `pub outdated` at https://dart.dev/tools/pub/cmd/pub-outdated
New command being developed here: https://github.com/dart-lang/pub/pull/2315.

This should be done before next stable sdk release

kwalrath:
Early draft is staged here:

https://kw-staging-dartlang-2.firebaseapp.com/tools/pub/cmd/pub-outdated

Am I on the right track, @sigurdm? Could you (or someone else) answer my questions in red? Thank you!

Fixes #2290


src/tools/pub/cmd/pub-outdated.md
```diff
@@ -0,0 +1,251 @@
...
+
+_Outdated_ is one of the commands of the [pub tool](/tools/pub/cmd).
+
+{% prettify nocode tag=pre+code %}
+$ pub outdated [options]
+{% endprettify %}
+
+Use `pub outdated` to identify out-of-date [package dependencies][]
+and get advice on how to update them.
+<span style="color:red">[PENDING: Should we motivate why you want to do this?
```
mit-mit, line 15: Yes, using latest stable versions is good guidance to ensure you have the latest bug fixes, and other improvements.

src/tools/pub/cmd/pub-outdated.md
```diff
@@ -0,0 +1,251 @@
...
+Null safety might be the biggest near-term motivation,
+but are there other major reasons
+that make it worth the effort to update and test your code?]</span>
+
+[package dependencies]: /tools/pub/dependencies
+
+Here's an overview of how you can use `pub outdated`
+to help you update package dependencies:
+
+1. In the top directory of your app or package,
```
mit-mit, line 25: I'd recommend running outdated before upgrade to see what the status is (similar to the blog post)

src/tools/pub/cmd/pub-outdated.md
```diff
@@ -0,0 +1,251 @@
...
+
+
+## Example
+
+You run `pub outdated` in the top directory of your package or app,
+in the same directory as your package or app's
+[`pubspec.yaml`](/tools/pub/pubspec) file.
+Here's an example of running `pub outdated`:
+
+```terminal
```
mit-mit, line 43: I think somewhere we should document what each of the four columns show.

src/tools/pub/cmd/pub-outdated.md
```diff
@@ -0,0 +1,251 @@
...
+$
+```
+<span style="color:red">[PENDING: file bug about "up-to-date" (shouldn't be hyphenated unless it comes before a noun)]
+
+{{site.alert.important}}
+  Test your code to verify that it still works as expected
+  after updating the packages.
+{{site.alert.end}}
+
+If any transitive dependencies are out of date,
```
mit-mit, line 118: I've found it easier to simply edit pubspec yaml to move the dependency that isn't resolvable, and then looking at the error output from pub get.

src/tools/pub/cmd/pub-outdated.md
```diff
@@ -0,0 +1,251 @@
...
+```
+
+As a rule, update the **dependencies** and **dev_dependencies** sections
+of your `pubspec.yaml` file
+to depend on the **resolvable** versions of out-of-date packages.
+<span style="color:red">
+  [PENDING: correct? explain why?]
+</span>
+
+Here are the meanings of the version numbers in the output:
```
mit-mit, line 170: Oh, here is what I had in mind. I think I would move this up before the first example.

src/tools/pub/cmd/pub-outdated.md
```diff
@@ -0,0 +1,251 @@
...
+## Options
+
+For options that apply to all pub commands, see
+[Global options](/tools/pub/cmd#global-options).
+
+`--json`
+: Use this option to generate output in JSON format.
+
+`--[no-]color`
+: Use this option to change whether the output has multiple colors.
```
mit-mit, line 226: Suggest s/has multiple colors/has color/, as white isn't a color, at least in physics ;-)

src/tools/pub/cmd/pub-outdated.md
kwalrath: I started out doing that, but it seemed a little silly to make them run outdated first, when they were almost certainly going to have to run upgrade.

mit-mit: Could just be me, but I find it helpful to see the inventory before making any changes to current versions. Helps understand the total set of packages that need updating, and which you should be on the lookout for new issues in.

src/tools/pub/cmd/pub-outdated.md
kwalrath: Unless you're talking light, as in computer screens. :)

sigurdm: Maybe 'is colorized'?

src/tools/pub/cmd/pub-outdated.md
```diff
@@ -0,0 +1,251 @@
...
+-  build_web_compilers: ^1.0.0
++  build_web_compilers: ^2.10.0
+```
+
+{{site.alert.tip}}
+  To see what changed in the new version, you can look at the package's changelog,
+  which you can find in the package page on
+  [pub.dev.]({{site.pub}})
+  For an example, see the
+  [build_web_compilers changelog.]({{site.pub-pkg}}/build_web_compilers#-changelog-tab-)
```
sigurdm, line 90: I think the link recently changed to: https://pub.dev/packages/build_web_compilers/changelog

src/tools/pub/cmd/pub-outdated.md
```diff
@@ -0,0 +1,251 @@
...
+transitive dependencies
+_fe_analyzer_shared  -         -           2.0.0       2.0.0     
+analyzer             0.35.4    0.35.4      0.39.5      0.39.5    
+...
+
+9  dependencies are constrained to versions that are older than a resolvable version.
+To update these dependencies, edit pubspec.yaml.
+```
+
+<span style="color:red">[PENDING: Is there a better example we should use?]</span>
```
sigurdm, line 60: Perhaps the example from the blog post.
I believe it has a bit more going on between the columns.

src/tools/pub/cmd/pub-outdated.md
kwalrath: Oh, that's much a much nicer URL. Although clicking on the in-page link still takes me to the ugly URL.

@jonasfj can you confirm which URLs I should be using? Will the Installing tab URLs be updated soon?

Once I know what URLs to use, I should do a global search for package tab references.

src/tools/pub/cmd/pub-outdated.md
kwalrath: I think I wanted a non-Flutter example (so as to keep dart.dev platform-agnostic), so I picked a random one from site-www. But a more interesting example is probably better.

/cc @mit-mit

mit-mit: Maybe an alternate option is to construct your own example; linking to something existing isn't very durable, as it will likely have it's versions rolled and then no longer exhibit the behaviour described in the docs?

src/tools/pub/cmd/pub-outdated.md
mit-mit: paging @jonasfj

src/tools/pub/cmd/pub-outdated.md
kwalrath: Transitive dependencies aren't in pubspec.yaml... Do you mean remove that dependency from pubspec.lock?

src/tools/pub/cmd/pub-outdated.md
kwalrath: Maybe "uses color for emphasis"?

kwalrath:
Still a WIP, but please take a look at the changes to the Package dependencies page:

https://kw-staging-dartlang-2.firebaseapp.com/tools/pub/dependencies
https://kw-staging-dartlang-2.firebaseapp.com/tools/pub/dependencies#best-practices

kwalrath:
I still need to update the example, and I'd like answers to the remaining questions in red. But... maybe this is otherwise presentable?

https://kw-staging-dartlang-2.firebaseapp.com/tools/pub/cmd/pub-outdated
https://kw-staging-dartlang-2.firebaseapp.com/tools/pub/dependencies
https://kw-staging-dartlang-2.firebaseapp.com/tools/pub/dependencies#best-practices

src/tools/pub/cmd/pub-outdated.md
```diff
@@ -0,0 +1,248 @@
...
+  the value is `-`.
+
+Upgradable
+: The latest version allowed by your `pubspec.yaml` file.
+  This is the version that `pub upgrade` resolves to.
+  The value is `-` if the value in the **Current** column is `-`.
+
+  <span style="color:red">
+    [PENDING: Is the description of `-` correct?
+    Is there a more helpful way to state it?]
```
jonasfj, line 174: This is correct.

`Upgradable` is the version you'll get if you run `pub upgrade`.

Another way to say it involves `pubspec.lock`, one could say something like: this is the version you get with `pub get` if you delete the version locking in `pubspec.lock`.
(I'm not sure I would phrase it that way).

**Context**: `pub upgrade` is the same as running `rm -f pubspec.lock && pub get`.

src/tools/pub/cmd/pub-outdated.md
```diff
@@ -0,0 +1,248 @@
...
+  and then run `pub upgrade`.
+  A value of `-` means that the package will no longer be needed.
+
+  <span style="color:red">
+    [PENDING: Correct?]
+  </span>
+
+{{site.alert.important}}
+  We don't recommend using `any` as the constraint,
+  for performance reasons.
```
jonasfj, line 190: I think we discourage `any` because `pub upgrade` will then install new possibly incompatible major versions of your dependencies, and this is often undesirable. Instead it is often desirable to require manual editing of the `pubspec.yaml` to get new possibly incompatible major versions -- as such new versions often require code changes too.

The "performance reasons" argument is not something I've seen substantiated. A SAT solver may take longer if there are more solutions in the solution space, but I wouldn't worry about this. That said I have no metrics to back this up in either direction :D

src/tools/pub/cmd/pub-outdated.md
```diff
@@ -0,0 +1,248 @@
...
+
+Resolvable
+: The latest version that can be resolved.
+  This version corresponds to what you get if
+  you set all version constraints in `pubspec.yaml` to `any`
+  and then run `pub upgrade`.
+  A value of `-` means that the package will no longer be needed.
+
+  <span style="color:red">
+    [PENDING: Correct?]
```
jonasfj, line 185: Correct :1st_place_medal:

src/tools/pub/cmd/pub-outdated.md
```diff
@@ -0,0 +1,248 @@
...
+    [PENDING: Is the description of `-` correct?
+    Is there a more helpful way to state it?]
+  </span>
+
+Resolvable
+: The latest version that can be resolved.
+  This version corresponds to what you get if
+  you set all version constraints in `pubspec.yaml` to `any`
+  and then run `pub upgrade`.
+  A value of `-` means that the package will no longer be needed.
```
jonasfj, line 182: I'm not sure how we want to phrase this. This is not a bad explanation, but we don't explain why we're showing this result.

**reasoning**: We compute the `latest resolvable version` because this is the latest version **you can get** in combination with the other dependencies you have.

So while you may not be able to get the latest versions of all packages, you can take the version number from `resolvable` column and insert that in `pubspec.yaml` with a `^` in-front.. and if you do that for all versions in the resolvable column then version resolution will work. So in cases where you can't get the latest version, _resolvable_ helps you discover what the latest version _you can get_ is.

src/tools/pub/dependencies.md
```diff
@@ -235,32 +223,62 @@ Instead, the typical workflow is:
...
+
+A version constraint that uses _traditional syntax_
+is a series of the following:
 
 `any`
 : The string `any` allows any version. This is equivalent to an empty
-  version constraint, but is more explicit. **While `any` is allowed,
-  we do not recommend it for performance reasons.**
+  version constraint, but is more explicit. **Although `any` is allowed,
+  we don't recommend it for performance reasons.**
```
jonasfj, line 281: Unrelated, and let's not open the can of worms here. But I don't think "performances reasons" is the argument we should be using :)

src/tools/pub/dependencies.md
```diff
@@ -397,23 +388,78 @@ dependency_overrides:
...
+  If you modify the pubspec and update to a new major version,
+  then you might encounter breaking changes,
+  so you need to test even more thoroughly.
+
+Don't check in the `pubspec.lock` file for library packages
+: Following the guidance in [What not to commit][],
+  don't check in `pubspec.lock` for library packages.
+  For app packages, you can commit the lock file if
+  it helps you track the dependencies that are used for
+  production or across a team.
```
jonasfj, line 442: I'm not sure we should make this recommendation a _best-practice_.

Honestly, this is one of the more controversial recommendations. Some people would always want to check-in `pubspec.lock` for applications (we do that) -- otherwise dependencies may change between QA and production. Some people may even want to check-in `pubspec.lock` for libraries to get a consistent development environment.

It depends a bit on workflow, hence, I don't think it's a _best practice_.

src/tools/pub/cmd/pub-outdated.md
kwalrath: I think it's important to have the example up high in the page, but I've added a link to this section.

src/tools/pub/cmd/pub-outdated.md
jonasfj: Yes, the URLs will change, but not in time for Dart 2.8, so I would suggest using the old URLs for now.

When the new URLs do launch I believe we will have javascript code redirecting the old ugly URLs such that these still work. We will want to update `dart.dev` docs when this happens to get a nicer experience. But it's not like `dart.dev` links are going to be broken all of a sudden.

src/tools/pub/cmd/pub-outdated.md
kwalrath: @mit-mit what do you think we should say? I've heard the performance reason, but I don't actually have evidence for it.

jonasfj: Maybe we could say:
> We discourage use of `any` as `pub upgrade` will install new possibly incompatible major versions.

src/tools/pub/dependencies.md
kwalrath: @mit-mit, @munificent, @kevmoo what do you think?

This section is completely new, and I kind of threw everything we've said about dependency management here... but we might want to start with less? Or to have a followup paragraph about practices that are more nuanced.

kevmoo: For libraries, I think it's a 'should not', – not a 'must not'.

For libs, if you check it in, any CI environment your in will honor it, but it won't be honored for your users. I think the guidance to not check in is a great idea.

jonasfj: @kevmoo, the guidance to not check in `pubspec.yaml` for libraries is reasonable. IMO a matter of personal preference: "stable CI" vs. "More relevant tests".

But for applications I think the guidance ought to strongly encourage checking `pubspec.lock` into git. Otherwise, you'll have discrepancy between devs, CI, QA and production apps.

If you don't check `pubspec.yaml` into git, you could easily ship untested code in an app store. In case a new version of a dependency is published between the time you do QA and the time you build+publish your app.

src/tools/pub/cmd/pub-outdated.md
mit-mit: I don't think you can ever have any outdated transitive dependencies unless you either have outdated main dependencies, dev_dependencies, or both.

To figure out why I can't get to the most recent main or dev dep (listed in Latest), I edit pubspec.yaml to pin to that exact version, run pub get, and look at the resolution error.

src/tools/pub/cmd/pub-outdated.md
```diff
@@ -0,0 +1,236 @@
...
+1. **Run `pub outdated`** to confirm that you've updated all the dependencies.
+1. **Test** your package to confirm that it still works as expected.
+
+
+## Example
+
+Here's an example of running `pub outdated` on
+[PENDING: link to source code]:
+
+```terminal
```
mit-mit, line 50: I suggest you create an example that has one or more outdated main dependencies (that's the main issue wrt. shipping bugs etc.).

src/tools/pub/cmd/pub-outdated.md
```diff
@@ -0,0 +1,236 @@
...
+Link to it so it's reproduceable.
+Update the text to match.]</span>
+
+The preceding output shows that all regular dependencies are up-to-date,
+but one [dev dependency][] (`build_web_compilers`) is out-of-date.
+Several [transitive dependencies][] are also out-of-date.
+For details on what each column shows, see the
+[output columns](#output-columns) section of this page.
+
+To fix the dev dependency,
```
mit-mit, line 77: I think it might be good if the example shows the three main cases:

  1. A dependency is outdated, but can be moved to latest by running `pub upgrade`.
  1. A dependency is outdated, but pubspec.yaml doesn't allow the latest version, so an edit to pubspec is needed
  1. A dependency is outdated, but the app is on the latest one that's mutually compatible

I personally find it easiest to explain this via the columns (hence why I lead with those in the blog post), which correspond to the three cases above like this:

First, the colors in the output are important here. They show the user where to start (the left-most non-red column)

  1. Current version is red, upgradeable is bold white, upgradeable/resolveable/latest show the same number: outdated but can be upgraded automatically; the dependency needs a `pub upgrade`
  1. Upgradeable is red, resolvable/latest show the same number: outdated and not automatically upgradeable; edit pubspec.yaml
  1. Resolvable is red: The app is already locked on the latest version that's mutually compatible. Nothing the developer can do (expect to file bugs on the packages that cause the mutual exclusion)

src/tools/pub/dependencies.md
```diff
@@ -397,23 +388,71 @@ dependency_overrides:
...
+  Also, it places an upper limit on the allowed version,
+  based on an assumption that packages use [semantic versions][],
+  where any version of path versioned `1.x` is compatible,
+  but where a new version `2.x` would be a major upgrade
+  that isn't semantically compatible with `1.x` versions. 
+
+Depend on the latest stable package versions
+: Use [`pub upgrade`][] to update to the latest package versions
+  that your pubspec allows.
+  To identify dependencies in your pubspec that
```
mit-mit, line 425: This is not entirely accurate as `outdated` looks at both `pubspec.yaml` and `pubspec.lock`.

Maybe say something like "To identify dependencies in your app or package that are not on the latest possible versions, use pub outdated."

mit-mit:
This is shaping up very nicely!

src/tools/pub/cmd/pub-outdated.md
kwalrath: Couldn't you have a regular or dev dependency that, in its latest version, depends on something that's out-of-date?

src/tools/pub/cmd/pub-outdated.md
```diff
@@ -0,0 +1,236 @@
...
+`--[no-]prereleases`
+: Use `--prereleases` to include prereleases when determining
+  the latest package versions.
+  By default, prerelease versions aren't considered.
+
+`--[no-]dev-dependencies`
+: Use `--no-dev-dependencies` to ignore [dev dependencies][dev dependency].
+
+`--[no-]dependency-overrides`
+: Use `--no-dependency-overrides` to ignore package dependencies
```
sigurdm, line 219: This is slightly off. The packages are not ignored, only the dependency_overrides on them are ignored.

Perhaps better would be:

`Use --no-dependency-overrides` to ignore [`dependency_overrides`][] when resolving package constraints.

src/tools/pub/cmd/pub-outdated.md
jonasfj: > I don't think you can ever have any outdated transitive dependencies unless you either have outdated main dependencies, dev_dependencies, or both.

<details>
<summary>You absolutely can :)</summary>
If you depend on `foo` and `foo` depends on `retry` version `^2.0.0`, then you won't get `retry` version `3.0.0` (despite the fact that it's only a transitive dependency).
</details>

kwalrath: So... is `pub deps` the best way to figure out which package's [direct or transitive] dependencies are causing the problem?

src/tools/pub/cmd/pub-outdated.md
kwalrath: OK, sounds like I should feature a screenshot so the colors are visible.

kwalrath: I... didn't know how to do the 3rd one. You can see the code I used at https://github.com/kwalrath/po-example.

mit-mit: See https://github.com/kwalrath/po-example/pull/1

src/tools/pub/cmd/pub-outdated.md
mit-mit: I think we have resolved this?

src/tools/pub/cmd/pub-outdated.md
mit-mit: Yes, `pub deps -s list` is good for this. Take the example I created in https://github.com/kwalrath/po-example/pull/1. After fixing the first two issues (args to 1.6.0, http to 0.12.0) what remains is path stuck on 1.6.2 with 1.7.0 being the latest. Here's is what the output of deps is:

```
dependencies:
- args 1.6.0
- http 0.12.1
  - http_parser >=0.0.1 <4.0.0
  - path >=0.9.0 <2.0.0
  - pedantic ^1.0.0
- path 1.6.2
- terminal_tools 0.0.0
  - path 1.6.2
  - meta 1.1.6

dev dependencies:
[...]

transitive dependencies:
[...]
```

From this we can see path is allowed to be 1.6.2 because terminal_tools allows just that single version, and similar for meta.

Another option is to update `pubspec.yaml` to temporarily have that dep directly:

```
dependencies:
  args: ^1.4.4
  http: ^0.12.0
  # Changed this to 1.7.0 to trigger error:
  path: ^1.7.0
  terminal_tools:
    path: ./pkg/terminal_tools/

```

The resolution will tell you what's going on:

```
mit-macbookpro4:po-example mit$ ~/sdk/dart-beta/dart-sdk/bin/pub get
Resolving dependencies...
Because every version of terminal_tools from path depends on path 1.6.2 and po_example depends on path ^1.7.0,
  terminal_tools from path is forbidden.
So, because po_example depends on terminal_tools from path, version solving failed.
```

For the transitive that's a bit harder; you need to temporarily add the transitive dep as a direct:
```
dependencies:
  args: ^1.4.4
  http: ^0.12.0
  path: ^1.5.0
  terminal_tools:
    path: ./pkg/terminal_tools/
  # begin temp dep to trigger error
  meta: 1.1.8
  # end temp dep
```

The resolution will tell you what's going on:

```
$ ~/sdk/dart-beta/dart-sdk/bin/pub get
Resolving dependencies...
Because po_example depends on terminal_tools from path which depends on meta 1.1.6, meta 1.1.6 is required.
So, because po_example depends on meta 1.1.8, version solving failed.
```

src/tools/pub/cmd/pub-outdated.md
```diff
@@ -0,0 +1,240 @@
...
+## Example
+
+Here's an example of running `pub outdated` on
+an example that depends on old versions of two packages:
+`args` and `pedantic`.
+As the following screenshot shows,
+`pub outdated` colorizes the output by default
+when you run it on the command line.
+
+![screenshot of pub outdated output; visible as text later in "Output columns" section](images/pub-outdated.png)
```
mit-mit, line 59: I find the colors hard to see in this screenshot with the green background. I think it's easier with a black background.

src/tools/pub/cmd/pub-outdated.md
```diff
@@ -0,0 +1,240 @@
...
+   the directory that contains your package's
+  [`pubspec.yaml`](/tools/pub/pubspec) file.
+1. **Run `pub outdated`**
+   to identify which package dependencies are out-of-date.
+   Note the affected packages,
+   so that later you can test the behavior of code that uses them.
+1. Follow the recommendations of `pub outdated` for updating the packages.
+   Some updates might require only running `pub upgrade`.
+   Others might require updating `pubspec.yaml` before running `pub upgrade`.
+1. **Run `pub outdated`** to confirm that you've updated all the dependencies.
```
mit-mit, line 41: Sometimes you cannot update all dependencies. So maybe change to "to confirm that you are using the latest compatible package versions" ?

src/tools/pub/cmd/pub-outdated.md
```diff
@@ -0,0 +1,240 @@
...
+when you run it on the command line.
+
+![screenshot of pub outdated output; visible as text later in "Output columns" section](images/pub-outdated.png)
+
+The preceding output shows that one regular dependency (`args`)
+and one [dev dependency][] (`pedantic`) are out-of-date.
+For details on what each column shows, see the
+[output columns](#output-columns) section of this page.
+
+As a rule, you update the **dependencies** and **dev_dependencies** sections
```
mit-mit, line 66: "As a rule" seems strong? "As a general guideline"?

src/tools/pub/cmd/pub-outdated.md
```diff
@@ -0,0 +1,240 @@
...
+
+{{site.alert.tip}}
+  To see what changed in the new version of a package
+  that's published on [pub.dev,]({{site.pub}})
+  look at the changelog in the package page.
+  For example, you can look at the **Changelog** tabs on the pages for the
+  [`args`][] and [`pedantic`][] packages.
+{{site.alert.end}}
+
+To fix the first dependency (`args`),
```
mit-mit, line 80: Maybe expand to "To fix the first dependency (`args`), which is listed as upgradeable," ?

src/tools/pub/cmd/pub-outdated.md
```diff
@@ -0,0 +1,240 @@
...
+
+```terminal
+$ pub upgrade
+Resolving dependencies... 
+> args 1.6.0 (was 1.5.0)
+  pedantic 1.8.0 (1.9.0 available)
+Changed 1 dependency!
+```
+
+To fix the second dependency (`pedantic`),
```
mit-mit, line 91: Maybe expand to "To fix the second dependency (`pedantic`), which is listed as resolveable," ?

src/tools/pub/cmd/pub-outdated.md
```diff
@@ -0,0 +1,240 @@
...
+  after updating the packages.
+{{site.alert.end}}
+
+
+## Output columns
+
+The output of `pub outdated` has four columns of version information
+for each out-of-date dependency:
+
+```terminal
```
mit-mit, line 136: Here would prefer to show all three cases incl. outdated transitive dependencies (I like that they were omitted above in the example).

src/tools/pub/cmd/pub-outdated.md
```diff
@@ -0,0 +1,240 @@
...
+Upgradable
+: The latest version allowed by your `pubspec.yaml` file.
+  This is the version that `pub upgrade` resolves to.
+  The value is `-` if the value in the **Current** column is `-`.
+
+Resolvable
+: The latest version that can be resolved,
+  when combined with all other dependencies.
+  This version corresponds to what you get if
+  you set all version constraints in `pubspec.yaml` to `any`
```
mit-mit, line 171: "you set" sounds a little too much like a recommendation for something to do. Maybe phrase like "corresponds to what you get if you had an `any` version constraints in `pubspec.yaml`  on all dependencies" ?

src/tools/pub/cmd/pub-outdated.md
```diff
@@ -0,0 +1,240 @@
...
+  because then running `pub upgrade` might surprise you by making
+  incompatible version changes.
+{{site.alert.end}}
+
+Latest
+: The latest version of the package available,
+  excluding prereleases unless you use the option `--prereleases`.
+
+For example, say your app depends on the `foo` and `bar` packages,
+but the latest version of `bar` depends on an old major version of `foo`.
```
mit-mit, line 186: Maybe change "latest version of `bar` depends on an old major version of `foo" to something like "latest version of `bar` allows only older major versions of `foo" ?

src/tools/pub/cmd/pub-outdated.md
mit-mit, line 136: Consider using a screenshot so we get the colors again

src/tools/pub/cmd/pub-outdated.md
kwalrath: I tried a black background but couldn't see it due to https://github.com/dart-lang/pub/issues/2473.

If you want to take screenshots, I could use yours...

src/tools/pub/cmd/pub-outdated.md
kwalrath: The problem with a screenshot is that it's very bad for accessibility and search.

src/tools/pub/cmd/pub-outdated.md
kwalrath: I've changed this to:

When you edit the pubspec.yaml file, you generally update the dependencies and dev_dependencies sections so that each package uses the versions in the Resolvable column.

src/tools/pub/cmd/pub-outdated.md
kwalrath: That means changing the example, right? If so, how would you suggest we do that?

src/tools/pub/cmd/pub-outdated.md
kwalrath: That seems a little harder to read. Is it important to talk about multiple major versions?

kwalrath:
ptal

src/tools/pub/cmd/pub-outdated.md
mit-mit: Oh, that's sad. I'm using this Terminal, that has the colors as expected: https://www.iterm2.com/. Do you want to try that one, or should I grab some for you?

src/tools/pub/cmd/pub-outdated.md
mit-mit: That's a good point. Can we get colors (specifically the red highlight) in markdown?

jonasfj: Use:
```bash
sudo apt-get install aha
TERM=xterm-256color script -qefc 'dart bin/pub.dart outdated' | aha -n
```

Then wrap the output with `<pre></pre>`. See `man aha` if you want colors for a black background, and other tweaks.

The `TERM=xterm-256color script -qefc 'dart ...'` wrapping just helps to trick dart:io into thinking it's running in a terminal with ANSI color support.

kwalrath: Michael, can you do this? I don't have access to a Linux machine. (I was able to use brew to install aha, but the script command is different on Mac — doesn't let you specify a command, e.g. — and it'd take quite a while to figure this out.)

mit-mit: https://gist.github.com/mit-mit/a20b65967d8cd0c6c33f93a8b0363a01

src/tools/pub/cmd/pub-outdated.md
mit-mit: Screenshot here with black if you need one:

https://user-images.githubusercontent.com/13644170/81102585-24ed4680-8f10-11ea-8d9f-f08466892c87.png

Screenshot 2020-05-05 at 20 36 39

src/tools/pub/cmd/pub-outdated.md
```diff
@@ -0,0 +1,284 @@
...
+  The value is `-` if the value in the **Current** column is `-`.
+
+Resolvable
+: The latest version that can be resolved,
+  when combined with all other dependencies.
+  This version corresponds to what `pub upgrade` would give you
+  if all version constraints in `pubspec.yaml` were `any`.
+  A value of `-` means that the package won't be needed.
+
+{{site.alert.important}}
```
mit-mit, line 215: I personally find the big yellow box here in the middle of the column explanations a bit confusing (I thought for a sec. that Latest wasn't described). Can we consider either 1) moving this below the description of Latest, or 2) rephrasing the explanation of Resolvable to not use `any` (e.g.: "The latest version that can be resolved, when combined with all other dependencies. This version corresponds to what pub upgrade would give you if all version constraints in pubspec.yaml were unbounded.")

src/tools/pub/cmd/pub-outdated.md
mit-mit: Forget this comment again; the example does include an outdated transitive dep (`meta`), so this is great!

kwalrath: Got it, thanks!

kwalrath:
Ready to go? I think we can publish this ahead of 2.8.

mit-mit:
One issue: Looks like some extra indentation was added here? I think there might be some leading spaces?

```
$ pub outdated

    Dependencies  Current    Upgradable  Resolvable  Latest  
    args          1.4.4      1.6.0       1.6.0       1.6.0   
    http          0.11.3+17  0.11.3+17   0.12.1      0.12.1  
    path          1.6.2      1.6.2       1.6.2       1.7.0   
```

kwalrath:
For some reason the travis build isn't showing up, but it succeeded.

PR #2336, opened 2020-04-07
Add GNU diffutils to the list of build prereqs

kwalrath:
I'm not sure it's required, but it seems wise to match the change in flutter/website#3916.


kwalrath:
Thanks, @chalin!

PR #2339, opened 2020-04-10
Fixed an instance of inconsistency regarding notes and other warnings on the async-await page

VarunChilukuri:


VarunChilukuri:
@googlebot I signed it!

PR #2358, opened 2020-04-16
add extension to list of keywords

johnpryan:
fixes #2357

PR #2359, opened 2020-04-17
Tweak keyword table

kwalrath:
`extension` is a built-in identifier, per https://github.com/dart-lang/site-www/pull/2358#issuecomment-614973233.

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -147,7 +147,7 @@ The following table lists the words that the Dart language treats specially.
 | [as][]{{bii}}         | [enum][]              | [in][]                | [switch][]        |
 | [assert][]            | [export][]{{bii}}     | [interface][]{{bii}}  | [sync][]{{ckw}}   |
 | [async][]{{ckw}}      | [extends][]           | [is][]                | [this][]          |
-| [await][]{{lrw}}      | [extension][]{{ckw}}  | [library][]{{bii}}    | [throw][]         |
+| [await][]{{lrw}}      | [extension][]{{bii}}  | [library][]{{bii}}    | [throw][]         |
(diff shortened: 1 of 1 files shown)


eernstg:
I have a couple of drive-by comments, cf. https://github.com/dart-lang/site-www/pull/2358#issuecomment-614973233.

The phrase 'contextual keyword' is not used in the language specification documents. I did use it in a comment in the [spec grammar](https://github.com/dart-lang/sdk/blob/ae0eca83188ce3bd99e3f012cf7eec4b0bdf2c4f/tools/spec_parser/Dart.g#L1539) , but only with `await` and `yield`. Those two words are special in that they are considered to be reserved words in the body of an `async` (they cannot be the name of any declaration, which is more strict than built-in identifiers that can't be the name of a type but can be the name of other things).

So it would make sense to single out just `await` and `yield` as being contextual keywords (they are currently presented as 'newer' because asynchrony was new some years back, but that's hardly helpful today). So they'd get superscript '1'.

All the words currently marked with superscript '1' would no more be marked with that superscript: `extension` will be a built-in identifier ('2'), and it's not obvious that we would need to mention `async`, `hide`, `of`, `on`, `show`, and `sync` as special at all: They can be used as the name of any declaration, and they can be used as identifiers in any expression. They do occur explicitly in the grammar, but it's not helpful to teach anybody to keep them in a mental list of special words (to avoid), because there are no special constraints on the use of these words as identifiers.

If we expect this to be a surprise to readers then we could mention `async`, `hide`, `of`, `on`, `show`, `sync` after the table, saying that there are no constraints on the use of those words as identifiers, even though they appear in the grammar.

kwalrath:
I'm going to file an issue about Erik's feedback. @johnpryan excepting that, lgty?

PR #2366, opened 2020-04-21
Make Flutter info more prominent

kwalrath:
We could do more... but at least this is more prominent than it was.

Staged: https://dartlang-org-staging-1.firebaseapp.com/get-dart


johnpryan:
Looks great, thanks Kathy!

PR #2369, opened 2020-04-22
Add a large Dart logo for Chocolatey (or whoever) to use

kwalrath:


kwalrath:
We do have SVG somewhere, but chocolatey needs a PNG in a public place.🤷🏽‍♀️

PR #2373, opened 2020-04-27
Add missing space to ref in SDK

DanTup:
Minor tweak, but I can't unsee this missing space before the first ref [here](https://dart.dev/tools/sdk/archive) 😃

Screenshot 2020-04-27 at 09 46 26

This adds the space (and also removes the superfluous space in the place that already had it for consistency).

PR #2375, opened 2020-04-27
Add Debian packages to the SDK archive

mit-mit:
@whesse was sad to see the dev-channel Debian package link get removed from https://dart.dev/get-dart in https://github.com/dart-lang/site-www/commit/081802849abd3beb04f2650c341fcc478f21ea3d. This re-introduces that link, but for all channels in the SDK archive.

@whesse I staged this here. Would this work? https://mit-staging.web.app/tools/sdk/archive

### DIFF
--- a/src/tools/sdk/dart_sdk_archive/lib/src/util.dart
+++ b/src/tools/sdk/dart_sdk_archive/lib/src/util.dart
@@ -33,7 +33,7 @@ String svnRevisionForVersion(String svnVersion) {
   return null;
 }
 
-Map<String, String> archiveMap = {
+const Map<String, String> archiveMap = {
   'Mac': 'macos',
   'Linux': 'linux',
   'Windows': 'windows',
@@ -45,19 +45,19 @@ Map<String, String> archiveMap = {
   'Dartium': 'dartium',
 };
 
-Map<String, String> directoryMap = {
+const Map<String, String> directoryMap = {
   'Dart SDK': 'sdk',
   'Dartium': 'dartium',
   'Debian package': 'linux_packages',
 };
 
-Map<String, String> suffixMap = {
+const Map<String, String> suffixMap = {
   'Dart SDK': '-release.zip',
   'Dartium': '-release.zip',
   'Debian package': '-1_amd64.deb',
 };
 
-Map<String, List<PlatformVariant>> platforms = {
+const Map<String, List<PlatformVariant>> platforms = {
   'Mac': [
     PlatformVariant('ia32', ['Dart SDK', 'Dartium']),
     PlatformVariant('x64', ['Dart SDK', 'Dartium']),
--- a/src/tools/sdk/dart_sdk_archive/lib/src/util.dart
+++ b/src/tools/sdk/dart_sdk_archive/lib/src/util.dart
@@ -42,14 +42,19 @@ Map<String, String> archiveMap = {
   'ARMv7': 'arm',
   'ARMv8 (ARM64)': 'arm64',
   'Dart SDK': 'dartsdk',
-  'Dartium': 'dartium'
+  'Dartium': 'dartium',
 };
 
-Map<String, String> directoryMap = {'Dart SDK': 'sdk', 'Dartium': 'dartium'};
+Map<String, String> directoryMap = {
+  'Dart SDK': 'sdk',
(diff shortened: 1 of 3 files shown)


mit-mit:
@athomas does the Url's here look sane?

@kevmoo sorry to hack your code. I couldn't get everything from the config.

src/tools/sdk/dart_sdk_archive/lib/src/version_selector.dart
```diff
@@ -193,15 +194,27 @@ class VersionSelector {
               }
             }
 
+            String baseFileName = '${archiveMap[pa]}-${archiveMap[name]}-'
+                '${archiveMap[platformVariant.architecture]}';
+
+            if (pa == 'Debian package') {
+              // Debian packages start with 2.0.0
+              if (versionInfo.version < Version(2, 0, 0)) {
```
kevmoo, line 202: Should probably store that version in a top-level final. So it's not created for every entry

src/tools/sdk/dart_sdk_archive/lib/src/util.dart
```diff
@@ -42,14 +42,19 @@ Map<String, String> archiveMap = {
   'ARMv7': 'arm',
   'ARMv8 (ARM64)': 'arm64',
   'Dart SDK': 'dartsdk',
-  'Dartium': 'dartium'
+  'Dartium': 'dartium',
 };
 
-Map<String, String> directoryMap = {'Dart SDK': 'sdk', 'Dartium': 'dartium'};
+Map<String, String> directoryMap = {
```
kevmoo, line 48: Make const?

src/tools/sdk/dart_sdk_archive/lib/src/util.dart
```diff
@@ -42,14 +42,19 @@ Map<String, String> archiveMap = {
...
+  'Dart SDK': 'sdk',
+  'Dartium': 'dartium',
+  'Debian package': 'linux_packages',
+};
 
 Map<String, String> suffixMap = {
```
kevmoo, line 54: same here?

src/tools/sdk/dart_sdk_archive/lib/src/util.dart
```diff
@@ -42,14 +42,19 @@ Map<String, String> archiveMap = {
   'ARMv7': 'arm',
   'ARMv8 (ARM64)': 'arm64',
   'Dart SDK': 'dartsdk',
-  'Dartium': 'dartium'
+  'Dartium': 'dartium',
 };
```
whesse, line 46: Do we need to keep dartium here?

kevmoo: Dartium *was* a valid download for a long time.

Although it's expired now.

Maybe it would be nice to drop...

mit-mit: Is it not still needed for older builds?

kevmoo:
It was optional for older builds – but it had a kill switch in it that
makes it not work after 6 months.

...so I think it's fine if we don't make it easy to find

src/tools/sdk/dart_sdk_archive/lib/src/version_selector.dart
mit-mit: You mean `Version(2, 0, 0)`? That's only used in this single place.

src/tools/sdk/dart_sdk_archive/lib/src/util.dart
mit-mit: Done

src/tools/sdk/dart_sdk_archive/lib/src/util.dart
mit-mit: Done

kwalrath:
@mit-mit / @johnpryan: Where is it documented how you update the .js?

(Asking for @DanTup, who has a PR out: #2373)

mit-mit:
Created https://github.com/dart-lang/site-www/pull/2378

mit-mit:
How about we do a second PR for Dartium? Makes is easier if we need to revert.

kevmoo:
make sure to rebuild!

PR #2376, opened 2020-04-27
Refactor sentence: you’ll learn the about futures, async, and await

farr64:
Original:
In the next sections you’ll learn the about futures, async, and await

Proposed: 
In the next sections you'll learn about futures and about working with futures (using async and await)


farr64:
@googlebot I signed it!

PR #2379, opened 2020-04-27
Remove Dartium downloads

mit-mit:
Remove the Dartium downloads as the existing ones no longer work (they were built to have a duration of six months), and we no longer rely on Dartium for debugging (see https://dart.dev/web/debugging).


mit-mit:
@kevmoo can I haz lgtm?

PR #2391, opened 2020-05-06
Update banner for 2.8

kwalrath:


kwalrath:
@mit-mit & @messi01 gave an out-of-band lgtm, so I plan to push this.

PR #2404, opened 2020-05-14
Clarify relative imports guidance

mit-mit:


src/_guides/language/effective-dart/usage.md
```diff
@@ -116,11 +113,18 @@ under `test` that other libraries in `test` import.
...
+library, which can lead to unexpected duplicated types.
+
+Follow these two rules:
 
 * An import path should never contain `/lib/`.
 * A library under `lib` should never use `../` to escape the `lib` directory.
 
+In summary, there is no profound reason to prefer the former—it's just shorter-
+but using a single style of imports within `lib` is required to avoid unexpected
+type duplication.
```
munificent, line 125: The summary is a little confusing because the first half seems to disagree with the latter half. It's really talking about two separate concepts: imports that reach in/out of "lib" and imports that don't. I'd just leave this sentence out.

mit-mit: Done!

kwalrath:
We do need to regenerate the index. (I'm not sure that works automagically, but one way or another, we need to update the list seen on the main Effective Dart page.) I think we also want to add an anchor for the old title, so that old links won't break.

mit-mit:
Thanks, I'll go ahead and merge this

Issue #2402, opened 2020-05-13
Remove "PREFER placing external 'package:' imports before other imports."


munificent:
I personally like the guideline, but we've never had automated tooling support for it and don't have any plans. I think the benefits of being able to easily automatically sort imports outweigh the benefits of seeing a package's own "package:" imports separately.

I think we should just remove the rule and let the previous rule cover all "package:" imports.

pq:
> I personally like the guideline, but we've never had automated tooling support for it

Unless I'm misunderstanding, I _think_ we do?

I believe that the sorting analysis server provides and you can call out to in IntelliJ via the Code Menu respects this ordering.

![image](https://user-images.githubusercontent.com/67586/81991446-1977da80-95f6-11ea-824e-fee5d3943ffb.png)

(fyi @scheglov)

Also as of really recently (https://github.com/dart-lang/sdk/commit/7fe16ba8125ccaa275dfb29bf029cf5d4566c105) I added a quickfix that I've been wanting forever for the corresponding `directives_ordering` lint so the tooling is slowly catching up... 🐌

munificent:
We have support for sorting imports, but (as far as I know) it doesn't separate out a package's own "package:" imports from "package:" imports of other packages. At least tidy_dart does not support that, and I think that's the one that is most widely used inside Google.

scheglov:
The sorting in Analysis Server also does not separate package's own `package:` imports from other  `package:` imports.

pq:
Ah, ok, cool.  Thanks!  In that case, bombs away :)

pq:
Is this  done?

kwalrath:
Fixed by #2405.

PR #2405, opened 2020-05-14
Remove external package import ordering

ncroll:
#2402 Remove the rule "PREFER placing external 'package:' imports before other imports."

### DIFF
--- a/src/_guides/language/effective-dart/style.md
+++ b/src/_guides/language/effective-dart/style.md
@@ -342,23 +342,6 @@ import 'util.dart';
 {% endprettify %}
 
 
-### PREFER placing external "package:" imports before other imports. {#prefer-placing-third-party-package-imports-before-other-imports}
-
-{% include linter-rule.html rule="directives_ordering" %}
-
-If you have a number of "package:" imports for your own package along with other
-external packages, place yours in a separate section after the external ones.
-
-{:.good}
-<?code-excerpt "misc/lib/effective_dart/style_lib_good.dart (third-party)" replace="/\w+\/effective_dart\///g;/(package):foo(.dart)/$1:my_package\/util$2/g"?>
-{% prettify dart tag=pre+code %}
-import 'package:bar/bar.dart';
-import 'package:foo/foo.dart';
-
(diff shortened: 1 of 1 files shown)


kwalrath:
@munificent I had the impression you were already fixing this. Is that right?

munificent:
> @munificent I had the impression you were already fixing this. Is that right?

No, I'm very happy to have it fixed, but didn't have any time to do it myself. Thanks, @ncroll!

kwalrath:
We'll need to update the index, as discussed in #2404, but I can handle both of those at the same time.

I should probably start of list of guidelines that've changed, as well. It could be in a new page similar to https://flutter.dev/docs/whats-new-archive, or... somewhere else. @munificent or @mit-mit, do you have an opinion?

ncroll:
Thanks @kwalrath and @munificent!

kwalrath:
Thank _you_, @ncroll!

pq:
Cheers for the contribution @ncroll!

I added some questions to #2404 (tl;dr are we sure we want to remove this?)

But assuming you really do want this removed, yep we'd want to update the lint and the index that @tenhobi maintains too.

munificent:
> It could be in a new page similar to https://flutter.dev/docs/whats-new-archive, or... somewhere else. @munificent or @mit-mit, do you have an opinion?

Great idea. :)

munificent:
@kwalrath, do you want to merge this in? Is there anything else we should do first?

kwalrath:
@munificent I'll merge the 2 PRs today and prepare the followup PR to update the index.

tenhobi:
Hi, not sure what needs to be changed from my side. (effective_dart package https://github.com/tenhobi/effective_dart)

This PR removes the rule which contained `{% include linter-rule.html rule="directives_ordering" %}` but that lint (`directives_ordering`) is used by many more rules. So I guess we are still fine and I don't have to change anything?

pq:
Sorry for the false alarm @tenhobi.  You're right: nothing for you to do.  Thanks for  following up!

PR #2407, opened 2020-05-15
Remove unnecessary highlight

MichalLankof:
Remove unnecessary highlight from "Example: asynchronous functions". The highlighted part is the same as in "Example: synchronous functions".

### DIFF
--- a/src/codelabs/async-await.md
+++ b/src/codelabs/async-await.md
@@ -329,14 +329,14 @@ Your order is: Instance of _Future<String>
 <div class="col-sm" markdown="1">
 #### Example: asynchronous functions
 
-<?code-excerpt "async_await/bin/get_order.dart" replace="/(\s+\/\/ )(Imagine.*? is )(.*)/$1$2$1$3/g; /async|await/[!$&!]/g; /(Future<\w+\W)( [^g])/[!$1!] $2/g; /4/2/g"?>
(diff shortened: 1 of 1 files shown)


MichalLankof:
@googlebot I signed it!

kwalrath:
Nice catch! I'll see if I can fix the build.

kwalrath:
Ugh, fixing this problem will require fixing the following regex, which is what added the highlighting:

```
<?code-excerpt "async_await/bin/get_order.dart" replace="/(\s+\/\/ )(Imagine.*? is )(.*)/$1$2$1$3/g; /async|await/[!$&!]/g; /(Future<\w+\W)( [^g])/[!$1!] $2/g; /4/2/g"?>
```

I think we'll have to either explicitly ignore `Future<String> fetch` or explicitly look for the other two occurrences of `Future<String>`.

You don't happen to be really good at regular expressions, do you?

MichalLankof:
@kwalrath I am not quite sure how this works. Is this script just reading the template file from 'async_await/bin/get_order.dart` and replaces the contents using the provided regular expression? If so, why is there the output stored in the markdown file as well? 

If we need to change the regular expression it should be 
```
/(Future<\w+\W)( [^f])/[!$1!]$2/g
```
I have changed 2 things:

1. `[^g]` to `[^f]` to filter out part starting with an 'f' (fetchUserOrder)
2. removed space before `$2` because the first space is already captured in the second group and this only produced double spaces

This new regexp will fail over the already produced part because it contains double spaces. 
I will commit the new regexp and fix the double spaces, but let me know if I understood this correctly.

kwalrath:
That's it! The function name must have started with g (probably "get...") before, and the regex didn't get updated when the function name did.

Thank you for fixing this!

PR #2409, opened 2020-05-16
Update Dart DevTools and web debugging pages, de-emphasize Chrome DevTools

### LINKED ISSUE #2130: Make web debugging page refer to new Dart DevTools
Can we correct the `Note` on this page:

https://dart.dev/web/debugging

to point the new Dart DevTools work instead?

I.e.,
> Note: A Dart fork of Chrome DevTools exists but isn’t ready yet for general use. For the latest status, see the devtools-frontend README.

Alternatively, we could delete it for now.  We have no further plans for that fork.

@kwalrath @grouma

kwalrath:
Fixes #2130 

Staged:
https://dartlang-org-staging-0.firebaseapp.com/web/debugging
https://dartlang-org-staging-0.firebaseapp.com/tools/dart-devtools


kwalrath:
@grouma am I on the right track? Could you look for "PENDING:" and answer questions I had? Thanks!

src/tools/dart-devtools.md
```diff
@@ -5,13 +5,113 @@ description: A suite of debugging and performance tools.
 
 Dart DevTools is a suite of debugging and performance tools
 for Dart and Flutter.
+These tools are distributed in the [devtools package,][devtools package]
+**[PENDING: how do you normally get devtools, aside from pub global activate?]**
```
grouma, line 9: `pub global activate` is one way to run the tool. It is also baked into `webdev` and the `Flutter` CLI.

src/web/debugging.md
```diff
@@ -3,247 +3,172 @@ title: Debugging Dart web apps
...
+     For example, if you press **Step in** at a call to `print()`,
+     you go to the next line, not into the SDK code that implements `print()`.
+   {{site.alert.end}}
 
-Chrome DevTools <!-- and Dart DevTools --> enables source maps, by default.
-You might want to temporarily disable source maps so that you can see
-the generated JavaScript code.
+1. Change your source code and reload the Chrome window that's running the app.
+   The app quickly rebuilds and reloads.
+   **[PENDING: mention something about losing your breakpoints?]**
```
grouma, line 155: Correct. @bkonyi is working on properly resolving this issue. We have a temporary workaround that is not being leveraged by `webdev` or the `Flutter` CLI. @devoncarew / @jonahwilliams thoughts on using the temporary workaround for the time being?

kwalrath:
@grouma I forgot to stage this. In case that helps, you can see the pages here:

https://dartlang-org-staging-0.firebaseapp.com/web/debugging
https://dartlang-org-staging-0.firebaseapp.com/tools/dart-devtools

src/tools/dart-devtools.md
kwalrath: It sounds like there's no reason to `pub global activate devtools` if you're writing a web app. Correct?

devoncarew: I think the most common way to run DevTools is from IDEs (VS Code and IntelliJ). The IDEs also support debugging web apps as well now.

Are we recommending web debugging through DevTools specifically, or documenting how to use from VS Code, IntelliJ, and DevTools?

grouma: > It sounds like there's no reason to pub global activate devtools if you're writing a web app (and not a command-line app). Correct?

Correct.

> Are we recommending web debugging through DevTools specifically, or documenting how to use from VS Code, IntelliJ, and DevTools?

Yes. That's one of the main changes with this PR.

src/web/debugging.md
bkonyi: Can you refresh my memory on this @grouma? :)

grouma: Right now we don't restore breakpoints if you refresh the application. `package:dwds` can handle refreshing the breakpoints through this deprecated [argument](https://github.com/dart-lang/webdev/blob/df632ec25bd68039a28b9624fa7a4dc4148a5649/dwds/lib/dwds.dart#L91). However, `webdev` nor the Flutter CLI make use of this functionality. It's only used by internal tools.

bkonyi: Ah okay, gotcha. So, this is what the [requirePermissionToResume](https://pub.dev/documentation/vm_service/latest/vm_service/VmService/requirePermissionToResume.html) RPC is for. Clients are responsible for resetting their breakpoints after reloads, so the `requirePermissionToResume` ensures that an isolate stays paused until all clients from whom the service requires permission have indicated the isolate can be resumed.

kwalrath:
Updated. ptal

Issue #2330, opened 2020-04-06
'Customizing static analysis' page issue


workerbee22:
Page URL: https://dart.dev/guides/language/analysis-options
Page source: https://github.com/dart-lang/site-www/tree/master/src/_guides/language/analysis-options.md

Found a typo? You can fix it yourself by going to the page source and clicking the pencil icon. Or finish creating this issue.

Description of issue:

Disabling Linter rules section is confusing. Two bullets given:

-Omit the dash (-) before the rule name.
-Add : false after the rule name.

Is this OR or AND ? 

After that the example given for enabling linter rules uses prefix '-' but disabling uses trailing ":true" type. 

So I'm left confused. Either or .... or both ?

domesticmouse:
The core of this confusion is understanding `YAML`'s syntax, near as I can tell. It is a case of both are required because you are turning it from a list of values, into a map of token to booleans, near as I can tell. How should we change this to make more sense?

workerbee22:
Actually I suspect it is either OR, not both. Both currently throws Dart Analysis errors in AS. Also so does just removing the leading '-'.

So suggest:
1. Fix issues so that both syntaxes work.
2. Make clear that either syntax is permitted, but not BOTH
3. Syncronis the Enable and Disable lint rule example code to both use just one of these methods. Or provide two examples, each using one method for both enable and disable lint rule.

Just my two cents, hope it helps.

domesticmouse:
Hey @bwilkerson and @pq, I'd love your input here. Thanks!

pq:
> Actually I suspect it is either OR, not both.

Sadly, I think this is right and IIRC YAML out of the box isn't hospitable to a mixed syntax.

The most expedient thing here would likely be  improving the docs.  To that end this feels like a good suggestion me:

> ... Or provide two examples, each using one method for both enable and disable lint rule.


Sorry for the confusion @workerbee22!

bwilkerson:
> How should we change this to make more sense?

I think you're right, that the confusion is a result of YAML's syntax. Perhaps the best solution is to explain a little bit about how we're using YAML's syntax. Perhaps changing the current text to something like the following:

### Enabling individual rules

To enable a single linter rule, add `linter:` to the analysis options file as a top-level key, followed by `rules:` as a second-level key. On subsequent lines, specify the rules that you want to apply, prefixed with dashes (the syntax for a YAML list). For example:

...

### Disabling individual rules

[...]

PR #2416, opened 2020-05-17
Clarifying how to enable and disable lints

domesticmouse:
Fixes https://github.com/dart-lang/site-www/issues/2330


kwalrath:
Actually, I think I'll revise the line breaks... For consistency's sake.

Issue #2312, opened 2020-03-23
Initial null safety docs for tech preview


mit-mit:
As of https://github.com/dart-lang/pub/pull/2465, pub will link to https://dart.dev/null-safety/migration-guide, so we need to provide a file or redirect for that URL.

kwalrath:
https://github.com/dart-lang/pub/pull/2465/files#diff-7192e860ddb78fc60e6bb87a113e4535 proposes a link to a migration guide, perhaps at /null-safety/migration-guide.

kwalrath:
Working on a PR now for a placeholder and redirect.

franklinyow:
Can this be closed?

kwalrath:
I think so. @mit-mit?

mit-mit:
Yes, we are no longer blocked, closing

PR #2424, opened 2020-05-19
Add null safety placeholder

kwalrath:
Addresses #2312.

Staged:
https://kw-staging-dartlang-2.firebaseapp.com/null-safety
https://kw-staging-dartlang-2.firebaseapp.com/null-safety/migration-guide


mit-mit:
Place holder LGTM, but I don't think we want a "migration guide" this soon. That may be misconceived as a migration call-to-action, and that's too soon.

kwalrath:
The migration guide is just a temporary redirect to the main null safety page. I thought that URL was mentioned in command output, so we needed that redirect?

mit-mit:
I don't know if that's the case or not. But if it is the case, I think the Url should go to an actual placeholder page for migration, which should say something like "As part of the final null safety feature, we plan on making a migration tool available. At the current stage, with null safety still in preview, we do not recommend doing anything but test migration"

kwalrath:
ptal

mit-mit:
Looks good, thanks!

Issue #2427, opened 2020-05-20
Git URL in dependencies examples are not working


lzhuor:
Example given is using `git://` which causes error during `pub get` command:

```yaml
dependencies:
  kittens:
    git:
      url: git://github.com/munificent/kittens.git
      ref: some-branch
```

![image](https://user-images.githubusercontent.com/14172220/82437597-429edc00-9aca-11ea-8848-4e2642dc4b39.png)

The correct one should be `git@git@github.com/XXX`

kwalrath:
Thanks for the issue & PR, @lzhuor! I've asked an engineer to review the PR.

jonasfj:
I have no problems cloning:
```yaml
name: hack
dependencies:
        retry:
                git:
                        url: git://github.com/google/dart-neats.git
                        ref: master
                        path: retry
```

On linux in a  clean docker image without any special configuration of my git client.

-----------
That said, it might be fine to improve the documentation by pointing out that any url that works with `git clone <url>` will do. And that the common forms for this on github is:
 * `git@github.com:munificent/kittens.git` for private repositories, and,
 * `https://github.com/munificent/kittens.git` for public repositories.

PR #2428, opened 2020-05-20
Update dependencies.md to fix git import URL in examples

lzhuor:
Fixes: #2427 

- Replaced `git://github.com` with `git@github.com` for `pubspec.yaml`.

### DIFF
--- a/src/tools/pub/dependencies.md
+++ b/src/tools/pub/dependencies.md
@@ -139,7 +139,7 @@ stored in a [Git][] repository.
 {% prettify yaml tag=pre+code %}
 dependencies:
   kittens:
-    git: git://github.com/munificent/kittens.git
+    git: git@github.com/munificent/kittens.git
 {% endprettify %}
 
 The `git` here says this package is found using Git, and the URL after that is
@@ -162,7 +162,7 @@ add a `ref` argument:
 dependencies:
   kittens:
     git:
-      url: git://github.com/munificent/kittens.git
+      url: git@github.com/munificent/kittens.git
       ref: some-branch
 {% endprettify %}
 
@@ -177,7 +177,7 @@ To specify a different location in the repo, use the `path` argument:
 dependencies:
   kittens:
     git:
-      url: git://github.com/munificent/cats.git
+      url: git@github.com/munificent/cats.git
       path: path/to/kittens
 {% endprettify %}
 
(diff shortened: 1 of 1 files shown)


kwalrath:
@jonasfj can you please review this (or find someone to do so)?

Looking in the site-www source, the only git URLs I see use https:

    git: https://github.com/chalin/code_excerpt_updater.git

kwalrath:
@kevmoo maybe you can review?

src/tools/pub/dependencies.md
```diff
@@ -139,7 +139,7 @@ stored in a [Git][] repository.
 {% prettify yaml tag=pre+code %}
 dependencies:
   kittens:
-    git: git://github.com/munificent/kittens.git
+    git: git@github.com/munificent/kittens.git
```
jonasfj, line 142: For github and similar the format is either:
 * `git@github.com:munificent/kittens.git` (this uses SSH, and requires you to have an SSH key registered with github)
 * `https://github.com/munificent/kittens.git` (provides anonymous read-only access)

**You are missing the colon** in `git@github.com:munificent/kittens.git`, which is a slash in HTTPS.

In general, `pub get` will shell out to `git` so any URL that works with `git clone <url>` should work.

Commonly the useful advice would be to use:
 * `git@github.com:munificent/kittens.git` for private repositories, and,
 * `https://github.com/munificent/kittens.git` for public repositories.

lzhuor: Fixed.

lzhuor:
@jonasfj @kwalrath Fixed. Please check. Thank you.

src/tools/pub/dependencies.md
```diff
@@ -139,7 +139,7 @@ stored in a [Git][] repository.
 {% prettify yaml tag=pre+code %}
 dependencies:
   kittens:
-    git: git://github.com/munificent/kittens.git
+    git: https://github.com/munificent/kittens.git
```
lzhuor, line 142: @jonasfj I left this as `https` URL to be an example for public repo.

lzhuor:
> Seems fine... we could also add a box-thingy explaining the two kinds of URLs...

Hi @jonasfj , I didn't add it because I found there is already a sentence about ssh URL:
image

PR #2430, opened 2020-05-20
Use num less often

### LINKED ISSUE #2423: Use num less often
From @timsneath:
 
> We have about 30 references to `num` in the language tour, but only one is needed (where we explain that it's available as a supertype of `int` and `double`). 
> 
> In the tour, it's repeatedly used for didactic purposes in the form of a `Point` class. In most real-world code, I'd expect that a `double` would be used here (or perhaps `int`, if `Point` represented a discrete integer-based pixel model). 
> 
> For someone new to the language who is using this to quickly learn Dart's syntax and general philosophy, can we simply replace `num` with one or the other subtype?

I suspect `num` might be overused in other areas of the site, as well.

kwalrath:
/cc @munificent @filiph because they wrote some of these snippets and text
/cc @timsneath @mit-mit because they care

Publicly visible changes are in the .md files, which start here:
https://github.com/dart-lang/site-www/pull/2430/files#diff-e1df7aee84d572a5d2aeea1fd3eb38f8

Staged:
https://dartlang-org-staging-1.firebaseapp.com/guides/language/effective-dart/usage
https://dartlang-org-staging-1.firebaseapp.com/guides/language/language-tour
https://dartlang-org-staging-1.firebaseapp.com/codelabs/dart-cheatsheet
https://dartlang-org-staging-1.firebaseapp.com/samples

Fixes #2423

### DIFF
--- a/examples/misc/lib/effective_dart/usage_good.dart
+++ b/examples/misc/lib/effective_dart/usage_good.dart
@@ -345,7 +345,8 @@ class C {
   // #docregion use-arrow
   num get area => (right - left) * (bottom - top);
 
-  bool isReady(double time) => minTime == null || minTime <= time;
+  bool isReady(double time) =>
+      minTime == null || minTime <= time;
 
   String capitalize(String name) =>
       '${name[0].toUpperCase()}${name.substring(1)}';
--- a/examples/misc/lib/effective_dart/usage_good.dart
+++ b/examples/misc/lib/effective_dart/usage_good.dart
@@ -31,7 +31,7 @@ void miscDeclAnalyzedButNotTested() {
     // #enddocregion adjacent-strings-literals
   }
 
-  (String name, num year, num birth) {
+  (String name, int year, int birth) {
     // #docregion string-interpolation
     'Hello, $name! You are ${year - birth} years old.';
     // #enddocregion string-interpolation
@@ -300,12 +300,12 @@ class LazyId {
 
 // #docregion cacl-vs-store
 class Circle {
-  num radius;
+  double radius;
 
   Circle(this.radius);
 
-  num get area => pi * radius * radius;
-  num get circumference => pi * 2.0 * radius;
+  double get area => pi * radius * radius;
+  double get circumference => pi * 2.0 * radius;
 }
 // #enddocregion cacl-vs-store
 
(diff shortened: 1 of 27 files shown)


johnpryan:
LGTM - some other places to consider updating in the future:

https://dart.dev/guides/language/effective-dart/design#do-annotate-when-dart-infers-the-wrong-type
https://dart.dev/guides/language/effective-dart/documentation#consider-including-code-samples-in-doc-comments
https://dart.dev/guides/language/extension-methods#api-conflicts (parseNum() method)

examples/misc/lib/effective_dart/usage_good.dart
```diff
@@ -338,14 +338,15 @@ class Treasure {
 }
 
 class C {
-  double left, right, top, bottom, minTime;
+  num left, right, top, bottom, minTime;
```
timsneath, line 341: Did you mean to turn this from `double` to `num`? 

(I'm actually surprised it's not `int` here...)

examples/misc/lib/effective_dart/usage_good.dart
```diff
@@ -338,14 +338,15 @@ class Treasure {
...
   Point center;
   Map<Chest, Treasure> _opened;
 
   // #docregion use-arrow
-  double get area => (right - left) * (bottom - top);
+  num get area => (right - left) * (bottom - top);
```
timsneath, line 346: I get the impression that this is representing a game board, so I would have _expected_ this to be an `int`, I think...

kwalrath:
Thanks for the careful review!

> LGTM - some other places to consider updating in the future:

Agreed. These weren't as easy to update, so I punted.

examples/misc/lib/effective_dart/usage_good.dart
kwalrath: Oops, missed your comments here before I committed. I don't recall why I changed these. I think there was an analysis error. I'll check.

examples/misc/lib/effective_dart/usage_good.dart
kwalrath: It's calculated from doubles (if I change the type that's on line 341).

timsneath: Right, I mean we're discussing an imaginary app here, so it's hard to do a code review ;) but I could see both these being `int` if they represent locations in a game. But we're into a nit on top of a nit now!

PR #2449, opened 2020-05-27
Update pubspec.lock commit guidance

### LINKED ISSUE #2408: [discussion]: Tweak pubspec.lock recommendations
See: https://dart.dev/guides/libraries/private-files#pubspeclock

**Today, we say:**
> **For library packages**, do **not** commit the `pubspec.lock` file.
> **For application packages**, it’s up to you whether you commit the `pubspec.lock` file. Consider checking it in if you want to track the dependencies that are used for production or across a team.

**I think, it would be better to say:** (_or something along these lines_)
> **For library packages**, it can be a good idea to not commit the `pubspec.lock` file. Committing the `pubspec.lock` file will not affect package consumers, they are likely to get the latest dependencies allowed by `pubspec.yaml`. Therefore it can be a good idea to avoid locking your dependencies and test against the latest versions of your dependencies. However, committing the `pubspec.lock` file may produce more consistent CI test results, and whether or not to commit it can be a matter of personal preference and workflow.

**I _strongly_ think, it would be better to say:** (_or something along these lines_)
> **For application packages**, always **do commit** the `pubspec.lock`. This will lock your dependencies, so that code changes in your dependencies doesn't sneak in as your application progresses from development, to integration testing, to product release, to product support.

----------------

## Rationale for library packages recommendations

When developing a library package the `pubspec.lock` will not affect the package consumers.
When `pub get` downloads a package it will **not read** the `pubspec.lock` from the package being downloaded. Package consumers will always get the latest dependencies compatible with the constraints in `pubspec.yaml`.

Hence, whether you choose to keep `pubspec.lock` under source control is a matter of personal preference (or workflow).

Reasons to **keep `pubspec.lock` in git** includes:
 * It is easier to reproduce behavior your collaborators experience, because all developers of the library package has the same dependencies.
 * Your CI results (on travis, jenkins, etc) won't break unexpectedly because some upstream package changed. Hence, your CI results are more consistent, and less intermittent.

For large teams working on a library package, locking dependencies might make sense. Because intermittent CI results can be disruptive to the workflow.

The counter argument is that you easily forget to test against latest dependencies if your CI system isn't doing it. And the package consumers will likely get the latest dependencies, so it's unfortunate that you're not testing against them.

I don't feel too strongly about recommendation we give, but making it clear that `pubspec.lock` won't affect package consumers seems like a good idea.

## Rationale for application packages recommendations

When developing an application package the `pubspec.lock` specifies part of the code that makes up your application. If you don't commit the `pubspec.lock` file, you may inadvertently upgrade your dependencies between testing and deployment. Common scenarios include:

 * Flutter deployment using tutorials for [Cirrus](https://flutter.dev/docs/deployment/cd#cloud-build-and-deploy-setup), [codemagic](https://blog.codemagic.io/getting-started-with-codemagic/) or [bitrise.io](https://devcenter.bitrise.io/getting-started/getting-started-with-flutter-apps/).
 * Deployment of any kind from [GitHub Actions](https://github.com/features/actions) or [GitLab CI/CD](https://docs.gitlab.com/ee/ci/).
 * Deploying to Platform-as-a-Service providers from git, examples includes heroku (which use `git push heroku` for deployment), Google Cloud Build which can deploy to AppEngine, Cloud Run, etc.
 * Deployment from a different working folder (git checkout) or physical machine than was used for primary development.

Doing any of this without a `pubspec.lock` under source control and you may easily deploy a configuration you have not tested. And once you have deployed this untested configuration, you may have a hard time reproducing issues reported by users, because you might have different dependencies (and it's hard to guess which packages was actually deployed into production).

Therefore it is in my opinion very important that application developers keep `pubspec.lock` under source control. 

---------
CC @sigurdm, @mit-mit, @natebosch, @jakemac53, please tune-in if you have any thoughts.

kwalrath:
Fixes #2408

### DIFF
--- a/src/_guides/libraries/private-files.md
+++ b/src/_guides/libraries/private-files.md
@@ -101,7 +101,7 @@ The `pubspec.lock` file is a special case,
 similar to Ruby's `Gemfile.lock`.
 
 **For library packages**, **don't commit** the `pubspec.lock` file.
-Recreating the `pubspec.lock` file lets you test your package
+Regenerating the `pubspec.lock` file lets you test your package
 against the latest compatible versions of its dependencies.
 
 **For application packages**, we recommend that you
--- a/src/_guides/libraries/private-files.md
+++ b/src/_guides/libraries/private-files.md
@@ -100,9 +100,11 @@ using [pub get](/guides/packages#getting-packages).
 The `pubspec.lock` file is a special case,
 similar to Ruby's `Gemfile.lock`.
 
-**For library packages**, do **not** commit the `pubspec.lock` file.
+**For library packages**, **don't commit** the `pubspec.lock` file.
+Recreating the `pubspec.lock` file lets you test your package
+against the latest compatible versions of its dependencies.
 
-**For application packages**, it's up to you whether you
-commit the `pubspec.lock` file.
-Consider checking it in if you want to track the dependencies that
-are used for production or across a team.
+**For application packages**, we recommend that you
+**commit** the `pubspec.lock` file.
+Saving `pubspec.lock` ensures that everyone working on the app
+uses the exact same versions.


src/_guides/libraries/private-files.md
```diff
@@ -100,9 +100,11 @@ using [pub get](/guides/packages#getting-packages).
...
 similar to Ruby's `Gemfile.lock`.
 
-**For library packages**, do **not** commit the `pubspec.lock` file.
-
-**For application packages**, it's up to you whether you
-commit the `pubspec.lock` file.
-Consider checking it in if you want to track the dependencies that
-are used for production or across a team.
+**For library packages**, **don't commit** the `pubspec.lock` file.
+Recreating the `pubspec.lock` file lets you test your package
```
parlough, line 104: `Regenerating` or simply `Generating` rather than `Recreating` seem more fitting, as `Recreating` to me seems to imply the user creates it by hand. I believe the word `generate` is used elsewhere as well.

kwalrath: Good point. Thanks!

Issue #2392, opened 2020-05-06
one example wrong in Language Style


amitbar05:
In section [DO capitalize acronyms and abbreviations longer than two letters like words](https://dart.dev/guides/language/effective-dart/style#do-capitalize-acronyms-and-abbreviations-longer-than-two-letters-like-words)
You got a conflict between:
Good examples:
DB, Id
Bad examples:
Db, ID

Just need to change in good example the "Id" to "ID".

kwalrath:
Ah, but ID is an abbreviation, not an acronym. It's covered by the following sentence:

>  (Two-letter abbreviations like ID and Mr. are still capitalized like words.)

Issue #2327, opened 2020-04-01
'Effective Dart' page issue


jessika:
Page URL: https://dart.dev/guides/language/effective-dart/style#do-capitalize-acronyms-and-abbreviations-longer-than-two-letters-like-words
Page source: https://github.com/dart-lang/site-www/tree/master/src/_guides/language/effective-dart/index.md

Found a typo? You can fix it yourself by going to the page source and clicking the pencil icon. Or finish creating this issue.

Description of issue:

The guide says "Two-letter abbreviations like ID and Mr. are still capitalized like words." but then it shows "Id" as one of the "good" examples and "ID" as one the "bad" examples. Is this incorrect? It's not clear to me if two-letter abbrevations can go either way or if they *must* be written in all-caps.

jessika:
Nevermind, "ID" is an "abbreviation" not an "acronym".

PR #2455, opened 2020-05-28
fix typos in Effective Dart: Stygle

weakish:


kwalrath:
Hi, these spellings are intentional, although you're not the first to think they aren't (e.g. see #2327 & #2392). I'll file an issue to clarify the text.

/cc @munificent

PR #2457, opened 2020-05-29
Update /null-safety to include actual docs

kwalrath:
Staged: https://dartlang-org-staging-1.firebaseapp.com/null-safety

Even if this is perfect (haha), we'll need to remove the QUESTION before publishing this page.

### DIFF
--- a/src/null-safety/index.md
+++ b/src/null-safety/index.md
@@ -1,38 +1,246 @@
 ---
 title: Null safety
-description: Watch this space for news about null safety
-toc: false
+description: Information about Dart's upcoming null safety feature
 ---
 
-Watch this page for updates about null safety.
-In the meantime, check out these resources:
+Sound null safety is coming to the Dart language!
+Once you opt into null safety,
+types are non-nullable by default, meaning that
+values in your code can’t be null _unless you say they can be._
+With null safety, your **runtime** null-reference exceptions
+turn into **edit-time** analysis errors.
+The compiler can optimize away internal null checks,
+enabling apps to be faster, smaller, and more reliable.
 
-* [Null-safe playground: nullsafety.dartpad.dev][nullsafety.dartpad.dev]
+You can practice using null safety in the web app
+[DartPad with Null Safety.][nullsafety.dartpad.dev]
+
+{{ site.alert.important }}
+  **Don't start migrating large projects yet.**
+  Do test the feature and [give us feedback.][]
+{{ site.alert.end }}
+
+[give us feedback.]: https://github.com/dart-lang/sdk/issues/new?title=Null%20safety%20feedback:%20[issue%20summary]&labels=NNBD&body=Describe%20the%20issue%20or%20potential%20improvement%20in%20detail%20here
+
+If you've used Kotlin, TypeScript, or C#,
+then some of the null-safety syntax might look familiar.
+That's by design: the Dart language aims to be unsurprising.
+
+
+## Creating and using variables (?, !, late)
+
+Use `?`, `!`, and `late` to tweak how the Dart analyzer treats your code.
+
+Here’s an example of declaring a **non-nullable integer variable**
+(assuming you’ve opted into null safety):
+
+```dart
+int aRealInt; // Can't be used until it's initialized to an integer.
+```
+
+If the variable _can_ have the value `null`,
+**add  `?`** to its type declaration:
+
+```dart
+int? aNullableInt = null;
+```
+
+When using a value that you know isn’t null but that has a nullable type,
+**add `!`** to tell the Dart analyzer that the value isn’t null:
+
+```dart
+int? a = 2;
+int b = a!; // The ! temporarily changes `a` from int? to int.
+```
+
+If you know that a non-nullable value will be
+initialized to a non-null value before it's used,
+but the Dart analyzer thinks otherwise,
+**add the `late` keyword** to the type:
+
+```dart
+late int aRealInt;
+```
+
+{% comment %}
+{{ site.alert.tip }}
+  Avoid using `late final` unless you really need it.
+  For example, don't use `late final` in functions,
+  and in constructors prefer using an [initializer list][].
+{{ site.alert.end }}
+
+[initializer list]: /guides/language/language-tour#initializer-list
+{% endcomment %}
+
+
+## Understanding list, set, and map types
+
+Lists, sets, and maps are commonly used collection types in Dart programs,
+so you need to know how they interact with null safety.
+Here are some examples of how Dart code uses these collection types:
+
+* Flutter layout widgets such as [`Column`][]
+  often have a `children` property that’s a
+  `List` of `Widget` objects.
+* The [Veggie Seasons][] sample uses a `Set` of `VeggieCategory` to
+  store a user’s food preferences.
+* The [GitHub Dataviz][] sample has a `fromJson()`method that
+  creates an object from JSON data that’s supplied in a
+  `Map<String, dynamic>`.
+
+[`Column`]: https://api.flutter.dev/flutter/widgets/Column-class.html
+[Veggie Seasons]: https://github.com/flutter/samples/tree/master/veggieseasons
+[GitHub Dataviz]: https://github.com/flutter/samples/tree/master/web/github_dataviz
+
+
+### List and set types {#list-and-set-types}
+
+When you’re declaring the type of a list or set, think about what can be null. The following table shows the possibilities for a list of strings
+if you opt into null safety.
+
+{% assign yes = '<b>Yes</b>' %}
+
+|------------------+----------------------------------+--------------------------+------------|
+|Type              | Can an item (string)<br>be null? | Can the list<br>be null? | Description|
+|------------------|----------------------------------|--------------------------|------------|
+| `List<String>`   | No      | No      | A non-null list that contains<br> non-null strings       |
+| `List<String>?`  | No      | {{yes}} | A non-null list that contains<br> strings that **might be null** |
+| `List<String?>`  | {{yes}} | No      | A list that **might be null** and that<br> contains non-null strings |
+| `List<String?>?` | {{yes}} | {{yes}} | A list that **might be null** and that<br> contains strings that **might be null**<br> (same as `List<String>` if you <br>don't opt into null safety) |
+{:.table .table-striped}
+
+When literals initialize a list or set,
+then instead of seeing a type like in the table above,
+you typically see a type annotation on the literal.
+For example, here’s what a `List<String?>` (called `nameList`) and
+a `Set<String?>` (called `nameSet`) might look like:
+
+```dart
+var nameList = <String?>['Andrew', 'Anjan', 'Anya'];
+var nameSet = <String?>{'Andrew', 'Anjan', 'Anya'};
+```
+
+### Map types {#map-types}
+
+Map types behave mostly like you’d expect, with one exception: **the returned value can be null**. Null is the value for a key that isn't present in the map.
+
+As an example, look at the following code. What do you think the value and type of `uhOh` are?
+
+```dart
+var myMap = <String, int>{'one': 1};
+var uhOh = myMap['two'];
+```
+
+The answer is that `uhOh` is null and has type `int?`. 
+
+Like lists and sets, maps can have a variety of type permutations.
+
+|----------------------+-------------------------------+-------------------------|
+|Type                  | Can an item (int)<br>be null? | Can the map<br>be null? |
+|----------------------|-------------------------------|-------------------------|
+| `Map<String, int>`   | No*                           | No                      |
+| `Map<String, int>?`  | No*                           | {{yes}}                 |
+| `Map<String, int?>`  | {{yes}}                       | No                      |
+| `Map<String, int?>?` | {{yes}}                       | {{yes}}                 |
+{:.table .table-striped}
+
+_\* Even when all the int values in the map are non-null,
+when you use an invalid key to do a map lookup, the returned value is null._
+
+Because map lookups can return null,
+you can't assign them to non-nullable variables:
+
+```dart
+// Assigning a lookup result to a non-nullable
+// variable causes an error.
+int value = <String, int>{'one': 1}['one']; // ERROR
+```
+
+**[QUESTION: Will the analyzer ever be able to detect that is non-null?]**
+
+One workaround is to change the type of the variable to be nullable:
+
+```dart
+int? value = <String, int>{'one': 1}['one']; // OK
+```
+
+Another way to fix the problem —
+if you're sure the lookup succeeds —
+is to add a `!`:
+
+```dart
+int value = <String, int>{'one': 1}['one']!; // OK
+```
+
+A safer approach is to use the lookup value only if it's not null.
+You can test its value using
+an [`if` statement][] or the [`??` operator][].
+Here's an example of using the value `0` if the lookup returns a null value:
+
+```dart
+var aList = <String, int>{'one': 1};
+...
+int value = aList['one'] ?? 0;
+```
+
+[`if` statement]: /guides/language/language-tour#if-and-else
+[`??` operator]: /guides/language/language-tour#conditional-expressions
+
+
+## Fixing analysis errors
+
+With sound null safety, the Dart analyzer generates errors when
+it finds a nullable value where a non-null value is required.
+That's not as bad as it sounds:
+the analyzer can often recognize when
+a variable or expression inside a function has
+a nullable type but can't have a null value.
+However, the analyzer can't model the flow of your whole application,
+so it can't predict the value of a global variable or class field.
+
+If you know a value can't be null
+but it has a nullable type,
+consider using `!`, `late`, or `as`.
+Examples of using `!` and `late` are in the [first section of this page][].
+Here's an example of using `as` to convert an `int?` to a `int`:
+
+```dart
+return maybeInt(3) as int;
+```
+
+[first section of this page]: #creating-and-using-variables---late
+
+If a value might indeed be null,
+then you can explicitly check for that.
+A common approach is to use an `if` statement or `??` expression
+to check for and handle a null value.
+The analyzer then considers the code for the non-null value to be null safe.
+Here's an example of using `??`:
+
+```dart
+int value = aNullableInt ?? 0; // 0 if it's null; otherwise, the integer
+```
+
+Here's an example of using an `if` statement:
+
+```dart
+int definitelyInt(int? aNullableInt) {
+  if (aNullableInt == null) {
+    return 0;
+  }
+  return aNullableInt; // Can't be null!
+}
+```
+
+
+## Where to learn more
+
+For more information about null safety, see the following resources:
+
+* [DartPad with Null Safety][nullsafety.dartpad.dev]
 * [Null safety tracking issue][110]
 * [Dart announcements group][Dart announce]
 * [Dart blog][]
-
-Here's some background about null safety
-from the [Dart 2.8 announcement:][Announcing Dart 2.8]
-
-<blockquote markdown="1">
-  A common source of app crashes is
-  code that attempts to use a variable that happens to be null.
-  Sir Tony Hoare, who introduced null references in the
-  ALGOL programming language in 1965,
-  famously called them his “billion-dollar mistake” in a QCon talk in 2009.
-  In some cases null values are useful;
-  the challenge is separating those from the cases where they aren’t.
-
-  For the past year we’ve been busy building support for
-  sound null safety in Dart.
-  This support will extend the type system to
-  express variables that are always non-nullable,
-  but in addition the type system will be fully sound:
-  the Dart compilers and runtimes will be able to trust those types,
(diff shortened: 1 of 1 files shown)


mit-mit:
@lrhn any chance you have time to review this today?

src/null-safety/index.md
```diff
@@ -1,39 +1,247 @@
...
+**add the `late` keyword** to the type:
+
+```dart
+late int aRealInt;
+```
+
+{% comment %}
+{{ site.alert.tip }}
+  Avoid using `late final` unless you really need it.
+  For example, don't use `late final` in functions,
```
lrhn, line 68: "In functions" just means "for local variables"?

Where does this advice come from?
There *are* uses for local late final variables, so "don't use" is too strong. "Use judiciously" would be better.

src/null-safety/index.md
```diff
@@ -1,39 +1,247 @@
...
+  and in constructors prefer using an [initializer list][].
```
lrhn, line 69: Constructors don't declare late final variables, so this is referring to instance variables?

If so, then maybe "in classes, prefer fields initialized using an [initializer list][]."

src/null-safety/index.md
```diff
@@ -1,39 +1,247 @@
...
+When you’re declaring the type of a list or set, think about what can be null. The following table shows the possibilities for a list of strings
+if you opt into null safety.
+
+{% assign yes = '<b>Yes</b>' %}
+
+|------------------+----------------------------------+--------------------------+------------|
+|Type              | Can an item (string)<br>be null? | Can the list<br>be null? | Description|
+|------------------|----------------------------------|--------------------------|------------|
+| `List<String>`   | No      | No      | A non-null list that contains<br> non-null strings       |
+| `List<String>?`  | No      | {{yes}} | A non-null list that contains<br> strings that **might be null** |
```
lrhn, line 107: The description is wrong. It should be swapped with the next line.

src/null-safety/index.md
```diff
@@ -1,39 +1,247 @@
...
+| `List<String?>`  | {{yes}} | No      | A list that **might be null** and that<br> contains non-null strings |
+| `List<String?>?` | {{yes}} | {{yes}} | A list that **might be null** and that<br> contains strings that **might be null**<br> (same as `List<String>` if you <br>don't opt into null safety) |
```
lrhn, line 109: The "same as" comment is confusing. It mixes old syntax with new syntax and it's not clear what is meant. 
Can be read as saying that "List<String?>?` is a non-nullable list of non-nullable strings if you don't opt into null safety. (That is, read as: "same as `List<String>` above if ..."). 

Just drop the parenthesized comment.

src/null-safety/index.md
```diff
@@ -1,39 +1,247 @@
...
+{:.table .table-striped}
+
+When literals initialize a list or set,
```
lrhn, line 112: initialize -> create. 
Or: If a list or set is created by a literal ...

src/null-safety/index.md
```diff
@@ -1,39 +1,247 @@
...
+then instead of seeing a type like in the table above,
+you typically see a type annotation on the literal.
+For example, here’s what a `List<String?>` (called `nameList`) and
+a `Set<String?>` (called `nameSet`) might look like:
```
lrhn, line 116: -> here's what variables containing a ....

The `nameList` variable is not a `List<String?>`. It's (current) value *is* a `List<String?`>, and its type is `List<String?>`. When *reading* a variable, it's sometimes OK to talk about it as if the variable is its value, but here we are declaring the variables, not reading them.

src/null-safety/index.md
```diff
@@ -1,39 +1,247 @@
...
+
+```dart
+var nameList = <String?>['Andrew', 'Anjan', 'Anya'];
+var nameSet = <String?>{'Andrew', 'Anjan', 'Anya'};
+```
+
+### Map types {#map-types}
+
+Map types behave mostly like you’d expect, with one exception: **the returned value can be null**. Null is the value for a key that isn't present in the map.
```
lrhn, line 125: -> ... the returned value of a lookup can ....

src/null-safety/index.md
```diff
@@ -1,39 +1,247 @@
...
+As an example, look at the following code. What do you think the value and type of `uhOh` are?
+
+```dart
+var myMap = <String, int>{'one': 1};
+var uhOh = myMap['two'];
+```
+
+The answer is that `uhOh` is null and has type `int?`. 
+
+Like lists and sets, maps can have a variety of type permutations.
```
lrhn, line 136: type permutations -> types

Not sure what is being *permuted* here. And "variety of type variants" sounds odd :)

src/null-safety/index.md
```diff
@@ -1,39 +1,247 @@
...
+int value = aList['one'] ?? 0;
+```
+
+[`if` statement]: /guides/language/language-tour#if-and-else
+[`??` operator]: /guides/language/language-tour#conditional-expressions
+
+
+## Fixing analysis errors
+
+With sound null safety, the Dart analyzer generates errors when
```
lrhn, line 192: Isn't "sound null safety" the phrase we use for the sound run-time mode.
Here it seems to be used just for "null safety", so maybe remove the "sound".

src/null-safety/index.md
```diff
@@ -1,39 +1,247 @@
...
+so it can't predict the value of a global variable or class field.
+
+If you know a value can't be null
+but it has a nullable type,
+consider using `!`, `late`, or `as`.
+Examples of using `!` and `late` are in the [first section of this page][].
+Here's an example of using `as` to convert an `int?` to a `int`:
+
+```dart
+return maybeInt(3) as int;
```
lrhn, line 208: Don't do (or recommend) using `as` like that. Always use `!` 
... unless you actually want to check for a different type, like `maybeNum() as int`.

src/null-safety/index.md
```diff
@@ -1,39 +1,247 @@
...
+description: Information about Dart's upcoming null safety feature
 ---
 
-Watch this page for updates about null safety.
-In the meantime, check out these resources:
+Sound null safety is coming to the Dart language!
+Once you opt into null safety,
+types are non-nullable by default, meaning that
+values in your code can’t be null _unless you say they can be._
+With null safety, your **runtime** null-reference exceptions
```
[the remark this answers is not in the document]
kwalrath: /cc @mit-mit & @filiph

src/null-safety/index.md
```diff
@@ -1,39 +1,247 @@
...
+```
+
+If the variable _can_ have the value `null`,
+**add  `?`** to its type declaration:
+
+```dart
+int? aNullableInt = null;
+```
+
+When using a value that you know isn’t null but that has a nullable type,
```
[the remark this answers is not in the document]
kwalrath: I did have it there once, but thought it might obscure the larger introductory clause.

src/null-safety/index.md
```diff
@@ -1,39 +1,247 @@
...
+**add `!`** to tell the Dart analyzer that the value isn’t null:
+
+```dart
+int? a = 2;
+int b = a!; // The ! temporarily changes `a` from int? to int.
```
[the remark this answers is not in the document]
kwalrath: Oh interesting!

src/null-safety/index.md
kwalrath: Interesting... I commented this out because we knew it was a complicated subject, and we'd witnessed people using `late final` prematurely.

src/null-safety/index.md
kwalrath: I can remove it from the table, but it seems worth pointing out somewhere that the old `List<String>` would be expressed as `List<String?>?`. (Could be in a footnote.) Or do people not need to know/understand this?

src/null-safety/index.md
```diff
@@ -1,39 +1,247 @@
...
-description: Watch this space for news about null safety
-toc: false
+description: Information about Dart's upcoming null safety feature
 ---
 
-Watch this page for updates about null safety.
-In the meantime, check out these resources:
+Sound null safety is coming to the Dart language!
+Once you opt into null safety,
+types are non-nullable by default, meaning that
```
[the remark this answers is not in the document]
kwalrath: I'm using "in your code" and removing the "in your code" from "values", later in the sentence.

src/null-safety/index.md
kwalrath: Yikes! Thanks for catching that. Btw, I've decided to swap the boolean columns so the bold explanation corresponds with the bold **yes**es. I think I got it right this time.

src/null-safety/index.md
lrhn: It's only partly true.

The old `List<String>` works like the new `List<String?>?` when assigning to it, but like the new `List<String>` when reading from it or calling methods. 

If you have `List<String?>? list; print(list.length);`, it's not valid. It would be for `List<String>` (both new and old interpretations).

I'm not against saying something, but it needs to be very, very clear when syntax is interpreted in the old way instead of the new way. I'm not sure what it's saying is actually worth the necessary effort.

src/null-safety/index.md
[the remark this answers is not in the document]
lrhn: You're the expert!

kwalrath:
I've updated the staging server. This could be good to go.

kwalrath:
Thanks for the careful review, @lrhn! I learn something every time you review my work.

src/null-safety/index.md
```diff
@@ -1,39 +1,257 @@
...
+[give us feedback.]: https://github.com/dart-lang/sdk/issues/new?title=Null%20safety%20feedback:%20[issue%20summary]&labels=NNBD&body=Describe%20the%20issue%20or%20potential%20improvement%20in%20detail%20here
+
+If you've used Kotlin, TypeScript, or C#,
+then some of the null-safety syntax might look familiar.
+That's by design: the Dart language aims to be unsurprising.
+
+
+## Creating and using variables (?, !, late)
+
+Use `?`, `!`, and `late` to tweak how the Dart analyzer treats your code.
```
mit-mit, line 32: 'tweak' sounds a bit odd here; these are not 'fine adjustments'. And it's not just for the analyzer, it's a fundamental type change which also effects compilers, etc.

src/null-safety/index.md
```diff
@@ -1,39 +1,257 @@
...
+
+If the variable _can_ have the value `null`,
+**add `?`** to its type declaration:
+
+```dart
+int? aNullableInt = null;
+```
+
+When using a value that you know isn’t null but that has a nullable type,
+**add `!`** to tell the Dart analyzer that the value isn’t null:
```
mit-mit, line 49: Again, it's not just the Dart analyzer.

src/null-safety/index.md
```diff
@@ -1,39 +1,257 @@
...
+int b = a!; // `a!` is an int (or throws if `a` is null).
+```
+
+If you know that a non-nullable variable will be
+initialized to a non-null value before it's used,
+but the Dart analyzer thinks otherwise,
+**insert `late`** before the variable's type:
+
+```dart
+late int aRealInt;
```
mit-mit, line 62: Can we show an example that includes the initialization? For example, show an example of setting the variable inside a constructor?

src/null-safety/index.md
mit-mit: Also, should we recommend doing a null check instead? Some guidance on when to use the (potentially bad) bang operator would be good.

src/null-safety/index.md
```diff
@@ -1,39 +1,257 @@
...
 
-Watch this page for updates about null safety.
-In the meantime, check out these resources:
+Sound null safety is coming to the Dart language!
+When you opt into null safety,
+types in your code are non-nullable by default, meaning that
+values can’t be null _unless you say they can be._
+With null safety, your **runtime** null-dereference errors
+turn into **edit-time** analysis errors.
+The compiler can optimize away internal null checks,
```
mit-mit, line 12: Can we elaborate a bit and say something like "Because null safety in Dart is sound, the compiler can..."

src/null-safety/index.md
```diff
@@ -1,39 +1,257 @@
...
+{{ site.alert.tip }}
+  Avoid using `late final` unless you really need it.
+  For example, prefer using an [initializer list][]
+  to initialize instance variables.
+{{ site.alert.end }}
+
+[initializer list]: /guides/language/language-tour#initializer-list
+{% endcomment %}
+
+
```
mit-mit, line 77: Consider mentioning the "null aware operator" (`?.`):

```dart
  double? d;  
  print(d?.floor());
```

Developers will see it mentioned in error messages:

```dart
  double? d;  
  print(d.floor());
```
=>

```
 Error: Method 'floor' cannot be called on 'double?' because it is potentially null.
Try calling using ?. instead.
  print(d.floor());
          ^^^^^
```

src/null-safety/index.md
```diff
@@ -1,39 +1,257 @@
...
+### List and set types {#list-and-set-types}
+
+When you’re declaring the type of a list or set,
+think about what can be null.
+The following table shows the possibilities for a list of strings
+if you opt into null safety.
+
+{% assign yes = '<b>Yes</b>' %}
+
+|------------------+----------------------------------+--------------------------+------------|
```
mit-mit, line 107: This table is great!

src/null-safety/index.md
```diff
@@ -1,39 +1,257 @@
...
+|Type              | Can the list<br>be null? | Can an item (string)<br>be null? | Description|
+|------------------|----------------------------------|--------------------------|------------|
+| `List<String>`   | No      | No      | A non-null list that contains<br> non-null strings |
+| `List<String>?`  | {{yes}} | No      | A list that **might be null** and that<br> contains non-null strings |
+| `List<String?>`  | No      | {{yes}} |  A non-null list that contains<br> strings that **might be null** |
+| `List<String?>?` | {{yes}} | {{yes}} | A list that **might be null** and that<br> contains strings that **might be null** |
+{:.table .table-striped}
+
+When a literal creates a list or set,
+then instead of seeing a type like in the table above,
```
mit-mit, line 117: Change 'seeing' to 'using'?

src/null-safety/index.md
mit-mit: For comparison, see how the Kotlin docs treats their `!` (which is `!!`) operator here: https://kotlinlang.org/docs/reference/null-safety.html

lrhn: I actually think the situation where you *know* that the value is non-null, but the type system can't see that (say, you know that when `state` is `running`, the `runningTask` is non-null, and you just checked `state`) is *exactly* the place to use `!`.

src/null-safety/index.md
lrhn: Agree. Something like "... to tell the compiler how your code handles null values."

src/null-safety/index.md
mit-mit: Sure, but here it's very early in our docs, and we've not described the other ways yet (like actually testing for null and relying on flow analysis), so someone new to null safety might get the idea that `!` is needed for any access on a nullable variable...

src/null-safety/index.md
mit-mit: Or maybe something like:

Use `?`, `!`, and `late` to specify the nullability of a variable. Because null safety in Dart is sound, this effects the behaviour of the Dart analyzer, and the Dart compilers & runtimes.

kwalrath: I'm using mit's first sentence but not the second. (If we want something like the second sentence, it should go in the intro to this page.)

src/null-safety/index.md
kwalrath: I've added a note and example about explicitly checking for null if you aren't sure about the value.

src/null-safety/index.md
kwalrath: OK, since this isn't really an analysis error, I've put this in a new section called **Other common code changes.**

src/null-safety/index.md
kwalrath: I deleted the verb :)

kwalrath:
OK, maybe good to go? I've updated the staged site:

https://dartlang-org-staging-1.firebaseapp.com/null-safety

src/null-safety/index.md
mit-mit: @kwalrath this wasn't addressed?

src/null-safety/index.md
mit-mit: I would like to entirely do away with the section header "Fixing analysis errors". That titles sounds like the developer is in some battle with the analyzer, which hopefully really isn't the case :-). I really like the content in the section, just not the title. Maybe something like "Working with nullable variable safely"?

Also, I think that section needs to be moved up and be right after "Creating and using" (it's all another form of using), and it's stuff that is far more important that developers understand early than the details of list, sets, and maps.

src/null-safety/index.md
```diff
@@ -1,41 +1,288 @@
...
+Because null safety in Dart is sound,
+the compiler can optimize away internal null checks,
+enabling apps to be faster, smaller, and more reliable.
 
-* [Null-safe playground: nullsafety.dartpad.dev][nullsafety.dartpad.dev]
+You can practice using null safety in the web app
+[DartPad with Null Safety.][nullsafety.dartpad.dev]
+
+{{ site.alert.important }}
+  **Don't start migrating large projects yet.**
```
mit-mit, line 20: This sounds a bit like it's OK to migrate smaller projects. We really don't want even small packages migrated and published yet.

Maybe change to something like "Don't migrate production code, don't publish any migrated code."

src/null-safety/index.md
mit-mit: Kathy, for an example of what I mean by this being part of "Creating and using" try to ready through the blog post draft, which has this content together with declaring.

src/null-safety/index.md
```diff
@@ -1,41 +1,288 @@
...
+if the operand might be null.
+Instead, you can use the null-aware version of that operator (`?.`):
+
+```dart
+double? d;  
+print(d?.floor()); // Uses `?.` instead of `.` to invoke `floor()`.
+```
+
+
+## Where to learn more
```
mit-mit, line 271: I think we need some details on how to actually run with null safety, either in the text above, or in a link here.

src/null-safety/index.md
mit-mit: Sorry, my mistake. I was reading the wrong code snippet!

src/null-safety/index.md
mit-mit: Or maybe slightly more nuanced: It's totally OK to migrate code (of any size), you should just not do it to your main master code base (like you could do it to a copy, or do it and then roll back).

src/null-safety/index.md
mit-mit: I really like your idea of expanding on this in the intro. Maybe something like:

## Sound null safety

When you use null safety to specify the nullability of your variables, the Dart analyzer will enforce that your code is safe. For example that you don't read a nullable variable without checking for null. You will see potential issues listed next to other warnings & errors from the Dart analyzer.

[screenshot of analyzer output?]

Further, unlike some other languages, null safety in Dart is sound: This means that Dart is 100% sure that a variable isn't null if the nullability specification says so. As a result, the Dart compilers & runtimes can produce optimized code which is smaller and faster.

src/null-safety/index.md
mit-mit: We could consider just linking to this? https://github.com/dart-lang/samples/blob/master/null_safety/calculate_lix/README.md

src/null-safety/index.md
kwalrath: This had some overlap with what we already had, so I've modified it heavily. ptal.

mit-mit: From a marketing PoV I'd really like to have the 'Sound null safety' heading. WDYT?

src/null-safety/index.md
```diff
@@ -1,41 +1,314 @@
...
+
+When creating a variable,
+you can use `?` and `late`
+to inform Dart of the variable's nullability.
+
+Here’s an example of declaring a **non-nullable integer variable**
+(assuming you’ve opted into null safety):
+
+```dart
+int aRealInt; // Can't be used until it's initialized to an integer.
```
mit-mit, line 56: I'm in doubt about this one. Given it's the very first example of a non-nullable variable on the page, should we not show what I think is the more regular case: initializing it with a value?

```dart
int aRealInt = 42; // Can't be assigned null.
```

mit-mit:
Getting very close, just a few comments!

src/null-safety/index.md
kwalrath: If it's assigned a value, wouldn't you normally use `var` instead?

mit-mit: Yup. The blog post has these examples:

```dart
// In null-safe Dart, none of these can ever be null.
var i = 42;
final b = Foo();
String m = '';
```

kwalrath:
OK, I think this PR is ready, and I've updated the staging server:

https://dartlang-org-staging-1.firebaseapp.com/null-safety

PR #2462, opened 2020-06-02
Roll Dart SDK version used for analysis

mit-mit:
I noticed on https://dart.dev/tools/sdk that we're still using 2.6.0


kwalrath:
Good catch! I need to make sure we have a process to update the SDK version.

kwalrath:
It's a little weird that the updated page says it's using 2.8.0, though. Did we even release 2.8.0?

PR #2467, opened 2020-06-04
Add a page about experiment flags

kwalrath:
Staged:
https://kw-staging-dartlang-2.firebaseapp.com/tools/experiment-flags
https://kw-staging-dartlang-2.firebaseapp.com/guides/language/analysis-options#the-analysis-options-file

/cc @kevmoo


kevmoo:
We should add a note in here about use `pub run --enable-experiment=xyz ...` too!

kevmoo:
This will be available in the next dev build

mit-mit:
> We should add a note in here about use pub run --enable-experiment=xyz ... too!

Isn't that already covered as part of 'Using experiment flags with command-line tools' ?

kwalrath:
@kevmoo I'm going to merge this, but we could add a version note that `pub run` will support this as of the dev release 2.9.0-13.0.dev.

Issue #2474, opened 2020-06-07
'Debugging Dart web apps' page issue


01abhishekjain:
Page URL: https://dart.dev/web/debugging.html
Page source: https://github.com/dart-lang/site-www/tree/master/src/web/debugging.md

Found a typo? You can fix it yourself by going to the page source and clicking the pencil icon. Or finish creating this issue.

Description of issue:
This document should also mention how to automatically break at start of user script, similar to node's `--inspect-brk` flag

kwalrath:
@grouma can you tell us how you'd automatically break? thanks!

PR #2475, opened 2020-06-07
Include --pause-isolates-on-start in Dart DevTools

01abhishekjain:
I think it would be helpful for people to know about this important and handy option. It took me some time to dig it up, and it would have helped if it was there upfront.

### DIFF
--- a/src/tools/dart-devtools.md
+++ b/src/tools/dart-devtools.md
@@ -147,7 +147,9 @@ Serving DevTools at http://127.0.0.1.9100
 ### 3. Start the target app
 
 Use the `dart --observe` command to execute the main file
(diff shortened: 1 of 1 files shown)


kwalrath:
I think this answers my question in #2474. :)

src/tools/dart-devtools.md
```diff
@@ -147,7 +147,7 @@ Serving DevTools at http://127.0.0.1.9100
 ### 3. Start the target app
 
 Use the `dart --observe` command to execute the main file
-for the Dart command-line app that you want to debug or observe:
+for the Dart command-line app that you want to debug or observe. Optionally, you can also include the `--pause-isolates-on-start` option which will automatically break the execution at start of the script:
```
kwalrath, line 150:
```suggestion
for the Dart command-line app that you want to debug or observe.
Optionally add `--pause-isolates-on-start`,
which automatically breaks execution at the start of the script.
```

PR #2477, opened 2020-06-10
Update /null-safety, centralize CSS for embedded DartPads

kwalrath:
This PR builds on #2457. 

Staged:
https://kw-staging-dartlang-2.firebaseapp.com/null-safety
https://kw-staging-dartlang-2.firebaseapp.com/codelabs/async-await


src/_assets/css/main.scss
```diff
@@ -1379,5 +1379,13 @@ blockquote {
   padding: 10px 20px 0 20px;
 }
 
+iframe[src*="dartpad.dev"] {
+  border: 1px solid #ccc;
+  margin-bottom: 1rem;
+  min-height: 150px;
```
johnpryan, line 1385: I think this might be a little too small. 250px seems to be a better height for most snippets.

When I try the null safety codelab, I lose visibility of the console output when I uncomment the line:

Screen Shot 2020-06-09 at 8 26 08 PM

mit-mit:
When I initially load the page I see this:
Screenshot 2020-06-10 at 12 34 28

If I then shift-reload, then I get the full UI:
Screenshot 2020-06-10 at 12 35 20

kwalrath:
fyi @munificent I'd started to work on real docs, but the PR has stalled.

kwalrath:
The CSS centralization should go into a separate PR.

PR #2481, opened 2020-06-10
Updating examples for "DON'T use `async` when it has no useful effect."

lucyget:
Updating examples for "DON'T use `async` when it has no useful effect." to follow "DO use `Future<void>` as the return type of asynchronous members that do not produce values."


kwalrath:
Thank you, @lucyget!

@munificent, if you like these changes, @lucyget or I can figure out why there are branch conflicts, and I can make sure the build works.

kwalrath:
Paging @munificent...

munificent:
Sorry for the delay! Yes, these changes LGTM.

kwalrath:
I think the merge issue had something to do with the prettify tags: `{% prettify dart tag=pre+code %}` Strangely, Travis doesn't seem to be trying to build this, but I think it'll commit ok. I'll do a little local test and then merge.

PR #2486, opened 2020-06-15
Rename type system page

mit-mit:
Using the term 'sound' here is getting a bit confusing, especially with sound nulls safety


kwalrath:
Did you mean for /codelabs/null-safety.md to be in this PR?

mit-mit:
Oh, good catch, I did not. Removed now.

src/_data/side-nav.yml
```diff
@@ -12,6 +12,8 @@
           permalink: /codelabs
         - title: Language cheatsheet
           permalink: /codelabs/dart-cheatsheet
+        - title: Null safety
```
kwalrath, line 15: Oops, remove this too!

PR #2500, opened 2020-06-29
Fix the build

kwalrath:
Followup to #2481

### DIFF
--- a/examples/misc/lib/effective_dart/usage_good.dart
+++ b/examples/misc/lib/effective_dart/usage_good.dart
@@ -179,13 +179,13 @@ void miscDeclAnalyzedButNotTested() {
   }
 
   {
-    // ignore_for_file: only_throw_errors
+    // ignore_for_file: only_throw_errors,
     // #docregion async
-    Future<void> usesAwait(Future<void> later) async {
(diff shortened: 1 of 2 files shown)


examples/misc/lib/effective_dart/usage_good.dart
```diff
@@ -180,7 +181,7 @@ void miscDeclAnalyzedButNotTested() {
   {
     // ignore_for_file: only_throw_errors
     // #docregion async
-    Future<void> usesAwait(Future<void> later) async {
+    Future<void> usesAwait(Future<Object> later) async {
```
johnpryan, line 184: nit: I don't think `<Object>` is required here, this could be changed to `Future later`.

kwalrath: @munificent what do you think? Clearly, `Future <void>` was the wrong type to use. :)

(I'm going to merge now, so the build is fixed, but I can easily change this later.)

munificent: Yeah, `void` is definitely wrong. `String` would probably be clear and simple since the result gets passed to `print()`.

PR #2508, opened 2020-07-01
New banner

kwalrath:
Staged: https://kw-www-dartlang-1.firebaseapp.com/

Should be similar to other banners, like: https://developers.google.com/assistant

### DIFF
--- a/src/_assets/css/_dash.scss
+++ b/src/_assets/css/_dash.scss
@@ -334,6 +334,7 @@ $dash-dartpad-border: #293542;
     a {
       color: white;
       text-decoration: underline;
+      font-weight: 700;
     }
   }
 
--- a/src/_assets/css/_dash.scss
+++ b/src/_assets/css/_dash.scss
@@ -328,12 +328,23 @@ $dash-dartpad-border: #293542;
 .banner {
   background-color: $dash_callout;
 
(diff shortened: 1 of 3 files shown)


src/_assets/css/_dash.scss
```diff
@@ -328,12 +328,24 @@ $dash-dartpad-border: #293542;
...
 
-  .banner__text {
+  .bw_banner__text {
     color: white;
+
+    a {
+      color: white;
+      text-decoration: underline;
+      font-weight: 700;
+    }
```
chalin, line 338: Great to see this banner!

Btw, defining a new class **breaks the formatting** a bit (notice that there's a bottom margin on the banner, the font isn't quite right, etc):

> image

You'll want to keep the `.banner__text` name and instead, say, introduce a class (like `a.bw`) below `.banner__text` so that it looks like this:

> image

chalin: Or do something similar to the Flutter website: define a `.banner__text--bw` style to use as _an additional_ class on the banner div paragraph.

kwalrath: Thank you! I saw the problem but assumed I'd just never noticed it before.

Issue #2412, opened 2020-05-16
Add "What's new" page


kwalrath:
Maybe start with a "What's new on this site" section to dart.dev/docs, like flutter.dev/docs.

Then we can add a page like https://flutter.dev/docs/whats-new-archive.

This came up in the course of Effective Dart changes: #2404, #2405.

nnhhiilliisstt:
can i work on this issue?

kwalrath:
Thanks for volunteering, @nnhhiilliisstt! You're welcome to submit a PR, but be warned, it might be fiddly work because we'll want to set a good precedent for future changes.

nnhhiilliisstt:
will try my best

PR #2513, opened 2020-07-08
Document a CHANGELOG.md format for use on pub.dev

jonasfj:
@kwalrath I'm not sure the explanations needs to be this long... But embedding an example would be very nice.

---------

New upcoming design of pub.dev will "parse" the changelog. And display
the section headers for versions differently. In the future more tooling
might parse or read changelogs. So this is an attempt to document a
format for changelogs that can be easily parsed.

This does not provide a formal specification of all CHANGELOG.md formats
accepted by pub.dev, but it should provide sufficient documentation that
package authors can write CHANGELOG.md files that will be parsed
correctly on pub.dev.

### DIFF
--- a/src/tools/pub/package-layout.md
+++ b/src/tools/pub/package-layout.md
@@ -128,12 +128,45 @@ enchilada/
   CHANGELOG.md
 {% endprettify %}
 
-To show users the latest changes to your package, you can include a changelog
-file where you can write a short note about the changes in your latest
-release. When you upload your package to the
-[pub.dev site,]({{site.pub}}) your package's `CHANGELOG.md` file (if any)
-appears in the changelog tab, rendered as [Markdown.][Markdown]
+To help users of your package upgrade between versions of your package, you
+should include a `CHANGELOG.md` file. This file should contain a section for
+each release with notes about changes.
 
+When you upload your package to the [pub.dev site,]({{site.pub}})
+your package's `CHANGELOG.md` file (if any)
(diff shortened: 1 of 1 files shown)


kwalrath:
Good idea! I proposed some changes, so please take a look.

We should also add this to the "What's new" page that I keep threatening to create: #2412

kwalrath:
Staged: https://kw-www-dartlang-1.firebaseapp.com/tools/pub/package-layout#changelogmd

Note that I changed the subheading to `##` instead of `###` because skipping a level is considered bad for accessibility. Got to set a good example!

src/tools/pub/package-layout.md
```diff
@@ -128,11 +128,51 @@ enchilada/
...
+As the example shows, you can add subsections.
+
+{% prettify markdown tag=pre+code %}
+# 1.0.1
+
+* Fixed missing exclamation mark in `sayHi()` method.
+
+# 1.0.0
+
+* **Breaking change:** Removed deprecated `sayHello()` method.
```
sigurdm, line 161: Perhaps we should move the example of a breaking change from v1.0.0 to v2.0.0.
I think there is a strong tendency for v1.0.0 to be a 'graduation' of a 0.x.y version without further change.

kwalrath: Although 0.1.1 did deprecate sayHello...

kwalrath:
I'm going to commit this, since it's better than what we have now. We can iterate...

PR #2531, opened 2020-07-17
Add link to Mustache in web library overview

jmewes:


jmewes:
@googlebot I signed it!

jmewes:
When doing my first steps with Dart I was looking for a simple templating tool but couldn't find it. Just now I learned that "pub.dev" is build with the "mustache" package which is kind of what I was looking for (see [pub-dev/../listing.dart#L33](https://github.com/dart-lang/pub-dev/blob/8b51d32f9e0e29a83d5d46180ca66cef79127591/app/lib/frontend/templates/listing.dart#L33)). So a link to the "mustache" library might be useful for others as well, to see that also such an approach is possible.

kwalrath:
Thanks for the PR! The one thing that scares me about mustache is that it hasn't been updated for over a year. Maybe someone like @theacodes or @filiph could weigh in?

jmewes:
> The one thing that scares me about mustache is that it hasn't been updated for over a year.

I had the same concern. But then I thought that, "If 'pub.dev' relies on the library and there is a problem with it, there will be enough of pressure to get it fixed or replaced."

See https://github.com/dart-lang/pub-dev/blob/master/app/pubspec.yaml#L29

theacodes:
Some libraries get to a state of maturity and don't really update often. Especially something like Mustache which is a port of an already well established, mature project.

PR #2532, opened 2020-07-18
Add "Understanding null safety" doc.

munificent:
I turned the "Null safety" link on the navigation into a menu with the
old short doc under "Overview". I figure we can eventually add
"Migration" under there too.

I kept this as one giant doc instead of splitting into separate pages.
The latter felt weird since it left no clear place for the introduction
and summary. It's still shorter than the language tour page. :)

The images will need some work to decide how they should be scaled. I
drew them in Google Docs (because I am a masochist) then exported them
as SVG and rasterized in Photoshop. What's the normal process for images
on the site?


kwalrath:
Oh, and the normal process for images used to be that we'd use Omnigraffle (one advantage: you can change text relatively easily), but draw.io (being renamed to diagrams.net) seems like the cool new thing and it produces SVG. (We have an internal instance of draw.io that you can search for.)

munificent:
Hi, @sfshaza2! Since @kwalrath is on vacation, do you want to review it in her absence? I know it's huge. Sorry. :-/

src/null-safety/understanding-null-safety/index.md
```diff
@@ -0,0 +1,1410 @@
...
+
+```dart
+bool isEmpty(String string) => string.length == 0;
+
+main() {
+  isEmpty(null);
+}
+```
+
+If you run this Dart program today, it will throw a `NoSuchMethodError`
```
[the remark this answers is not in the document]
munificent: Done.

src/null-safety/understanding-null-safety/index.md
```diff
@@ -0,0 +1,1410 @@
...
+exception on the call to `.length`. The `null` value is an instance of the
+`Null` class, and `Null` has no "length" getter. Runtime failures suck. This is
```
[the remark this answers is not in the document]
munificent: Let's let her advocate for herself when she returns then. :)

src/null-safety/understanding-null-safety/index.md
```diff
@@ -0,0 +1,1410 @@
...
+especially true in a language like Dart that is designed to run on an end-user's
+device. If a server application fails, you can often restart it before anyone
+notices. But when a Flutter app crashes on a user's phone, they are not happy.
+When you're users aren't happy, you aren't happy.
```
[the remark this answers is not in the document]
munificent: Done.

src/null-safety/understanding-null-safety/index.md
```diff
@@ -0,0 +1,1410 @@
...
+
+There is no one true solution to this problem. Rust and Kotlin both have their
+own approach that makes sense in the context of those languages. This doc walks
+through all the details of our answer for Dart, which we call "null safety". It
+includes a set of changes to the static type system and also a constellation of
+other changes and new language features to let you not only write null-safe code
+but hopefully to *enjoy* doing so.
+
+This document is long. If you want something shorter that covers just what you
+need to know to get up and running, start with the [overview][overview]. When
```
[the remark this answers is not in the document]
munificent: Done.

src/null-safety/understanding-null-safety/index.md
```diff
@@ -0,0 +1,1410 @@
...
+you are ready to for a deeper understanding and have the time, come back here so
```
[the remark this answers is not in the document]
munificent: Done.

src/null-safety/understanding-null-safety/index.md
```diff
@@ -0,0 +1,1410 @@
...
+you can understand deeply *how* the language handles `null`, *why* we designed
+it that way, and how to write idiomatic, modern, null-safe Dart. (Spoiler alert:
+it ends up being surprisingly close to how you write Dart today.)
+
+As we said before, there are several ways a language can tackle null reference
+errors, each of which have their pros and cons. We had a few principles that
```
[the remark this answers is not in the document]
munificent: Even better: just "These".

src/null-safety/understanding-null-safety/index.md
```diff
@@ -0,0 +1,1410 @@
...
+guided the choices we made:
+
+*   **Code should be safe by default.** If you write new Dart code and don't use
+    any explicitly unsafe features, it will never throw a null reference error
+    at runtime. All possible null reference errors are caught statically. If you
+    want to defer some of that checking to runtime to get greater flexibility,
+    you can, but you have to choose that by using some feature that is textually
```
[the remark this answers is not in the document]
munificent: That's what I had in mind, yeah.

src/null-safety/understanding-null-safety/index.md
```diff
@@ -0,0 +1,1410 @@
...
+    In other words, we aren't giving you a life jacket and leaving it up to you
+    to remember to put it on every time you go out on the water. Instead, we
+    give you a boat that doesn't sink. Unless you jump overboard, you'll stay
+    dry.
+
+*   **Null safe code should be easy to write.** Most existing Dart code is
+    dynamically correct and does not throw null reference errors. You like your
+    Dart program the way it looks now, and we want you to be able to keep
+    writing code that way. Safety shouldn't require sacrificing usability,
+    paying penance to the type checker, or having to significantly change the
```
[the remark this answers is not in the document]
munificent: I do my best. :)

src/null-safety/understanding-null-safety/index.md
```diff
@@ -0,0 +1,1410 @@
...
+Getting rid of implicit downcasts and removing `Null` as a bottom type covers
+all of the main places that types flow through a program across assigments and
+from arguments into parameters on function calls. The main remaining places
+where `null` can sneak in are when a variable first comes into being and when
+you leave a function. So there are some additional compile errors:
+
+### Invalid returns
+
+If a function has a non-nullable return type, then every path through the
+function must reach a `return` statement that returns a value. Dart before null
```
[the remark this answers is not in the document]
munificent: Done.

src/null-safety/understanding-null-safety/index.md
```diff
@@ -0,0 +1,1410 @@
...
+before:
+
+```dart
+List<int> filterEvens(List<int> ints) {
+  return ints.where((n) => n.isEven);
+}
+```
+
+Spot the bug? The `.where()` method is lazy, so it returns an `Iterable`, not a
+`List`. This program will compile but then throw an exception at runtime when it
```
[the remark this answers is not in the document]
munificent: Done.

src/null-safety/understanding-null-safety/index.md
```diff
@@ -0,0 +1,1410 @@
...
+
+There is a region of non-nullable types. Those types let you access all of the
+interesting methods, but can never ever contain `null`. And then there is a
+parallel family of all of the corresponding nullable types. Those permit `null`,
+but you can't do much with them. We let values flow from the non-nullable side
+to the nullable side because doing so is safe, but not the other direction.
+
+That seems like nullable types are basically useless. They have no methods and
+you can't get away from them. Don't worry, we have a whole suite of features to
+learn to help you move values from the nullable half over to the other side
```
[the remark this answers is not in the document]
munificent: Yeah, good idea.

src/null-safety/understanding-null-safety/index.md
```diff
@@ -0,0 +1,1410 @@
...
+
+*   If you want to indicate that you allow a value of any type, use `Object?`
+    instead of `Object`. In fact, it becomes pretty rare to use `Object` since
+    that type means "could be any possible value except this one weirdly
+    prohibted value `null`".
+
+*   On the rare times you need a bottom type, use `Never` instead of `Null`.
+    If you don't know if you need it, you probably don't.
+
+## Ensuring Correctness
```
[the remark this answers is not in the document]
munificent: Done. (Fixed all the other headers too.)

src/null-safety/understanding-null-safety/index.md
```diff
@@ -0,0 +1,1410 @@
...
+    return object.isEmpty;
+  } else {
+    return false;
+  }
+}
+```
+
+[control flow analysis]: https://en.wikipedia.org/wiki/Control_flow_analysis
+
+Note how on the marked line, we can call `isEmpty` on `object`. That method is
```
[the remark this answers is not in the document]
munificent: Oops, marked!

src/null-safety/understanding-null-safety/index.md
```diff
@@ -0,0 +1,1410 @@
...
+defined on `List`, not `Object`. This works because the type checker looks at
+all of the `is` expressions and the control flow paths in the program. If the
+body of some control flow construct only executes when a certain `is` expression
+on a variable is true, then inside that body the variable's type is "promoted"
+to the tested type.
+
+In the example here, the then block of the `if` statement only runs when
```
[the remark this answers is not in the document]
munificent: It was correct but awkward. Changed to "the then branch". Is that better?

src/null-safety/understanding-null-safety/index.md
```diff
@@ -0,0 +1,1410 @@
...
+`object` actually contains a list. Therefore, Dart promotes `object` to type
+`List` instead of its declared type `Object`. This is a handy feature, but it's
+pretty limited. The following functionally identical program does not work:
+
+```dart
```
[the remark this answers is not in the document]
munificent: Added a marker on the erroneous line.

src/null-safety/understanding-null-safety/index.md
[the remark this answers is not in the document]
leafpetersen: There's a minor quibble that could be made here about whether dynamic invocations count as textually visible.

munificent:
> I wonder how we should adjust the overview page (/null-safety) to point to your page and to have a clear division between what's in each page.

Yes, I had a clear narrative in mind so I just started writing and didn't think about how to integrate with the existing page. (I actually didn't know we had one...) Oops.

>  I can't help but say... all the headings should use sentence case. (Ducking and running now!)

I got it *mostly* right this time so... progress?

> Where is the source for all of these new diagrams? We generally like to save the source in case we have to modify the images down the line.

I drew them in Docs (I have a high pain tolerance). I can email you and co. the links. Or if you'd rather me redo them in OmniGraffle or draw.io (?), I'm happy to do that too. Just let me know.

src/null-safety/understanding-null-safety/index.md
```diff
@@ -0,0 +1,1410 @@
...
+```
+
+This program should run without error. Using `as T` accomplishes that. Using
+`!` would throw an exception.
+
+Other generic types have some bound that restricts the kinds of type arguments
+that can be applied:
+
+```dart
+Rect<T extends num> {
```
leafpetersen, line 1223: `class Rect`?

src/null-safety/understanding-null-safety/index.md
```diff
@@ -0,0 +1,1410 @@
...
+fields and variables uninitialized. But you can call any methods on values of
+the type parameter type that are declared on its bound. Having a non-nullable
+bound does, however, prevent *users* of your generic class from instantiating it
+with a nullable type argument. That's probably a reasonable limitation for most
+classes.
+
+You can also use a nullable *bound*:
+
+```dart
+Rect<T extends num?> {
```
leafpetersen, line 1243: `class Rect`?

src/null-safety/understanding-null-safety/index.md
```diff
@@ -0,0 +1,1410 @@
...
+Other generic types have some bound that restricts the kinds of type arguments
+that can be applied:
+
+```dart
+Rect<T extends num> {
+  final T left, top, right, bottom;
+
+  Rect(this.left, this.top, this.right, this.bottom);
+
+  T get area => (right - left) * (bottom - top);
```
leafpetersen, line 1228: Will this actually work with out a downcast with Lasse's current proposal?

src/null-safety/understanding-null-safety/index.md
```diff
@@ -0,0 +1,1410 @@
...
+
+```dart
+Rect<T extends num?> {
+  final T left, top, right, bottom;
+
+  T get area => (right - left) * (bottom - top);
+}
+```
+
+This means that in the body of the class you get the flexibility of treating the
```
leafpetersen, line 1250: It's a little confusing here whether you intend the code to be invalid or not.  As written, the definition of `area` is invalid, so maybe call that out in some way?

src/null-safety/understanding-null-safety/index.md
```diff
@@ -0,0 +1,1410 @@
...
+you can also set it to a *longer* length to pad the list with uninitialized
+elements.
+
+If you were to do that with a list of a non-nullable type, you'd violate
+soundness when you later accessed those unwritten elements. To prevent that, the
+`length` setter will throw a runtime exception if (and only if) the list has a
+non-nullable element type *and* you set it to a *longer* length. It is still
+fine to truncate lists of all types, and you can grow lists of nullable types.
+
+There is an important consequence of this if you define your own list types that
```
leafpetersen, line 1348: This feels a little oddly stated.  The core issue is that you have to implement `add` yourself, because the version in ListMixin has to use the length setter.  I think `insert` actually may be fixed to not use `add` now?  cc @natebosch was this fixed?

natebosch: No, we have to use `add` for null safe code. Part of the migration does need to be to implement `add` if you are using `ListMixin`.

src/null-safety/understanding-null-safety/index.md
```diff
@@ -0,0 +1,1410 @@
...
+<img src="understanding-null-safety/top-and-bottom.png" width="50%">
+
+In practice, this means:
+
+*   If you want to indicate that you allow a value of any type, use `Object?`
+    instead of `Object`. In fact, it becomes pretty rare to use `Object` since
+    that type means "could be any possible value except this one weirdly
+    prohibted value `null`".
+
+*   On the rare times you need a bottom type, use `Never` instead of `Null`.
```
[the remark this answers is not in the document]
sfshaza2: agree, "On the rare occasion" is good.

src/null-safety/understanding-null-safety/index.md
```diff
@@ -0,0 +1,1410 @@
...
+would still be nullable. Instead, we've added a new bottom type named `Never`:
+
+<img src="understanding-null-safety/top-and-bottom.png" width="50%">
+
+In practice, this means:
+
+*   If you want to indicate that you allow a value of any type, use `Object?`
+    instead of `Object`. In fact, it becomes pretty rare to use `Object` since
+    that type means "could be any possible value except this one weirdly
+    prohibted value `null`".
```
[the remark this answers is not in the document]
sfshaza2: Oops, missed that one!

src/null-safety/understanding-null-safety/index.md
```diff
@@ -0,0 +1,1410 @@
...
+
+  // Constructor and hashCode...
+}
+```
+
+This program analyzes without error. Notice that the last line of the `==`
+method accesses `.x` and `.y` on `other`. It has been promoted to `Point` even
+though the function doesn't have any `return` or `throw`. The control flow
+analysis knows that the declared type of `wrongType()` is `Never` which means
+the then body of the `if` statement *must* abort somehow. Since the second
```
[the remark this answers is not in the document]
sfshaza2: OOOOOHH, now i get it. Yes, please code font.  :)

src/null-safety/understanding-null-safety/index.md
```diff
@@ -0,0 +1,1410 @@
...
+subtype of it. Dart has no *named* top type. If you need a top type, you want
+`Object?`. Likewise, `Null` is no longer the bottom type. If it was, everything
+would still be nullable. Instead, we've added a new bottom type named `Never`:
+
+<img src="understanding-null-safety/top-and-bottom.png" width="50%">
+
+In practice, this means:
+
+*   If you want to indicate that you allow a value of any type, use `Object?`
+    instead of `Object`. In fact, it becomes pretty rare to use `Object` since
```
natebosch, line 320: Do we have data on the relative usage of `Object?` and `Object` ? I'm suspicious that `Object` is less rare than indicated here.

src/null-safety/understanding-null-safety/index.md
```diff
@@ -0,0 +1,1410 @@
...
+    that type means "could be any possible value except this one weirdly
+    prohibted value `null`".
+
+*   On the rare times you need a bottom type, use `Never` instead of `Null`.
+    If you don't know if you need it, you probably don't.
```
natebosch, line 325: Which is the "it" here? `Never` or `Null`?

src/null-safety/understanding-null-safety/index.md
```diff
@@ -0,0 +1,1410 @@
...
+      int atDeclaration = 0;
+      int initializingFormal;
+      int initializationList;
+
+      SomeClass(this.initializingFormal)
+          : initializationList = 0;
+    }
+    ```
+
+    In other words, as long as the field has a value by the time you reach the
```
natebosch, line 420: [nit] I'm worried that "by the time you reach" might still get misinterpreted as allowing initialization in the constructor body. I think "before you reach" is slightly more clear.

src/null-safety/understanding-null-safety/index.md
```diff
@@ -0,0 +1,1410 @@
...
+    instead of `Object`. In fact, it becomes pretty rare to use `Object` since
+    that type means "could be any possible value except this one weirdly
+    prohibted value `null`".
+
+*   On the rare times you need a bottom type, use `Never` instead of `Null`.
+    If you don't know if you need it, you probably don't.
+
+## Ensuring correctness
+
+We bifurcated the universe of types into nullable and non-nullable halves. In
```
filiph, line 329: [nit] I like the word "bifurcate" as much as anyone, but I wonder if it's gratuitous language here. Unless _bifurcate_ is a term Dart developers should know, or an important term they might _want_ to know, I'd go with something simpler.

```suggestion
We split the universe of types into nullable and non-nullable halves. In
```

src/null-safety/understanding-null-safety/index.md
```diff
@@ -0,0 +1,1410 @@
...
+String missingReturn() {
+  // No return.
+}
+```
+
+If you analyze this, you get a gentle *hint* that *maybe* you forgot a return,
+but if not, no big deal. That's because if execution reaches the end of a
+function body then Dart implicitly returns `null`. Since every type is nullable,
+*technically* this function is safe, even though it's probably not what you
+want.
```
filiph, line 356: I wonder if it makes sense to use past tense in paragraphs like this. People scan documents, or they just don't get the meaning of a sentence above, and they might get the impression that the behavior explained here is the current one. It might help to put all the "before null safety" in past tense, and all the "after" as present tense.

This will also make the document more long-lived as-is. I think we expect most readers to read this when null safety is either almost here, or completely here.

src/null-safety/understanding-null-safety/index.md
```diff
@@ -0,0 +1,1410 @@
...
+
+### Uninitialized variables
+
+When you declare a variable, if you don't give it an explicit initializer, Dart
+default initializes the variable with `null`. That's convenient, but obviously
+totally unsafe if the variable's type is non-nullable. So we have to tighten
+things up for non-nullable variables:
+
+*   **Top level variable and static field declarations must have an
+    initializer.** Since these are lazy and can be accessed and assigned from
```
filiph, line 391: [nit] If there's a place to link to from "lazy", I'd do it. Not everyone will know what you mean.

src/null-safety/understanding-null-safety/index.md
```diff
@@ -0,0 +1,1410 @@
...
+
+      SomeClass(this.initializingFormal)
+          : initializationList = 0;
+    }
+    ```
+
+    In other words, as long as the field has a value by the time you reach the
+    constructor body, you're good.
+
+*   Local variables are the fun one. A non-nullable local variable *doesn't*
```
filiph, line 423: [nit] "are a fun one" reads to me like "the semantics are kind of crazy here". This is almost the opposite of what you're trying to say, I assume. 

```suggestion
*   With local variables, we can be a lot smarter. A non-nullable local variable *doesn't*
```

src/null-safety/understanding-null-safety/index.md
```diff
@@ -0,0 +1,1410 @@
...
+*   **Optional parameters must have a default value.** If you don't pass an
+    argument for an optional positional or named parameter, then the language
+    fills it in with the default value. If you don't specify a default value,
+    the default default value is `null`, and that doesn't fly if the parameter's
+    type is non-nullable.
+
+    So, if you want a parameter to be optional, you need to either make it
+    nullable or specify a valid non-`null` default value.
+
+These restrictions sound onerous, but in they aren't too bad in practice. They
```
filiph, line 454:
```suggestion
These restrictions sound onerous, but they aren't too bad in practice. They
```

src/null-safety/understanding-null-safety/index.md
```diff
@@ -0,0 +1,1410 @@
...
+  print(definitelyString.length);
+}
+
+main() {
+  String? maybeString = null; // Or not!
+  requireStringNotNull(maybeString);
+}
+```
+
+This program is not safe and we shouldn't allow it. However, Dart has this thing
```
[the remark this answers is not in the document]
munificent: I don't want to change the tense of a bunch of paragraphs. Instead, changed this to "However, Dart has always had
this thing...". I think that hints that the behavior will change without having to put it all in the past tense.

src/null-safety/understanding-null-safety/index.md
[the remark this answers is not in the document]
munificent: Good call.

src/null-safety/understanding-null-safety/index.md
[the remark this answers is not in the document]
munificent: Done.

src/null-safety/understanding-null-safety/index.md
```diff
@@ -0,0 +1,1410 @@
...
+defined on `List`, not `Object`. This works because the type checker looks at
+all of the `is` expressions and the control flow paths in the program. If the
+body of some control flow construct only executes when a certain `is` expression
+on a variable is true, then inside that body the variable's type is "promoted"
+to the tested type.
+
+In the example here, the then branch of the `if` statement only runs when
+`object` actually contains a list. Therefore, Dart promotes `object` to type
+`List` instead of its declared type `Object`. This is a handy feature, but it's
+pretty limited. The following functionally identical program does not work:
```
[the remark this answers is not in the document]
munificent: Done.

src/null-safety/understanding-null-safety/index.md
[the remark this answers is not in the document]
munificent: Changed to "then branch" like the other places use. Putting it in code font seems weird to me since "then" isn't a keyword in Dart. I could hyphenate like "then-branch" if you think that helps. A couple of textbooks I have do that to refer to programming language constructs.

src/null-safety/understanding-null-safety/index.md
```diff
@@ -0,0 +1,1410 @@
...
+promotion. The general goal is that if the code is dynamically correct and it's
+reasonable to figure that out statically, the analysis should be clever enough
+to do so.
+
+### Unnecessary code warnings
+
+Having smarter reachability analysis and knowing where `null` can flow through
+your program helps ensure that you *add* code to handle `null`. But we can also
+use that same analysis to detect code that you *don't* need. Before null safety,
+if you write something like:
```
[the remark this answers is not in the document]
munificent: Done.

src/null-safety/understanding-null-safety/index.md
munificent: You mean we aren't making `class` optional? :) Done.

src/null-safety/understanding-null-safety/index.md
munificent: Hmm, not sure. Replaced this class entirely with a simpler example. PTAL.

src/null-safety/understanding-null-safety/index.md
munificent: Oops, I just totally forgot to update the code. Replaced it with a better example.

src/null-safety/understanding-null-safety/index.md
munificent: Reworded it some.

src/null-safety/understanding-null-safety/index.md
munificent: No official data. This is based on my experience migrating the core libraries and tests. Most places that were previously using `Object` ended up getting changed to `Object?`. The only exceptions I can think of offhand are code dealing with errors/exceptions because the language specifically prohibits throwing `null`, and overrides of `==` because, again, the language prohibits `null`.

src/null-safety/understanding-null-safety/index.md
munificent: Bottom type. Clarified.

src/null-safety/understanding-null-safety/index.md
munificent: Good point. Done.

src/null-safety/understanding-null-safety/index.md
munificent: Aww. Fine. How about "divided"?

src/null-safety/understanding-null-safety/index.md
munificent: Done.

src/null-safety/understanding-null-safety/index.md
munificent: Changed to "most flexible case".

src/null-safety/understanding-null-safety/index.md
natebosch: I wouldn't consider the core libraries and tests to be particularly representative of the ecosystem.

I'm worried about sending an unintentionally prescriptive message with this.

src/null-safety/understanding-null-safety/index.md
munificent: Yeah, tense and temporality is hard. I'm trying to balance both where the language is at the point in time that a reader reads this, and where they are as they make progress through the document. Also, large passages in convoluted tenses can be painful to read. ("Large passages that had been being in convoluted tenses will have been painful to have been reading.")

You're right this paragraph reads better as past tense. Done. Tweaked some other places too.

src/null-safety/understanding-null-safety/index.md
munificent: Actually, laziness doesn't come into play anyway. Deleted.

src/null-safety/understanding-null-safety/index.md
```diff
@@ -0,0 +1,1410 @@
...
+    and older legacy code from before Dart supported null safety. In these
+    "mixed-mode" programs, null reference errors may still occur. During the
+    process of migrating your code to null safety, you get all of the static
+    safety benefits in the portions of your program that are migrated. But you
+    don't get the full runtime soundness until you're fully null safe.
+
+Note that *eliminating* `null` is not a goal. There's nothing wrong with `null`.
+On the contrary, it's really useful to be able to represent the *absence* of a
+value. Building support for a special "absent" value directly into the language
+lets us making working with absence flexible and usable. It underpins optional
```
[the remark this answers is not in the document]
munificent: Fixed.

munificent:
@sfshaza2, I think you're the only one who can give an approving review. Are you ready for me to merge this?

sfshaza2:
@munificent, do you want to add a link to the new page from /null-safety, or should I?

munificent:
> @munificent, do you want to add a link to the new page from /null-safety, or should I?

If you wouldn't mind doing that, I would definitely appreciate it. :)

PR #2535, opened 2020-07-20
Revise guidance on using "dynamic".

munificent:
These days, we no longer suggest "dynamic" for cases where the type
system can't express a type. Instead, "dynamic" really only comes into
play when you want dynamic dispatch.

### DIFF
--- a/src/_guides/language/effective-dart/design.md
+++ b/src/_guides/language/effective-dart/design.md
@@ -1533,7 +1533,7 @@ The new syntax is a little more verbose, but is consistent with other locations
 where you must use the new syntax.
 
 
-### AVOID using `dynamic` unless you want dynamic dispatch.
+### AVOID using `dynamic` unless you want to disable static checking.
 
 Some operations work with any possible object. For example, a `log()` method
 could take any object and call `toString()` on it. Two types in Dart permit all
--- a/src/_guides/language/effective-dart/design.md
+++ b/src/_guides/language/effective-dart/design.md
@@ -1565,9 +1565,9 @@ bool convertToBool(Object arg) {
 
 The main exception to this rule is when working with existing APIs that use
 `dynamic`, especially inside a generic type. For example, JSON objects have type
-`Map<String, dynamic>` and your code will thus need to accept that same type.
-Even so, when using a value from one of these APIs, it's often a good idea to
-cast it to a more precise type before accessing members.
+`Map<String, dynamic>` and your code will need to accept that same type. Even
+so, when using a value from one of these APIs, it's often a good idea to cast it
+to a more precise type before accessing members.
 
 
 ### DO use `Future<void>` as the return type of asynchronous members that do not produce values.
--- a/src/_guides/language/effective-dart/design.md
+++ b/src/_guides/language/effective-dart/design.md
@@ -1533,7 +1533,7 @@ The new syntax is a little more verbose, but is consistent with other locations
 where you must use the new syntax.
 
 
-### DO annotate with `Object` instead of `dynamic` to indicate any object is allowed.
+### AVOID using `dynamic` unless you want dynamic dispatch.
 
 Some operations work with any possible object. For example, a `log()` method
 could take any object and call `toString()` on it. Two types in Dart permit all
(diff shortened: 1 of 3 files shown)


src/_guides/language/effective-dart/design.md
```diff
@@ -1533,36 +1533,42 @@ The new syntax is a little more verbose, but is consistent with other locations
...
 simply want to state that you allow all objects, use `Object`, as you would in
 Java or C#.
 
-Using `dynamic` sends a more complex signal. It may mean that Dart's type system
-isn't sophisticated enough to represent the set of types that are allowed, or
-that the values are coming from interop or otherwise outside of the purview of
-the static type system, or that you explicitly want runtime dynamism at that
-point in the program.
+The type `dynamic` not only accepts all objects, but it also permits all
+*operations*. Any member access on a value of type `dynamic` is allowed at
```
chalin, line 1545: Is "operations" the best term here? Maybe "methods" (or, as the following sentence suggests: "members")?

src/_guides/language/effective-dart/design.md
```diff
@@ -1533,36 +1533,42 @@ The new syntax is a little more verbose, but is consistent with other locations
 where you must use the new syntax.
 
 
-### DO annotate with `Object` instead of `dynamic` to indicate any object is allowed.
+### AVOID using `dynamic` unless you want dynamic dispatch.
```
chalin, line 1536: How about, more simply:

> ### AVOID using `dynamic`

---

I'm not sure that "dynamic dispatch" is appropriate here. By nature of Dart being OO, dynamic dispatching happens "all the time" for all types; it isn't specific to `dynamic`.

If you really want an extra qualifier for the title how about:

> ### AVOID using `dynamic` unless you want to disable static checks

(or "... static checking")

src/_guides/language/effective-dart/design.md
```diff
@@ -1533,36 +1533,42 @@ The new syntax is a little more verbose, but is consistent with other locations
...
   if (arg is bool) return arg;
-  if (arg is String) return arg == 'true';
+  if (arg is String) return arg.toLowerCase() == 'true';
   throw ArgumentError('Cannot convert $arg to a bool.');
 }
 {% endprettify %}
 
+The main exception to this rule is when working with existing APIs that use
+`dynamic`, especially inside a generic type. For example, JSON objects have type
+`Map<String, dynamic>` and your code will thus need to accept that same type.
```
chalin, line 1568: Consider dropping "thus".

src/_guides/language/effective-dart/design.md
munificent: "Methods" is overly precise. "Members" is pretty good, but reads as awkwardly redundant with the next sentence to me. Also, though the paragraph doesn't spell this out, using `dynamic` permits you to pass the value as an argument to a function expecting any type, so there are other "operations" that are allowed outside of member access.

chalin: Ok, sold (on the use of "operations").

src/_guides/language/effective-dart/design.md
munificent: > I'm not sure that "dynamic dispatch" is appropriate here. By nature of Dart being OO, dynamic dispatching happens "all the time" for all types; it isn't specific to `dynamic`.

This is one of those corners where I wish we could more easily rely on everyone have the same shared lingo. I personally make a distinction between *dynamic* and *virtual* dispatch. The former means looking up a member entirely by name with no reference to any static type the value may or may not have. It's what you get in JS, Ruby, Python, etc. The latter allows polymorphism and methods to be overridden, but still relies on the static type of the value to determine whether a method call is allowed.

But this is probably something I can't rely on readers understanding. I like your suggestion. Done.

src/_guides/language/effective-dart/design.md

src/_guides/language/effective-dart/design.md
chalin: > This is one of those corners where I wish we could more easily rely on everyone have the same shared lingo. 

I agree.

> I personally make a distinction between dynamic and virtual dispatch. 

Nice distinction.

chalin:
Travis is green. @sfshaza2 or @munificent - you'll have to merge, I don't have the permissions anymore.

PR #2536, opened 2020-07-20
Add more two-letter acronym examples.

### LINKED ISSUE #2511: Include example of concatenated two-letter acronyms
Under "DO capitalize acronyms and abbreviations longer than two letters like words.", it would be good to show an example of how an identifier formed from sequential acronyms is cased when those acronyms are and are not two letters long. Something like:

```
// Good:
DBIOPort
TVVcr

// Bad:
DbIOPort
TvVcr
```

munificent:
Fix #2511.

### DIFF
--- a/src/_guides/language/effective-dart/style.md
+++ b/src/_guides/language/effective-dart/style.md
@@ -84,7 +84,7 @@ class C { ... }
 {% include linter-rule.html rule="camel_case_extensions" %}
 
 Like types, extensions should capitalize the first letter of each word
-(including the first word), 
+(including the first word),
 and use no separators.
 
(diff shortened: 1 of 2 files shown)


src/_guides/language/effective-dart/style.md
```diff
@@ -264,7 +265,8 @@ UiHandler
 IoStream
 HTTPRequest
 ID
-Db
+DbIOPort
```
chalin, line 268: Would you want this "bad" example to be `DbIoPort` instead?

munificent: Done.

PR #2557, opened 2020-08-04
Assert: throw error rather than exception

suragch:
Given 

- the difference in Dart of Exception vs Error
- and `assert` throws an `AssertionError`

It seems more clear to say "throw an error" rather than "throw an exception" when used in the context of `assert`. 

However, I may be mistaken about the meaning of "throw an exception" if that phrase is broad enough to include errors. If so, though, that's not very clear in the documentation.

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -307,7 +307,7 @@ assert(lineCount == null);
 
 {{site.alert.note}}
   Production code ignores the `assert()` call. During development, on the other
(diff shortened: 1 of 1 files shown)


suragch:
Hmm, actually when an assert fails the message is "Exception has occurred". So either the issue is bigger than the documentation or exceptions actually do include errors. Either way this pull request should probably be rejected.

kwalrath:
Yes, errors are just a kind of exception. The terminology isn't the most obvious, but it's fairly standard (from at least Java days).

PR #2560, opened 2020-08-05
Update Flutter version that has dart CLI tool

mit-mit:
This was postponed out a release

### DIFF
--- a/src/get-dart.md
+++ b/src/get-dart.md
@@ -11,14 +11,14 @@ The Dart SDK has the libraries and command-line tools that you need to develop
 Dart command-line, server, and non-Flutter web apps.
 For details, see the [Dart SDK overview](/tools/sdk).
 
-**As of Flutter 1.20, the [Flutter SDK][flutter] includes the Dart SDK.**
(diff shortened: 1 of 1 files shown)


src/get-dart.md
```diff
@@ -11,14 +11,14 @@ The Dart SDK has the libraries and command-line tools that you need to develop
 Dart command-line, server, and non-Flutter web apps.
 For details, see the [Dart SDK overview](/tools/sdk).
 
-**As of Flutter 1.20, the [Flutter SDK][flutter] includes the Dart SDK.**
+**As of Flutter 1.21, the [Flutter SDK][flutter] includes the Dart SDK.**
```
kwalrath, line 14: This should remain 1.20, right?

mit-mit: Nope, Flutter 1.20 uses Dart 2.9 which doesn't have it (it did temporarily, but 1.20 stable doesn't)

kwalrath:
These changes were made as part of a larger PR (#2562), so I'm going to close this PR.

PR #2563, opened 2020-08-05
Tiny typo fix

kwalrath:
/cc @gspencergoog (follow-up to #2548)


gspencergoog:
LOL, just goes to illustrate that making a change always runs the risk of introducing another bug.
Thanks for catching that.

PR #2568, opened 2020-08-06
Make in-progress spec easier to find

### LINKED ISSUE #2567: Make it easier to find the PDF version of the in-progress language spec
Per https://github.com/dart-lang/sdk/issues/42924#issuecomment-670098243.

kwalrath:
Fixes #2567

Staged: https://kw-www-dartlang-1.firebaseapp.com/guides/language/spec

### DIFF
--- a/src/_guides/language/spec.md
+++ b/src/_guides/language/spec.md
@@ -12,7 +12,7 @@ For a gentler introduction to Dart, see the
 
 The Dart 2 language specification is available in PDF format:
 
-  * [Formal specification (version 2.2)][formal spec]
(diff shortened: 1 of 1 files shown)


mit-mit:
Looks good!

@eernstg is there a Dart language/SDK version that the specification version 2.2 corresponds to? (Is that actually Dart 2.2?)

eernstg:
I believe language versions were introduced much later than 2.2, so we probably can't tell precisely what "language version 2.2" should mean, but I think it's reasonable to call it 'Dart 2.2'.

PR #2570, opened 2020-08-11
Upgrade to jekyll 3.9

johnpryan:
Upgrades to kramdown 2.3.0, which patches CVE-2020-14001

This change runs `bundle update`. When I ran `bundle list` I see that
kramdown 2.3.0 is now installed, (although the Gemfile.lock says
kramdown (>= 1.17, < 3))


johnpryan:
I applied the same fix to the Flutter site here: https://github.com/flutter/website/pull/4466

johnpryan:
Staged: https://jr-dartlang-site.firebaseapp.com/

kwalrath:
It looks like code excerpts don't work. :( E.g., look at the bottom of the homepage and the top of https://jr-dartlang-site.firebaseapp.com/guides/language/language-tour.

I think Patrice used to generate diffs (with the footer fixed to avoid unnecessary diffs due to date changes) to catch this sort of thing.

johnpryan:
Still looking at this... It's not clear to me yet if _plugins/markdown_with_code_excerpts.rb is throwing an exception or not running at all.

johnpryan:
@kwalrath I think this is working now. I haven't fully checked all of the snippets, but I've parsed through the Language Tour and Effective Dart and don't see any issues.

staged: 

- https://jr-dartlang-site.firebaseapp.com/
- https://jr-dartlang-site.firebaseapp.com/guides/language/language-tour
- https://jr-dartlang-site.firebaseapp.com/guides/language/effective-dart/usage

kwalrath:
Does the embedded DartPad on the homepage work for you? I'm still seeing code-excerpt instead of code, but I seem to recall that caching can be a problem (even though I cleared the browser cache).

image

kwalrath:
Everything else is looking great, so far, including red & green backgrounds, and highlighting.

johnpryan:
good catch, yeah that needs to be fixed.

johnpryan:
#2124 is broken after upgrading to 3.9. We could either revert that change (stop testing these snippets), leave the site on Jekyll 3.8.6, or figure out some other way to test the snippets.

kwalrath:
Oy, testing the snippets recently found a problem: https://github.com/dart-lang/site-www/pull/2575/files#diff-5516df53c13434e15d4ca7bd4fb9cc8b. Although I wonder if you knew about it but wanted to keep the code simple, since it doesn't cause a visible error.

How about we NOT auto-include the snippets for now, but file a bug to reenable auto-including them (or at least syncing the code so the latest version is always tested).

johnpryan:
Sounds good to me

johnpryan:
Deployed: https://jr-dartlang-site.firebaseapp.com

PR #2576, opened 2020-08-13
Fix Flutter version number

theUncanny:
In this page says "As of Flutter 1.21, the Flutter SDK includes the Dart SDK." but the correct version it's `1.2.1` (just a typo, a missing dot/period).
- Source: https://flutter.dev/docs/development/tools/sdk/release-notes/release-notes-1.2.1#dart

### DIFF
--- a/src/get-dart.md
+++ b/src/get-dart.md
@@ -11,7 +11,7 @@ The Dart SDK has the libraries and command-line tools that you need to develop
 Dart command-line, server, and non-Flutter web apps.
 For details, see the [Dart SDK overview](/tools/sdk).
 
-**As of Flutter 1.21, the [Flutter SDK][flutter] includes the Dart SDK.**
+**As of Flutter 1.2.1, the [Flutter SDK][flutter] includes the Dart SDK.**
 So if you have Flutter installed,
 you might not need to explicitly download the Dart SDK.
 Consider downloading the Dart SDK if
--- a/src/get-dart.md
+++ b/src/get-dart.md
@@ -18,7 +18,7 @@ Consider downloading the Dart SDK if
 any of the following are true:
 
 * You don't use Flutter.
-* You use a pre-1.21 version of Flutter.
+* You use a pre-1.2.1 version of Flutter.
 * You want to reduce disk space requirements or download time,
   and your use case doesn't require Flutter.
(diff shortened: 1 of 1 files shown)


domesticmouse:
I suspect the confusion here is that Flutter has always included a cut down version of the Dart SDK, the library is mainly written in Dart after all. However, up until Flutter 1.21 the Dart SDK included in Flutter has always been a partial copy. The Dart SDK shipping with Flutter 1.21 will be the full thing, which is what this document is referring to.

/cc @kwalrath to correct me if I flubbed it. =)

kwalrath:
You got it right, @domesticmouse.

src/get-dart.md
```diff
@@ -11,14 +11,14 @@ The Dart SDK has the libraries and command-line tools that you need to develop
 Dart command-line, server, and non-Flutter web apps.
 For details, see the [Dart SDK overview](/tools/sdk).
 
-**As of Flutter 1.21, the [Flutter SDK][flutter] includes the Dart SDK.**
+**As of Flutter 1.2.1, the [Flutter SDK][flutter] includes the Dart SDK.**
```
domesticmouse, line 14: s/includes the Dart SDK/includes the full Dart SDK/

@kwalwrath thoughts?

theUncanny:
> I suspect the confusion here is that Flutter has always included a cut down version of the Dart SDK, the library is mainly written in Dart after all. However, up until Flutter 1.21 the Dart SDK included in Flutter has always been a partial copy. The Dart SDK shipping with Flutter 1.21 will be the full thing, which is what this document is referring to.

> I was wondering if this would be confusing. :/ I'm not sure how to fix it, but 1.21 (in the future) is the release we mean.

@domesticmouse Thank you for the clarification. I thought it was a typo since version 1.21 has not yet been released (on the stable channel) and the quoted phrase seemed to imply that it was something "current" not future (immediate).

Perhaps the best thing to do would be to close this pull request, leave it as it is and wait for the release of Flutter version 1.21 in the stable channel (especially if it is going to be released in a short time)?

kwalrath:
"full SDK" is probably the right wording to use, @domesticmouse.

I'll close this PR, @theUncanny, but thank you for letting us know that it was confusing!

PR #2584, opened 2020-08-17
New use cases for Factory Constructor

### LINKED ISSUE #2202: Mention initializing final variables using factory constructors
https://dart.dev/guides/language/language-tour#factory-constructors

About only time I have used factory constructor is to have a class with final fields, but need some logic in the constructor to initialize them that cannot be expressed in initializer list.

The current documentation for factory constructor mention that it's for cases when implementing a constructor that doesn’t always create a new instance of its class which doesn't cover this use case and is slightly misleading for new users (in my opinion). I had to be hinted initially to use the factory constructor to initialize finals, and also giving code review it also doesn't get through all the time

stupendoussuperpowers:
Attempt to resolve #2202

### DIFF
--- a/examples/misc/lib/language_tour/classes/logger.dart
+++ b/examples/misc/lib/language_tour/classes/logger.dart
@@ -13,7 +13,7 @@ class Logger {
     return _cache.putIfAbsent(
         name, () => Logger._internal(name));
   }
-  
+
   factory Logger.fromJson(Map<String, Object> json) {
     return Logger(json['name'].toString());
   }
@@ -30,7 +30,7 @@ void main() {
   // #docregion logger
   var logger = Logger('UI');
   logger.log('Button clicked');
-  
+
   var logMap = {'name': 'UI'};
   var loggerJson = Logger.fromJson(logMap);
   // #enddocregion logger
--- a/examples/misc/lib/language_tour/classes/logger.dart
+++ b/examples/misc/lib/language_tour/classes/logger.dart
@@ -33,7 +33,7 @@ void main() {
   var logger = Logger('UI');
   logger.log('Button clicked');
   
-  Map logMap = jsonDecode(jsonString);
+  Map<dynamic, dynamic> logMap = {'name': 'UI'};
   var loggerJson = Logger.fromJson(logMap);
   // #enddocregion logger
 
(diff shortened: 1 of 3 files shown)


kwalrath:
Nice work, @stupendoussuperpowers! I'm going to let @rkj do the technical review of the text, since they filed the original bug. Once that's done, I can help you get the build working and maybe tweak the text a bit.

kwalrath:
@rkj says this looks great to him, so I'll aim to fix the build tomorrow.

stupendoussuperpowers:
Let me know if there's anything I can help out with!

kwalrath:
I think I fixed the issues with the example (please take a look at the diffs), but the build is still failing for some reason. The issue is in two files not edited in this PR (see below).

@theacodes do you have a theory as to why this PR is failing analysis when the main build is fine?

```
   error • The argument type 'int' can't be assigned to the parameter type 'String'. • lib/library_tour/core/collections.dart:5:14 • argument_type_not_assignable
  error • A value of type 'dynamic' can't be assigned to a variable of type 'Person'. • lib/library_tour/core/hash_code.dart:24:21 • invalid_assignment
```

stupendoussuperpowers:
I checked out the diffs, kind of embarassed about that mistake in the code.

As far as the files are concerned, not sure how useful my little detective work is but:

This is the code for the file /hash_code.dart, which has documentation which implies that we are supposed to ignore the invalid_assigment error that we saw above?

```
if (other is! Person) return false;  
// ignore: stable, dev, invalid_assignment, // https://github.com/dart-lang/sdk/issues/32236  
Person person = other;  
```

As far as the collections.dart is concerned, the error is pretty straightforward, we are storing an int in a List<String>

Is there any chance that these errors popped up during the last build as well? And something changed in how these files are being loaded/deployed?

theacodes:
Build is green now, just need to update a string in an affected test.

Should be good to go now. The beta build is failing because of an unrelated analysis change, but I can send a follow-up PR to fix that.

kwalrath:
Oops, I forgot to update that test. :(

kwalrath:
Thank you, @theacodes! I was so mystified by the failure message that I missed the obvious.

@stupendoussuperpowers are you OK with my code and text changes? I've staged them:

https://kw-www-dartlang-1.firebaseapp.com/guides/language/language-tour#factory-constructors

stupendoussuperpowers:
Yes, I went over the changes and I am fine with them

theacodes:
No worries. It took me longer to find the error than it did to fix it. :)

PR #2587, opened 2020-08-18
add a redirect / shorten link for the nnbd migration tool

devoncarew:
- add a redirect / short link for the nnbd migration tool
- this will allow us to replace the long 'https://github.com/dart-lang/sdk/tree/master/pkg/nnbd_migration#providing-feedback' link (referenced from the `dart migrate` command line tool) with a shorter `dart.dev/go/null-safety-migration` link.

This idea is borrowed from the flutter.dev 'go' redirects: https://github.com/flutter/website/blob/master/firebase.json#L122.

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -126,6 +126,7 @@
       { "source": "/events/2016{,/**}", "destination": "https://events.dartlang.org/2016/summit", "type": 301 },
       { "source": "/events{,/**}", "destination": "https://events.dartlang.org", "type": 301 },
       { "source": "/flutter", "destination": "https://flutter.io", "type": 301 },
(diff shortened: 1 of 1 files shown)


kwalrath:
Works for me. The funny thing is that the final URL (/null-safety/migration) might be even shorter. But since we can easily change the go URL and we probably don't want to imply that the content is actual on dart.dev, this go link makes sense.

devoncarew:
Yeah, the link to the current markdown doc is pretty long, but even if we get a shorter one in the future I like that we can encode the 'go' link in tools and then adjust where it points to after the fact from the web site.

devoncarew:
Note that this link is show from the help of the `dart migrate` command and tool, so is currently focused on migrations using that tool.

PR #2588, opened 2020-08-19
Make Arrow syntax solution comment consistent

Martin005:


Martin005:
@googlebot I signed it!

PR #2591, opened 2020-08-22
Fixed backslash removing from code exceprts

limexp:
Backslash is not a special character in HTML, so `CGI.escapeHTML` kept it intact and passed into the frame_code function.
At the same time backslash is a special character for `sub`, so it must be escaped before using as a replacement string.

The error effect now is present in the code here https://dart.dev/codelabs/dart-cheatsheet#factory-constructors
```print('I don\'t recognize $typeName');```
became
```print('I donescaped_code#39;t recognize $typeName');```


kwalrath:
Wow, nice catch and fix! I'm going to ask one of our DPEs to review this.

johnpryan:
This looks good to me. I should have put this file in site-shared, since [this was taken from the flutter website](https://github.com/flutter/website/blame/master/src/_plugins/code_excerpt_framer.rb)

PR #2615, opened 2020-09-01
add a short redirect link for docs referenced by 'run run test' help

devoncarew:
- add a short redirect link for docs referenced by 'run run test' help

This will redirect from:

```
https://dart.dev/go/test-docs/json_reporter.md
```

to:

```
https://github.com/dart-lang/test/blob/master/pkgs/test/doc/json_reporter.md
```

Currently the `pun run test` command line help is using a goo.gl link, but:
- this isn't possible to update after the tool is shipped (the URL has been out of date twice), and
- we've used the `go` link pattern a few other places in our CLI tools.

@kwalrath @grouma @natebosch


devoncarew:
See the goo.gl reference here: https://github.com/dart-lang/test/blob/master/pkgs/test_core/lib/src/runner/configuration/reporters.dart#L47.

PR #2621, opened 2020-09-03
Try to clarify optional parameter restrictions

jamesderlin:
The Language Tour currently states:

> Optional parameters can be either named or positional, but not
> both.

That statement can be interpreted to mean that any individual
parameter cannot be both named and positional.  However, the actual
restriction is stronger: a function cannot have both optional named
and optional positional parameters (see
https://github.com/dart-lang/language/issues/1076).

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -1129,7 +1129,7 @@ is sometimes referred to as _arrow_ syntax.
 
 A function can have two types of parameters: _required_ and _optional_.
 The required parameters are listed first, followed by any optional parameters.
-Optional parameters can be _named_ or _positional_.
+Optional parameters can be either _named_ or _positional_.
 
 {{site.alert.note}}
   Some APIs — notably [Flutter][] widget constructors — use only named
@@ -1139,7 +1139,8 @@ Optional parameters can be _named_ or _positional_.
 
 ### Optional parameters
 
-Optional parameters can be either named or positional, but not both.
+A function can have optional positional parameters or optional named parameters,
+but not both.
 
 #### Named parameters
 
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -1127,20 +1127,23 @@ is sometimes referred to as _arrow_ syntax.
   there, but you can use a [conditional expression](#conditional-expressions).
 {{site.alert.end}}
 
-A function can have two types of parameters: _required_ and _optional_.
-The required parameters are listed first, followed by any optional parameters.
-Optional parameters can be either _named_ or _positional_.
+### Parameters
+
+A function can have any number of *required positional* parameters. These may be
+followed either by *optional positional* parameters or by *optional named*
+parameters.  A function can have optional positional parameters or optional
+named parameters, but not both.
+{% comment %}
+NULLSAFE: Replace "optional named" above with just "named" and add:
+
+A named parameter can be either optional or `required`.
+{% endcomment %}
 
 {{site.alert.note}}
   Some APIs — notably [Flutter][] widget constructors — use only named
   parameters, even for parameters that are mandatory. See the next section for
   details.
 {{site.alert.end}}
-
-### Optional parameters
-
-A function can have optional positional parameters or optional named parameters,
-but not both.
 
 #### Named parameters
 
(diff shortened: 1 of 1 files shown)


kwalrath (approved):
The writing LGTM, but I'll defer to Lasse as to correctness.

src/_guides/language/language-tour.md
```diff
@@ -1129,7 +1129,7 @@ is sometimes referred to as _arrow_ syntax.
 
 A function can have two types of parameters: _required_ and _optional_.
 The required parameters are listed first, followed by any optional parameters.
-Optional parameters can be _named_ or _positional_.
+Optional parameters can be either _named_ or _positional_.
```
lrhn, line 1132: This is basically all wrong for Null Safety. We can have required named parameters placed after optional named parameters.

How about:

> A function can have any number of _required positional_ parameters. 
> These may be followed either by _optional positional_ parameters, 
> or by _named_ parameters. 
> A named parameter can be either optional or `required`.

If this is describing pre-Null Safety behavior, then I'd say:

> A function can have any number of _required_ parameters. 
> These may be followed either by _optional positional_ parameters, 
> or by _optional named_ parameters.

src/_guides/language/language-tour.md
```diff
@@ -1139,7 +1139,8 @@ Optional parameters can be _named_ or _positional_.
 
 ### Optional parameters
 
-Optional parameters can be either named or positional, but not both.
+A function can have optional positional parameters or optional named parameters,
+but not both.
```
lrhn, line 1143: For null safety, we'll need to remove the "optional" before "named".

lrhn (approved):

kwalrath:
We should have a standard string to mark where we know things have to change for null safety. Maybe just NULLSAFE. Something like this:

```
{% comment %}
NULLSAFE: Remove the "optional" before "named".
{% endcomment %}
```

jamesderlin:
I've made adjustments to try to be friendlier for Null Safety when it becomes enabled by default.  Notably I've removed the "Optional parameters" subheading since optional/required and positional/named will be independent axes.  I've instead added a more general "Parameters" heading and merged part of the "Functions" section into it.

src/_guides/language/language-tour.md
```diff
@@ -1127,20 +1127,24 @@ is sometimes referred to as _arrow_ syntax.
   there, but you can use a [conditional expression](#conditional-expressions).
 {{site.alert.end}}
 
-A function can have two types of parameters: _required_ and _optional_.
-The required parameters are listed first, followed by any optional parameters.
-Optional parameters can be _named_ or _positional_.
+### Parameters
```
kwalrath, line 1130: Thanks for adding this section head! I don't know why we didn't have it before.

src/_guides/language/language-tour.md
```diff
@@ -1127,20 +1127,24 @@ is sometimes referred to as _arrow_ syntax.
...
+
+A function can have any number of *required positional* parameters. These may be
+followed either by *optional positional* parameters or by *optional named*
+parameters.  A function can have optional positional parameters or optional
+named parameters, but not both.
```
kwalrath, line 1135: I think this can be a little shorter (and also we [avoid may](https://developers.google.com/style/word-list#letter-m)). Also, we should switch the order to match the section names... and maybe it isn't necessary to say "optional" before named — especially since we don't use "Optional" in the section name? 

```suggestion
A function can have any number of *required positional* parameters. These can be
followed either by *named* parameters or by *optional positional* parameters
(but not both).
```

src/_guides/language/language-tour.md
```diff
@@ -1127,20 +1127,24 @@ is sometimes referred to as _arrow_ syntax.
...
+{% comment %}
+NULLSAFE: Replace "optional named" above with just "named" and add:
+
+A named parameter can be either optional or `required`.
```
kwalrath, line 1139: How about we add this now? Something like this.

Named parameters are optional unless they're specifically marked as required.

kwalrath (commented):
I'm wondering if we can simplify this a bit and also make some of the future changes now.

src/_guides/language/language-tour.md
jamesderlin: Done.

src/_guides/language/language-tour.md
jamesderlin: Sounds good.  I added that to the "Named parameters" section.

PR #2624, opened 2020-09-07
Use dart analyze

mit-mit:
Try to use `dart analyze` instead of `dartanalyzer`.


mit-mit:
The current CI job relies on passing multiple directories (e.g. `dartanalyzer lib test web`), which `dart analyze` doesn't support. But why do we do that? Can we just run from the root directory?

johnpryan:
The directory needs to be specified, so `dartanalyzer --no-implicit-casts` should be `dartanalyzer --no-implicit-casts .`

johnpryan:
We shouldn't have to specify the project subdirectory. @mit-mit Here's a PR into your branch: https://github.com/mit-mit/site-www/pull/1

johnpryan:
I'm trying to understand this function so we can figure out how to adjust it:

```bash
function toggleInFileAnalyzerFlags() {
  # Arguments: [disable|reenable] path path...
  local action=$1; shift;
  local mark='!'; local toggle=' '
  if [[ $action == 'disable' ]]; then
    mark=' '; toggle='!'
  fi

  find $* -name "*.dart" ! -path "**/.*" \
    -exec perl -i -pe "s{//$mark(ignore(_for_file)?: .*?\b(stable|dev)\b)}{//$toggle\$1}g" {} \;
}
```

It's called once per project in the `examples/` directory before calling `dartanalyzer`, presumably overwrite certain ignore_for_file annotations in some way. I also can't see why it's called with a second argument, but it only handles the first argument (`$1`).

johnpryan:
@mit-mit does GitHub give you the option to give maintainers edit access to this branch?

mit-mit:
@johnpryan, "Allow edits by maintainers" should already be set on this PR, I think

johnpryan:
@mit-mit @kwalrath the build is passing now. The beta and dev channels are still failing because the `dart analyze` output varies slightly:

old:

```
error • A value of type 'Object' can't be assigned to a variable of type 'String'. • lib/assignment.dart:11:14 • invalid_assignment
```
new:

```
error - A value of type 'Object' can't be assigned to a variable of type 'String' at lib/assignment.dart:11:14 - (invalid_assignment)
```

I added a tool (`tool/dart_tools/bin/update_analyzer_txt.dart`) to to regenerate the `analyzer-results.txt` files when 2.10 becomes the stable channel.

kwalrath:
> I added a tool (`tool/dart_tools/bin/update_analyzer_txt.dart`) to to regenerate the `analyzer-results.txt` files when 2.10 becomes the stable channel.

Nice! I think we can create analyzer-results-beta.txt and analyzer-results-dev.txt files to make everything pass... but of course we'll still need to update those files when the stable release goes out.

examples/analysis/analyzer-results.txt
```diff
@@ -1,4 +1,4 @@
-Analyzing analysis_options.yaml, lib...
+Analyzing analysis...
```
kwalrath, line 1: Really? "Analyzing analysis"? 😆 

Not a big deal, but maybe we could say something like `Analyzing examples/analysis...`

kwalrath (approved):
LGTM if it LGTMit

examples/analysis/analyzer-results.txt
johnpryan: Ha, this will always be the name of the directory, since `dart analyze` doesn't take an additional argument.

johnpryan:
Created a follow-up issue to fix the `beta` and `dev` channel builds: https://github.com/dart-lang/site-www/issues/2652

I'll file an SDK issue if we have any problems with changing `dartanalyzer lib web test` to just  `dartanalyzer .`

Issue #2617, opened 2020-09-01
Transition doc over to the new `dart` tool


kwalrath:
Although the existing Dart CLI tools will be around for quite a while, we need to transition over to the [new `dart` tool](https://dart.dev/tools/dart-tool):

1. Find pages that use the command line, and substitute the new `dart` tool for existing tools.
2. Convert the build and CI to use the new `dart` tool. (See #2623, #2624, and [this comment](https://github.com/dart-lang/site-www/issues/2617#issuecomment-807693733))

Separately, we'll need to transition some of the non-pub command reference pages over to new URLs (#2719). But for now, **keep using the old URLs**, and avoid editing the pages to be redirected.

Here's a list (a reordered version of the output of ``fgrep -Rl "\`\`\`terminal" src``) of the 38 pages we should look at and eventually update (there might be more that don't show up due to improper formatting): 

***Update: All known occurrences in pages are fixed! We should open new issues for anything else that turns up.***

- [x]  src/tools/pub/cmd/pub-deps.md (fixed by #3116)
- [x] src/tools/experiment-flags.md (fixed by #3118)
- ~~src/tools/pub/cmd/pub-run.md~~  (covered by #2719) 
- [x] ~~_Assigned to @KevinRohn:_ src/tools/pub/cmd/pub-outdated.md~~ (fixed by #2789)
- [x] ~~_Assigned to @KevinRohn:_ src/tools/pub/cmd/pub-upgrade.md~~ (fixed by #2825)
- [x] ~~_Assigned to @chethanprabhakar:_ src/tools/pub/cmd/pub-downgrade.md~~ (fixed by #2847)
- [x] ~~_Assigned to @Abhishek01039:_ src/tools/pub/cmd/pub-global.md~~ (fixed by #2843)
- [x] ~~src/tools/pub/cmd/pub-lish.md~~ (fixed by #2816)
- [x] ~~src/tools/pub/cmd/pub-uploader.md~~ (fixed by #2883)
- [x] ~~_Assigned to @KevinRohn:_ src/tools/pub/cmd/pub-cache.md~~ (fixed by #2825)
- [x] ~~_Assigned to @KevinRohn:_ src/tools/pub/cmd/pub-get.md~~ (fixed by #2826)
- [x] ~~src/tools/pub/troubleshoot.md~~ (fixed by #2854)
- [x] ~~src/tools/pub/publishing.md~~ (fixed by #2816)
[...]

PR #2629, opened 2020-09-10
Change pub pages to show the new dart tool

felpsisonfire:
@kwalrath Can you check this out, please?


kwalrath (commented):
We aren't ready to make visible changes yet, so I'll hold onto this PR for a while.

Did you test the commands to make sure they acted like the doc says they do?

felpsisonfire:
> We aren't ready to make visible changes yet, so I'll hold onto this PR for a while.
> 
> Did you test the commands to make sure they acted like the doc says they do?

At the end of the day, I think I still got the instructions wrong. Sorry about that, this is my first contribution ever. 
So, actually I need to test the commands and do comment the .md file with a description of the behavior and if it does match with the documentation. Right?

kwalrath:
> So, actually I need to test the commands and do comment the .md file with a description of the behavior and if it does match with the documentation. Right?

Yes, exactly! The main goal is to test the new `dart` command. A secondary (but still important) goal is preparing to update the docs with the right commands.

I think for these files, if your testing confirms that changing `pub` to `dart pub` is all that's necessary, all you need to do is to add a comment about that to each `.md` file. Maybe something like this:

```
{% comment %}
NEW DART COMMAND:
Throughout this file, everywhere you see `pub outdated`, change it to `dart pub outdated`.
Tested in dart version 2.10.0-59.0.dev (dev)
{% endcomment %}
```

If there are any surprises, then please create an issue in this repo (for dart.dev issues) or in github.com/dart-lang/sdk (for SDK/help issues), and refer to #2617 from that issue.

src/tools/pub/cmd/pub-deps.md
```diff
@@ -28,10 +28,10 @@ dependencies:
   markdown: ^0.7.2
 {% endprettify %}
 
-Here's an example of the `pub deps` output for markdown_converter:
+Here's an example of the `dart pub deps` output for markdown_converter:
 
 ```terminal
-$ pub deps
+$ datt pub deps
```
sumandipanshu, line 34: You should correct it to `dart pub deps`

src/tools/pub/cmd/pub-outdated.md
```diff
@@ -1,65 +1,69 @@
...
 an example that has several out-of-date dependencies.
 Three of the dependencies (`args`, `http`, and `path`) are direct,
 and one is transitive (`meta`).
 As the following example shows,
-`pub outdated` colorizes the output by default
+`dart pub outdated` colorizes the output by default
 when you run it on the command line.
 
-{% include_relative _pub-outdated-output.html %}
+{% include_relative _dart-pub-outdated-output.html %}
```
sumandipanshu, line 66: "_pub-outdated-ouput.html" references to a file on the relative path. So I think you should not change it to "_dart-pub-outdated-ouput.html" as there is no file with this name on the relative path

src/tools/pub/cmd/pub-deps.md
sumandipanshu, line 34: You should correct it to `dart pub deps`

src/tools/pub/cmd/pub-outdated.md
```diff
@@ -161,13 +165,13 @@ it should be possible to update this package.
...
 
-The output of `pub outdated` has four columns of version information
+The output of `dart pub outdated` has four columns of version information
 for each out-of-date dependency.
 Here is the part of the [example](#example) output
 that shows the four version columns:
 Current, Upgradable, Resolvable, and Latest.
 
-{% include_relative _pub-outdated-output-columns.html %}
+{% include_relative _dart-pub-outdated-output-columns.html %}
```
sumandipanshu, line 174: "_pub-outdated-ouput-columns.html" references to a file on the relative path. So I think you should not change it to "_dart-pub-outdated-ouput-columns.html" as there is no file with this name on the relative path

kwalrath:
I'm going to close this and unassign all but one of the pages, so the conversion task is simpler and we can divvy up the rest of the pages to whoever's willing.

PR #2637, opened 2020-09-13
docs: update for-in example

crimx:
Remove ambiguity in list index and value.

In language like JavaScript `for-in` gets list index instead. Example is updated to highlight the difference.

### DIFF
--- a/examples/misc/test/language_tour/control_flow_test.dart
+++ b/examples/misc/test/language_tour/control_flow_test.dart
@@ -29,14 +29,14 @@ void main() {
   test('collection', () {
     _test() {
       // #docregion collection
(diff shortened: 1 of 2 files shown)


crimx:
@googlebot I signed it!

kwalrath:
I didn't know that about JavaScript! Thanks for the fix, @crimx.

Will you be able to update examples/misc/test/language_tour/control_flow_test.dart? (The code is auto-included from a tested Dart file, so we need to update that file, as well.)

crimx:
@kwalrath test updated.

kwalrath (approved):
LGTM!

Issue #2631, opened 2020-09-10
Change how we use the term "element"


kwalrath:
This came up in a discussion with @johnpryan and @munificent.

"Element" is overloaded to mean "a syntactic construct that can appear inside a collection literal" and "a value that can appear in a collection object". The former is evaluated at runtime and may produce zero or more of the latter.

UI-as-code features like collection if, collection for, and spread are elements in the former sense, but not in the latter.

We should probably move away from using "element" to mean a value in a list object. We can already use "value" or "object" to refer to that and those are clearly runtime terms. So elementAt() returns a value, not an "element" (despite the name).

This affects the language tour and anywhere else that uses the term "element" (up to 34 pages, but some of those probably refer to HTML elements).

Aman9026:
@kwalrath  The writers have used "element" at two places which caught my attention: 

- In examples having "halogens", if we go with that context, elements would be possibly referring to "chemical elements"

- In examples of "list", this is ambiguous and needs to be fixed

kwalrath:
#2638 fixed this in the language tour, but we need to make a sweep through other uses.

atsansone:
`element` appears 541 results in 40 files. Should be a straightforward edit.

atsansone:
@munificent : This inconsistency exists as to how we refer to the items in collection structures. This issue wants to move toward using `value` whenever we are talking about list, map, or set contents. The [spec](https://github.com/dart-lang/language/blob/master/accepted/2.3/control-flow-collections/feature-specification.md#syntax) uses element when talking about control flow structures / literals. Do you have guidance as to which would be the best term, even if different terms should be used in different circumstances?

Current terms under consideration are `value` (used in language tour only, but also in Python), `element` (used almost everywhere else, but also in Java and Scala), or `item`.

munificent:
I'd probably use "value" when referring to the thing you can get out of a list or set, and "entry" for the key-value pairs a map contains.

I do have to say that "element" feels pretty natural to me to refer to the runtime concept of a value returned from a list, and it's a shame that I overloaded the term to mean the syntactic construct. I was never able to come up with a better name for "the kind of thing you can write inside a collection literal".

johnpryan:
I don't remember the context here, but "element" is also okay with me after reading through this issue three years later.

PR #2638, opened 2020-09-14
Replaced ambiguous "element" with value

Aman9026:
Replaced "element" with "value" where ever it was appropriate.
Also, note that "Elements" has been used at multiple places as an example for relating it to chemical "elements"
This PR replaces the ambiguous usage of "element" with list.

Fixes: #2631 (for language tour)


kwalrath:
@munificent is this what you had in mind?

munificent (approved):

kwalrath:
Thanks, @Aman9026 & @munificent!

PR #2640, opened 2020-09-15
Added redirect for fetchdata tutorial

zhesha:


kwalrath:
Turns out we had an old redirect that was bad!

kwalrath (approved):

PR #2653, opened 2020-09-22
Tweaks to Understanding null safety

kwalrath:
This PR has a few changes to make me feel OK about not editing this perfectly wonderful page to follow Google dev docs style.

* Add byline so it's OK to not follow the [Google style guidelines](https://developers.google.com/style/highlights).
* Use site conventions for external links & Dart strings.

### DIFF
--- a/src/null-safety/understanding-null-safety/index.md
+++ b/src/null-safety/understanding-null-safety/index.md
@@ -1,14 +1,17 @@
 ---
 title: Understanding null safety
-description: A deep dive into all of the changes related to null safety and how to use them.
+description: A deep dive into Dart language and library changes related to null safety.
 ---
+
+_Written by Bob Nystrom<br>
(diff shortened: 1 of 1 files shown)


sfshaza2 (approved):
All those single quotes are a thing of beauty. LGTM (But I won't merge until Bob reviews...)

munificent (approved):
I'm not morally opposed to changing it to follow the Google style if that's what you'd prefer. I naturally write in my own voice, but if it's better for users to have something more neutral, that's OK too.

kwalrath:
> I'm not morally opposed to changing it to follow the Google style if that's what you'd prefer. I naturally write in my own voice, but if it's better for users to have something more neutral, that's OK too.

I don't think the benefit (for this page) would outweigh the effort. Easier to just keep it in your voice. :)

Issue #1942, opened 2019-09-20
SCSS variable $site-color-body is used but not defined


chalin:
The `bootstrap_config` file imported from site-shared at this line:
https://github.com/dart-lang/site-www/blob/c2f2476258c5ebde96d0116369a50a64aba4b1ff/src/_assets/css/main.scss#L6

Contains uses of `$site-color-body`, such as here: 
https://github.com/dart-lang/site-shared/blob/8f783bfa10019afb023d4917f9cd8e614227bd3d/src/_assets/css/_bootstrap_config.scss#L4-L7

```scss
// Color config
$primary: $site-color-primary;
$body-color: $site-color-body;
$code-color: $site-color-body;
```

But `$site-color-body` isn't defined in `src/_assets/css/_variables.scss` (as it is under flutter/website, for example).

Issue #1710, opened 2019-06-28
Issues with contrast on the site


deakjahn:
But really? Who thought that this would be legible?

sfshaza2:
What? Where? I gather the contrast of something is insufficient..

deakjahn:
The site. Am I in the wrong place? I assumed it was about the web site eg. https://dart.dev/guides/language/language-tour

Main text area is so-so, border case, I'd use a more definite color but can be disputed. Code samples are downright pale, especially where the background matches the foreground (red on red, gray on gray). Right side menu is abysmal, both by eye and by measurement.

What I really don't understand is the basic rationale. I know this is the trend in fashion and lifestyle websites but I don't think a reference material for a programming language should follow fashion trends. Simply two different worlds—fortunately, I'd say :-)

sfshaza2:
Thanks for that clarification, @deakjahn.

johnpryan:
I made some changes [here ](https://github.com/dart-lang/site-www/pull/1574)a while back that might improve this.

sfshaza2:
Good point, @johnpryan. We should revisit this, or maybe run accessibility tools over some of the pages.

PR #2658, opened 2020-09-28
Improve site legibility

mit-mit:
Fixes https://github.com/dart-lang/site-www/issues/1710
Fixes https://github.com/dart-lang/site-www/issues/1942

Staged on https://mit-staging.web.app/

### DIFF
--- a/src/_assets/css/_code.scss
+++ b/src/_assets/css/_code.scss
@@ -2,7 +2,7 @@ $pre-bg: #F5F5F5; // TODO(chalin): use $site-color-codeblock-bg?
 
 @mixin pre-defaults {
   background-color: $pre-bg;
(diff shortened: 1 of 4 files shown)


kwalrath:
I think we need to change the section head styles, as well. E.g. if you go to https://mit-staging.web.app/guides/language/language-tour#constructors, the h3 ("Constructors") is pretty close to the body font, and the h4 ("Default constructors") is actually smaller than the body font.

Looking at the computed styles for each:
* h3: "Google Sans" 17 px 400 (same as before)
* h4:  "Google Sans" 15 px 400 (ditto)
* p: Roboto 16px 400 (was 14px 300)

kwalrath (approved):
LGTM!

PR #2664, opened 2020-09-30
Update README.md #Hackoctoberfest

GRACEMARYMATHEW:


kwalrath:
Thanks, but we want to keep the website readme as simple as possible.

PR #2667, opened 2020-10-01
Update get-started.md

Thusal06:
Fixed some typos.


Thusal06:
@googlebot I signed it!

kwalrath (commented):
Thanks for the PR, but we chose [entrypoint](https://dart.dev/tools/pub/glossary#entrypoint) on purpose. (I'm not crazy about it, but at least we're consistent.) The comma after "First" is optional and it's not confusing to omit it, so we opted not to include it.

PR #2673, opened 2020-10-02
null-safety index.md: Correct variable name

SimonInman:
The given example is for a Map, not a List. Update the variable name to aMap to reflect this.


kwalrath (approved):
Nice catch. Thanks for the PR!

PR #2681, opened 2020-10-03
Add a note with further resources on the choice of underscore rather than access modifier keywords in Dart.

### LINKED ISSUE #2671: Explain why Dart uses '_' for privacy rather than a keyword in the documentation.
Coming from C#, I was incredibly confused why Dart chose to use a prefix of `_` for private variables, rather than the typical `public`, `private`, and `protected` modifiers. After reading through https://github.com/dart-lang/sdk/issues/33383, however, I understand why Dart chose this method of visibility modification, and I don't mind it as much. However, it seems strange to me that no where in the docs does it even reference an explanation (unless I missed something, and if I did, please let me know). I think adding a section somewhere in the docs to explain the underscore notation (maybe a note in the tour?) would make people less likely to immediately discount Dart based off of this choice (I know I did, and looking at that issue, so did a lot of people).

Any opinions?

kfish610:
Closes #2671.

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -135,6 +135,10 @@ mind:
     from executing at all; a run-time error results in an
     [exception](#exceptions) being raised while the code executes.
 
+{{site.alert.note}}
+  If you're curious why Dart uses underscores instead of access modifier keywords like `public` or `private`, see [SDK issue 33383](https://github.com/dart-lang/sdk/issues/33383).
+{{site.alert.end}}
+
 
 ## Keywords
 
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -136,7 +136,9 @@ mind:
     [exception](#exceptions) being raised while the code executes.
 
 {{site.alert.note}}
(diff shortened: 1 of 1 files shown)


kfish610:
@googlebot I signed it!

src/_guides/language/language-tour.md
```diff
@@ -135,6 +135,10 @@ mind:
     from executing at all; a run-time error results in an
     [exception](#exceptions) being raised while the code executes.
 
+{{site.alert.note}}
+  If you're curious why Dart uses underscores instead of access modifier keywords like `public` or `private`, see [SDK issue 33383](https://github.com/dart-lang/sdk/issues/33383).
```
kwalrath, line 139:
```suggestion
  If you're curious why Dart uses underscores instead of
  access modifier keywords like `public` or `private`, see
  [SDK issue 33383](https://github.com/dart-lang/sdk/issues/33383).
```

kwalrath (approved):
LGTM, after breaking on phrases (to reduce line length to <80 chars and reduce possible diffs in the future).

kwalrath:
Closes #2671

PR #2689, opened 2020-10-06
Added notes on Dart SDK tools pages to inform about the new dart unified tool

### LINKED ISSUE #2678: Create notes on Dart SDK tools pages to inform about the new dart unified tool
Page URL: https://dart.dev/tools/dartanalyzer.html
Page source: https://github.com/dart-lang/site-www/tree/master/src/tools/dartanalyzer.md

According to the new Dart 2.10, is preferable to use the new `dart` unified command line tool over the standalone ones, such as `dartanalyzer`, `dartfmt`, `dart2js`, `dart2native` and `pub`.

So it would be cool to create an informative note on each one of these pages, linking to the following page:
https://dart.dev/tools/dart-tool

sumandipanshu:
Fixes #2678

### DIFF
--- a/src/tools/dart2js.md
+++ b/src/tools/dart2js.md
@@ -18,6 +18,13 @@ which performs a similar analysis but has a different implementation.
 This page tells you how to use dart2js on the command line. It also give tips
 on debugging the JavaScript that dart2js generates.
 
+{{ site.alert.version-note }}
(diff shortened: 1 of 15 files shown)


sumandipanshu:
@googlebot I signed it!

kwalrath (commented):
We aren't quite ready to add these notes yet. When we do, we'll probably start small (one page) and then copy a customized version of that message to other pages.

(But if we did end up using boilerplate that's the same, then we'd use an include.)

PR #2690, opened 2020-10-07
/docs -> /guides in topnav

kwalrath:
Not sure why we had /docs, but it results in a bunch of redirects, which seems unnecessary and bloats the CI output.

Let's see what happens when we change /docs to /guides.

### DIFF
--- a/src/_includes/navigation-main.html
+++ b/src/_includes/navigation-main.html
@@ -10,7 +10,7 @@
       If you edit this list, also edit navigation-side.html (mobile).
     {% endcomment %}
     <li class="mainnav__get-started">
-      <a href="/docs" class="nav-link
+      <a href="/guides" class="nav-link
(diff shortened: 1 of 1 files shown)


kwalrath:
Reduces output of the first job from 1698 lines to 1043 lines.
Reduces occurrences of "(200)" from 136 to 12 (which I might actually fix!).

mit-mit:
I *think* we had `/docs` to be consistent with Flutter. If we prefer that (I have no strong opinions here, but it seems a bit nicer?), then I guess the alternate fix would be to rename `/guides` to `/docs` ?

kwalrath:
> I _think_ we had `/docs` to be consistent with Flutter. If we prefer that (I have no strong opinions here, but it seems a bit nicer?), then I guess the alternate fix would be to rename `/guides` to `/docs` ?

If we were to be consistent with flutter.dev, that would require a lot more than just changing /guides to /docs. Flutter has all docs under `/docs` and, in general, has a deeper file hierarchy. For example:

https://dart.dev/codelabs vs. https://flutter.dev/docs/codelabs
https://dart.dev/tutorials vs. https://flutter.dev/docs/reference/tutorials
https://dart.dev/tools vs. https://flutter.dev/docs/development/tools
https://dart.dev/tools/sdk/archive vs. https://flutter.dev/docs/development/tools/sdk

It wouldn't be bad to be parallel... but it'd be a big job. Using /guides instead of /docs (while keeping the redirect) is a lot simpler, and would make it easier to find real issues in the CI output.

mit-mit:
Ah, good points. This PR lgtm then.

mit-mit (approved):

PR #2691, opened 2020-10-07
Language tour: rework prose related to operators

### LINKED ISSUE #425: Language tour: "Overridable operators" should be rephrased
The section title and section on [Overridable operators](https://www.dartlang.org/guides/language/language-tour#overridable-operators) needs a little rework IMHO. The qualification of operators as (only) _overridable_ (as opposed to simply "definable") is misleading.

From the [DLS](http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-408.pdf), Section 10.1.1:
> _Operators_ are instance methods with special names.
> ...
> The following names are allowed for **user-defined operators**: <, >, <=, >=,
==, -, +, /, ˜/, *, %, |, ˆ, &, <<, >>, []=, [], ˜.

Maybe use of the title "User-definable operators" would be better than the current "Overridable operators"? And it should rapidly be made clear that "_Operators_ are instance methods with special names." I.e., one cannot define an operator in any other context.

chalin:
Closes #425

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -1559,9 +1559,9 @@ assert(foo() == null);
 
 ## Operators
 
-Dart defines the operators shown in the following table.
-You can override many of these operators, as described in
-[Overridable operators](#overridable-operators).
+Dart supports the operators shown in the following table.
+You can define your own operators, as described in the section on
+[class methods](#_operators).
 
 |--------------------------+------------------------------------------------|
 |Description               | Operator                                       |
@@ -1621,9 +1621,9 @@ if (n % i == 0 && d % i == 0) ...
 ```
 
 {{site.alert.warning}}
-  For operators that work on two operands, the leftmost operand determines which
-  version of the operator is used. For example, if you have a Vector object and
-  a Point object, `aVector + aPoint` uses the Vector version of +.
+  For operators that take two operands, the leftmost operand determines which
+  function is used. For example, if you have a `Vector` object and
+  a `Point` object, then `aVector + aPoint` uses `Vector` addition (`+`).
 {{site.alert.end}}
 
 
@@ -1761,7 +1761,7 @@ you are sure that the object is of that type. Example:
 (emp as Person).firstName = 'Bob';
 ```
 
-If you aren't sure that the object is of type `T`, then use `is T` to check the 
+If you aren't sure that the object is of type `T`, then use `is T` to check the
 type before using the object.
 <?code-excerpt "misc/lib/language_tour/classes/employee.dart (emp is Person)"?>
 ```dart
@@ -2954,7 +2954,7 @@ constructor might return an instance from a cache, or it might
 return an instance of a subtype.
 Another use case for factory constructors is
 initializing a final variable using
-logic that can't be handled in the initializer list. 
+logic that can't be handled in the initializer list.
 
 
 In the following example,
@@ -3032,6 +3032,51 @@ class Point {
   }
 }
 ```
+
+#### Operators {#_operators}
+
+Operators are instance methods with special names.
+Dart allows you to define operators with the following names:
+
+`<`  | `+`  | `|`  | `[]`
+`>`  | `/`  | `^`  | `[]=`
+`<=` | `~/` | `&`  | `~`
+`>=` | `*`  | `<<` | `==`
+`–`  | `%`  | `>>`
+{:.table}
+
+{{site.alert.note}}
+  You may have noticed that some [operators](#operators), like `!=`, are not in
+  the list of names. That's because they're just syntactic sugar. For example,
+  the expression `e1 != e2` is syntactic sugar for `!(e1 == e2)`.
+{{site.alert.end}}
+
+An operator declaration is identified using the built-in identifier `operator`.
+The following example defines vector addition (`+`) and subtraction (`-`):
+
+<?code-excerpt "misc/lib/language_tour/classes/vector.dart"?>
+```dart
+class Vector {
+  final int x, y;
+
+  Vector(this.x, this.y);
+
+  Vector operator +(Vector v) => Vector(x + v.x, y + v.y);
+  Vector operator -(Vector v) => Vector(x - v.x, y - v.y);
+
+  // Operator == and hashCode not shown.
+  // ···
+}
+
+void main() {
+  final v = Vector(2, 3);
+  final w = Vector(2, 2);
+
+  assert(v + w == Vector(4, 5));
+  assert(v - w == Vector(0, 1));
+}
+```
+
 
 #### Getters and setters
 
@@ -3197,11 +3242,11 @@ class SmartTelevision [!extends!] Television {
 {% endprettify %}
 
 
-
+<a name="overridable-operators"></a>
 
 #### Overriding members
 
-Subclasses can override instance methods, getters, and setters.
+Subclasses can override instance methods (including [operators](#_operators)), getters, and setters.
 You can use the `@override` annotation to indicate that you are
 intentionally overriding a member:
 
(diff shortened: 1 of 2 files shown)


src/_guides/language/language-tour.md
```diff
@@ -3033,6 +3033,51 @@ class Point {
...
+#### Operators {#_operators}
+
+Operators are instance methods with special names.
+Dart allows you to define operators with the following names:
+
+`<`  | `+`  | `|`  | `[]`
+`>`  | `/`  | `^`  | `[]=`
+`<=` | `~/` | `&`  | `~`
+`>=` | `*`  | `<<` | `==`
+`–`  | `%`  | `>>`
```
chalin, line 3045: It might be worth double checking this list of operators against the language spec, ... just in case it has changed.

kwalrath:
Thanks, @chalin! I'll be happy to look at the words/formatting once @lrhn (or designate :) makes sure the facts are correct.

kwalrath:
FYI, I just sent email to find a technical reviewer for this PR.

kwalrath:
@eernstg and @munificent have been suggested as appropriate reviewers.

eernstg:
I'll take a look.

src/_guides/language/language-tour.md
```diff
@@ -3197,11 +3242,11 @@ class SmartTelevision [!extends!] Television {
 {% endprettify %}
 
 
-
+<a name="overridable-operators"></a>
```
eernstg, line 3245: Could this be spurious? I can't see other occurrences of `<a name="...">` in this file (having expanded the whole file here, and also searching in https://raw.githubusercontent.com/dart-lang/site-www/master/src/_guides/language/language-tour.md), and it's surprising that, if I understand this mechanism correctly, a click on 'operators' will take us to this point, showing a section called 'Overriding members'.

eernstg (approved):
LGTM, noting that I'd recommend using the old wording rather than the new one in one location, and noting that there may be a surprising target for a link with the text 'operator'.

src/_guides/language/language-tour.md
```diff
@@ -3033,6 +3033,51 @@ class Point {
...
+`<=` | `~/` | `&`  | `~`
+`>=` | `*`  | `<<` | `==`
+`–`  | `%`  | `>>`
+{:.table}
+
+{{site.alert.note}}
+  You may have noticed that some [operators](#operators), like `!=`, are not in
+  the list of names. That's because they're just syntactic sugar. For example,
+  the expression `e1 != e2` is syntactic sugar for `!(e1 == e2)`.
+{{site.alert.end}}
```
lrhn, line 3052: Maybe worth mentioning that `??`, `&&` and ||` are not in the list either because they are not, and cannot be, methods because of their short-circuiting nature.

(And `!` is not there for ... probably historical reasons. No real reason I can think of.)

lrhn (approved):

src/_guides/language/language-tour.md
```diff
@@ -1559,9 +1559,9 @@ assert(foo() == null);
 
 ## Operators
 
-Dart defines the operators shown in the following table.
-You can override many of these operators, as described in
-[Overridable operators](#overridable-operators).
+Dart supports the operators shown in the following table.
+You can define your own operators, as described in the section on
```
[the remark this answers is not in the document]
chalin: @eernstg - Good points. In rereading, I agree that "define your own" is misleading.

How about:

```diff
- You can override many of these operators, 
+ You can implement and override many of these operators, 
```

My concern with the original formulation is that it might give readers who are new to Dart the impression that one can _only_ override operators.

src/_guides/language/language-tour.md
chalin: Thanks for the feedback.

@kwalrath - thoughts?

AFAIK, in the language tour we generally try to avoid (too many) low-level technical details (Kathy can correct me if I'm wrong). So ...

> Maybe worth mentioning that ??, && and || are not in the list either because they are not, and cannot be, methods because of their short-circuiting nature.

... would make a great commentary in the language spec (AI for @eernstg? 😄)

> And ! is not there for ... probably historical reasons. No real reason I can think of.

Interesting! (Also possibly a good commentary for the DLS?)

src/_guides/language/language-tour.md
chalin: This ensures that links, from outside the site, that target `#overridable-operators` will still work. 

I can certainly drop this if you prefer. @kwalrath?

src/_guides/language/language-tour.md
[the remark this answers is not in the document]
eernstg: Right, we don't want to imply that there must be an override relation, but even 'implement' is slightly dangerous (if it is taken to mean that there must be an `implements` relationship): We keep hinting that an operator cannot be added to an arbitrary interface. It's difficult to keep it precise as well as concise, but here's one more way to say it:

```suggestion
With many of these operators (e.g., `+`), a class can declare an implementation (`operator +`), thus adding that operator to the class or overriding a declaration in a superinterface. This is described in the section on
```

src/_guides/language/language-tour.md
eernstg: OK, then it makes sense to keep it. But maybe the link from the same file shouldn't have the link text 'operators' when it jumps to 'Overriding members'.

src/_guides/language/language-tour.md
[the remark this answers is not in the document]
chalin: Thanks for the suggestion.

> It's difficult to keep it precise as well as concise

Yes. Here's one more try at keeping it concise; how about (additions in bold):

> You can **implement and** override many of these operators **as class members**, ...

Or more simply:

> You can **implement** many of these operators **as class members**, ...

(leaving it understood that operators, like other members, can be overridden).

Trying to find a balance here for the Language Tour, in being sufficiently accurate without getting too much into details. @kwalrath - any further suggestions or comments?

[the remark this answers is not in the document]
eernstg: I like the 'more simply' variant, and I'm sure @kwalrath has good further suggestions. ;-)

src/_guides/language/language-tour.md
chalin: Oh, thanks for pointing that out! I missed the in-page link to `#overridable-operators` (strange, I thought that I had searched for those before); that needs to be changed to `#_operators`.

Hmm, I've also found another place in the tour that needed a slight update. Please see my latest commit.

eernstg (approved):
LGTM.

src/_guides/language/language-tour.md
kwalrath: Agreed that we don't want to show these kinds of details in the tour. (It'd be fine to note them or this comment in `{% comment %}` sections, so future writers know why they're omitted.)

src/_guides/language/language-tour.md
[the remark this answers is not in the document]
kwalrath: I like the "more simply" version, too.

src/_guides/language/language-tour.md
chalin: Good idea, I'll add those as comments.

src/_guides/language/language-tour.md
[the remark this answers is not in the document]
chalin: Done. Changed the text as follows:

> Dart supports the operators shown in the following table.
You can implement many of these [operators as class members](#_operators).

PTAL @kwalrath

src/_guides/language/language-tour.md
```diff
@@ -1588,6 +1587,11 @@ You can override many of these operators, as described in
   Operator precedence is an approximation of the behavior of a Dart parser.
   For definitive answers, consult the grammar in the
   [Dart language specification][].
+  {%- comment %}
+  Internal note from https://github.com/dart-lang/site-www/pull/2691#discussion_r506184100:
+  -  `??`, `&&` and `||` are excluded because they are lazy / short-circuiting operators
+  - `!` is probably excluded for historical reasons
+  {% endcomment %}
```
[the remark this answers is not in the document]
chalin: Oh my! Apologies for my multitasking blunder. Done.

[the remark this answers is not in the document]
chalin: Actually, just to be sure (since 3066 falls inside a code block for me), I've moved it just after the Note that starts with "You may have noticed that some [operators](#operators), like `!=`, are not in the list of names."

PR #2693, opened 2020-10-08
Add missing type annotations

komainu8:
Hello,

I suppose that this example sets the types better than omit the types.
Because Dart recommends type annotations.

I just started studying Dart. So, this PR maybe not a good...

thanks,

### DIFF
--- a/examples/misc/test/language_tour/basic_test.dart
+++ b/examples/misc/test/language_tour/basic_test.dart
@@ -6,12 +6,12 @@ void main() {
   test('basic', () {
     // #docregion
     // Define a function.
-    printInteger(int aNumber) {
+    void printInteger(int aNumber) {
       print('The number is $aNumber.'); // Print to console.
     }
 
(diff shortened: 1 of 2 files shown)


komainu8:
@googlebot I signed it!

kwalrath:
You updated the .dart file as well as the .md file, well done! I think we'll accept your change... but I'll check. When we originally wrote the language tour, it was a selling point that you could omit types until you were ready to add them. I'm not sure how we feel about omitted types nowadays.

@mit-mit or @munificent, want to opine? This is the sample at the top of the language tour: https://dart.dev/guides/language/language-tour#a-basic-dart-program. Adding `void` would also give us a chance to mention it in the discussion just after the sample.

munificent (approved):
LGTM.

kwalrath:
Thank you, @komainu8!

komainu8:
Thank you for merge my pull request.

PR #2694, opened 2020-10-09
Clarify capitalization of Id

aizatazhar:
Closes #2456

PR #2628, opened 2020-09-10
Add in-progress doc for null safety migration.

munificent:
This isn't ready to land on master yet. It has a few TODOs for things like version numbers that we won't know until closer to shipping. My plan is to fix all the stuff in it we can fix now, sit on the PR until closer to shipping, and then fix the stuff once everything is pinned down then land.

So, for now, don't worry too much about the TODOs (unless you know anything about them). But please do take a look at everything else. Thanks!

PR #2696, opened 2020-10-10
Language versioning docs

kwalrath:
One PENDING question:
When was per-library language version selection introduced? 2.10? (If so, I should probably add it to the 2.10 item, no?)

I started making related changes (lower SDK constraints) in other files, but I'll probably finish those in another PR.

Staged:
https://kw-dartlang-3.firebaseapp.com/guides/language/evolution#language-versioning

Contributes to #2694
Related PR: #2628

/cc @kevmoo @mit-mit

### DIFF
--- a/src/_guides/language/evolution.md
+++ b/src/_guides/language/evolution.md
@@ -12,7 +12,6 @@ If you want details about the currently supported language, see the
 To use a language feature that was introduced after 2.0,
 specify [SDK constraints][] that are no lower than
 the release when the feature was first supported.
-(Details are in the [language versioning section][] of this page.)
 For example, to use extension methods, which were [supported starting in 2.7][],
 the `pubspec.yaml` file can have 2.7.0 as the lower constraint:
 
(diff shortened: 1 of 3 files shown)


src/_guides/language/evolution.md
```diff
@@ -127,25 +151,99 @@ void main() {
...
+Before language versioning,
+Dart tools assumed that all Dart code could use all features
+that were supported by the tool's SDK version.
+This assumption causes problems when a package
+that uses a _newer_ language feature
+has [SDK constraints][] that are too low.
+For example, consider a package that uses set literals (2.2)
+but has the SDK constraints `">=2.0.0 <3.0.0"`.
+People using Dart 2.0.0 or 2.1.0 can download the package,
+but their tools can't understand the set literals.
```
munificent, line 184: I believe even before language versioning, our tools would warn if you used a language feature that was added to Dart after your SDK's minimum version. That's a fairly straightforward set of warnings we could support whenever if we wanted to.

Language versioning is quite a bit more complex. It's really about being able to handle the fact that the *same* Dart code may mean *different* things in different versions of Dart. Concretely, in 2.10:

```dart
int i;
```

Declares a variable whose value can be `null`, and has no errors. Once null safety ships, that same program now has a compile error.

The example you describe with set literals is one of *forwards* incompatibility: an old SDK can't handle new set literals. That's definitely an issue, but language versioning is more about handling *backwards* incompatibility. It lets us change the language in incompatible ways without breaking existing code.

src/_guides/language/evolution.md
```diff
@@ -127,25 +151,99 @@ void main() {
...
+
+With language versioning,
+a single Dart SDK simultaneously supports
+multiple different versions of the Dart language.
+When you compile some Dart code,
+Dart figures out what version the code is targeting,
+and you're warned if the code uses a feature
+that was introduced after the version in the lower SDK constraint.
```
munificent, line 192: It's more than warnings. It means Dart may interpret the code in an entirely different way.

src/_guides/language/evolution.md
```diff
@@ -127,25 +151,99 @@ void main() {
...
+
+On rare occasions, Dart introduces a
+backwards-incompatible feature like [null safety][].
+Previously valid code might not work with null safety,
+so to get the benefits of null safety you'll need to migrate your code.
+And because migrating to null safety can't be completely automated,
+you'll need to be able to migrate one library
+(usually one Dart file) at a time.
```
munificent, line 200: It can't be completely automated, but using the migration tool, the normal workflow is still to migrate an entire package at a time.

The file comments are intended to be a fairly rarely-used feature for most application developers.

src/_guides/language/evolution.md
```diff
@@ -127,25 +151,99 @@ void main() {
...
+
+{% comment %}
+  Once /null-safety/migrating-to-null-safety exists,
+  the previous paragraph should link to it.
+{% endcomment %}
+
+To enable file-by-file migration,
+Dart 2.10 introduced **per-library language version selection**.
+**[PENDING: was it really a 2.10 feature?]**
```
munificent, line 209: I could be wrong, but I think language versioning shipped in 2.7.0. (@lrhn would know more.) We didn't announce it because it didn't really do anything yet (except for give you warnings if you tried to use extension methods in packages whose SDK constraints were too low).

src/_guides/language/evolution.md
```diff
@@ -127,25 +151,99 @@ void main() {
...
+  (not `///` or `/*`),
+  and it must appear before any Dart code in the file.
+  Whitespace (tabs and spaces) doesn't matter,
+  except within the `@dart` and version strings.
+  As the example above shows,
+  other comments can appear before the `@dart` comment.
+
+  The library language version is usually lower than
+  the package default language version,
+  but it doesn't have to be:
```
munificent, line 241: It almost always should be lower than the package default version or you're going to have a bad time. I would consider not even mentioning that it can be a higher number. That exists almost entirely for code generators.

src/_guides/language/evolution.md
lrhn: I *think* it shipped in 2.8. (That's why `package:package_config` defaults code with only a `.packages` file to version 2.7, the last version prior to language versioning existing. From 2.8 you should stop using `.packages`.)
There is no entry in our changelog for 2.8.0. We go directly to 2.8.1 and states that there are no new language features.
The only mention of "language versioning" in the CHANGELOG.md is when pub starts generating the `package_config.json` file in 2.7, which it did prior to launching language versioning). There is a link to the package versioning spec in the 2.8.1 notes, in a fix to Pub.

Which also means that it's shouldn't be affecting extension methods, which were released in 2.7 (and enabled for preview in 2.6).

I think language versioning snuck under the radar because it was never an *experiment*. It's just a compiler feature.

kwalrath:
ptal @munificent: https://kw-dartlang-3.firebaseapp.com/guides/language/evolution#language-versioning

munificent (approved):
Looks great! :)

PR #2704, opened 2020-10-15
Add more details about final and const, fix formatting

### LINKED ISSUE #1203: `final` explanation from 'A Tour of the Dart Language' is a bit misleading
Page URL: https://www.dartlang.org/guides/language/language-tour
Page source: https://github.com/dart-lang/site-www/tree/master/src/_guides/language/language-tour.md

I think that the explanation for the `final` keyword from the Dart language tour is a bit misleading.  It states:

>You can’t change the value of a final variable:

While technically true, coming from a C++ background, my expectation was that `final` is similar to `const` in C++ where it is predominantly used with pointers and references to specify that the object cannot be (visibly) mutated via that pointer/reference. It would help if the Dart language tour explicitly mentioned that `final` does not prevent the object from being mutated.

(As an aside, I think that `final` makes cases like:
```dart
final x = foo();
bar(x); // Does this mutate x?
```
misleading to readers... Should we possibly consider discouraging that on one of the Effective Dart pages?)

micpap25:
Fixes #1203
Adds a note at the bottom of the section regarding how `final` and `const` handle mutations of objects.
Additionally, makes usage of code-blocking on `final` and `const` more consistent.

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -333,7 +333,7 @@ the first time it's used.
   or in the constructor's [initializer list](#initializer-list).
 {{site.alert.end}}
 
-Here's an example of creating and setting a final variable:
+Here's an example of creating and setting a `final` variable:
 
 <?code-excerpt "misc/lib/language_tour/variables.dart (final)"?>
 ```dart
@@ -341,7 +341,7 @@ final name = 'Bob'; // Without a type annotation
 final String nickname = 'Bobby';
 ```
 
-You can't change the value of a final variable:
+You can't change the value of a `final` variable:
 
 {:.fails-sa}
 <?code-excerpt "misc/lib/language_tour/variables.dart (cant-assign-to-final)"?>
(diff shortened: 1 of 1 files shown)


src/_guides/language/language-tour.md
```diff
@@ -405,6 +405,14 @@ const list = [i as int]; // Use a typecast.
 const map = {if (i is int) i: "int"}; // Use is and collection if.
 const set = {if (list is List<int>) ...list}; // ...and a spread.
 ```
+
+{{site.alert.note}}
+  Although a `final` object cannot be modified,
+  its fields can be mutated. 
+  In comparison, both a `const` object and its properties
+  cannot be reassigned.
```
kwalrath, line 413: Let's make the wording more consistent within this note. Maybe:

```suggestion
  its fields can be changed. 
  In comparison, a `const` object and its fields
  cannot be changed: they're _immutable_.
```

kwalrath (commented):
Nice changes! I have some suggestions for the wording of the note.

micpap25:
Made the changes.

PR #2705, opened 2020-10-15
Fix wrong version number in 'brew switch' command

### LINKED ISSUE #2672: version number in 'brew switch' command might wrong
on https://dart.dev/get-dart

Command in Get Dart/Installing SDK/Mac

` brew switch dart 2.1.0`

Error: 

```
Error: dart does not have a version "2.1.0" in the Cellar.
dart's installed versions: 2.10.0
```

Suggestion: 
`brew switch dart 2.10.0`

komainu8:
Fixes #2672

I fixed the version number for the moment.
However, we will need to fix whenever dart version up in this correcting method.

Would it be better if we don't write directly version number as below?

```diff
- $ brew switch dart 2.10.1
+ $ brew switch dart <version>
```

Could you let me know if you have any other better ideas?

### DIFF
--- a/src/tools/sdk/_mac.md
+++ b/src/tools/sdk/_mac.md
@@ -22,7 +22,7 @@ To switch between locally installed Dart releases, use
 `brew switch dart <version>`. Example:
 
 ```terminal
-$ brew switch dart 2.1.0
+$ brew switch dart 2.10.1
 ```
 
 To see which versions of Dart you've installed:


kwalrath:
I think it's fine to let the version number lag — after all, this command is for switching between version numbers, which implies that you have at least two stable versions installed.

Thanks for the PR!

kwalrath (approved):

komainu8:
I got it. Thank you for your explanation.

PR #2706, opened 2020-10-15
Simplify default parameters example

### LINKED ISSUE #1242: Default parameter values example don't need null check
On this example `device` has a default parameter value, so there's no need to null check it:

```
String say(String from, String msg,
    [String device = 'carrier pigeon', String mood]) {
  var result = '$from says $msg';
  if (device != null) {
    result = '$result with a $device';
  }
  if (mood != null) {
    result = '$result (in a $mood mood)';
  }
  return result;
}

assert(say('Bob', 'Howdy') ==
    'Bob says Howdy with a carrier pigeon');
```

micpap25:
Fixes #1242 

Simplifies the default parameters example so it is much more clear by removing unecessary `if` statements and the `mood` variable.

### DIFF
--- a/examples/misc/test/language_tour/functions_test.dart
+++ b/examples/misc/test/language_tour/functions_test.dart
@@ -27,7 +27,8 @@ void main() {
 
   test('optional-positional-param-default', () {
     // #docregion optional-positional-param-default
-    String say(String from, String msg, [String device = 'carrier pigeon']) {
+    String say(String from, String msg, 
+        [String device = 'carrier pigeon']) {
       var result = '$from says $msg with a $device';
       return result;
     }
--- a/examples/misc/test/language_tour/functions_test.dart
+++ b/examples/misc/test/language_tour/functions_test.dart
@@ -27,7 +27,7 @@ void main() {
 
   test('optional-positional-param-default', () {
     // #docregion optional-positional-param-default
-    String say(String from, String msg, 
+    String say(String from, String msg,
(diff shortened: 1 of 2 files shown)


kwalrath:
To get the build working again, we'll need to update the example code in examples/misc/test/language_tour/functions_test.dart

kwalrath:
I suspect we'll also need to break the first line where it was broken before, to pass dartfmt. (We have a <80 line length for code snippets, for maximum code visibility under a variety of page widths.)

kwalrath (commented):
The code LGTM (thank you!). @micpap25 are you able to make the changes I described, or do you want me to add to this PR?

micpap25:
I think I can make the change, I'll let you know if I cannot.

micpap25:
@kwalrath Hmm, I'm not sure why it's breaking now. Any idea?

kwalrath:
> @kwalrath Hmm, I'm not sure why it's breaking now. Any idea?

I copied it locally and did tool/dartformat.sh and tool/refresh-code-excerpts.sh, and apparently there's a space at the end of a line! I'll try deleting it and see what happens.

micpap25:
awesome, looks like it works now. Thank you so much.

kwalrath (approved):

kwalrath:
I wish the Travis CI output were clearer, but I'm glad we got it working. Thank you, @micpap25!

PR #2707, opened 2020-10-17
style: prevent table overflow in mobile

g1eny0ung:
When viewing https://dart.dev/guides/language/language-tour in the mobile viewport, because the `keywords` table has a slightly longer width, there will be an offset across the page.

image


g1eny0ung:
@googlebot I signed it!

kwalrath (approved):
Thank you, @g1eny0ung! I didn't know we could (or needed to) do this!

PR #2708, opened 2020-10-17
Added `void` to the basic Dart program section of language tour.

### LINKED ISSUE #2699: Add `void` to the basic Dart program section of language tour
#2693 added`void` to the function examples. We should mention `void` in the discussion that follows.

sumandipanshu:
Fixes #2699.
Added brief description about `void`.
Is this enough? Or I should add more detailed info?
Please review @kwalrath .

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -71,6 +71,9 @@ apps:
     equivalent inside of a string literal. For more information, see
     [Strings](#strings).
 
+`void`
+:   A type(here, used as a return type of function). It can take anything as it's value, but it's value cannot be used anywhere. For more information, see [this article](https://medium.com/dartlang/future-void-vs-future-null-whats-the-difference-84a42d616e7d).
+
 `main()`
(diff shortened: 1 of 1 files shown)


kwalrath:
I tried to edit this PR but couldn't. I think we should move `void` up (to use the same order as the code) and point to a different article. What do you think, @sumandipanshu? 

```
`void`
:   A special type that indicates a value that's never used.
    Functions like `printInteger()` and `main()` that don't explicitly return a value
    have the `void` return type.
    For more information, see [this article][void article].
    
[void article]: https://medium.com/dartlang/dart-2-legacy-of-the-void-e7afb5f44df0

`int`
:   Another type, indicating an integer.
    Some additional [built-in types](#built-in-types)
    are `String`, `List`, and `bool`.
```

kwalrath (changes requested):
Thanks for the PR! See my comment for changes I think we should make.

sumandipanshu:
Thanks for the review. I will update the PR with the suggested changes.

PR #2709, opened 2020-10-17
Removing unnecessary `new` keyword

JHBitencourt:
According to https://dart.dev/guides/language/effective-dart/usage#dont-use-new


kwalrath (approved):
Nice catch. Thanks, @JHBitencourt!

Ideally these would be auto-included and thus analyzed to catch issues like this (#1803). But until that day...

PR #2716, opened 2020-10-24
Add trailing comma info

### LINKED ISSUE #1404: add info about trailing commas
Trailing commas are allowed in places like list literals, set literals, map literals, enum declarations, argument lists, and parameter lists. I think that the example at the bottom of https://www.dartlang.org/guides/language/language-tour#sets is the first usage of a trailing comma in the language tour, but maybe we should introduce them in the list literal section.

We should mention that it doesn't matter whether the trailing comma is there or not, but it can help prevent errors when copy-pasting code.

More info: https://github.com/dart-lang/sdk/issues/26644

micpap25:
Fixes #1404
Adds info about trailing commas and an example for them.

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -718,6 +718,19 @@ var list = [1, 2, 3];
   [type inference.](/guides/language/type-system#type-inference)
 {{site.alert.end}}
 
+Trailing commas are allowed in Dart collections. 
+They don't affect the collection, but having them can 
+prevent errors when copy-pasting code. 
+
+<?code-excerpt "misc/lib/language_tour/built_in_types.dart (trailing-commas)"?>
+```dart
+var list = [
+  'Car',
+  'Boat',
+  'Plane',
+];
+```
+
 Lists use zero-based indexing, where 0 is the index of the first value
 and `list.length - 1` is the index of the last value. You can get a
 list’s length and refer to list values just as you would in
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -718,9 +718,9 @@ var list = [1, 2, 3];
   [type inference.](/guides/language/type-system#type-inference)
 {{site.alert.end}}
 
-Trailing commas are allowed in Dart collections. 
-They don't affect the collection, but having them can 
-prevent errors when copy-pasting code. 
+You can add a comma after the last item in a Dart collection literal.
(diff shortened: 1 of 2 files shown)


kwalrath (changes requested):
I think the build failed because dartfmt wants to change the code. Maybe it's because of the indentation difference? I'll try changing that in this PR.

src/_guides/language/language-tour.md
```diff
@@ -718,6 +718,19 @@ var list = [1, 2, 3];
   [type inference.](/guides/language/type-system#type-inference)
 {{site.alert.end}}
 
+Trailing commas are allowed in Dart collections. 
+They don't affect the collection, but having them can 
+prevent errors when copy-pasting code.
```
kwalrath, line 723: I'd like to clarify the meaning of "trailing comma" and avoid the ambiguous "they/them". What do you think of this?

```suggestion
You can add a comma after the last item in a Dart collection literal.
This _trailing comma_ doesn't affect the collection,
but it can help prevent copy-paste errors.
```

kwalrath: #1404 mentions other places where trailing commas work, too. I started trying to update this to include those additional places. (Grepping for `','?` and `(',')?` in the spec, it looks like it's enums, args, params, & asserts — probably all the comma-delimited things in the language except perhaps initializers.) But it started getting too complicated. Maybe we should cover each of those cases where they come up?

micpap25:
I added one small section for each of the topics you noted. 

While doing this, I noted that the language tour assumes that the user knows what an argument is, and never actually formally introduces them. It was a bit hard to find a place to mention trailing commas when it came to args and params. Additionally, I can't imagine trailing commas being very useful for passing arguments or defining functions, although that's just my 2 cents. 

While I added explanations for all four of the types that can use trailing commas, I think it would make sense to delete some of them. However, we can keep all of them if you think it's a good idea.

micpap25:
As a Flutter dev - wow, I can't believe I forgot that, lol. I was thinking about it on a single-line context.

kwalrath:
Failing build is bogus (see #2720).

kwalrath:
You can see our latest changes staged here:

https://kw-staging-dartlang-2.firebaseapp.com/guides/language/language-tour

It could always be improved (e.g. the anchored link goes a little lower than it should), but I think it's better than what's there, and it fixes the bug!

kwalrath (approved):

PR #2734, opened 2020-11-06
[DONT MERGE] Update /null-safety for beta

mit-mit:
Some updates:

- Incorporate the principles we laid our over last few blog posts
- Tone down warnings about use now that it's in beta
- Correct various paths
- Remove steps no longer needed

PR #2739, opened 2020-11-07
Real content for /null-safety/migration-guide

kwalrath:
If any updates to /null-safety are necessary, they should go in #2734, not here.

Staged at: ~~https://kw-dartlang-3.firebaseapp.com/null-safety/migration-guide~~
https://kw-www-dartlang-1.firebaseapp.com/null-safety/migration-guide and
https://kw-www-dartlang-1.firebaseapp.com/guides/language/evolution#mixed-mode-programs

Replaces #2628 
Related PR: #2764

~~TODO: Add `/*late*/`, `/*late final*/`, `/*required*/`.~~


kwalrath:
@munificent @kevmoo @mit-mit I've moved the mixed-mode info to the [evolution page](https://kw-www-dartlang-1.firebaseapp.com/guides/language/evolution#mixed-mode-programs) and edited it to match the Google doc style. Is there anything else we can trim from there, or anything I should add back? Anywhere it can be clearer? I wasn't sure of a few points, marked with PENDING.

kwalrath:
After discussing mixed mode with @mit-mit and @vsmenon, it seems like it isn't generally applicable, so I'm going to move the discussion out of the evolution doc and into the null safety docs (probably the migration page).

src/_guides/language/evolution.md
```diff
@@ -269,10 +269,82 @@ except within the `@dart` and version strings.
 As the example above shows,
 other comments can appear before the `@dart` comment.
 
-
 For more information about how language versioning works, see the
 [language versioning specification][language versioning feature].
 
+
+## Mixed-mode programs
```
mit-mit, line 276: I'm not a big fan of the term "mixed-mode". Null safety really isn't a mode, and I think the term makes the non-sound mode sound more appealing than it is.

My personal preference is to call it "partial null safety"; that makes it clear that you are not getting all soundness.

cc @leafpetersen

src/null-safety/migration-guide.md
```diff
@@ -1,23 +1,395 @@
...
+
+**_[PENDING:
+See [Experimenting with null safety in Flutter](https://github.com/flutter/flutter/wiki/Experimenting-with-null-safety-in-Flutter)
+for details (on testing and otherwise; it's sort of this doc, but for pre-beta.
+]_**
+
+
+## Publish your code
+
+We encourage you to publish non-app packages as prereleases
```
kevmoo, line 329: This wording...worries me.

`publish non-app packages` – are we worried that folks confuse publishing an app (to a store) vs publishing a package (to pub.dev)?

maybe we should have seperate sections: `Publishing packages to pub.dev` vs `Publishing your application to a store`?

kwalrath: I originally had two sections, but since we don't really want people to publish apps yet, we deleted that section. We don't really have anything to say about the app case, anyway.

I don't think most people think of apps as packages, so I should be able to reword without lying.

src/null-safety/migration-guide.md
```diff
@@ -1,23 +1,416 @@
...
+   **publish** the null-safe version as a **prerelease** version.
 
-[null safety homepage]: /null-safety
-[Dart announcements]: {{site.group}}/d/forum/announce
-[Dart blog]: https://medium.com/dartlang
+{{ site.alert.info }}
+  **Migrating an app is the same as migrating a package.**
+  Before fully migrating an app,
+  consider waiting until null safety is in a stable release.
+  In the meantime, you can add migration hints to your app
```
tvolkert, line 27: Do we expect them to add migration hints manually?  The tool will try to upgrade them (prematurely in this case), but it also presents a handy way to see where hints might be needed or useful...

@stereotype441 we may want to support a mode in which users can fire up the migration tool for the sole purpose of seeing what would happen, and adding hints.  This can be done as-is by simply not applying the suggestions, but is that an intended/well-supported use case?

@kwalrath if we expect them to do so manually, then making "migration hints" link to a page or section that documents what the supported hints are would be useful.

src/null-safety/migration-guide.md
```diff
@@ -1,23 +1,416 @@
...
+**[PENDING: check!!!]**
+
+```yaml
+environment:
+  sdk: '>=2.12.0-18.0.beta <3.0.0'
+```
+
+
+{{ site.alert.warn }}
+  When you return to working on production code, remember to
```
tvolkert, line 102: But that will downgrade their Dart SDK and make any null safety edits they've made turn into compiler errors..

src/_guides/language/evolution.md
kwalrath: We've decided on _unsound null safety_, per #2748.

src/null-safety/migration-guide.md
mit-mit: I don't think we need to be so strong as to say "you cannot publish apps". We have plenty of Flutter beta features that are used in published apps.

I think we should make it clear that a beta feature is potentially less stable, and then let app developers choose themselves if they want to take the risk.

src/null-safety/migration-guide.md
mit-mit: I'm not sure it's worthwhile to generally recommend several migration passes to most app developers. I'd rather loosen the guidance against migrating apps.

rafalbednarczuk:
`dart migrate` has bugs.
I was following migration guide with my package [curved_navigation_bar](https://pub.dev/packages/curved_navigation_bar). I wanted to try migration tool.

<details>
<summary>Here first run of `dart migrate`</summary>
<p>

```
Migrating /home/rafal/AndroidStudioProjects/curved_navigation_bar

See https://dart.dev/go/null-safety-migration for a migration guide.

Note: more than one project found; migrating the top-level project.

Analyzing project...
[-----------------------------------------------------------------------------|]
16 analysis issues found:
  error • The parameter 'key' can't have a value of 'null' because of its type, and no non-null default value is provided at lib/curved_navigation_bar.dart:18:9 • (missing_default_value_for_parameter)
  error • The parameter 'items' can't have a value of 'null' because of its type, and no non-null default value is provided at lib/curved_navigation_bar.dart:19:20 • (missing_default_value_for_parameter)
  error • The parameter 'buttonBackgroundColor' can't have a value of 'null' because of its type, and no non-null default value is provided at lib/curved_navigation_bar.dart:22:10 • (missing_default_value_for_parameter)
  error • The parameter 'onTap' can't have a value of 'null' because of its type, and no non-null default value is provided at lib/curved_navigation_bar.dart:24:10 • (missing_default_value_for_parameter)
  error • Non-nullable instance field '_startingPos' must be initialized at lib/curved_navigation_bar.dart:40:10 • (not_initialized_non_nullable_instance_field)
  error • Non-nullable instance field '_pos' must be initialized at lib/curved_navigation_bar.dart:42:10 • (not_initialized_non_nullable_instance_field)
  error • Non-nullable instance field '_icon' must be initialized at lib/curved_navigation_bar.dart:44:10 • (not_initialized_non_nullable_instance_field)
  error • Non-nullable instance field '_animationController' must be initialized at lib/curved_navigation_bar.dart:45:23 • (not_initialized_non_nullable_instance_field)
  error • Non-nullable instance field '_length' must be initialized at lib/curved_navigation_bar.dart:46:7 • (not_initialized_non_nullable_instance_field)
  error • The parameter 'onTap' can't have a value of 'null' because of its type, and no non-null default value is provided at lib/src/nav_button.dart:10:19 • (missing_default_value_for_parameter)
  error • The parameter 'position' can't have a value of 'null' because of its type, and no non-null default value is provided at lib/src/nav_button.dart:10:31 • (missing_default_value_for_parameter)
  error • The parameter 'length' can't have a value of 'null' because of its type, and no non-null default value is provided at lib/src/nav_button.dart:10:46 • (missing_default_value_for_parameter)
  error • The parameter 'index' can't have a value of 'null' because of its type, and no non-null default value is provided at lib/src/nav_button.dart:10:59 • (missing_default_value_for_parameter)
  error • The parameter 'child' can't have a value of 'null' because of its type, and no non-null default value is provided at lib/src/nav_button.dart:10:71 • (missing_default_value_for_parameter)
  error • Non-nullable instance field 's' must be initialized at lib/src/nav_custom_painter.dart:9:3 • (not_initialized_non_nullable_instance_field)
  error • Non-nullable instance field 'loc' must be initialized at lib/src/nav_custom_painter.dart:9:3 • (not_initialized_non_nullable_instance_field)

The migration tool didn't start, due to analysis errors.

The following steps might fix your problem:
1. Set the lower SDK constraint (in pubspec.yaml) to a version before 2.12.
2. Run `dart pub get`.
3. Try running `dart migrate` again.
```

</p>
</details>  

I expected migration suggestions with an UI as described in the guide.

I decided to fix the errors by myself.
<details>
<summary>Here is the run of `dart migrate` after fixing errors</summary>
<p>

```
Migrating /home/rafal/AndroidStudioProjects/curved_navigation_bar

See https://dart.dev/go/null-safety-migration for a migration guide.

Note: more than one project found; migrating the top-level project.

Analyzing project...
[----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|]No analysis issues found.

Generating migration suggestions...
[-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------]

Compiling instrumentation information...
Bad state: Seem to be migrating code that's already migrated
#0      NullabilityMigrationImpl.finish (package:nnbd_migration/src/nullability_migration_impl.dart:204:7)
#1      NonNullableFix.finish (package:nnbd_migration/src/front_end/non_nullable_fix.dart:108:15)
#2      _FixCodeProcessor.runLaterPhases (package:nnbd_migration/migration_cli.dart:1109:29)
<asynchronous suspension>
#3      MigrationCliRunner.run (package:nnbd_migration/migration_cli.dart:674:24)
<asynchronous suspension>
#4      MigrateCommand.run (package:nnbd_migration/migration_cli.dart:185:7)
<asynchronous suspension>
#5      CommandRunner.runCommand (package:args/command_runner.dart:197:13)
<asynchronous suspension>
#6      DartdevRunner.runCommand (package:dartdev/dartdev.dart:249:24)
<asynchronous suspension>
#7      runDartdev (package:dartdev/dartdev.dart:112:14)
<asynchronous suspension>
#8      main (file:///b/s/w/ir/cache/builder/src/third_party/dart/pkg/dartdev/bin/dartdev.dart:11:3)
<asynchronous suspension>

```

</p>
</details>  

At the end `dart migrate` didn't help me at all. Errors given by `dart migrate` were the same as provided by `dart analysis` so I had them in my IDE anyway.

```
Dart SDK version: 2.12.0-37.0.dev (dev) (Thu Nov 12 14:03:34 2020 -0800) on "linux_x64"
```

kwalrath:
> `dart migrate` has bugs.

Could you please file these at https://github.com/dart-lang/sdk/issues?

src/null-safety/migration-guide.md
kwalrath: We don't say "cannot publish apps". We just say "we generally recommend waiting for a stable SDK release."

src/null-safety/migration-guide.md
kwalrath: I changed it to "If". The intent was just to remind them, especially if they're just checking on the status of dependencies and not in the midst of really migrating.

src/null-safety/migration-guide.md
```diff
@@ -1,23 +1,399 @@
...
+For example, in the preceding screenshot,
+the tool infers that the `bar2()` function returns a non-nullable type (`int`).
+
+{{ site.alert.tip }}
+  To see the reason for each change (or non-change),
+  click the type.
+{{ site.alert.end }}
+
+When analysis infers the wrong nullability,
+you can override its choice by inserting temporary `/*?*/` and `/*!*/` markers.
```
mit-mit, line 228: I think we should consider elaborating here. It's not that clear when to add these hints. The readme has a nice example, I think:
https://github.com/dart-lang/sdk/tree/master/pkg/nnbd_migration#interactive-feedback-to-the-tool

src/null-safety/migration-guide.md
```diff
@@ -1,23 +1,399 @@
...
+(for example, A imports B which imports C, and C imports A),
+consider migrating those libraries together.
+
+To migrate a package by hand, follow these steps:
+
+1. Modify the SDK constraints in the package's `pubspec.yaml` file,
+   setting the language version to 2.12 beta or later:
+   ```yaml
+environment:
+  sdk: '>=2.12.0-18.0.beta <3.0.0' [PENDING: check version; would 2.12.0-0 work?]
```
mit-mit, line 280: Yes, any 2.12 version will work. `2.12.0-0` is the shortest one.

src/null-safety/migration-guide.md
```diff
@@ -1,23 +1,399 @@
...
+   every library in your package to 2.12,
+   opting them all in to null safety.
+
+
+3. Open the package in your IDE. <br>
+   You're likely to see a lot of analysis errors.
+   That's OK.
+
+4. _Optional:_ Add a [language version comment][] to the top of
+   any Dart files that you don't want to consider during your current migration:
```
mit-mit, line 299: I wonder if this should be omitted here, and then all mentioning of unsound null safety & mixed version should be elsewhere?

src/null-safety/migration-guide.md
```diff
@@ -1,23 +1,399 @@
...
+
+```terminal
+View the migration suggestions by visiting:
+
+  http://127.0.0.1:60278/Users/you/project/mypkg.console-simple?authToken=Xfz0jvpyeMI%3D
+```
+
+Visit that URL in a Chrome browser
+to see an interactive UI
+where you can guide the migration process:
```
mit-mit, line 213: There is a bunch more details here for the suggested flow when using the tool; some of this might be relevant:
https://github.com/dart-lang/sdk/blob/master/pkg/nnbd_migration/README.md#using-the-tool

src/null-safety/migration-guide.md
```diff
@@ -1,23 +1,399 @@
...
+### Check dependency status
+
+Get the migration state of your package's dependencies,
+using the following command:
+
+```terminal
+$ dart pub outdated --mode=null-safety
+```
+
+If the output says **all the packages fully support null safety**,
```
mit-mit, line 116: I think we have have to answer the question here of whether developers have to wait for both direct and dev (and transitive of both) dependencies.

src/null-safety/migration-guide.md
kevmoo: As long as it IS elsewhere. This is going to mess up a bunch of folks.

src/null-safety/migration-guide.md
kwalrath: I've elaborated the heck out of this. Testing the readme's example, I found that it had some nonstandard code and the results were different, so I used my own versions of those. ptal @mit-mit.

src/null-safety/migration-guide.md
kwalrath: The top of this section motivates the whole migrating by hand thing using incremental migration, and it mentions mixed-version programs. So... I'm not sure that deleting this step would be helpful/enough.

src/null-safety/migration-guide.md
kwalrath: I got some of that in there, but I didn't get `pub get` in there. Will do...

(I was surprised that comments got written in right away, btw. I thought they'd just be suggestions that wouldn't be applied until you clicked Apply. I ended up editing the source files directly a fair amount.)

src/null-safety/migration-guide.md
```diff
@@ -1,23 +1,507 @@
...
+```
+3. Run `dart pub upgrade`.
+4. Run `dart pub outdated --mode=null-safety` again,
+   to make sure that all dependencies are null safe and up-to-date.
+
+{{ site.alert.info }}
+  **Why do all dependencies need to support null safety?**
+  When all direct dependencies support null safety,
+  you can _run your app_ with sound null safety.
+  When all dev dependencies support null safety,
```
leafpetersen, line 148: This is an oversimplification  - you may have dev dependencies which are irrelevant to running your tests in sound mode - in fact this is fairly likely.

src/null-safety/migration-guide.md
```diff
@@ -1,23 +1,507 @@
...
+  You can, however, make changes like refactoring
+  that don't depend on null-safety features.
+
+  When you've finished editing your code,
+  click **Rerun from sources** to pick up your changes.
+
+A single hint can have ripple effects elsewhere in the code.
+In the example from before,
+adding a `/*!*/` marker where `zero` is assigned its value (on line 2)
+makes the migration tool infer the type of `zero` as `int?` instead of `int`.
```
leafpetersen, line 292: Shouldn't this be the other way around?  `int` instead of `int?` ?

src/null-safety/migration-guide.md
```diff
@@ -1,23 +1,507 @@
...
+</table>
+
+#### Applying changes
+
+When you like all of the changes
+that the migration tool proposes, click
+**Apply migration** to tell the tool to save the migrated code
+and to update the minimum SDK in your pubspec.
+
+The next step is to [test your code](#step3-test).
```
leafpetersen, line 337: The next step is to ensure that your code has no static errors.  The migration tool may not be able to migrate 100% of your code, particularly if you override it's decisions.  After applying the changes from the tool, you should run the analyzer to ensure that the code now passes static analysis.

src/null-safety/migration-guide.md
```diff
@@ -1,23 +1,507 @@
...
+$ dart pub get
+$ dart analyze
+$ dart test
+```
+
+***[PENDING: What are the commands for flutter?]***
+
+You might need to update tests that expect null values.
+
+If analysis or testing show the you need to make
```
leafpetersen, line 429: "show the" -> "show that"?

src/null-safety/migration-guide.md
kwalrath: Oops, yes.

src/null-safety/migration-guide.md
kwalrath: I thought it was, and even had a comment about it, but no one else said anything! Building was one thing I thought of. What are other common ones?

Is it possible that they won't even be able to _build_ their app if their dev dependencies aren't null safe?

src/null-safety/migration-guide.md
mit-mit: Well, we say "we generally recommend waiting for a stable SDK release". That seems a bit late to say here under step 4. We already have a general warning at the top of the page, so I think we can remove the warning here from step 4.

src/null-safety/migration-guide.md
kwalrath: This is answered in a note later, but I'll move it up before the example.

src/null-safety/migration-guide.md
kwalrath: How about this:

>   When all of an app's direct dependencies support null safety,  you can _run the app_ with sound null safety. When all the dev dependencies support null safety, you can _run tests_ with sound null safety. You might also need null-safe dev dependencies for other reasons, such as building.

src/null-safety/migration-guide.md
mit-mit: @kwalrath to clarify, I'm suggesting we change:

"Migrating an app is the same as migrating a package. Before fully migrating an app, consider waiting until null safety is in a stable release. In the meantime, you can learn about null safety, perhaps adding migration hints to your app or refactoring your app code."

to:

"Migrating an app is technically the same as migrating a package. Before fully migrating an app, consider waiting until null safety is in a stable release, and all your dependencies are ready. In the meantime, you can learn about null safety."

src/null-safety/migration-guide.md
mit-mit: Love it, thanks much!

src/null-safety/migration-guide.md
mit-mit: What I meant was to potentially move the whole "Migrating by hand" section to a new "Advanced manual null safety" migration topic, which could also hold https://kw-www-dartlang-1.firebaseapp.com/guides/language/evolution#per-library-language-version-selection. It would simplify the flow here, which is already complicated.

src/null-safety/migration-guide.md
mit-mit: The `pub get` part shouldn't be needed here in step 2 as already covered that in step 1.

I think we can resolve this comment?

src/null-safety/migration-guide.md
```diff
@@ -1,23 +1,507 @@
...
+[Dart SDK archive]: /tools/sdk/archive#beta-channel
+
+{{ site.alert.warn }}
+  If you return to working on production code, remember to
+  **switch back to a stable release**
+  (for example, by running `flutter channel stable`).
+{{ site.alert.end }}
+
+
+### Opt into null safety
```
mit-mit, line 90: @jonasfj @munificent is this needed here while checking if dependencies are ready, or could that be done with a 2.10 SDK? I guess it might not be able to resolve deps with bad SDK constraints, but generally it should not be needed, right?

Here the section is in the staged content:
https://kw-www-dartlang-1.firebaseapp.com/null-safety/migration-guide#opt-into-null-safety

src/_guides/language/evolution.md
```diff
@@ -269,10 +269,82 @@ except within the `@dart` and version strings.
...
+
+Mixed mode lets you migrate to new features at your own pace.
+It also frees package maintainers to migrate their code,
+with the knowledge that even legacy users can get new
+bug fixes and other improvements.
+However, mixed-mode programs don't get all the advantages
+that features such as null safety can bring.
+
+**_[PENDING: Is it OK to talk about mixed mode generically,
+or is it tied tightly to null safety?]_**
```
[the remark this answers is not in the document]
mit-mit: Yes, I'm generally in favor of 1) tying mixed version docs with null safety, and 2) considering moving the mixed version + manual migration content out into a separate advanced topic so that this page can focus on the recommended flow (migrating in order).

src/null-safety/migration-guide.md
kwalrath: I think the `pub get` is necessary because of moving the min SDK back to 2.9.

src/null-safety/migration-guide.md
```diff
@@ -119,14 +120,25 @@ then you can start migrating.
...
-For example, in the following output for a simple package,
-the green checkmarked versions support null safety:
+{{ site.alert.info }}
+  **Why do all dependencies need to support null safety?**
+  When all of an app's direct dependencies support null safety,
+  you can _run the app_ with sound null safety.
+  When all the dev dependencies support null safety,
+  you can _run tests_ with sound null safety.
+  You might also need null-safe dev dependencies for other reasons,
+  such as building.
```
mit-mit, line 130: Not sure what building is here? Perhaps "code generation" is another likely example?

src/null-safety/migration-guide.md
mit-mit: That part shouldn't be needed either; we cherry picked this earlier today:
https://github.com/dart-lang/sdk/issues/44207

It still needs to be rolled to Flutter. After that we can both retest it.

kwalrath: Updated, ptal

src/null-safety/migration-guide.md
```diff
@@ -1,23 +1,516 @@
...
+## 2. Migrate {#step2-migrate}
+
+Most of the changes that your code needs to be null safe
+are easily predictable.
+For example, if a variable can be `null`,
+its type needs a `?` suffix.
+A named parameter that shouldn't be nullable
+needs to be marked `required`.
+
+The migration tool can make most of these changes,
```
devoncarew, line 141: Perhaps add a link around the 'migration tool' text (to the heading just below), to parallel what we do for the 'migrate your code by hand' option? I think that would highlight that there are two migration paths. I'd also consider making these bullet points - that would also highlight that there are two options.

src/null-safety/migration-guide.md
```diff
@@ -1,23 +1,453 @@
...
+
+
+## 2. Migrate {#step2-migrate}
+
+Most of the changes that your code needs to be null safe
+are easily predictable.
+For example, if a variable can be `null`,
+its type needs a `?` suffix.
+A named parameter that shouldn't be nullable
+needs to be marked `required`.
```
leafpetersen, line 139: Maybe link again to the "understanding null safety docs here"?

src/null-safety/migration-guide.md
```diff
@@ -1,23 +1,453 @@
...
+from before null safety:
+
+```dart
+var ints = const <int>[0, null];
+var zero = ints[0];
+var one = zero + 1;
+var zeroOne = <int>[zero, one];
+```
+
+The default migration is backward compatible but not ideal:
```
leafpetersen, line 209: I don't actually see this migration if I run the tool on the code above.  The suggestion I see is:

```
 var ints = const <int?>[0, null];
  var zero = ints[0]!;
  var one = zero + 1;
  var zeroOne = <int >[zero, one];
```

src/null-safety/migration-guide.md
```diff
@@ -1,23 +1,453 @@
...
+  click **Rerun from sources** to pick up your changes.
+
+A single hint can have ripple effects elsewhere in the code.
+In the example from before,
+adding a `/*!*/` marker where `zero` is assigned its value (on line 2)
+makes the migration tool infer the type of `zero` as `int` instead of `int?`.
+This type change can affect code that directly or indirectly uses `zero`.
+
+```dart
+var zero = ints[0]/*!*/;
```
leafpetersen, line 256: Is this actually what the tool will offer?  I'm a little confused here.  I believe there are two kinds of /*!*/ hints: there are the hints that say "infer this variable as non-nullable", and the hints that say "insert a ! operator here".  I think it's a little unfortunate that the same hint is overloaded, but I think this discussion is confusing/conflating the two?  I'm not sure though.  I don't think the tool will suggest that kind of hint here though, will it?  cc @stereotype441 @devoncarew

src/null-safety/migration-guide.md
```diff
@@ -1,23 +1,453 @@
...
+If it's valid, then [test your code](#step4-test).
+Then, if you've published your code on pub.dev,
+[publish a null-safe prerelease](#step5-publish).
+
+
+### Migrating by hand
+
+If you prefer not to use the migration tool,
+you can migrate manually.
+Because Dart supports [mixed-version programs][Unsound null safety],
```
leafpetersen, line 307: This isn't consistent with the workflow below.  It's true you can do this, but not if you follow the workflow below - in order to migrate one library at a time, you have to opt all of the other libraries out explicitly by marking them with `// @dart = 2.9`.  @mit-mit @devoncarew do we really want to go into that here?

src/null-safety/migration-guide.md
stereotype441: I believe you're right: the tool accepts `/*!*/` hints on expressions, but never offers to add them--you have to add them manually.  The only `/*!*/` hints it offers to add on your behalf are on types.

src/null-safety/migration-guide.md
mit-mit: As discussed, we changed out mind and would like to keep "hand migration" here (but can still move mixed-version elsewhere). Resolving.

src/null-safety/migration-guide.md
```diff
@@ -1,23 +1,453 @@
...
+  When all the dev dependencies support null safety,
+  you can _run tests_ with sound null safety.
+  You might also need null-safe dev dependencies for other reasons,
+  such as code generation.
+{{ site.alert.end }}
+
+Here's an example of the output for a simple package.
+The green checkmarked version for each package supports null safety:
+
+![Output of dart pub outdated, showing that examples_util has no releases with null safety support, but pedantic and test do](/null-safety/pub-outdated-output.png)
```
mit-mit, line 109: The image seems outdated? It doesn't show the null safe version of `examples_util` that have in step 2 below.

Also, I think it would be even better if we had both a direct and dev dependencies.

The test case I've been using is:

`dart create sampleapp`

Edit `pubspec.yaml` to:
```
environment:
  sdk: '>=2.8.1 <3.0.0'

dependencies:
  path: ^1.7.0
  process: 3.0.13

dev_dependencies:
  pedantic: ^1.9.0
```

Run `pub get`

Then `dart pub outdated --mode=null-safety` gives me:

```
direct dependencies:
path          ✗1.7.0   ✗1.7.0      ✓1.8.0-nullsafety.3   ✓1.8.0-nullsafety.3
process       ✗3.0.13  ✗3.0.13     ✓4.0.0-nullsafety.4   ✓4.0.0-nullsafety.4

dev_dependencies:
pedantic      ✗1.9.2   ✗1.9.2      ✓1.10.0-nullsafety.3  ✓1.10.0-nullsafety.3
```

src/null-safety/migration-guide.md
mit-mit: +1

src/null-safety/migration-guide.md
mit-mit: Looks good!

src/null-safety/migration-guide.md
mit-mit: I don't think so. I think the present page should focus on sound migration (that is what we preached in Step 1. above), and then cover how to do a sound migration with either the migration tool or the analyzer.

We have a dedicated topic for unsound underway (https://github.com/dart-lang/site-www/pull/2764) so we can cover all the gory details of that here.

Concretely, on the current page, I suggest we remove "Because Dart supports mixed-version programs, you can migrate one library (generally one Dart file) at a time, while still being able to run your app and its tests."

src/_guides/language/evolution.md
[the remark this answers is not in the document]
kwalrath: The mixed version + detailed manual migration content are in #2764.

src/null-safety/migration-guide.md
kwalrath: Ah, the pleasures of documenting tools that are still changing. Both what you see and what I saw are different from what's in https://github.com/dart-lang/sdk/tree/master/pkg/nnbd_migration#interactive-feedback-to-the-tool (even before I changed the code to make it more standard).

src/null-safety/migration-guide.md
kwalrath: Yes, that's what I saw. I had to add this hint by hand.

src/null-safety/migration-guide.md
kwalrath: I was trying to illustrate the point the you can't migrate yet (step 1: wait for examples_util to be migrated). But apparently that isn't clear. Switching a different example sounds good to me; I'll work on that.

kwalrath:
@stereotype441 and/or @leafpetersen could you please review the wording in the table of hint markers: https://kw-www-dartlang-1.firebaseapp.com/null-safety/migration-guide#hint-markers

src/null-safety/migration-guide.md
```diff
@@ -1,23 +1,472 @@
...
+  When you've finished editing your code,
+  click **Rerun from sources** to pick up your changes.
+
+The following table shows the hint markers that you can use
+to change the migration tool's proposed edits.
+
+|------------------+--------------------------------------------------------------------------|
+| Hint marker      | Effect on the migration tool                                             |
+|------------------|--------------------------------------------------------------------------|
+| `/*!*/`          | Mark the preceding expression or type as non-null or non-nullable.       |
```
stereotype441, line 260: This is a little misleading because the meaning of `/*!*/` is quite different depending whether you're applying it to an expression or a type.  If you're applying it to a type it means "don't make this type nullable", which I think is consistent with what you're saying here.  If you're applying it to an expression, it means "add a `!` here when migrating, so that the expression is null checked"; that's really different from marking the expression as non-nullable; in point of fact usually the only reason you would want to add a `!` to an expression is if you expect that the expression's type to be nullable (otherwise there would be no point in null checking it).

I wonder if it would be clearer to break these two cases out into separate entries in the table, e.g.:

| Hint marker | Effect on the migration tool |
|-|-|
| _expression_ `/*!*/` | Ensure that _expression_ is null checked in the migrated output. |
| _type_ `/*!*/` | Mark _type_ as non-nullable. |

src/null-safety/migration-guide.md
```diff
@@ -1,23 +1,472 @@
...
+| `/*?*/`          | Mark the preceding expression or type as nullable.                       |
```
stereotype441, line 261: `/*?*/` can only be applied to types, so this should be "Mark the preceding type as nullable."

src/null-safety/migration-guide.md
kwalrath: Thanks, @stereotype441!

kwalrath: For the first one, we use different terminology in the null safety docs, so I'm thinking of something like this:

> Adds a `!` to the migrated code, which casts _expression_ to its underlying non-nullable type.

stereotype441: > For the first one, we use different terminology in the null safety docs, so I'm thinking of something like this:
> 
> > Adds a `!` to the migrated code, which casts _expression_ to its underlying non-nullable type.

Excellent, that's way better than my suggestion 😃

src/null-safety/migration-guide.md
```diff
@@ -1,23 +1,472 @@
...
+For example, if package C depends on package B, which depends on package A,
+then A should be migrated to null safety first, then B, then C.
+
+![Illustration of C/B/A sentence](/null-safety/null-safety-migration-order.png){:width="454px"}<br>
+
+Although you _can_ migrate before your dependencies support null safety,
+you might have to change your code when your dependencies migrate.
+For example, if you predict that a function will take a nullable parameter but
+the package migrates it to be non-nullable,
+then passing a nullable argument becomes a compile error.
```
jonasfj, line 45: **nit**: Is it perhaps worth pointing out that to run code migrated out-of-order you'll have to:
* (A) have an entrypoint that is opted-out of null-safety (possibly using a _language-version-marker_, e.g. add `// @dart=2.7` to `main.dart`), or,
 * (B) Run the code with `dart --no-sound-null-safety main.dart`.

For this reason package authors are strongly discouraged from publishing packages migrated out-of-order.
Since the downstream consumers of the published packages will be surprised if they have to pass `--no-sound-null-safety`.

This might be the wrong place to write this, maybe there is an FAQ where we can give a longer explanation of the downsides to migration out-of-order (perhaps with small examples of what can go wrong). AFAIK reasons are:
 * (i) `null` values may be able to sneak in places where they are not allowed, trigger run-time errors that wouldn't be possible in a full null-safety code base.
 * (ii) semantics of your dependencies will likely change as they migrate, potentially breaking your code.
 * (iii) users of your package will get a degraded experience because they have to pass `--no-sound-null-safety` (and they will suffer (i) and (ii), which is why this flag is required, as a warning to them).

Note: (iii) is unimportant if you're developing an application, but if you are migrating a package published on `pub.dev` this is a very good reason to postpone migration until dependencies are migrated.

mit-mit: @kwalrath I think this was intended to be covered in https://github.com/dart-lang/site-www/pull/2764? If so, we could perhaps link to that from 'can migrate'?

src/null-safety/migration-guide.md
```diff
@@ -1,23 +1,472 @@
...
+(for example, A imports B which imports C, and C imports A),
+consider migrating those libraries together.
+
+To migrate a package by hand, follow these steps:
+
+1. Modify the SDK constraints in the package's `pubspec.yaml` file,
+   setting the language version to 2.12 beta or later:
+   ```yaml
+environment:
+  sdk: '>=2.12.0-0 <3.0.0'
```
mit-mit, line 353: Tim suggested in the blog post draft that we highlight the `-0` part, which I think is a good idea (this is not something most have seen before). I added a commit for this over on the other page; perhaps we can use something similar here?

https://github.com/dart-lang/site-www/commit/5b32909f51bd43449fc132d5aebf09e17dcea138

src/null-safety/migration-guide.md
kwalrath: Actually... even after updating I get these suggestions. Are you sure you don't with the latest flutter master release?

mit-mit:
@kwalrath, @timsneath left some great "things developers are likely to ask about migration" questions in the blog post. They were too detailed to get into there, but I think here they would be great to answer:

> a) should I just wait if they're not ready?

We mostly answer this in Step 1. Maybe we could add something like "If there is a particular package missing null safety support in your dependency list, perhaps reach out to the package owner to let them know. You can find contact details for a package on pub.dev"

> b) are the Dart / Flutter teams migrating their packages? what's the roadmap / plan?

I believe we covered this in the blog post.

> c) do I have to wait on both direct and dev dependencies?

I believe the docs here cover that.

> d) is there anything I should do to encourage my dependencies to migrate?

See answer in a)

> e) if I migrate my package, will it break downstream dependencies that aren't yet null safe? (link to "incrementally adoptable", obvs).

I think it would be good to cover this! We have the "mixed version" doc, but that mostly covers the question of "can I migrate before my deps". It might be good to let package owners know that they *can* create a null safety version and still allow developers to use that package from an app that doesn't use null safety yet. We could even give the example that this is what we've done with our core libs and the Flutter framework (which is all opted in, but we still allow callers that are null safe or not).

src/null-safety/migration-guide.md
kwalrath: We should probably be more explicit about some of these downsides in #2764, but mit's right, that's where we intend to cover this.

src/null-safety/migration-guide.md
mit-mit: If you want the build that is going out tomorrow, then you should use latest Flutter **dev** channel.

But I get the same results in latest master (Dart `2.12.0-50.0.dev`) and dev (Dart `2.12.0-29.10.beta`):
Screenshot 2020-11-18 at 23 36 24

kwalrath: I wonder if Leaf's putting this code in a function. I have it outside (top level).

kwalrath: Confirmed that I get the same results as Leaf if I move the code into a function.

src/null-safety/migration-guide.md
kwalrath: I've updated the discussion to mention that the proposed edits differ depending on the location of the code.

kwalrath:
> I think it would be good to cover this! We have the "mixed version" doc, but that mostly covers the question of "can I migrate before my deps". It might be good to let package owners know that they _can_ create a null safety version and still allow developers to use that package from an app that doesn't use null safety yet. We could even give the example that this is what we've done with our core libs and the Flutter framework (which is all opted in, but we still allow callers that are null safe or not).

I just added a note about this, so ptal and feel free to tweak it if necessary.

mit-mit (approved):

Issue #2383, opened 2020-04-30
Improve search pages & CSE


kwalrath:
From flutter/website#4000, here are some problems with the search pages on dart.dev & flutter.dev that we'd love some help with:

* ~~When the text field is empty on both /search & /search-all, it's too small. Fixed on dart.dev by #2741.~~
* It'd be nice to copy the search query from /search over to /search-all.
* We should probably tweak the ownership of the CSEs, so a role account is the owner instead of me. (This is the only part that probably requires a Googler.)

Page links:
* https://dart.dev/search
* https://dart.dev/search-all
* https://flutter.dev/search
* https://flutter.dev/search-all

CSE links:
* Flutter.dev: https://cse.google.com/cse?cx=011220921317074318178:qhv1ojyoytz
  ([admin link](https://cse.google.com/cse/setup/basic?cx=011220921317074318178:qhv1ojyoytz))
* Flutter.dev and more: https://cse.google.com/cse?cx=011220921317074318178%3Adeii8vtsimy
  ([admin link](https://cse.google.com/cse/setup/basic?cx=011220921317074318178%3Adeii8vtsimy))
* Dart.dev: https://cse.google.com/cse?cx=011220921317074318178%3Aufsd-ukx1ii
  ([admin link](https://cse.google.com/cse/setup/basic?cx=011220921317074318178%3Aufsd-ukx1ii))
* Dart.dev and more: https://cse.google.com/cse?cx=011220921317074318178%3A_yy-tmb5t_i
  ([admin link](https://cse.google.com/cse/setup/basic?cx=011220921317074318178%3A_yy-tmb5t_i))

micpap25:
@kwalrath Should the Dart /search and /search-all files be simplified down to the way they are in Flutter?

micpap25:
@johnpryan made some changes to the site that fix the first problem you listed, I'm bringing them over to the Flutter site right now.

kwalrath:
> @kwalrath Should the Dart /search and /search-all files be simplified down to the way they are in Flutter?

Yes, I think so. When I updated the Flutter search, I just copy-pasted what was in the custom search console. Going back to the console, here's what it says for the dart.dev-only search:

```
<script async src="https://cse.google.com/cse.js?cx=011220921317074318178:ufsd-ukx1ii"></script>
<div class="gcse-search"></div>
```

And for the "more sites" search:

```
[...]

PR #2741, opened 2020-11-07
Improve search pages

micpap25:
Related to #2383 

Restructures the code to make it simpler and match that of the Flutter site. Maintains the current CSEs. 

Not sure if this the issue should be closed or not, because there are other changes listed on the issue that need to be made and I can't tell if they've already been done.


kwalrath (approved):
I staged it locally, and it seems to work great. Thank you, @micpap25!

PR #2742, opened 2020-11-08
Move diagrams out of src

### LINKED ISSUE #1920: Move `diagrams` out of `src`
The [diagrams](https://github.com/dart-lang/site-www/tree/master/src/diagrams) folder contains Graffle sources and other resources. None of these files are directly used by Jekyll to generate the site. 

(Jekyll has an `exclude` feature, but IMHO, that's for files and directories that we'd have a hard time keeping out of `src`.)

micpap25:
Fixes #1920 
Moves `diagrams` out of `src` and removes the now unnecessary exclusions of it. 

I don't see where `diagrams` actually gets used on the site, so hopefully this doesn't break anything.


kwalrath:
> Fixes #1920
> Moves `diagrams` out of `src` and removes the now unnecessary exclusions of it.

Thank you for doing this! It's one of those things I've wanted to do for ages, but never got around to.

> I don't see where `diagrams` actually gets used on the site, so hopefully this doesn't break anything.

It isn't used anywhere, really. It's just source files that we use to generate annotated images and diagrams.

@sfshaza2 I've filed https://github.com/flutter/website/issues/4974 to remind us to do the same on flutter/website.

kwalrath (approved):

PR #2744, opened 2020-11-10
More direct language in language tour

### LINKED ISSUE #2718: "//uncommenting this causes an error" in Language Tour
Page: https://dart.dev/guides/language/language-tour#lists
Page Source: https://github.com/dart-lang/site-www/blob/master/src/_guides/language/language-tour.md

Issue: There are a few instances where a code example has a section that says "uncommenting this causes an error". This is pretty obviously misleading and could cause confusion.

micpap25:
Fixes #2718
A bit of a nit-pick on my part, uses more direct language to make it clear to users that the commented lines are examples of what *not* to do when working with constants.

### DIFF
--- a/examples/misc/lib/language_tour/built_in_types.dart
+++ b/examples/misc/lib/language_tour/built_in_types.dart
@@ -99,7 +99,7 @@ void miscDeclAnalyzedButNotTested() {
   {
     // #docregion const-list
     var constantList = const [1, 2, 3];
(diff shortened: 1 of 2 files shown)


kwalrath (approved):
We usually avoid "will", and I suspect we could still be clearer... but I don't think this change hurts the doc. And maybe it'll help!

PR #2747, opened 2020-11-11
Forward /go/null-safety-migration to draft doc

kwalrath:
I think we're ready to get input from more people on this doc (https://kw-www-dartlang-1.firebaseapp.com/null-safety/migration-guide). If you'd still like people to have access to https://github.com/dart-lang/sdk/tree/master/pkg/nnbd_migration, I can add a link to the draft doc.

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -128,7 +128,7 @@
       { "source": "/flutter", "destination": "https://flutter.io", "type": 301 },
 
       { "source": "/go/experiments", "destination": "/tools/experiment-flags", "type": 301 },
-      { "source": "/go/null-safety-migration", "destination": "https://github.com/dart-lang/sdk/tree/master/pkg/nnbd_migration", "type": 301 },
+      { "source": "/go/null-safety-migration", "destination": "https://kw-www-dartlang-1.firebaseapp.com/null-safety/migration-guide", "type": 302 },
       { "source": "/go/test-docs/:page*", "destination": "https://github.com/dart-lang/test/blob/master/pkgs/test/doc/:page*", "type": 301 },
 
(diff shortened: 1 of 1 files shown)


kevmoo (approved):

devoncarew (commented):
👍 

For context on where this is shown, it's the url displayed by `dart migrate` for the migration guide (so, as well as being the general migration guide, the target of this url should also mention the migration tool).

kwalrath:
> 👍
> 
> For context on where this is shown, it's the url displayed by `dart migrate` for the migration guide (so, as well as being the general migration guide, the target of this url should also mention the migration tool).

It mentions it a lot. :)

firebase.json
```diff
@@ -128,7 +128,7 @@
       { "source": "/flutter", "destination": "https://flutter.io", "type": 301 },
 
       { "source": "/go/experiments", "destination": "/tools/experiment-flags", "type": 301 },
-      { "source": "/go/null-safety-migration", "destination": "https://github.com/dart-lang/sdk/tree/master/pkg/nnbd_migration", "type": 301 },
+      { "source": "/go/null-safety-migration", "destination": "https://kw-www-dartlang-1.firebaseapp.com/null-safety/migration-guide", "type": 302 },
```
devoncarew, line 131: I'm curious if there was a reason to change this from a 301 to a 302 (the other redirects in this file use a 301).

kwalrath: 302 redirects are supposed to be temporary, so I used that because we'll change the target soon. It's my (perhaps misguided) hope that search engines might be more likely to pick up a change to a 302 redirect.

PR #2750, opened 2020-11-12
Remove Unused Images

### LINKED ISSUE #1789: Make sure image files are used
More info: https://github.com/dart-lang/site-www/pull/1780#issuecomment-514753731

We might have other image files that are unused... it wouldn't be bad to do a pass to find those.

micpap25:
Fixes #1789
Removes all images from `src/assets/dash` that aren't used on the website.
I think it's 100% fair if not all these images are removed; that's a discussion that might need to happen, because there might be future development of the site that uses these images.


kwalrath:
Thank you, @micpap25!

@RedBrogdon do you think it's safe to remove all these images? (Want to review?)

kwalrath:
Sorry for the delay, @micpap25. We've been a little distracted by getting null safety beta out the door! I've asked Andrew if he's able to review this.

kwalrath:
I'm going to close this because #2889 is covering the same territory. (I believe the 1x images are used, after all... but we might get rid of them in favor of SVGs everywhere.)

Issue #1949, opened 2019-09-25
automate updates of the diagnostic messages page


kwalrath:
See #1892 for details

kwalrath:
Here's the process I'm doing by hand:

1. Create a new branch of my site-www repo.
2. Go to https://github.com/dart-lang/sdk/blob/master/pkg/analyzer/tool/diagnostics/diagnostics.md, click the **Raw** button, and **Save as** _pathToRepo_/src/tools/diagnostic-messages.md.
3. View the diffs, and make sure they look reasonable.
4. Inform @bwilkerson of any issues. E.g., `title: Diagnostics` should be `title: Diagnostic messages`. Maybe edit the file if I don't feel like waiting.
5. Once the file is ready, create a PR.

kwalrath:
Thinking about how to automate this, I'm not positive it should be automated, but perhaps it'd be good to have a build check (failable) about whether this file should be updated.

chalin:
Yes, that is what I had in mind; similar to what we do with the code-excerpt refresh. I.e., actual code refresh is under site-maintainer control (and done manually using a single command), but the build scripts ensure that everything is up-to-date.

parlough:
@devoncarew I'm not sure if you're the right person to ask, but I know you've been working with automation a bit recently. Let me know if I should direct my question elsewhere.

We're expecting the source file to change, perhaps to one or more YAML files, but we'd still like to automatic moving of 1+ diagnostic related files from the SDK to this repository in our work to better standardizing linter rule documentation as well as diagnostics.

I was thinking of using a GitHub workflow which checks for updates once a day, then if there are differences, open a new branch and pull request with the changes. Then use the existing autosubmit logic to automatically merge in the changes if all tests pass.

Does this make sense? If so, does Google have any rules for GitHub bots, particularly in relation to the CLA? Can I use the default github-actions bot, should I use one that the Dart team already has, work with the devrel team to set up a new one? 

Thanks so much for any thoughts you have!

parlough:
Thanks for your thoughts!

I'd like to avoid the check that fails builds, since we've had a few in the past and they slow down writers.

The idea of shipping it alongside the SDK may be interesting and I'll consider that as we figure out what the files exactly look like and include. Another similar options is including the SDK as a submodule, that I've been mostly avoiding, but the PRs come easily with dependabot and may be useful for other use cases as well.

antfitch:
[...]

PR #2753, opened 2020-11-12
Typo fix

faustodc:
@bwilkerson Completed missing sentence in diagnostic-messages.md following #1949 steps.

### DIFF
--- a/src/tools/diagnostic-messages.md
+++ b/src/tools/diagnostic-messages.md
@@ -5644,7 +5644,7 @@ class C {
 }
 {% endprettify %}
 
-If the instances of the class should be mutable, then remove the
+If the instances of the class should be mutable, then remove the `@immutable` annotation:
 
 {% prettify dart tag=pre+code %}
 class C {
--- a/src/tools/diagnostic-messages.md
+++ b/src/tools/diagnostic-messages.md
@@ -87,12 +87,12 @@ when it’s passed as an argument to `print`:
 
 ```dart
 void f() {
-  late String s;
+  String s;
   print(s);
 }
 ```
 
-But in the following code, the variable `s` is definitely assigned: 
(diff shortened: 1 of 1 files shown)


src/tools/diagnostic-messages.md
```diff
@@ -5644,7 +5644,7 @@ class C {
 }
 {% endprettify %}
 
-If the instances of the class should be mutable, then remove the
+If the instances of the class should be mutable, then remove the `@immutable` annotation:
```
bwilkerson, line 5647: Thanks. I dropped a space in the source code from which this was generated which kept a whole line of text from being included in the generated output. The original reviewed text was:

If the instances of the class should be mutable, then remove the annotation, or choose a different superclass if the annotation is inherited:

The original source code will be fixed by https://dart-review.googlesource.com/c/sdk/+/171810. If you could apply the fix manually, that would be much appreciated.

bwilkerson (approved):

bwilkerson:
@kwalrath This lgtm, but I'm not sure what the process is here and whether you want to be involved.

kwalrath:
I'm happy to wait on the sidelines unless @faustodc needs my help.

src/tools/diagnostic-messages.md
faustodc: Great, I updated it with the original source code comment.

bwilkerson (approved):
Thanks!

bwilkerson:
@kwalrath I don't have the ability to merge the PR.

kwalrath (approved):
Looks great. Thank you, @faustodc and @bwilkerson!

faustodc:
Thank you for your work @kwalrath  and @bwilkerson 😊

PR #2764, opened 2020-11-18
[DONT MERGE] Create unsound-null-safety.md

kwalrath:
Staged: https://kw-dartlang-3.firebaseapp.com/null-safety/unsound-null-safety

I wasn't able to satisfactorily test the `--no-sound-null-safety` instructions (at the bottom) — in particular, `dart test` didn't seem to accept that flag — so I'd appreciate special attention to that section.

We should probably add this to the sidenav... but that can happen later.

### DIFF
--- a/src/null-safety/unsound-null-safety.md
+++ b/src/null-safety/unsound-null-safety.md
@@ -1,0 +1,186 @@
+---
+title: Unsound null safety
+description: Mixing language versions lets you migrate to null safety at your own pace, with some of the benefits of null safety.
+---
+
+An app can use some libraries that
+are [null safe][] and some that aren't.
+These apps — called **mixed-version programs** —
+execute with **unsound null safety**.
+
+[null safe]: /null-safety
+
+Mixing language versions lets you migrate to null safety at your own pace.
+It also frees package maintainers to migrate their code,
+with the knowledge that even legacy users can get new
+bug fixes and other improvements.
+However, mixed-version programs don't get all the advantages
+that null safety can bring.
+
+This page describes the differences between sound and unsound null safety,
+to help you decide whether to migrate before
+all your dependencies support null safety.
+After the conceptual discussion are instructions for migrating incrementally,
+followed by details on analyzing, testing, and running mixed-version programs.
+
+{{ site.alert.note }}
+  We recommend that, if possible, you wait for dependencies to migrate
+  before you migrate your app.
+  For details, see the [migration guide][].
+{{ site.alert.end }}
+
+[migration guide]: /null-safety/migration-guide
+
+
+## Sound and unsound null safety
+
+Dart provides sound null safety through a combination of
+static and runtime checks.
+Each Dart library that opts in to null safety gets
+all the _static_ checks, with stricter compile-time errors.
+This is true even in a mixed-version program that contains
+null-unsafe libraries.
+You start getting these benefits
+as soon as you start migrating some of your code to null safety.
+
+However, a mixed-version program can't have the
+_runtime_ soundness guarantees that a fully null-safe app has.
+It's possible for `null` to leak out of the null-unsafe libraries
+into the null-safe code, because
+preventing that would break the existing behavior of the unmigrated code.
+
+To maintain runtime compatibility with legacy libraries
+while offering soundness to completely null-safe programs,
+Dart tools support two modes:
+
+* When you run a mixed-version program,
+  it runs with **unsound null safety**.
+  It's possible for `null` reference errors to occur at runtime,
+  but only because a `null` or nullable type escaped from
+  some null-unsafe library and got into your null-safe code.
+
+* When your program is fully migrated and _all_ libraries are null safe,
+  then your program automatically runs with **sound null safety**.
+  That mode gives you the full null safety experience with
+  all of the guarantees and compiler optimizations that soundness enables.
+
+Sound null safety is what you want if possible.
+Dart tools automatically run your program in sound mode if
+the main entrypoint library of your program has opted into null safety.
+If you import a null-unsafe library,
+the tools print a warning to let you know that
+they can only [run with unsound null safety](#analyzing-and-testing).
+
+{{ site.alert.info }}
+  Except for a few edge cases,
+  a mixed-version program should do exactly what you expect.
+  One example of an edge case is an `is` test
+  on an instance of a generic type from a closure
+  that crosses the boundary between null-safe and null-unsafe libraries.
+{{ site.alert.end }}
+
+
+## Migrating incrementally
+
+The [migration tool][] is aimed at migrating a whole package at once.
+If you want to migrate your package piecemeal,
+the migration tool can still help,
+but you'll need to do more by hand.
+
+[migration tool]: /null-safety/migration-guide#step2-migrate
+
+Because Dart supports mixed-version programs,
+you can migrate one library (generally one Dart file) at a time,
+while still being able to run your app and its tests.
+
+We recommend that you **first migrate leaf libraries** —
+libraries that don't import other files from the package.
+Then migrate libraries that directly depend on the leaf libraries.
+End by migrating the libraries that have the most
+intra-package dependencies.
+
+For example, say you have a `lib/src/util.dart` file
+that imports other (null-safe) packages and core libraries,
+but that doesn't have any `import '<local_path>'` directives.
+Consider migrating `util.dart` first,
+and then migrating simple files that depend only on `util.dart`.
+If any libraries have cyclic imports
+(for example, A imports B which imports C, and C imports A),
+consider migrating those libraries together.
+
+To migrate a package by hand, follow these steps:
+
+1. Modify the SDK constraints in the package's `pubspec.yaml` file,
+   setting the language version to 2.12 beta or later:
+   ```yaml
+environment:
+  sdk: '>=2.12.0-0 <3.0.0'
+```
+
+2. Update dependencies: <br>
+   ```terminal
+$ dart pub get
+```
+
+   Running `dart pub get` with a lower constraint of 2.12.0
+   sets the default language version of
+   every library in your package to 2.12,
+   opting them all in to null safety.
+
+
+3. Open the package in your IDE. <br>
+   You're likely to see a lot of analysis errors.
+   That's OK.
+
+4. _Optional:_ Add a [language version comment][] to the top of
+   any Dart files that you don't want to consider during your current migration:
+   ```dart
+// @dart=2.9
+```
+
+   Using language version 2.9 for a library that's in a 2.12 package
+   can reduce analysis errors (red squiggles) if you disable sound null safety.
+   However, **unsound null safety reduces the
+   information the analyzer can use.**
+   For example, the analyzer might assume a
+   parameter type is non-nullable,
+   even though a 2.9 file might pass in a null value.
+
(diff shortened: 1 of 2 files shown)


src/null-safety/unsound-null-safety.md
```diff
@@ -0,0 +1,186 @@
...
+
+To analyze, test, or run mixed-version code,
+you need to disable sound null safety.
+You can do this in two ways:
+
+* Disable sound null safety using the `--no-sound-null-safety` flag.
+  Example:
+
+  ```terminal
+  $ dart run --no-sound-null-safety
```
mit-mit, line 166: The flag needs to be before the verb, so:  `dart --no-sound-null-safety run`

mit-mit: See also https://github.com/dart-lang/sdk/issues/44022

mit-mit:
With this new topic, I think we should consider removing the "Per-library language version selection" section from https://dart.dev/guides/language/evolution#per-library-language-version-selection

src/null-safety/unsound-null-safety.md
```diff
@@ -0,0 +1,186 @@
...
+
+An app can use some libraries that
+are [null safe][] and some that aren't.
+These apps — called **mixed-version programs** —
+execute with **unsound null safety**.
+
+[null safe]: /null-safety
+
+Mixing language versions lets you migrate to null safety at your own pace.
+It also frees package maintainers to migrate their code,
```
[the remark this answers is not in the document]
kwalrath: I think this is more or less from @munificent (although I probably changed it, perhaps changing the meaning), so maybe he can help us clear it up.

src/null-safety/unsound-null-safety.md
```diff
@@ -0,0 +1,186 @@
...
+
+Sound null safety is what you want if possible.
+Dart tools automatically run your program in sound mode if
+the main entrypoint library of your program has opted into null safety.
+If you import a null-unsafe library,
+the tools print a warning to let you know that
+they can only [run with unsound null safety](#analyzing-and-testing).
+
+{{ site.alert.info }}
+  Except for a few edge cases,
```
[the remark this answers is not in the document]
kwalrath: Always happy to delete text!

src/null-safety/unsound-null-safety.md
```diff
@@ -0,0 +1,186 @@
...
+   That's OK.
+
+4. _Optional:_ Add a [language version comment][] to the top of
+   any Dart files that you don't want to consider during your current migration:
+   ```dart
+// @dart=2.9
+```
+
+   Using language version 2.9 for a library that's in a 2.12 package
+   can reduce analysis errors (red squiggles) if you disable sound null safety.
```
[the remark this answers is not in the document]
kwalrath: I'm happy to delete that text, as it doesn't seem necessary, but you seem to be implying that soundness is automatically disabled. The command-line doesn't do that — it makes you specify the flag yourself. Do IDEs automatically disable sound null safety?

kwalrath:
> With this new topic, I think we should consider removing the "Per-library language version selection" section from https://dart.dev/guides/language/evolution#per-library-language-version-selection

That probably makes sense, but I'd like to do it later. I'll file an issue.

src/null-safety/unsound-null-safety.md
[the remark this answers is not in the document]
kwalrath: Actually, the command-line `dart run` didn't seem to need the flag in my very simple test. So... when *do* you need the flag?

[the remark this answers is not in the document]
mit-mit: Sorry, my words were confusing. Running any app that has a >=2.12 SDK constraint and mixed-versions (i.e. at least one opt-ed out library) requires either 1) the `--no-sound-null-safety` flag, **or**, 2) a `<2.12` language version comment in the **entry point**.

src/null-safety/unsound-null-safety.md
```diff
@@ -0,0 +1,178 @@
...
+   That's OK.
+
+4. Add a [language version comment][] to the top of
+   any Dart files that you don't want to consider during your current migration:
+   ```dart
+// @dart=2.9
+```
+
+   Using language version 2.9 for a library that's in a 2.12 package
+   can reduce analysis errors (red squiggles).
```
mit-mit, line 135: Maybe extend a bit here, like "can reduce analysis errors (red squiggles) coming from unmigrated code." ?

src/null-safety/unsound-null-safety.md
```diff
@@ -0,0 +1,178 @@
+---
+title: Unsound null safety
+description: Mixing language versions lets you migrate to null safety at your own pace, with some of the benefits of null safety.
+---
+
+An app can use some libraries that
```
mit-mit, line 6: Packages can do this too. Given we're focusing on package migration for beta, perhaps clarify that here.

src/null-safety/unsound-null-safety.md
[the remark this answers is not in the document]
mit-mit: @munificent can you clarify here? Here are the parts that confuse me:

> Mixing [language versions][] lets you migrate to null safety at your own pace.

Yup, this part I agree with. Perhaps "own pace" is a bit imprecise; I think the core value in mixed versions is that you can migrate parts of a package or app rather than all of it at once?

> It also frees package maintainers to migrate their code, with the knowledge that even legacy users can get new bug fixes and other improvements.

This part I don't get. How does mixed mode "free" the maintainer? Is this about it enabling them to migrate "out of order", i.e. migrate before their dependencies have migrated to null safety?

> However, mixed-version programs don't get all the advantages that null safety can bring.

Yup, this is good.

src/null-safety/unsound-null-safety.md
```diff
@@ -0,0 +1,178 @@
...
+
+5. Migrate the code of each Dart file,
+   using the analyzer to identify static errors. <br>
+   Eliminate static errors by adding `?`, `!`, `required`, and `late`,
+   as needed.
+
+
+## Analyzing, testing, and running mixed-version programs
+
+To analyze, test, or run mixed-version code,
```
munificent, line 150: Sound null safety is a mode for *running or compiling* code, but not a static property of the program itself, so when just analyzing it doesn't come into play. (Think of it like debug and release mode. Your IDE doesn't care whether you're running in debug or release. As far as it's concerned it's all just code.)

I'd just removing "analyzing" from here and the subheader.

src/null-safety/unsound-null-safety.md
```diff
@@ -0,0 +1,178 @@
...
+
+Sound null safety is what you want if possible.
+Dart tools automatically run your program in sound mode if
+the main entrypoint library of your program has opted into null safety.
+If you import a null-unsafe library,
+the tools print a warning to let you know that
+they can only [run with unsound null safety](#analyzing-and-testing).
+
+
+## Migrating incrementally
```
munificent, line 77: Is there a reason we document incremental manual migration here, but not using the migration tool?

src/null-safety/unsound-null-safety.md
```diff
@@ -0,0 +1,178 @@
...
+  Example:
+
+  ```terminal
+  $ dart --no-sound-null-safety run
+  ```
+
+* Alternatively, set the language version in the entrypoint —
+  the file that contains `main()` function — to 2.9.
+  In Flutter apps, this file is often named `lib/main.dart`.
+  In command-line apps, this file is often named `bin/<packageName>.dart`.
```
munificent, line 164: Probably good to mention that all of your test files under `test/` are entrypoints too and can be opted out this way.

src/null-safety/unsound-null-safety.md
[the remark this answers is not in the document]
mit-mit: Ah, I see what you mean. Mixed-version comes into play in two use cases:

1. A developer migrates a package, all the way to sound nulls safety. Because of mixed-version support, this package can be used _not only_ by apps that themselves use null safety, but also by "legacy" apps that don't yet use null safety. This is a form of mixed version, but it's perhaps easier to understand as backwards compatibility.

   This is a *good way* of using null safety.

2. A developer has a package foo with two deps, bar and baz. Bar has migrated to null safety, but baz hasn't. The developer doesn't feel like waiting on baz, so they migrate their package foo out of order. This is also a form of mixed version.

   This is a potentially more troublesome way of using null safety. You can do it, but we should recommend against it. The core issue is that once that last dependency baz migrates, the run-time checking from soundness might be breaking to consumers of **bar**.

Did I get that right?

src/null-safety/unsound-null-safety.md
kwalrath: I'm not sure... I think we wanted to add better support for incremental migration to the tool, so maybe that's why.

src/null-safety/unsound-null-safety.md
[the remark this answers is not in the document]
kwalrath: I modified the text a bit to simplify it and focus on the case of migrating from the leaf dependencies. ptal

PS: I should probably add the info from https://github.com/dart-lang/site-www/pull/2739#discussion_r526430314, but maybe not in the first published version.

src/null-safety/unsound-null-safety.md
```diff
@@ -0,0 +1,178 @@
...
+To migrate a package by hand, follow these steps:
+
+1. Modify the SDK constraints in the package's `pubspec.yaml` file,
+   setting the language version to 2.12 beta or later:
+   ```yaml
+environment:
+  sdk: '>=2.12.0-0 <3.0.0'
+```
+
+2. Update dependencies: <br>
```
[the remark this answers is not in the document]
kwalrath: I'll update the migration guide, too.

src/null-safety/unsound-null-safety.md
```diff
@@ -0,0 +1,179 @@
...
+
+  ```dart
+  // @dart=2.9
+  import 'src/my_app.dart';
+
+  main() {
+    //...
+  }
+  ```
+
```
leafpetersen, line 176: I think it's worth re-emphasizing here that this is not a desirable end state.  I'd say something like:

```
While opting out your test files may be useful to allow you to run tests **during** your incremental migration process, doing so means that you won't be testing your code with full null safety enabled.  Therefore, it's important to opt your tests back in once you have finished the incremental migration of your libraries.
```

src/null-safety/unsound-null-safety.md
leafpetersen: Yeah, the tool doesn't really support this yet, except via a really kludgy workflow.  I think it's better to follow up here later if/when the tool gets support for this.  cc @stereotype441

leafpetersen (approved):
LGTM

src/null-safety/unsound-null-safety.md
mit-mit: I'll do this in a follow-up PR

mit-mit (approved):

src/null-safety/unsound-null-safety.md
mit-mit: https://github.com/dart-lang/site-www/pull/2770

src/null-safety/unsound-null-safety.md
stereotype441: FYI, incremental migration work is being tracked at https://github.com/dart-lang/sdk/issues/44124

PR #2768, opened 2020-11-18
Use Class Members Language Tour Fix

### LINKED ISSUE #2627: Suggestion change 'Using class members code example'
Hi, sorry for disturbing you guys but the code exemple shown on [Using class members](https://dart.dev/guides/language/language-tour#using-class-members) is a bit confusing, using the [Point](https://api.dart.dev/stable/2.9.2/dart-math/Point-class.html) class as an exemple give me an error because we can't assign another value to a final

```
// Set the value of the instance variable y.
p.y = 3; // The final variable 'y' can only be set once.
```

I know it isn't a big deal, but it'd nice to have an easy working example.
Cheers!

micpap25:
Fixes #2627
Changes a section in the Language Tour so the example no longer tries to change the value of a final variable.

### DIFF
--- a/examples/misc/test/language_tour/classes_test.dart
+++ b/examples/misc/test/language_tour/classes_test.dart
@@ -43,7 +43,7 @@ void main() {
   test('object-members', () {
     // #docregion object-members
     var p = Point(2, 2);
-      
+
     // Get the value of y.
     assert(p.y == 2);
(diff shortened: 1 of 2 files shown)


kwalrath (approved):
Thank you, @micpap25!

PR #2769, opened 2020-11-19
Update firebase.json to create a go/dart-fix redirect

devoncarew:
- update firebase.json to create a `dart.dev/go/dart-fix` redirect

@kwalrath for review

@pq @bwilkerson - this depends on having a placeholder `pkg/dartdev/doc/dart-fix.md` file to point to; we can update this re-direct when there's something more official to use


kwalrath (approved):
LGTM once the target exists... or maybe before, if you need it now?

devoncarew:
Thanks for the review - this should be safe to merge now (I don't have rights).

PR #2775, opened 2020-11-19
Mention that the migration tool deletes hint markers

kwalrath:
Clarification suggested [via tweet](https://twitter.com/programmer_just/status/1329548356331446278)


leafpetersen (approved):
Thanks!

PR #2782, opened 2020-11-26
Clarify wording for bang operator in all modes

natebosch:
The `!` operator may be used on values that aren't strictly "user
passed". Expand the wording to make it clear that the expression might
not only be an argument.


domesticmouse (approved):

ryanmcginnis (approved):
thanks! i've updated the internal FAQ.

kwalrath:
Thanks, everyone!

PR #2792, opened 2020-12-07
Add clarification for null-safe prereleases

johnpryan:
This resolves a question where it was incorrectly assumed that Pub requires that you publish a semantic version higher than the highest published version.

cc: @kevmoo


kwalrath (approved):
LGTM if my wording tweaks LGTY, @johnpryan.

johnpryan (commented):
looks good.

PR #2793, opened 2020-12-08
add docs of static method

Abhishek01039:
I have added some Documentation of static method.

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -3536,9 +3536,11 @@ Static variables aren’t initialized until they’re used.
 
 #### Static methods
 
-Static methods (class methods) do not operate on an instance, and thus
-do not have access to `this`. Static method can be invoked directly on a class, and can access static data member.
-For example:
+Static methods (class methods) don't operate on an instance, and thus
+don't have access to `this`.
+They do, however, have access to static variables.
+As the following example shows,
+you invoke static methods directly on a class:
 
 <?code-excerpt "misc/lib/language_tour/classes/point_with_distance_method.dart"?>
 ```dart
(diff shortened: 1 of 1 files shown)


Abhishek01039:
@mit-mit 
can you please review this PR?
Thanks.

mit-mit:
Isn't this part (calling static methods) already covered right above the code snippet? cc @kwalrath

Abhishek01039:
This part is covered by the code snippet.
but, is this not better to have well-documented these things also?
Thanks.

mit-mit:
@kwalrath is a better wordsmith than I, but my point was that I think this belongs in the paragraph above the code snippet. That currently has:

```
Static methods (class methods) do not operate on an instance, and thus do not have access to this. For example:
```

Perhaps it could have:

```
Static methods (class methods) do not operate on an instance, and thus do not have access to this. 
Static method can be invoked directly on a class, and can access can access static data member.
For example:

```

Abhishek01039:
Yeah 
Thanks, @mit-mit

kwalrath:
@leafpetersen could you (or someone) review this for correctness and terminology?

kwalrath (approved):
I'm going to say this is correct, since it matches what my tests have shown.

Abhishek01039:
Thanks @kwalrath

kwalrath:
> Thanks @kwalrath

Thank _you_ @Abhishek01039!

src/_guides/language/language-tour.md
```diff
@@ -3536,8 +3536,11 @@ Static variables aren’t initialized until they’re used.
 
 #### Static methods
 
-Static methods (class methods) do not operate on an instance, and thus
-do not have access to `this`. For example:
+Static methods (class methods) don't operate on an instance, and thus
+don't have access to `this`.
```
leafpetersen, line 3540: You could also comment on the fact that they can't reference generic parameters to the class (or not).

leafpetersen (commented):
LGTM

Issue #2790, opened 2020-12-06
Code sample needs format/description improvements


emilieroberts:
The code sample for [Do Annotate When Dart Infers The Wrong Type](https://dart.dev/guides/language/effective-dart/design#do-annotate-when-dart-infers-the-wrong-type) has the same code sample twice, once with green "do", and the second with grey. It follows a long list of green "dos" and red "don'ts" so it is confusing because the code is identical.

![Screenshot 2020-12-06 at 1 05 50 PM](https://user-images.githubusercontent.com/8290920/101279633-d2d11400-37c3-11eb-90d3-167922af41fb.png)

Abhishek01039:
@emilieroberts 
Please see
``` num highest = 0;```
 and ```var highest = 0;```
It is different in both examples.


Thanks

emilieroberts:
Wow, ok, that is really not clear. Maybe this could be improved in the
following ways:
- fix the box formatting so it says "don't" and is red like the other code
samples
- mention in the proceeding paragraph what the dev is supposed to be
looking for, eg. "notice the type of the variable 'highest' in the
following sample"
- and/or use colour to highlight the important line or "var/num"
declaration that is different.

Il mar 8 dic 2020, 05:35 Abhishek Ghaskata <notifications@github.com> ha
scritto:

munificent:
Duplicate of #2132.

PR #2794, opened 2020-12-10
change the code snippet

Abhishek01039:
fixes https://github.com/dart-lang/site-www/issues/2790

### DIFF
--- a/src/_guides/language/effective-dart/design.md
+++ b/src/_guides/language/effective-dart/design.md
@@ -1273,7 +1273,7 @@ can later assign some other sibling type to the variable:
 <?code-excerpt "misc/lib/effective_dart/design_good.dart (inferred-wrong)"?>
 {% prettify dart tag=pre+code %}
 num highScore(List<num> scores) {
-  num highest = 0;
+  <span style="color: green;">num highest = 0;</span>
   for (var score in scores) {
     if (score > highest) highest = score;
   }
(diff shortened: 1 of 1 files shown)


Abhishek01039:
@kwalrath 
are these colors is ok or should I use a different color?
Thanks

kwalrath:
Unfortunately, that approach won't work because we need to compare against working code excerpts.

I'd probably use a combination of comments (duplicated in the .md file and the code excerpts' .dart files) and moving down the "bad" code marker for the second example. And... looking closer... there's the issue that the "bad" code doesn't actually work, which is why we disconnected it from the code excerpt system. Maybe I shouldn't have marked this as beginner. :) 

Can you come up with a "bad" excerpt that passes analysis but illustrates the problem? If not, then maybe this rule is no longer needed?

kwalrath:
/cc @munificent 
The affected rule is https://dart.dev/guides/language/effective-dart/design#do-annotate-when-dart-infers-the-wrong-type.

Abhishek01039:
ok @kwalrath 
I will find the example and add it here!
Thanks

kwalrath:
Actually... We need to stop work on this page. @munificent is about to make a pass through Effective Dart, so we need to avoid modifying those docs for a while.

Sorry for the churn! Maybe you could work on one of the pages under #2617?

Issue #2719, opened 2020-10-27
Change URLs for command-line tools being replaced by `dart`


kwalrath:
We'll follow a variant of the pub precedent but NOT update any pub URLs: (/tools/pub/cmd/pub-outdated, e.g.). Here are the new URLs we'll use:

* [x] /tools/dart-analyze (redirect from /tools/dartanalyzer)
* [x] /tools/dart-compile (redirect from /tools/dart2native and, eventually from /tools/dart2js)
* [x] /tools/dart-create 
* [x] /tools/dart-format (redirect from /tools/dartfmt) (#2982)
* [x] /tools/dart-pub (created for URL guessers; redirect **to** /tools/pub/cmd) (#3367)
* [x] /tools/dart-run (redirect from /tools/dart-vm and probably /tools/pub/cmd/pub-run) (#3154)
* [x] /tools/dart-test (replaces `pub run test`) (#3369)

Other notes:
* Should probably redirect from /tools/dart-vm to /tools/dart-run.
* We'll need to do a search for the old command names _and_ for `dart-tool`, so we can update links. 
* We'll keep the old command name on the page, for searchability and to help users who find/know about the old name.
* No change (for now, at least) to the following pages under /tools/:
    * build_runner
    * dart-devtools
    * dartaotruntime (deliberately a standalone file; the smallest VM we have for running code, usable only on the output of `dart compile aot-snapshot`)
    * dartdevc (not sure whether this will stay a separate command; /cc @kevmoo)
    * dartdoc (verb still to be added)
    * dartpad
    * webdev

kwalrath:
I'm starting work on the dart-test page and dart-pub redirect.

PR #2803, opened 2020-12-15
Added command for dart pub

Abhishek01039:
Added `dart pub` in the place of `pub`.

### DIFF
--- a/src/tools/pub/cmd/pub-run.md
+++ b/src/tools/pub/cmd/pub-run.md
@@ -1,12 +1,12 @@
 ---
-title: dart pub run
-description: Use dart pub run to run a Dart script in your package.
+title: dart run
+description: Use dart run to run a Dart script in your package.
 ---
 
 _Run_ is one of the commands of the [pub tool](/tools/pub/cmd).
 
 {% prettify nocode tag=pre+code %}
-$ dart pub run [--enable-asserts] <executable> [args...]
+$ dart run [--enable-asserts] <executable> [args...]
 {% endprettify %}
 
 Use this command to run a Dart script in your package,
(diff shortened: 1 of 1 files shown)


Abhishek01039:
@kwalrath 
can you please review this PR?
Thanks

kwalrath:
I wonder... Can we just use `dart run` instead of `dart pub run`? (Have you tried it?)

/cc @mit-mit

mit-mit:
Yes, you'll want `dart run`, not `dart pub run`.

The only exception is `pub run test`, which is just `dart test`.

Abhishek01039:
ohh ! yes
@kwalrath 
my bad !

Abhishek01039:
can you please review now ? @kwalrath

kwalrath:
I'm closing this PR because we're going to need some more changes — removing most mentions of the `pub` tool, referring back to the dart-tool page (which is something I need to add to my PR #2804), changing the filename, and adding a redirect. I'll take care of it as part of my work on #2719 (which I've updated). I've also updated #2617 to reflect what we've learned about pub-run.md.

@Abhishek01039, your work helped us realize the need to transition this page over to the dart-tool work, so thank you for that, and sorry we couldn't use this PR!

Would you like to work on another page in #2617? (If so, I'll update #2617 to assign your chosen page to you; it didn't matter this time, but we should use that bug to keep track of who's working on what.)

Abhishek01039:
Ah! ok, Thanks @kwalrath 
 I will look into this.

kwalrath:
> Ah! ok, Thanks @kwalrath
> I will look into this.

Thanks for understanding!

PR #2805, opened 2020-12-15
Iterables codelab example migration to Null Safety

miquelbeltran:
Hi! :wave: I am working with @RedBrogdon on migrating documentation to null safety, first focusing on codelabs.

In this PR, I have copied the `iterables` example code into the `null_safety_examples` folder.
I had to copy as well the `util` code to it, because `iterables` depends on it.
Migrating the codelab and updating the `code-excerpts` will come in a separated PR.

I also applied the following changes:

- sdk version set to `">=2.12.0-0 <3.0.0"`
- `dev_dependencies` point to the `nullsafety` versions of pedantic and test packages.
- in `util` I had to fix the `dartMajorVers` and the `ellipsis` code snippets to support null safety.
- the code in `iterables` was already null safe after applying the sdk change.
- I did not copy the `analysis_options.yaml`, and instead I pointed to the one in `examples`.

Let me know if you would like to do anything differently.

cc. @kwalrath

### DIFF
--- a/null_safety_examples/iterables/analysis_options.yaml
+++ b/null_safety_examples/iterables/analysis_options.yaml
@@ -1,1 +1,1 @@
-include: ../../examples/analysis_options.yaml
+include: ../analysis_options.yaml
--- a/null_safety_examples/iterables/analysis_options.yaml
+++ b/null_safety_examples/iterables/analysis_options.yaml
@@ -1,0 +1,1 @@
+include: ../../examples/analysis_options.yaml
--- a/null_safety_examples/iterables/test/iterables_test.dart
+++ b/null_safety_examples/iterables/test/iterables_test.dart
@@ -1,0 +1,107 @@
+import 'package:test/test.dart';
+
+void main() {
+  group('iterables codelab', () {
+    test('any_example', () {
+      void any(Iterable<String> items, Function(String) print) {
+        // #docregion any-false
+        if (items.any((element) => element.contains('Z'))) {
+          print('At least one element contains "Z"');
+        } else {
+          print('No element contains "Z"');
+        }
+        // #enddocregion any-false
+      }
+
+      var items = ['Zoo', 'Home'];
+      any(items, (output) {
+        expect(output, 'At least one element contains "Z"');
+      });
+
+      items = ['Home'];
+      any(items, (output) {
+        expect(output, 'No element contains "Z"');
+      });
+    });
+
+    test('every_example', () {
+      bool bad(Iterable<String> items) {
+        // #docregion every-bad
+        for (var item in items) {
+          if (item.length < 5) {
+            return false;
+          }
+        }
+        return true;
+        // #enddocregion every-bad
+      }
+
+      bool good(Iterable<String> items) {
+        // #docregion every-good
+        return items.every((element) => element.length >= 5);
+        // #enddocregion every-good
+      }
+
+      expect(bad(['12345']), true);
+      expect(bad(['1234']), false);
+      expect(good(['12345']), true);
+      expect(good(['1234']), false);
+    });
+
+    test('firstWhere_example', () {
+      final iterable = ['a', '123456', 'abcdef'];
+      // #docregion firstwhere
+      String element = iterable.firstWhere((element) => element.length > 5);
+      // #enddocregion firstwhere
+      expect(element, '123456');
+    });
+
+    test('iterable_example', () {
+      // #docregion iterable
+      // #docregion iterable-elementat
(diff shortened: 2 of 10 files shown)


kwalrath:
LGTM, thank you! We should probably get a review from someone like @RedBrogdon or @johnpryan or @filiph, as well. I've committed the utils package myself  (a few times, while testing `dart migrate` :), and I wondered we might find a better solution than just adding `?` and `!`.

In future PRs, could you have an initial commit that's just a copy of the original files? That way it'll be easier to see what changed. (I have two local copies of the repo, so I just put your changes in the one I keep for PRs and diffed it with the other repo... but it'd be nice to be able to do that comparison in the GitHub UI.)

domesticmouse:
+1 to an initial PR that is the current code and follow up PR with the changes.

I'm curious if `int.parse(_majorVersNum.firstMatch(Platform.version)![1]!);` is really the best we can do. It seems noisy, but I can't see a better way.

miquelbeltran:
> could you have an initial commit that's just a copy of the original files?

Will do!

miquelbeltran:
Rebased with current master.

null_safety_examples/iterables/test/iterables_test.dart
```diff
@@ -0,0 +1,107 @@
...
+    test('firstWhere_example', () {
+      final iterable = ['a', '123456', 'abcdef'];
+      // #docregion firstwhere
+      String element = iterable.firstWhere((element) => element.length > 5);
+      // #enddocregion firstwhere
+      expect(element, '123456');
+    });
+
+    test('iterable_example', () {
+      // #docregion iterable
```
RedBrogdon, line 60: I didn't realize you could interleave regions like this.

null_safety_examples/iterables/analysis_options.yaml
```diff
@@ -0,0 +1 @@
+include: ../../examples/analysis_options.yaml
```
RedBrogdon, line 1: nit: newline needed.

I'd actually recommend copying the file at `/examples/analysis_options.yaml` to `null_safety_examples/analysis_options.yaml`. It's possible (even likely) that they'll diverge at some point, and I'd also like to avoid inter-dependencies between the two sets of examples.

RedBrogdon (approved):
One nit. LGTM otherwise!

null_safety_examples/iterables/analysis_options.yaml
miquelbeltran: makes sense, thanks! I will do that

RedBrogdon (approved):

kwalrath:
@miquelbeltran I'm thinking of two things, which could/should be in a separate PR (which either you or I could create):

1. Move `null_safety_examples` to `examples_null_safety` (since my eyes keep missing this directory, and it makes sense to keep the two directories alphabetically close).

2. Add a liquid/jekyll macro that uses a site-wide variable to choose between the current and null-safe version of each example. (To simplify things, you wouldn't switch an example to this macro until a null-safe version exists, and you wouldn't use it for examples that exist only in the null-safe example directory.)

What do you think, Miguel?

kwalrath (approved):

miquelbeltran:
Hi @kwalrath 

1. Agreed! makes sense.
2. I see an issue with it, is that codelabs mix both examples (code is in the examples folder) and dartpads (code is found in the markdown). As well, the explanation around them may also change. So it is more complicated than just replacing one example by the other, as it requires to touch up the rest of the markdown file that contains it.

kwalrath:
> 2. I see an issue with it, is that codelabs mix both examples (code is in the examples folder) and dartpads (code is found in the markdown). As well, the explanation around them may also change. So it is more complicated than just replacing one example by the other, as it requires to touch up the rest of the markdown file that contains it.

Oh, good point. Maybe we could just use something like the following to point to the new code. (I'd like to have a easy-to-search-for convention so we don't have to look inside the example directories and can easily find the diffs.) What do you think of this (which you could add to this PR):

```
{% comment %}
Converted to null safety: https://github.com/dart-lang/site-www/pull/2805.
{% endcomment %}
```

miquelbeltran:
Yes, we can add that, I will add that to the upcoming codelabs migration.

Another thing :) Would it be OK to merge this PR? I have the code for the iterables codelab and the async-await as well ready, but it needs this one to be merged first.

kwalrath:
> Another thing :) Would it be OK to merge this PR? I have the code for the iterables codelab and the async-await as well ready, but it needs this one to be merged first.

Sure thing. 😆

PR #2808, opened 2020-12-16
Clarify how dartdoc handles getters and setters with doc comments.

### LINKED ISSUE #1078: Better guideline on adding DocComment on setter
At the bottom of [this Effective Dart guideline](https://www.dartlang.org/guides/language/effective-dart/documentation#prefer-starting-variable-getter-or-setter-comments-with-noun-phrases), there is this sentence:

`Avoid having a doc comment on both the setter and the getter, as DartDoc will show only one (the one on the getter.)`

This is a bit ambiguous, especially because, I don't know how DartDoc gets generated.

Does that sentence mean "DartDoc will only show the one on the getter, when a comment is added on both getter and setter" or does that mean, "DartDoc will always only show the one on the getter"?

For angular components, it is pretty common to split an `@Input()` into a getter/setter pair.
When you do, the `@Input()` annotation goes on the setter, as that is the part of the public API.
And, in those cases, I have seen it to be a common practice to put the DartDoc above that annotation, which means, you are only annotating the setter.

What happens in this case?

And can we, may be, tweak that sentence to make it less ambiguous?

munificent:
Fix #1078.

### DIFF
--- a/src/_guides/language/effective-dart/documentation.md
+++ b/src/_guides/language/effective-dart/documentation.md
@@ -260,7 +260,7 @@ int get checkedCount => ...
 If a property has both a getter and a setter, then create a doc comment for
 only one of them. Dartdoc treats the getter and setter like a single field,
 and if both the getter and the setter have doc comments, then
-dartdoc discards the setter's doc comments.
+dartdoc discards the setter's doc comment.
 
 ### PREFER starting library or type comments with noun phrases.
 
--- a/src/_guides/language/effective-dart/documentation.md
+++ b/src/_guides/language/effective-dart/documentation.md
@@ -257,10 +257,10 @@ int weekday;
 int get checkedCount => ...
 {% endprettify %}
 
-When you have a corresponding getter and setter with the same name, document
-only one of them. Dartdoc will merge them together and document them as if they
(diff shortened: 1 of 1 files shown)


src/_guides/language/effective-dart/documentation.md
```diff
@@ -257,8 +257,10 @@ int weekday;
 int get checkedCount => ...
 {% endprettify %}
 
-Avoid having a doc comment on both the setter and the getter, as DartDoc will show
-only one (the one on the getter.)
+When you have a corresponding getter and setter with the same name, document
+only one of them. Dartdoc will merge them together and document them as if they
+are a single field. When it does that, if both the getter and setter have doc
+comments, it discards the setter's.
```
kwalrath, line 263:
```suggestion
If a property has both a getter and a setter, then create a doc comment for
only one of them. Dartdoc treats the getter and setter like a single field,
and if both the getter and the setter have doc comments, then
dartdoc discards the setter's doc comments.
```

kwalrath: My suggestions are an attempt to remove some possible ambiguity, and we (conforming to Google's style guide) also prefer present tense.

kwalrath (approved):

PR #2811, opened 2020-12-17
Fix some null safety errors in "Effective Dart" example code context.

munificent:
I wanted to get a sense of how many of the examples would be affected
by null safety, so I temporarily enabled that and went through the
errors. The good news is that it's not bad. Probably half a dozen with
trivial fixes.

The bad news is that those fixes do require using null safety feature
syntax in the examples themselves ("?", "required", etc.). So I'm not
going to make those changes yet.

But, while I was at it, I went ahead and fixed a bunch of null safety
errors in the surrounding context code. These changes are entirely
backwards compatible and fine to land now on master so I figured I'd
send them out first.

### DIFF
--- a/examples/misc/lib/effective_dart/design_bad.dart
+++ b/examples/misc/lib/effective_dart/design_bad.dart
@@ -19,7 +19,7 @@ void miscDeclAnalyzedButNotTested() {
     monsters.filter((monster) => monster.hasClaws);
     // #enddocregion code-like-prose
 
-    Iterable theCollectionOfErrors;
+    Iterable theCollectionOfErrors = [];
     // #docregion code-like-prose-overdone
     if (theCollectionOfErrors.isEmpty) {/*-...-*/}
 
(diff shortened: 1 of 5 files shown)


kwalrath:
> The bad news is that those fixes do require using null safety feature
> syntax in the examples themselves ("?", "required", etc.). So I'm not
> going to make those changes yet.

Is there any reason not to put your code-only changes in the /null-safety-examples directory?

kwalrath:
I suspect you can fix the build by running tool/dartformat.sh. Even if you formatted things perfectly for 80 chars, we often have shorter line lengths to avoid horizontal scrolling.

munificent:
> Is there any reason not to put your code-only changes in the /null-safety-examples directory?

Not in particular, I was just trying to make the easy non-disruptive changes first.

kwalrath:
I'm surprised you didn't need to update the .md files, too, but the code changes LGTM.

kwalrath (approved):

munificent:
> I'm surprised you didn't need to update the .md files, too, but the code changes LGTM.

All of these changes were just in the contextual code used to make the snippets actually compile. :)

PR #2815, opened 2020-12-18
Update language-tour.md

timsneath:
Remove references to long-dead language versions. At this point, there's no need to document features added in 2.3 or call out that async keywords aren't supported in Dart 1.x :)

### DIFF
--- a/examples/misc/lib/language_tour/variables.dart
+++ b/examples/misc/lib/language_tour/variables.dart
@@ -57,7 +57,7 @@ void miscDeclAnalyzedButNotTested() {
 
   {
     // #docregion const-dart-25
-    // Valid compile-time constants as of Dart 2.5.
+    // Some examples of valid compile-time constants
     const Object i = 3; // Where i is a const Object with an int value...
     const list = [i as int]; // Use a typecast.
     const map = {if (i is int) i: "int"}; // Use is and collection if.
--- a/examples/misc/lib/language_tour/variables.dart
+++ b/examples/misc/lib/language_tour/variables.dart
@@ -57,7 +57,6 @@ void miscDeclAnalyzedButNotTested() {
 
   {
     // #docregion const-dart-25
-    // Some examples of valid compile-time constants
     const Object i = 3; // Where i is a const Object with an int value...
     const list = [i as int]; // Use a typecast.
     const map = {if (i is int) i: "int"}; // Use is and collection if.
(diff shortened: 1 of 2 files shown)


src/_guides/language/language-tour.md
```diff
@@ -403,14 +402,13 @@ You can't change the value of a `const` variable:
...
 
-As of Dart 2.5, you can define constants that use
+You can define constants that use
 [type checks and casts](#type-test-operators) (`is` and `as`),
 [collection if](#collection-operators),
 and [spread operators](#spread-operator) (`...` and `...?`):
 
 <?code-excerpt "misc/lib/language_tour/variables.dart (const-dart-25)"?>
 ```dart
-// Valid compile-time constants as of Dart 2.5.
```
kwalrath, line 413: To fix the build we'll need to update the source of this sample, too. (examples/misc/lib/...)

kwalrath (commented):
I'm always nervous about removing history, but I'm willing to do it if people like @mit-mit and @kevmoo and @RedBrogdon are OK with it.

timsneath:
Yeah, I think the problem is that all these references clutter the document with extraneous information ("why are they telling me this? what version am I running? should I be running a different version?"). 

If we _want_ a changelog to track when new language features arrived, then we should have one, but I don't believe this is the place. Conversely, if this is primarily targeted at new-to-Dart developers (as I believe it is), I'm keen that we document the world as it is, not the journey that took us there.

mit-mit:
I'm fine with removing the version information; we have that in our evolution page: https://dart.dev/guides/language/evolution

kevmoo (approved):
Godo to move to the future!

src/_guides/language/language-tour.md
timsneath: Great catch. Fixed in subsequent commit.

kwalrath:
I'll look into the build failure

examples/misc/lib/language_tour/variables.dart
```diff
@@ -57,7 +57,7 @@ void miscDeclAnalyzedButNotTested() {
 
   {
     // #docregion const-dart-25
-    // Valid compile-time constants as of Dart 2.5.
+    // Some examples of valid compile-time constants
```
kwalrath, line 60: I don't think we need this comment.

```suggestion
```

kwalrath:
Thanks, @timsneath!

PR #2817, opened 2020-12-21
Update and fix old pub link

### LINKED ISSUE #2377: Update pub.dev tab URLs
Per the conversation at https://github.com/dart-lang/site-www/pull/2332#pullrequestreview-394558946, pub.dev package tabs will have prettier URLs soon. Once those URLs are implemented, we should use them. (Low priority because the old URLs will continue to work.)

parlough:
It seems the new format of links are used by default, even when navigating between tabs. In fact, it appears this old link didn't even navigate to the example tab. This was the only link I identified which navigated to a tab with the old fashion.

Fixes #2377


kwalrath (approved):
Thanks!

PR #2818, opened 2020-12-21
Add font-weight 700 Roboto to improve bold text clarity

parlough:
Since we use font-weight: 700 for some bold text, not including that weight causes some slight blurriness for bold text. It's hard to notice, but it's there.

It's hard to showcase in screenshots due to compression, but here I'll try. Refer to the `Note`: 

Previously:
image

After:
image

### DIFF
--- a/src/_includes/head.html
+++ b/src/_includes/head.html
@@ -41,7 +41,7 @@
   <meta property="og:type" content="{{og_type}}" />
   {% endif -%}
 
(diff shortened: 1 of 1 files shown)


kwalrath (commented):
Thanks, @parlough! @johnpryan LGTY?

johnpryan:
LGTM, thanks! It's very slight, but now I can't unsee it.

kwalrath (approved):

PR #2820, opened 2020-12-21
Update faq.md

### LINKED ISSUE #2799: 'Null safety: frequently asked questions' page issue
Page URL: https://dart.dev/null-safety/faq.html
Page source: https://github.com/dart-lang/site-www/tree/master/src/null-safety/faq.md

Description of issue:
I was just wondering about the following point:

> ## How do I signal that the return value from a Map is non-nullable?
> 
> The
> [lookup operator](https://api.dart.dev/stable/dart-core/Map/operator_get.html)
> on Map (`[]`) by default returns a nullable type. There's no way to signal to
> the language that the value is guaranteed to be there.
> 
> In this case, you should use the bang operator (`!`) to cast the value back to
> V:
> 
> ```dart
> if (blockTypes.containsKey(key)) {
>   return blockTypes[key]!; // blockTypes[key] is non-nullable
> } else {
>   throw ArgumentError('Could not read block type.');
> }
> ```

From my point of understanding this should be changed to something like this:

> ## How do I signal that the return value from a Map is non-nullable?
> Do not try to do that. Simply test whether the return value is null or not.
> 
> ```dart
> final blockType = blockTypes[key]
> if (blocktype != null) {
>   return blocktype;
> } else {
>   throw ArgumentError('Could not read block type.');
> }
> ```

The argument being, that in multithreaded code the code snippet as present in the FAQ now could actually fail with a null pointer exception, due to the value at key being possibly removed between the `containsKey()` check and the actual use of the lookup operator `[]`.

SebastianJL:
Changes according to discussion in #2799.
Closes #2799.

### DIFF
--- a/src/null-safety/faq.md
+++ b/src/null-safety/faq.md
@@ -221,11 +221,15 @@ In this case, you should use the bang operator (`!`) to cast the value back to
 V:
 
 ```dart
-if (blockTypes.containsKey(key)) {
-  return blockTypes[key]!; // blockTypes[key] is non-nullable
-} else {
-  throw ArgumentError('Could not read block type.');
-}
+return blockTypes[key]!;
+```
+
+Which will throw if the map returns null. If you want explicit handling for that case:
+
+```dart
+var result = blockTypes[key];
+if (result != null) return result;
+// Handle the null case here, e.g. throw with explanation.
 ```
 
 ## Why is the generic type on my List/Map nullable?


SebastianJL:
@googlebot I signed it!

davidmorgan:
Looks good, thanks! Added Ryan for review as I don't have permission to merge here.

ryanmcginnis:
thanks! over to Kathy (i don't have permission either). i've started the changes for the internal document.

kwalrath (approved):

kwalrath:
LGTM! (And I'll get @ryanmcginnis commit permission.)

PR #2823, opened 2020-12-26
doc update to use new `dart` tool in `pub upgrade`

KevinRohn:
Updated documentation for `src/tools/pub/cmd/pub-upgrade.md` to use the new dart tool.

I also updated the `src/tools/pub/cmd/_pub-outdated-output.html` file and changed the terminal preview (html file) from `pub outdated` to `dart pub outdated`.

See #2617


KevinRohn:
I closed this pr, because it's duplicated by #2825.

PR #2824, opened 2020-12-26
doc update to use new `dart` tool in `pub cache`

KevinRohn:
Tested the new `dart` tool commands and updated the doc with the new syntax.

To keep things simple I updated the example code `pub cache add barback --version "<=0.8.0 <0.110"` to `dart pub cache add http --version "0.12.2"`. The `barback` package is already archived and if someone reads it, I think a package name like `http` is more clear to understand.

See: #2617


KevinRohn:
I closed this pr, because it's duplicated by #2825.

PR #2826, opened 2020-12-26
doc update to use new `dart` tool in `pub get`

KevinRohn:
tested the new `dart` tool commands for `src/tools/pub/cmd/pub-get.md` and updated the syntax.

See: https://github.com/dart-lang/site-www/issues/2617

### DIFF
--- a/src/tools/pub/cmd/pub-get.md
+++ b/src/tools/pub/cmd/pub-get.md
@@ -1,12 +1,12 @@
 ---
-title: pub get
-description: Use pub get to retrieve the dependencies used by your Dart application.
+title: dart pub get
+description: Use dart pub get to retrieve the dependencies used by your Dart application.
 ---
 
 _Get_ is one of the commands of the [pub tool](/tools/pub/cmd).
 
 {% prettify nocode tag=pre+code %}
-$ pub get [--offline]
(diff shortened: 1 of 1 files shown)


kwalrath (commented):
LGTM! @johnpryan how does this look to you? I believe this is the first of the `pub` command pages to be updated to `dart pub`, so I want to make sure we get the conventions right.

Perhaps we should add a note at the top about this being equivalent to `pub get`, but that we expect `pub` as a separate command to go away, in favor of `dart pub`?

src/tools/pub/cmd/pub-get.md
```diff
@@ -21,7 +21,7 @@ Got dependencies!
 ```
```
kwalrath, line 21: Oops, 2 lines above this `pub get` should be `dart pub get`

kwalrath:
I've staged this at https://kw-www-dartlang-1.firebaseapp.com/tools/pub/cmd/pub-get.

I think we can take this once the terminal example is fixed. I don't think we need to say something about this replacing `pub get`, because I can just add something to https://kw-www-dartlang-1.firebaseapp.com/tools/pub/cmd about this command being available via the `dart` command.

kwalrath (approved):

PR #2832, opened 2021-01-05
Clarify that migrating imposes an SDK version requirement

mit-mit:


src/null-safety/migration-guide.md
```diff
@@ -53,6 +53,10 @@ then passing a nullable argument becomes a compile error.
   don't use null safety yet.
   For example, the Dart and Flutter core libraries are null safe,
   and they're still usable by apps that haven't migrated to null safety.
+  
+  Note, however, that any package or app that uses a null safety-enabled
+  package needs to use an SDK version that understands null safety. Currently
+  this is Dart 2.12 beta or later.
```
kwalrath, line 59:
```suggestion
  Note: Any package or app that uses a null safety-enabled
  package **must use an SDK version that can support null safety.**
  Currently this version is Dart 2.12 beta or later.
```

Alternatively, we might change the text above. Something like this:

Your null-safe package is usable by packages and apps that don’t use null safety yet,
**if they use an SDK version that supports null safety**
(currently, Dart 2.12 beta or later).
For example, the latest beta versions of the Dart and Flutter core libraries are null safe,
and they’re still usable by apps that haven’t migrated to null safety.

mit-mit: @kwalrath

kwalrath: @mit-mit let's talk (I've added it to our agenda :)

mit-mit:
@kwalrath can you approve with that last edit?

kwalrath (approved):

Issue #2859, opened 2021-01-14
Address TODOs related to style 


woinbo:
TODO's related to style:

Style Related
- [ ] TODO(chalin): use $site-color-codeblock-bg? [link to file](https://github.com/dart-lang/site-www/blob/main/src/_sass/components/_code.scss#L1)
- [ ] TODO(chalin): 1.5rem; [link to file](https://github.com/dart-lang/site-www/blob/main/src/_sass/components/_code.scss#L7)
- [ ] TODO: drop once heading class styles (like .h4) are fixed. [link to file](https://github.com/dart-lang/site-www/blob/main/src/_sass/_books.scss#L13)
- [ ] TODO: use / override BS defaults named $hX-font-size. [link to file](https://github.com/dart-lang/site-www/blob/main/src/_sass/core/_variables.scss#L77-L84)

@kwalrath (#2119 )

PR #2865, opened 2021-01-14
Update TODO with relevant issue filed in #2859

woinbo:
Update TODO in file src/_assets/css/_variables.scss with relevant issue filed in #2859

### DIFF
--- a/src/_assets/css/_variables.scss
+++ b/src/_assets/css/_variables.scss
@@ -74,7 +74,7 @@ $google-font-family:        "Product Sans", $font-family-base;
 // Sidenav, toc, and footer
 $font-size-small:         14px;
 
-// TODO: use / override BS defaults named $hX-font-size. #2859
+// TODO #2859: use / override BS defaults named $hX-font-size.
 // Headings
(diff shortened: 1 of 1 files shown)


src/_assets/css/_variables.scss
```diff
@@ -74,7 +74,7 @@ $google-font-family:        "Product Sans", $font-family-base;
 // Sidenav, toc, and footer
 $font-size-small:         14px;
 
-// TODO: use / override BS defaults named $hX-font-size.
+// TODO: use / override BS defaults named $hX-font-size. #2859
```
kwalrath, line 77: I'm thinking it might be good to have the issue number right next to the TODO. Something like this:

```suggestion
// TODO #2859: use / override BS defaults named $hX-font-size.
```

(It was hard to find a precedent for this; my initial search turned up auto-creation of issues from TODOs, but although that might be cool, I'm more concerned about the other direction. https://github.com/phetsims/chipper/issues/815 had some discussion of this.)

kwalrath (approved):
Thanks, @woinbo!

PR #2873, opened 2021-01-18
[NOT READY FOR MERGING] Mark `dart:ffi` as stable

dcharkes:
Should be merged when https://dart-review.googlesource.com/c/sdk/+/179765 is merged.

### DIFF
--- a/src/_guides/libraries/c-interop.md
+++ b/src/_guides/libraries/c-interop.md
@@ -12,13 +12,8 @@ _FFI_ stands for [_foreign function interface._][FFI]
 Other terms for similar functionality include _native interface_
 and _language bindings._
 
-{{site.alert.info}}
-  As of Dart 2.10, [dart:ffi is in beta,][ffi issue]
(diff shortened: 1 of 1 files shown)


mit-mit:
I like the change, but for now I think we should keep a note that says "As of Dart 2.12 beta, FFi has been marked as 1.0, and will be fully stable once a 2.12 stable SDK is released". WDYT @kwalrath ?

kwalrath:
> I like the change, but for now I think we should keep a note that says "As of Dart 2.12 beta, FFi has been marked as 1.0, and will be fully stable once a 2.12 stable SDK is released". WDYT @kwalrath ?

SGTM. Also, the note should probably switch from site.alert.info to site.alert.version-note.

kwalrath (approved):

kwalrath:
I believe the build break is due to the age of this PR. (It seems to be fine.) So I'm going to merge 🤞

PR #2876, opened 2021-01-19
Document main channel builds

mit-mit:


kwalrath (approved):
LGTM if my changes LGTY

athomas (approved):
LGTM

At some point, it would be nice if we could hide the implementation details on GCS from these URLs via some sort of redirects (something like https://dart.dev/get-dart/dev/linux/arm/latest -> redirects to GCS URL). The current bucket is in a GCP project we don't control (and we've had outages because of similar situations) and there are old sins in the `be` subfolder that if cleaned-up could be seen as a breaking change (so please don't go into more details on what is in the `be` folder than we do with this PR).

PR #2879, opened 2021-01-20
Redirect for SDK archives

mit-mit:


mit-mit:
@athomas what we discussed. But not clear if we willing to take the dependency on Firebase Hosting for this or not?

athomas:
Yes, might be better to keep direct links to GCS. The benefits are mostly "nice to have" and firebase hosting is a real risk. And changing to a different URL later is no more breaking that getting the community to switch to this one.

PR #2857, opened 2021-01-13
Migrate Async-Await Codelab to Null Safety

miquelbeltran:
Continuing the work from the #2805 PR.

This PR adds the migrated version of the Async-Await Collections codelab in the `async-await_migrated.md`, following the migration guide.

This PR is split in different commits:

1. the original copy of the codelab and the examples folder.
2. contains the modification of the example code (no code was changed in the dart files, but the pubspec changed)
3. contains the modification of the codelab markdown and code.

I removed the references to the iframes hosted as gists, because I cannot update them and I assume they are not deprecated.

Live preview: https://mb-dart-dev-2.firebaseapp.com/codelabs/async-await_migrated

cc. @kwalrath @RedBrogdon

PR #2880, opened 2021-01-21
Migrate Dart Cheat-Sheet Codelab to Null Safety

miquelbeltran:
Continuing the work from the #2857 PR.

This PR adds the migrated version of the Dart cheat-sheet codelab, following the migration guide.

This PR is split in different commits, a larger commit (the last one) shows the changes in the codelab markdown.

This codelab has been modified a little bit to accommodate null safety better:

- Added a short introduction about null safe variable declaration, linking to the null safety docs.
- Some sections have extra explanations.
- Some dartpats show errors by default, this is intended because the learner must complete the dartpad for the errors to go away.

It is larger than the other two PRs, please feel free to edit the content!

Link preview: https://mb-dart-dev-1.firebaseapp.com/codelabs/dart-cheatsheet_migrated

cc. @kwalrath @RedBrogdon

### DIFF
--- a/null_safety_examples/misc/bin/named_constructor.dart
+++ b/null_safety_examples/misc/bin/named_constructor.dart
@@ -1,0 +1,20 @@
+// ignore_for_file: sort_constructors_first
+
+// #docregion
+class Point {
+  double x, y;
+
+  Point(this.x, this.y);
+
+  Point.origin() {
+    x = 0;
+    y = 0;
+  }
+}
+// #enddocregion
+
+void main() {
+  final myPoint = Point.origin();
+  print(myPoint.x);
+  print(myPoint.y);
+}
--- a/null_safety_examples/misc/bin/named_constructor.dart
+++ b/null_safety_examples/misc/bin/named_constructor.dart
@@ -2,7 +2,7 @@
 
 // #docregion
 class Point {
-  double x, y;
+  late double x, y;
 
   Point(this.x, this.y);
 
--- a/src/codelabs/dart-cheatsheet_migrated.md
+++ b/src/codelabs/dart-cheatsheet_migrated.md
@@ -99,36 +99,47 @@ reference them inside single quotes, with a space in between.
 
 ## Nullable variables
 
-In Dart, types in your code are non-nullable by default, meaning that values can’t be null
-unless you say they can be.
+Dart 2.12 introduced sound null safety,
+meaning that (when you [enable null safety][])
+values can’t be null unless you say they can be.
+In other words, types are non-nullable by default.
 
-For example, consider the following code, which is **invalid**, because `null` cannot be assigned to `a`.
-
-{% prettify dart tag=pre+code %}
-int a = [!null!];
-{% endprettify %}
-
-When creating a variable, you can use ? to inform Dart of the variable’s nullability.
+For example, consider the following code,
+which is **invalid** because (with null safety)
+a variable of type `int` can't have the value `null`:
 
 {% prettify dart tag=pre+code %}
-int? a = null;
+int a = [!null!]; // INVALID in null-safe Dart.
 {% endprettify %}
 
-However, nullable variables are null by default, so it is not necessary to assign null to it.
+When creating a variable in Dart 2.12 or higher,
+you can add `?` to the type to indicate
+that the variable can be null:
 
 {% prettify dart tag=pre+code %}
-int? a; // The initial value of a is null
+[!int?!] a = null; // Valid in null-safe Dart.
 {% endprettify %}
 
-Read the [Sound null safety guide](/null-safety) for more information about null safety on Dart.
+You can simplify that code a bit because, in all versions of Dart,
+`null` is the default value for uninitialized variables:
+
+{% prettify dart tag=pre+code %}
+int? a; // The initial value of a is null.
+{% endprettify %}
+
+For more information about null safety in Dart,
+read the [sound null safety guide](/null-safety).
+
+[enable null safety]: https://dart.dev/null-safety#enable-null-safety
+
 
 ### Code example
 
 Try to declare two variables below:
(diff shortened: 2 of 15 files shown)


miquelbeltran:
looking at the CI issues

.firebaserc
```diff
@@ -1,6 +1,6 @@
 {
   "projects": {
-    "default": "dart-dev",
+    "default": "mb-dart-dev-1",
```
kwalrath, line 3: This should be local only, right?

miquelbeltran: oops :) amended the mistake

src/codelabs/dart-cheatsheet_migrated.md
```diff
@@ -439,35 +508,35 @@ bool hasEmpty = aListOfStrings.any((s) => s.isEmpty);
 
 Try finishing the following statements, which use arrow syntax.
 
-```dart:run-dartpad:height-345px:ga_id-arrow_syntax
+```dart:run-dartpad:height-345px:ga_id-arrow_syntax:null_safety-true
 {$ begin main.dart $}
 class MyClass {
-  int _value1 = 2;
```
kwalrath: It seems to me that the variables are intentionally private, although underscore and its relationship to privacy are never discussed. (Whoops. I'll try to remedy that.)

Do you think there's a good reason to make these variables public?

kwalrath: Actually, on second thought, I think there's no good reason to use `_` here. But I should still add something about privacy and underscore, probably in "Getters and setters".

null_safety_examples/misc/bin/named_constructor.dart
```diff
@@ -0,0 +1,20 @@
+// ignore_for_file: sort_constructors_first
+
+// #docregion
+class Point {
+  late double x, y;
+
+  Point(this.x, this.y);
+
+  Point.origin() {
```
parlough, line 9: Based off the comments in the codelab this seems to also be used to intentionally show of the functionality of `late`.

If the desire to illustrate the functionality of `late` here stands, perhaps a different example could be used as this case nor the solution need to use `late`, with the constructor becoming `Point.origin(): this(0, 0);`.

I say this because how I see it, `late` shouldn't be used unless necessary and in this case it's essentially just a hint to the analyzer.

src/codelabs/dart-cheatsheet_migrated.md
parlough: I'll just leave a comment here in favor of the private fields. This class seems to intentionally encapsulate some of the behavior with a method to increment a specific value, calculate a product, join them, etc with the intention to base it off of those fields. 

While I know this is purely a theoretical class created as an example, if a user of this object could manipulate those freely, it seems like it would be unexpected/unintended by the writer. Why else would they have the form of encapsulation.

kwalrath (approved):
LGTM if my changes are OK with you.

Other changes I considered (but decided not to do at this time, as these aren't crucial AND they predate your changes):
- Get rid of "Code example" headers (at least from the auto-generated TOC)
- Fix line lengths to be <= 80 chars (and semantic)
- Where appropriate, change periods to colons before samples
- Make sure we use backtick everywhere we should

After this is published I'd like to ask for testers to help us find places where we could provide better hints and test feedback. But... not yet.

null_safety_examples/misc/bin/named_constructor.dart
kwalrath: I suspect the `late` was just a side effect of migrating the example. But you're right, maybe we should feature it more.

kwalrath:
Of course the build needs to work... and we should update the code examples to match whatever we end up using in the page.

src/codelabs/dart-cheatsheet_migrated.md
miquelbeltran: The reason I changed the visibility of the fields is because it would show a warning  in the dartpad otherwise:
image

As most other examples in the codelab also used public fields, I thought it would be a good idea to make them public as well.

I could switch back to private fields and then have a couple of statements in the test section that could use them, so there's no warnings in the dartpad. Or we could simple tell the learner to ignore the warnings.

Let me know what you prefer! :)

null_safety_examples/misc/bin/named_constructor.dart
miquelbeltran: Yes, the use of `late` is a side effect of explaining Named Constructors.

The actual good solution is what @parlough  said, but that solution is explained a bit later in the Redirecting Constructors section.

In the codelab there's a line explaining: "This example uses the late annotation to indicate that x and y will be initalized later in the named constructor."

I could probably add a line explaining that we no longer need to annotate with `late` any more because because we are using a redirecting constructor in the other section.

src/codelabs/dart-cheatsheet_migrated.md
```diff
@@ -949,7 +949,7 @@ Their default value is null unless you provide another default value:
 <?code-excerpt "../null_safety_examples/misc/bin/optional_positional_args2.dart"?>
 ```dart
 int sumUpToFive(int a, [int b = 2, int c = 3, int d = 4, int e = 5]) {
-  // ···
+// ···
```
miquelbeltran, line 952: unfortunately this is generated by the code-excerpt integration so we can't add two spaces here

kwalrath: I'm pretty sure there's a way around that... but I don't remember what it is, so lgtm.

null_safety_examples/misc/bin/named_constructor.dart
kwalrath: On second thought, I'm not sure we want to feature `late` too much, so let's leave it as-is, for now. (We don't want people to use it unnecessarily.)

PR #2883, opened 2021-01-22
Update src/tools/pub/cmd/pub-uploader.md to use new dart tools

woinbo:
Update src/tools/pub/cmd/pub-uploader.md to new dart tools


kwalrath (approved):
LGTM. Thanks, @woinbo!

PR #2889, opened 2021-01-26
Add SVG versions of homepage images

kwalrath:


kwalrath:
@mit-mit I think this stalled because we weren't sure whether using SVGs would work everywhere. We do use them in the overview (Dart-platforms.svg) and 10 other places, but the homepage is more likely to attract people using random browsers. 

Should I close this PR?

Either way, we should update the screenshot of DevTools (3rd bullet under **Productive development**).

domesticmouse:
Fwiw, according to caniuse.com SVG in HTML IMG element support is almost 99% coverage: https://caniuse.com/svg-img

I don't see any reason not to move this forward.

mit-mit:
SGTM. But let's remove the PNGs then?

kwalrath:
> SGTM. But let's remove the PNGs then?

And use the new ones. :) Yep! I'll make that happen.

kwalrath:
@mit-mit can we get an SVG version of the new image in #3413?

kwalrath:
I've updated the PR and staged it, and I've noticed a couple of issues.

On both sites (i.e. this is a **bug right now on https://dart.dev**), when you're on a computer but in single-column width, section 3 ("Fast on all platforms") shows only one image. It doesn't switch between images, like it does at full size. In fact, this image is sometimes the first (3-1,`$ dart compile exe hello.dart`...) and sometimes one that's displayed in section 1 (1-3, familiar syntax, examples of Dart/Kotlin/Swift/TypeScript).

Only on the staged site (https://kw-staging-dartlang-2.web.app), text doesn't look as pretty — it's pixellated. (The dart.dev copy is pretty but has less contrast and seems a bit blurry.) I think the text in the SVG is drawn as shapes, not text, and I wonder if saving it as real text is possible and would help. It'd be nice to be able to edit the text in a regular editor!

mit-mit:
> @mit-mit can we get an SVG version of the new image in #3413?

yup, will add that shortly

mit-mit:
> when you're on a computer but in single-column width, section 3 ("Fast on all platforms") shows only one image

Fixed in https://github.com/dart-lang/site-www/pull/3419

mit-mit:
> text doesn't look as pretty

I can't see much difference. Which browser are you using? On what kind of machine + display?

mit-mit:
[3-1 - aot compile.svg.zip](https://github.com/dart-lang/site-www/files/6868600/3-1.-.aot.compile.svg.zip)

svg for compile (remove `.zip` manually)

kwalrath:
> > text doesn't look as pretty
> 
> I can't see much difference. Which browser are you using? On what kind of machine + display?

I'm on a Mac, Chrome Canary & Stable, HP ZR24w Display (25.5-inch, 1920 × 1200). Here's a screenshot that shows what I mean (click it to see it at full resolution; there isn't a noticeable difference otherwise):

image

Left: https://kw-staging-dartlang-2.web.app 
Right: https://dart.dev

It's still legible, and I'm OK with it, but I suspect that someone who cares about graphics might be bothered by the pixelation.

mit-mit:
That is really odd. Here on my setup (macbook with 27" monitor scaled to 2560x1440, chrome 91 stable), the look pretty similar:
Screenshot 2021-07-28 at 12 33 48

kwalrath:
@parlough can you see a difference? I wonder if the problem is with my screen resolution.

parlough:
@kwalrath It's a bit tough to tell, but at least in your image I can slightly see what you mean. 

On my Linux machine though, Chrome Dev, 1440p and 2560p, I didn't see much of a difference. Then on my MacBook Pro (2880x1800) it seems similar, unless zoomed in then the SVGs of course do much better. 25in is quite large for 1080p so perhaps it is your screen resolution, I'm not sure. 

I think overall this is an improvement, but like you mentioned, if it would be possible, using SVG text elements would likely result in better scaling.

kwalrath:
I think we might be good to go! Staged here:

https://kw-staging-dartlang-2.web.app/

parlough (approved):

PR #2890, opened 2021-01-27
Mention MacPorts in Mac installation instructions

amake:
The Dart SDK is also available in MacPorts.

### DIFF
--- a/src/_includes/get-sdk.md
+++ b/src/_includes/get-sdk.md
@@ -39,12 +39,16 @@ which (as of Flutter 1.21) includes the full Dart SDK.
 </div>
 
 <div id="tab-sdk-install-mac" class="tabs__content" markdown="1">
-  With [Homebrew,](https://brew.sh/)
+  With [Homebrew](https://brew.sh/) or [MacPorts](https://www.macports.org/),
   installing Dart is easy.
 
   ```terminal
   $ brew tap dart-lang/dart
   $ brew install dart
+  ```
+  or
+  ```terminal
+  $ port install dart-sdk
   ```
 </div>
 


amake:
@googlebot I signed it!

kwalrath:
I hadn't heard of MacPorts! Looks like its dart-sdk port has been around since August 2018.

@mit-mit what do you think?

mit-mit:
It looks like this only has our stable channel?

amake:
> It looks like this only has our stable channel?

Currently, yes. It would be trivial to add the dev channel; it would be burdensome to try to support specific versions.

kwalrath:
We also have a beta chanel, which is probably more useful than the dev channel (although we currently just refer people to the archive page for anything that isn't stable).

kwalrath:
Sorry for letting this PR languish...

I don't think supporting only stable channel is a problem; we no longer document how to download non-stable versions, except by using the archive. Note that this particular include isn't used in https://dart.dev/get-dart; it's used in https://dart.dev/tutorials/server/get-started & https://dart.dev/tutorials/web/get-started, which are as simple (in appearance) as possible.

However, https://www.scivision.dev/homebrew-macports-fink/ worries me... it sounds like MacPorts requires root? Also, that page has figures indicating that Homebrew appears to be _much_ more popular than MacPorts:
 
> GCC: homebrew: 73,800; macports: 127
> CMake: homebrew: 139,730; macports: 502
> Clang: homebrew: 12,613; macports: 70
> Python: homebrew: 869,020; macports: 87

So I'm going to close this PR, although I'm open to reconsidering it. For example, we could maybe mention MacPorts in the /get-dart page instead of or in addition to putting it in this include.

PR #2891, opened 2021-01-27
Migrated language_tour code to null safety

miquelbeltran:
As requested by @kwalrath.

In most cases the changes are transparent to the Language Tour guide and happen outside the code excerpts that get imported.

The most complex cases to solve is what to do with simple classes with one or more members. I went case by case and made them nullable, gave them a default value or marked them as `late` depending on what made more sense.

I also left some comments pointing to changes that should be done in the markdown file as well.

### DIFF
--- a/null_safety_examples/misc/lib/language_tour/async.dart
+++ b/null_safety_examples/misc/lib/language_tour/async.dart
@@ -1,0 +1,87 @@
+// ignore_for_file: unused_element, unused_local_variable, unawaited_futures
+typedef Async0 = Future Function();
+typedef Async1 = Future Function(dynamic);
+typedef Async2 = Future Function(dynamic, dynamic);
+
+Future miscDeclAnalyzedButNotTested() async {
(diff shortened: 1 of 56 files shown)


kwalrath (approved):
Thank you so much, @miquelbeltran!

kwalrath:
Travis has been borked for a while, so I just ran `dart analyze` using the beta SDK. Here's the output (omitting 3 issues in overview_pi.dart, because it isn't yours):

```
.../null_safety_examples/misc (pr/2891)$ dart analyze
Analyzing misc...                      5.2s

   info • Unused import: 'package:meta/meta.dart' at lib/language_tour/functions.dart:3:8 • (unused_import)
          Try removing the import directive.
...
   info • The operand can't be null, so the condition is always true at test/language_tour/control_flow_test.dart:46:17 • (unnecessary_null_comparison)
          Remove the condition.

5 issues found.
```

PR #2892, opened 2021-01-27
Initial cut at system requirements

mit-mit:
cc @athomas @kwalrath one option for how to document system requirements.

Or, would we rather have this on a separate page?

### DIFF
--- a/src/get-dart.md
+++ b/src/get-dart.md
@@ -24,7 +24,6 @@ any of the following are true:
   For example, you might have a continuous integration (CI)
   setup that requires Dart but not Flutter.
 
-
 ## Installing the Dart SDK {#install}
 
 As the following instructions show,
@@ -46,7 +45,7 @@ so it's easy to find (but not more tempting than package managers).
 <ul class="tabs__top-bar">
   <li class="tab-link current" data-tab="tab-sdk-install-windows">Windows</li>
   <li class="tab-link" data-tab="tab-sdk-install-linux">Linux</li>
-  <li class="tab-link" data-tab="tab-sdk-install-mac">Mac</li>
+  <li class="tab-link" data-tab="tab-sdk-install-mac">macOS</li>
 </ul>
 <div id="tab-sdk-install-windows" class="tabs__content current" markdown="1">
 {% include_relative tools/sdk/_windows.md %}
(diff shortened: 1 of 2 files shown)


kwalrath:
When I thought about this a while ago, in https://github.com/dart-lang/sdk/issues/41644#issuecomment-754886793 I proposed putting this in /get-dart#system-requirements.

kwalrath:
I also got some info in email, which I'll record here (but we need more):

* OS:
  * Windows:
    * We probably work on Vista or later. We rely on some locking/condition-variable APIs that were added in Vista, but there might be other dependencies requiring something later.
  * Linux:
    * We probably work on Debian Wheezy or later.
* CPU:
  * X86-64: supported
  * ARM64: supported
  * IA32:
    * We require at least SSE2.
  * 32-bit ARM:
    * We require at least ARMv7.
  * (Past versions supported little-endian MIPS, ARMv6 and ARMv5TE.)
* Misc:
  * Windows: Some features require administrator privileges to work (in particular, symlink creation).
  * In general, we don't actually know what our minimal requirements are for OS or libc versions. "One would have to get a hold of some older systems and test."

athomas:
What we're currently testing on is this:
* Windows 10 version 1703 (Windows-10-15063)
* Ubuntu-16.04.6 (we are using a Debian Jessie sysroot, so that should work, too).
* Mac-10.13.6-17G65
* Android L & N

Flutter tests Dart on a host of additional platforms and there's some more internal platforms.

From an development perspective, we should upgrade these to be more modern:
* Windows 10 version 1703
   * This version is arbitrary, but we're sharing this image with Chromium. Chromium does have other versions so we could 
      choose what makes sense from a product perspective and use something close to that.
* Ubuntu-16.04.6
    * EOL, so we should move to 18.04 soon.
* Mac-10.13.6-17G65
    * We need to upgrade soon. This version doesn't support the clang version we need for the latest xcode.
    * We will also have Apple Silicon devices that will run the latest macOS.
    * Preferred would be to move all infra to 11.2.x (Big Sur).
* Android L & N

Ideally, we'd use what we test on as the minimum requirements. Earlier platforms may work, or not. If there are issues found on earlier platforms, they will be difficult/time consuming for our developers to work on (especially for macOS).

kwalrath (approved):

kwalrath:
Staged: https://kw-www-dartlang-1.firebaseapp.com/get-dart#system-requirements

parlough (approved):

src/get-dart.md
```diff
@@ -58,6 +57,26 @@ so it's easy to find (but not more tempting than package managers).
...
+## System requirements
+
+The Dart SDK is supported on the following operating systems:
+
+#### Windows
+* _Supported versions_: Windows 10.
+* _Supported architectures_: x64, ia32.
+
+#### Linux
+* _Supported versions_: Most recent Linux versions; we however only test with Ubuntu 16.04.
```
athomas: Perhaps recent Linux versions?

src/get-dart.md
```diff
@@ -58,6 +57,26 @@ so it's easy to find (but not more tempting than package managers).
...
+* _Supported architectures_: x64, ia32, arm, arm64.
+* _Notes_: On arm, we require Glibc 2.23 or newer due to [a bug](https://sourceware.org/bugzilla/show_bug.cgi?id=14341) in the dynamic linker.
+
+#### macOS
+* _Supported versions_: Latest three major versions, currently:
```
athomas: Would that mean that if Apple releases a new version we just update the list below and drop support for the oldest version?

src/get-dart.md
```diff
@@ -46,7 +45,7 @@ so it's easy to find (but not more tempting than package managers).
 <ul class="tabs__top-bar">
   <li class="tab-link current" data-tab="tab-sdk-install-windows">Windows</li>
   <li class="tab-link" data-tab="tab-sdk-install-linux">Linux</li>
-  <li class="tab-link" data-tab="tab-sdk-install-mac">Mac</li>
+  <li class="tab-link" data-tab="tab-sdk-install-mac">macOS</li>
```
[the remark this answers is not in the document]
mit-mit: Yes, I think we should. "macOS" is the operating system (which is what we're referring to), "Mac" is a type of Apple hardware.

src/get-dart.md
mit-mit: Done

src/get-dart.md
mit-mit: Yes, that is my suggestion. We support 3 last versions in a rolling window. Typically Apple releases one new version a year, usually in the fall.

mit-mit:
@athomas anything else or can you approve?

athomas (approved):

PR #2895, opened 2021-01-29
Clean up ImmutablePoint.origin

parlough:
Since the field is static, we can specify a single `const` instead of `final` in the declaration.


kwalrath (approved):
Thanks, @parlough!

PR #2899, opened 2021-01-30
Update /tools/pub/cmd

### LINKED ISSUE #2849: Update /tools/pub/cmd
The page needs to talk about the `dart` command, as well as the `flutter` command.

kwalrath:
Fixes #2849.

Staged: https://kw-staging-dartlang-2.firebaseapp.com/tools/pub/cmd


src/tools/pub/cmd/index.md
```diff
@@ -1,34 +1,57 @@
 ---
 title: The pub tool
-description: Pub, a package management tool for Dart, supports a variety of commands.
+description: Pub, a package management tool for Dart, supports a variety of subcommands.
 ---
 
-The `pub` tool has commands for managing packages
+The pub tool has subcommands for managing packages
 and for deploying packages and command-line apps.
```
parlough, line 7: I know you aren't writing this sentence here, but I couldn't help but feel it is a bit strange. Perhaps the double use of "packages", I'm not sure. 

Maybe just `The pub tool has subcommands for managing packages and command-line applications` would be fine, as I feel the term "managing" can also include deployment. 

You're the skillful copy editor though, so perhaps it's fine as is or you have a more clear suggestion :)

kwalrath: Changing just one word didn't work for me. :) ptal

kwalrath: Ugh, just noticed two "use"s in a row. Must... change...

kwalrath:
And... now I suspect that we need to update/delete the [global options section](https://kw-staging-dartlang-2.firebaseapp.com/tools/pub/cmd#global-options). @mit-mit? @sigurdm?

kwalrath:
@mit-mit can we publish, or should we update/delete global options first?

parlough:
I think the global options still make sense, besides `--version` I'm not sure works at all with pub being a `dart` subcommand.

parlough (approved):
Looks good to me and the description clean up is a nice bonus!

src/tools/pub/cmd/index.md
```diff
@@ -1,132 +1,152 @@
 ---
 title: The pub tool
```
mit-mit, line 2: I suggest s/The pub tool/The pub client/ now that it's no longer a standalone tool.

src/tools/pub/cmd/index.md
```diff
@@ -1,132 +1,152 @@
...
-For general information about using the pub package manager, see
-[How to use packages](/guides/packages).
+The [pub package manager](/guides/packages) has a command-line interface
+that works with either the
+[`flutter` tool][flutter-cli] or the [`dart` tool][dart-cli].
+With either tool, add the `pub` command followed by
+a subcommand such as `get`:
 
-{% include flutter-packages.md %}
+```terminal
```
mit-mit, line 12: Swap these two as we're on the Dart site.

src/tools/pub/cmd/index.md
```diff
@@ -1,132 +1,152 @@
...
-## Running command-line apps
-
-Two commands let you run Dart scripts from the command line:
+### Running command-line apps
 
-* The [`pub run`](/tools/pub/cmd/pub-run) command invokes a Dart script in your
-  package, or in one of its dependencies.
+The [`global`](/tools/pub/cmd/pub-global) subcommand lets you 
+make a package globally available, 
+so you can run scripts from that package’s `bin` directory.
```
mit-mit, line 117: Maybe add `Note this requires adding the pub global directory [to path](https://dart.dev/tools/pub/cmd/pub-global#running-a-script-from-your-path).` ?

mit-mit (commented):
Just a few smaller comments. Otherwise LGTM.

src/tools/pub/cmd/index.md
kwalrath: I had two reasons for ordering it this way:
* I wanted to make sure people saw that they might want to run it using the `flutter` command. Everywhere else on this site, we show only the `dart` command.
* It made it easy to just say "any other package" for the `dart` comment.

kwalrath:
I'd like to defer/discuss 2 suggestions, but I think this page might be ready. ptal @mit-mit. (I've updated https://kw-staging-dartlang-2.firebaseapp.com/tools/pub/cmd#running-command-line-apps.)

src/tools/pub/cmd/index.md
mit-mit: OK

src/tools/pub/cmd/index.md
mit-mit: I really, really don't like listing Flutter first on the Dart.dev site. It just seems wrong.

I get the "any other part", but I also don't think that level of detail is needed in the comments (they could just say "Get dependencies for a Dart package"/"Get dependencies for a Flutter package")

kwalrath: That seems wrong to me because Flutter packages _are_ Dart packages. I'd say, instead "Get dependencies for a non-Flutter package".

I guess what it comes down to is: Is there harm in using `dart` when you should use `flutter`? Or vice versa? Which mistake is most likely to occur and cause problems?

If using `dart` when you need `flutter` isn't likely to cause confusion, then I'm OK with putting `dart` first.

mit-mit: There are some small differences, but those are being removed.

For what it's worth, I've heard lots of folks call "Dart packages that contain Flutter specific code" for simply "Flutter packages" and never experienced much confusion over that.

mit-mit (approved):

src/tools/pub/cmd/index.md
kwalrath: I agree that "Flutter packages" seems fairly clear. It's "Dart packages" that is ambiguous, because it's unclear whether it includes Flutter packages.

mit-mit: I get the general point, but not sure I find "Dart packages" ambiguous when it's listed right next to "Flutter packages". I don't care that strongly though.

kwalrath: I've swapped them and tweaked the comments.

PR #2900, opened 2021-01-31
A simple typo on null-safety page

JHBitencourt:
The word default is duplicated.


parlough:
This one is quite funny. I actually think in this case it may have been intentional as `null` is the _default_ default value. 

Not saying this change is wrong, just enjoyed the sentence 😁

JHBitencourt:
Well, if that's the case and intention, this can be ignored :D

kwalrath:
Thanks for the PR @JHBitencourt, and for the interpretation @parlough! It does seem ambiguous. @munificent wrote this and gets to choose whether to change it. :D

mit-mit:
I'm also of the impression that we mean "the _default_ default". Adding italics like that hopefully would make that clearer?

kwalrath (approved):

PR #2901, opened 2021-02-01
Keep .pub-cache/bin at the end of your PATH

jonasfj:
Just a nice suggestion so it's less likely to overwrite any other commands you have on the system.


kwalrath (approved):
Thanks, @jonasfj!

PR #2914, opened 2021-02-04
Add documentation alert about brew switch being disabled

aaronlademann-wf:
The `brew switch` command is advertised as being the supported method of switching between Dart SDK installs on a Mac. However, versions of Homebrew beyond `2.6.0` no longer support it.

This documentation alert explains how to work around that.

> More context: https://github.com/Homebrew/discussions/discussions/339

### DIFF
--- a/src/tools/sdk/_mac.md
+++ b/src/tools/sdk/_mac.md
@@ -25,10 +25,9 @@ To switch between locally installed Dart releases, use
 $ brew switch dart 2.10.1
 ```
 
-{{site.alert.important}}
-  The `brew switch` command was removed after the `2.6.0` release of Homebrew.
-  
-  To continue using it, you need to explicitly downgrade Homebrew to the `2.6.0` version by running:
+{{site.alert.version-note}}
+  The `brew switch` command was removed after the 2.6 release of Homebrew.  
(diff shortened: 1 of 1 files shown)


kwalrath:
> The `brew switch` command is advertised as being the supported method of switching between Dart SDK installs on a Mac. However, versions of Homebrew beyond `2.6.0` no longer support it.
> 
> This documentation alert explains how to work around that.
> 
> > More context: [Homebrew/discussions#339](https://github.com/Homebrew/discussions/discussions/339)

Thanks for the PR! I hadn't noticed that `brew switch` no longer worked. It makes me slightly nervous to still recommend `brew switch`. Do you think it's useful enough to homebrew users that we should keep those instructions? An alternative is just to drop the `brew switch` discussion, with the understanding that people can download the zipfile if they need an old release.

aaronlademann-wf:
I think its very useful.

kwalrath:
@aaronlademann-wf are my changes OK by you? (Mostly formatting tweaks, but also I changed `2.6.0` to 2.6, since the command shows 2.6.2.)

aaronlademann-wf (commented):
@kwalrath absolutely!

+1

kwalrath (approved):

kwalrath:
Thanks, @aaronlademann-wf!

PR #2915, opened 2021-02-05
Fix the build: make check-code work

kwalrath:


kwalrath:
More to do, but this is a start.

parlough (approved):

kwalrath:
Thanks, @parlough!

tool/dartformat.sh
```diff
@@ -23,6 +23,29 @@ if [[ $NULL_SAFETY == 1 ]]; then
   dart format $* `find $NULL_SAFETY_EXAMPLES -name "*.dart" \
       ! -path "**/.*" \
       ! -path "**/build/**"`
+
+  dart format -l 60 \
```
johnpryan, line 27: What's the reason for `-l 60` here?

kwalrath: We decided a long time ago that these examples needed a shorter line length, probably to avoid scrolling with common page widths.

PR #2919, opened 2021-02-05
update /samples to null safety

johnpryan:
This adds `src/samples/index_migrated.md` and the corresponding snippets to `null_safety_examples/`

This doesn't add samples for null safety or add DartPad embeds, I plan to follow up with PRs for these.

staged: https://jr-dartlang-site.firebaseapp.com/samples/index_migrated.html


src/samples/index_migrated.md
```diff
@@ -0,0 +1,407 @@
...
+
+<?code-excerpt "../../null_safety_examples/misc/test/samples_test.dart (import)" plaster="none"?>
+{% prettify dart tag=pre+code %}
+// Importing core libraries
+import 'dart:math';
+
+// Importing libraries from external packages
+import 'package:test/test.dart';
+
+
```
kwalrath, line 155:
```suggestion
```

kwalrath (commented):
I'm not used to seeing `../../null_safety_examples`. Should it be `../null_safety_examples` globally? (Did you run `pub get; tool/refresh-code-excerpts.sh`?)

I was a little freaked by not having an initial commit of the original files, but then I realized that I could just diff index.md & index_migrated.md!

null_safety_examples/misc/lib/samples/spacecraft.dart
```diff
@@ -0,0 +1,87 @@
...
+class Orbiter extends Spacecraft {
+  double altitude;
+
+  Orbiter(String name, DateTime launchDate, this.altitude)
+      : super(name, launchDate);
+}
+// #enddocregion extends
+
+// #docregion mixin
+class Piloted {
```
parlough, line 41: Can we use an explicit `mixin` here?

null_safety_examples/misc/lib/samples/spacecraft.dart
```diff
@@ -0,0 +1,87 @@
+// ignore_for_file: sort_constructors_first
+// #docregion class
+class Spacecraft {
```
parlough, line 3: Is this supposed to extend `Describable`?

johnpryan: Right now the `Describable` class exists on it's own. This isn't something I changed, though.

null_safety_examples/misc/lib/samples/spacecraft.dart
johnpryan: That makes sense, added in https://github.com/dart-lang/site-www/pull/2919/commits/32c509f52811be97a5ee91ba8d0bc67130d3ec44

null_safety_examples/misc/lib/samples/spacecraft.dart
johnpryan: It seems like it should, on second glance...

johnpryan: Ah, but it would need to implement `describeWithEmphasis` as well. I vote to leave it as-is for now.

parlough: Yeah either way is fine by me! I was just curious if that was the intention. 

I thought it wouldn't need to implement `describeWithEmphasis` if using `extend Describable`, as the class provides a default implementation. If you use `implements Describable` though that would then require an implementation.

johnpryan: ah right. I don't actually know what the intention was here, but I suspect it should extend Describable

johnpryan:
@kwalrath this is ready for another look - I'll follow up with some null safety samples and DartPad support

kwalrath (approved):

PR #2921, opened 2021-02-08
Migrate to GitHub Actions

athomas:


athomas:
Unfortunately, the GitHub Actions UI is pretty poor at displaying allowed failures. `continue-on-error` at least keeps the checks green, but you have to click all the way into the log to see what's going on (see https://github.com/actions/toolkit/issues/399 for the feature request to improve that UI).

athomas:
@kwalrath not sure who would be the best reviewer here, feel free to add more appropriate ones.

This seems to work, but I'm not sure how to test the deploy step. Landing this PR should do a no-op deployment (only change to the site should be in the footer where it links to the build).

tool/analyze-and-test-examples.sh
```diff
@@ -116,7 +116,7 @@ function analyze_and_test() {
   if [[ -z $TEST_FILES ]]; then
     echo "No browser-only tests."
   else
-    travis_fold start analyzeAndTest.tests.browser
+    echo "::group::analyzeAndTest.tests.browser"
     echo Running browser tests ...
     PLATFORM=chrome
     if [[ -n $TRAVIS ]]; then PLATFORM=travischrome; fi
```
mit-mit, line 122: Still a travis reference?

.github/workflows/build.yml
```diff
@@ -0,0 +1,88 @@
...
+      - name: Install node dependencies
+        # TOOD: caching requires a lockfile for node deps, so does npm ci.
+        run: npm install
+      - name: Install dart dependencies
+        run: dart pub get
+      - run: ./tool/shared/write-ci-info.sh -v
+        if: contains(matrix.task, 'build.sh')
+      - run: ${{ matrix.task }}
+        continue-on-error: ${{ matrix.experimental }}
+      - run: ./tool/shared/deploy.sh --robots ok default
```
mit-mit, line 80: I wonder if it would be easier to read if this was a separate job?

tool/analyze-and-test-examples.sh
athomas: Thanks, not sure how I missed that one. Though, the browser tests pass on GitHub Actions without that `--no-sandbox` so I just removed the whole travischrome noise.

.github/workflows/build.yml
athomas: What I did is closest to what Travis used to do. I'm agnostic, the "easier to read" comes at the cost of overall efficiency and latency because the `build.sh` needs to be run twice.

This is what it would look like as a job instead of a step:
```yaml
  deploy:
    needs: test
    runs-on: ubuntu-latest
    if: |
      github.event == 'push'
      && github.ref == 'refs/heads/master'
      && github.repository == 'dart-lang/site-www'
    steps:
      - uses: actions/checkout@v2
        with:
          # submodules param for checkout action to get site-shared, 'true' might be enough
          submodules: "recursive"
      - run: mkdir -p $TMP
      - uses: actions/setup-node@v1
        with:
          node-version: 12
      - uses: ruby/setup-ruby@v1
        with:
          bundler-cache: true
          ruby-version: 2.6.5
      - uses: dart-lang/setup-dart@v0.3
        with:
          sdk: stable
      - name: Install node dependencies
        # TOOD: caching requires a lockfile for node deps, so does npm ci.
        run: npm install
      - name: Install dart dependencies
        run: dart pub get
      - run: ./tool/shared/write-ci-info.sh -v
      - run: ./tool/build.sh
      - run: ./tool/shared/deploy.sh --robots ok default
        env:
          FIREBASE_TOKEN: ${{ secrets.FIREBASE_TOKEN }}
```

mit-mit: Ah, I think what you have originally is good then!

mit-mit (approved):

parlough (approved):

kwalrath (approved):

kwalrath:
OK, let's see what happens! I can always deploy by hand if that part is broken.

.github/workflows/build.yml
athomas: I've changed my mind, we should do this because otherwise the deploy step might run before all tests are finished (PR: #2933).

PR #2922, opened 2021-02-09
/platforms -> /overview[#platform]

kwalrath:
I should've realized we'd need some redirects. :)


mit-mit:
Sorry, I could have sworn I did redirects, but I clearly wasn't very thorough then.

mit-mit (approved):

parlough (approved):

PR #2924, opened 2021-02-09
Update C Interop Tutorial to use new dart tool

shubhanshu02:
Fixes part of #2617 
## Description
Update the C Interop Tutorial (`src/_guides/libraries/c-interop.md`) to use the new dart tool. Also, I noticed the SDK version in the `hello_world` example had a lower bound of 2.6, but the lower bound in the tutorial was 2.5, so I updated it.

@kwalrath, please review the changes.


domesticmouse (approved):

kwalrath (approved):
Thanks, @shubhanshu02!

PR #2928, opened 2021-02-10
Add a different color for the banner links on hover/focus

parlough:
Previously since the custom color for the banner links overrode all color styles, there was no hover/focus style change, making it hard to identify a difference.

With this change, a user can now properly identify which link they focused or hovered.

image


kwalrath (approved):
I like it! Thanks, @parlough!

From the screenshot I was worried that the contrast would be too low, but I deployed locally, and it looks fine to me.

johnpryan (approved):

PR #2933, opened 2021-02-11
Move deploy step into a separate job

athomas:
This prevents it from running until all tests have run and makes it more visible in the Actions UI.


mit-mit (approved):

kwalrath:
Good change, thanks!

PR #2939, opened 2021-02-11
Update Making a POST request from a standalone client in httpserver.md

zakariaz:
Hi,
What i learned in the previous tutorials is that an async function or method returns a `Future` which complete with a value so i think the section explaining the behavior of the `post()` method should use the same terminology so new people like me who are learning Dart will not be confused.

Sorry i saw that i merged it against my master branch XD.
Apologies


kwalrath (approved):
Thank you, @zakariaz!

PR #2945, opened 2021-02-12
Switch from stagehand to webdev for pub global example

parlough:
I thought `webdev` would be a more appropriate example since `stagehand` is not often used by users directly and `dart create` is the more appropriate options these days anyway.

### DIFF
--- a/src/tools/pub/cmd/pub-global.md
+++ b/src/tools/pub/cmd/pub-global.md
@@ -12,13 +12,12 @@ After [activating a package](#activating-a-package), you can
 [Deactivating a package](#deactivating-a-package) removes it from
 your list of globally available packages.
 
-For example, say you want to run
-[Stagehand]({{site.pub}}/packages/stagehand)
-the Dart project generator, from the command line.
+For example, say you want to use [webdev][], to serve
+your Dart web application from the command line.
 
 ```terminal
-$ dart pub global activate stagehand
-$ stagehand
+$ dart pub global activate webdev
+$ webdev serve
 ```
 
 If this doesn't work, you might need to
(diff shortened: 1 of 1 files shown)


src/tools/pub/cmd/pub-global.md
```diff
@@ -12,13 +12,12 @@ After [activating a package](#activating-a-package), you can
 [Deactivating a package](#deactivating-a-package) removes it from
 your list of globally available packages.
 
-For example, say you want to run
-[Stagehand]({{site.pub}}/packages/stagehand)
-the Dart project generator, from the command line.
+For example, say you want to use [webdev][], to serve
```
kwalrath, line 15:
```suggestion
For example, say you want to use [webdev][] to serve
```

src/tools/pub/cmd/pub-global.md
```diff
@@ -137,9 +136,9 @@ may vary for different versions of Windows.
 You can now directly invoke the command:
 
 {% prettify none tag=pre+code %}
-$ mkdir angular_project
-$ cd angular_project
-$ [!stagehand web-angular!]
+$ mkdir web_project
```
kwalrath, line 139: This made sense for stagehand, but let's assume a prepopulated app for webdev.

```suggestion
```

kwalrath (commented):
Thank you, @parlough! Looks good to me after a couple of tweaks.

src/tools/pub/cmd/pub-global.md
parlough: Good point!

parlough:
Thanks for the feedback, it's been addressed :>

kwalrath (approved):

PR #2946, opened 2021-02-12
Switch favicon to new icon

parlough:


src/_includes/head.html
```diff
@@ -15,7 +15,7 @@
   <title>{% if page.short-title %}{{ page.short-title }}{% else %}{{ page.title }}{% endif %} | {{ site.title }}</title>
 
   <!-- Favicon / Touch Icons -->
-  <link rel="icon" sizes="64x64" href="/assets/shared/dart/icon/64.png">
+  <link rel="icon" sizes="64x64" href="{% asset shared/dart/icon/64.png @path %}">
```
johnpryan, line 18: Does this change the icon being used?

parlough: It's a bit weird, but it should. Using this `asset` directive we should get something with the appended hash or whatever. For example:

Old logo:
https://dart.dev/assets/shared/dart/icon/64.png

New logo:
https://dart.dev/assets/shared/dart/icon/64-63351a53a8a87676dbb345d3458246abc3d7cfaa7be1d331952a019866e33b6a.png

johnpryan (approved):

PR #2953, opened 2021-02-13
Add num declaration example to Language Tour

### LINKED ISSUE #1165: Update description & examples of integer and double literals
https://www.dartlang.org/guides/language/language-tour#numbers could use the following changes:

* [x] Consider not using types when declaring most integers/doubles. E.g.:
   `int x = 1;` -> `var x = 1; // The inferred type of x is int`
   `double y = 1.1;` -> `var y = 1.1; // The inferred type of y is double`
* [ ] Maybe show declaring a `num` variable:
   `num x = 1; // x can have both int and double values`
* [x] When 2.1 ships, show using an integer literal to initialize a double, with a note that this didn't work before 2.1. E.g. `double x = 1;`

Related: #1164

micpap25:
Fixes #1165
Adds an example of declaring a `num` in the Language Tour. This will resolve the last item on the list and fix the issue.


kwalrath:
My impression is that `num`s are used less than they were, since you can assign int literals to doubles now. But still, good to cover `num`.

(@micpap25 although this looks like a good change, I'm in the midst of updating the language tour, which is why all the language tour issues are assigned to me right now. I'm not sure that I'll be able to merge this change.)

kwalrath:
Sorry for the long delay!

@micpap25 I like this change, but we'll need to move the included code over to /null_safety_examples. Are you up for doing that?

(@lrhn, last chance to chime in. Any objection to us adding this?)

src/_guides/language/language-tour.md
```diff
@@ -505,6 +505,14 @@ var y = 1.1;
 var exponents = 1.42e5;
```
lrhn, line 505: Not part of this change, but this doesn't cover things like `var exponent2 = 8e5` or `double d = 5;`, which are both doubles without a "decimal" (should that be "decimal point" above, or does "decimal" also mean a fractional digit?)

src/_guides/language/language-tour.md
```diff
@@ -505,6 +505,14 @@ var y = 1.1;
...
 ```
 
+You can also declare a variable as a num. If you do this, the variable
+can have both integer and double values.
+
+<?code-excerpt "misc/lib/language_tour/built_in_types.dart (declare-num)"?>
+```dart
+num x = 1; // x can have both int and double values
+```
```
lrhn, line 514: Would it be worth showing the variable containing a double? Maybe follow it with:
```dart
x += 2.5;
```

lrhn (approved):

micpap25:
Yep, I'll get to this later today.

src/_guides/language/language-tour.md
micpap25: Including `var exponent2 = 8e5;`, I think your point on `double d = 5;` is sufficiently covered by other sections.

kwalrath (approved):

kwalrath:
Build's working! Thanks @micpap25 for making this change and for being willing to return to it after so much time had passed!

micpap25:
No problem. Are language tour issues still on pause for now, or can I move forward with resolving some of those?

kwalrath:
Please feel free to resolve more, @micpap25! The language tour was only paused because of null safety going to stable (which happened in early March).

PR #2962, opened 2021-02-18
Clean up and fix named constructor portion of cheatsheet

parlough:
Switches to an initializer list in the related code to avoid need to use `late` and removes the testing code which depended on the now removed `LateInitializationError`.

Fixes https://github.com/dart-lang/dart-pad/issues/1766


theacodes (approved):

theacodes:
This looks great, thanks for doing this!

PR #2966, opened 2021-02-19
Revamp the guidelines around types.

### LINKED ISSUE #996: Generalize type annotation guidelines
Effective Dart used to have two general guidelines around annotating APIs:

* DO type annotate public APIs
* PREFER type annotating private declarations.

Now the corresponding rules are:

* PREFER type annotating public fields and top-level variables if the type isn’t obvious.
* CONSIDER type annotating private fields and top-level variables if the type isn’t obvious.
* PREFER annotating with dynamic instead of letting inference fail.

Since unannotated parameters and returns usually aren't inferred, I believe the effect of the new guidelines is technically the same, but it certainly ain't obvious. The best fix is probably to just generalize the first two to:

* PREFER type annotating public APIs if the type isn’t obvious.
* CONSIDER type annotating private APIs if the type isn’t obvious.

Then in the prose explain that it applies to top-level variables, fields, parameters, and return types. Explain that in the absence of an initializer, the type is basically never obvious.

### LINKED ISSUE #1387: Clarify ambiguity about explicit types for parameters with initialization expressions
Given the following method (or constructor):

`void foo({color = SomeColorEnum.orange})`

The `color` parameter should have an explicit type, but Effective Dart doesn't explicitly cover that scenario.

Readers may expect it to follow the rule about using [implicit typing for variables / fields that have initializers](https://www.dartlang.org/guides/language/effective-dart/design#avoid-type-annotating-initialized-local-variables), but the compiler doesn't actually apply that same kind of inference for parameters.

In the case of constructors, they may think that the rule about [not type-annotating initializing formals](https://www.dartlang.org/guides/language/effective-dart/usage#dont-type-annotate-initializing-formals) applies, but when it's a regular parameter that doesn't apply either. This can be easy to mistake when changing code like:

`Thing({this.foo = 123})`

to something like:

`Thing({foo = 123}) : _foo = foo;`

munificent:
This is a pretty big textual change, but the actual effective change is
small. It's mostly clarifying existing practices and covering omissions.

Fix #996. Fix #1387.

I'd like to run this by the Dart readability folks before this lands. Is there an easy way I can stage this and share it with them in some kind of readable form?

### DIFF
--- a/null_safety_examples/misc/lib/effective_dart/design_good.dart
+++ b/null_safety_examples/misc/lib/effective_dart/design_good.dart
@@ -239,7 +239,7 @@ void miscDeclAnalyzedButNotTested() {
 
   {
     // #docregion infer-dynamic
-    Map<String, dynamic> readJson() { ellipsis(); }
+    Map<String, dynamic> readJson() => ellipsis();
 
     void printUsers() {
       var json = readJson();
--- a/null_safety_examples/misc/lib/effective_dart/design_good.dart
+++ b/null_safety_examples/misc/lib/effective_dart/design_good.dart
@@ -1,4 +1,4 @@
-// ignore_for_file: type_annotate_public_apis, unused_element, unused_local_variable, sort_constructors_first
+// ignore_for_file: close_sinks, type_annotate_public_apis, unused_element, unused_local_variable, sort_constructors_first
 
 import 'dart:async';
 import 'dart:collection';
@@ -158,6 +158,13 @@ void miscDeclAnalyzedButNotTested() {
   }
 
   {
+    // #docregion non-inferred-type-args
+    var playerScores = <String, int>{};
+    final events = StreamController<Event>();
+    // #enddocregion non-inferred-type-args
+  }
+
+  {
     // #docregion omit-types-on-locals
     List<List<Ingredient>> possibleDesserts(Set<Ingredient> pantry) {
       var desserts = <List<Ingredient>>[];
@@ -170,6 +177,24 @@ void miscDeclAnalyzedButNotTested() {
       return desserts;
     }
     // #enddocregion omit-types-on-locals
+  }
+
+  {
+    // #docregion annotate-return-types
+    String makeGreeting(String who) {
+      return "Hello, $who!";
+    }
+    // #enddocregion annotate-return-types
+  }
+
+  {
+    // #docregion annotate-parameters
+    void sayRepeatedly(String message, {int count = 2}) {
+      for (var i = 0; i < count; i++) {
+        print(message);
+      }
+    }
+    // #enddocregion annotate-parameters
   }
 
   (AstNode node) {
@@ -194,21 +219,34 @@ void miscDeclAnalyzedButNotTested() {
   // #enddocregion inferred-wrong
 
   {
-    // #docregion redundant
-    Set<String> things = Set();
-    // #enddocregion redundant
+    // #docregion explicit
+    var items = Future.value([1, 2, 3]);
+    // #enddocregion explicit
   }
 
   {
-    // #docregion explicit
-    var things = Set<String>();
-    // #enddocregion explicit
+    // #docregion incomplete-generic
+    List<num> numbers = [1, 2, 3];
+    var completer = Completer<Map<String, int>>();
+    // #enddocregion incomplete-generic
   }
 
   {
     // #docregion prefer-dynamic
     dynamic mergeJson(dynamic original, dynamic changes) => ellipsis();
     // #enddocregion prefer-dynamic
+  }
+
+  {
+    // #docregion infer-dynamic
+    Map<String, dynamic> readJson() { ellipsis(); }
+
+    void printUsers() {
+      var json = readJson();
+      var users = json['users'];
+      print(users);
+    }
+    // #enddocregion infer-dynamic
   }
 
   // #docregion avoid-Function
@@ -271,6 +309,33 @@ void miscDeclAnalyzedButNotTested() {
     // #enddocregion avoid-mandatory-param
   };
 }
+
+//----------------------------------------------------------------------------
+
+// #docregion dont-type-init-formals
+class Point1 {
+  double x, y;
+  Point1(this.x, this.y);
+}
+// #enddocregion dont-type-init-formals
+
+//----------------------------------------------------------------------------
+
+// #docregion inferred-type-args
+class Downloader {
+  final Completer<String> response = Completer();
+}
+// #enddocregion inferred-type-args
+
+//----------------------------------------------------------------------------
+
+// #docregion redundant
+class Downloader1 {
+  final Completer<String> response = Completer();
+}
+// #enddocregion redundant
+
+//----------------------------------------------------------------------------
 
 class MyIterable<T> {
   // #docregion function-type-param
--- a/null_safety_examples/misc/lib/effective_dart/design_good.dart
+++ b/null_safety_examples/misc/lib/effective_dart/design_good.dart
@@ -182,7 +182,7 @@ void miscDeclAnalyzedButNotTested() {
   {
     // #docregion annotate-return-types
     String makeGreeting(String who) {
-      return "Hello, $who!";
+      return 'Hello, $who!';
     }
     // #enddocregion annotate-return-types
   }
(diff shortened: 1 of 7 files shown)


domesticmouse (approved):
Looks like some tests are failing. Plz fix?

kwalrath:
> I'd like to run this by the Dart readability folks before this lands. Is there an easy way I can stage this and share it with them in some kind of readable form?

I've staged it here: https://kw-www-dartlang-1.firebaseapp.com/guides/language/effective-dart/design_migrated#type-inference

kwalrath:
I'll wait for the readability folks to have their say before I review this. You might be able to fix the build failure by running `./tool/dartformat.sh` and then `./tool/refresh-code-excerpts.sh`.

parlough (approved):

null_safety_examples/misc/lib/effective_dart/design_good.dart
```diff
@@ -272,6 +310,33 @@ void miscDeclAnalyzedButNotTested() {
...
+class Downloader {
+  final Completer<String> response = Completer();
+}
+// #enddocregion inferred-type-args
+
+//----------------------------------------------------------------------------
+
+// #docregion redundant
+class Downloader1 {
+  final Completer<String> response = Completer();
```
gresrun, line 334: Following "DO type annotate fields and top-level variables if the type isn’t obvious." I would expect this to say:
```dart
class Downloader1 {
  final response = Completer<String>();
}
```
Line 326 is carving out an optional way to restate this but I think we should reinforce here that types should not be duplicated, if possible.

munificent: We don't consider `Foo foo = Foo()` to be repeating the type because the thing on the right is a constructor call, not a type annotation. And for fields and top-level variables, I think there is a general preference (but not a hard rule) towards annotating the variable instead of putting type arguments on the invocation.

munificent: I'm gonna mark this resolved because I think the guidelines here say what we intend to say (and are in line with what I think has long been the recommended guidance), but I'm still open to this discussion if you want to.

src/_guides/language/effective-dart/design_migrated.md
```diff
@@ -1218,90 +1289,167 @@ var names = people.map((person) => person.name);
...
+  double x, y;
+  Point(double this.x, double this.y);
+}
+{% endprettify %}
+
+
+### DO write type arguments on generic invocations that aren't inferred.
+
+Dart is pretty smart about inferring type arguments in generic invocations. It
+will look at the expected type where the expression occurs and the types of
```
kwalrath, line 1327:
```suggestion
looks at the expected type where the expression occurs and the types of
```

We try to stick to present tense: https://developers.google.com/style/tense

src/_guides/language/effective-dart/design_migrated.md
```diff
@@ -1218,90 +1289,167 @@ var names = people.map((person) => person.name);
...
+{% prettify dart tag=pre+code %}
+var playerScores = {};
+final events = StreamController();
+{% endprettify %}
 
-A type argument is redundant if inference would fill in the same type. If the
-invocation is the initializer for a type-annotated variable, or is an argument
-to a function, then inference usually fills in the type for you:
+Sometimes the invocation occurs as the initializer to a variable declaration. If
+variable is *not* a local then instead of writing the type argument list on the
```
kwalrath, line 1347:
```suggestion
the variable is *not* local, then instead of writing the type argument list on the
```

src/_guides/language/effective-dart/design_migrated.md
```diff
@@ -1218,90 +1289,167 @@ var names = people.map((person) => person.name);
...
 
-### DO annotate when Dart infers the wrong type.
+### AVOID writing incomplete generic types.
 
-Sometimes, Dart infers a type, but not the type you want. For example, you may
-want a variable's type to be a supertype of the initializer's type so that you
-can later assign some other sibling type to the variable:
+When writing a type annotation or type argument, it's still possible to not pin
+down a complete type by writing the name of a generic type but omitting its type
+arguments. In Java, these are called "raw types". For example:
```
kwalrath, line 1415: I'm having a hard time parsing the first sentence. Could you please rewrite it?

src/_guides/language/effective-dart/design_migrated.md
```diff
@@ -1218,90 +1289,167 @@ var names = people.map((person) => person.name);
...
-assign double values from `scores` to it. In cases like this, explicit
-annotations make sense.
+Here, `numbers` has a type annotation, but the annotation doesn't provide a type
+argument to the generic `List`. Likewise, the `Map` type argument to `Completer`
+isn't fully specified. In cases like this, Dart will *not* try to "fill in" the
+rest of the type for you using the surrounding context. Instead, it silently
+fills in any missing type arguments with with `dynamic` (or the bound if the
+class has one). That's rarely what you want.
 
+Instead, if you are writing a generic type in a type annotation or as a type
```
kwalrath, line 1431: This sentence is a bit hard to parse. Here's an attempt at making it a bit clearer grammatically:

```suggestion
Instead, if you're writing a generic type either in a type annotation or as a type
```

src/_guides/language/effective-dart/design_migrated.md
```diff
@@ -1218,90 +1289,167 @@ var names = people.map((person) => person.name);
...
+### DO annotate with `dynamic` instead of letting inference fail.
 
-However, it's not the most *clear* way. A casual reader of your code who sees an
-annotation is missing has no way of knowing if you intended it to be `dynamic`,
-expected inference to fill in some other type, or simply forgot to write the
-annotation.
+When inference doesn't fill in a type, it usually defaults to `dynamic`. If
+`dynamic` is the type you want, this is technically the most terse way to get
+it. However, it's not the most *clear* way. A casual reader of your code who
+sees an annotation is missing has no way of knowing if you intended it to be
```
kwalrath, line 1447:
```suggestion
sees that an annotation is missing has no way of knowing if you intended it to be
```

src/_guides/language/effective-dart/design_migrated.md
```diff
@@ -1315,6 +1463,26 @@ dynamic mergeJson(dynamic original, dynamic changes) => ...
...
+  var users = json['users'];
+  print(users);
+}
+{% endprettify %}
+
+Here, Dart infers `Map<String, dynamic>` for `json` and then from that infers
+`dynamic` for `users`. It's fine to leave `users` without a type annotation. The
+distinction is a little subtle. It's OK to allow inference to *propagate*
+`dynamic` through your code from a `dynamic` type annotation somewhere else, but
+you don't want it to summon `dynamic` *ex nihilo* when nothing in your code ever
```
kwalrath, line 1484: I'd never seen *ex nihilo* before, and I've seen a lot of Latin! I'm going to pretend I didn't see it, since it's probably against style but is totally cool (and doesn't really affect the meaning of the sentence).

kwalrath (approved):
Some suggestions (because I'm me), but in general this is clear and a fun read.

src/_guides/language/effective-dart/design_migrated.md
```diff
@@ -1176,23 +1198,72 @@ List<List<Ingredient>> possibleDesserts(Set<Ingredient> pantry) {
...
-<?code-excerpt "design_good.dart (uninitialized-local)"?>
+<?code-excerpt "design_good.dart (annotate-return-types)"?>
 {% prettify dart tag=pre+code %}
-List<AstNode> parameters;
-if (node is Constructor) {
-  parameters = node.signature;
-} else if (node is Method) {
-  parameters = node.parameters;
+String makeGreeting(String who) {
+  return "Hello, $who!";
```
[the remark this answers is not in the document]
munificent: Fixed. (I will seemingly never break this habit.)

src/_guides/language/effective-dart/design_migrated.md
```diff
@@ -1176,23 +1198,72 @@ List<List<Ingredient>> possibleDesserts(Set<Ingredient> pantry) {
...
+{% endprettify %}
+
+There are two cases that seem like exceptions but aren't:
+
+*   Function *expressions* ("lambdas" or "anonymous functions") do infer a
+    return type from the body. In fact, the syntax doesn't even allow a return
+    type annotation. But these aren't what we consider function *declarations*.
+
+*   The return type of an overriding method can be inferred from the method it
+    overrides. However, this is a rarely-known feature and even in this case
```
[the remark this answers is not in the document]
munificent: I like "little-known".

src/_guides/language/effective-dart/design_migrated.md
munificent: Done!

munificent:
Thanks for the review!

PR #2968, opened 2021-02-19
Remove extra comma causing scss parse warning

parlough:
Remove extra comma causing scss parse warning


kwalrath (approved):
Thanks, @parlough!

PR #2970, opened 2021-02-19
Flip is! explanation to improve clarity - Fixes #2913

### LINKED ISSUE #2913: is! description should be flipped
https://dart.dev/guides/language/language-tour#type-test-operators

The operator/description table looks like this:

Operator | Meaning
-- | --
as | Typecast (also used to specify library prefixes)
is | True if the object has the specified type
is! | False if the object has the specified type
-- | --

But the actual behaviour for `is!` should say: False if the object **doesn't** have the specified type

parlough:


kwalrath:
Thanks for making the change in both places, @parlough! Fortunately, I'm not in the guts of migrating the language tour at this moment, so I can accept this.

kwalrath (approved):

PR #2973, opened 2021-02-19
Centralize DartPad embed styling - Fixes #2951

### LINKED ISSUE #2951: Centralize embedded DartPad/iframe styling
A lot of files with embedded DartPads have the following at the top:

```
{% comment %}
TODO(chalin): move this into one of our SCSS files
{% endcomment -%}
<style>
iframe[src^="https://dartpad"] {
  border: 1px solid #ccc;
  margin-bottom: 1rem;
  min-height: 150px;
  resize: vertical;
  width: 100%;
}
</style>
```

We should, you know, actually do that. :)

parlough:
I decided to add the centralized style to the `_code.scss` file as it seemed the most appropriate.

To avoid changing the styling for the homepage try dart embed, I added some code to the `dartpad_picker` handling to set an id of `try-dart-pad` on the `iframe` so we can avoid selecting on that one.

As for the `min-height` of `400px`, it was the largest currently used and easiest to share of the choices. Also I found the shorter min-heights made the UI hard to use, especially once there were issues with code. I'm open to picking a different default or overriding this style locally, but I think you'll be happy with the default.

I recommend trying this change out locally to verify it matches what you expect as well.

### DIFF
--- a/src/_assets/css/_code.scss
+++ b/src/_assets/css/_code.scss
@@ -18,6 +18,15 @@ pre {
   }
 }
 
+// Border and resizability for dartpad
+iframe[src^="https://dartpad"]:not(#try-dart-pad) {
+  border: 1px solid #ccc;
+  margin-bottom: 1rem;
+  min-height: 400px;
+  resize: vertical;
+  width: 100%;
+}
+
 @media screen {
   .com { color: #6E6E70; }
   // TODO(chalin): dropping the following overrides would allow us to be consistent with Flutter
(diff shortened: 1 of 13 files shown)


kwalrath:
Nice, @parlough! This TODO has bugged me every time I work on a page with embedded DartPads.

johnpryan (approved):
looks great 👍

kwalrath (approved):
LGTM. I did a spot check of the rendered pages, and I'm fine with seeing how people like the (often) larger DartPads. Thank you, @parlough!

PR #2974, opened 2021-02-19
Add mention of stream_transform to useful libraries - Fixes #2261

### LINKED ISSUE #2261: Link to stream_transform package
Perhaps we should add a link from https://dart.dev/guides/libraries/useful-libraries to the stream_transform package: https://pub.dev/packages/stream_transform

/cc @natebosch @mit-mit

parlough:


kwalrath (commented):
LGTM, but I'd like an engineer to confirm.

@natebosch maybe?

natebosch:
I would prefer not to merge this. After the dust settles on the null safety releases we are likely going to merge that package into `package:async` under a separate top level import - so it would be `package:async/stream_transform.dart`. https://github.com/dart-lang/stream_transform/issues/111

If we decide not to do that for some reason I'll make a note in that issue to revive this.

parlough:
:O I didn't realize that was happening, pretty exciting then :) 

I'll close this for now, thanks for the heads up!

PR #2804, opened 2020-12-15
Start adding pages for `dart` commands

kwalrath:
* Add /tools/dart-compile and /tools/dart-analyze
* Remove /tools/dart2native, /tools/dartanalyzer
* Update tool index pages
* Update pages that refer to compilation and analyzing tools
* Add/update redirects

Staged:
https://kw-dartlang-3.firebaseapp.com/tools/dart-compile
https://kw-dartlang-3.firebaseapp.com/tools/dart-analyze

Contributes to #2719
Fixes #2858

PR #2978, opened 2021-02-19
Create /tools/dart-fix

kwalrath:
Once #2804 is merged, we can update its table to point to this page.

Staged: https://kw-staging-dartlang-2.firebaseapp.com/tools/dart-fix

Fixes #2978

### DIFF
--- a/src/tools/dart-fix.md
+++ b/src/tools/dart-fix.md
@@ -1,0 +1,31 @@
+---
+title: The dart fix command
+description: Help for bringing your code up-to-date.
+toc: false
+---
+
+The `dart fix` command finds and fixes two types of issues:
+
+* Analysis issues that have associated automated fixes
+* Issues that have associated package API migration information
+
+{{site.alert.info}}
+  For more information about this and other `dart` commands,
+  see the [Dart command-line tool page](/tools/dart-tool).
+{{site.alert.end}}
+
+To preview proposed changes, use the `--dry-run` flag:
+
+```terminal
+$ dart fix --dry-run
+```
+
+To apply the proposed changes, use the `--apply` flag:
+
+```terminal
+$ dart fix --apply
+```
+
+For information about how automated package API changes work,
+see [dart.dev/go/dart-fix](/go/dart-fix).
+
--- a/src/tools/dart-fix.md
+++ b/src/tools/dart-fix.md
@@ -7,6 +7,7 @@ toc: false
 The `dart fix` command finds and fixes two types of issues:
 
 * Analysis issues that have associated automated fixes
+  (sometimes called _quick fixes_)
 * Issues that have associated package API migration information
 
 {{site.alert.info}}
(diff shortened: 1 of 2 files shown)


src/tools/dart-fix.md
```diff
@@ -0,0 +1,31 @@
+---
+title: The dart fix command
+description: Help for bringing your code up-to-date.
+toc: false
+---
+
+The `dart fix` command finds and fixes two types of issues:
+
+* Analysis issues that have associated automated fixes
```
devoncarew, line 9: We also refer to the automated fixes as 'quick fixes' - I don't know if we want to mention that name here or not.

devoncarew (commented):
lgtm!

cc @bwilkerson

devoncarew (approved):

src/tools/dart-fix.md
kwalrath: Yes, I think it's good to mention synonyms that people might look for, for searchability/understability reasons. Thanks!

devoncarew:
Actually, thinking about it, I suspect that we'll want the content of https://github.com/dart-lang/sdk/blob/master/pkg/dartdev/doc/dart-fix.md in-lined into here (and then we'd likely remove the markdown file in the sdk repo).

kwalrath:
> Actually, thinking about it, I suspect that we'll want the content of https://github.com/dart-lang/sdk/blob/master/pkg/dartdev/doc/dart-fix.md in-lined into here (and then we'd likely remove the markdown file in the sdk repo).

If I don't get to that in this PR, I'll file a bug to do that.

A couple of questions:
* What happens if you don't include a command-line flag when you invoke `dart fix`? 
* If you don't want all of the changes it suggests, do you need to edit your code by hand? (either to undo some of the changes, or to apply/tweak the changes you want)

devoncarew:
> * What happens if you don't include a command-line flag when you invoke `dart fix`?

It prints the help for the command and exits without making any changes on disk.

> * If you don't want all of the changes it suggests, do you need to edit your code by hand? (either to undo some of the changes, or to apply/tweak the changes you want)

It's an all-or-nothing tool. You can preview the changes that it would make with `--dry-run`. If there are some you don't want, currently you options are to apply the ones you do want by hand, or, use git, apply all the changes, and back out the ones you don't want.

src/tools/dart-fix.md
bwilkerson: The term 'quick-fix' was is used by IntelliJ and Eclipse. VS Code (and hence LSP) uses the term 'code action'. We should probably include both to cover a wider range of users.

bwilkerson (approved):

kwalrath:
I'm going to commit this PR and file an issue to make it replace https://github.com/dart-lang/sdk/blob/master/pkg/dartdev/doc/dart-fix.md.

PR #2979, opened 2021-02-20
Update the equality operator Effective Dart guideline for null safety.

munificent:
Missed this one in the initial migration. Nothing major here, just rationalizing with how `==()` works with null safety.


kwalrath (commented):
I think the wording could be clearer, but aside from that this change looks good.

Staged: https://kw-staging-dartlang-2.firebaseapp.com/guides/language/effective-dart/design_migrated#dont-make-the-parameter-to--nullable

kwalrath:
Should we say something about the non-null-safe case, in a version-note? I'm thinking that people might assume they should check for null if they haven't opted into null safety.

This comment applies to more than just this change, of course. Although we want to encourage null safety and consider it the default, we might be confusing people who are stuck with non-null-safe code for a while. We could even introduce a new note type so it'll be easy to take out these comments in the future.

kwalrath (approved):

munificent:
> Should we say something about the non-null-safe case, in a version-note? I'm thinking that people might assume they should check for null if they haven't opted into null safety.

Nope, even then they don't need to. (That fact is *why* we made the method take a non-nullable parameter.) Added a note to explain that.

> This comment applies to more than just this change, of course. Although we want to encourage null safety and consider it the default, we might be confusing people who are stuck with non-null-safe code for a while. We could even introduce a new note type so it'll be easy to take out these comments in the future.

I'm hesitant to add too many notes like this because it might just make things more confusing. But I think it's probably a good idea in a few places like this one.

PR #2980, opened 2021-02-20
Add null safety guidance

### LINKED ISSUE #2445: Update Effective Dart for null safety
Examples:
* Don't use `late final` for a local variable.
* Use `:` syntax in constructors when possible, rather than `late final`.

What other examples might we want?

/cc @munificent @mit-mit @johnpryan @kevmoo

munificent:
The first commit just moves some existing rules around to get ready for more. The second commit adds a new guideline and adds an exception to an existing one.

I've got a couple more planned but I'll do those in a separate PR.

Fix #2445. (There are other null safety changes coming, but this fixes the one concrete request on that issue.)

### DIFF
--- a/src/_guides/language/effective-dart/usage_migrated.md
+++ b/src/_guides/language/effective-dart/usage_migrated.md
@@ -191,7 +191,7 @@ void error([String? message = null]) {
 {% endprettify %}
 
 
-### DO use `??` to convert `null` to a boolean value.
+### PREFER using `??` to convert `null` to a boolean value.
 
 This rule applies when an expression can evaluate to `true`, `false`, or `null`,
 and you need to pass the result to something that expects a non-nullable boolean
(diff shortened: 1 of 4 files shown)


src/_guides/language/effective-dart/usage_migrated.md
```diff
@@ -240,6 +240,92 @@ preferred for three main reasons:
...
+  }
+
+  return 0;
+}
+{% endprettify %}
+
+
+### CONSIDER copying a nullable field to a local variable to enable type promotion.
+
+Checking that a nullable variable is not `null` promotes it to a non-nullable type so that you can then call methods on the variable. Unfortunately, that is only sound for local variables and parameters, so fields and top-level variables are not promoted.
```
natebosch, line 277: [nit] Some other content is formatted to 80 columns. Should this be as well?

src/_guides/language/effective-dart/usage_migrated.md
```diff
@@ -172,6 +240,93 @@ preferred for three main reasons:
...
```
natebosch, line 277: Do we expect readers to understand that "call methods" also covers reading/writing fields/getters?

natebosch, line 277: [nit]  "is not `null`" sounds closer to `is! Null` to me than `!= null`. Should this say "Checking that a nullable variable is not equal to `null`"? Or is that too pedantic?

natebosch (approved):

src/_guides/language/effective-dart/usage_migrated.md
kwalrath: Yes, all content except maybe long URLs should be 80 columns or fewer. (We prefer [semantic line breaks](https://github.com/dart-lang/site-shared/blob/master/doc/writing-for-dart-and-flutter-websites.md#semantic-line-breaks), but Bob gets a religious exemption.)

munificent: Oops, yes.

src/_guides/language/effective-dart/usage_migrated.md
munificent: Good point. Reworded.

munificent: Reworded.

natebosch (approved):

parlough (approved):
Looks good to me after the new clarifications!

src/_guides/language/effective-dart/usage_migrated.md
```diff
@@ -172,6 +240,100 @@ preferred for three main reasons:
     expression is `null`. With `== true`, you have to think through the boolean
     logic to realize that means that a `null` gets converted to *false*.
 
+**Exception:** Using a null-aware operator on a variable inside a condition does
+not promote the variable to a non-nullable type. If you want the variable to be
+promoted inside the body of the if statement, it may be better to use an
+explicit `!= null` check instead of `?.` followed by `??`:
```
kwalrath, line 246:
```suggestion
**Exception:** Using a null-aware operator on a variable inside a condition
doesn't promote the variable to a non-nullable type. If you want the variable
to be promoted inside the body of the `if` statement, it might be better to use
an explicit `!= null` check instead of `?.` followed by `??`:
```

For the may -> might change, see https://developers.google.com/style/word-list#letter-m.

I have to point out that these diffs would've been easier to read if we were using semantic line breaks. (I'll try not to proselytize any more!)

src/_guides/language/effective-dart/usage_migrated.md
```diff
@@ -172,6 +240,100 @@ preferred for three main reasons:
...
+{% endprettify %}
+
+
+### CONSIDER copying a nullable field to a local variable to enable type promotion.
+
+Checking that a nullable variable is not equal to `null` promotes the variable
+to a non-nullable type. That lets you access members on the variable and pass it
+to functions expecting a non-nullable type. Unfortunately, promotion is only
+sound for local variables and parameters, so fields and top-level variables are
+not promoted.
```
kwalrath, line 281:
```suggestion
sound for local variables and parameters, so fields and top-level variables
aren't promoted.
```

src/_guides/language/effective-dart/usage_migrated.md
```diff
@@ -172,6 +240,100 @@ preferred for three main reasons:
...
+
+One pattern to work around this is to copy the field's value to a local
+variable. Null checks on that variable do promote and then you can safely treat
```
kwalrath, line 284:
```suggestion
variable. Null checks on that variable do promote, so you can safely treat
```

src/_guides/language/effective-dart/usage_migrated.md
```diff
@@ -172,6 +240,100 @@ preferred for three main reasons:
...
+          "(error code ${response.errorCode}): ${response.reason}.";
+    }
+
+    return "Could not upload (no response).";
+  }
+}
+{% endprettify %}
+
+This can be cleaner and safer than using `!` every place the field or top level
+variable is used.
```
kwalrath, line 309:
```suggestion
Copying to a local variable can be cleaner and safer than using `!` every place
the field or top-level variable is used:
```

src/_guides/language/effective-dart/usage_migrated.md
```diff
@@ -172,6 +240,100 @@ preferred for three main reasons:
...
+
+    return "Could not upload (no response).";
+  }
+}
+{% endprettify %}
+
+This pattern must be applied with some care. If you need to write back to the
+field make sure you do so, and not just to the local variable. Also, if there
+may be other changes to the field while the local is still in scope, it may end
+up having a stale value. Sometimes it is best to simply use `!` on the field.
```
kwalrath, line 334:
```suggestion
Be careful when using a local copy. If you need to write back to the field,
then make sure you do so, and don't just write to the local variable. Also, if
the field might change while the local is still in scope, then the local might
have a stale value. Sometimes it's best to simply use `!` on the field.
```

kwalrath (commented):
Some wording suggestions, but otherwise LGTM

src/_guides/language/effective-dart/usage_migrated.md
munificent: I know semantic line breaks are better. You're absolutely right. They just hurt my soul to look at. I'm sorry.

munificent:
Thanks for all the improvements!

munificent:
@kwalrath, I think I need the official approval from you. :)

kwalrath (approved):

src/_guides/language/effective-dart/usage_migrated.md
kwalrath: It helps to think of them as free-form poetry. (I'm probably using that terminology wrong; @RedBrogdon would know the real poetical terms.)

PR #2982, opened 2021-02-21
Move dartfmt to dart-format

kwalrath:
I made minimal changes to this page. Is anything crucial missing/wrong/misleading? I'd also like to hear about what could be improved but isn't urgent... but probably won't fix those in this PR.

Staged: https://kw-dartlang-3.firebaseapp.com/tools/dart-format

Contributes to #2719


munificent (approved):
Looks great, thanks!

You might want to show an example of using `dart format .` to format all files in the current directory (recursively), since that's the most common way to run it.

PR #2986, opened 2021-02-23
More safety changes for "Effective Dart".

munificent:
- Add note suggesting late fields.
- Add guideline about late public fields.
- Remove outdated guideline about nullable primitives.
- Add guideline about nullable container types.
- Explain initializing late fields at the declaration.
- Add guideline encouraging constructor initialization lists.

### DIFF
--- a/null_safety_examples/misc/lib/effective_dart/usage_bad.dart
+++ b/null_safety_examples/misc/lib/effective_dart/usage_bad.dart
@@ -374,6 +374,18 @@ class Point0 {
 
 //----------------------------------------------------------------------------
 
+// #docregion late-init-list
+class Point1 {
+  late double x, y;
+  Point1.polar(double theta, double radius) {
+    x = cos(theta) * radius;
+    y = sin(theta) * radius;
+  }
+}
+// #enddocregion late-init-list
+
+//----------------------------------------------------------------------------
+
 // #docregion semicolon-for-empty-body
 class Point2 {
   double x, y;
--- a/src/_guides/language/effective-dart/design_migrated.md
+++ b/src/_guides/language/effective-dart/design_migrated.md
@@ -947,11 +947,12 @@ rarely what you want. Fields are usually marked late so that they can be
 initialized *internally* at some point in the instance's lifetime, often inside
 the constructor body.
 
-Unless you do want users to call the setter, it's better to either:
+Unless you *do* want users to call the setter, it's better to pick one of the
+following solutions:
 
-* not use `late`,
-* ensure the `late` field is initialized at its declaration,
-* or make the `late` field private and define a public getter to access it.
+* Don't use `late`.
+* Use `late`, but initialize the `late` field at its declaration.
+* Use `late`, but make the `late` field private and define a public getter for it.
 
 ### AVOID returning nullable `Future`, `Stream`, and collection types.
 
--- a/src/_guides/language/effective-dart/design_migrated.md
+++ b/src/_guides/language/effective-dart/design_migrated.md
@@ -947,10 +947,11 @@ rarely what you want. Fields are usually marked late so that they can be
 initialized *internally* at some point in the instance's lifetime, often inside
 the constructor body.
 
-Unless you do intend to expose the setter publicly and allow users to invoke it,
-it's better to not use `late`, ensure the `late` field is initialized at its
-declaration, or make the `late` field private with a public getter.
+Unless you do want users to call the setter, it's better to either:
 
+* not use `late`,
+* ensure the `late` field is initialized at its declaration,
+* or make the `late` field private and define a public getter to access it.
 
 ### AVOID returning nullable `Future`, `Stream`, and collection types.
 
--- a/src/_guides/language/effective-dart/design_migrated.md
+++ b/src/_guides/language/effective-dart/design_migrated.md
@@ -785,6 +785,11 @@ Of course, it is often useful to have mutable data. But, if you don't need it,
 your default should be to make fields and top-level variables `final` when you
 can.
 
+Sometimes an instance field doesn't change after it has been initialized, but
+can't be initialized until after the instance is constructed. For example, it
+may need to reference `this` or some other field on the instance. In cases like,
+that, consider making the field `late final`.
+
 
 ### DO use getters for operations that conceptually access properties.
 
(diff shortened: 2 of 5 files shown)


kwalrath:
I've staged this: https://kw-www-dartlang-1.firebaseapp.com/guides/language/effective-dart/usage_migrated

kwalrath:
I'll wait for @natebosch to review this before I take a look

src/_guides/language/effective-dart/usage_migrated.md
```diff
@@ -1152,6 +1161,45 @@ named parameter whose name doesn't match the name of the field you are
...
+  late double x, y;
+  Point.polar(double theta, double radius) {
+    x = cos(theta) * radius;
+    y = sin(theta) * radius;
+  }
+}
+{% endprettify %}
+
+
+The initialization list gives you access to constructor parameters and lets you initialize fields before they can be read. So, if it's possible to use that, that's better than making the field `late` and losing some static safety and performance.
```
natebosch, line 1200: [nit] 80 columns?

null_safety_examples/misc/lib/effective_dart/usage_bad.dart
```diff
@@ -374,6 +374,18 @@ class Point0 {
 
 //----------------------------------------------------------------------------
 
+// #docregion late-init-list
+class Point1 {
+  late double x, y;
+  Point1.polar(double theta, double radius) {
+    x = cos(theta) * radius;
```
natebosch, line 381: Should we look for a case where the assignment cannot happen in the initializer list?

src/_guides/language/effective-dart/design_migrated.md
```diff
@@ -933,20 +938,31 @@ not intended to be invoked from Dart code and don't need a corresponding getter.
...
-mindset.
+When an API returns a container type, it has two ways to indicate the absence of
+data: It can return an empty container or it can return `null`. Users generally
+assume and prefer that you use an empty container to indicate "no data". That
+way, they have a real object that they can call methods on like `isEmpty`.
 
-It can be occasionally useful to have a "nullable primitive" type in your API,
-for example to indicate the absence of a value for some key in a map, but these
-should be rare.
+Prefer returning an empty collection, a future that doesn't complete, or a
```
natebosch, line 960: A future that doesn't complete is not often what users would want. More likely you want a non-nullable future with a nullable generic.

src/_guides/language/effective-dart/design_migrated.md
```diff
@@ -785,6 +785,11 @@ Of course, it is often useful to have mutable data. But, if you don't need it,
 your default should be to make fields and top-level variables `final` when you
 can.
 
+Sometimes an instance field doesn't change after it has been initialized, but
+can't be initialized until after the instance is constructed. For example, it
+may need to reference `this` or some other field on the instance. In cases like,
+that, consider making the field `late final`.
```
natebosch, line 791:
```suggestion
that, consider making the field `late final`. When a field is `late final` it may also be possible to use an initializing expression on the field declaration.
```

src/_guides/language/effective-dart/design_migrated.md
```diff
@@ -933,20 +938,31 @@ not intended to be invoked from Dart code and don't need a corresponding getter.
 [angular]: {{site.angulardart}}
 
 
-### AVOID returning `null` from members whose return type is `bool`, `double`, `int`, or `num`.
+### AVOID public `late final` fields.
+
+Unlike other final fields, a late final field *does* define a setter. If
```
natebosch, line 943:
```suggestion
Unlike other final fields, a late final field without an initializing expression *does* define a setter. If
```

src/_guides/language/effective-dart/usage_migrated.md
```diff
@@ -1110,6 +1110,15 @@ class ProfileMark {
...
 {% endprettify %}
 
+One common reason that fields are *not* initialized at the declaration is
+because they need to reference `this` or call methods on the containing
+instance. However, if the field is marked `late`, then the initializer *can*
+access `this`. In cases where the initialization doesn't depend on constructor
+parameters, you may be able to move the initialization to the field's
+declaration. That doesn't necessarily mean you *should* use `late` just to be
+able to initialize at the declaration. But if you want the field to be `late`,
+it gives you additional flexibility.
```
natebosch, line 1120: I don't think I understand this.

Why else would someone want the field to be late?

Is an author going to read this an push back against a CR comment saying "make this field late final so we can move the initialization out of the constructor" with "You should only do that if you want the field to be `late` and I don't have another reason to want that"?

src/_guides/language/effective-dart/design_migrated.md
munificent: Ah, good call. Done.

src/_guides/language/effective-dart/design_migrated.md
munificent: Good suggestion. I'll reword this a little and apply it to my change.

src/_guides/language/effective-dart/design_migrated.md
munificent: 😮 Today I learned.

Reworded and updated the guideline title too.

src/_guides/language/effective-dart/usage_migrated.md
munificent: My intent was to not come across as strongly encouraging users to use `late` unless they actually need it. But I guess you're probably right that in cases where this applies, they probably *do* want to use `late` so it's fine. Removed those sentences.

null_safety_examples/misc/lib/effective_dart/usage_bad.dart
munificent: I don't think so. "Effective Dart" is already getting pretty big and cumbersome, so I'm hesitant to start introducing counterexamples to show when guidelines *don't* apply. That road seems to lead towards a lot of verbosity.

natebosch (approved):

kwalrath:
Looking now...

src/_guides/language/effective-dart/usage_migrated.md
```diff
@@ -1110,6 +1110,11 @@ class ProfileMark {
 }
 {% endprettify %}
 
+One common reason that fields are *not* initialized at the declaration is
+because they need to reference `this` or call methods on the containing
+instance. However, if the field is marked `late`, then the initializer *can*
+access `this`.
```
kwalrath, line 1116: This can be read as if fields always need access to `this`. How about something like this (which I think you could improve):

```suggestion
Some fields can't be initialized at declaration because they need to reference
`this` — to use other fields or call methods, for example. However, if the
field is marked `late`, then the initializer *can* access `this`.
```

src/_guides/language/effective-dart/usage_migrated.md
```diff
@@ -1152,6 +1157,48 @@ named parameter whose name doesn't match the name of the field you are
 initializing. But when you *can* use initializing formals, you *should*.
 
 
+### DON'T use `late` when a constructor initialization list will do.
+
+Sound null safety requires Dart to ensure that a non-nullable field must be
```
kwalrath, line 1162: "ensure" and "must be" seem redundant. Maybe:

```suggestion
Sound null safety requires Dart to ensure that a non-nullable field is
```

src/_guides/language/effective-dart/usage_migrated.md
```diff
@@ -1152,6 +1157,48 @@ named parameter whose name doesn't match the name of the field you are
...
+  Point.polar(double theta, double radius) {
+    x = cos(theta) * radius;
+    y = sin(theta) * radius;
+  }
+}
+{% endprettify %}
+
+
+The initialization list gives you access to constructor parameters and lets you
+initialize fields before they can be read. So, if it's possible to use that,
```
kwalrath, line 1197:
```suggestion
initialize fields before they can be read. So, if it's possible to use an initialization list,
```

(There was quite a distance between "that" and its antecedent.)

Also... I just noticed that you use _initialization list_ but the language tour uses _initializer list_. I just checked the language spec, and it uses _initializer list_. Could you make that change throughout Effective Dart?

kwalrath (commented):
Some wording suggestions

src/_guides/language/effective-dart/usage_migrated.md
munificent: Will do.

src/_guides/language/effective-dart/design_migrated.md
```diff
@@ -933,20 +939,33 @@ not intended to be invoked from Dart code and don't need a corresponding getter.
...
+define a setter. If that field is public, then the setter is public. This is
+rarely what you want. Fields are usually marked late so that they can be
+initialized *internally* at some point in the instance's lifetime, often inside
+the constructor body.
+
+Unless you do want users to call the setter, it's better to either:
+
+* not use `late`,
+* ensure the `late` field is initialized at its declaration,
+* or make the `late` field private and define a public getter to access it.
```
kwalrath, line 954: So much easier to read! Some further tweaks... Apparently, "either <3 things>" is controversial (https://developers.google.com/style/word-list#letter-e). Also, list items should be parallel. So maybe:

```suggestion
Unless you *do* want users to call the setter, it's better to pick one of the
following solutions:

* Don't use `late`.
* Use `late`, but initialize the `late` field at its declaration.
* Use `late`, but make the `late` field private and define a public getter for it.
```

kwalrath (approved):
One fix, then LGTM!

PR #2989, opened 2021-02-23
Simplify null safety homepage

kwalrath:
* Add instructions for creating a null-safe app/package from scratch
* Move tour into its own page (it might go away soon, in favor of the null safety codelab)

Staged:
https://dartlang-org-staging-0.firebaseapp.com/null-safety
https://dartlang-org-staging-0.firebaseapp.com/null-safety/tour


src/null-safety/index.md
```diff
@@ -314,9 +60,23 @@ To find the most recent releases, see the
 Dart SDK archive, or the **Beta channel** section of the
 [Flutter SDK archive.][flutter-sdks]
 
+### Creating a null-safe package or app
 
-Set the [SDK constraints](/tools/pub/pubspec#sdk-constraints)
```
mit-mit, line 318: I feel the migration use case is getting a bit lost in between the `## Enabling null safety` section and the `### Behind the scenes` below.

I think I'd prefer something like:

```
## Enabling null safety

Current intro...

### Migrating existing packages or apps

Migrating existing code require a number of steps; see the [migration guide][].

### Creating a null-safety package or app
...
```

kwalrath:
ptal @mit-mit. I've updated the staged version: https://dartlang-org-staging-0.firebaseapp.com/null-safety

src/null-safety/index.md
```diff
@@ -314,9 +60,29 @@ To find the most recent releases, see the
...
+
+For instructions on how to migrate your code to null safety,
+see the [migration guide][].
+
+
+### Creating a null-safe package or app {#create}
 
-Set the [SDK constraints](/tools/pub/pubspec#sdk-constraints)
-to require a [language version][] that has null safety support.
+The templates used by the [`dart create` command][`dart create`] and IDEs
```
mit-mit, line 71: nit: i find this a bit hard to read with the link being both on the `dart create` and the word 'command'. I think it would be easier to read if the link was just on `dart create`

src/null-safety/index.md
```diff
@@ -43,266 +49,6 @@ Dart null safety support is based on the following three core design principles:
   and its dependencies to null safety, you reap the full benefits of soundness
   —- not only fewer bugs, but smaller binaries and faster execution.
 
-## A tour of the null safety feature
```
mit-mit, line 46: I wonder if the page is a bit hard to consume for a new developer now that it has zero code examples on it? I think it's great to move the majority of this elsewhere, but would it make sense to include just a bit of code to exemplify what null safety is?

A concrete suggestion: Just before the `## Null safety principles` section, include just the piece from the current tour about declaring variables that are null and non-null (that is the first two code snippets in that section), and then end that with "For an example-driven summary...." changed to "For further examples...". WDYT?

src/null-safety/index.md
```diff
@@ -13,9 +13,9 @@ turn into **edit-time** analysis errors.
 
 
 You can
-[try null safety in your normal development environment](#enable-null-safety),
-[migrate your existing code][migration-guide] to use null safety,
-or you can practice using null safety in the web app
+[use null safety](#enable-null-safety) in your normal development environment,
+[migrate existing code][migration guide] to use null safety,
+or practice using null safety in the web app
```
mit-mit, line 18: Maybe change 'practice using' to something broader now, like just 'use'?

mit-mit (commented):
Looks great overall, just a few more comments

src/null-safety/index.md
kwalrath: Sure. I tend to think that links on code font are hard to see, but I do agree that it looks weird. I'll change it.

src/null-safety/index.md
kwalrath: I swapped it around a bit... ptal

mit-mit (approved):
Looks great!

PR #2997, opened 2021-02-25
Library tour text

### LINKED ISSUE #2992: Migrate library tour to null safety
https://dart.dev/guides/libraries/library-tour should have null safe code. I'm on it.

filiph:
Fixes #2992.

### DIFF
--- a/src/_guides/libraries/library-tour_migrated.md
+++ b/src/_guides/libraries/library-tour_migrated.md
@@ -1,0 +1,1689 @@
+---
+title: A tour of the core libraries
+description: Learn about the major features in Dart's libraries.
+short-title: Library tour
+---
+<?code-excerpt replace="/ *\/\/\s+ignore_for_file:[^\n]+\n//g; /\n? *\/\/\s+ignore:[^\n]+//g"?>
+<?code-excerpt plaster="none"?>
+
+This page shows you how to use the major features in Dart’s core libraries.
+It’s just an overview, and by no means comprehensive.
+Whenever you need more details about a class,
+consult the [Dart API reference.][Dart API]
+
+
+[dart:core](#dartcore---numbers-collections-strings-and-more)
+: Built-in types, collections, and other core functionality.
+  This library is automatically imported into every Dart program.
+
+[dart:async](#dartasync---asynchronous-programming)
+: Support for asynchronous programming, with classes such as Future and Stream.
+
+[dart:math](#dartmath---math-and-random)
+: Mathematical constants and functions, plus a random number generator.
+
+[dart:convert](#dartconvert---decoding-and-encoding-json-utf-8-and-more)
+: Encoders and decoders for converting between different data representations, including JSON and UTF-8.
+
+[dart:html](#darthtml)
+: DOM and other APIs for browser-based apps.
+
+[dart:io](#dartio)
+: I/O for programs that can use the Dart VM,
+  including Flutter apps, servers, and command-line scripts.
+
+This page is just an overview;
+it covers only a few dart:* libraries
+and no third-party libraries.
+
+Other places to find library information are the
+[pub.dev site]({{site.pub}}) and the
+[Dart web developer library guide][webdev libraries].
+You can find API documentation for all dart:* libraries in the
+[Dart API reference][Dart API] or, if you're using Flutter,
+the [Flutter API reference.][docs.flutter]
+
+{{site.alert.info}}
+  **DartPad tip:** You can play with the code in this page by copying it into a
+  [DartPad.]({{site.dartpad}})
+{{site.alert.end}}
+
+
+## dart:core - numbers, collections, strings, and more
+
+The dart:core library ([API reference][dart:core])
+provides a small but critical set of built-in functionality.
+This library is automatically imported into every Dart program.
+
+
+### Printing to the console
+
+The top-level `print()` method takes a single argument (any Object)
+and displays that object's string value (as returned by `toString()`)
+in the console.
+
+<?code-excerpt "misc/test/library_tour/core_test.dart (print)"?>
+```dart
+print(anObject);
+print('I drink $tea.');
+```
+
+For more information on basic strings and `toString()`, see
+[Strings](/guides/language/language-tour#strings) in the language tour.
+
+
+### Numbers
+
+The dart:core library defines the num, int, and double classes, which
+have some basic utilities for working with numbers.
+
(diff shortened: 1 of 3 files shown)


src/_guides/libraries/library-tour_migrated.md
```diff
@@ -443,12 +443,15 @@ assert(ingredients.length == 3);
 // Remove an item from a set.
 ingredients.remove('gold');
 assert(ingredients.length == 2);
+
+// You can also create sets using a literal.
```
kwalrath, line 447: Should we have the literal at the top?

Also... It seems weird to not have the generic type for the Set. Shouldn't we model specifying the type? This might be an issue in other code samples, too. A quick look finds the questionable looking `var teamAssignments = {};` and I'm sure there are more.

src/_guides/libraries/library-tour_migrated.md
```diff
@@ -1278,10 +1275,10 @@ Dart language features, see the
 To get each value as it arrives, either use `await for` or
 subscribe to the stream using the `listen()` method:
 
-<?code-excerpt "misc/lib/library_tour/async/stream_web.dart (listen)" replace="/listen/[!$&!]/g"?>
+<?code-excerpt "../null_safety_examples/misc/lib/library_tour/async/stream_web.dart (listen)" replace="/listen/[!$&!]/g"?>
 {% prettify dart tag=pre+code %}
 // Find a button by ID and add an event handler.
-querySelector('#submitInfo').onClick.[!listen!]((e) {
+querySelector('#submitInfo')!.onClick.[!listen!]((e) {
```
kwalrath, line 1281: I wonder if we should have a note about the `!`, since not all readers might be familiar with that syntax, and not all readers might be using null safety yet.

kwalrath (commented):
All the changes LGTM, but I wonder if we should be more aggressive about favoring literals and specifying generic types.

src/_guides/libraries/library-tour_migrated.md
filiph: I think here it makes more sense to start with the `Set` constructor:

* There is no empty set literal (unless you pair it with generics). `{}` means an empty Map.
* We're talking about the library, and its classes — not language features. If I could, I would start the `List` section with `List()` as well, but it's not possible.

As for the generic type: Yeah, I wasn't sure about that. The tour seems to take a pretty laissez-faire approach. I wonder if it's remnants from pre-Dart-2?

But, on the other hand, I like to avoid introducing too much stuff at once. Like, when we're looking at the library, and the learner isn't possibly all that familiar with generics, let's keep them away as much as possible.

In _this_ particular case, the generic for `{79, 22, 54}` is inferred. So I wouldn't write in the generic even in production code.

src/_guides/libraries/library-tour_migrated.md
filiph: In that case, wouldn't it be better to extract the `querySelector` and have something like:

```suggestion
submitButton.onClick.[!listen!]((e) {
```

I tried to imagine how I'd address the `!`, and it seems weird. I'd spend more time explaining the `!` than the topic at hand.

src/_guides/libraries/library-tour_migrated.md
kwalrath: > As for the generic type: Yeah, I wasn't sure about that. The tour seems to take a pretty laissez-faire approach. I wonder if it's remnants from pre-Dart-2?


Almost certainly, yes.
 
> But, on the other hand, I like to avoid introducing too much stuff at once. Like, when we're looking at the library, and the learner isn't possibly all that familiar with generics, let's keep them away as much as possible.

I agree with being as simple as possible, but I also don't want to model bad behavior. We could explain the first time we need to use generics, and/or we could link aware to a friendly place where generics are talked about. (https://dart.dev/guides/language/language-tour#generics maybe?)

src/_guides/libraries/library-tour_migrated.md
filiph: Resolved with latest commit

src/_guides/libraries/library-tour_migrated.md
filiph: Ok, resolved with latest commit

src/_guides/libraries/library-tour_migrated.md
```diff
@@ -0,0 +1,1707 @@
...
+<?code-excerpt "../null_safety_examples/misc/test/library_tour/core_test.dart (compareTo)"?>
+```dart
+var fruits = ['bananas', 'apples', 'oranges'];
+
+// Sort a list.
+fruits.sort((a, b) => a.compareTo(b));
+assert(fruits[0] == 'apples');
+```
+
+Lists are parameterized types ([generics]({{site.www}}/guides/language/language-tour#generics)), so you can specify the type that a list
```
kwalrath, line 407: Fixing the line length and the link. (We no longer use site.www and should remove/replace it. I've filed #3003 to track this.)

```suggestion
Lists are parameterized types
([generics](/guides/language/language-tour#generics)),
so you can specify the type that a list
```

src/_guides/libraries/library-tour_migrated.md
```diff
@@ -0,0 +1,1707 @@
...
+<?code-excerpt "../null_safety_examples/misc/lib/library_tour/core/collections.dart (List-of-String)"?>
+```dart
+fruits.add(5); // Error: 'int' can't be assigned to 'String'
+```
+
+{{site.alert.note}}
+  In many cases, you don't
+  need to explicitly specify generic
+  types, because Dart will
+  [infer]({{site.www}}/guides/language/type-system#type-inference)
```
kwalrath, line 430:
```suggestion
  [infer](/guides/language/type-system#type-inference)
```

src/_guides/libraries/library-tour_migrated.md
```diff
@@ -0,0 +1,1707 @@
...
+{% prettify dart tag=pre+code %}
+// Add an event handler to a button.
+submitButton.onClick.[!listen!]((e) {
+  // When the button is clicked, it runs this code.
+  submitData();
+});
+{% endprettify %}
+
+In this example, the `onClick` property is a Stream object provided by
+the "submitInfo" button.
```
kwalrath, line 1309:
```suggestion
In this example, the `onClick` property is a `Stream` object provided by
the submit button.
```

kwalrath (approved):
LGTM! Needs just a few tweaks, and then we're good.

PR #3000, opened 2021-02-26
Document abstract fields in "Understanding Null Safety".

### LINKED ISSUE #2695: Document abstract fields
Details: dart-lang/sdk#43712

munificent:
I tried
using semantic line breaks.
It makes my technical prose
read like some weird form
of nerd poetry
(which I sort of like).

Fix #2695.

### DIFF
--- a/src/null-safety/understanding-null-safety/index.md
+++ b/src/null-safety/understanding-null-safety/index.md
@@ -1157,7 +1157,7 @@ It sees that `contents` declaration as a real field.
 And, unfortunately, that field is non-nullable and has no initializer,
 so you get a compile error.
 
-One fix is to use the explicit abstract getter/setter declarations
+One fix is to use explicit abstract getter/setter declarations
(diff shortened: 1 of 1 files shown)


src/null-safety/understanding-null-safety/index.md
```diff
@@ -1122,6 +1122,56 @@ non-nullable types (if they have a default value).
...
+  set contents(Beverage);
+}
+```
+
+But Dart doesn't *know* that this class will never be used as a concrete type.
+It sees that `contents` declaration as a real field.
+And, unfortunately, that field is non-nullable and has no initializer,
+so you get a compile error.
+
+One fix is to use the explicit abstract getter/setter declarations
```
kwalrath, line 1160: ?

```suggestion
One fix is to use explicit abstract getter/setter declarations
```

kwalrath (approved):
Nice! And poetical!

PR #3005, opened 2021-02-26
Update null default value discussion

kwalrath:
Staged: https://kw-www-dartlang-1.firebaseapp.com/guides/language/language-tour_migrated#default-value

### DIFF
--- a/src/_guides/language/language-tour_migrated.md
+++ b/src/_guides/language/language-tour_migrated.md
@@ -324,15 +324,10 @@ String name = 'Bob';
 
 ### Default value
 
-Uninitialized variables have an initial value of `null`. Even variables
-with numeric types are initially null, because numbers—like everything
-else in Dart—are objects.
-
-{% comment %}
-TODO #2950: Cover `?` earlier.
-21 out of 958 excerpts were updated in the migration of examples to null safety.
-This is the first one, and the first occurrence of `?`.
-{% endcomment %}
+Uninitialized variables have an initial value of `null`,
+whether or not you've enabled [null safety][ns].
+Even variables with numeric types are initially null,
+because numbers—like everything else in Dart—are objects.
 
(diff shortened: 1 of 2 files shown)


parlough (approved):

src/_guides/language/language-tour_migrated.md
```diff
@@ -341,15 +341,10 @@ String name = 'Bob';
...
-with numeric types are initially null, because numbers—like everything
-else in Dart—are objects.
-
-{% comment %}
-TODO #2950: Cover `?` earlier.
-21 out of 958 excerpts were updated in the migration of examples to null safety.
-This is the first one, and the first occurrence of `?`.
-{% endcomment %}
+Uninitialized variables have an initial value of `null`,
+whether or not you've enabled [null safety][ns].
```
munificent, line 345: I think this is a little misleading. Nullable variables have a default initial value of `null` and the language is happy to let you omit the initializer and use the variable because of that.

Non-nullable (and potentially nullable) variables do not have any initial value at all. The language absolutely prevents you from ever extracting a value from one of those variables that you yourself did not store there first. It could be default initializing them with the full text of the Battle Hymn of the Republic for all a user knows, because they'll never be able to see it.

This is somewhat important because saying that all variables are default-null initialized might imply to a reader that the compiler has to treat all variable storage locations as supporting null. That has negative perf implications which aren't actually true. Null safety lets the compiler store many non-nullable variables unboxed specifically *because* they are never initialized to null or accessible without being initialized.

munificent (approved):
One thing but otherwise LGTM!

src/_guides/language/language-tour_migrated.md
kwalrath: How about something like this:

> Uninitialized variables that have a nullable type have an initial value of `null`. (If you haven't opted into  null safety, then every variable has a nullable type.)

PR #3008, opened 2021-02-26
Add late variable section to language tour

kwalrath:
Staged: https://kw-staging-dartlang-2.firebaseapp.com/guides/language/language-tour_migrated#late-variables

### DIFF
--- a/src/_guides/language/language-tour_migrated.md
+++ b/src/_guides/language/language-tour_migrated.md
@@ -402,6 +402,7 @@ then the initializer runs the first time the variable is used.
 This lazy initialization is handy in a couple of cases:
 
 * The variable might not be needed,
+  and initializing it is costly.
 * You're initializing an instance variable,
   and its initializer needs access to `this`.
 
--- a/src/_guides/language/language-tour_migrated.md
+++ b/src/_guides/language/language-tour_migrated.md
@@ -368,13 +368,13 @@ assert(lineCount == null);
 
 Dart 2.12 added the `late` modifier, which has two use cases:
 
-* To tell Dart that a non-nullable variable is set before it's used.
-* To lazily initialize a variable.
+* Declaring a non-nullable variable that's initialized after its declaration.
+* Lazily initializing a variable.
 
 Often Dart's control flow analysis can detect when a non-nullable variable
 is set to a non-null value before it's used,
 but sometimes analysis fails.
-One common case is top-level variables:
+Two common cases are top-level variables and instance variables:
 Dart often can't determine whether they're set,
 so it doesn't try.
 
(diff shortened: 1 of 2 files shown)


src/_guides/language/language-tour_migrated.md
```diff
@@ -363,6 +363,51 @@ assert(lineCount == null);
   _condition_ is false. For details, see [Assert](#assert).
 {{site.alert.end}}
 
+
+### Late variables
+
+Dart 2.12 added the `late` modifier, which has two use cases:
+
+* To tell Dart that a non-nullable variable is set before it's used.
```
munificent, line 371: "Is" might lead a reader astray, when it's really something more like "should be". If they forget to actually set it, the language won't stop them and will instead blow up at runtime.

Maybe something like: "To allow you to declare a non-nullable variable that is initialized after its declaration."

src/_guides/language/language-tour_migrated.md
```diff
@@ -363,6 +363,51 @@ assert(lineCount == null);
...
+* To lazily initialize a variable.
+
+Often Dart's control flow analysis can detect when a non-nullable variable
+is set to a non-null value before it's used,
+but sometimes analysis fails.
+One common case is top-level variables:
```
munificent, line 377: Fields are probably the more common case.

src/_guides/language/language-tour_migrated.md
```diff
@@ -363,6 +363,51 @@ assert(lineCount == null);
...
+### Late variables
+
+Dart 2.12 added the `late` modifier, which has two use cases:
+
+* To tell Dart that a non-nullable variable is set before it's used.
+* To lazily initialize a variable.
+
+Often Dart's control flow analysis can detect when a non-nullable variable
+is set to a non-null value before it's used,
+but sometimes analysis fails.
```
munificent, line 376: This is generally true. But there's a fairly big split between how local variables are analyzed and how all other variables are. Even with locals there are edge cases where the flow analysis fails. But for fields and top-level variables, there is no flow analysis at all. If you don't initialize it at the declaration (or initializer list for instance fields), it's an error.

src/_guides/language/language-tour_migrated.md
```diff
@@ -363,6 +363,51 @@ assert(lineCount == null);
...
+
+<?code-excerpt "../null_safety_examples/misc/lib/language_tour/variables.dart (var-late-top-level)" replace="/late/[!$&!]/g"?>
+```dart
+[!late!] String description;
+
+void main() {
+  description = 'Feijoada!';
+  print(description);
+}
+```
```
munificent, line 393: Do you think it's worth explaining that if you fail to initialize it, you'll get a runtime error when you try to use the variable?

src/_guides/language/language-tour_migrated.md
```diff
@@ -363,6 +363,51 @@ assert(lineCount == null);
...
+
+Use `late` for lazy initialization when a variable might not be needed,
+and initializing it is costly.
+When you mark a variable as `late` but initialize it at its declaration,
+then the initializer doesn't run until the first time the variable is used.
```
munificent, line 398: A nice benefit of lazy initialization is that late instance fields do have access to `this`, unlike other instance fields. It might be worth spelling that out, since it's not an obvious capability, but is pretty useful.

src/_guides/language/language-tour_migrated.md
kwalrath: How about:

* Declaring a non-nullable variable that's initialized after its declaration.
* Lazily initializing a variable.

kwalrath:
ptal @munificent

munificent (approved):

PR #3010, opened 2021-02-26
Revise import guidelines.

### LINKED ISSUE #2496: Recommend relative paths for inside-package imports
See https://github.com/flutter/samples/pull/463#discussion_r439219534 for details.

### LINKED ISSUE #2606: Effective Dart - which import style is preferred for package's own imports?
Need to clarify which import style is preferred for package's own imports using simple rules to follow.

Currently, the [docs](https://dart.dev/guides/language/effective-dart/usage#do-use-relative-paths-when-importing-libraries-within-your-own-packages-lib-directory) says:

> When referencing a library inside your package’s lib directory from another library in that same package, use a relative URI, not an explicit `package:` URI.

And then:

> Libraries inside `lib` can import other libraries inside `lib` (or in subdirectories of it). Libraries outside of `lib` can use relative imports to reach other libraries outside of `lib`.

And there is ["Creating packages" guide](https://dart.dev/guides/libraries/create-library-packages#importing-library-files):

> When importing a library file from your own package, use a relative path when both files are inside of lib, or when both files are outside of lib. Use `package:` when the imported file is in `lib` and the importer is outside.

It still is not clear which import to use. What does the "inside" and "outside" mean? Is `lib/src/foo.dart` "inside" or "outside" `lib`? Following this statement, I should import `lib/src/foo.dart` ("inside") from within `test/src/foo_test.dart` ("outside") using `package:` import, which is discouraging, as `lib/src/foo.dart` is considered private and shouldn't be imported using `package:` import. Examples don't help either.

It would be better to provide clear rules to follow for, e. g. such cases:

- import a library implementation from a library using relative import;
- import a library from test using `package:` import;
- import a library implementation from test using (I don't know) import.

Terminology:

- a *library* is a Dart source directly under `lib` or sub-directories, but not in `lib/src`;
- a *library implementation* is a Dart source directly under `lib/src` or sub-directories.

Personally I found docs very convoluted, scattered and hard to grasp as a whole. I'm compelled to keep three to five browser tabs open with different Dart guides in order to understand single topic.

munificent:
Make it clearer that you should prefer relative imports even outside of
"lib".

Fix #2496. Fix #2606.

### DIFF
--- a/src/_guides/language/effective-dart/usage_migrated.md
+++ b/src/_guides/language/effective-dart/usage_migrated.md
@@ -73,53 +73,98 @@ change to the package.
 That means that if you import some other package's private library, a minor,
 theoretically non-breaking point release of that package could break your code.
 
-### DO use relative paths when importing libraries within your own package's `lib` directory.
 
-{% include linter-rule.html rule1="avoid_relative_lib_imports" rule2="prefer_relative_imports" %}
+### DON'T allow an import path to reach into or out of `lib`.
 
-When referencing a library inside your package's `lib` directory from another
-library in that same package, use a relative URI, not an explicit `package:`
-URI.
+{% include linter-rule.html rule="avoid_relative_lib_imports" %}
+
+Dart supports `package:` imports to let you access
+the libraries inside a package's `lib` directory
+regardless of whether the package is physically stored on your computer.
+This can lead to a very confusing situation
+if you have a `package:` import of a library and another import
+that physically traverses into that same package's `lib` directory.
 
 For example, say your directory structure looks like:
 
 ```text
 my_package
 └─ lib
-   ├─ src
-   │  └─ utils.dart
    └─ api.dart
+   test
+   └─ api_test.dart
 ```
 
-If `api.dart` wants to import `utils.dart`, it should do so using:
-
-{:.good}
-{% prettify dart tag=pre+code %}
-import 'src/utils.dart';
-{% endprettify %}
-
-And not:
+If `api_test.dart` contains:
 
 {:.bad}
 {% prettify dart tag=pre+code %}
-import 'package:my_package/src/utils.dart';
+import 'package:my_package/api.dart';
+import '../lib/api.dart';
 {% endprettify %}
 
-The "within your own package's `lib` directory" part is important. Libraries
-inside `lib` can import other libraries inside `lib` (or in subdirectories of
-it). Libraries outside of `lib` can use relative imports to reach other
-libraries outside of `lib`. For example, you may have a test utility library
-under `test` that other libraries in `test` import.
-
-But you can't "cross the streams". A library outside of `lib` should never use a
-relative import to reach a library under `lib`, or vice versa. Doing so will
-break Dart's ability to correctly tell if two library URIs refer to the same
-library, which can lead to unexpected duplicated types.
-
-Follow these two rules:
+Then Dart thinks those are imports of two completely unrelated libraries.
+This causes all sorts of errors and confusion.
+To avoid that, follow these two rules:
 
 * An import path should never contain `/lib/`.
 * A library under `lib` should never use `../` to escape the `lib` directory.
+
+Instead, when you need to reach into a package's `lib` directory
+(even from the same package's `test` or other top-level directories),
+use a `package:` import.
+A package should never reach *out* of its `lib` directory and
+import libraries from other places in the package.
+
+
+### PREFER relative import paths.
+
+{% include linter-rule.html rule="prefer_relative_imports" %}
+
+Whenever the previous rule doesn't come into play, follow this one.
+When an import does *not* reach across `lib`, prefer using relative imports.
+They're shorter.
+For example, say your directory structure looks like this:
+
+```text
+my_package
+└─ lib
+   ├─ src
+   │  └─ stuff.dart
+   │  └─ utils.dart
+   └─ api.dart
+   test
+   │─ api_test.dart
+   └─ test_utils.dart
+```
+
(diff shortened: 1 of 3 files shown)


src/_guides/language/effective-dart/usage_migrated.md
```diff
@@ -73,53 +73,98 @@ change to the package.
...
-library in that same package, use a relative URI, not an explicit `package:`
-URI.
+{% include linter-rule.html rule="avoid_relative_lib_imports" %}
+
+Dart supports `package:` imports to let you access
+the libraries inside a package's `lib` directory
+regardless of whether the package is physically stored on your computer.
+This can lead to a very confusing situation
+if you have a `package:` import of a library and another import
+that physically traverses into that same package's `lib` directory.
```
kwalrath, line 86: I had trouble understanding this intro, even though I think I know what you're getting at. It seems a little too high level. Maybe:

```suggestion
If you have a `package:` import of a library and another import
that physically traverses into that same package's `lib` directory,
the result can be all sorts of errors and confusion.
```

src/_guides/language/effective-dart/usage_migrated.md
```diff
@@ -73,53 +73,98 @@ change to the package.
...
 
 For example, say your directory structure looks like:
```
kwalrath, line 88:
```suggestion
For example, say your directory structure looks like this:
```

src/_guides/language/effective-dart/usage_migrated.md
```diff
@@ -73,53 +73,98 @@ change to the package.
...
 
+```text
+my_package
+└─ lib
+   └─ api.dart
+   test
+   └─ api_test.dart
+```
+
+If `api_test.dart` contains:
```
kwalrath, line 98:
```suggestion
Say `api_test.dart` imports `api.dart` in two ways:
```

src/_guides/language/effective-dart/usage_migrated.md
```diff
@@ -73,53 +73,98 @@ change to the package.
...
+
+{:.bad}
+{% prettify dart tag=pre+code %}
+import 'package:my_package/api.dart';
+import '../lib/api.dart';
+{% endprettify %}
+
+Then Dart thinks those are imports of two completely unrelated libraries.
+This causes all sorts of errors and confusion.
```
kwalrath, line 107:
```suggestion
Dart thinks those are imports of two completely unrelated libraries.
```

src/_guides/language/effective-dart/usage_migrated.md
```diff
@@ -73,53 +73,98 @@ change to the package.
...
+To avoid that, follow these two rules:
```
kwalrath, line 108:
```suggestion
To avoid confusing Dart and yourself, follow these two rules:
```

src/_guides/language/effective-dart/usage_migrated.md
```diff
@@ -73,53 +73,98 @@ change to the package.
...
+
+* An import path should never contain `/lib/`.
+* A library under `lib` should never use `../` to escape the `lib` directory.
```
kwalrath, line 111:
```suggestion
* Don't use `/lib/` in an import path.
* Don't use `../` to escape the `lib` directory.
```

src/_guides/language/effective-dart/usage_migrated.md
```diff
@@ -73,53 +73,98 @@ change to the package.
...
+
+Instead, when you need to reach into a package's `lib` directory
+(even from the same package's `test` or other top-level directories),
```
kwalrath, line 114:
```suggestion
(even from the same package's `test` directory
or any other top-level directory),
```

kwalrath (commented):
I had a little trouble understanding the new rule, so I have some suggestions for making it more direct. (Your changes to the existing rule look great.)

src/_guides/language/effective-dart/usage_migrated.md
kwalrath: "physically traverses" could probably be simpler, now that I look at this. "physically" isn't really accurate, and "traverses" is not a word I'd expect an ESL reader to know.

Change it to "crosses"?

munificent: Yeah, this is hard to explain concisely. I'll take another stab at it.

src/_guides/language/effective-dart/usage_migrated.md
```diff
@@ -78,12 +78,16 @@ theoretically non-breaking point release of that package could break your code.
...
-the libraries inside a package's `lib` directory
-regardless of whether the package is physically stored on your computer.
-This can lead to a very confusing situation
-if you have a `package:` import of a library and another import
-that physically traverses into that same package's `lib` directory.
+A `package:` import lets you access
+a library inside a package's `lib` directory
+without having to worry about where the package is stored on your computer.
+For this to work, you cannot have imports that require the `lib`
+to be in some location on disc relative to other files.
```
kwalrath, line 85:
```suggestion
to be in some location on disk relative to other files.
```

kwalrath (approved):
Much clearer! Just one small change (to use the 'merican/CS spelling of disk).

munificent:
> Much clearer! Just one small change (to use the 'merican/CS spelling of disk).

One day I will remember and get this right.

kwalrath:
> > Much clearer! Just one small change (to use the 'merican/CS spelling of disk).
> 
> One day I will remember and get this right.

It doesn't help that the sound recording kind of disc is spelled with a 'c'.

PR #3012, opened 2021-02-28
Beefed up built-in types section

### LINKED ISSUE #2485: Document the new `Never` type
As part of the wider null safety changes, Dart 2.9 has a new type, `Never`. This can be used as a return type:

  *  `void foo() {}`: A function which usually returns normally, but doesn't return a (meaningful) result

  *   `Never foo() {}`: A function which returns abnormally (e.g. throws or runs an infinite loop), e.g. the `exit()` function.

kwalrath:
* Added coverage of `Null` and `Never` (contributing to #2485)
* Code fonted more class names

Staged: https://dartlang-org-staging-0.firebaseapp.com/guides/language/language-tour_migrated#built-in-types

### DIFF
--- a/src/_guides/language/language-tour_migrated.md
+++ b/src/_guides/language/language-tour_migrated.md
@@ -62,10 +62,7 @@ apps:
 :   A special type that indicates a value that's never used.
     Functions like `printInteger()` and `main()` that don't explicitly return a value
     have the `void` return type.
-    For more information, see [this article][void article].
     
-[void article]: https://medium.com/dartlang/dart-2-legacy-of-the-void-e7afb5f44df0
-
 `int`
 :   Another type, indicating an integer.
     Some additional [built-in types](#built-in-types)
@@ -110,7 +107,7 @@ mind:
 -   Everything you can place in a variable is an *object*, and every
     object is an instance of a *class*. Even numbers, functions, and
     `null` are objects.
-    All objects except `null` inherit from the [Object][] class.
+    All objects except `null` inherit from the [`Object`][] class.
 
 -   Although Dart is strongly typed, type annotations are optional
     because Dart can infer types. In the code above, `number`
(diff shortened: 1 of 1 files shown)


src/_guides/language/language-tour_migrated.md
```diff
@@ -466,35 +452,63 @@ For more information on using `const` to create constant values, see
...
+[Understanding null safety][].
+
+{% comment %}
+If we decide to cover `dynamic` more,
+here's a nice example that illustrates what dynamic does:
+  dynamic a = 2;
+  String b = a; // No problem! Until runtime, when you get an uncaught error.
+
+  Object c = 2;
+  String d = c;  // Problem!
```
munificent, line 503: Another interesting example might be:

```dart
class Dog {
  void bark() { print('Ruff!'); }
}

class Tree {
  void bark() { print('Rough!'); }
}

main(List<String> arguments) {
  Object? thing = arguments.first == 'dog' ? Dog() : Tree();
  thing.bark(); // Error.

  dynamic dynamicThing = thing;
  thing.back(); // OK. Picks right method at runtime.
}
```

Whether code like this is a good thing is an open question. :)

src/_guides/language/language-tour_migrated.md
```diff
@@ -4111,8 +4129,8 @@ When you need to lazily produce a sequence of values,
 consider using a _generator function_.
 Dart has built-in support for two kinds of generator functions:
 
-* **Synchronous** generator: Returns an **[Iterable]** object.
-* **Asynchronous** generator: Returns a **[Stream]** object.
+* **Synchronous** generator: Returns an **[`Iterable`]** object.
+* **Asynchronous** generator: Returns a **[`Stream`]** object.
```
munificent, line 4133: Maybe delete the `**` bold since it probably doesn't play nice with the code font?

munificent (approved):

PR #3013, opened 2021-02-28
Fix assert in migrated list tour example

parlough:
I've updated the `vegetables` `assert` as it should no longer be empty but rather be filled with a lot of broccoli. I hope it's roasted - yum!

### DIFF
--- a/null_safety_examples/misc/test/library_tour/core_test.dart
+++ b/null_safety_examples/misc/test/library_tour/core_test.dart
@@ -239,8 +239,8 @@ void main() {
 
       // You can also create a List using one of the constructors.
       var vegetables = List.filled(99, 'broccoli');
+      assert(vegetables.every((v) => v == 'broccoli'));
       // #enddocregion List
-      assert(vegetables.isEmpty);
     });
 
     test('indexOf', () {
--- a/src/_guides/libraries/library-tour.md
+++ b/src/_guides/libraries/library-tour.md
@@ -378,6 +378,7 @@ assert(fruits.isEmpty);
 
 // You can also create a List using one of the constructors.
 var vegetables = List.filled(99, 'broccoli');
+assert(vegetables.every((v) => v == 'broccoli'));
 ```
 
 Use `indexOf()` to find the index of an object in a list:


kwalrath:
@filiph could you please review the changes to the migrated library tour?

@parlough thanks for the PR! I don't expect to take the changes to the old library tour, since it should be replaced by the migrated tour soon, but I really appreciate your thoughtfulness about how we should present these APIs.

examples/misc/test/library_tour/core_test.dart
```diff
@@ -212,10 +212,7 @@ void main() {
   group('Collections: List:', () {
     test('constructor', () {
       // #docregion List
-      // Use a List constructor.
-      var vegetables = List();
```
domesticmouse, line 216: I'm curious why we are deleting the empty list constructor. I'd argue for writing it as `<type>[]` instead tho.

src/_guides/libraries/library-tour.md
```diff
@@ -351,10 +351,7 @@ items to and removing items from lists.
 
 <?code-excerpt "misc/test/library_tour/core_test.dart (List)"?>
 ```dart
-// Use a List constructor.
-var vegetables = List();
```
domesticmouse, line 355: Same as previous, empty lists are useful.

domesticmouse (changes requested):

src/_guides/libraries/library-tour.md
parlough: With null safety the empty list constructor is now disallowed.

domesticmouse: Really? I doubt that. 

A pre-sized list of non optional type is disallowed, as it can’t be pre-filled with nulls. But an empty list of non optional sis still completely valid. And useful.

parlough: image

examples/misc/test/library_tour/core_test.dart
parlough: We could definitely add a new example using an empty list literal, I was just trying to match the old example with the new one. I definitely think that would be worth it and in a separate PR we can add a new empty example back.

parlough:
As we don't need to include the change in the old example and I've redefined this PR as simply updating the assert to match the new behavior.

null_safety_examples/misc/test/library_tour/core_test.dart
```diff
@@ -233,10 +233,10 @@ void main() {
       fruits.clear();
       assert(fruits.isEmpty);
 
-      // You can also create a List using one of the constructors.
+      // You can also create a List using one of the factory constructors.
```
filiph, line 236: I'm not 100% sure if it's needed to qualify the constructors as factory. It's [technically correct](https://i.kym-cdn.com/photos/images/original/000/909/991/48c.jpg), of course, but may imply the existence of non-factory constructors.

Here, we're talking about constructors, specifically, because, above, we've been using literals.

```suggestion
      // You can also create a List using one of the constructors.
```

null_safety_examples/misc/test/library_tour/core_test.dart
```diff
@@ -233,10 +233,10 @@ void main() {
...
       var vegetables = List.filled(99, 'broccoli');
+      assert(vegetables.every((v) => v == 'broccoli'));
```
filiph, line 238: Good one!

null_safety_examples/misc/test/library_tour/core_test.dart
```diff
@@ -233,10 +233,10 @@ void main() {
...
       // #enddocregion List
-      assert(vegetables.isEmpty);
```
filiph, line 239: I'm a bit worried now. How did this pass the tests? (Not a question for you, @parlough, I'm just asking the universe I guess.)

src/_guides/libraries/library-tour_migrated.md
```diff
@@ -372,8 +372,9 @@ assert(fruits.length == 4);
 fruits.clear();
 assert(fruits.isEmpty);
 
-// You can also create a List using one of the constructors.
+// You can also create a List using one of the factory constructors.
```
filiph, line 375:
```suggestion
// You can also create a List using one of the constructors.
```

filiph (approved):
LGTM, but I strongly suggest removing the "factory" qualifier in the comment.

kwalrath (approved):

parlough:
@filiph That makes sense and I suppose it adds unnecessary cruft, I've removed both `factory` qualifiers.

src/_guides/libraries/library-tour.md
filiph: @parlough is right. There is no `List()` anymore. Only `List.empty()`, and if you want to ever add to it, you have to do `List.empty(growable: true)`.

domesticmouse: There is also `<type>[]` which has been the idiomatic type constructor for a while.

parlough: I can create a follow up PR with an example of that.

domesticmouse: I stand corrected https://dart.dev/tools/diagnostic-messages#default_list_constructor

however, I still advocate for replacing the `List()` constructor with the List literal `[]` instead of deleting it.

kwalrath: List literals are covered in the language tour, and there's a link supposedly to that coverage in the text above... but it turns out to be a self-link!

kwalrath: I see @parlough has a PR to add back an empty list example: #3017

parlough:
I've reset this on top of the consolidated library tour, good to go on my end :)

kwalrath (approved):
I believe that Brett was OK with the final changes, so I'm going to merge this.

PR #3014, opened 2021-02-28
Minor corrections/improvements to language tour's doc comments section

parlough:
I'm not sure it's really the compiler that does the resolution in the doc comments but rather the analyzer and surrounding tools, so I updated `the Dart compiler` to `Dart program analysis` but if desired I think `the Dart analyzer` could work too, but I'm not sure if most users know what that is.

Also, instead of calling dartdoc the SDK's documentation tool I thought it was a bit more friendly and easy to understand to just call it `Dart's documentation tool`. 

I've also made `Food` an inline code block.

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -4538,7 +4538,7 @@ classes and arguments:
 class Llama {
   String? name;
 
-  /// Feeds your llama [Food].
+  /// Feeds your llama [food].
   ///
   /// The typical llama eats one bale of hay per week.
   void feed(Food food) {
@@ -4553,7 +4553,7 @@ class Llama {
 }
 ```
 
-In the generated documentation, `[Food]` becomes a link to the API docs
+In the generated documentation, `[food]` becomes a link to the API docs
 for the `Food` class.
 
 To parse Dart code and generate HTML documentation, you can use Dart's
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -4520,7 +4520,7 @@ Documentation comments are multi-line or single-line comments that begin
 with `///` or `/**`. Using `///` on consecutive lines has the same
 effect as a multi-line doc comment.
 
-Inside a documentation comment, the Dart compiler ignores all text
+Inside a documentation comment, Dart progam analysis ignores all text
 unless it is enclosed in brackets. Using brackets, you can refer to
 classes, methods, fields, top-level variables, functions, and
 parameters. The names in brackets are resolved in the lexical scope of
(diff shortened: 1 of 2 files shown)


kwalrath (commented):
LGTM, but I'd like Mr. Analyzer (@bwilkerson) to take a look.

src/_guides/language/language-tour.md
```diff
@@ -4520,7 +4520,7 @@ Documentation comments are multi-line or single-line comments that begin
 with `///` or `/**`. Using `///` on consecutive lines has the same
 effect as a multi-line doc comment.
 
-Inside a documentation comment, the Dart compiler ignores all text
+Inside a documentation comment, Dart progam analysis ignores all text
```
bwilkerson, line 4523: Specific to this change, I think the alternate that you suggested of using 'the analyzer' might read better than 'Dart program analysis'. I think enough users have an understanding of what the analyzer is that it won't be confusing. Either way it's better than 'compiler' because all of the compilers ignore doc comments.

@kwalrath I don't think I've ever looked at this part of the documentation before. I don't know whether it matters much to users, but I wanted to point out that the analyzer and dartdoc treat the text slightly differently. As stated on line 4526, the analyzer will only resolve names in brackets if they're in the lexical scope (that is, if there's an import for them) whereas dartdoc will do extra work and create links for things even if they haven't been imported. Analyzer will ignore identifiers in brackets if they aren't in the lexical scope.

Also, a minor nit, but on line 4541, it's idiomatic to use the name of a parameter rather than the type of the parameter when documenting a method, so consider changing `[Food]` to `[food]`. (Using the type is confusing when there are multiple parameters with the same type.)

bwilkerson (approved):

src/_guides/language/language-tour.md
parlough: Thanks for the suggestions @bwilkerson, I've gone ahead and switched the wording to `the analyzer` which I now noticed is used elsewhere in the page anyway. I've also updated the parameter references to `[food]`. 

As for the dartdoc vs analyzer lexical scope issue, I tried to modify the piece to provide more explanation but I couldn't find a great way to word it. If it is desired to provide more specifics about those differences, perhaps it would be better handled by @kwalrath as a commit to this branch or a separate PR.

src/_guides/language/language-tour.md
```diff
@@ -4553,11 +4553,11 @@ class Llama {
 }
 ```
 
-In the generated documentation, `[Food]` becomes a link to the API docs
-for the Food class.
+In the generated documentation, `[food]` becomes a link to the API docs
+for the `Food` class.
```
bwilkerson, line 4557: This is no longer true. The text `[food]` becomes a link to the parameter. The user can still navigate to the class from the type annotation on the parameter declaration, but not from the dartdoc reference.

parlough: @bwilkerson I tried to add some class documentation to cover the desired results, what do you think about these latest changes?

bwilkerson: Those changes look fine to me, but I'll leave it to @kwalrath to decide whether they add to the quality of the documentation.

src/_guides/language/language-tour.md
```diff
@@ -4553,11 +4556,12 @@ class Llama {
...
-In the generated documentation, `[Food]` becomes a link to the API docs
-for the Food class.
+In the class's generated documentation, `[feed]` becomes a link
+to the docs for the `feed` method,
+and `[Food]` becomes a link to the docs for the `Food` class.
 
-To parse Dart code and generate HTML documentation, you can use the SDK’s
-[documentation generation tool.](https://github.com/dart-lang/dartdoc#dartdoc)
+To parse Dart code and generate HTML documentation, you can use Dart's
+[documentation generation tool, dartdoc.](https://github.com/dart-lang/dartdoc#dartdoc)
```
kwalrath, line 4564:
```suggestion
[documentation generation tool, dartdoc.](/tools/dartdoc)
```

kwalrath (approved):
I'm going to tweak one link, but your changes look good to me. Thank you, @parlough!

PR #3015, opened 2021-02-28
Specify map generics to actually be equivalent to literal example

parlough:
If we want the the `Map` constructor example to be equivalent to the preceding literal example which we specify by saying ("You can create the same objects using a Map constructor:") we need to specify the generic type arguments or else they're assumed to be `dynamic`.


kwalrath (commented):
Good catch! I'm in the midst of working on migrating the language tour, so I can't accept this change right now, but I'll do so asap.

parlough:
I've reset this on top of the consolidated library tour, ready for review on my end :)

kwalrath (approved):

PR #3016, opened 2021-02-28
Add ! to "Important concepts" section

kwalrath:
Not staged (I'm running out of servers!), but it looks fine on my local server.

### DIFF
--- a/src/_guides/language/language-tour_migrated.md
+++ b/src/_guides/language/language-tour_migrated.md
@@ -110,22 +110,26 @@ mind:
 -   Everything you can place in a variable is an *object*, and every
     object is an instance of a *class*. Even numbers, functions, and
     `null` are objects.
-    All objects except `null` inherit from the [Object][] class.
+    With the exception of `null` (if you enable [sound null safety][ns]),
+    all objects inherit from the [Object][] class.
+
+    {{site.alert.version-note}}
+      [Null safety][ns] was introduced in Dart 2.12.
+    {{site.alert.end}}
 
 -   Although Dart is strongly typed, type annotations are optional
     because Dart can infer types. In the code above, `number`
     is inferred to be of type `int`.
 
--   If you enable [sound null safety][ns],
+-   If you enable [null safety][ns],
     variables can’t contain `null` unless you say they can.
     You can make a variable nullable by
     putting a question mark (`?`) at the end of its type.
     For example, a variable of type `int?` might be an integer,
     or it might be `null`.
-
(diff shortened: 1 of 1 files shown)


parlough:
Do we want to create/link to some more specific guidance on this? I know a lot of packages and development have a stance of no usage of `!` and use it more of a tool during in progress migration. The end goal is to avoid the situations where its needed, provide a default value, or have a check that enables type promotion if appropriate.

kwalrath:
> Do we want to create/link to some more specific guidance on this? I know a lot of packages and development have a stance of no usage of `!` and use it more of a tool during in progress migration. The end goal is to avoid the situations where its needed, provide a default value, or have a check that enables type promotion if appropriate.

We don't have much guidance on using `!`. The only thing I see in the migrated Effective Dart is [CONSIDER copying a nullable field to a local variable to enable type promotion](https://dart.dev/guides/language/effective-dart/usage_migrated#consider-copying-a-nullable-field-to-a-local-variable-to-enable-type-promotion), which is super targeted. There's also a mild warning in [Understanding null safety](https://dart.dev/null-safety/understanding-null-safety), which I could perhaps replicate:

> Of course, like any cast, using ! comes with a loss of static safety. The cast must be checked at runtime to preserve soundness and it may fail and throw an exception. But you have control over where these casts are inserted, and you can always see them by looking through your code.

/cc @munificent

src/_guides/language/language-tour_migrated.md
```diff
@@ -110,22 +110,26 @@ mind:
...
     variables can’t contain `null` unless you say they can.
     You can make a variable nullable by
     putting a question mark (`?`) at the end of its type.
     For example, a variable of type `int?` might be an integer,
     or it might be `null`.
-
-    {{site.alert.version-note}}
-      [Null safety][ns] was introduced in Dart 2.12.
-    {{site.alert.end}}
+    If you _know_ an expression is non-null but Dart disagrees,
```
munificent, line 130: I try to use "non-nullable" and "nullable" to refer to types but not values. Maybe "If you know an expression will not evaluate to null..."

src/_guides/language/language-tour_migrated.md
```diff
@@ -110,22 +110,26 @@ mind:
...
+    you can add `!` to make it be treated as non-null.
```
munificent, line 131: "be treated as" is vague about what happens if it does evaluate to null. Maybe: "you can add `!` to assert that it won't be null". And then clarify that `!` throws an exception otherwise?

PR #3017, opened 2021-02-28
Add back empty list example to migrated library tour

parlough:
/cc @domesticmouse @filiph


domesticmouse (approved):

domesticmouse:
Thanks for the quick fix @parlough!

PR #3020, opened 2021-03-02
Update guidance around using collections.

### LINKED ISSUE #2872: Effective Dart: Update with collection-for/if/spread operators
The rule [CONSIDER using higher-order methods to transform a sequence](https://dart.dev/guides/language/effective-dart/usage#consider-using-higher-order-methods-to-transform-a-sequence) should probably be mention that collection-for/if/spread are fine too (or even preferred in some cases?)

munificent:
- Remove incorrect sentence about deprecate List constructor.

- Encourage using new spread, if, and for.

- Remove rule about higher-order methods. Most Dart users seem to know
  about them now, and we don't necessarily want to stress them over the
  new spread, if, and for.

Fix #2872.


kwalrath (approved):
Nice!

PR #3021, opened 2021-03-02
Add guideline about faking overloading.

### LINKED ISSUE #1205: Style: Prefer overload-style methods to dynamic calls or arguments
Based on work/analysis with @srawlins fixing issues across google3. This might be better suited for _multiple_ style rules instead of a single one, but I wanted to file a single issue for posterity and start there.

**BAD**: Using dynamic calls to perform intersection-type-like calls.

This causes poor code to be emitted on both DDC and Dart2JS (probably Dart AOT too), as well as drops all useful type information (users can pass `bool` to `checkLength`, for example, and get no static checking).

```dart
void checkLength(dynamic /*String | List*/ element) {
  if (element.length > 5) {
    print('Too big!');
  }
}
```

**GOOD**: Use overload-like methods.

Sure, it is strictly speaking more typing, but gives you (a) monomorphic functions and (b) static type checking. In a Dart 2.0 (AOT) world, both of these make infinitely more sense than the above option.

```dart
void checkList(List<void> elements) {
  if (element.length > 5) {
    print('Too big!');
  }
}

void checkString(String text) {
  if (text.length > 5) {
    print('Too big!');
  }
}
```

-----

**BAD**: Using `dynamic` to mean "unknown" type.

Using the `dynamic` keyword encourages accidental dynamic calls, which are slower on some runtimes, as well as omit any type checking, automated refactoring, and so on. 

```dart
void checkAny(dynamic /* String | bool */ stringOrBool) {
  // ...
}
```

**GOOD**: It is preferred to _Use overload-like methods_ where possible (see above), but if not possible, use `Object` and the `is` operator in order to emit only statically checked calls:

```dart
void checkAny(Object /* String | bool */ stringOrBool) {
  if (stringOrBool is String) {
    // ...
  } else if (stringOrBool is bool) {
    // ...
  } else {
    throw new ArgumentError('stringOrBool');
  }
}
```

munificent:
Fix #1205.

### DIFF
--- a/src/_guides/language/effective-dart/design_migrated.md
+++ b/src/_guides/language/effective-dart/design_migrated.md
@@ -947,12 +947,13 @@ not intended to be invoked from Dart code and don't need a corresponding getter.
 
 It's common for an API to support similar operations
 on different types of parameters.
-To emphasize the similarity, some languages support *overloading*
+To emphasize the similarity, some languages support *overloading*,
 which lets you define multiple methods
 that have the same name but different parameter lists.
 At compile time, the compiler looks at the actual argument types to determine
 which method to call.
-Dart does not have overloading.
+
+Dart doesn't have overloading.
 You can define an API that looks like overloading
 by defining a single method and then using `is` type tests
 inside the body to look at the runtime types of the arguments and perform the
@@ -963,16 +964,14 @@ into a choice that happens at *runtime*.
 If callers usually know which type they have
 and which specific operation they want,
 it's better to define separate methods with different names
-to let them select the right one directly.
+to let callers select the right operation.
 This gives better static type checking and faster performance
 since it avoids any runtime type tests.
-However, if users may have an object of an unknown type
-and *want* the API to interally use `is` to pick the right operation,
-then a single method with whose parameter is a supertype
+
+However, if users might have an object of an unknown type
+and *want* the API to internally use `is` to pick the right operation,
+then a single method where the parameter is a supertype
 of all of the supported types might be reasonable.
-In other words, if they would have wanted the API to use runtime type tests
-even if Dart *did* have overloading, then that's a good place
-to use a single method.
 
 
 ### AVOID public `late final` fields without initializers.
--- a/src/_guides/language/effective-dart/design_migrated.md
+++ b/src/_guides/language/effective-dart/design_migrated.md
@@ -943,6 +943,38 @@ not intended to be invoked from Dart code and don't need a corresponding getter.
 [angular]: {{site.angulardart}}
 
 
+### AVOID using runtime type tests to fake overloading.
+
+It's common for an API to support similar operations
+on different types of parameters.
+To emphasize the similarity, some languages support *overloading*
+which lets you define multiple methods
+that have the same name but different parameter lists.
+At compile time, the compiler looks at the actual argument types to determine
+which method to call.
+Dart does not have overloading.
+You can define an API that looks like overloading
+by defining a single method and then using `is` type tests
+inside the body to look at the runtime types of the arguments and perform the
+appropriate behavior.
+However, faking overloading this way turns a *compile time* method selection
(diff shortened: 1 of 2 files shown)


kwalrath (commented):
We'll need to update the TOC, too

src/_guides/language/effective-dart/design_migrated.md
```diff
@@ -943,6 +943,38 @@ not intended to be invoked from Dart code and don't need a corresponding getter.
 [angular]: {{site.angulardart}}
 
 
+### AVOID using runtime type tests to fake overloading.
+
+It's common for an API to support similar operations
+on different types of parameters.
+To emphasize the similarity, some languages support *overloading*
```
kwalrath, line 950:
```suggestion
To emphasize the similarity, some languages support *overloading*,
```

src/_guides/language/effective-dart/design_migrated.md
```diff
@@ -943,6 +943,38 @@ not intended to be invoked from Dart code and don't need a corresponding getter.
...
+which lets you define multiple methods
+that have the same name but different parameter lists.
+At compile time, the compiler looks at the actual argument types to determine
+which method to call.
+Dart does not have overloading.
```
kwalrath, line 955:
```suggestion

Dart doesn't have overloading.
```

src/_guides/language/effective-dart/design_migrated.md
```diff
@@ -943,6 +943,38 @@ not intended to be invoked from Dart code and don't need a corresponding getter.
...
+by defining a single method and then using `is` type tests
+inside the body to look at the runtime types of the arguments and perform the
+appropriate behavior.
+However, faking overloading this way turns a *compile time* method selection
+into a choice that happens at *runtime*.
+
+If callers usually know which type they have
+and which specific operation they want,
+it's better to define separate methods with different names
+to let them select the right one directly.
```
kwalrath, line 966:
```suggestion
to let callers select the right operation.
```

src/_guides/language/effective-dart/design_migrated.md
```diff
@@ -943,6 +943,38 @@ not intended to be invoked from Dart code and don't need a corresponding getter.
...
+This gives better static type checking and faster performance
+since it avoids any runtime type tests.
+However, if users may have an object of an unknown type
```
kwalrath, line 969:
```suggestion

However, if users might have an object of an unknown type
```

src/_guides/language/effective-dart/design_migrated.md
```diff
@@ -943,6 +943,38 @@ not intended to be invoked from Dart code and don't need a corresponding getter.
...
+and *want* the API to interally use `is` to pick the right operation,
```
kwalrath, line 970:
```suggestion
and *want* the API to internally use `is` to pick the right operation,
```

src/_guides/language/effective-dart/design_migrated.md
```diff
@@ -943,6 +943,38 @@ not intended to be invoked from Dart code and don't need a corresponding getter.
...
+then a single method with whose parameter is a supertype
```
kwalrath, line 971:
```suggestion
then a single method where the parameter is a supertype
```

src/_guides/language/effective-dart/design_migrated.md
```diff
@@ -943,6 +943,38 @@ not intended to be invoked from Dart code and don't need a corresponding getter.
...
+of all of the supported types might be reasonable.
+In other words, if they would have wanted the API to use runtime type tests
+even if Dart *did* have overloading, then that's a good place
+to use a single method.
```
kwalrath, line 975: This last sentence is pretty complex. I started trying to rewrite it, but then I started wondering whether the sentence was even necessary.

```suggestion
```

kwalrath (approved):

PR #3023, opened 2021-03-03
Add guideline around late variables and boolean flags.

### LINKED ISSUE #2633: Document when, if ever, is it better to use a bool to check late initialization
I'm writing code where I will lazily initialize a cache field on my class.

I could make the variable `Foo? _cachedFoo;` or `late Foo _cachedFoo;` with a bool to track whether I've set it at least once.

Reading https://dart.dev/null-safety#creating-variables, it's not entirely clear to me whether either of these is preferable, but it seems like one should be. It'd be helpful if that documentation explained this kind of case, and how to make a good decision about which pattern to prefer.

/cc @leafpetersen @munificent - sorry if this is the wrong repo

munificent:
Fix #2633.

### DIFF
--- a/src/_guides/language/effective-dart/usage_migrated.md
+++ b/src/_guides/language/effective-dart/usage_migrated.md
@@ -321,6 +321,27 @@ int measureMessage(String? message) {
 {% endprettify %}
 
 
+### AVOID `late` variables if you need to check whether they are initialized.
+
+Dart offers no way to tell if a `late` variable
+has been initialized or assigned to.
+If you access it, it either immediately runs the initializer
+if it has one or throws an exception.
+Sometimes you have some state that is lazily initialized
+where `late` would be a good fit,
+but you also need to be able to *tell* if the initialization has happened yet.
+
+You could do that by storing the state in a `late` variable
+and then have a separate boolean field
+to track whether the variable has been set.
+This works, but is redundant since Dart *internally*
+still needs to maintain the initialized status of the `late` variable.
+Instead, it's usually clearer to make the variable non-`late` and nullable.
+Then you can check the variable for `null` to see if it has been initialized.
+(Of course, if `null` is a valid initialized value for the variable,
+then it does make sense to have a separate boolean field.)
+
+
 ### CONSIDER copying a nullable field to a local variable to enable type promotion.
 
 Checking that a nullable variable is not equal to `null` promotes the variable
--- a/src/_guides/language/effective-dart/usage_migrated.md
+++ b/src/_guides/language/effective-dart/usage_migrated.md
@@ -326,20 +326,22 @@ int measureMessage(String? message) {
 Dart offers no way to tell if a `late` variable
 has been initialized or assigned to.
 If you access it, it either immediately runs the initializer
-if it has one or throws an exception.
-Sometimes you have some state that is lazily initialized
-where `late` would be a good fit,
+(if it has one) or throws an exception.
+Sometimes you have some state that's lazily initialized
+where `late` might be a good fit,
 but you also need to be able to *tell* if the initialization has happened yet.
 
-You could do that by storing the state in a `late` variable
-and then have a separate boolean field
-to track whether the variable has been set.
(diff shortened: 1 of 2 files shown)


kwalrath (commented):
We'll want to add this to the toc

src/_guides/language/effective-dart/usage_migrated.md
```diff
@@ -321,6 +321,27 @@ int measureMessage(String? message) {
 {% endprettify %}
 
 
+### AVOID `late` variables if you need to check whether they are initialized.
+
+Dart offers no way to tell if a `late` variable
+has been initialized or assigned to.
+If you access it, it either immediately runs the initializer
+if it has one or throws an exception.
```
kwalrath, line 329:
```suggestion
(if it has one) or throws an exception.
```

src/_guides/language/effective-dart/usage_migrated.md
```diff
@@ -321,6 +321,27 @@ int measureMessage(String? message) {
...
+Sometimes you have some state that is lazily initialized
```
kwalrath, line 330:
```suggestion
Sometimes you have some state that's lazily initialized
```

src/_guides/language/effective-dart/usage_migrated.md
```diff
@@ -321,6 +321,27 @@ int measureMessage(String? message) {
...
+where `late` would be a good fit,
```
kwalrath, line 331:
```suggestion
where `late` might be a good fit,
```

src/_guides/language/effective-dart/usage_migrated.md
```diff
@@ -321,6 +321,27 @@ int measureMessage(String? message) {
...
+but you also need to be able to *tell* if the initialization has happened yet.
+
+You could do that by storing the state in a `late` variable
+and then have a separate boolean field
+to track whether the variable has been set.
+This works, but is redundant since Dart *internally*
+still needs to maintain the initialized status of the `late` variable.
```
kwalrath, line 338:
```suggestion
Although you could detect initialization by storing the state in a `late` variable
and having a separate boolean field
that tracks whether the variable has been set,
that's redundant because Dart *internally*
maintains the initialized status of the `late` variable.
```

src/_guides/language/effective-dart/usage_migrated.md
```diff
@@ -321,6 +321,27 @@ int measureMessage(String? message) {
...
+Instead, it's usually clearer to make the variable non-`late` and nullable.
+Then you can check the variable for `null` to see if it has been initialized.
+(Of course, if `null` is a valid initialized value for the variable,
+then it does make sense to have a separate boolean field.)
```
kwalrath, line 342: I thought a separate paragraph might make this a little more scannable.

```suggestion

Of course, if `null` is a valid initialized value for the variable,
then it probably does make sense to have a separate boolean field.
```

src/_guides/language/effective-dart/usage_migrated.md
```diff
@@ -321,6 +321,27 @@ int measureMessage(String? message) {
...
+where `late` would be a good fit,
+but you also need to be able to *tell* if the initialization has happened yet.
+
+You could do that by storing the state in a `late` variable
+and then have a separate boolean field
+to track whether the variable has been set.
+This works, but is redundant since Dart *internally*
+still needs to maintain the initialized status of the `late` variable.
+Instead, it's usually clearer to make the variable non-`late` and nullable.
+Then you can check the variable for `null` to see if it has been initialized.
```
kwalrath, line 340: Might be nice to have `null` at the end of the paragraph, so it's easier to pick out.

```suggestion
Then you can see if the variable has been initialized
by checking for `null`.
```

kwalrath (commented):
We might be able to make this a little clearer. Maybe.

kwalrath (approved):

munificent:
Thanks for all the improvements!

kwalrath (approved):

PR #3028, opened 2021-03-03
Update Try Dart homepage pad to use null safety

parlough:
Update the "Try Dart" DartPad on the home page to support null safety and update examples appropriately.

cc @johnpryan


kwalrath (commented):
Nice work, @parlough!

tool/dartpad_picker/web/dartpad_picker_main.dart
```diff
@@ -52,13 +52,13 @@ bool isMobile() {
...
-var helloWorld = r'''
-main() {
+const helloWorld = r'''
+void main() {
   print("Hello, World!");
 }
-'''
-    .trim();
-var functions = r'''
+''';
```
johnpryan, line 59: Is trimming not necessary?

parlough: It changed the extra behavior of the line at the end, but in my latest commit I've moved the multi-line literal right after the previous bracket to match the earlier behavior so the strings are equivalent while still allowing them to be const.

johnpryan: cool, LGTM

johnpryan (approved):

PR #3030, opened 2021-03-03
Fix base and focus styles of Try Dart select

parlough:
Before this change it was almost impossible to tell the element was focused.

### Previously:

Base: 
image

Focus:
image

### After:

Base:
image

Focus:
image

I rounded the corners of the select to better match the other buttons on the page.


kwalrath:
Thanks, @parlough!

PR #3033, opened 2021-03-03
Remove link to unsupported russian site

parlough:
The link is currently dead and it wasn't updated in over a year according to the [Github repository](https://github.com/iboldurev/dart-lang). Best not to confuse users or bring them to quite outdated information if it does come back up.

If it comes back up and has updates, then this would definitely be worth being re-included.


kwalrath (approved):
Thanks, @parlough!

PR #3034, opened 2021-03-03
Update mention of Flutter Web in FAQ now that it is stable

parlough:
Now that flutter for web is stable and no longer a technical preview I thought it was fitting to update this mention.


kwalrath (approved):
Nice catch, @parlough!

kevmoo (approved):

PR #3035, opened 2021-03-03
Remove mention of asm.js

parlough:
Since `asm.js` is no longer developed and has long been superseded by WebAssembly, this section doesn't make much sense anymore.

This could be replaced by a section about WebAssembly, but it seems that answer is still in flux but perhaps @mraleph could follow up with a replacement piece, but it's not necessary.


kwalrath (approved):
lgtm, but @kevmoo should take a look

kevmoo:
Yup. This is good to drop. A discussion of wasm may be worth adding at some point.

PR #3036, opened 2021-03-03
Switch mention from error codes to diagnostics

parlough:
The list of diagnostics found at https://dart.dev/tools/diagnostic-messages is much more useful than a literal list of the error strings(https://github.com/dart-lang/sdk/blob/master/pkg/analyzer/lib/error/error.dart).


kwalrath (approved):
Thanks!

PR #3038, opened 2021-03-03
Unfork effective dart

munificent:
Here it is! I broke it up into smaller commits to try to make it easier to review. The unfork commit is literally just file moves with no textual changes.


kwalrath (approved):
lgtm! I'll just wait for the build to finish. I'll probably also stage locally, so I can do a little manual link clicking.

kwalrath (approved):
lgtm!

src/_guides/language/effective-dart/usage.md
```diff
@@ -292,39 +556,19 @@ Instead, there are faster and more readable getters: `.isEmpty` and
...
 
 {:.bad}
-<?code-excerpt "misc/lib/effective_dart/usage_bad.dart (dont-use-length)"?>
+<?code-excerpt "usage_bad.dart (dont-use-length)"?>
 {% prettify dart tag=pre+code %}
 if (lunchBox.length == 0) return 'so hungry...';
 if (!words.isEmpty) return words.join(' ');
 {% endprettify %}
 
-### CONSIDER using higher-order methods to transform a sequence.
```
sungamagnus, line 308: Is there a reason why `CONSIDER using higher-order methods to transform a sequence.` is deleted?  It's still referenced in Overview page 

https://dart.dev/guides/language/effective-dart#:~:text=CONSIDER%20using%20higher-order%20methods%20to%20transform%20a%20sequence.

kwalrath: That's weird. @munificent?

munificent: I added that guideline many many moons ago when programming in a functional style was relatively novel. I saw lots of Dart code being written like Java where every time someone wanted to transform a list, they wrote a big block of imperative code. The world has moved on and now everyone seems to be entirely comfortable with a functional style.

Also, now that we have `for` and `...` in lists, there are many ways to build and transform collections. Having this CONSIDER guideline inadvertently seemed to imply that using those methods was *preferred* over using collection-for and friends, which is not the case.

**TL;DR: It wasn't needed anymore and was kind of misleading.**

kwalrath: The really weird thing is that it's still in the TOC, even though it disappeared. Why is that?

parlough: Likely wasn't regenerated, seems to be gone with https://github.com/dart-lang/site-www/commit/c6fb01eaefa34e7ac4f87d4c787d5d773dfdc1a0#diff-d55f6b13cd55d9aa5d5d21b2dd633abf1bca9e93a1164841276b670fcc15f936L116

PR #3044, opened 2021-03-05
Update prettier definition for Dart

parlough:
We probably could(/should) switch to using the shared assets folder but for now just updates the Dart prettify definition.

### DIFF
--- a/src/_assets/js/vendor/code-prettify/lang-dart.js
+++ b/src/_assets/js/vendor/code-prettify/lang-dart.js
@@ -54,10 +54,10 @@ PR['registerLangHandler'](
       [PR['PR_KEYWORD'], /^\b(?:assert|async|await|break|case|catch|continue|default|do|else|finally|for|if|in|is|new|on|rethrow|return|super|switch|sync|this|throw|try|while|yield)\b/i],
 
       // Declaration keywords.
-      [PR['PR_KEYWORD'], /^\b(?:abstract|const|covariant|extends|external|factory|final|get|implements|native|operator|set|static|typedef|var|with)\b/i],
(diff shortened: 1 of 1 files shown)


kwalrath:
Thanks, @parlough! Where should I look to see the difference between the current version and my staged version of this PR?

parlough:
@kwalrath I would look at a page like [Understanding Null Safety](https://dart.dev/null-safety/understanding-null-safety). The changes are subtle due to the current style used have but the primary change you'll notice is `late` and `required` will now be colored like other keywords(pink) instead of like variables(purple).

kwalrath:
> @kwalrath I would look at a page like [Understanding Null Safety](https://dart.dev/null-safety/understanding-null-safety). The changes are subtle due to the current style used but the primary change you'll notice is `late` and `required` will now be colored like other keywords(pink) instead of like variables(purple).

I'm still seeing the purple, even though I've deleted all my old files, rebuilt everything, and restarted my local server. I even tried using an incognito window. E.g. in my locally staged version of https://dart.dev/null-safety/understanding-null-safety#late-final-variables, I still see a purple `late`:

image

domesticmouse (approved):
Just confirming, these changes have been submitted upstream? I don't want to lose these next time we update the vendored code. =)

parlough:
@kwalrath I'll test locally again and respond here with what I find.

@domesticmouse I would love to, but unfortunately my other patches from 1+ years ago still haven't been accepted and the [repo is archived](https://github.com/googlearchive/code-prettify). 

The goal is to switch to(or create) a better long term replacement, so this is more of a bandaid for now.

parlough:
@kwalrath I'm sorry, I forgot we only use prettify for a few situations, most of which are not dart related. I don't think we currently have any examples that will utilize these improvements.

kwalrath:
It sounds like accepting this PR is harmless, so I'll go ahead and do it.

PR #3045, opened 2021-03-08
Document options for pub get and upgrade

### LINKED ISSUE #2898: Add pages for new pub commands
Also, commands -> subcommands.

* Subcommands we should probably document:
  * New subcommands: add, login, remove
  * Other subcommands: logout
* Subcommands we should remove: run

2.10 stable's output from `dart help pub`:
```
...
Available commands:
  cache       Work with the system cache.
  deps        Print package dependencies.
  downgrade   Downgrade the current package's dependencies to oldest versions.
  get         Get the current package's dependencies.
  global      Work with global packages.
  logout      Log out of pub.dartlang.org.
  outdated    Analyze your dependencies to find which ones can be upgraded.
  publish     Publish the current package to pub.dartlang.org.
  run         Run an executable from a package.
  upgrade     Upgrade the current package's dependencies to latest versions.
  uploader    Manage uploaders for a package on pub.dartlang.org.
  version     Print pub version.
```

2.12.0-133.7.beta output from the same:
```
...
Available subcommands:
  add         Add a dependency to pubspec.yaml.
  cache       Work with the system cache.
  deps        Print package dependencies.
  downgrade   Downgrade the current package's dependencies to oldest versions.
  get         Get the current package's dependencies.
  global      Work with global packages.
  login       Log into pub.dev.
  logout      Log out of pub.dev.
  outdated    Analyze your dependencies to find which ones can be upgraded.
  publish     Publish the current package to pub.dartlang.org.
  remove      Removes a dependency from the current package.
  upgrade     Upgrade the current package's dependencies to latest versions.
  uploader    Manage uploaders for a package on pub.dartlang.org.
```

micpap25:
Partially fixes #2898

Documents unlisted options for `dart pub get` and `dart pub upgrade`. I don't know too much about precompile, so it would be useful if someone else went more in-depth about what exactly that option implies.

Also, perhaps there is something to be done about the amount of overlap between the two articles?

This is also just a general issue I noticed, but the wording between the different `help` sections on the CLI are very inconsistent. I might file an issue for this on the main repository, because we can't make the docs for these commands consistent when the CLI's built-in documentation isn't.

### DIFF
--- a/src/tools/pub/cmd/pub-get.md
+++ b/src/tools/pub/cmd/pub-get.md
@@ -143,16 +143,22 @@ command-line argument, as mentioned above.
 For options that apply to all pub commands, see
 [Global options](/tools/pub/cmd#global-options).
 
+### `--offline`
+
+Uses cached packages rather than downloading
+from the network. See above for a more detailed
+explanation.
+
 ### `--dry-run` or `-n`
 
-With this, pub reports the dependencies that would
-but does not make the changes. This is useful if you
+Reports the dependencies that would be changed,
+but doesn't make the changes. This is useful if you
 want to analyze updates before making them.
 
 ### `--precompile`
 
-This option makes pub compile executables in its dependencies
-before running the get process.
+Creates snapshots of the
+project's executables in direct dependencies.
 
 <aside class="alert alert-info" markdown="1">
 *Problems?*
--- a/src/tools/pub/cmd/pub-get.md
+++ b/src/tools/pub/cmd/pub-get.md
@@ -137,23 +137,20 @@ run [`dart pub upgrade`](/tools/pub/cmd/pub-upgrade) to upgrade to a later versi
 
 ## Options
 
-The `dart pub get` command supports the `--offline`
-command-line argument, as mentioned above.
-
 For options that apply to all pub commands, see
 [Global options](/tools/pub/cmd#global-options).
-
-### `--offline`
-
-Uses cached packages rather than downloading
-from the network. See above for a more detailed
-explanation.
 
 ### `--dry-run` or `-n`
 
 Reports the dependencies that would be changed,
 but doesn't make the changes. This is useful if you
 want to analyze updates before making them.
+
+### `--offline`
+
+Uses cached packages rather than downloading
+from the network.
+For details, see [Getting while offline](#getting-while-offline).
 
 ### `--precompile`
 
--- a/src/tools/pub/cmd/pub-get.md
+++ b/src/tools/pub/cmd/pub-get.md
@@ -6,7 +6,7 @@ description: Use dart pub get to retrieve the dependencies used by your Dart app
 _Get_ is one of the commands of the [pub tool](/tools/pub/cmd).
 
 {% prettify nocode tag=pre+code %}
-$ dart pub get [--offline]
+$ dart pub get [--offline] [--dry-run] [--precompile]
 {% endprettify %}
 
 This command gets all the dependencies listed in the
@@ -142,6 +142,18 @@ command-line argument, as mentioned above.
 
 For options that apply to all pub commands, see
 [Global options](/tools/pub/cmd#global-options).
+
+### `--dry-run` or `-n`
+
+With this, pub reports the dependencies that would
+but does not make the changes. This is useful if you
+want to analyze the updates before making them,
+as you cannot undo the changes afterwards.
+
+### `--precompile`
+
+This option makes pub compile executables in its dependencies
+before running the get process.
 
 <aside class="alert alert-info" markdown="1">
 *Problems?*
--- a/src/tools/pub/cmd/pub-get.md
+++ b/src/tools/pub/cmd/pub-get.md
@@ -147,8 +147,7 @@ For options that apply to all pub commands, see
 
 With this, pub reports the dependencies that would
 but does not make the changes. This is useful if you
-want to analyze the updates before making them,
-as you cannot undo the changes afterwards.
+want to analyze updates before making them.
 
 ### `--precompile`
 
--- a/src/tools/pub/cmd/pub-get.md
+++ b/src/tools/pub/cmd/pub-get.md
@@ -6,7 +6,7 @@ description: Use dart pub get to retrieve the dependencies used by your Dart app
 _Get_ is one of the commands of the [pub tool](/tools/pub/cmd).
 
 {% prettify nocode tag=pre+code %}
-$ dart pub get [--offline] [--dry-run] [--precompile]
+$ dart pub get [args]
 {% endprettify %}
 
 This command gets all the dependencies listed in the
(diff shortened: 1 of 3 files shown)


kwalrath:
Thanks for the PR, @micpap25! Reviewing now...

src/tools/pub/cmd/pub-get.md
```diff
@@ -143,6 +143,17 @@ command-line argument, as mentioned above.
 For options that apply to all pub commands, see
 [Global options](/tools/pub/cmd#global-options).
 
+### `--dry-run` or `-n`
+
+With this, pub reports the dependencies that would
+but does not make the changes. This is useful if you
+want to analyze updates before making them.
+
```
kwalrath, line 151: Let's add the `--offline` flag to this section. (It can mostly point to previous sections.)

src/tools/pub/cmd/pub-get.md
```diff
@@ -143,6 +143,17 @@ command-line argument, as mentioned above.
 For options that apply to all pub commands, see
 [Global options](/tools/pub/cmd#global-options).
 
+### `--dry-run` or `-n`
```
kwalrath, line 146: Interesting approach! At first it seemed weird to use h3s (so big, so emphatic), but I like having the options visible in the TOC.

The only badness is that the code font doesn't look great in headings (because of the color maybe? the huge size?) and isn't picked up in the TOC. Still, we could change the code font color/size in h3s to mitigate this.

@parlough, do you have an opinion on making h3s the standard way to document options for tools? Currently we use `<dl>`s instead (e.g. at https://dart.dev/tools/pub/cmd/pub-cache).

The only place I've found where it wouldn't work is in http://localhost:4000/tools/dart2js#options, which has so many options that they're divided into groups. (Weirdly, that page uses h4s instead of h3s, so even the groups don't appear in the TOC.) But I think that exception is fine.

src/tools/pub/cmd/pub-get.md
```diff
@@ -143,6 +143,17 @@ command-line argument, as mentioned above.
...
+
+With this, pub reports the dependencies that would
+but does not make the changes. This is useful if you
```
kwalrath, line 149: Let's start these with a _-s_ verb. Also, we use [contractions](https://developers.google.com/style/contractions) (especially for _not_) and [semantic line breaks](https://github.com/dart-lang/site-shared/blob/master/doc/writing-for-dart-and-flutter-websites.md#semantic-line-breaks).

```suggestion
Reports the dependencies that would be made,
but doesn't make the changes.
This option is useful if you
```

src/tools/pub/cmd/pub-get.md
```diff
@@ -143,6 +143,17 @@ command-line argument, as mentioned above.
...
+want to analyze updates before making them.
+
+### `--precompile`
+
+This option makes pub compile executables in its dependencies
+before running the get process.
```
kwalrath, line 155: This doesn't sound right... Wouldn't it compile the executables _after_ getting them?

```suggestion
Creates snapshots of the executables in direct dependencies.
```

kwalrath (commented):
I started with `pub get`, which set some standards that I'd like to make sure we all like before continuing with the next command.

src/tools/pub/cmd/pub-get.md
parlough: @kwalrath I think using h3s seems like a fair idea, for both being present in the TOC but also to simplify contributing and later editing.

Do you have anything in particular in mind in terms of the style change you mentioned? It does seem slightly strange to mix code style with normal text in headers.

src/tools/pub/cmd/pub-get.md
```diff
@@ -143,6 +143,17 @@ command-line argument, as mentioned above.
 For options that apply to all pub commands, see
 [Global options](/tools/pub/cmd#global-options).
 
+### `--dry-run` or `-n`
+
+With this, pub reports the dependencies that would
```
parlough, line 148: I'm not sure the full intention here. Reports the dependencies that would _what_?

micpap25:
Updated with the changes suggested by @kwalrath , and thank you @parlough for catching my typo in `--dry-run`.

src/tools/pub/cmd/pub-get.md
micpap25: I used h3's primarily because I wanted to keep consistent with other pages that listed options ( my basis was https://dart.dev/tools/pub/cmd/pub-lish ).

kwalrath (approved):

kwalrath (changes requested):
Oops, I just realized that pub-upgrade needs some changes to match the changes I just made to pub-get. @micpap25 can you make them?

(I forgot to list it, but I also put the options in alpha order. I tend to default to that, unless there's a good reason not to.)

micpap25:
@kwalrath Alphabetically ordered the options in pub upgrade

kwalrath:
> @kwalrath I alphabetically ordered the options in pub upgrade

Could you also change "See above for a more detailed explanation" to be similar to what I did for pub get? (switching the sentence around, and linking to the actual section in the page)

(If you can't get to it, I'll do it.)

Thanks!

kwalrath:
And... I should actually review the pub upgrade text, so you might want to wait for that. Yeesh, I keep forgetting what I've looked at.

kwalrath (commented):
OK, I _think_ this covers everything else.

kwalrath (approved):
Thank you for your good work and perseverance, @micpap25!

kwalrath:
I wonder if we shouldn't move the "Commit the pubspec.yaml file" tip up to the top of any doc for commands that modify the pubspec. But that can be a separate PR.

src/tools/pub/cmd/pub-upgrade.md
```diff
@@ -105,10 +105,55 @@ run `dart pub upgrade` again to upgrade to a later version.
...
+
+[`dart pub outdated`]: /tools/pub/cmd/pub-outdated
+
+{{ site.alert.tip }}
+  Commit the `pubspec.yaml` file before running this command,
+  so that you can undo the changes if necessary.
+{{ site.alert.end }}
+To check which dependencies will be upgraded,
+you can use `dart pub upgrade --major-versions --dry-run`.
+
```
jonasfj, line 131: @micpap25 also it's actually possible to do:
```
dart pub upgrade --major-versions foo bar ...
```
Assume you only want to upgrade major versions for packages `foo` and `bar`...

When running in this mode, it will still allow upgrades of other packages, but only if those upgrades don't require changes in `pubspec.yaml`. So in effect this will only do major version upgrades for packages `foo` and `bar`... other packages will get normal upgrade, if necessary.

--------

This partial behavior is also available for `--null-safety` with `dart pub upgrade --null-safety <package>...`, but here it's discouraged, because partial null-safety migration is a bad idea. So I don't think we should highlight the partial null-safety migration functionality.

Besides `dart pub upgrade --null-safety` will, if it fails, print a message saying which packages didn't have a null-safety compatible version, and then print the command for running a partial null-safety migration -- and discourage users from doing so :rofl: 

But if you really need to do partial null-safety migration, you'll discover it when you run `dart pub upgrade --null-safety` so let's not complicate documentation with such details.

PR #3053, opened 2021-03-10
Add language info strings to Markdown code blocks in Effective Dart

### LINKED ISSUE #3052: Avoid language-free backticks in Effective Dart
https://dart.dev/guides/language/effective-dart/documentation#prefer-backtick-fences-for-code-blocks has an example with a backtick block, starting with three backticks, which does not include a language specification (or an ["info string"](https://github.github.com/gfm/#info-string)). However, https://github.com/dart-lang/dartdoc/issues/2520 was opened against dartdoc, and now dartdoc warns about language-free code blocks, as of https://github.com/dart-lang/dartdoc/pull/2559.

I think Effective Dart should match dartdoc's warnings in terms of the good examples.

parlough:
Since we are adding a warning to dartdoc that users should specify a language, best to lead by example.

Fixes #3052

### DIFF
--- a/src/_guides/language/effective-dart/documentation.md
+++ b/src/_guides/language/effective-dart/documentation.md
@@ -421,7 +421,7 @@ a flavor of what's supported:
 ///
 /// Code blocks are fenced in triple backticks:
 ///
-/// ```
+/// ```dart
 /// this.code
 ///     .will
 ///     .retain(its, formatting);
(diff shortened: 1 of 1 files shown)


src/_guides/language/effective-dart/documentation.md
```diff
@@ -421,7 +421,7 @@ a flavor of what's supported:
 ///
 /// Code blocks are fenced in triple backticks:
 ///
-/// ```
+/// ```dart
```
kwalrath, line 424: Looks the build will fail until we update the included file, docs_good.dart

kwalrath (commented):
Thanks, @parlough! I think you just need to make the .dart file change to fix the build.

src/_guides/language/effective-dart/documentation.md
parlough: Thanks. I always forget to include changes in both places...

PR #3063, opened 2021-03-12
Remove mention of dartdevc from analysis options page

parlough:
I'm not sure this mention of a resource makes a lot of sense anymore as it's a tool for compiling and debugging web applications rather than a static analysis tool. 

I believe in the past this page had a lot of mentions of strong mode and how dartdevc only supported it or something along those line, but those mentions aren't relevant and do not exist anymore.

_Sorry if this PR ends up appearing multiple times, Github is being flaky and providing me with lots of unicorns!_


kwalrath (approved):
Thanks, @parlough!

PR #3066, opened 2021-03-13
Correct sample code of 'Effective Dart: Usage' page

ngdangtu-vn:
On production site: https://dart.dev/guides/language/effective-dart/usage#dont-allow-an-import-path-to-reach-into-or-out-of-lib

The sample code above the line '_Dart thinks those are imports of two completely unrelated libraries._' has wrong a format. It supports to be 2 sample codes. One's good and one's bad each has one line. But it got mixed into one bad sample. Therefore I separated it.

This isn't rocket science, so I didn't do a local build to test. But I make sure to double check the syntax so it should display properly. (The reasons I couldn't do that yet due to I having some trouble with `rvm`.)


ngdangtu-vn:
@googlebot I signed it!

kwalrath (changes requested):
Actually, I'm pretty sure that @munificent meant this sample to be the way it is — he's illustrating that the same file imports `api.dart` in two ways.

However, this rule _could_ use a _good_ example, down at the bottom. Maybe something like this:

> Instead, when you need to reach into a package’s lib directory (even from the same package’s test directory or any other top-level directory), use a package: import:
>
> import 'package:my_package/api.dart';
>
> A package should never reach out of its lib directory and import libraries from other places in the package.

ngdangtu-vn (commented):
Then I will remove the 'bad' example which leaves sample only. Also adding the good example as you suggested. How's about it?

src/_guides/language/effective-dart/usage.md
```diff
@@ -101,7 +101,6 @@ my_package
 
 And say `api_test.dart` imports `api.dart` in two ways:
 
-{:.bad}
```
kwalrath, line 104: We should leave this line in. It's **bad** to import the same library two different ways.

src/_guides/language/effective-dart/usage.md
```diff
@@ -117,6 +116,13 @@ Instead, when you need to reach into a package's `lib` directory
 (even from the same package's `test` directory
 or any other top-level directory),
 use a `package:` import.
+
+{:.good}
+<?code-excerpt "test/api_test.dart"?>
```
kwalrath, line 121: We'll need to change the excerpt tag and the test/api_test.dart file to make the build work, but I can handle that.

kwalrath:
Fixes #3124

kwalrath (approved):

PR #3068, opened 2021-03-15
Remove library tour hyperlink

hrishikesh-kadam:
May be not required because `_dart-io-tour.md` page itself is displayed on the `library-tour.md` page.


kwalrath (approved):
Nice, catch. Thanks!

PR #3080, opened 2021-03-18
Migrated 'extension methods' to null safety

miquelbeltran:
I saw that the changes are now being made directly into the site sources rather than making a copy, I still followed the process to copying over the example to the `null_safety_examples`.

I have upgraded the pubspec and as well the references to the code excerpts.

In this case there were no other code changes required.

@kwalrath @sfshaza2 @RedBrogdon


kwalrath (approved):
Thanks, @miquelbeltran! It's nice to see that it's sometimes possible to migrate code without changing anything outside of the pubspec.

PR #3064, opened 2021-03-13
Include linter rules on Dart.dev

parlough:
The goal is to standardize hosting rule information on [dart.dev](https://dart.dev) similar to the [analyzer diagnostics](https://dart.dev/tools/diagnostic-messages). As a result, the style followed implementing this followed closely to that.

**Parker's must complete tasks:**

- [x] Add dart specifier to linter code blocks (https://github.com/dart-lang/linter/pull/2508)
- [x] Actually write out an appropriate description of the linter
- [x] Write descriptions for the types of rules
- [x] Add explanations for the maturity levels
- [x] Improve sidebar spacing and formatting so entire lint rule can be read (https://github.com/dart-lang/site-www/pull/3089)
- [x] Implement some GitHub action to indicate this needs to be updated (https://github.com/dart-lang/site-www/pull/3171)

**Long term tasks for _after_ landing:**

- Flesh out each individual lint documentation, standardizing on a style similar to analyzer diagnostics
- Update other locations to refer to this link where appropriate
- Include rule set(s) in the listings in some fashion(need to iterate on designs). This will be important when the recommended rule sets are released.

**You can find a staged version of this here:**
_Last updated 05-09-21_
https://parlough-dart-dev.web.app/tools/linter-rules

PR #3089, opened 2021-03-21
Improvements for long text in TOC

parlough:
Currently for pages like [Diagnostic messages](https://dart.dev/tools/diagnostic-messages) and the [linter page](https://github.com/dart-lang/site-www/pull/3064) in the future, the full-width TOC cuts off the full name of long entries, making it hard to determine what you're navigating to. Especially since a lot of rules share the same beginning, so there is no way to differentiate between them.

This PR slightly increases the width that is available to the TOC when needed as well as allowing wrapping in the middle of words.

**After change:**
image

**Before change:**
image

### DIFF
--- a/src/_assets/css/main.scss
+++ b/src/_assets/css/main.scss
@@ -584,7 +584,16 @@ img {
   }
   .toc-entry {
     padding-bottom: 0.5rem;
+
+    // Indent subsequent lines
+    margin-left: 1rem;
+    text-indent: -1rem;
+
+    & .toc-entry {
+      margin-left: 0;
+    }
   }
+
   .nav {
     display: block;
     padding-top: 0.5rem;
--- a/src/_assets/css/main.scss
+++ b/src/_assets/css/main.scss
@@ -563,11 +563,13 @@ img {
   @include sidenav-top-position;
 
   position: fixed;
-  bottom: 0px;
-  width: 226px;
-  right: 0px;
+  bottom: 0;
+  min-width: 226px;
+  max-width: 256px;
+  right: 0;
   overflow-x: hidden;
   overflow-y: auto;
+  overflow-wrap: break-word;
   z-index: 999;
 
   // override shared/_toc.scss
--- a/src/_assets/css/main.scss
+++ b/src/_assets/css/main.scss
@@ -565,7 +565,7 @@ img {
   position: fixed;
   bottom: 0;
   min-width: 226px;
-  max-width: 256px;
+  max-width: 246px;
   right: 0;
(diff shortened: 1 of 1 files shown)


kwalrath:
LGTM (but I am not a designer). I don't see any problems with other pages, either. I've staged it, and compared these pages with a window wide enough to see the right-hand nav:

https://dart.dev/tools/diagnostic-messages
https://kw-staging-dartlang-2.firebaseapp.com/tools/diagnostic-messages
https://dart.dev/guides/language/language-tour
https://kw-staging-dartlang-2.firebaseapp.com/guides/language/language-tour

The only objection I can think of is that the designer picked the TOC position & width for a reason. But it seems like a bug to have a long word be chopped off at the right. It's a little ugly to continue the word on the next line, but it's much more usable. And that weird break doesn't happen as long as the words aren't too long.

E.g. the new way (on top) is a little ugly when the name is long, but it conveys much more info than the old way:

image

image

The new way (on top) reduces the space between the content and the rhs toc, but that looks fine to me.

image

image

parlough:
I also wasn't super happy with the word breaks, so I'm especially open to other suggestions. I'll explore a bit more myself. 

I tried showing the full title on hover, but it didn't help with glanceability(not sure if this is even a word :p)

As for the width, I can definitely set that back to the old width. I tried to avoid cutting off more titles and this didn't seem to break anything.

kwalrath:
> I also wasn't super happy with the word breaks, so I'm especially open to other suggestions. I'll explore a bit more myself.

Maybe use an ellipsis at the beginning of the second line? Or just a little indentation?

Are there any cases where the headings is >2 lines? If so, and if it's possible to just display the end of the string (big _if_), then starting the 2nd (final) line with an ellipsis would make sense.

> I tried showing the full title on hover, but it didn't help with glanceability(not sure if this is even a word :p)

It should be :D. I think _scannability_ is the word I'd usually use.

> As for the width, I can definitely set that back to the old width if that is desired. I tried to avoid cutting off more titles and this didn't seem to break anything.

I'm actually happy with the new width. @johnpryan might have an opinion, too.

johnpryan (approved):
I like this change - the new width doesn't seem to cause any issues.

parlough:
@kwalrath 

With just basing it off of HTML and CSS the best I could figure out is a little indentation, I'm not sure how I feel about it though:
image

What do you think?

johnpryan (approved):

kwalrath (approved):
LGTM. Thank you, @parlough!

PR #3102, opened 2021-03-25
Reorder "For loops" examples of the Language Tour

jamesderlin:
The "For loops" section of the Language Tour described
`Iterable.forEach` before `for-in` and suggested `Iterable.forEach`
as a "good option".  This currently is not the recommendation from
Effective Dart, which discourages using `Iterable.forEach` with
anonymous functions.

* Reorder the examples in the "For loops" section to put greater
  emphasis on `for-in`.
* Describe `Iterable.forEach` as "an option" rather than as a "good
  option".
* Adjust the code snippets so that the `Iterable.forEach` example
  follows the Effective Dart guideline by using a tear-off instead
  of an anonymous function.

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -2237,9 +2237,9 @@ callbacks.forEach((c) => c());
 The output is `0` and then `1`, as expected. In contrast, the example
 would print `2` and then `2` in JavaScript.
 
-If the object that you are iterating over is an Iterable (such as List or Set),
-you can use the `for-in` form of [iteration](/guides/libraries/library-tour#iteration)
-if you don't need to know the current iteration counter:
+If the object that you are iterating over is an Iterable (such as List or Set)
+and if you don't need to know the current iteration counter, 
+you can use the `for-in` form of [iteration](/guides/libraries/library-tour#iteration):
 
 <?code-excerpt "../null_safety_examples/misc/lib/language_tour/control_flow.dart (collection)"?>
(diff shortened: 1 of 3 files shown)


src/_guides/language/language-tour.md
```diff
@@ -2237,24 +2237,23 @@ callbacks.forEach((c) => c());
 The output is `0` and then `1`, as expected. In contrast, the example
 would print `2` and then `2` in JavaScript.
 
-If the object that you are iterating over is an Iterable, you can use the
-[forEach()][] method. Using `forEach()` is a good option if you don’t need to
-know the current iteration counter:
+If the object that you are iterating over is an Iterable (such as List or Set),
+you can use the `for-in` form of [iteration](/guides/libraries/library-tour#iteration)
+if you don't need to know the current iteration counter:
```
parlough, line 2242: Now that you made this one sentence, it seems a bit strange to have the two conditions separated here. Perhaps combining the them with `and` or keeping them as separate sentences like before.

I'm leaning a bit towards keeping the conditional sentences separate as you technically "can" use `for-in` even if you do need the current iteration counter or if an iteration counter doesn't even make sense for the collection type.

jamesderlin: I'll combine them with "and".  If someone needs an iteration counter, they should prefer `for` instead of `for-in`.

kwalrath (approved):

parlough (approved):

PR #3120, opened 2021-04-05
Update get-dart.md

negitorogithub:
Added 'Flutter is built with a particular Dart SDK, and can only be used with a given version of that SDK.'
(reference: https://github.com/flutter/flutter/issues/62790#issuecomment-668201880 )

### DIFF
--- a/src/get-dart.md
+++ b/src/get-dart.md
@@ -24,6 +24,7 @@ any of the following are true:
   For example, you might have a continuous integration (CI)
   setup that requires Dart but not Flutter.
 
+**Flutter is built with a particular Dart SDK, and can only be used with a given version of that SDK.**
 
 ## Installing the Dart SDK {#install}
 


negitorogithub:
@googlebot I signed it!

negitorogithub:
Should I close this issue?

domesticmouse:
@negitorogithub I'm curious what the driving issue is for this change. We like pairing a change like this one with a bug report that explains the problem, and motivates the need for the fix.

My understanding is that Flutter bundles it's own copy of Dart at the correct version level. There is no issue with having a system installed version of Dart which differs from the one required by Flutter, IIUC.

Or am I missing something?

negitorogithub:
@domesticmouse 
Thank you for replying.		
>My understanding is that Flutter bundles it's own copy of Dart at the correct version level. There is no issue with having a system installed version of Dart which differs from the one required by Flutter

It's true, but I didn't know that 
(https://github.com/flutter/flutter/issues/62790#issuecomment-668201880)
>Flutter is built with a particular Dart SDK, and can only be used with a given version of that SDK.

In other words, some people(and I) may think that they can use a certain version of flutter with another version(not bundled version) of dart.
For example, https://github.com/flutter/flutter/issues/62790#issuecomment-668378234.
> i did not find this sentence can only be used with a given version of that SDK

domesticmouse:
I'd argue that documenting how Flutter uses Dart, and that it has a dependence on a specific version, is a Flutter implementation detail. I'd suggest raising a bug on the flutter/flutter issue tracker to document how Flutter integrates Dart into it's build chain, for example. 

@mit-mit I'd appreciate your input here.

mit-mit:
I agree that if this should be documented, it should be on dart.dev.

PR #3121, opened 2021-04-06
Troubleshooting pub socket error

jonasfj:
Some lessons from https://github.com/dart-lang/pub-dev/issues/4690

### DIFF
--- a/src/tools/pub/troubleshoot.md
+++ b/src/tools/pub/troubleshoot.md
@@ -138,8 +138,22 @@ pub get failed (server unavailable) -- attempting retry 1 in 1 second...
 This can happen for many reasons, check your internet connection, and verify that
 you do not have firewalls or other security software that block internet access from `dart`.
 
-Users have reported issues using `dart pub` along with:
- * `Kaspersky Internet Security`
+Users have reported firewall configuration instructions for using `dart pub` along with:
 
-It might be possible that `dart pub` can be used along side such firewalls,
-if firewalls are correctly configured.
+* <details><summary><b>Kaspersky Internet Security</b></summary>
+
+   Even with _Kaspersky Internet Security_ protection turned off in the menu bar, the VPN application filter `sysextctrld` will still run in 
+   the background causing one not to be able to connect to `https://pub.dartlang.org`.
+   The URL must be added to the _"Trusted Zone"_. Here's how:
+
+   - Open Kaspersky Internet Security and click on the "Privacy" icon.
+   - Under the "Block website tracking" section click on the "Preferences" button.
+   - On the top icon bar select "Threats".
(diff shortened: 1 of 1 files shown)


src/tools/pub/troubleshoot.md
```diff
@@ -123,3 +123,23 @@ Try this workaround, which uses the command line to complete sign-in:
...
+```
+
+This can happen for many reasons, check your internet connection, and verify that
+you do not have firewalls or other security software that block internet access from `dart`.
+
+Users have reported issues using `dart pub` along with:
+ * `Kaspersky Internet Security`
+
+It might be possible that `dart pub` can be used along side such firewalls,
+if firewalls are correctly configured.
```
jonasfj, line 145: I'm not sure if there is a good way to warn users, but it seems that many of these security products contain multiple security features. And it's easy for a user to disable one of the security features (thinking they've disabled all), observe that disabling it didn't resolve the issue for them, and thus, conclude that it is not their security software that is interfering.

EtienneBauscher: @jonasfj - I must apologise to you and István Soós, I installed the Security Software without configuring trusted zones. I reinstalled it now and added `https://pub.dartlang.org` to the trusted zone and `flutter pub get` ran without a problem. My humble apologies for the inconvenience caused, it was due to my own inexperience.

jonasfj: @EtienneBauscher could you provide a few hints on how to configure a trust zone.

Then maybe we change this section to say:
> Configuring trust zones for ...

Or if you publish those steps as a gist we can just link there for people using this software..

It's better than some vague comments saying that some users had trouble using this with that..

EtienneBauscher: @jonasfj - Will the below suffice?

https://gist.github.com/EtienneBauscher/5d8cefe797e0400dd001d461ecc6d904

kwalrath: @jonasfj did you plan to incorporate some of Etienne's text?

kwalrath: @jonasfj do you plan to update the trust zone info?

jonasfj: Yeah, this seems useful, I've cleaned up a bit and included it as an expandable section.

If we get other reports on how to configure firewalls, we can add the instructions here.

kwalrath:
I've staged my suggested (and committed :) changes:

https://dartlang-org-staging-0.firebaseapp.com/tools/pub/troubleshoot#pub-get-socket-error

kwalrath (approved):

parlough (approved):

kwalrath:
@jonasfj are you OK with my changes?

PR #3125, opened 2021-04-06
Update sidenav wrt pub and dartfmt

mit-mit:
- Remove redundant `dartfmt` entry
- Rename `pub` to `dart pub`


kwalrath (approved):
Can't believe I didn't catch the dartfmt duplication...

PR #3126, opened 2021-04-06
Update sidenav wrt dart web

mit-mit:
- Move `dart2js` and `dartdevc` down next to webdev, as most developers should be spawning them via `webdev`


kevmoo (approved):

kwalrath (changes requested):
This list was alphabetical. Rather than change that, I'd prefer to either take these pages out of the sidenav or indent them under web, somehow. (Not all the commands need to be in the sidenav, as long as there's a parent page that can represent them. E.g. see https://dart.dev/tools/pub/cmd/pub-cache.)

mit-mit:
How about we just delete them then? They are already linked from several other pages, incl. https://dart.dev/tools/dart-compile#js and https://dart.dev/tools/webdev.

parlough (approved):
I agree removing them makes the most sense, as `webdev` should be their primary destination.

You'll have to merge master in or rebase, there's a conflict with your other sidebar changes.

kwalrath (approved):

PR #3136, opened 2021-04-09
Update language spec

mit-mit:


kwalrath (approved):

kwalrath:
Oops, I merged without seeing that @eernstg was supposed to review this. I guess we can always revert if necessary...

PR #3143, opened 2021-04-11
Change MAC to macOS

### LINKED ISSUE #3137: Mac -> macOS
Per the discussion in #2892.

bharat-biradar:
fixes #3137

### DIFF
--- a/src/tools/sdk/archive/_archives_table.html
+++ b/src/tools/sdk/archive/_archives_table.html
@@ -7,7 +7,7 @@
     <label for="{{ include.channel }}-os">OS:</label>
     <select id="{{ include.channel }}-os">
       <option value="all">All</option>
-      <option value="macos" id="{{include.channel}}-macos" class="macos-option">Mac</option>
+      <option value="macos" id="{{include.channel}}-macos" class="macos-option">macOS</option>
       <option value="linux" id="{{include.channel}}-linux" class="linux-option">Linux</option>
       <option value="windows" id="{{include.channel}}-windows" class="windows-option">Windows</option>
     </select>


kwalrath (changes requested):
It's a good start, but we'll need to update the code, too, to avoid having Mac in some places and macOS in others. I think updating the code will require a local copy so we can regenerate the JavaScript.

Also, even once we update the script, the issue probably needs to stay open, since there are lots of other occurrences of _Mac_ on this site.

image

bharat-biradar:
I completely missed noticing it. I'll try to make the changes as soon as possible. Incase of any doubt I'll ask them here.

bharat-biradar:
Could you please elaborate on what you meant by `regenerate javascript`?

kwalrath:
> Could you please elaborate on what you meant by `regenerate javascript`?

There's some info at https://github.com/dart-lang/site-www/blob/master/src/tools/sdk/archive/_README.md.

It can take a bit of time and effort to get the local build of site-www working, so I'll understand completely if you don't want to go to the trouble!

bharat-biradar:
Thank you. I'll take a look at the site working before deciding.

bharat-biradar:
@kwalrath I could not figure out how to change the version table. But to the already made changes i.e `index.html` file should I regenerate the javascript as it says there that you should do it only when dart code is modified and I am changing an HTML file here. If I should I'll just regenerate the javascript and push it to the current PR so that at least this change can be completed.

kwalrath:
I believe you'd need to modify the Dart code; otherwise, JavaScript regeneration won't do anything.

bharat-biradar:
Then is there no need to regenerate javascript for the currently made changes? i.e in this PR.

Issue #2940, opened 2021-02-11
Add promotion info/workaround to /null-safety


kwalrath:
Details in https://github.com/dart-lang/sdk/issues/44900.

[Example from Lasse & mit](https://dartpad.dev/6e35cb85754ce3d44b1e94bf4d69caec?null_safety=true)

Note that `BadEvent` needs to use `year!` because `year` could have been changed by an overriding getter in a subclass. `GoodEvent` shadows `year` with a local variable (`var year = this.year`) so that the code can just use `year`.

kwalrath:
I believe this is covered by https://dart.dev/guides/language/effective-dart/usage_migrated#consider-copying-a-nullable-field-to-a-local-variable-to-enable-type-promotion (which will be moved to https://dart.dev/guides/language/effective-dart/usage once null safety is in the stable SDK).

I wonder if this should also be in the FAQ, which is based on an internal FAQ.

/cc @ryanmcginnis @davidmorgan

kwalrath:
I'm working on this and hope to have something for @stereotype441 to look at early next week. I/O permitting. :)

kwalrath:
The updated page is out: https://dart.dev/tools/non-promotion-reasons

PR #3148, opened 2021-04-14
Placeholder page & go links for non-promotion docs

kwalrath:
Contributes to #2940 and https://github.com/dart-lang/sdk/issues/44900.

I'm not sure how to frame this page, so a few questions (which don't necessarily have to be answered definitively before we merge this PR):

* Is everything on this page going to be about promotion from nullable to non-nullable types? Or might we get some visits related to other forms of type promotion?
* What should the title be? What's going to set expectations correctly?
* Where on dart.dev should this page really live? Initially I thought it was like https://dart.dev/tools/diagnostic-messages, but now I'm not sure how tool-tied it is.

Staged:
https://dartlang-org-staging-0.firebaseapp.com/tools/non-promotion-reasons
https://dartlang-org-staging-0.firebaseapp.com/go/non-promo-write
https://dartlang-org-staging-0.firebaseapp.com/go/non-promo-property
https://dartlang-org-staging-0.firebaseapp.com/go/non-promo-this

/cc @bwilkerson @mit-mit @munificent @sfshaza2

### DIFF
--- a/src/tools/non-promotion-reasons.md
+++ b/src/tools/non-promotion-reasons.md
@@ -1,0 +1,45 @@
+---
+title: Fixing type promotion failures
+description: Solutions for cases where you know more about a field's type than Dart can determine.
+toc: false
+---
+
+{{ site.alert.warn}}
+  **This page is under construction**
+  ([issue #2940][]).
+{{ site.alert.end}}
+
+This page will have information to help you understand
+why type promotion failures occur,
+and give tips on how to fix them.
+For example, consider the following code:
+
+```dart
+class C {
+  int? i;                  // (1)
+  f() {
+    if (i == null) return;
+    i.isEven;              // (2)
+  }
+}
+```
+
+The Dart compiler produces an error message for (2)
+that points to (1) and explains that
+`i` can't be promoted to a non-nullable type
+because it's a field.
+The usual fix is to either use `i!`
+or create a local variable
+of type `int` that holds the value of `i`.
+
+Until this page is complete, see the following resources:
+
+* [Working with nullable fields][ns-fields]:
+  A section in [Understanding null safety][].
+* [Discussion in SDK issue #44900][sdk-44900-comment]:
+  A detailed list of reasons why type promotion might fail.
+
+[issue #2940]: https://github.com/dart-lang/site-www/issues/2940
+[sdk-44900-comment]: https://github.com/dart-lang/sdk/issues/44900#issuecomment-807117343
+[ns-fields]: /null-safety/understanding-null-safety#working-with-nullable-fields
+[Understanding null safety]: /null-safety/understanding-null-safety
--- a/firebase.json
+++ b/firebase.json
@@ -137,6 +137,9 @@
       { "source": "/go/experiments", "destination": "/tools/experiment-flags", "type": 301 },
       { "source": "/go/null-safety-migration", "destination": "/null-safety/migration-guide", "type": 301 },
       { "source": "/go/test-docs/:page*", "destination": "https://github.com/dart-lang/test/blob/master/pkgs/test/doc/:page*", "type": 301 },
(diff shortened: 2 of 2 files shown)


stereotype441 (approved):
Lgtm

stereotype441:
> Contributes to #2940 and [dart-lang/sdk#44900](https://github.com/dart-lang/sdk/issues/44900).
> 
> I'm not sure how to frame this page, so a few questions (which don't necessarily have to be answered definitively before we merge this PR):
> 
> * Is everything on this page going to be about promotion from nullable to non-nullable types? Or might we get some visits related to other forms of type promotion?

Long term the goal is to be able to tell users about non-promotion reasons for all kinds of failed promotions.  At the moment it's only working well for failed promotions related to nullability, but it's possible that the messages will occasionally appear for other kinds of  failed promotions.  (Null safety is built around a more general-purpose type promotion mechanism in the analyzer and CFE, so even though I focused on addressing null-related non-promotions first, it would have been difficult to restrict the feature to just null-related non-promotions). 

> * What should the title be? What's going to set expectations correctly?

How about something like: "Reasons type promotion might fail, and what to do about it"

> * Where on dart.dev should this page really live? Initially I thought it was like https://dart.dev/tools/diagnostic-messages, but now I'm not sure how tool-tied it is.

I don't have a strong opinion here; I'll defer to others.

bwilkerson:
> Where on dart.dev should this page really live? Initially I thought it was like https://dart.dev/tools/diagnostic-messages, but now I'm not sure how tool-tied it is.

The reason diagnostics seem tool related to me is that, while the specification says what conditions must be reported, it doesn't specify _how_ they are to be reported; that's up to the tools. The same is true for the context messages that Paul added that will point to these pages, so I think that viewing them as being tool related is fine.

kwalrath:
> > * Is everything on this page going to be about promotion from nullable to non-nullable types? Or might we get some visits related to other forms of type promotion?
> 
> Long term the goal is to be able to tell users about non-promotion reasons for all kinds of failed promotions. At the moment it's only working well for failed promotions related to nullability, but it's possible that the messages will occasionally appear for other kinds of failed promotions. (Null safety is built around a more general-purpose type promotion mechanism in the analyzer and CFE, so even though I focused on addressing null-related non-promotions first, it would have been difficult to restrict the feature to just null-related non-promotions).

Cool, that helps.

> > * What should the title be? What's going to set expectations correctly?
> 
> How about something like: "Reasons type promotion might fail, and what to do about it"

I like it, but that's really long. I don't think it'd fit on one line!

Are you OK with **Fixing type promotion failures** for now? I figure that devs' primary interest is in fixing issues, not necessarily understanding them (even if they might need to understand them to fix them). We could substitute a more positive word for _failures_, but if they're going to see the word _fail_ in the message anyway...

stereotype441:
> > > * What should the title be? What's going to set expectations correctly?
> > 
> > 
> > How about something like: "Reasons type promotion might fail, and what to do about it"
> 
> I like it, but that's really long. I don't think it'd fit on one line!
> 
> Are you OK with **Fixing type promotion failures** for now? I figure that devs' primary interest is in fixing issues, not necessarily understanding them (even if they might need to understand them to fix them). We could substitute a more positive word for _failures_, but if they're going to see the word _fail_ in the message anyway...

Works for me

src/tools/non-promotion-reasons.md
```diff
@@ -0,0 +1,45 @@
...
+toc: false
+---
+
+{{ site.alert.warn}}
+  **This page is under construction**
+  ([issue #2940][]).
+{{ site.alert.end}}
+
+This page will have information to help you understand
+why type promotion failures occur,
```
mit-mit, line 13: I'm a bit confused about the word 'failure' here. To me that sounds a bit like something that might occur when there is a bug in the analyzer... Is there a another word that might make it more clear that this isn't happening simply because it's something we're currently not capable of doing?

PR #3150, opened 2021-04-17
added documentation for dart pub add/remove

jp803333:
I am actually very new to open source and this will be my very first PR so any review, suggestion, advice or changes are welcomed.

So there was an [issue](https://github.com/flutter/flutter/issues/79646) on the flutter repository where someone said that documentation of `dart pub add` and `flutter pub add` are not available and the it was suggested to update the related websites.

I will try to add more documentation for other new sub command of  `dart pub`.

### DIFF
--- a/src/tools/pub/cmd/pub-add.md
+++ b/src/tools/pub/cmd/pub-add.md
@@ -12,21 +12,27 @@ $ dart pub add <package>[:<constraint>] [options]
 
 This command adds the specified package to the pubspec as a dependency,
 and then gets the dependency.
-
 For example, the following command is equivalent to
-editing `pubspec.yaml` (adding `http: ^0.12.1` under `dependencies`)
+editing `pubspec.yaml` to add the `http` package,
 and then calling `dart pub get`:
 
 ```terminal
-$ dart pub add http:^0.12.1
+$ dart pub add http
 ```
 
-By default, `dart pub add` gets the package from pub.dev.
-You can use options to specify other
-[dependency sources](/tools/pub/dependencies#dependency-sources) —
-Git repos, the Flutter SDK, alternate package servers, or
-a path within your local file system.
+By default, `dart pub add` uses the
+latest stable version of the package from pub.dev.
+For example, if 0.13.1 is the latest stable version of the `http` package,
+then `dart pub add http` adds
+`http: ^0.13.1` under `dependencies` in the pubspec.
 
+To add a [dev dependency][], use the `--dev` option:
+
+[dev dependency]: /tools/pub/dependencies#dev-dependencies
+
+```terminal
+$ dart pub add --dev test
+```
 
 ## Options
 
--- a/src/tools/pub/cmd/pub-add.md
+++ b/src/tools/pub/cmd/pub-add.md
@@ -1,0 +1,54 @@
+---
+title: dart pub add
+description: Use dart pub add to add dependecies to pubspec.yaml.
+toc: false
+---
+
+_Add_ is one of the commands of the [pub tool](/tools/pub/cmd).
+
+```
+$ dart pub add <package> [options]
+```
+Example:
+
+```
+$ dart pub add http:0.12.1
+```
+
+## Options
+
+For options that apply to all pub commands, see
+[Global options](/tools/pub/cmd#global-options).
+
+<dl>
+    <dt><code>-h, --help</code></dt>
+        <dd>Print this usage information.</dd>
+    <dt><code>-d, --dev</code></dt>
+        <dd>Adds package to the development dependencies instead.</dd>
+    <dt><code>--git-url</code></dt>
+        <dd>Git URL of the package.
+        </dd>
+        {% prettify nocode tag=pre+code %}
+        $ dart pub add package-name --git-url package-git-url
+        {% endprettify %} 
+    <dt><code>--git-ref</code></dt>
+        <dd>Git branch or commit to be retrieved</dd>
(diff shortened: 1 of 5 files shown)


jp803333:
@googlebot I signed it!

kwalrath:
Thank you, @jp803333!

jp803333:
@kwalrath I create a include file for all the pub subcommands and have include it in two places mentioned in [issue#3156](https://github.com/dart-lang/site-www/issues/3156). Can you check if i have done it correctly!?

kwalrath:
> @kwalrath I create a include file for all the pub subcommands and have include it in two places mentioned in [issue#3156](https://github.com/dart-lang/site-www/issues/3156). Can you check if i have done it correctly!?

Wow, I thought I'd have to do this change myself, but your change works great! Two things:

* We should probably just delete the old lists instead of commenting them out. (Not a big deal, but this would reduce file size slightly and reduce fgrep results — I seem to do a lot of grepping for command names.)
* I'm curious why you used HTML for the list formatting. Did you try using markdown? (It's been a while since I've done an include, and I don't remember if there were issues.)

kwalrath:
Staged:

https://kw-www-dartlang-1.firebaseapp.com/tools/pub/cmd#list-of-subcommands
https://kw-www-dartlang-1.firebaseapp.com/guides/packages#pub-subcommands
https://kw-www-dartlang-1.firebaseapp.com/tools/pub/cmd/pub-add
https://kw-www-dartlang-1.firebaseapp.com/tools/pub/cmd/pub-remove

The `dart pub remove` page needs changes similar to the ones I made to `dart pub add`. Do you want to make those changes, @jp803333, or would you like me to?

jp803333:
@kwalrath No, I want to do it. So I will change the list formatting from html to markdown, and remove the commented list from both places and for `dart pub remove` should I do required changes and send it you on discord for review before I commit all the changes or should I just commit them directly?

kwalrath (approved):
LGTM! But I'd like to have an engineer who works on pub review this, as well.

src/tools/pub/cmd/pub-add.md
```diff
@@ -0,0 +1,70 @@
...
+```
+$ dart pub add <package>[:<constraint>] [options]
+```
+
+This command adds the specified package to the pubspec as a dependency,
+and then gets the dependency.
+
+For example, the following command is equivalent to
+editing `pubspec.yaml` (adding `http: ^0.12.1` under `dependencies`)
+and then calling `dart pub get`:
```
jonasfj, line 18:
```suggestion
and then calling `dart pub get <package>`:
```

src/tools/pub/cmd/pub-add.md
```diff
@@ -0,0 +1,70 @@
...
+
+```terminal
+$ dart pub add http:^0.12.1
```
jonasfj, line 21: I would highlight `dart pub add http` as the primary example...

And then instead say that this is equivalent to adding `http: ^a.b.c` under dependencies, where `a.b.c` is the latest version of  package `http` that is compatible your other dependencies.

src/tools/pub/cmd/pub-add.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Options
+
+For options that apply to all pub commands, see
+[Global options](/tools/pub/cmd#global-options).
+
+<dl>
+    <dt><code>-d, --dev</code></dt>
+        <dd>Adds the package as a dev dependency,
+        instead of as a regular dependency.</dd>
```
jonasfj, line 39: You might want to high an example like `dart pub add --dev test` which adds `package:test` as a dev-dependency.

src/tools/pub/cmd/pub-add.md
jonasfj: Could we do an example `pubspec.yaml` and show before/after? Just an idea, maybe it's better to do it in a followup PR.

src/tools/pub/cmd/pub-add.md
kwalrath: I didn't know you could do this! (We don't document it in https://kw-www-dartlang-1.firebaseapp.com/tools/pub/cmd/pub-get, and I don't see it in the command-line help either.)

jonasfj: my bad... you can't do `dart pub get <package>`, only `dart pub add <package>`.

src/tools/pub/cmd/pub-add.md
jonasfj, line 21: I meant to suggest:
```suggestion
$ dart pub add http
```

src/tools/pub/cmd/pub-add.md
jp803333: Okay I will add this

src/tools/pub/cmd/pub-add.md
kwalrath: @jp803333 I'll handle this, since it's going to be mostly about finessing the wording.

src/tools/pub/cmd/pub-add.md
kwalrath: @jp803333 I'll handle this too, since I'll be in there anyway.

src/tools/pub/cmd/pub-add.md
kwalrath: I hesitate to show any code the user doesn't need to provide, because (in my experience) people who are skimming think they need to write that code.

PR #3154, opened 2021-04-19
Create `dart run` page

mit-mit:
Migrate `pub run` to `dart run`

Staged here: https://mit-staging.web.app/tools/dart-run


mit-mit:
@jonasfj can you please sanity check that this is actually how the tool still works?

kwalrath:
I plan to take a look after @jonasfj does

src/tools/dart-run.md
```diff
@@ -37,7 +44,7 @@ to the name of the script.
 For example, to run `foo.dart` in the `example/sub` directory:
 
 ```terminal
-$ pub run example/sub/foo arg1 arg2
+$ dart run example/sub/foo arg1 arg2
```
jonasfj, line 47: This won't work, `dart run example/sub/foo.dart` will work

src/tools/dart-run.md
```diff
@@ -47,17 +54,13 @@ in the pubspec, specify the package name.
 For example, to run `bar.dart` in the foo package:
```
jonasfj, line 54: I would rephrase this as something like: runs `bin/bar.dart` from `package:foo`.

jonasfj: TODO: we should also highlight the shorthand `dart run foo` is equivalent to `dart run foo:foo`, which means it runs `bin/foo.dart` from `package:foo`.

mit-mit:
Updated fairly substantially, PTAL @jonasfj

src/tools/dart-run.md
```diff
@@ -0,0 +1,74 @@
...
+```
+
+## Running a script in a dependency
+
+To run a script from the `bin` directory of a package that you depend on
+in the pubspec, specify the package name.
+For example, to run `bar.dart` in the `baz` package:
+
+```terminal
+$ dart run baz:bar [arg1 arg2 ...]
```
jonasfj, line 65: Here it might be nice to illustrate the shorthand again... This is usually where it's useful.

`dart test` is actually a shorthand for `dart run test` which is a shorthand for `dart run test:test`.

And this is actually the case where you're running a script in a dependency.

jonasfj (approved):
LGTM

src/tools/dart-run.md
```diff
@@ -0,0 +1,95 @@
...
+$ dart create myapp
+$ cd myapp
+$ dart run
+```
+
+## Running a program in a Dart file
+
+This is the simplest form, where you pass a relative path to a `.dart` file.
+
+To run a program in the file `internal/debug.dart`,
```
jonasfj, line 29: nit: `test/debug.dart` might be better... `internal/` is not a convention.

src/tools/dart-run.md
```diff
@@ -0,0 +1,95 @@
...
+which omits the package name and just specifies the program name,
+to run `bin/baz.dart`:
+
+```terminal
+$ dart run :baz [arg1 arg2 ...]
+```
+
+To run the main program -- the one which has the same name as the package --
+you can use these short forms,
+which will run `<current directory>/bin/<current directory>.dart`:
```
jonasfj, line 78: Actually, this is only if `<current directory> == "name" from pubspec.yaml`.

Name the of the current package is not the name of the current directory, it's the `name:` property from `./pubspec.yaml`.

jonasfj (approved):

src/tools/dart-run.md
mit-mit: done

src/tools/dart-run.md
mit-mit: clarified

mit-mit:
@kwalrath ready for you now

src/tools/dart-run.md
```diff
@@ -0,0 +1,96 @@
+---
+title: dart run
+deprogramion: Use dart run to run a Dart program in your package.
+---
+
+The `dart run` command supports running a Dart program located in a file,
+or a package, or in one of its dependencies, from the command line.
```
parlough, line 7: Perhaps the first "or" isn't needed here.

```suggestion
a package, or in one of its dependencies, from the command line.
```

src/tools/dart-run.md
```diff
@@ -0,0 +1,96 @@
+---
+title: dart run
+deprogramion: Use dart run to run a Dart program in your package.
```
parlough, line 3: I'm assuming you meant `description` here and from the content it seems we want to describe running programs in more than just "your" package.

```suggestion
description: Use dart run to run a Dart program.
```

src/tools/dart-run.md
```diff
@@ -0,0 +1,96 @@
...
+$ dart run :baz [arg1 arg2 ...]
+```
+
+To run the main program
+-- the one which has the same name as the package --
+use these short forms:
+
+```terminal
+$ dart run
+$ dart run [-- arg1 arg2 ...]
```
parlough, line 83: Is it necessary to outline both here? To me, these really seems like the same "short form".

src/tools/dart-run.md
```diff
@@ -0,0 +1,96 @@
...
+```
+
+To run a program in the current package but not in the `bin` directory,
+pass a relative path as discussed earlier:
+
+```terminal
+$ dart run tool/debug.dart [arg1 arg2 ...]
+```
+
+## Running with debugging
```
parlough, line 93: Since we're including some information about the debugging flags, should we mention the ability to enable experiments and perhaps link to https://dart.dev/tools/experiment-flags?

parlough (commented):
Also as a note, your stage version seems outdated.

kwalrath:
I did a fairly big edit, staged here:

https://kw-staging-dartlang-2.firebaseapp.com/tools/dart-run

Please take a careful look, @mit-mit and @jonasfj.

mit-mit:
@kwalrath are you still working on rewriting this?

kwalrath:
> @kwalrath are you still working on rewriting this?

Oops, looks like this slipped off my stack of things to do. I think I'll get to it today.

kwalrath:
> We lost the part here with the `--` separator before the args (which is needed when not passing the name of the entry point).

That's not working for me. I modified `bin/myapp.dart` (in the `myapp` package & directory) to be this:

```dart
void main(List<String> arguments) {
  print('Hello ${arguments.isEmpty ? 'World' : arguments[0]}!');
}
```

`dart run` works, and I expected `dart run -- Kathy` to work... but it doesn't. Instead, I need to use `dart run myapp Kathy`. Did I do something wrong?

```terminal
$ dart run
Hello World!
$ dart run -- Kathy
Could not find package `Kathy` or file `Kathy`
$ dart run myapp Kathy
Hello Kathy!
```

kwalrath:
ptal. I've restaged:

https://kw-staging-dartlang-2.firebaseapp.com/tools/dart-run

mit-mit:
> That's not working for me.

I just tried again, and it's not working for me either. So let's just keep what you have here!

kwalrath (approved):

PR #3155, opened 2021-04-19
Update typedef example comment

omarduarte:
### Context

The example in question is explaining how typedef works.

In the first  example we have:
``` 
  ...
  Function compare;
  ...
  SortedCollection(int f(Object a, Object b)) : compare = f;
  ...
  // Initial, broken implementation.
  int sort(Object a, Object b) => 0;
```

In the following example we have:
```
typedef Compare = int Function(Object a, Object b);
...
Compare compare;
...
// Initial, broken implementation.
int sort(Object a, Object b) => 0;
```

### Proposal

My proposed changes changes the comment of the typedef example from:

`// Initial, broken implementation.` to `// Now matching the defined function type.`


omarduarte:
@googlebot I fixed it.

PR #3157, opened 2021-04-20
Updating null safety codelab example

RedBrogdon:
The third exercise in the null safety codelab was assigning a value to `aNullableListOfStrings`, and as a result that variable wasn't functioning as it should in the exercise.

`aNullableListOfStrings` and `aListOfNullableStrings` are intended (at least back when I did the snippets) to teach people that both `List<String>?` and `List<String?>` are valid, but used for different things. If `aNullableListOfStrings` is assigned a value, you never need to change its type to `List<String>?` to make the analyzer happy, so that bit of learning opportunity is lost.


RedBrogdon:
Bah! Looks like I need to do this for real and update the associated example project. Closing this for now, and I'll redo it in a proper fork.

PR #3160, opened 2021-04-21
Remove reference to angel framework

kevmoo:
Hasn't been touched in 1+ years

Also cleanup gRPC link

### DIFF
--- a/src/server/libraries.md
+++ b/src/server/libraries.md
@@ -50,11 +50,10 @@ the packages listed in the previous table
 and [general-purpose packages][] such as `logging`:
 
 | **Package**                   | **Description** |
-| [angel_framework]({{site.pub-pkg}}/angel_framework) | A high-powered HTTP server with support for dependency injection & sophisticated routing. |
 | [crypto]({{site.pub-pkg}}/crypto) | Implements cryptographic hashing functions for algorithms such as SHA-1, SHA-256, MD5, and HMAC. |
-| [grpc]({{site.pub-pkg}}/grpc) | Implements [gRPC,][] a high performance, open source, general RPC framework that puts mobile and HTTP/2 first. |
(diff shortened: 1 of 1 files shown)


src/server/libraries.md
```diff
@@ -50,11 +50,10 @@ the packages listed in the previous table
 and [general-purpose packages][] such as `logging`:
 
 | **Package**                   | **Description** |
-| [angel_framework]({{site.pub-pkg}}/angel_framework) | A high-powered HTTP server with support for dependency injection & sophisticated routing. |
 | [crypto]({{site.pub-pkg}}/crypto) | Implements cryptographic hashing functions for algorithms such as SHA-1, SHA-256, MD5, and HMAC. |
-| [grpc]({{site.pub-pkg}}/grpc) | Implements [gRPC,][] a high performance, open source, general RPC framework that puts mobile and HTTP/2 first. |
+| [grpc]({{site.pub-pkg}}/grpc) | Implements [gRPC], a high performance, open source, general RPC framework that puts mobile and HTTP/2 first. |
```
parlough, line 54: You still need the extra bracket set after [gRPC] to enable it to resolve.

parlough (changes requested):

src/server/libraries.md
kevmoo: Fixed! Thanks!

parlough (approved):

kwalrath (approved):

src/server/libraries.md
kwalrath: Nice find, @parlough. Technically I think it would resolve without the bracket set, but we like the bracket set because it's an easy way to (1) indicate that it's an intentional link and (2) help us find malformed links in the rendered page (by searching for `][`). There might've been more reasons, but I don't recall what they were.

PR #3164, opened 2021-04-22
Update information about web libraries

sigmundch:
WebSQL's standard was dropped 5+ years ago and is unsupported on most major browsers except Chrome. We'll be deprecating it too, but we feel there is no need to highlight it here.

`js_util` is not quite deprecated, it is still the recommended API to workaround limitations in `dart:html` and `package:js`, so we want to make the labeling here more accurate.

### DIFF
--- a/src/_guides/libraries/index.md
+++ b/src/_guides/libraries/index.md
@@ -63,8 +63,7 @@ The following table lists the Dart core libraries that work on the
 | [`dart:web_audio`][dart-web_audio]      <br> High-fidelity audio programming in the browser. | |
 | [`dart:web_gl`][dart-web_gl]            <br> 3D programming in the browser. | |
 | [`dart:js`][dart-js]                    <br> _Don't use._ Instead, use the `js` package, as described in [JavaScript interoperability][]. | *DEPRECATED* |
-| [`dart:js_util`][dart-js_util]          <br> _Don't use._ Instead, use the `js` package, as described in [JavaScript interoperability][]. | *DEPRECATED* |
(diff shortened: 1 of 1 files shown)


kevmoo (approved):

mit-mit:
@sigmundch this table is expected to have entries for all core libraries here: https://dart.dev/guides/libraries#web-platform-libraries

Are you removing this from API docs also? Or removing `dart:web_sql` entirely?

sigmundch:
> Are you removing this from API docs also? 

Yes - I have CL out for that here: https://dart-review.googlesource.com/c/sdk/+/196461 that will be part of 2.14. 

> Or removing dart:web_sql entirely?

Also :), but we'd like to do so in 2.15 to give time to make an breaking change announcement and give users enough notice.

For what it is worth, I've confirmed that there is no relevant usage at the moment in packages for which we have visibility. That is, no uses in internal code, and also checked 15K packages with pub_crawl and only found 2 uses. Both are benign and I feel confident the breaking change wont be an issue there: 1 was from a discontinued package that we own, the other is from code automatically generated by mockito (which I believe would immediately go away if it were to be regenerated after the library is removed.)

sigmundch:
Since the deprecation and removal of dartdoc wont happen until 2.14, would you prefer that we readd the entry in this table until 2.14 hits stable?

mit-mit:
> Since the deprecation and removal of dartdoc wont happen until 2.14, would you prefer that we readd the entry in this table until 2.14 hits stable?

No, I sounds like we're already on "a glidepath" to having this be gone, that I feel totally fine having the table not mention it.

PR #3167, opened 2021-04-26
Use getters and setters in null safe codelab

parlough:
The thought here is to use a structure closer to idiomatic Dart when possible.

I kept the block structure for the setter as part of the codelab discusses commenting out its body, so this makes that easier to do.


kwalrath:
@RedBrogdon or @domesticmouse, would you please review?

domesticmouse (approved):

kwalrath (approved):

kwalrath:
I've confirmed that this works as expected in my locally staged version. Thank you, @parlough!

PR #3168, opened 2021-04-26
Update old Apple favicons so Dart favicon works on mobile

### LINKED ISSUE #3165: Why do we use the touch-icon dev logos?
In [head.html](https://github.com/dart-lang/site-www/blob/master/src/_includes/head.html) is there a reason we use the touch-icon- (found in [_assets/image](https://github.com/dart-lang/site-www/tree/master/src/_assets/image)) dev logos instead of the Dart logo?

This results in that icon showing up instead of the Dart logo in Google search on mobile.

image

Perhaps these just were never updated, not sure. I'd be happy to update/remove them, but wanted to make sure it wasn't intentional first.

parlough:
Fixes #3165


kwalrath:
I don't know how to test this, but I've staged it:

https://kw-staging-dartlang-2.firebaseapp.com

johnpryan:
Apple still [recommends having them](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/ConfiguringWebApplications/ConfiguringWebApplications.html) so it might be better to update the logo

parlough:
That's quite the website :P. Do we have these sizes anywhere or should I just generate them from the svg?

johnpryan:
We keep the latest assets in a [drive folder](https://drive.google.com/corp/drive/folders/1KXNtO9My2AMpDOF9A9Y_4aj4_BcgmDDT), but if the sizes are wrong we might need a [favicon generator](https://realfavicongenerator.net/).

parlough:
There correct sizes weren't present, so I generated the new pngs from the SVG.

kwalrath (approved):

kwalrath:
Wow, what a weird and tangled thing favicon size is! @parlough, can you point me to where you found the icon sizes to use?

I compared the icons in this PR to the sizes listed in [Apple's docs](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/ConfiguringWebApplications/ConfiguringWebApplications.html):

* touch-icon-iphone.png is 60x60 (docs don't say what size it should be, but other sources imply it should be 120x120)
* touch-icon-ipad.png is 76x76, but shouldn't it be 152x152? (copy over the current touch-icon-ipad-retina.png?)
* touch-icon-iphone-retina.png is 120x120, but shouldn't it be 180x180?
* touch-icon-ipad-retina.png is 152x152, but shouldn't it be 167x167?

Another page I looked at, which seems to agree with these sizes, is https://webhint.io/docs/user-guide/hints/hint-apple-touch-icons/. 

More info on the mess that is favicons:
* https://evilmartians.com/chronicles/how-to-favicon-in-2021-six-files-that-fit-most-needs 
* https://github.com/audreyfeldroy/favicon-cheat-sheet

parlough:
@kwalrath I just stuck with what we already had. After looking it up, it does seem they have updated the sizes over the years, so perhaps these sizes are from previous standards. 

I'll update the images and the HTML with the sizes you found. Thanks for looking into it further!

kwalrath (approved):
Thank you for going to all that trouble, @parlough! These LGTM.

johnpryan (approved):

PR #3171, opened 2021-04-26
Add tool to check if diagnostic files are updated

parlough:
The main goal of this CL is to introduce a check to see if the analyzer diagnostic message's file is updated and eventually the linter rules as well(https://github.com/dart-lang/site-www/pull/3064). 

This will run as a GitHub Action job that does not block the test or deploy jobs on a push to the master branch. This way it won't block contributors creating PRs but it will indicate to repository maintainers they need to update the local copy.

Since I was using null safety, I also updated the other tools in the directory to support null safety. The primary file of concern is `tool/diagnostic_update_check.dart`.

CC @bwilkerson @pq


tool/create_code_with_tooltips.dart
```diff
@@ -188,19 +190,24 @@ Iterable<List<String>> _tooltips(String toolTipDataPath) sync* {
...
 
     while (liClosingTag == null && lines.moveNext()) {
       final line = lines.current;
       final match = _tooltipLineEndRE.firstMatch(line);
       _log.info(
           '  ${match == null ? "Skipping data:" : "Processing data:"}: $line.');
-      tooltip = _join(tooltip, match[1]);
-      liClosingTag = match[2];
+      if (match == null) {
+        _log.info('Skipping data: $line');
```
bwilkerson, line 205: We're now duplicating the logged information. I'm guessing you intended to remove lines 202-203.

Also, although it was probably a bug that the previous logged info contained two colons, I'd want confirmation of that just in case it's important for some reason.

tool/diagnostic_update_check.dart
```diff
@@ -0,0 +1,41 @@
...
+
+Future<bool> checkAsset(final String url, final String path) async {
+  final remote = await http.get(Uri.parse(url));
+  if (remote.statusCode == 200) {
+    final local = File(path);
+    final remoteBytes = remote.bodyBytes;
+    final localBytes = local.readAsBytesSync();
+    if (remoteBytes.length == localBytes.length) {
+      final remoteHash = crypto.sha256.convert(remoteBytes).toString();
+      final localHash = crypto.sha256.convert(localBytes).toString();
```
bwilkerson, line 36: Is there any value in comparing the sha's rather than just comparing the contents directly?

tool/convert_book.dart
```diff
@@ -22,7 +22,7 @@ convertFile(String fileName) {
   print(fileName);
   var file = new File(fileName);
   var contents = file.readAsStringSync();
-  var title = new RegExp(r'<title>(.*)</title>').firstMatch(contents)[1];
+  var title = new RegExp(r'<title>(.*)</title>').firstMatch(contents)?[1];
```
bwilkerson, line 25: It might make more sense to migrate this using a null check (`!`). It matches the previous behavior and avoids needing to make `title` nullable.

bwilkerson (commented):
I'm not sure this is a positive change, but I don't maintain this repo. The downside is that changes made in the analyzer package to the markdown file will cause the next PR for this site to fail due to circumstances outside the control of the PR's author. That isn't very friendly. And while I generally like the idea that what's published can't get out of date, it does mean that we can't commit changes in analyzer that we're not yet ready to publish (such as for new language features). The latter probably isn't a significant issue, but I thought I'd point it out.

I'd also want @kwalrath to look over the changes to `diagnostic-messages.md` as I believe that she often proofreads them again while updating them.

kwalrath:
> The downside is that changes made in the analyzer package to the markdown file will cause the next PR for this site to fail due to circumstances outside the control of the PR's author.

Agreed that we shouldn't fail unrelated PRs. I do, however, appreciate knowing that we might need to update the diagnostics page.

> I'd also want @kwalrath to look over the changes to `diagnostic-messages.md` as I believe that she often proofreads them again while updating them.

Yep, I do a spot check. :) It's mostly to make sure the formatting isn't broken, but sometimes I'll notice something else — like the title case section head (Mixin Application) that I'd love to convert to sentence case.

tool/convert_book.dart
kwalrath: This script is very old, and I don't find any references to it. Delete it?

parlough:
I agree failing PRs is likely not the best solution. I believe GitHub actions allows steps to surface as warnings and not as required for the PR to merge, do you think that would be a fine solution? @kwalrath 

@bwilkerson Do beta/dev SDKs surface links to in progress/unreleased diagnostics? Is there a concern of users using dev SDKs navigating to dead links? That was my motivation for making sure this file stays updated. Perhaps we can add some sort of documentation to the page that indicates if a diagnostic is experimental?

tool/diagnostic_update_check.dart
parlough: I was just following the `pub-dev` implementation for their comparison checks, but after some basic benchmarking, comparing the contents is faster and more straightforward anyway. Thanks for the suggestion!

tool/convert_book.dart
parlough: Sounds good to me, deleted!

tool/create_code_with_tooltips.dart
parlough: Yeah it was a bug as far as I can tell, we don't use the output for anything besides logging purposes.

bwilkerson:
> Do beta/dev SDKs surface links to in progress/unreleased diagnostics? Is there a concern of users using dev SDKs navigating to dead links?

No, and no. After new documentation has been reviewed it's added to the source code in `sdk` and a new file is generated into `analyzer`. After the generated file been copied over and published, then we manually mark each of the documented error codes as having published docs, and only after that's happened do we start returning a URL with the diagnostic. So while the process is manual, I think it's fairly error-proof.

We do have to be careful in the other direction. If we remove an error code then older versions of the SDK could produce a dead link. To solve that problem we manually add a section to the end of the generated docs that the obsolete links can land on.

> Perhaps we can add some sort of documentation to the page that indicates if a diagnostic is experimental?

The more I think about it the more convinced I am that it isn't a problem. We won't start adding error codes for new language features until there's an experiment flag defined for it, and if there's a flag then a user could enable the experiment. If they run into a diagnostic that has documentation for it, then I suppose that's actually a positive outcome.

kwalrath:
> I agree failing PRs is likely not the best solution. I believe GitHub actions allows steps to surface as warnings and not required for a PR to merge, do you think that would be a fine solution? @kwalrath

Yep, that works for me. Ideally I'd notice (or, better yet, have a tool that somehow informs us) when an OK-to-fail test has newly switched from passing to failing.

kwalrath (commented):
I staged this locally, and the changes to the diagnostic-messages.md file lgtm (once you delete the test :) except for one case of excessive capitalization.

/cc @bwilkerson

src/tools/diagnostic-messages.md
```diff
@@ -154,6 +156,40 @@ For additional details, see the
 
 [definiteAssignmentSpec](https://github.com/dart-lang/language/blob/master/resources/type-system/flow-analysis.md)
 
+### Mixin Application
```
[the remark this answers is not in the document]
bwilkerson: I'll get that updated on the analyzer side tomorrow. If you want, we can go ahead and land this in the meantime even though there will be a warning.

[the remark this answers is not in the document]
parlough: Thanks! I'll hold off until it gets landed on the analyzer side.

Unfortunately GitHub actions does not seem to have a way of indicating a warning/okay failure as far as I can tell, so I am going to investigate other avenues as well. Perhaps a PR being created with the updated file, not sure yet.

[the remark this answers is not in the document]
kwalrath: @parlough we use `experimental: true` in https://github.com/dart-lang/site-www/blob/master/.github/workflows/build.yml to make a test failure OK.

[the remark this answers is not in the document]
parlough: @kwalrath I considered that but then I believe then there's no indication in the PR UI that anything failed, limiting the usefulness :(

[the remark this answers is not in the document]
kwalrath: > To implement the issue/PR idea there would need to be a GITHUB token for an account to create it and is likely more complicated. Perhaps this check should just run when pushing to master, so it doesn't block PRs but is easy to identify? It can be a separate job so it doesn't block deploy as well. How does that sound?

I like that idea!

.github/workflows/build.yml
```diff
@@ -78,6 +78,19 @@ jobs:
       - run: ${{ matrix.task }}
         continue-on-error: ${{ matrix.experimental }}
 
+  check:
```
kwalrath, line 81: "check" is vague (people might think it means a linkcheck, e.g.), so it's a little unnerving to see it be skipped in the Actions tab. Could we be more specific? I don't know the conventions for names, but if length and hyphens are OK, one possibility is something like sdk-files-are-up-to-date.

bwilkerson:
https://dart-review.googlesource.com/c/sdk/+/197101 has landed and should fix the inconsistency issue.

parlough:
@bwilkerson Thanks for your help and feedback :) I've gone ahead and updated the included file.

.github/workflows/build.yml
```diff
@@ -69,7 +69,7 @@ jobs:
         with:
           sdk: ${{ matrix.sdk }}
       - name: Install node dependencies
-        # TOOD: caching requires a lockfile for node deps, so does npm ci.
+        # TODO: caching requires a lockfile for node deps, so does npm ci.
```
kwalrath, line 72: 😆

kwalrath (approved):
LGTM!

PR #3172, opened 2021-04-27
Added example for 'version'

yosus:
Added example for 'version' syntax.


kwalrath:
Thanks for the PR, but this isn't the place for an example. This section just introduces the next two sections, which have the examples.

PR #3179, opened 2021-04-29
Update Google AdWords to Google Ads

parlough:
My understanding is Google AdWords has been completely renamed to Google Ads:

https://en.wikipedia.org/wiki/Google_Ads


parlough:
We could likely add some more consumer facing examples here, such as Google Pay and Stadia, but that can be in a follow-up change.

kwalrath:
This seems low risk, so I'm going to commit it. I'll create an issue to update the list of deployments.

kwalrath (approved):

PR #3181, opened 2021-04-30
Improve OS ordering and update mac to macOS in archive

### LINKED ISSUE #2046: SDK Archive sorting is weird
When I go to the [Dart SDK Archive page](https://dart.dev/tools/sdk/archive) to download a .zip file, for Linux I see:

image

1. The "32-bit" and "64-bit" entries don't specify what CPU architecture they're for.  They should be labeled "ia32" and "x64" instead (or "x86" and "x64", but "ia32" matches the filename).

2. The "Bits" column is misnamed. (It probably made more sense before ARM versions were added.) "Architecture"/"Arch." would be better.

3. The sort order seems odd.  I expect that most people want the ia32/x64 versions instead of the ARM ones, yet the ARM ones are listed first.  64-bit versions probably should be favored over 32-bit ones too.

(This also would be consistent with the "Download URLs" section, which refers to "`<architecture>`" in the URL format.)

### LINKED ISSUE #3137: Mac -> macOS

parlough:
Orders 64-bit downloads before 32-bit ones and updates Mac references to macOS in the archive.

Also updates some dependencies and minor doc changes.

Fixes #2046 and Closes #3137

### DIFF
--- a/src/tools/sdk/_linux.md
+++ b/src/tools/sdk/_linux.md
@@ -24,7 +24,7 @@ $ sudo apt-get install dart
 
 #### Install a Debian package
 
-Alternatively, download Dart SDK [as Debian package](#){:.debian-link-stable}
+Alternatively, download Dart SDK [as a Debian package](#){:.debian-link-stable}
 in the `.deb` package format. 
 
 #### Modify PATH for access to all Dart binaries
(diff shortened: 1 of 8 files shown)


domesticmouse (approved):

kwalrath (approved):
Nice. Thanks, @parlough!

PR #3195, opened 2021-05-04
Update nullable type parameter code example

parlough:
Original motivation from https://github.com/dart-lang/site-www/pull/3157:

> The third exercise in the null safety codelab was assigning a value to `aNullableListOfStrings`, and as a result that variable wasn't functioning as it should in the exercise.  
>
> `aNullableListOfStrings` and `aListOfNullableStrings` are intended (at least back when I did the snippets) to teach people that both `List<String>?` and `List<String?>` are valid, but used for different things. If `aNullableListOfStrings` is assigned a value, you never need to change its type to `List<String>?` to make the analyzer happy, so that bit of learning opportunity is lost.

@RedBrogdon I believe this is what you were going for. The correct example also had the nullability suffix in the wrong place, so I adjusted that as well.


RedBrogdon (approved):
Yep, that's it. LGTM.

PR #3200, opened 2021-05-05
Fix the xOrigin and yOrigin const declaration location

### LINKED ISSUE #3197: Language tour: confusing xOrigin yOrigin in named constructor
The named constructor .origin uses xOrigin and yOrigin, but it's not clear that those are a reference to an earlier example defining them as const for the class.  Perhaps the const definitions can be repeated?

parlough:
The organization of these values caused confusion as they were used in an early section but then declared in a later section that didn't even use them.

Fixes #3197


kwalrath (approved):

kwalrath:
Thanks, @parlough!

PR #3207, opened 2021-05-05
Clarification in example for 'Type test operators' section

GGurbanov:
Clarified example `obj is Object` when `null` is assigned to `obj`

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -1899,7 +1899,7 @@ runtime.
 {:.table .table-striped}
 
 The result of `obj is T` is true if `obj` implements the interface
-specified by `T`. For example, `obj is Object` is always true.
+specified by `T`. For example, `obj is Object` is always true (if `obj != null`).
 
 Use the `as` operator to cast an object to a particular type if and only if
 you are sure that the object is of that type. Example:


parlough:
Thanks for creating a pull request for this @GGurbanov, great catch! I wonder if instead of adding the non-null clarification though, we should instead change it to the following to make it more straightforward. Especially since the analyzer can consistently report this as being unnecessary code and always evaluating to `true`.

> For example, `obj is Object?` is always true.

Let me know what you think :)

GGurbanov:
> Thanks for creating a pull request for this @GGurbanov, great catch! I wonder if instead of adding the non-null clarification though, we should instead change it to the following to keep it simple. Especially since the analyzer can consistently report this check as being unnecessary code and always evaluating to `true`.
> 
> > For example, `obj is Object?` is always true.
> 
> Let me know what you think :)

Agree!)

src/_guides/language/language-tour.md
```diff
@@ -1899,7 +1899,7 @@ runtime.
 {:.table .table-striped}
 
 The result of `obj is T` is true if `obj` implements the interface
-specified by `T`. For example, `obj is Object` is always true.
+specified by `T`. For example, `obj is Object` is always true (if `obj != null`).
```
parlough, line 1902:
```suggestion
specified by `T`. For example, `obj is Object?` is always true.
```

kwalrath (approved):

PR #1324, opened 2019-01-23
Replace absolute links with macros

sfshaza2:
Please review, @kwalrath.

(I ran the checklink script until it was clear.)

PR #3212, opened 2021-05-06
Fix links for site.pub-pkg and braces

micpap25:
A response to #1324

Changes all instances of `{{site.pub}}/packages` to `{{site.pub-pkg}}`.

In addition, changes all double-curly-brace notation around singular objects in `src` so they have no space inside them. Changing multiple-object double-curly-brace notations could be an option in the future. 

This is all meant to clean up the backend and promote consistent style in the future.

### DIFF
--- a/src/_articles/archive/zones.md
+++ b/src/_articles/archive/zones.md
@@ -750,7 +750,7 @@ stack_trace
 : With the stack_trace library's
   [Chain class]({{site.pub-api}}/stack_trace/latest/stack_trace/Chain-class.html)
   you can get better stack traces for asynchronously executed code.
-  See the [stack_trace package]({{site.pub}}/packages/stack_trace)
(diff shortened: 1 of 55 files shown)


parlough:
@micpap25 Thanks for working on this!

One question, was there any particular reason or comment you went with removing the spaces in the Liquid tags? It seems the Liquid documentation and a lot of popular Jekyll sites use the spacing in the tags, so I wonder if we should standardize on having the spaces to make contribution more standard for those that have used Liquid templates before.

micpap25:
Hi @parlough , 
My goal was to standardize the use of Liquid tags to one style, and within the site, Liquid tags without any pipes more commonly did not have any spaces. Although it's pretty low-priority, maybe that's a good discussion to be had. 
For now, I'll edit this PR so it only changes the `site.pub` issue. I'll open a new issue for the spacing.

PR #3217, opened 2021-05-07
Update firebase.json

devoncarew:
- add a go link for a description page for the new dart core (and recommended) lints

Currently this points to the package:lints repo; in the future this might instead point into dart.dev. This doc link will be references from the analysis_options.yaml file that `daft create` generates for all new dart projects.

@pq @kwalrath


devoncarew:
(also, I sorted the go links)

parlough (approved):

PR #3233, opened 2021-05-13
Change nullable to non-nullable to fix logic

### LINKED ISSUE #3161: 'Understanding null safety' page issue

Description of issue:

In the first paragraph of [Required named parameters](https://dart.dev/null-safety/understanding-null-safety#required-named-parameters) section, I think this sentence 

>  What if you want to have a named parameter with a _**nullable**_ type and no default value? That would imply that you want to require the caller to always pass it. In other words, you want a parameter that is named but not optional.

should be this. (nullable -> non-nullable)

> What if you want to have a named parameter with a _**non-nullable**_ type and no default value? That would imply that you want to require the caller to always pass it. In other words, you want a parameter that is named but not optional.

parlough:
Fixes #3161

### DIFF
--- a/src/null-safety/understanding-null-safety/index.md
+++ b/src/null-safety/understanding-null-safety/index.md
@@ -1088,10 +1088,10 @@ local lazy evaluation.
 
 To guarantee that you never see a `null` parameter with a non-nullable type, the
 type checker requires all optional parameters to either have a nullable type or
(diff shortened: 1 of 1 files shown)


kwalrath:
@munificent could you please review this?

munificent (approved):
Oops, thank you!

PR #3234, opened 2021-05-13
Remove now unnecessary stagehand references

### LINKED ISSUE #3101: [Next Release] Update references to and documentation around Stagehand
Various references will be need to be removed, updated, or added to with the transition to what was stagehand just being part of `dart create` as the explicit references to stagehand won't be so relevant or necessary anymore.

https://github.com/dart-lang/sdk/commit/f6ce7a829f288267fb9720a30346ff4c086001ab

parlough:
Leaves the mention in `src/tools/dart-tool.md` as that is used to indicate to people who used `stagehand` before that `dart create` is what should be used going forward.

Fixes #3101


kwalrath (approved):
Thank you, @parlough!

PR #3237, opened 2021-05-13
Migrate dart:html library docs to null safety

parlough:
I'm not sure the best way to migrate some web usage after null safety. In a lot of cases `!` is the equivalent and makes sense, since you generally know what is in your document. I think I found an okay balance, but am open to suggestions.

I'm thinking we can add a note explaining when it is appropriate to use the null assertion operator, the conditional access operator, and when you should check for null. I'm not yet sure what best practices to suggest though, nor where this should go.


kwalrath:
@munificent might be able to suggest some guidelines (or point to someone who can) for using null safety with web APIs.

domesticmouse (approved):

domesticmouse:
I agree with your reasoning @parlough. Interfaces between different systems (in this case the DOM and Dart's type system) always have some level of impedence mismatch.

kwalrath (approved):
Great work, @parlough!

PR #3245, opened 2021-05-16
Migrate Creating Streams article to support null safety

parlough:
Should be in a good state, but not completely confident on my usages of `late` here, so open to heavy feedback there.

The GitHub links will be dead as these changes are not yet committed.


null_safety_examples/misc/lib/articles/creating-streams/stream_controller.dart
```diff
@@ -84,7 +82,7 @@ void showBasicUsage() {
 void demoPause() {
   var counterStream =
       Stream<int>.periodic(Duration(seconds: 1), (x) => x).take(15);
-  StreamSubscription<int> subscription;
+  late StreamSubscription<int> subscription;
```
domesticmouse, line 85: Is there any reason not to merge the declaration and initialisation of this variable? It'd allow skipping the late declaration

null_safety_examples/misc/lib/articles/creating-streams/stream_controller_bad.dart
```diff
@@ -49,7 +49,7 @@ void listenAfterDelay() async {
 // #docregion pause-problem
 void listenWithPause() {
   var counterStream = timedCounter(const Duration(seconds: 1), 15);
-  StreamSubscription<int> subscription;
+  late StreamSubscription<int> subscription;
```
domesticmouse, line 52: Ditto

domesticmouse (approved):

null_safety_examples/misc/lib/articles/creating-streams/stream_controller.dart
parlough: These were the usages I wasn't completely sure about but to get the original samples working I found the early declaration with `late` is needed as the `subscription` variable is used in the callback. Do you think there is perhaps a different structure that would work better?

domesticmouse: Actually, if it is being intialised in a callback, that's pretty much `late` by definition. SGTM

src/_articles/libraries/code/stream_controller_bad.dart
```diff
@@ -1,62 +0,0 @@
-// ignore_for_file: cancel_subscriptions
-import 'dart:async';
-
-// #docregion flawed-stream
-// NOTE: This implementation is FLAWED!
```
domesticmouse, line 5: Don't we still need a bad stream controller for documentation purposes?

parlough: This file was used to link to a file from the article, but it is just a duplicate of the other `stream_controller_bad` file which can cause issues for keeping them in sync.

Like some of our other, newer documentation, instead of linking directly to the file(which opens the raw file in the browser), I've modified this to link to the original file in GitHub.

kwalrath (approved):

kwalrath:
Thank you, @parlough!

PR #3247, opened 2021-05-17
Fixed provably typo to probably

JasonJamal:


JasonJamal:
@googlebot I signed it!

mit-mit:
I believe this is actually intended; "provable" as in "can be proved", right @munificent ?

mit-mit:
But I also seem to recall we got this feedback before, so perhaps better to rewrite?

munificent:
Yes, I intended "provably" here. :)

PR #3253, opened 2021-05-17
Add details for non-promotion causes/workarounds

### LINKED ISSUE #2940: Add promotion info/workaround to /null-safety

kwalrath:
Fixes #2940

Staged: https://kw-staging-dartlang-2.firebaseapp.com/tools/non-promotion-reasons


kwalrath:
@stereotype441 here's an early draft of how we might approach the no-promo page. The first case — only promoting local variables — seems like it might be the most common one by far, so I thought it might deserve its own section. Plus it seemed like a good example to feature (since we already did). But if people will always be dropping into individual sections, that doesn't make as much sense.

If people will mostly be dropping into individual sections, and if we don't have workarounds for each section, then we'll need a link back up to the common workaround section (https://kw-staging-dartlang-2.firebaseapp.com/tools/non-promotion-reasons#other-causes-and-workarounds).

Anyway, let me know what you think. I didn't want to get too far into this before talking with you.

/cc @bwilkerson

src/tools/non-promotion-reasons.md
```diff
@@ -1,45 +1,380 @@
...
+
+{:.bad}
+{% prettify dart tag=pre+code %}
 class C {
   int? i;                  // (1)
-  f() {
+  void f() {
     if (i == null) return;
-    i.isEven;              // (2)
+    i.isEven;              // (2) ERROR
```
stereotype441, line 32: I'm guessing you got this example from me, because I write a lot test cases using `int` and `isEven` like this.  I wonder if it would be better to try to come up with less contrived examples since this is user-facing documentation.  I'm spitballing here, but maybe something like this?

```dart
class Order {
  DateTime? shipDate; // (1)
  void checkForShipmentDelay() {
    if (shipDate == null) return;
    if (DateTime.now().difference(shipDate) > Duration(days: 7)) { // (2) ERROR
      print('Shipment is late');
    }
  }
}
```

Unfortunately, I don't really have good ideas for making the other examples less contrived (they are less common than field promotion, so it's harder to invent concrete examples of where they might occur).  Maybe it's better to stick with `int` and `isEven` for now, and wait and see if anyone complains?  I don't know.

At the very least, consider changing this line to `print(i.isEven);` so that it doesn't just look like a statement with no effect 😃

src/tools/non-promotion-reasons.md
```diff
@@ -1,45 +1,380 @@
...
+
+Example:
+
+{:.bad}
+{% prettify dart tag=pre+code %}
+f(int? i, int? j) {
+  if (i == null) return;
+  try {
+    i = j;                 // (1)
+    if (i == null) return; // (2)
```
stereotype441, line 215: Consider adding some placeholder text in the try block so that it looks more plausible that an exception might actually occur, e.g.:

```dart
try {
  i = j; // (1)
  ... other logic ...
  if (i == null) return; // (2)
  ... other logic ...
} catch (...) {
  print(i.isEven); // ERROR
}
```

src/tools/non-promotion-reasons.md
```diff
@@ -1,45 +1,380 @@
...
+{% endprettify %}
+
+In this example, `o` is promoted to `Comparable` at (1), but
+it's not promoted to `Pattern` at (2),
+because `Pattern` is not a subtype of `Comparable`.
+(The rationale is that if we did promote,
+then you wouldn't be able to access methods on `Comparable` anymore).
+Note that just because `Pattern` isn't a subtype of `Comparable`
+doesn't mean the code at (3) is dead;
+if you define a class that
```
stereotype441, line 266: Aha!  I just realized that the standard library already has a class that implements both `Comparable` and `Pattern`: the `String` class!  So if you want you could change this sentence to something like:

> Note that just because `Pattern` isn't a subtype of `Comparable` doesn't mean the code at (3) is dead; `o` could be a `String` (which implements both `Comparable` and `Pattern`).

stereotype441 (approved):
I like where you're going with this.  I added a few comments on details, but nothing really major.

src/tools/non-promotion-reasons.md
kwalrath: Using print() would be nice. :) I do like keeping it simple, and contrived is fine as long as it helps readability/understandability.

kwalrath:
What's your thinking on how people will reach this page? Do you expect it to mostly be through anchors (sections) or mostly to the top of the page? If we're unsure, I should probably link back up from each section to the common workaround section.

src/tools/non-promotion-reasons.md
kwalrath: I just noticed a weird URL when I put the code in DartPad:

image

kwalrath: (It should be http://dart.dev/go/non-promo-property.)

parlough: This renders properly for me in WebStorm, so perhaps it's an issue with how DartPad sends or displays the diagnostic message.

Seems like it's received liked that from dart-services:
image

cc @domesticmouse

kwalrath:
ptal @stereotype441. Can we give more guidance on how to fix each of these issues?

kwalrath:
I've updated the staging server:

https://kw-staging-dartlang-2.firebaseapp.com/tools/non-promotion-reasons
https://kw-staging-dartlang-2.firebaseapp.com/go/non-promo-property
https://kw-staging-dartlang-2.firebaseapp.com/go/non-promo-this
https://kw-staging-dartlang-2.firebaseapp.com/go/non-promo-write

src/tools/non-promotion-reasons.md
```diff
@@ -1,45 +1,399 @@
...
+}
+{% endprettify %}
+
+This example features an instance field,
+but it could instead use an instance getter, a static field or getter,
+a top-level variable or getter, or `this`.
+(Although promoting `this` would be sound,
+implementing it would be difficult and not very useful.)
+
+{{site.alert.tip}}
```
parlough, line 77: This tip fits in well here, I like it!

src/tools/non-promotion-reasons.md
```diff
@@ -1,45 +1,399 @@
...
+{% endprettify %}
+
+In this case, a human can see that the promotion is safe because
+the only write to `i` uses a non-null value and
+happens before `foo` is ever created.
+But [flow analysis isn't that smart][1536].
+
+[1536]: https://github.com/dart-lang/language/issues/1536
+
+***[PENDING: what's the fix?]***
```
stereotype441, line 366: The best fix for this "particularly nasty" case is probably to make a fresh variable, e.g.:
```dart
void f(int? i) {
  var j = i ?? 0;
  var foo = () {
    print(j.isEven); // OK
  };
}
```
This works because `i ?? 0` has type non-nullable `int`; this will be used as the inferred type of `j`, so regardless of whether there's an assignment to `j` later, it can never take on a non-null value.

src/tools/non-promotion-reasons.md
```diff
@@ -1,45 +1,399 @@
...
+}
+{% endprettify %}
+
+Flow analysis reasons that there's no way of telling
+what order `foo` and `bar` might be executed in;
+in fact, `bar` might even get executed halfway through executing `foo`
+(due to `foo` calling something that calls `bar`).
+So it isn't safe to promote `i` at all inside `foo`.
+
+***[PENDING: what's the fix?]***
```
stereotype441, line 398: The best bet here is probably to do a separate variable, e.g.:
```dart
void f(int? i, int? j) {
  var foo = () {
    var i2 = i;
    if (i2 == null) return;
    print(i2.isEven); // OK; i2 can be promoted because it's local to this closure
  };
  var bar = () {
    i = j;
  };
}
```

stereotype441 (approved):
> ptal @stereotype441. Can we give more guidance on how to fix each of the problems (see my PENDING comments)?

Done!  I'm going to go ahead and mark this "approved" so that you don't have to wait for me if you want to land it as is, and iterate to make improvements later.  Or we can continue the conversation in this PR until it's fully polished - your call 😃

kwalrath:
Great examples, @stereotype441. Thank you!

kwalrath:
I've incorporated the examples but want to make another pass over it before asking for another review.

kwalrath:
@stereotype441 I think we might be close to something I'd publish, but I'd like to figure out a few things:

* The right word(s) to use for "separate variable" / "fresh variable". I wonder if the word "copy" or "local" should be in there.
* Clarification on what makes `!` or `as` redundant.
* (could be later) The scope of this page, and when type promotion can and cannot occur. This is related to a discussion I had with @RedBrogdon today, regarding a question that @leafpetersen answered about type promotion via definite assignment.

In the meantime, I'll work on making the good examples be auto-included. We're close!

src/tools/non-promotion-reasons.md
```diff
@@ -1,45 +1,601 @@
...
+{% endprettify %}
+
+Flow analysis reasons that there's no way to determine
+when `foo` might get called,
+so it might get called after the assignment at (2),
+and thus the promotion might no longer be valid.
+As with loops, this demotion happens regardless of the type of
+the right hand side of the assignment.
+
+A solution is to use a fresh variable:
```
bwilkerson, line 505: I'll propose uniformly using a phase such as 'create a new variable', 'add a new variable', or even 'define a new variable' when a new variable is being introduced to solve the problem. For example: 'A solution is to create a new variable:'. Adding 'local' would also be reasonable: 'A solution is to create a new local variable:'.

bwilkerson (commented):
> Clarification on what makes ! or as redundant.

It's redundant in the sense that the condition has already been checked but has to be checked again because inference wasn't able to use the result of the previous check.

src/tools/non-promotion-reasons.md
kwalrath: I like the more specific one: "create a new local variable". Thanks, @bwilkerson!

kwalrath: Upon further thought, the "new" seems redundant, so I'm thinking "create a local variable".

bwilkerson: I agree. sgtm

kwalrath:
I've created a .dart file to hold all the good examples, but I don't think the code's actually being analyzed. I'll look at that when I get back to this (Monday, I think).

The staging server is up-to-date:

https://kw-staging-dartlang-2.firebaseapp.com/tools/non-promotion-reasons

src/tools/non-promotion-reasons.md
```diff
@@ -1,45 +1,615 @@
...
+      print(j.isEven);
+      j = k;
+      continue label;
+  }
+}
+{% endprettify %}
+
+
+
+### In catch after possible write in try
```
parlough, line 259: Perhaps another simplified anchor may be useful in this case as well.

parlough (approved):
Reads great  to me!

src/tools/non-promotion-reasons.md
kwalrath: Good catch, I'll do that before merging.

Issue #3254, opened 2021-05-18
Update images on 'Fetch data dynamically' page


parlough:
Page URL: https://dart.dev/tutorials/web/fetch-data
Page source: https://github.com/dart-lang/site-www/tree/master/src/_tutorials/web/fetch-data.md

The images on the page use the old `dartlang.org` link, `http` instead of `https`, can be blurry on some screen sizes, and could use a style to match better with the rest of the website. We could create SVGs using a more user accessible tool such as https://www.diagrams.net/

kwalrath:
@parlough what do you think of adding **help wanted** and **beginner** tags to this?

PR #3255, opened 2021-05-18
Migrate fetch data tutorial to null safety

parlough:


domesticmouse (approved):

kwalrath (approved):
I found a couple of minor issues:

* In the text it says `showJson`, but the code now uses `_showJson`
* The figures say `http://www.dartlang.org` instead of `https://dart.dev` (although I'm pretty sure the source code has been changed to use dart.dev)

If you want to file an issue for the figures, we could fix them in a future PR.

parlough:
@kwalrath I've updated the text to `_showJson` and I previously filed https://github.com/dart-lang/site-www/issues/3254 to update the figures.

kwalrath:
> @kwalrath I've updated the text to `_showJson` and I previously filed #3254 to update the figures.

Ah right, I forgot about that.

PR #3256, opened 2021-05-18
Update some language tour dartpad embeds

parlough:


null_safety_examples/misc/lib/language_tour/callable_classes.dart
```diff
@@ -7,4 +7,4 @@ class WannabeFunction {
 var wf = WannabeFunction();
 var out = wf('Hi', 'there,', 'gang');
 
-main() => print(out);
+void main() => print(out);
```
kwalrath, line 10: If we do this, can we remove the ignore_for_file comment?

src/_guides/language/language-tour.md
```diff
@@ -1553,25 +1540,22 @@ prints a string that includes the value at the specified index.
...
-{{site.dartpad}}/5d70bc1889d055c7a18d35d77874af88
-{% endcomment %}
-
-<iframe
-src="{{site.dartpad-embed-inline}}?id=5d70bc1889d055c7a18d35d77874af88&split=60&ga_id=anonymous_functions"
-    width="100%"
-    height="400px"
-    style="border: 1px solid #ccc;">
-</iframe>
+```dart:run-dartpad:height-400px:ga_id-anonymous_functions:null_safety-true
```
kwalrath, line 1551: Can we add a code-excerpt element above here?

kwalrath (approved):
The changes look good, thank you! I think we could tweak a couple of things.

null_safety_examples/misc/lib/language_tour/callable_classes.dart
parlough: This is also used for the `wf` and `out` as they are considered public APIs. I can remove the ignore and type annotate those if desired though.

src/_guides/language/language-tour.md
parlough: Good idea!

null_safety_examples/misc/lib/language_tour/callable_classes.dart
kwalrath: Nope, that's fine. Never mind. :)

kwalrath:
Would the 3 unusual build failures be fixed by running tool/dartformat.sh? (I didn't look too closely)

parlough:
@kwalrath Yes! Thanks for pointing that out. I messed up the formatting when I fixed the merge conflict on GitHub. Should be good now.

PR #3259, opened 2021-05-18
Add Docker on /get-dart

mit-mit:


kwalrath (approved):
I promised an engineer that I'd keep the zipfile link last in the paragraph. So unless you want engineers to use Docker images instead, we should keep that promise. (If we do want to change the order, we should update/remove the "NOTE to editors".)

kevmoo (approved):
:shipit:

mit-mit:
> I'd keep the zipfile link last in the paragraph

SGTM

kwalrath:
Actions seem to be frozen... I'll try to kick it off again.

mit-mit:
GitHub Actions is having general issues: https://www.githubstatus.com/

```
Update - We have confirmed that run starts are delayed for a number of customers. We have identified the issue and are working on a fix.
May 18, 11:06 UTC
Investigating - We are investigating reports of degraded performance for GitHub Actions.
May 18, 08:45 UTC
```

PR #3262, opened 2021-05-18
Remove the banner

kwalrath:
Don't worry, removing the banner doesn't mess up dart.dev the way it does flutter.dev :) (as far as I know)


sfshaza2 (approved):
That is correct. LGTM!

parlough (approved):

Issue #3235, opened 2021-05-13
Switch examples from pedantic to recommended


parlough:
With the new `recommended` set of rules from the [lints package](https://pub.dev/packages/lints) being used in `dart create` in the future and being the main recommended set, we should eventually migrate examples from pedantic to those, at least for the examples migrated to null safety.

There's some overlap so hopefully it shouldn't require too many updates.

PR #3263, opened 2021-05-18
Migrate to using recommended lints instead of pedantic

parlough:
Fixes https://github.com/dart-lang/site-www/issues/3235


kwalrath (approved):
Nice, work, @parlough! @mit-mit any problem with committing this now?

PR #3265, opened 2021-05-19
Evolution: add 2.13, update example

kwalrath:
Staged:

https://kw-www-dartlang-1.firebaseapp.com/guides/language/evolution


src/_guides/language/evolution.md
```diff
@@ -180,6 +180,18 @@ turn into edit-time analysis errors.
 
 In Dart 2.12, **[Dart FFI][]** graduated from beta to the stable channel.
 
+### Dart 2.13
+
+Dart 2.13 added support for **[type aliases][]**,
```
parlough, line 185: Do we need to specify that support was added for "non-function type aliases" or were the function typedefs not referred as type aliases originally?

parlough (approved):
Looks good to me, one question though.

src/_guides/language/evolution.md
mit-mit: Yeah, either "added support for non-function type aliases" or perhaps "added complete support for type aliases"?

mit-mit (approved):

src/_guides/language/evolution.md
kwalrath: I tried something slightly different:

Dart 2.13 expanded support for type aliases (typedef), which used to work only for function types but now work for any type. 

wdyt?

PR #3267, opened 2021-05-19
fix to appengine package link

lambdaR:
hovering the mouse over 'appengine package' only affects 'package'.

### DIFF
--- a/src/server/google-cloud.md
+++ b/src/server/google-cloud.md
@@ -72,13 +72,13 @@ For more information, see the [GKE overview][].
 [App Engine flexible environment][], which does not
 [autoscale to zero instances][scale to zero], so we recommend **Cloud Run** for new
 server-side Dart code.
-If you _want_ to use App Engine, consider using the [`appengine` package][].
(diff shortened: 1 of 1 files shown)


lambdaR:
@googlebot I signed it!

parlough:
Hey @lambdaR! Thanks for the pull request :)

It seems intentional that the "inline code" style was used for `appengine` as it is a package that can be utilized and imported, so perhaps the issue is that we instead need to introduce a hover style for links that are inline code(`a code`). What do you think?

lambdaR:
@parlough  probably, or maybe it is better to exclude `appengine` from the link, i just did a PR.
also, there is a subtle difference of what you see in github preview and the actual web site https://dart.dev/server/google-cloud regarding the "inline code"

lambdaR:
there must be a better way to do this ... anyway happy I/O

kwalrath:
> there must be a better way to do this ... anyway happy I/O

I hope you've been enjoying I/O!

PR #3276, opened 2021-05-25
Finalize cmdline example migrations and updates

parlough:


examples/misc/test/samples_test.dart
```diff
@@ -23,16 +23,6 @@ void main() {
   const oneSecond = Duration(seconds: 0);
   final someDate = DateTime(1999);
 
-  test('hello world', () {
```
kwalrath, line 26: Why remove this?

kwalrath (approved):
LGTM, but I'm curious about the test. Also... should this PR update the try-dart section to use single quotes for the Hello World example?

examples/misc/test/samples_test.dart
parlough: Not sure that I remember any good reasoning. I've added it back!

parlough:
Good point on fixing the try-dart section here, thanks! Luckily the compile resulted in minimal changes.

PR #3281, opened 2021-05-27
Update daily news and tests to null safety and latest recommended lints

parlough:
These files are not currently used anywhere as far as I can tell, but for now we can at least update the example in-case we use or refer to it again in the future.


kwalrath:
This example used to be featured in a "futures & error handling" page, but I don't remember what happened to that, or whether anything else pointed to it. I'll file an issue to remind us to consider removing this example.

kwalrath (approved):

PR #3282, opened 2021-05-28
Fix null safe example to remove error and introduce warning

### LINKED ISSUE #3139: 'Understanding null safety' page issue

Description of issue:

Taking the code written [here](https://dart.dev/null-safety/understanding-null-safety#unnecessary-code-warnings)

```dart
String checkList(List? list) {
  if (list == null) return 'No list';
  if (list?.isEmpty) { 
    return 'Empty list';
  }
  return 'Got something';
}
```

Instead of showing a warning like "The receiver can't be null, so the null-aware operator '?.' is unnecessary.", it gives an error:

> A nullable expression can't be used as a condition. 

image

parlough:
The `list?.isEmpty` evaluates to `bool?` resulting in an error instead of a warning, so this modifies the example to introduce the expected _warning_. 

I'm open to other ways to fix the conditional statement as well!

Also adds the return type to be consistent with the other surrounding examples.

Fixes #3139

### DIFF
--- a/src/null-safety/understanding-null-safety/index.md
+++ b/src/null-safety/understanding-null-safety/index.md
@@ -726,9 +726,9 @@ redundantly check it again for `null`:
 
 ```dart
 // Using null safety:
-checkList(List? list) {
+String checkList(List? list) {
   if (list == null) return 'No list';
-  if (list?.isEmpty) {
+  if (list?.isEmpty ?? false) {
     return 'Empty list';
   }
   return 'Got something';
--- a/src/null-safety/understanding-null-safety/index.md
+++ b/src/null-safety/understanding-null-safety/index.md
@@ -701,8 +701,8 @@ if you wrote something like:
 
 ```dart
 // Using null safety:
-String checkList(List list) {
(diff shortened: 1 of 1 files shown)


src/null-safety/understanding-null-safety/index.md
```diff
@@ -726,9 +726,9 @@ redundantly check it again for `null`:
 
 ```dart
 // Using null safety:
-checkList(List? list) {
+String checkList(List? list) {
   if (list == null) return 'No list';
-  if (list?.isEmpty) {
+  if (list?.isEmpty ?? false) {
```
munificent, line 731: While you're at it, line 705 needs the same fix.

Also, would you mind changing `List` to `List<Object>`? I think we generally try to avoid raw types in docs.

munificent (approved):
This is great, thanks!

src/null-safety/understanding-null-safety/index.md
parlough: Good suggestions, thanks. I've added the generic type and fixed the other example as well.

PR #3291, opened 2021-06-02
Fix numbering of steps on cmdline page

parlough:
I guess I'm still learning to count...


kwalrath (approved):
No big, since markdown ignores the number, but it does feel better to have it look right in the source!

PR #3293, opened 2021-06-03
Add exception for methods declared as void

Mayank-9018:
> All functions return a value.

Added exception for method declared as `void`


parlough:
Hey @Mayank-9018, thanks so much for the contribution! 

I'm not too familiar with the formal language specification, but I think there are some intricacies here as I think technically there is always some value returned from a function, but the `void` type indicates that this returned value should not have any meaning.

Perhaps @leafpetersen could provide some more context here then we can decide how to move forward. I do think highlighting `void` and its behavior in this section would definitely be worth it, I just want to make sure we avoid potential inconsistencies or confusion. 

@kwalrath Do you have any thoughts here?

kwalrath (commented):
What @parlough said is right: `void` doesn't mean that a function doesn't return a value; it means that the value should/must not be used. Although this is implied by https://dart.dev/guides/language/language-tour#a-basic-dart-program, it would be nice to mention it again in this section.

I don't recall whether the language/tools have gotten more restrictive about using `void` values. @leafpetersen?

leafpetersen:
Correct, void functions may return a value (for various reasons), the void return type just indicates that the value is not intended to be meaningful.

Mayank-9018:
I just went through it and yes `void` function can return a value, but it cannot be used. But I'm not so sure how to highlight the point there.

And also, while I was there, I noticed [there](https://dart.dev/guides/language/language-tour#return-values):
> If no return value is specified, the statement `return null; ` 

This would be applicable only if the method is not type annotated (which would then be dynamic). If the method is type annotated, then the type should be null-safe for it to return `null`. Do you think that this should be mentioned there?

Mayank-9018:
@kwalrath 
 > I don't recall whether the language/tools have gotten more restrictive about using `void` values.

I think they have, Plugins do raise an error, and trying to use them returns this error: `Error: This expression has type 'void' and can't be used.
`

kwalrath:
> I just went through it and yes `void` function can return a value, but it cannot be used. But I'm not so sure how to highlight the point there.
> 
> And also, while I was there, I noticed [there](https://dart.dev/guides/language/language-tour#return-values):
> 
> > If no return value is specified, the statement `return null; `
> 
> This would be applicable only if the method is not type annotated (which would then be dynamic). If the method is type annotated, then the type should be null-safe for it to return `null`. Do you think that this should be mentioned there?

I suspect that no `return` would still result in `return null`; it'd just be a static error if the return type isn't nullable. 

@leafpetersen can you confirm?

parlough:
This seems to be the case and there's an analyzer diagnostic for this specific situation(https://dart.dev/tools/diagnostic-messages#body_might_complete_normally).

PR #3307, opened 2021-06-07
Add extension as a prettify declaration type keyword

parlough:


kwalrath (approved):
LGTM, but it's weird to see this code again! Does it really need to be duplicated?

parlough:
> LGTM, but it's weird to see this code again! Does it really need to be duplicated?

I agree. Currently we don't use the site-shared ones as far as I can tell. I created an issue to track this work: https://github.com/dart-lang/site-www/issues/3311

parlough:
Okay it seemed to work without any problems, so took less time than I expected: https://github.com/dart-lang/site-www/pull/3312

PR #3308, opened 2021-06-07
Standardize on Function style for parameters in MyStream example

### LINKED ISSUE #3298: TODO(chalin): I believe that handleError's test argument should be declared
TODO link: https://github.com/dart-lang/site-www/blob/master/null_safety_examples/misc/lib/tutorial/stream_interface.dart#L4

parlough:
It seems the SDK code uses the-non Function style for most of its declarations for Streams, but since `use_function_type_syntax_for_parameters` is part of the recommended rule set, we'll stay consistent with that here.

Updating to this is still compatible with the original declaration when used as an @override so everything should be good.

Fixes #3298


kwalrath:
LGTM, but we should get a pickier code reviewer to take a look.

@johnpryan? @domesticmouse?

johnpryan (approved):

PR #3312, opened 2021-06-08
Use vendor assets from site-shared

### LINKED ISSUE #3311: Use vendor assets from site-shared
Currently we duplicate jquery and the code-prettify files in site-www, but they are also present in site-shared. We should try to share these files if possible.

parlough:
This also updates Jquery from 3.4.1->3.6.0 as introduced by https://github.com/dart-lang/site-shared/pull/133

Fixes #3311


kwalrath (approved):
Seems to work well, and it's great to reduce the duplication. Thanks, @parlough!

PR #3318, opened 2021-06-08
Add a dart2js section to the null safety FAQ

### LINKED ISSUE #3314: Update null safety FAQ
We added a section about dart2js to the internal null safety FAQ (377214967). We should add it to the https://dart.dev/null-safety/faq, as well.

@sigmundch unless you already have a PR in flight, I can make this change.

/cc @ryanmcginnis

kwalrath:
Closes #3314

### DIFF
--- a/src/null-safety/faq.md
+++ b/src/null-safety/faq.md
@@ -301,6 +301,77 @@ and situations where a null value is really expected. So the tool tells you what
 it knows ("it looks like this condition will always be false!") and lets you
 decide what to do.
 
+## What should I know about dart2js and null safety?
+
+For a long time, the dart2js compiler has had optimizations specifically
+targeting null values and null checks. Because of that, null safety is not
+expected to affect much the output of the compiler.
+
+A few notes that are worth highlighting:
+
+* dart2js emits `!` null assertions, but you may not notice them.
+  That's because pre null-safety dart2js already emitted null checks
+  (so they are already in the existing programs).
+
+* dart2js emits these null assertions regardless of sound/unsound null safety
+  and regardless of optimization level. In fact, dart2js doesn't remove
+  `!` when using `-O3` or `--omit-implicit-checks`.
+
+* dart2js may optimize away unnecessary null checks. This is because the same
+  optimizations done by dart2js in the past will be able to eliminate
+  unnecessary null checks when it knows the value is not null.
+
+* By default dart2js would generate parameter subtype checks (runtime checks
+  used to ensure covariant virtual calls are given appropriate arguments).
+  These are elided with the `--omit-implicit-checks` flag just as before.
+  Recall that this flag can make programs have unexpected behavior if types
+  are invalid, so we continue to recommend that the code has strong test
+  coverage to avoid any surprises. In particular, dart2js optimizes code based
+  on the fact that inputs should comply with the type declaration. If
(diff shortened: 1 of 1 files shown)


kwalrath:
Staged: https://kw-www-dartlang-1.web.app/null-safety/faq#what-should-i-know-about-dart2js-and-null-safety

src/null-safety/faq.md
```diff
@@ -301,6 +301,77 @@ and situations where a null value is really expected. So the tool tells you what
...
+targeting null values and null checks. Because of that, null safety is not
+expected to affect much the output of the compiler.
+
+A few notes that are worth highlighting:
+
+* dart2js emits `!` null assertions, but you may not notice them.
+  That's because pre null-safety dart2js already emitted null checks
+  (so they are already in the existing programs).
+
+* dart2js emits these null assertions regardless of sound/unsound null safety
```
parlough, line 316: Based on questions I've seen around this exact output, I can see this section causing more questions about "why" this is the case. Is there anything we can do to more directly connect to or elaborate the "why"?

parlough (approved):
Looks great to me and I learned a bit :)

I had one question, but I think this would still be great without addressing it.

src/null-safety/faq.md
```diff
@@ -301,6 +301,77 @@ and situations where a null value is really expected. So the tool tells you what
 it knows ("it looks like this condition will always be false!") and lets you
 decide what to do.
 
+## What should I know about dart2js and null safety?
+
+For a long time, the dart2js compiler has had optimizations specifically
```
parlough, line 306: Maybe having a link to [dart2js](https://dart.dev/tools/dart2js) somewhere would be worthwhile?

src/null-safety/faq.md
```diff
@@ -301,6 +301,77 @@ and situations where a null value is really expected. So the tool tells you what
...
+  and regardless of optimization level. In fact, dart2js doesn't remove
+  `!` when using `-O3` or `--omit-implicit-checks`.
+
+* dart2js may optimize away unnecessary null checks. This is because the same
+  optimizations done by dart2js in the past will be able to eliminate
+  unnecessary null checks when it knows the value is not null.
+
+* By default dart2js would generate parameter subtype checks (runtime checks
+  used to ensure covariant virtual calls are given appropriate arguments).
+  These are elided with the `--omit-implicit-checks` flag just as before.
```
sigmundch, line 326: maybe we can say:

 > ... with the `-O3` and `-O4` flags just as before.

In our internal build systems we use the `--omit-implicit-checks` flags, which is why I used that for the internal FAQ, but external users are more likely to be familiar with the `-O` flags.

sigmundch (approved):

src/null-safety/faq.md
parlough: Thanks! Yeah I think even a small clarification around maintaining program semantics within this point would be sufficient, as otherwise it may not be immediately clear to the average reader.

src/null-safety/faq.md
parlough: Good point. On the website we don't document the flag `--omit-implicit-checks` but rather mention it as part of the `-O` flags.

kwalrath:
I'm going to commit this (since it matches the internal version) and then address the review comments in another PR & CL.

PR #3327, opened 2021-06-11
Set html background color

kevmo314:
This sets the background color for overscrolling on mac os to be the same as the header and footer instead of the default white. The background for the remainder of the body remains white because the `body` tag has its own `background-color: white` declaration.

Before:

Screen Shot 2021-06-11 at 1 53 28 PM

After:

Screen Shot 2021-06-11 at 1 56 09 PM


parlough:
Thanks for the contribution, a consistent color definitely seems better than the stark switch to white. I've allowed the checks to run and assigned a reviewer :)

johnpryan (approved):

parlough:
Thanks again @kevmo314 and thanks for the review @johnpryan!

PR #3328, opened 2021-06-13
📝 Indicates `dart create` without nnbd before 2.13

AlexV525:
Currently `dart create` will create with a nnbd template, so it won't require further migrations.

### DIFF
--- a/src/null-safety/index.md
+++ b/src/null-safety/index.md
@@ -74,18 +74,17 @@ Sound null safety is available in Dart 2.12 and Flutter 2.
 For instructions on how to migrate your code to null safety,
 see the [migration guide][].
 
+{{site.alert.version-note}}
+  Before Dart 2.13, the templates used by the [`dart create`][] command
+  and IDEs aren't null safe, so you need to migrate the code they create.
+  For example:
 
-### Creating a null-safe package or app {#create}
-
-The templates used by the [`dart create`][] command and IDEs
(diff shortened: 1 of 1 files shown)


parlough:
Thanks for this, good catch! What do you think about instead focusing the documentation around new versions which already have null-safe templates and just need to use `dart create` without migrating. Then add an an extra `version-note` explaining the extra requirement to migrate before 2.13? As time progresses most new users reading this should be using the latest stable SDKs with no need to migrate and for users who aren't yet they'll still have the version note to refer to.

As an example, [`dart analyze`](https://dart.dev/tools/dart-analyze) has this:
image

Which was implemented like:

```markdown
{{site.alert.version-note}}
  Before Dart 2.13, `dart analyze` supported only directory arguments.
{{site.alert.end}}
```

AlexV525:
That would be great too!

AlexV525:
I've removed the `create` anchor completely, with the version note alert instead.

AlexV525:
Should I add my name to AUTHORS too?

parlough (approved):
Thanks for this, looks good to me.

kwalrath (commented):
Looks great!

(I forgot we even had an AUTHORS file. flutter/website has one, too, although it's been updated a little more recently. I'll look into either deleting these or marking them as no longer maintained and pointing to GitHub history.)

PR #3330, opened 2021-06-14
Clarify that FFI is not just about functions

mit-mit:
There was an email thread where a customer was confused by the name of FFI, that it might only be about calling functions.

This is a small edit to at least mention memory access. But perhaps we should also add an example of this?


dcharkes (approved):

mit-mit:
@dcharkes wdyt about the sample question I had?

dcharkes:
> @dcharkes wdyt about the sample question I had?

We could, but the API docs and further samples make this clear already right? (It's kind of obvious if your function signatures have pointer types that you would need to be able to construct these. Or is that just obvious to me?)

domesticmouse:
+1 to adding an example that clarifies this question. Daco, for better or worse the percentage of programmers that know explicitly that interop with C via a FFI includes dealing with manual memory management is dropping, year over year.

src/_guides/libraries/c-interop.md
```diff
@@ -7,7 +7,8 @@ sqlite: "https://github.com/dart-lang/sdk/tree/master/samples/ffi/sqlite"
 ---
 
 Dart mobile, command-line, and server apps running on the [Dart Native
-platform](/overview#platform/) can use the dart:ffi library to call native C APIs.
+platform](/overview#platform/) can use the dart:ffi library to call native C APIs,
+and to access and manipulate native memory.
```
kwalrath, line 11: _access_ is a bit unclear (https://developers.google.com/style/word-list#letter-a), and I'm not sure what you mean by _manipulate_. This is publishable, but I'd prefer to be more specific. Would this be accurate:

```suggestion
and to read, write, allocate, and deallocate native memory.
```

kwalrath:
Ping...🔊

mit-mit:
Took your rewrite suggestion

parlough (approved):

PR #3337, opened 2021-06-15
Update implicit-casts flag example

### LINKED ISSUE #3304: Update old examples involving implicit downcasts
Update examples in analyzer documentation and type system documentation around implicit downcasts since they were removed with null safety. There may be other cases as well.

The example in [Enabling stricter type checks](https://dart.dev/guides/language/analysis-options#enabling-additional-type-checks) is no longer relevant as it won't work without the option anyway.

As we migrate [The Dart type system](https://dart.dev/guides/language/type-system) article we'll need to keep this in mind as well.

parlough:
Fixes #3304

### DIFF
--- a/src/_guides/language/analysis-options.md
+++ b/src/_guides/language/analysis-options.md
@@ -166,14 +166,14 @@ You can use the flags together or separately; both default to `true`.
 
 `implicit-casts: <bool>`
 : A value of `false` ensures that the type inference engine never
-  implicitly casts to a more specific type.
+  implicitly casts from `dynamic` to a more specific type.
   The following valid Dart code
   includes an implicit downcast that would be caught by this flag:
 
 {:.fails-sa}
 <?code-excerpt "../null_safety_examples/analysis/lib/assignment.dart (implicit-downcast)" replace="/(s = )(o)/$1[!$2!]/g"?>
 {% prettify dart class="analyzer" %}
-Object o = ...
+dynamic o = ...
 String s = [!o!]; // Implicit downcast
 String s2 = s.substring(1);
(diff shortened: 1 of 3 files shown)


src/_guides/language/analysis-options.md
```diff
@@ -166,24 +166,30 @@ You can use the flags together or separately; both default to `true`.
...
 ```nocode
-error - A value of type 'Object' can't be assigned to a variable of type 'String' at lib/assignment.dart:11:14 - (invalid_assignment)
+error - A value of type 'dynamic' can't be assigned to a variable of type 'String' at lib/assignment.dart:11:14 - (invalid_assignment)
 ```
 
+{{site.alert.version-note}}
+With a [language version][] before Dart 2.12, 
+this flag also catches implicit downcasts from others types,
+but support for those was removed from the language.
+{{site.alert.end}}
```
kwalrath, line 191: I had a slightly different version in the works... not sure it's better.
Also, I think we usually indent these notes.

```suggestion
{{site.alert.version-note}}
  In packages that use a [language version][] before 2.12
  (when support for [null safety][] was introduced),
  code can implicitly downcast from non-`dynamic` types such as `Object`.
  The `implicit-casts` flag can catch those non-`dynamic` downcasts,
  even if you're using a more recent Dart SDK.
{{site.alert.end}}

  [null safety]: /null-safety
```

parlough: I like that this yours is more specific, not only about the language version meaning but also in what cases this will catch before 2.12. Thanks!

kwalrath (approved):

PR #3342, opened 2021-06-16
Fix errors in FAQ question about List.filled

stereotype441:
There was a syntax error (`List?<Foo?>`).  Also the example talked about initializing a list with the same value when in fact the "bad" example was initializing each element of the list with a different value.  This change fixes the syntax error and splits the one example into two (one using `List.filled`, for when there really is the same value, and the other using `List.generate`).

### DIFF
--- a/src/null-safety/faq.md
+++ b/src/null-safety/faq.md
@@ -262,8 +262,7 @@ _jellyCounts = List<int>.filled(jellyMax + 1, 0); // List initialized with fille
 
 If you are setting the elements of the list via an index, or you are populating
 each element of the list with a distinct value, you should instead use the
-[`generate`](https://api.dart.dev/stable/dart-core/List/List.generate.html)
-constructor to build the list.
+list literal syntax to build the list.
 
 {:.bad}
 {% prettify dart tag=pre+code %}
@@ -275,7 +274,10 @@ for (var i = 0; i <= jellyMax; i++) {
 
 {:.good}
 {% prettify dart tag=pre+code %}
-_jellyPoints = List<Vec2D>.generate(jellyMax + 1, (i) => Vec2D()); // List initialized with generate constructor
+_jellyPoints = [
+  for (var i = 0; i <= jellyMax; i++)
+    Vec2D() // Each list element is a distinct Vec2D
+];
 {% endprettify %}
 
 ## What happened to the default List constructor?
--- a/src/null-safety/faq.md
+++ b/src/null-safety/faq.md
@@ -262,7 +262,8 @@ _jellyCounts = List<int>.filled(jellyMax + 1, 0); // List initialized with fille
 
 If you are setting the elements of the list via an index, or you are populating
 each element of the list with a distinct value, you should instead use the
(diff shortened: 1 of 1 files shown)


src/null-safety/faq.md
```diff
@@ -249,20 +249,33 @@ use the [`filled`](https://api.dart.dev/stable/dart-core/List/List.filled.html)
...
-_jellyPoints = List<Vec2D>.filled(jellyMax + 1, Vec2D()); // List initialized with filled constructor
+_jellyCounts = List<int>.filled(jellyMax + 1, 0); // List initialized with filled constructor
 {% endprettify %}
 
-If you are setting the elements of the list via an index, you should instead use
-the `add` function to build the list. That is less error-prone and more
-readable.
+If you are setting the elements of the list via an index, or you are populating
+each element of the list with a distinct value, you should instead use the
+[`generate`](https://api.dart.dev/stable/dart-core/List/List.generate.html) constructor to build the list.
```
kwalrath, line 265: Let's keep the line length <= 80, where possible. (Normally we'd also use a macro for the API doc link, but since this doc is a copy of internal doc... nope.)

```suggestion
[`generate`](https://api.dart.dev/stable/dart-core/List/List.generate.html)
constructor to build the list.
```

kwalrath (approved):
Thanks, @stereotype441!

src/null-safety/faq.md
```diff
@@ -249,20 +249,33 @@ use the [`filled`](https://api.dart.dev/stable/dart-core/List/List.filled.html)
...
+
+{:.bad}
+{% prettify dart tag=pre+code %}
+_jellyPoints = List<Vec2D?>(jellyMax + 1);
+for (var i = 0; i <= jellyMax; i++) {
```
natebosch, line 270: [nit] The rewrite I would do is

```dart
_jellyPoints = [
  for (var i = 0; i <= jellyMax; i++)
    Vec2D(),
 ];
 ```
 
 We could make this slightly more realistic if we add a `growable: false` argument to `List.generate` which makes it infeasible to use the list literal. I'm not sure if doing that would add noise that detracts from the point though?

stereotype441: Honestly that's the rewrite I would have done too.  But I have the impression that `List.generate` is what we're trying to steer users to these days (and that the VM and dart2js folks have put special effort into making sure it's performant), so I figured 
I would go with that.  @leafpetersen / @munificent, do either of you have a preference for what we recommend here?

leafpetersen: List.generate was optimized: https://github.com/dart-lang/sdk/issues/42283 . Not sure about the collection literal, maybe @alexmarkov , @mraleph or @rakudrama can comment here.  In general though, I'd probably steer people first and foremost towards the most maintainable/readable code (so long as perf is adequate).  So my vote is probably for the collection literal.

stereotype441: Ok, switched to a collection literal.

stereotype441:
@natebosch @leafpetersen @kwalrath I believe I've adequately addressed review comments, but I don't have the necessary permissions to land this PR.  Can one of you land it?

PR #3343, opened 2021-06-17
Deprecate pedantic in linter rule set list

parlough:
Updates the pedantic listing here with information about its deprecation based off the information added to https://pub.dev/packages/pedantic

### DIFF
--- a/src/tools/linter-rules.md
+++ b/src/tools/linter-rules.md
@@ -31,7 +31,9 @@ consider starting with a linter rule set,
 which the following packages provide:
 
 [lints][]
-: The rule sets that the Dart team encourages using. 
+: The rule sets that the Dart team encourages using
+  for command line and server applications. 
+  For Flutter programs, see [flutter_lints][].
   Dart and Flutter packages uploaded to [pub.dev]({{site.pub}}) 
   are [scored]({{site.pub}}/help/scoring) 
   with the `core` set of these rules.
--- a/src/tools/linter-rules.md
+++ b/src/tools/linter-rules.md
@@ -30,13 +30,13 @@ To avoid the need to individually select compatible linter rules,
 consider starting with a linter rule set,
 which the following packages provide:
 
-[lints]({{site.pub-pkg}}/lints)
+[lints][]
 : The rule sets that the Dart team encourages using. 
   Dart and Flutter packages uploaded to [pub.dev]({{site.pub}}) 
   are [scored]({{site.pub}}/help/scoring) 
   with the `core` set of these rules.
 
-[flutter_lints]({{site.pub-pkg}}/flutter_lints)
+[flutter_lints][]
 : The set of rules that the Flutter team encourages you to use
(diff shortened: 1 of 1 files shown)


src/tools/linter-rules.md
```diff
@@ -30,13 +30,13 @@ To avoid the need to individually select compatible linter rules,
 consider starting with a linter rule set,
 which the following packages provide:
 
-[lints]({{site.pub-pkg}}/lints)
+[lints][]
 : The rule sets that the Dart team encourages using.
```
domesticmouse, line 34: Consider adding "for Dart command line and server programs. See `flutter_lints` for Flutter programs."

domesticmouse (approved):

src/tools/linter-rules.md
```diff
@@ -30,13 +30,15 @@ To avoid the need to individually select compatible linter rules,
 consider starting with a linter rule set,
 which the following packages provide:
 
-[lints]({{site.pub-pkg}}/lints)
-: The rule sets that the Dart team encourages using. 
+[lints][]
+: The rule sets that the Dart team encourages using
+  for command line and server applications. 
+  For Flutter programs, see [flutter_lints][].
```
kwalrath, line 36: This doesn't seem quite right, since we encourage these lints for _all_ Dart code, and command line and server aren't all the non-Flutter use cases for Dart. Also not sure we need a link to `flutter_lints`, since that's the next item.

Maybe:
```suggestion
: The rule sets that the Dart team encourages using.
  If you're writing Flutter code,
  use the `flutter_lints` rule set,
  which builds on `lints`.
```

src/tools/linter-rules.md
```diff
@@ -45,11 +47,21 @@ which the following packages provide:
...
+[effective_dart][]
 : A set of rules corresponding to the guidelines in [Effective Dart][].
 
-[pedantic]({{site.pub-pkg}}/pedantic)
-: The set of rules used for all Google-internal Dart code
+[pedantic][] (_deprecated_)
+: The deprecated set of rules previously used to match
+  the rules used for all Google-internal Dart code.
+  Consider migrating to one of the [package:lints][lints] rule sets 
+  or [package:flutter_lints][flutter_lints] if you use Flutter.
```
kwalrath, line 57: I avoid using `package:blah` in the docs. How about:

```suggestion
  Consider migrating to one of the rule sets in
  the [lints][] or [flutter_lints][] package.
```

src/tools/linter-rules.md
parlough: That makes a lot of sense with `flutter_lints` building on them. This seems nicer and less cluttered, thanks!

parlough:
Thanks for the suggestions @kwalrath, these changes get the point across in a more succinct way.

kwalrath:
> Thanks for the suggestions @kwalrath, these changes get the point across in a more succinct way.

I'm all about the succinct. :)

kwalrath (approved):

PR #3348, opened 2021-06-18
Minor cleanup and dependency updates on webdev page

parlough:
Updates the web and build dev dependencies to their latest, null-safe versions.


kwalrath (approved):
Thanks, @parlough!

PR #3350, opened 2021-06-20
Fix hyperlinks in packages.md

hrishikesh-kadam:


parlough (approved):
Not sure how no one noticed this earlier...thanks a lot for these fixes!

hrishikesh-kadam:
@parlough This happened in #3150 PR which migrates the list to `pub-subcommands.md`

PR #3352, opened 2021-06-21
Introduce a dart create page

### LINKED ISSUE #3329: Create a dedicated page for dart create
I just came across some writing where a link to a dedicated dart create page may come in handy, particularly with the removal of `stagehand`.

We could at the very least document its core arguments and flags, as well as discuss at a high level what each template is useful for.

It should be located with the other `dart` tools at `/src/tools/dart-create.md`:

image

parlough:
Fixes #3329

### DIFF
--- a/src/tools/dart-create.md
+++ b/src/tools/dart-create.md
@@ -10,10 +10,6 @@ The same functionality is available in IDEs.
 
 {% include tools/dart-tool-note.md %}
 
-The command creates a directory with the specified name, 
-creates necessary files, configures static analysis, 
-and retrieves any necessary dependencies.
-
 When you run `dart create`, it first creates a directory with the project files. 
 Then it gets package dependencies (unless you specify the `--no-pub` flag.) 
 
@@ -48,7 +44,7 @@ These templates result in a file structure being created
 which follows the guidelines and suggestions
 outlined in [Package layout conventions](/tools/pub/package-layout).
 
-If the specified directory already exists, the command fails. 
+If the specified directory already exists, `dart create` fails. 
 You can force project generation with the `--force` flag:
 
 ```terminal
--- a/src/tools/dart-create.md
+++ b/src/tools/dart-create.md
@@ -41,23 +41,9 @@ the package and directory name `my_cli`, run the following command:
 $ dart create my_cli
 ```
 
-This results in a package structure similar to the following:
-
-```bash
-my_cli/                  # The projects parent directory
-  bin/                   # The bin directory for the CLI tools
-  .gitignore             # Prevents various Dart files from being added to Git staging
-  analysis_options.yaml  # Configures static analysis for the project
-  CHANGELOG.md           # Allows you to specify changes between releases
-  pubspec.lock           # Pins down the dependency information for your package
-  pubspec.yaml           # Configures package information and dependencies
-  README.md              # Provides a description for your application
-```
-
-{{site.alert.info}}
-  To learn more about this and the other templates' generated structures,
-  see [Package layout conventions](/tools/pub/package-layout).
-{{site.alert.end}}
+This and the other templates, result in a file structure being created
+which follows the guidelines and suggestions
+outlined in [Package layout conventions](/tools/pub/package-layout).
 
 To instead utilize one of the other templates, 
 specify its id with the `-t` flag:
--- a/src/tools/dart-create.md
+++ b/src/tools/dart-create.md
@@ -14,35 +14,23 @@ The command creates a directory with the specified name,
 creates necessary files, configures static analysis, 
 and retrieves any necessary dependencies.
 
-```terminal
-$ dart create [ARGUMENTS] <DIRECTORY>
-```
+When you run `dart create`, it first creates a directory with the project files. 
+Then it gets package dependencies (unless you specify the `--no-pub` flag.) 
 
-The file structure created and the dependencies specified
-are configured through a choice of templates. By default, 
-the `console-simple` template is used, 
-creating a simple command-line application.
-
-To use that default template and create an application with
-the package and directory name `my_cli`, run the following command:
+Here's an example of using `dart create` to create a directory named `my_cli` 
+that contains a simple console app (the default template):
 
 ```terminal
 $ dart create my_cli
 ```
 
-This and the other templates, result in a file structure being created
-which follows the guidelines and suggestions
-outlined in [Package layout conventions](/tools/pub/package-layout).
-
-To instead utilize one of the other templates, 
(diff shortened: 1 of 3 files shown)


src/tools/dart-create.md
```diff
@@ -0,0 +1,127 @@
...
+```bash
+my_cli/                  # The projects parent directory
+  bin/                   # The bin directory for the CLI tools
+  .gitignore             # Prevents various Dart files from being added to Git staging
+  analysis_options.yaml  # Configures static analysis for the project
+  CHANGELOG.md           # Allows you to specify changes between releases
+  pubspec.lock           # Pins down the dependency information for your package
+  pubspec.yaml           # Configures package information and dependencies
+  README.md              # Provides a description for your application
+```
```
parlough, line 55: Do you think it's worth having an example outlining a created file structure like this? With the comments and everything?

kwalrath: I'd prefer to have a link or a shared graphic/text. That makes it less likely that the info will get out of date.

parlough: Good point, I've instead consolidated this part to a link to existing package conventions. I wasn't so confident on the wording though, so open to suggestions. Thanks!

src/tools/dart-create.md
```diff
@@ -0,0 +1,113 @@
...
+{{site.alert.note}}
+  The command will attempt to run [`dart pub get`](/tools/pub/cmd/pub-get) 
+  after the project is created.
+
+  To avoid this you can specify the `--no-pub` flag:
+
+  ```terminal
+  $ dart create --no-pub <DIRECTORY>
+  ```
+{{site.alert.end}}
```
kwalrath, line 30: This doesn't seem crucial enough to go above the fold. What do you think of deleting this note in favor of the `--no-pub` mention above?

src/tools/dart-create.md
```diff
@@ -0,0 +1,113 @@
...
+
+The `dart create` command allows creating Dart projects
+from varying available templates through a command line
+or many IDE project creation workflows.
+
+{% include tools/dart-tool-note.md %}
+
+The command creates a directory with the specified name, 
+creates necessary files, configures static analysis, 
+and retrieves any necessary dependencies.
```
kwalrath, line 15: I'd like the first screenful to show or hint at the commands that people are most likely to use — more like https://dart.dev/tools/dart-analyze. Maybe something like this:

-------------

When you run `dart create`, it first creates a directory with the project files. Then it gets package dependencies (unless you specify the `--no-pub` flag.) Here's an example of using `dart create` to create a directory named `my_cli` that contains a simple console app (the default template):

```terminal
$ dart create my_cli
```


To use a different template, add a template argument:

```terminal
$ dart create -t web-simple my_web_app
```

The following table shows the templates you can use:
...

src/tools/dart-create.md
```diff
@@ -0,0 +1,113 @@
...
+
+```terminal
+$ dart create -t <TEMPLATE> <DIRECTORY>
+```
+
+Beyond the `console-simple` template, these are the currently available
+templates:
+
+|------------------+---------------------------------------------+------------------------------------------------------|
+| Template         | Example of use                              | More information                                     |
```
kwalrath, line 59: I think we can drop the second (example of use) column.

src/tools/dart-create.md
```diff
@@ -0,0 +1,113 @@
...
+
+As an example, to create a simple web application
+with the directory and name `my_web_app`,
+you can run the following command:
+
+```terminal
+$ dart create -t web-simple my_web_app
+```
+
+If the specified directory already exists, the command will fail.
```
kwalrath, line 78:
```suggestion
If the specified directory already exists, the command fails. 
```

(We try to keep to [present tense](https://developers.google.com/style/tense).)

kwalrath (commented):
Thank you for creating this, @parlough! The info looks good, but I'd like to make it a bit more direct, and to feature what the reader's most likely to do (especially in the first screenful).

src/tools/dart-create.md
```diff
@@ -0,0 +1,82 @@
...
+```terminal
+$ dart create -t web-simple my_web_app
+```
+
+The following table shows the templates you can use:
+
+|------------------+------------------------------------------------------|
+| Template         | Description                                          |
+|------------------|------------------------------------------------------|
+| `console-simple` | A simple command-line application.                   |
```
kwalrath, line 34: Google style is to [use _app_ instead of _application_](https://developers.google.com/style/word-list#app) except where you want to convey complexity. Also, I wanted to make clear that console-simple is the default.

```suggestion
| `console-simple` | A simple command-line app (the default template).                   |
```

src/tools/dart-create.md
```diff
@@ -0,0 +1,82 @@
...
+| `console-full`   | A complete command-line application.                 |
+| `package-simple` | A starting point for Dart libraries or applications. |
+| `server-shelf`   | A server application built using [shelf][].          |
+| `web-simple`     | A web application built using core Dart libraries.   |
```
kwalrath, line 38:
```suggestion
| `console-full`   | A complete command-line app.                 |
| `package-simple` | A starting point for Dart libraries or apps. |
| `server-shelf`   | A server built using [shelf][].          |
| `web-simple`     | A web app built using core Dart libraries.   |
```

src/tools/dart-create.md
```diff
@@ -0,0 +1,82 @@
...
+{:.table .table-striped .nowrap}
+
+[shelf]: {{site.pub-pkg}}/shelf
+
+These templates result in a file structure being created
+which follows the guidelines and suggestions
+outlined in [Package layout conventions](/tools/pub/package-layout).
```
kwalrath, line 45:
```suggestion
These templates result in a file structure that follows
[package layout conventions](/tools/pub/package-layout).
```

src/tools/dart-create.md
```diff
@@ -0,0 +1,82 @@
...
+---
+
+The `dart create` command creates a Dart project,
+using one of several supported templates.
+The same functionality is available in IDEs.
+
+{% include tools/dart-tool-note.md %}
+
+When you run `dart create`, it first creates a directory with the project files. 
+Then it gets package dependencies (unless you specify the `--no-pub` flag.)
```
kwalrath, line 14:
```suggestion
Then it gets package dependencies (unless you specify the `--no-pub` flag).
```

kwalrath (approved):
LGTM once you apply a few minor suggestions. Thank you,  @parlough!

parlough:
Thanks for all the feedback and suggestions @kwalrath! I've included your latest suggestions and reformatted the table.

PR #3355, opened 2021-06-24
Initialize error field so focus can be on the toUpperCase error

### LINKED ISSUE #2586: 'Understanding null safety' page issue

Description of issue:

https://dart.dev/null-safety/understanding-null-safety#null-assertion-operator has an example:

```dart
// Using null safety, incorrectly:
class HttpResponse {
  final int code;
  final String? error;

  HttpResponse.ok() : code = 200;
  HttpResponse.notFound()
      : code = 404,
        error = "Not found";

  String toString() {
    if (code == 200) return "OK";
    return "ERROR $code ${error.toUpperCase()}";
  }
}
```

The `HttpResponse.ok` constructor neglects to initialize the `error` member, which currently must be initialized because it is `final`.  I assume that we do still want to require explicit initialization `final` members even if they're nullable.

parlough:
Fixes #2586

### DIFF
--- a/src/null-safety/understanding-null-safety/index.md
+++ b/src/null-safety/understanding-null-safety/index.md
@@ -879,11 +879,14 @@ class HttpResponse {
   final int code;
   final String? error;
 
-  HttpResponse.ok() : code = 200;
+  HttpResponse.ok()
+      : code = 200,
+        error = null;
   HttpResponse.notFound()
       : code = 404,
         error = 'Not found';
 
+  @override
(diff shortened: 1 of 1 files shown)


kwalrath (approved):

kwalrath:
We should test all the code, now that we can... but that can wait for another PR.

Related issue: #1803

parlough:
@kwalrath Yep, it's on my radar after I finish all the type system work :)

PR #3357, opened 2021-06-24
Remove reference to fast-startup as it is now the only emitter for dart2js

### LINKED ISSUE #1997: Update /web/deployment
https://dart.dev/web/deployment#use-dart2js-flags-to-produce-better-javascript

Mentions `fast-startup` flag. For more info, it links to https://dart.dev/tools/dart2js, which does not mention this flag.

And then from the CHANGELOG: https://github.com/dart-lang/sdk/blob/master/CHANGELOG.md#dart2js-2

> The old "full emitter" back-end is removed and dart2js always uses the "fast startup" back-end. The generated fast startup code is optimized to load faster, even though it can be slightly larger. The --fast-startup and --no-fast-startup are allowed but ignored. They will be removed in a future version.

parlough:
Fixes #1997

This page does require other updates and improvements but that work can be done as further followup.

### DIFF
--- a/src/web/deployment.md
+++ b/src/web/deployment.md
@@ -38,28 +38,24 @@ For more information, see the [documentation for webdev][webdev].
 
 Google's apps often use the following [dart2js options](/tools/dart2js#options):
 
-- `--fast-startup`
 - `--minify`
 - `--trust-primitives`
 - `--trust-type-annotations`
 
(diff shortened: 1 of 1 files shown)


kwalrath (approved):

sigmundch (approved):
Thanks!

PR #3363, opened 2021-06-28
Fix various missing anchors

parlough:


kwalrath (approved):
I trust you tested these, so I'm not going to. Thanks, @parlough!

PR #3365, opened 2021-07-01
Update documentation for deprecated linter rule packages

parlough:
Removes the instructions for enabling `pedantic` and `effective_dart` as they are now deprecated.


kwalrath (approved):
I didn't know these had been deprecated. Thanks for this PR, @parlough!

PR #3367, opened 2021-07-01
Redirect /tools/dart-pub

kwalrath:
Contributes to #2719. Tested locally.


parlough (approved):
LGTM and works great locally!

parlough:
Do you think it would be desired to direct from something like **http://localhost:5000/tools/dart-pub/add** to **https://dart.dev/tools/pub/cmd/pub-add**?

Not necessary as part of this change, just curious?

kwalrath:
> Do you think it would be desired to direct from something like **http://localhost:5000/tools/dart-pub/add** or **http://localhost:5000/tools/dart-pub/pub-add** to **https://dart.dev/tools/pub/cmd/pub-add**?

Probably not unless we see 404s for those. Which I haven't noticed, but I'll make a note to check for that.

PR #3375, opened 2021-07-08
Update old dart create links/text and add new ones

parlough:


src/_tutorials/libraries/shared-pkgs.md
```diff
@@ -51,22 +51,24 @@ your app must itself be a package.
...
+to see what kinds of template files it can generate:
 
 ```terminal
-$ dart create
+$ dart create --help
 ```
 
-You'll see a list of generators, including various web and server-side apps.
+You'll see a list of templates, including various web and server-side apps.
 One of the generators is named **console-full**.
```
kwalrath, line 68:
```suggestion
One of the templates is named **console-full**.
```

kwalrath (approved):
Thanks! Just one quibble

src/_tutorials/libraries/shared-pkgs.md
parlough: Good catch, thanks!

PR #3376, opened 2021-07-08
Minimal updates to bring old cheatsheet relatively up to date

parlough:
I'm not sure how valuable this article is with the codelab cheatsheet and the language tour, but I thought I could for now at least make some minor updates.

Does this article get traffic?

### DIFF
--- a/src/_guides/language/cheatsheet.md
+++ b/src/_guides/language/cheatsheet.md
@@ -34,7 +34,7 @@ Equivalent: `new Map<>();`
 
 | `var` | Generic `var` with type inference |
 | `final` | Same as `var` but cannot be reassigned |
-| `const` | Compile time constant |
+| `const` | Compile-time constant |
 {:.table}
 
 
--- a/src/_guides/language/cheatsheet.md
+++ b/src/_guides/language/cheatsheet.md
@@ -20,7 +20,6 @@ Equivalent: `‘Substitution’ + val`
 ### `<type>[ ]`
 
 Creates an object of type `List<type>`.
-Equivalent: `new List<type>();`
 
 ### `const [1, 2, 3]`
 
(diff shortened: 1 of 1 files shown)


domesticmouse (approved):

kwalrath:
> I'm not sure how valuable this article is with the codelab cheatsheet and the language tour, but I thought I could for now at least make some minor updates.

Thank you! @filiph and I have discussed creating a real, printable one-pager (for I/O 2020, sigh), so I think a Dart cheatsheet could be very handy. It should be reformatted/rewritten and then made easier to find, but keeping it correct is a good first step.

> Does this article get traffic?

It gets some traffic. It is the 106th most visited page on dart.dev (out of ~150), just behind /tutorials/libraries/shared-pkgs and ahead of /web/debugging. 🤷‍♀️  In terms of average time-on-page, it's 38th, which I think indicates that people find it interesting.

Btw the highest time-on-page page, by far, is /codelabs/async-await.

src/_guides/language/cheatsheet.md
```diff
@@ -33,9 +32,9 @@ Equivalent: `new Map<>();`
 
 ## Declaring fields
 
-| `var` | Generic `var` (dynamic) |
-| `final` | Same as `var` but final |
-| `const` | Constant |
+| `var` | Generic `var` with type inference |
+| `final` | Same as `var` but cannot be reassigned |
+| `const` | Compile time constant |
```
kwalrath, line 37:
```suggestion
| `const` | Compile-time constant |
```

kwalrath (approved):
LGTM. Technically the compound modifier (compile-time) should be hyphenated because it's before the noun (https://developers.google.com/style/hyphens#compound-modifiers).

parlough:
@kwalrath A printable cheatsheet sounds great and once I get a chance I'd also like to explore the concept more.

Thanks for the stats, that's interesting. I guess async/await can be pretty confusing.

PR #3377, opened 2021-07-08
Add language repo link and update old spec link

parlough:
That old Ecma link doesn't lead to the PDF anymore, so updating the link to what it actually leads to.


kwalrath (approved):
Good catch. Thanks, @parlough!

PR #3383, opened 2021-07-11
Fix putting-it-all-together exercise in Iterables Codelab

Mayank-9018:
`isValidEmailAddress` method should be a part of main.dart and solution.dart; and should be excluded from test.dart

### DIFF
--- a/src/codelabs/iterables.md
+++ b/src/codelabs/iterables.md
@@ -969,8 +969,7 @@ It’s time to practice what you learned, in one final exercise.
 
 This exercise provides the class `EmailAddress`,
 which has a constructor that takes a string.
-Another provided function is `
-    EmailAddress()`,
+Another provided function is `EmailAddress()`,
 which tests whether an email address is valid.
 
 |----------------------+-------------------+----------------------------------|
--- a/src/codelabs/iterables.md
+++ b/src/codelabs/iterables.md
@@ -1019,6 +1019,10 @@ Iterable<EmailAddress> validEmailAddresses(Iterable<EmailAddress> emails) {
   TODO('Implement this method');
 }
 
+bool isValidEmailAddress(EmailAddress email) {
+  return email.address.contains('@');
+}
+
 class EmailAddress {
   String address;
 
@@ -1052,6 +1056,10 @@ Iterable<EmailAddress> validEmailAddresses(Iterable<EmailAddress> emails) {
   return emails.where((email) => isValidEmailAddress(email));
 }
 
+bool isValidEmailAddress(EmailAddress email) {
+  return email.address.contains('@');
+}
+
 class EmailAddress {
   String address;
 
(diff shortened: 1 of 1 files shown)


parlough (approved):
Thanks for this @Mayank-9018!

I think it's technically fine in the test file as it's mentioned as a provided function, but if we include it explicitly, the example is easier to follow without referring back to earlier in the page for the signature.

`EmailAddress` is included in the user's exercise code in a similar fashion, so I think including this makes sense too. Good idea including it as the last function, keeping focus on the unimplemented ones :)

Mayank-9018:
@parlough  Wait, So how can I call `isValidEmailAddress()` if it is not a part of main.dart? In my local development environment, I receive `Method not found: 'isValidEmailAddress'`.

kwalrath:
> @parlough Wait, So how can I call `isValidEmailAddress()` if it is not a part of main.dart? In my local development environment, I receive `Method not found: 'isValidEmailAddress'`. I initially thought it was an error, that instead of including the method in main.dart, it was included in test.dart

These exercises are intended to be run in the embedded DartPad, which uses test.dart. We avoided adding too much detail to the main file because we didn't want to distract from the purpose of the exercise.

parlough:
By that I meant that's it technically fine to include it just in `test.dart`, but if you want to include in `main.dart` and `solution.dart`, you cannot include it in `test.dart`. The testing works by appending its code to `main.dart`, resulting in duplicate definitions of the method:

Screenshot from 2021-07-13 11-03-57

kwalrath:
I don't know that we want to show the code for `isValidEmailAddress`... it's not exactly a real test. It's more of a quick check that might be more distracting than helpful.

Do we need to improve the text around these exercises to make clear that the code works only in the embedded DartPads?

parlough:
@kwalrath I agree, I think it's fine to not include it, as it keeps focus on the goal of the codelab.

I think specifying something per page may be a bit much, but since we already mention that the page uses embedded DartPads and link to the troubleshooting page, perhaps it's worth a mention there?

kwalrath:
> I think specifying something per page may be a bit much, but since we already mention that the page uses embedded DartPads at the top and link to the troubleshooting page, perhaps it's worth a mention on that page instead?

Oops, I thought I answered this. Yes, that makes sense to me. @Mayank-9018 would that have helped you?

Mayank-9018:
> > I think specifying something per page may be a bit much, but since we already mention that the page uses embedded DartPads at the top and link to the troubleshooting page, perhaps it's worth a mention on that page instead?
> 
> Oops, I thought I answered this. Yes, that makes sense to me. @Mayank-9018 would that have helped you?

Yes, something along those lines. In other codelabs, such as Asynchronous, I've noticed that already provided methods are implemented in test.dart and it makes sense; it keeps the focus on the exercise's goal rather than on what is already provided.

kwalrath:
I'm closing this in favor of #3408. Thanks, @Mayank-9018, for bringing this confusing point to our attention!

PR #3393, opened 2021-07-14
Delete /tools/dart-vm

kwalrath:
Contributes to #2719

Staged:

https://kw-staging-dartlang-2.web.app/tools/dart-vm


kevmoo:
I'm okay with it, but it might be good to double check with @mit-mit – who is OOO until next week IIRC

parlough (approved):
This seems like a good idea to me. I thought calling this page "The Dart VM" (dart-vm) was a bit confusing as it only talked about the VM's ability to run programs.

kwalrath:
I converted this to a draft so we wouldn't accidentally merge this change before @mit-mit gets a chance to see it.

mit-mit (approved):

PR #3340, opened 2021-06-15
Initial migration of type system examples

parlough:
This changes updates the type system and sound problem articles to null safety. This doesn't add much new content, which may be a good idea to work on afterwards, but instead makes sure what is still present is up to date with the latest changes to Dart.

Beyond updating the code samples, this PR works to analyze more of the code and verify more of it and the analysis results are checked properly for exclusion.

In a followup to this PR, the `null_safety_examples` will be merged with `examples` and work will be done to consolidate the existing GitHub actions and make sure they're not ignoring failures.

Fixes https://github.com/dart-lang/site-www/issues/2618
Fixes https://github.com/dart-lang/site-www/issues/2444
Fixes https://github.com/dart-lang/site-www/issues/2590

_Staged instance last updated July 6th_
**Staged Type System article:** https://parlough-dart-dev.web.app/guides/language/type-system
**Staged Sound Problems article:** https://parlough-dart-dev.web.app/guides/language/sound-problems

PR #3396, opened 2021-07-15
Fix and ignore more lint warnings

parlough:
This brings the remaining warnings almost to those just remaining in the strong examples to be fixed in https://github.com/dart-lang/site-www/pull/3340


kwalrath (approved):
lgtm!

PR #3400, opened 2021-07-18
Document using Deprecated annotation with message parameter

### LINKED ISSUE #3397: Switch primary documentation from @deprecated to @Deprecated in language tour
In the [Metadata section](https://dart.dev/guides/language/language-tour#metadata) of the language tour we document `deprecated` but not the `Deprecated(String message)` annotation.

Since the rule(https://dart.dev/tools/linter-rules#provide_deprecation_message) to use the message constructor is included in the core set of lints, we should at least mention it, but also focus the documentation on it.

parlough:
Closes #3397


parlough:
Rebased on the main branch to fix conflicts.

kwalrath (approved):

PR #3405, opened 2021-07-20
Add arm64 (Apple M1) support in SDK archive

mit-mit:


src/tools/sdk/dart_sdk_archive/lib/src/version_selector.dart
```diff
@@ -144,13 +144,30 @@ class VersionSelector {
...
+        // No Mac 32-bit SDK builds after 2.80
         if (name == 'macOS' && platformVariant.architecture == 'ia32') {
-          // No Mac 32-bit SDK builds >= 2.80
           if (versionInfo.version > Version(2, 7, 0)) {
             continue;
           }
         }
 
+        // No Mac arm64 SDK builds before 2.14.0, after July 1st 2021, and not in stable yet.
+        // TODO: After 2.14 simply change these tests to versionInfo.version >= Version(2,14,0).
```
athomas, line 155: What about simply checking for `>= 2.14.0-216.0.dev` (the first dev version that had signed builds)?

We could also pick any arbitrary later dev version (`2.14.0-281.0.dev` would be the first in July that was published to dart.dev).

Then it could be:

```dart
versionInfo.version >= Version(2, 14, 0, pre: '281.0.dev')
```

Older beta versions would be before that and all 2.13.x stable versions as well so we wouldn't need the date or stable channel check.

athomas (approved):

src/tools/sdk/dart_sdk_archive/lib/src/version_selector.dart
mit-mit: PR updated, PTAL

mit-mit:
Also added the compiled output

src/tools/sdk/dart_sdk_archive/lib/src/version_selector.dart
```diff
@@ -144,13 +144,26 @@ class VersionSelector {
...
         }
 
+        // No Mac arm64 SDK builds before 2.14.0-281.0.dev, and not in stable yet.
+        // TODO: After 2.14 simply change these tests to versionInfo.version < Version(2,14,0).
+        if (name == 'macOS' && platformVariant.architecture == 'arm64') {
+          if (versionInfo.version < Version(2, 14, 0, pre: '281.0.dev')) {
+            continue;
+          }
+
+          if (versionInfo.channel == 'stable') {
```
athomas, line 161: Why keep that? For now, it's dead code (`2.13.x` is less than `2.14.0-281.0.dev` so on stable this if is never reached) and when 2.14.0 is released it will be wrong to skip stable (at least according to the TODO above).

mit-mit: I don't want to automatically show up in stable until we know for sure that everything works are intended.

athomas: LGTM then.

src/tools/sdk/dart_sdk_archive/lib/src/util.dart
```diff
@@ -39,6 +39,7 @@ const Map<String, String> archiveMap = {
   'Windows': 'windows',
   'ia32': 'ia32',
   'x64': 'x64',
+  'arm64': 'arm64',
   'ARMv7': 'arm',
   'ARMv8 (ARM64)': 'arm64',
```
kwalrath, line 44: It's a little confusing to see both arm64 & ARM64 in the table. I supposed you only see it when **OS** is **All**, but... can we make them more consistent?

image

mit-mit: We can certainly change `arm64` to `ARM64` on macOS, and it looks like Apple also uses that:
https://developer.apple.com/documentation/xcode/writing-arm64-code-for-apple-platforms

PR #3407, opened 2021-07-20
Update and clean up SDK archive tools

parlough:
- Update dependencies
- Switch back to official platform detect now that it has a null safe release
- Switch to recommended lints
- Format and order Dart files
- Switch from pedantic's `unawaited` to an ignore comment
- Minor doc updates

### DIFF
--- a/src/tools/sdk/archive/pubspec.yaml
+++ b/src/tools/sdk/archive/pubspec.yaml
@@ -1,11 +1,13 @@
 name: download_archive
 description: Display a list of downloads for dart.dev.
 environment:
-  sdk: '>=2.12.0 <3.0.0'
+  sdk: '>=2.13.0 <3.0.0'
+
 dependencies:
   dart_sdk_archive:
     path: ../dart_sdk_archive
+
 dev_dependencies:
   build_runner: ^2.0.0
-  pedantic: ^1.11.0
   build_web_compilers: ^3.0.0
+  lints: ^1.0.0
--- a/src/tools/sdk/dart_sdk_archive/lib/src/run_app.dart
+++ b/src/tools/sdk/dart_sdk_archive/lib/src/run_app.dart
@@ -2,7 +2,6 @@ import 'dart:html';
 
 import 'package:dart_sdk_archive/src/version_selector.dart';
 import 'package:http/browser_client.dart';
-import 'package:pedantic/pedantic.dart';
 import 'package:sdk_builds/sdk_builds.dart';
 
 Future<void> runApp() async {
@@ -32,7 +31,10 @@ Future<void> runApp() async {
     querySelector('#dev-os') as SelectElement,
   );
 
-  unawaited(stableSelector.init());
-  unawaited(betaSelector.init());
-  unawaited(devSelector.init());
+  // ignore: unawaited_futures
+  stableSelector.init();
+  // ignore: unawaited_futures
+  betaSelector.init();
+  // ignore: unawaited_futures
+  devSelector.init();
 }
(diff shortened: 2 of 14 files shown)


src/tools/sdk/archive/pubspec.yaml
```diff
@@ -1,11 +1,14 @@
...
 dependencies:
   dart_sdk_archive:
     path: ../dart_sdk_archive
+
 dev_dependencies:
   build_runner: ^2.0.0
-  pedantic: ^1.11.0
   build_web_compilers: ^3.0.0
+  lints: ^1.0.0
+  pedantic: ^1.11.0
```
kwalrath, line 14: Do we really need pedantic?

src/tools/sdk/dart_sdk_archive/pubspec.yaml
```diff
@@ -1,19 +1,21 @@
...
 
 dev_dependencies:
-  build_runner: ^1.12.0
-  build_web_compilers: ^2.16.0
+  build_runner: ^2.0.0
+  build_web_compilers: ^3.0.0
   dart_style: ^2.0.0
   grinder: ^0.9.0
+  lints: ^1.0.0
   pedantic: ^1.11.0
```
kwalrath, line 20: Delete pedantic?

kwalrath (approved):
I see a couple of references to pedantic remain... Do we really need them? Otherwise, LGTM!

kwalrath:
None of the code changes looks controversial, but I still hesitate to merge without an engineer taking a look.

parlough:
@kwalrath We were using pedantic to mark unawaited futures, but since pedantic is deprecated and we can do that with comments instead, I've removed it in the latest commit.

src/tools/sdk/archive/_README.md
```diff
@@ -6,4 +6,4 @@ If you modify that Dart code for this page, don't forget to regenerate
 the JavaScript:
 
     cd src/tools/sdk/archive
-    pub run build_runner build --release --output out
+    dart run build_runner build --release --output out
```
domesticmouse, line 9: This works now? Or should it be `dart pub run build_runner build --release --output out` ?

src/tools/sdk/dart_sdk_archive/lib/src/run_app.dart
```diff
@@ -32,7 +31,10 @@ Future<void> runApp() async {
     querySelector('#dev-os') as SelectElement,
   );
 
-  unawaited(stableSelector.init());
-  unawaited(betaSelector.init());
-  unawaited(devSelector.init());
+  // ignore: unawaited_futures
```
domesticmouse, line 34: I have a vague recollection that `unawaited` was on it's way into the core libraries? @johnpryan probably knows this better than I do.

domesticmouse (approved):

domesticmouse:
Just a couple of open questions, but not meant to be blocking

src/tools/sdk/archive/_README.md
mit-mit: Yes, this works

src/tools/sdk/dart_sdk_archive/lib/src/run_app.dart
parlough: Yep, but it won't be added until at least the next stable release, so don't want to rely on it yet. Here's a link to the new `unawaited` docs: https://api.dart.dev/dev/dart-async/unawaited.html

parlough:
Rebased this on the latest arm64 -> ARM64 changes and waiting for tests to pass. Thanks for the reviews.

src/tools/sdk/archive/_README.md
parlough: This is gone over in more detail in the (newer) `dart run` page: https://dart.dev/tools/dart-run#in-a-depended-on-package

PR #3409, opened 2021-07-21
Documenting `false_secrets` in `pubspec.yaml`

jonasfj:
Implementation on `pub` client: https://github.com/dart-lang/pub/pull/3049.

--------

Many of the patterns and heuristics for detection of leaks are adopted from [How Bad Can It Git?](https://www.ndss-symposium.org/wp-content/uploads/2019/02/ndss2019_04B-3_Meli_paper.pdf) and [awslabs/git-secrets](https://github.com/awslabs/git-secrets).
We could perhaps consider mentioning these in the documentation, to help people better understand what kind of heuristics is employed for leak detection.

--------

This won't ship until Dart 2.15, I'm not planning to sneak in new features right before the window closes.

### DIFF
--- a/src/tools/pub/pubspec.md
+++ b/src/tools/pub/pubspec.md
@@ -70,6 +70,11 @@ A pubspec can have the following fields:
 : Optional. Specify where to publish a package.
   [_Learn more._](#publish_to)
 
+`false_leaks`
+: Optional. Specify git-ignore style patterns for files to ignore when warning
+  against potential leak of secrets before publishing.
+  [_Learn more._](#false_leaks)
+
 Pub ignores all other fields,
 
 <aside class="alert alert-info" markdown="1">
@@ -272,6 +277,38 @@ publish_to: none
 {% endprettify %}
 
 
+### False_leaks
+
+When publishing leak detection on the client will abort publishing if it finds
+potential leaks of secret credentials, API keys and cryptographic keys in files
+to be published.
+While detection of possible secrets relies on advanced heuristics using regular
+expressions and entropy thresholds, the detection is not perfect.
+If the leak detection reports a _false positive_ you can specify a list of
+[git-ignore patterns][git-ignore-format] under `false_leaks` in `pubspec.yaml`.
+
+The leak detection will ignore any files matching a pattern in `false_leaks`.
+The example below ignores the file `lib/src/file_with_hardcoded_api_key.dart`
+and all `.pem` files in the `test/localhost_certificates/` folder. Starting a
+[git-ignore pattern][git-ignore-format] with slash `/` ensures that is it only
+considered relative to the root folder.
+
+{% prettify yaml tag=pre+code %}
+false_leaks:
+ - /lib/src/file_with_hardcoded_api_key.dart
+ - /test/localhost_certificates/*.pem
+{% endprettify %}
+
+{{site.alert.version-note}}
+  Leak detection will never prevent all leaks. It's your responsibility to
+  manage your credentials, prevent accidental leaks, and revoke credentials
+  that are accidentally leaked.
+
+  You cannot rely on leak detection prevent leaks, it will only detect a
+  limited set of common patterns, and is only intended to mitigate common
+  mistakes.
+{{site.alert.end}}
+
 ### SDK constraints
 
 A package can indicate which versions of its dependencies it supports, but
@@ -348,3 +385,4 @@ accidentally install packages that need Flutter.
 
 [pubsite]: {{site.pub}}
 [semantic versioning]: https://semver.org/spec/v2.0.0-rc.1.html
+[git-ignore-format]: https://git-scm.com/docs/gitignore#_pattern_format
--- a/src/tools/pub/pubspec.md
+++ b/src/tools/pub/pubspec.md
@@ -1,5 +1,6 @@
 ---
 title: The pubspec file
+description: Reference guide for the fields in pubspec.yaml.
 ---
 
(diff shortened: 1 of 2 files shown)


src/tools/pub/pubspec.md
```diff
@@ -272,6 +277,38 @@ publish_to: none
...
+The example below ignores the file `lib/src/file_with_hardcoded_api_key.dart`
+and all `.pem` files in the `test/localhost_certificates/` folder. Starting a
+[git-ignore pattern][git-ignore-format] with slash `/` ensures that is it only
+considered relative to the root folder.
+
+{% prettify yaml tag=pre+code %}
+false_leaks:
+ - /lib/src/file_with_hardcoded_api_key.dart
+ - /test/localhost_certificates/*.pem
+{% endprettify %}
```
jonasfj, line 300: @mit-mit what do you think about introducing the `false_leaks` key in `pubspec.yaml`?

kwalrath: That name LGTM. I'll have some tweaks to the text later, but they should wait until the technical details are all settled.

mit-mit: Hmm, not sure I really like the name `false_leaks`. It just sounds wrong somehow. Did we have any alternative names? Maybe reverse it and use something like `allowed_secrets`?

jonasfj: naming ideas:
 * `false_leaks` (because these are _false positives found by leak detection_)
 * `allowed_secrets`
 * `ignored_leaks`

So could be `(allowed|ignored|false)_(leaks|secrets)`.

I started leaning towards `false_leaks`, because what we're asking you to do is to explicitly state that these files do not _leak any secrets_, as in _false-positive leaks_.

The term `allowed_secrets` or `ignores_secrets` implies that _there are secrets present_. This is not necessarily the case, it may just be a false-positive match. Indeed, once published you can argue that by the nature of the package being public, its contents cannot contain a secret (or at-least _the secret_ is no-longer secret).

I do wonder if there is any value in include this ignore list in the `pubspec.yaml`, it has the downside that it will likely increase API response sizes for package version listings. I'm not too worried about that, because this feature will rarely be used, and we could filter it out of these API responses if we really want to (or ever need to).

Making the list of "ignored false-positive leaks detected" part of the package, whether by having it in `pubspec.yaml` or another file that is included in the package, makes sense to me. Because if a security researcher finds a leaked API key in a published package, a list such as `false_leaks` is a strong indicator that the package author actually intended to publish this API key. Hence, we can reasonably assume that the API has been locked down appropriately.

jonasfj: let's go with `false_secrets`

sigurdm (approved):
LGTM

src/tools/pub/pubspec.md
```diff
@@ -272,6 +277,39 @@ publish_to: none
 {% endprettify %}
 
 
+### False_secrets
+
+When publishing leak detection on the client will abort publishing if it finds
+potential leaks of secret credentials, API keys and cryptographic keys in files
+to be published.
+While detection of possible secrets relies on advanced heuristics using regular
```
sigurdm, line 285: Maybe 'advanced' sounds a bit too much for basically running though a bunch of regexps and doing a bit of checking on matches.

src/tools/pub/pubspec.md
```diff
@@ -272,6 +277,39 @@ publish_to: none
...
+[git-ignore pattern][git-ignore-format] with slash `/` ensures that is it only
+considered relative to the root folder.
+
+{% prettify yaml tag=pre+code %}
+false_secrets:
+ - /lib/src/file_with_hardcoded_api_key.dart
+ - /test/localhost_certificates/*.pem
+{% endprettify %}
+
+{{site.alert.version-note}}
```
kwalrath, line 303: This seems like a warning, not a version note. Shouldn't we have both?

kwalrath (approved):
LGTM. @jonasfj are my changes all right by you?

kwalrath:
Staged: https://dartlang-org-staging-0.web.app/tools/pub/pubspec#false_secrets

This page had some changes, so I staged the merged version (a clean merge).

src/tools/pub/pubspec.md
jonasfj: We do have entropy requirements, but just saying heuristics is fine.

jonasfj:
> @jonasfj are my changes all right by you?

Yeah, this looks great :D

src/tools/pub/pubspec.md
```diff
@@ -272,6 +278,50 @@ publish_to: none
...
+When you try to [publish a package][],
+pub conducts a search for potential leaks of
+secret credentials, API keys, or cryptographic keys.
+If pub detects a potential leak in a file that would be published,
+then pub warns you and refuses to publish the package.
+
+Leak detection isn't perfect.
+To avoid false positives,
+you can tell pub not to search for leaks in certain files,
+using [`gitignore` patterns][] under
```
mit-mit, line 292: I suggest changing:

"you can tell pub not to search for leaks in certain files,
using [`gitignore` patterns][]"

To:

"you can tell pub not to search for leaks in certain files,
by creating an allowlist
using [`gitignore` patterns][]"

mit-mit: @jonasfj WDYT?

mit-mit: I'll send a separate PR for that potential change

mit-mit: https://github.com/dart-lang/site-www/pull/3704

PR #3413, opened 2021-07-22
Update AOT compile detail on dart.dev

mit-mit:


kwalrath (approved):

parlough (commented):
Are the 1x and 2x pictures intentionally different? 

If so, in the one with "completed in just 8ms" on its own line, that line kind of seems to not be part of the same output/comment due to being on the next line without any comment marker. 

Screenshot from 2021-07-22 10-40-46

mit-mit:
> Are the 1x and 2x pictures intentionally different?

No, that was a good catch. Working on getting fixed images.

PR #3417, opened 2021-07-22
Add info about only being able to include a single file.

gspencergoog:


src/_guides/language/analysis-options.md
```diff
@@ -126,7 +126,8 @@ The sample illustrates the most common top-level entries:
 
 - Use <code>include: <em>url</em></code> to
   bring in options from the specified URL — in this case,
-  from a file in the `lints` package.
+  from a file in the `lints` package. Because YAML doesn't allow duplicate
+  keys, only one file or package may be included.
```
kwalrath, line 130: I didn't realize — thanks for the addition! I've tweaked the line breaks and wording a bit, so please check my changes. 

Also... can you really include a package? I assumed you could include only a `.yaml`-formatted file/resource.

```suggestion
  from a file in the `lints` package.
  Because YAML doesn't allow duplicate keys,
  you can include at most one file.
```

gspencergoog: Thanks!

> Also... can you really include a package? I assumed you could include only a .yaml-formatted file/resource.

You can! Well, technically, you can only include a YAML file _from_ a package, but it's the same as including a `.dart` file from a package in a Dart import.  It's what we do so people can use the `flutter_lints` package:

```
include: package:flutter_lints/flutter.yaml
```

kwalrath: Right, like https://dart.dev/guides/libraries/create-library-packages#importing-library-files. I think of `package:` as being a special URI scheme, but you're really importing the file/resource/library that's in the appropriate format. So "importing a package" sounds wrong to me, in either Dart import or YAML import usage.

gspencergoog: Yeah, I agree, it's not technically accurate to say "importing a package" when you're only really importing one file from the package.

However, if you import the main file from the package (e.g. in a Dart package, the eponymous package file: "package:foo/foo.dart") people think of it as importing the whole package. I'm not sure it's a distinction that we have to call out, though, it might just get confusing.

kwalrath (approved):
Thanks, @gspencergoog!

Issue #3399, opened 2021-07-16
Describe .pubignore


jjangga0214:
Hi :)

Through https://github.com/dart-lang/pub/pull/2787, `.pubignore` is introduced.

I think it should be explained from https://dart.dev/tools/pub/publishing.

Thanks!

parlough:
Thanks for the suggestion @jjangga0214, definitely would be nice to see this documented :)

QiXi:
does .pubignore currently support pub.dev?

jjangga0214:
@QiXi Accurately speaking, as far as I know, `.pubignore` does not have a direct relationship with pub.dev.
`pub` ( = `dart pub`) filters by `.pubignore` and `.gitignore` from publishing content _before publishing_. Then pub.dev receives 'unfiltered' files, so pub.dev does not know .pubignore.

QiXi:
I was faced with the task of publishing some files to git, but not allowing them to be published on pub.dev, after which a search brought me to topics with `.pubignore`. Perhaps I did not formulate the question well, but I would like to know if I can use it today or is it still under development?

parlough:
I don't think `.pubignore` functionality is available in the current stable builds of the SDK, but it should be available in the latest dev and beta builds of the SDK. Keep in mind though that this may result in changes in future behavior until the 2.14 release is made stable. You can find the changelog entry at the top of the 2.14 pub subsection: 3c14d86a67db7207bbc9f654ac49ee60e08e5240

QiXi:
@parlough thanks for the answer

kwalrath:
Closed in #3418

PR #3418, opened 2021-07-23
Document .pubignore

jonasfj:
Fixing https://github.com/dart-lang/site-www/issues/3399

Note: `.pubignore` support is set to ship in 2.14

### DIFF
--- a/src/tools/pub/publishing.md
+++ b/src/tools/pub/publishing.md
@@ -210,22 +210,27 @@ Here's how to transfer a package to a verified publisher:
 
 ## What files are published?
 
-**All files** in your package are included in the published package, with
-the following exceptions:
+**All files** in your local package root folder are included in the published
+package, with the following exceptions:
 
-* Any `packages` directories.
-* Your package's [lockfile](/tools/pub/glossary#lockfile).
-* If you aren't using Git, all _hidden_ files (that is,
-  files whose names begin with `.`).
-* If you're using Git, any files ignored by your `.gitignore` file.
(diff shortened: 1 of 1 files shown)


jonasfj:
Feel free to edit it... it's a little hard to explain concisely.

If we want to do better maybe we should make a separate page where we illustrate how it works using a few examples.

In short: we respect `.gitignore`, unless a folder contains a `.pubignore` then we allow the `.pubignore` to override the `.gitignore` files.

The hard part is the that it's not just the root folder that can contain a `.gitignore` or `.pubignore` file. And ignores from all parent folders under the package root-folder is applied to determine if a file/folder is ignored. For each parent folder we consider `.pubignore` (or `.gitignore` if the given folder doesn't contain a `.pubignore` file).

parlough:
Do you think it would be worth adding a version note or some mention of when this was introduced, to avoid confusion by users who are not using 2.14+ yet?

kwalrath:
I did a rewrite (of my words as much as Jonas's) and staged my version here:

https://kw-www-dartlang-1.web.app/tools/pub/publishing#what-files-are-published

We don't _usually_ publish docs before the stable release, so we might want to hold onto this for a while. Or not. At least there's a version note...

kwalrath:
@jonasfj could you please make this a draft, so we don't accidentally push it early? Also, are my changes OK with you?

kwalrath:
@jonasfj just checking again... are my changes OK by you?

jonasfj:
> just checking again... are my changes OK by you?

@kwalrath LGTM!

kwalrath (approved):

PR #3419, opened 2021-07-23
Fix image selection in narrow view

mit-mit:


parlough (approved):
Works well and fixes the issue for me when running locally, thanks.

PR #3420, opened 2021-07-23
Fix skipWhile explanation and update surrounding snippets

### LINKED ISSUE #2419: 'Iterable collections' page issue skipWhile() explanation
Page URL: https://dart.dev/codelabs/iterables.html
Page source: https://github.com/dart-lang/site-www/tree/master/src/codelabs/iterables.md

Description of issue:
> On the other hand, skipWhile() returns an Iterable while skipping all the elements before the one that satisfies the predicate. Note that the element that satisfies the predicate is also included. 
>
The explanation is little confusing and possibly misleading.  **_Note that the element that satisfies the predicate is also included._** is incorrect.
Elements for predicate true are skipped but **remaining collection of elements are returned starting from and including the first element for which test resulted in false**. 

Referred from the explanation on [dart:core skipWhile()](https://api.dart.dev/stable/2.8.2/dart-core/Iterable/skipWhile.html)
> The returned iterable provides elements by iterating this iterable, but skipping over all initial elements where test(element) returns true. If all elements satisfy test the resulting iterable is empty, otherwise it iterates the remaining elements in their original order, starting with the first element for which test(element) returns false.
>

parlough:
Fixes #2419


parlough:
@kwalrath When you get a chance, could you take a look at this one? Thanks!

kwalrath (approved):
Can't believe we had this wrong all this time. Thanks for the fix, @parlough!

PR #3426, opened 2021-07-27
Add details for redirecting constructors

nshahan:
I thought it was confusing there was no mention of the fact that the redirecting constructor syntax requires accessing the other constructor using `this`. WDYT? I'm totally open to rewording.

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -3050,7 +3050,9 @@ void main() {
 
 Sometimes a constructor’s only purpose is to redirect to another
 constructor in the same class. A redirecting constructor’s body is
-empty, with the constructor call appearing after a colon (:).
+empty, with the constructor call appearing after a colon (:). The
+constructor being redirected to must be accessed using the `this`
+keyword.
 
 <?code-excerpt "misc/lib/language_tour/classes/point_redirecting.dart"?>
 ```dart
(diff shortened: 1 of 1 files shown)


src/_guides/language/language-tour.md
```diff
@@ -3050,7 +3050,9 @@ void main() {
 
 Sometimes a constructor’s only purpose is to redirect to another
 constructor in the same class. A redirecting constructor’s body is
-empty, with the constructor call appearing after a colon (:).
+empty, with the constructor call appearing after a colon (:). The
+constructor being redirected to must be accessed using the `this`
+keyword.
```
kwalrath, line 3055: Yep, our laconic style is perhaps a bit too terse here. Since there are multiple uses of `this` in the following example, how about something like (wait for it) this:

```suggestion
empty, with the constructor call
(using `this` instead of the class name)
appearing after a colon (:).
```

We could also highlight `: this(x, 0);`. I can figure out the code for that if you think it's a good idea.

munificent (approved):

src/_guides/language/language-tour.md
nshahan: Thanks I like your wording. I don't think we necessarily need to highlight that code.

nshahan:
Feel free to merge, I'm not authorized to.

PR #3427, opened 2021-07-27
Minor improvements to Iterables codelab

parlough:
Aligns better to recommended lints, formatting, and good standards to avoid dynamic.

Also fixes some textual formatting and wording.

This is to separate update work from adding code excerpts, which I will submit after this lands.


src/codelabs/iterables.md
```diff
@@ -972,12 +970,12 @@ which has a constructor that takes a string.
 Another provided function is `isValidEmailAddress()`,
 which tests whether an email address is valid.
 
-|----------------------+-------------------+----------------------------------|
-| Constructor/function | Type signature    | Description                       |
-|----------------------|-------------------|----------------------------------|
-| EmailAddress() | `EmailAddress(String address)` | Creates an `EmailAddress` for the specified address. | 
+|-----------------------+------------------------------------------+---------------------------------------------------------|
```
parlough, line 973: I know formatting this isn't necessary, but I find it easier to read and edit this way.

kwalrath (approved):
LGTM! (And TIL that hyphenating _error-prone_ appears to be acceptable in all uses.)

PR #3431, opened 2021-07-28
Adding and updating excerpts for cheatsheet - Part 1

parlough:


src/codelabs/dart-cheatsheet.md
```diff
@@ -629,8 +634,9 @@ Using cascades, you can chain together operations that
 would otherwise require separate statements.
 For example, consider this code:
 
+<?code-excerpt "misc/bin/cheatsheet/cascades.dart (query-without-cascades)"?>
 ```dart
-var button = querySelector('#confirm');
+var button = querySelector('#confirm')!;
```
kwalrath, line 639: I feel like adding the `!` muddies this a bit. Should we add a null check?

kwalrath: Or maybe we could change this to use some API that always returns a non-null value.

parlough: You have a good point, any suggestions for a core API which would be good to use?

We could also instead use the conditional cascade to showcase an even larger benefit of cascades:

```dart
// #docregion query-without-cascades
var button = querySelector('#confirm');
button?.text = 'Confirm';
button?.classes.add('important');
button?.onClick.listen((e) => window.alert('Confirmed!'));
// #enddocregion query-without-cascades

// #docregion query-with-cascades
querySelector('#confirm')
  ?..text = 'Confirm'
  ..classes.add('important')
  ..onClick.listen((e) => window.alert('Confirmed!'));
// #enddocregion query-with-cascades
```

kwalrath: > We could also instead use the conditional cascade to showcase an even larger benefit of cascades:

I love that! We'll need to explain it a bit, but that would be a totally cool, useful thing to show.

parlough: Okay, I tried writing something, based off the language tour which I found already uses a similar snippet. I wasn't sure the best way to word/structure it though, so I would be happy to have some feedback. Thanks!

kwalrath (approved):
LGTM! We should probably have an engineer review the code, just to be sure.

Maybe @domesticmouse?

src/codelabs/dart-cheatsheet.md
kwalrath: For some reason, GH wouldn't let me make suggestions, so I just edited the text. (Nothing major, just tweaks.) The example looks great to me.

domesticmouse (approved):

PR #3433, opened 2021-07-29
Clean up links to `dart` commands

### LINKED ISSUE #2719: Change URLs for command-line tools being replaced by `dart`

kwalrath:
Fixes #2719

Staged at URLs like https://kw-dartlang-3.web.app/tools/pub/package-layout#public-tools


parlough (approved):
Nice improvements, LGTM! The new `dart run` page is great so updating these links is a good step.

PR #3438, opened 2021-08-02
Auto include and test remaining testable Iterables code snippets

parlough:


kwalrath (commented):
LGTM, but I'd like to have an engineer review the code.

examples/iterables/test/iterables_test.dart
```diff
@@ -56,6 +56,51 @@ void main() {
...
+      // #docregion first-where-long
+      bool predicate(String element) {
+        return element.length > 5;
+      }
+
+      void main() {
+        const items = ['Salad', 'Popcorn', 'Toast', 'Lasagne'];
+
+        // You can find with a simple expression:
+        var element1 = items.firstWhere((element) => element.length > 5);
```
domesticmouse, line 69: s/element/item/ because the group is `items`?

domesticmouse (approved):

examples/iterables/test/iterables_test.dart
parlough: Good suggestion, thanks. Makes it better match similar, lengthier examples. How does my most recent commit look?

kwalrath (approved):

PR #3445, opened 2021-08-03
Updated directions for installation in _windows.md

### LINKED ISSUE #1073: 'Dart SDK' choco installation confusing. Backed out.
Hey guys,

Just a heads up, the Windows `choco` installation of Dart, cited here: https://www.dartlang.org/tools/sdk, does not add any Dart tools to the path. Even if this is intended, **it is very difficult to find where exactly Dart tools are installed**. I had to remove Dart in the end because of this and use the normal .msi installer.

Little things like this, that can be easily fixed or clarified in the instructions hurts overall adoption.

mohitsaxenaknoldus:
Changes:

1. Added location of default installation directory.
2. Added information about the dart-sdk package and how to change the default installation directory while installing using choco.
3. Added steps to set path.

Fixes #1073

### DIFF
--- a/src/tools/sdk/_windows.md
+++ b/src/tools/sdk/_windows.md
@@ -2,12 +2,18 @@ How to install?
 
 You can install the Dart SDK using [Chocolatey.][Chocolatey]
 
-{{site.alert.important}} These commands require administrative privileges. 
-To start command prompt as an administrator following these steps:
-1. Press <kbd>Windows+R</kbd> to open the **Run** window. 
-2. Type “cmd” into the box. 
-3. Press <kbd>Ctrl+Shift+Enter</kbd> to run the command as an administrator.
+{{site.alert.important}} 
+  These commands require administrative privileges. 
 
+  For example, to start command prompt as an administrator,
+  follow these steps:
+  
+  1. Press <kbd>Windows+R</kbd> to open the **Run** window. 
+  
+  2. Type `cmd` into the box. 
+  
+  3. Press <kbd>Ctrl+Shift+Enter</kbd> to run the command as an administrator.
+  
 {{site.alert.end}}
 
 To install the Dart SDK:
--- a/src/tools/sdk/_windows.md
+++ b/src/tools/sdk/_windows.md
@@ -1,41 +1,48 @@
-###How to install?
+How to install?
 
 You can install the Dart SDK using [Chocolatey.][Chocolatey]
 
-> **_NOTE:_** These commands require administrative privileges. To start command prompt as an administrator following these steps:
-> 
-> - Press Windows+R to open the “Run” box. 
-> - Type “cmd” into the box 
-> - Press Ctrl+Shift+Enter to run the command as an administrator.
+{{site.alert.important}} These commands require administrative privileges. 
+To start command prompt as an administrator following these steps:
+1. Press <kbd>Windows+R</kbd> to open the **Run** window. 
+2. Type “cmd” into the box. 
+3. Press <kbd>Ctrl+Shift+Enter</kbd> to run the command as an administrator.
 
-####To install the Dart SDK:
+{{site.alert.end}}
+
+To install the Dart SDK:
 
 ```terminal
 C:\> choco install dart-sdk
 ```
 
-#####To upgrade the Dart SDK:
+To upgrade the Dart SDK:
 
 ```terminal
 C:\> choco upgrade dart-sdk
 ```
 
-####Default installation directory:
+Default installation directory:
 
 ```terminal
 C:\tools\dart-sdk
 ```
 
-###How to change installation directory?
+How to change installation directory?
 
-This package uses the help Get-BinRoot, which means that you can set an environment variable called ChocolateyToolsLocation, which controls where this package is extracted to. For more information click <a ahref=https://docs.chocolatey.org/en-us/create/functions/get-toolslocation>here</a>.
+This package uses the help Get-BinRoot, which means that 
+you can set an environment variable called ChocolateyToolsLocation, which 
+controls where this package is extracted to. 
+For more information check out [get-toolslocation].
 
-###How to set path?
+How to set path?
 
-- In the Windows search bar type env, you will see "edit the system environment variables" option, open it
-- Click on "Environment Variables...".
-- Under user variables, select "Path" and click edit.
-- Click new and add the path to your dart-sdk folder.
-- Press ok and apply.
+1. In the Windows search bar type "env". 
+2. Click on **edit the system environment variables**.
+3. Click on **Environment Variables...**.
+4. Under user variables, select **Path** and click edit.
+5. Click **New..** and add the path to your dart-sdk folder.
+6. Press **OK** and **Apply**.
 
 [Chocolatey]: https://chocolatey.org
+[get-toolslocation]: https://docs.chocolatey.org/en-us/create/functions/get-toolslocation
--- a/src/tools/sdk/_windows.md
+++ b/src/tools/sdk/_windows.md
@@ -1,17 +1,13 @@
 You can install the Dart SDK using [Chocolatey.][Chocolatey]
 
-{{site.alert.important}} 
-  These commands require administrative privileges. 
+{{site.alert.important}}
+  These commands require administrator rights.
+  Here's one way to open a Command Prompt window
+  that has admin rights:
 
-  For example, to start command prompt as an administrator,
-  follow these steps:
-  
-  1. Press <kbd>Windows+R</kbd> to open the **Run** window. 
-  
-  2. Type `cmd` into the box. 
-  
-  3. Press <kbd>Ctrl+Shift+Enter</kbd> to run the command as an administrator.
-  
+  1. Press <kbd>Windows+R</kbd> to open the **Run** window.
+  2. Type `cmd` into the box.
+  3. Press <kbd>Ctrl+Shift+Enter</kbd>.
 {{site.alert.end}}
 
 To install the Dart SDK:
@@ -27,23 +23,20 @@ C:\> choco upgrade dart-sdk
 ```
 
 By default, the SDK is installed at `C:\tools\dart-sdk`.
+You can change that location by setting
+the [`ChocolateyToolsLocation`][] environment variable
+to your chosen installation directory.
 
-To change the default installation location:
+If you can't use the Dart SDK executables,
+add the SDK location to your PATH:
 
-Set the `ChocolateyToolsLocation` environment variable
-to the desired extraction location.
-To read more about how the installer uses this configuration, 
-see [Chocolatey's Get-ToolsLocation reference][get-toolslocation].
-
-If you can't use the Dart executables,
-add the extraction location to your PATH:
-
-1. In the Windows search bar type "env". 
-2. Click on **edit the system environment variables**.
-3. Click on **Environment Variables...**.
-4. Under user variables, select **Path** and click edit.
-5. Click **New..** and add the path to your dart-sdk folder.
-6. Press **OK** and **Apply**.
+1. In the Windows search box, type `env`.
+2. Click **Edit the system environment variables**.
+3. Click **Environment Variables...**.
+4. In the user variable section, select **Path** and click **Edit...**.
+5. Click **New**, and enter the path to the `dart-sdk` directory.
+6. In each window that you just opened,
+   click **Apply** or **OK** to dismiss it and apply the path change.
 
 [Chocolatey]: https://chocolatey.org
-[get-toolslocation]: https://docs.chocolatey.org/en-us/create/functions/get-toolslocation
+[`ChocolateyToolsLocation`]: https://stackoverflow.com/questions/19752533/how-do-i-set-chocolatey-to-install-applications-onto-another-drive/68314437#68314437
--- a/src/tools/sdk/_windows.md
+++ b/src/tools/sdk/_windows.md
@@ -26,18 +26,14 @@ To upgrade the Dart SDK:
 C:\> choco upgrade dart-sdk
 ```
 
-Default installation directory:
+By default, the SDK is installed at `C:\tools\dart-sdk`.
 
-```terminal
-C:\tools\dart-sdk
-```
+To change the default installation location:
 
-How to change installation directory?
-
-This package uses the help Get-BinRoot, which means that 
-you can set an environment variable called ChocolateyToolsLocation, which 
-controls where this package is extracted to. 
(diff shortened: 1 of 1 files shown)


mohitsaxenaknoldus:
@googlebot I signed it!

src/tools/sdk/_windows.md
```diff
@@ -1,23 +1,41 @@
+###How to install?
```
parlough, line 1: In this header, and others make sure to put a space after the final hash(`#`) or you'll end up with headers such as these:

image

Also, I'm not sure we want to use headers within these OS specific instructions anyway, as then they will show up in the TOC even if a user does not have the Windows tab open. Then when clicking these with the Mac or Linux tab open, these TOC entries will not work(at least not properly). So perhaps we want to stick with the similar structure as before.

image

src/tools/sdk/_windows.md
```diff
@@ -1,23 +1,41 @@
...
 You can install the Dart SDK using [Chocolatey.][Chocolatey]
 
-{{site.alert.important}}
-These commands require administrator privileges.
-If you need help on starting an administrator-level command prompt,
-try a search like
-<em><a href="https://www.google.com/search?q=cmd+admin"
-  target="blank">cmd admin</a>.</em>
-{{site.alert.end}}
+> **_NOTE:_** These commands require administrative privileges. To start command prompt as an administrator following these steps:
```
parlough, line 5: I think the previous `{{site.alert.important}}` structure matches the rest of the website better, but if you think it's highlighted too much you can use `{{site.alert.note}}` instead which introduces the Note text and icon.

src/tools/sdk/_windows.md
```diff
@@ -1,23 +1,41 @@
...
+> 
+> - Press Windows+R to open the “Run” box. 
+> - Type “cmd” into the box 
+> - Press Ctrl+Shift+Enter to run the command as an administrator.
```
parlough, line 9: I agree that telling the user to search isn't the best solution, but I'm not sure this is as users may not want to use the old style run command prompt and may be using something like the new Windows Terminal.  I'll give some thought on how we can perhaps make this more generic.

parlough, line 9: For formatting keyboard keys and combinations, use `<kbd>` or backticks (`) as specified in https://developers.google.com/style/ui-elements?hl=en#keyboard

src/tools/sdk/_windows.md
```diff
@@ -1,23 +1,41 @@
...
 ```terminal
 C:\> choco upgrade dart-sdk
 ```
 
-[Chocolatey]: https://chocolatey.org
+####Default installation directory:
+
+```terminal
+C:\tools\dart-sdk
+```
```
parlough, line 27: Where did you find this? I don't have a Windows PC at the moment to try this, but I found on [Chocolatey's FAQ](https://docs.chocolatey.org/en-us/faqs#where-does-chocolatey-install-by-default) that it by default installs at:

> As of version 0.9.8.24, binaries, libraries and Chocolatey components install in C:\ProgramData\chocolatey (environment variable %ProgramData%) by default.

src/tools/sdk/_windows.md
```diff
@@ -1,23 +1,41 @@
...
+
+This package uses the help Get-BinRoot, which means that you can set an environment variable called ChocolateyToolsLocation, which controls where this package is extracted to. For more information click <a ahref=https://docs.chocolatey.org/en-us/create/functions/get-toolslocation>here</a>.
+
+###How to set path?
+
+- In the Windows search bar type env, you will see "edit the system environment variables" option, open it
+- Click on "Environment Variables...".
+- Under user variables, select "Path" and click edit.
+- Click new and add the path to your dart-sdk folder.
+- Press ok and apply.
```
parlough, line 39: A few notes:

- If this is sequential, perhaps you can use an [ordered list](https://www.markdownguide.org/basic-syntax/#ordered-lists). 
- For buttons and other UI elements, put their names in bold(**) instead of quotes: https://developers.google.com/style/ui-elements?hl=en#formatting
- Also match the capitalization of the UI elements

src/tools/sdk/_windows.md
```diff
@@ -1,23 +1,41 @@
...
-[Chocolatey]: https://chocolatey.org
+####Default installation directory:
+
+```terminal
+C:\tools\dart-sdk
+```
+
+###How to change installation directory?
+
+This package uses the help Get-BinRoot, which means that you can set an environment variable called ChocolateyToolsLocation, which controls where this package is extracted to. For more information click <a ahref=https://docs.chocolatey.org/en-us/create/functions/get-toolslocation>here</a>.
```
parlough, line 31: For this section and other parts of the document, we try to stick to under 80 characters long per line and use semantic line breaks to accomplish this. You can read more about how to do this here: https://github.com/dart-lang/site-shared/blob/master/doc/writing-for-dart-and-flutter-websites.md#semantic-line-breaks

Let me know if you have any questions about doing so.

Also for links, use the [Markdown link syntax](https://www.markdownguide.org/basic-syntax/#links) and try to create a more thoughtful statement instead of "click here". You can learn more about this in the documentation for [References to other documents](https://developers.google.com/style/cross-references?hl=en#references-to-other-documents).

parlough (commented):
Hey @mohitsaxenaknoldus, thanks so much for working on this. Always nice to have more clear instructions. I've left some initial questions, comments, and suggestions on various pieces. Feel free to push back on anything and ask any questions about my feedback or the website, just tag me. 

I recommend, if possible, setting up the website so you can see what it looks like after building, but I understand if that is too many steps right now. If you would like to though, you can read about how to do so here: https://github.com/dart-lang/site-www#before-you-build-this-site

src/tools/sdk/_windows.md
parlough: Also I'm not super sure I understand the "help Get-BinRoot" piece of this instruction. What are you trying to convey there?

src/tools/sdk/_windows.md
mohitsaxenaknoldus: Sure, will update.

src/tools/sdk/_windows.md
mohitsaxenaknoldus: Ok i'll try that.

src/tools/sdk/_windows.md
mohitsaxenaknoldus: We can also add instructions for Windows Powershell which is the new Windows terminal. Although, only step 1 will change and rest will remain the same.

mohitsaxenaknoldus: Ok sure

src/tools/sdk/_windows.md
mohitsaxenaknoldus: Yes, that's the default path but not every package is installed there. In the same link, read under the heading: "Why doesn't a package install software to Program Files?". 

Apparently this is why everyone was having difficulty finding since everyone was trying to look inside program data/files.

I tested on my PC and it installed inside the tools folder.

src/tools/sdk/_windows.md
mohitsaxenaknoldus: Point 3 is already like that. I'll implement the first 2 pointers. Thanks!

src/tools/sdk/_windows.md
mohitsaxenaknoldus: Ok i'll do that.

And as for "Get-BinRoot", it's an inbuilt environment variable which Chocolatey uses for some packages which allows the user to change the installation directory. 

So our package dart-sdk uses that as I checked it on its page on Chocolatey. So that was just a minor detail about the package for people familiar with Chocolatey installations so that they find it easier to change the installation directory.

mohitsaxenaknoldus:
Sure! I'll gladly do it and let you know.

mohitsaxenaknoldus:
PR updated, please take a look @parlough

parlough:
Thanks @mohitsaxenaknoldus! I'll take a look later today :)

src/tools/sdk/_windows.md
```diff
@@ -1,11 +1,13 @@
...
-try a search like
-<em><a href="https://www.google.com/search?q=cmd+admin"
-  target="blank">cmd admin</a>.</em>
+{{site.alert.important}} These commands require administrative privileges. 
+To start command prompt as an administrator following these steps:
+1. Press <kbd>Windows+R</kbd> to open the **Run** window. 
+2. Type “cmd” into the box. 
+3. Press <kbd>Ctrl+Shift+Enter</kbd> to run the command as an administrator.
+
 {{site.alert.end}}
```
parlough, line 11: I'm thinking since there are multiple ways to run commands as an administrator, through the old PowerShell and command prompt UIs as well as with the new Windows Terminal, we can present this as one example. I've also added some extra formatting. How does this look?

```suggestion
{{site.alert.important}} 
  These commands require administrative privileges. 

  For example, to start command prompt as an administrator
  following these steps:
  
  1. Press <kbd>Windows+R</kbd> to open the **Run** window. 
  
  2. Type `cmd` into the box. 
  
  3. Press <kbd>Ctrl+Shift+Enter</kbd> to run the command as an administrator.
  
{{site.alert.end}}
```

src/tools/sdk/_windows.md
```diff
@@ -1,11 +1,13 @@
+How to install?
+
```
parlough, line 2: I think we'd be fine without this, as the section of the page this is in is called "Installing the Dart SDK". 

```suggestion
```

src/tools/sdk/_windows.md
parlough: Thanks for the context. I'm not sure all the extra details here are necessary for a user to properly install Dart and may actually cause confusion. So is the essential idea that users can configure the `ChocolateyToolsLocation` environment variable to change the install location? If so, perhaps we can simply state something about that, and after that direct users with a statement like "For more information, see [Chocolatey's GET-TOOLSLOCATION documentation](https://docs.chocolatey.org/en-us/create/functions/get-toolslocation).".

src/tools/sdk/_windows.md
```diff
@@ -14,10 +16,33 @@ To install the Dart SDK:
...
+```
+
+How to change installation directory?
+
+This package uses the help Get-BinRoot, which means that 
+you can set an environment variable called ChocolateyToolsLocation, which 
+controls where this package is extracted to. 
+For more information check out [get-toolslocation].
+
+How to set path?
```
parlough, line 38: As a clarifying question before I review this section further, does choco not automatically add packages to the path?

src/tools/sdk/_windows.md
mohitsaxenaknoldus: sure

src/tools/sdk/_windows.md
mohitsaxenaknoldus: looks great!

src/tools/sdk/_windows.md
mohitsaxenaknoldus: It does, but for some cases (as reported in the original issue) it was not being added. So I added this information as well just in case someone struggles with it. But yes, we can put this under a troubleshooting heading.

src/tools/sdk/_windows.md
```diff
@@ -14,10 +20,33 @@ To install the Dart SDK:
...
+```terminal
+C:\tools\dart-sdk
+```
+
+How to change installation directory?
+
+This package uses the help Get-BinRoot, which means that 
+you can set an environment variable called ChocolateyToolsLocation, which 
+controls where this package is extracted to. 
+For more information check out [get-toolslocation].
```
parlough, line 40: Okay, I gave this section some thought and perhaps we should use more of a prose structure to describe these pieces while also avoiding some of the technical details about Chocolatey that may confuse readers. How about something like this, assuming my understanding of the `ChocolateyToolsLocation` environment variable is correct.

```suggestion
By default, the SDK is installed at `C:\tools\dart-sdk`.

To change the default installation location:

Set the `ChocolateyToolsLocation` environment variable
to the desired extraction location.
To read more about how the installer uses this configuration, 
see [Chocolatey's Get-ToolsLocation reference][get-toolslocation].
```

mohitsaxenaknoldus: Yes that's correct! Definitely looks more clear now.

src/tools/sdk/_windows.md
parlough: Thanks for including this then, it will definitely be helpful. 

Instead of starting this section with a question, how about we switch to some sort of troubleshooting statement like you mentioned. I'm not exactly sure how to structure or state this, do you have any ideas? It could be something basic like:

```suggestion
If you don't have access to the Dart executables,
add the extraction location to your PATH:
```

parlough:
Hey @mohitsaxenaknoldus, thanks so much for your help and patience so far! It's turning out to be a great addition. Once the final pieces are addressed, I'll add some final reviewers to review the Chocolatey functionality and someone to review the writing changes.

src/tools/sdk/_windows.md
mohitsaxenaknoldus: Yes, looks cleaner now.

mohitsaxenaknoldus:
> Hey @mohitsaxenaknoldus, thanks so much for your help and patience so far! It's turning out to be a great addition. Once the final pieces are addressed, I'll add some final reviewers to review the Chocolatey functionality and someone to review the writing/structural changes.

I think I've addressed all the pieces. Let me know if I missed something.

kwalrath (approved):

kwalrath:
I made some copy edits and staged the page here:

https://dartlang-org-staging-0.web.app/get-dart

I consulted the [Microsoft style guide](https://docs.microsoft.com/en-us/style-guide) for terms like _rights_, _search box_, and _Command Prompt window_, but I probably missed some. I checked the PATH instructions using a housemate's PC; I was surprised by the number of **OK** buttons I needed to click — at least 3! (**Apply** was grayed out the one place I saw it.)

We could probably be more helpful by giving an example of the path value. For example, would it be `C:\tools\dart-sdk` if the SDK is in its default location?

Also, we could say something about having to open a new command prompt to get the new path — if that's true.

But I suspect this is helpful enough. We can always iterate!

mohitsaxenaknoldus:
> Also, we could say something about having to open a new command prompt to get the new path — if that's true.

This won't be required since changing the path is a pre-install thing (before running `choco install` command). This shouldn't affect the command prompt even if someone already has one open in a rare case.

kwalrath:
Merged, and should be visible within minutes. Thank you, @mohitsaxenaknoldus and @parlough!

PR #3449, opened 2021-08-04
Discourage use of runtimeType

### LINKED ISSUE #3354: 'Effective Dart: Usage' - discourage use of .runtimeType
/cc @munificent 

Every now and then a user runs into a problem because they believe they can freely use `.runtimeType`, but in practice it is an API that needs to be used with a lot of caution. See https://github.com/flutter/flutter/issues/78041

What are your thoughts on using the guide to educate about this issue? While it would be great to fix the language and libraries to avoid such pitfalls (see https://github.com/dart-lang/sdk/issues/46373), maybe our effective guide can encourage positive patterns?

Two requests:
* [ ] Add a new guidance to encourage `is` tests instead of `.runtimeType == string`, and emphasize that the string representation is unstable in production environments, like `dart2js -O2` or higher.
* [ ] Remove the use of `runtimeType` from the [List.from example][1].  And change it to use the recommended pattern. For example:
```dart
// Prints: List<int>
print(iterable.toList().runtimeType);
```
is updated to:
```dart
// Prints true
print(iterable.toList is List<int>);
```
and similarly the bad case as
```dart
// Prints false because the type is a List<dynamic> instead
print(List.from(iterable) is List<int>);
```

Thoughts? 

[1]: https://dart.dev/guides/language/effective-dart/usage#dont-use-listfrom-unless-you-intend-to-change-the-type-of-the-result

micpap25:
Fixes #3354
Follows the guidelines listed, encourages better use of runtimeType.

### DIFF
--- a/examples/misc/test/effective_dart_test.dart
+++ b/examples/misc/test/effective_dart_test.dart
@@ -30,8 +30,8 @@ void main() {
         // Creates a List<int>:
         var iterable = [1, 2, 3];
 
-        // Prints true
-        print(iterable.toList() is List<int>);
+        // Prints "List<int>":
+        print(iterable.toList().runtimeType);
         // #enddocregion list-from-good
       }, prints('List<int>\n'));
 
@@ -40,8 +40,8 @@ void main() {
         // Creates a List<int>:
         var iterable = [1, 2, 3];
 
-        // Prints false because the type is a List<dynamic> instead
-        print(List.from(iterable) is List<int>);
+        // Prints "List<dynamic>":
+        print(List.from(iterable).runtimeType);
         // #enddocregion list-from-bad
       }, prints('List<dynamic>\n'));
     });
--- a/examples/misc/test/effective_dart_test.dart
+++ b/examples/misc/test/effective_dart_test.dart
@@ -30,8 +30,8 @@ void main() {
         // Creates a List<int>:
         var iterable = [1, 2, 3];
 
-        // Prints "List<int>":
-        print(iterable.toList().runtimeType);
+        // Prints true
+        print(iterable.toList() is List<int>);
         // #enddocregion list-from-good
       }, prints('List<int>\n'));
 
@@ -40,8 +40,8 @@ void main() {
         // Creates a List<int>:
         var iterable = [1, 2, 3];
 
-        // Prints "List<dynamic>":
-        print(List.from(iterable).runtimeType);
+        // Prints false because the type is a List<dynamic> instead
+        print(List.from(iterable) is List<int>);
         // #enddocregion list-from-bad
       }, prints('List<dynamic>\n'));
     });
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -2766,6 +2766,12 @@ which returns a [`Type`][] object.
 print('The type of a is ${a.runtimeType}');
 ```
 
+{{site.alert.warn}}
+  Use a [type test operator][] rather than `runtimeType` to test an object's type.
+  In production environments, the test `object is string` is more stable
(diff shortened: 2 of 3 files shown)


parlough:
Looks like you'll need to update the analyzed examples as well: https://github.com/dart-lang/site-www/blob/master/examples/misc/test/effective_dart_test.dart

munificent:
I think we should leave the Effective Dart examples as they were before. It is *generally* a bad idea to use `runtimeType` in normal code, but in the example here we're using it specifically because we really do want to show the exact runtime type. Using `is` might confuse readers because then they have to think about the subtyping rules around generic types, which can be pretty confusing for many users.

examples/misc/test/effective_dart_test.dart
```diff
@@ -40,8 +40,8 @@ void main() {
         // Creates a List<int>:
         var iterable = [1, 2, 3];
 
-        // Prints "List<dynamic>":
-        print(List.from(iterable).runtimeType);
+        // Prints false because the type is a List<dynamic> instead
+        print(List.from(iterable) is List<int>);
```
munificent, line 44: I think we should leave the Effective Dart examples as they were before. It is generally a bad idea to use runtimeType in normal code, but in the example here we're using it specifically because we really do want to show the exact runtime type. Using is might confuse readers because then they have to think about the subtyping rules around generic types, which can be pretty confusing for many users.

micpap25:
@munificent Should I revert that or would you rather wait to hear from others?

munificent:
You can just revert that part. I think I'm enough of an authority on "Effective Dart" to make the call. :)

I do like your addition to the language tour.

micpap25:
Reverted.

src/_guides/language/language-tour.md
```diff
@@ -2766,6 +2766,12 @@ which returns a [`Type`][] object.
 print('The type of a is ${a.runtimeType}');
 ```
 
+{{site.alert.warn}}
+  Use a [type test operator][] rather than `runtimeType` to test an object's type.
+  In production environments, the test `object is string` is more stable
+  than the test `object.runtimeType == string`.
```
parlough, line 2772: What do you think about using `Type` here instead of `string`?

For example:

```suggestion
  Use a [type test operator][] rather than `runtimeType`
  to test an object's type.
  In production environments, the test `object is Type` is more stable
  than the test `object.runtimeType == Type`.
```

parlough (approved):

kwalrath (approved):

kwalrath:
Thanks, @micpap25!

examples/misc/test/effective_dart_test.dart
sigmundch: @munificent  - any thoughts on how to deliver the Effective Dart message without using `.runtimeType`?

I first suggested `is` because I was trying to imagine how dropping the type parameter will fail later in the program. My first thought was that it would fail when you try to assign it to a variable/parameter of the original type. So at that point you are already having to address subtyping rules to understand a cast failure.

Another point of failure is missing errors when you expected them (e.g. `List.from(iterable).add("invalid-value-type")`), but that may be more confusing.

munificent: Oh, dear. I'm sorry, I gave completely inconsistent feedback between here and on the bug. :(

I'll follow up here and on the bug.

munificent:
Hey, I'm sorry @micpap25. I gave you really inconsistent feedback. On the issue you fixed I said one thing and then said the exact opposite here. :( Too much task switching on my part, I guess.

I'm going to re-open the bug and I'll add a comment there with some suggestions for fixes we can make. I really appreciate you work here.

PR #3454, opened 2021-08-05
Fix misc examples analysis issue breaking build

parlough:
The order of diagnostics was standardized in https://github.com/dart-lang/sdk/commit/1689a82229d5264d250c0ca93ec19ec14e02c440 from error->hint->info.


kwalrath (approved):
Interesting! Thanks, @parlough

PR #3459, opened 2021-08-07
Fixes remaining linkcheck warnings by properly enabling no_toc_section

parlough:
This was causing cards not intended for the TOC on the samples and tutorials pages to show up in their respective TOC with no working target.

Previously(from latest commit to main branch):
image

Now:
image


kwalrath (approved):
Nice, thanks!

PR #3481, opened 2021-08-10
Analyze and test more cheatsheet snippets

### LINKED ISSUE #2295: Add and auto-include cheatsheet code samples
We need to add the cheatsheet code samples in the repo so that we can link the code-excerpts in the source file.

Context: #1803 

cc: @kwalrath

parlough:
Fixes #2295


kwalrath (commented):
LGTM! We should also get an engineer to review the code.

@domesticmouse? @johnpryan?

examples/misc/test/cheatsheet/arguments_test.dart
```diff
@@ -0,0 +1,46 @@
...
+    as optional_args;
+import 'package:examples/cheatsheet/optional_positional_args2.dart'
+    as defaulted_args;
+import 'package:test/test.dart';
+
+void main() {
+  test('print_name', () {
+    expect(() {
+      printNameTest();
+    }, prints('Avinash Gupta \nPoshmeister Moneybuckets IV\n'));
```
johnpryan, line 12: nit: I realize this is testing an already existing snippet, but this string might be too wacky / goofy: https://developers.google.com/style/tone

johnpryan (approved):

examples/misc/test/cheatsheet/arguments_test.dart
parlough: That's fair, we don't want the examples to introduce unnecessary confusion or distraction. There's a few examples like this and I'm not sure of the best replacement at the moment, so how about we move this to a separate, larger issue to consider the voice/tone/global understanding of various examples. If not, do you have any short term suggestions for more understandable arguments I could switch these for.

johnpryan: Sounds good, let's consider this separately https://github.com/dart-lang/site-www/issues/3485

PR #3482, opened 2021-08-10
Remove daily news example

### LINKED ISSUE #3289: Remove daily_news example and test?
See #3281 for details.

parlough:
If we for some reason do need to reference it again, can always refer to Git history.

Fixes #3289


kwalrath (approved):
LGTM assuming the build doesn't break

PR #3488, opened 2021-08-13
Add dart:svg and dart:ffi

mit-mit:
Add `dart:ffi` to the core libraries table

Also, fix sorting in web table


mit-mit:
Add `dart:ffi` to the core libraries table

parlough (approved):
These changes look good to me, but taking a look at the sorted web libraries, I realized `dart:svg` is not present. Is this intentional?

mit-mit:
> I realized dart:svg is not present

I had an email out to enquire about that one, and just got confirmation that it indeed should be listed.

PR #3494, opened 2021-08-15
Removes extra blank line

fxn:
Easy one!

### DIFF
--- a/examples/misc/test/samples_test.dart
+++ b/examples/misc/test/samples_test.dart
@@ -11,8 +11,8 @@ import 'package:test/test.dart';
 // #enddocregion import
 import 'package:examples_util/print_matcher.dart' as m;
 import 'package:examples/samples/spacecraft.dart';
-// #docregion import
 
+// #docregion import
 // Importing files
 import 'path/to/my_other_file.dart';
 // #enddocregion import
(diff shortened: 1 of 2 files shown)


mit-mit:
Thanks @fxn, can you kindly sign the CLA?

fxn:
@mit-mit I did right after opening the PR, see image.

Screen Shot 2021-08-16 at 10 10 36

Does not square on your end?

parlough:
Make sure it's under the same email as the commit is and then reply to the PR with:

```
@googlebot I signed it!
```

parlough:
Thanks for this @fxn, I agree the extra line isn't necessary.

We have a system where code snippets come from testable examples, causing the build to fail as they no longer match. In this case, this snippet is using the (import) `#docregion` in [examples/misc/test/samples_test.dart](https://github.com/dart-lang/site-www/blob/master/examples/misc/test/samples_test.dart).

There's a few ways to remove the line from the docregion, but I think you could just swap [line 14 and 15](https://github.com/dart-lang/site-www/blob/master/examples/misc/test/samples_test.dart#L14-L15) putting the docregion comment directly about the import. 

Let me know if you have any issues or questions!

parlough (approved):
I went ahead and added the line to the example as well due to it being a simple change.

Thanks so much for catching this and submitting a fix, it's appreciated!

PR #3498, opened 2021-08-17
Include rule sets in linter rule descriptions

parlough:
- Differentiates between the `core` and `recommended` rule sets for linking purposes
- Adds links to the rule sets
- Add a list of linked rule sets to each linter rule
- Simplifies the includes to share more behavior across all rule types.
- Fixes extra space before comma in incompatible rules list


kwalrath (commented):
I started trying to work on the wording of the individual rule sets, but I think we might need to use the package-level structure instead. Otherwise we'll end up repeating ourselves (about using `flutter_lints`), and it isn't clear that `core` and `recommended` are in the same package.

What do you think of having the same "Predefined rule sets" structure as before, but beefing up the description of `lints`? Something like this:

`lints`
Rule sets curated by the Dart team. We recommend using at least the `core` rule set, which is used when scoring packages uploaded to pub.dev. Or, better yet, use the `recommended` rule set, a superset of `core` that identifies additional issues and enforces style and format. If you’re writing Flutter code, use the rule set in the `flutter_lints` package, which builds on `lints`.

parlough:
@kwalrath I originally split them up for link targets, but I think beefing up the `lints` description and linking to it for both sets seems reasonable while avoiding the issues you mentioned. Your modified description works well.

Thanks for the great suggestions. I'll make the modifications and see how it looks tomorrow :)

src/tools/linter-rules.md
```diff
@@ -30,35 +30,42 @@ To avoid the need to individually select compatible linter rules,
...
   which builds on `lints`.
-  Dart and Flutter packages uploaded to [pub.dev]({{site.pub}}) 
-  are [scored]({{site.pub}}/help/scoring) 
-  with the `core` set of these rules.
 
+<a id="flutter_lints"></a>
 [flutter_lints][]
-: The set of rules that the Flutter team encourages you to use
+: Contains the `flutter` set of rules
+  that the Flutter team encourages you to use
```
kwalrath, line 49:
```suggestion
: Contains the `flutter` rule set,
  which the Flutter team encourages you to use
```

kwalrath (approved):
LGTM! Just one suggestion.

PR #3502, opened 2021-08-18
Add constructor to interval example in understanding null safety

### LINKED ISSUE #2763: 'Understanding null safety' page issue

Description of issue:
In the part [nullability-and-generics](https://dart.dev/null-safety/understanding-null-safety#nullability-and-generics) is the the following example: 

```dart
// Using null safety:
class Interval<T extends num?> {
  T min, max;

  bool get isEmpty {
    var localMin = min;
    var localMax = max;

    // No min or max means an open-ended interval.
    if (localMin == null || localMax == null) return false;
    return localMax <= localMin;
  }
}
```
and the after that the statement "Note we have no constructor this time, and that’s OK.". 

That's seems to be not true. I tried the example on my machine and the initialization for the fields is required, regardless if it's `<T extends num>` or `<T extends num?>`. 

Thats seems right to me because a few sentence later it is mentioned that " a nullable bound does not prevent users from instantiating the class with non-nullable types" which would lead to an exception if the fields are not initialized. So I think the documentation is wrong here.  

I'm currently using Flutter 1.24.0-8.0.pre.252, channel master and Dart 2.12.0 (build 2.12.0-45.0.dev).

parlough:
This is because without the types being T? the fields aren't implicitly initialized to null.

Fixes #2763

### DIFF
--- a/src/null-safety/understanding-null-safety/index.md
+++ b/src/null-safety/understanding-null-safety/index.md
@@ -1350,6 +1350,8 @@ You can also use a nullable *bound*:
 class Interval<T extends num?> {
   T min, max;
 
+  Interval(this.min, this.max);
+
   bool get isEmpty {
     var localMin = min;
     var localMax = max;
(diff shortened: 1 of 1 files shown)


kwalrath:
I just pinged Bob via email.

kwalrath:
Bob was out, so I just sent another reminder.

munificent (approved):
This is great, thank you!

PR #3506, opened 2021-08-19
Switch Effective Dart linter rule links to the local versions

parlough:
What do you think about switching these linter rule links to the local version? Allows users to stay on the same website.

For example, this guide:

image

Will link to: https://dart.dev/tools/linter-rules#camel_case_extensions

Instead of: https://dart-lang.github.io/linter/lints/camel_case_extensions.html

Staged at https://kw-dartlang-3.web.app. E.g. [before](https://dart.dev/guides/language/effective-dart/style#do-name-extensions-using-uppercamelcase), [after](https://kw-dartlang-3.web.app/guides/language/effective-dart/style#do-name-extensions-using-uppercamelcase)

### DIFF
--- a/src/_guides/language/effective-dart/design.md
+++ b/src/_guides/language/effective-dart/design.md
@@ -367,7 +367,7 @@ previous guidelines state, either:
 
 ### PREFER naming a method `to___()` if it copies the object's state to a new object.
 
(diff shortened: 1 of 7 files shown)


pq (approved):
👍

kwalrath (approved):
LGTM

PR #3508, opened 2021-08-20
Update effective-dart-rules tool to null safety and fix new set of lints

parlough:


kwalrath (approved):
LGTM, but we should get an engineer to take a look in case I missed something.

parlough:
@domesticmouse Could you take a quick look when you get a chance? This is mostly just lint fixes. I appreciate it!

deploy/effective-dart-rules/bin/main.dart
```diff
@@ -9,53 +8,45 @@ import 'package:path/path.dart' as path;
...
-    "documentation.md",
-    "usage.md",
-    "design.md"
-  ];
+Future<void> main(List<String> arguments) async {
+  const dirPath = 'src/_guides/language/effective-dart';
+  const fileNames = ['style.md', 'documentation.md', 'usage.md', 'design.md'];
 
-  var sections = filenames.map((name) => Section(dirPath, name)).toList();
+  var sections = fileNames.map((name) => Section(dirPath, name)).toList();
```
domesticmouse, line 15: I reacted to this change. Turns out, I consider `filename` to be an actual word, not a compound phrase. 

See https://english.stackexchange.com/questions/5366/which-is-correct-filename-file-name-or-filename for some fun commentary.

deploy/effective-dart-rules/pubspec.yaml
```diff
@@ -1,9 +1,13 @@
 name: effective_dart_rules
-version: 0.0.2
+version: 0.0.3
 description: Builds a file with all the rules in the different sections of Effective Dart.
 environment:
-  sdk: '>=2.10.0 <3.0.0'
+  sdk: '>=2.13.0 <3.0.0'
```
domesticmouse, line 5: I'm curious if this code uses Dart 2.13 specific code, or if 2.12 would have been sufficient.

domesticmouse (approved):

deploy/effective-dart-rules/bin/main.dart
kwalrath: Good catch! We use one word: https://developers.google.com/style/word-list#filename

parlough: I forgot about this usage, thanks for catching this :)

Issue #1265, opened 2018-12-14
'A Tour of the Dart Libraries' page issue


feinstein:
Page URL: https://www.dartlang.org/guides/libraries/library-tour
Page source: https://github.com/dart-lang/site-www/tree/master/src/_guides/libraries/library-tour.md

Description:

The "Chaining multiple asynchronous methods" chapter says:

>If the callback registered with then() returns a Future, then() returns an equivalent Future.

This "equivalent Future" is not clear what it is, is it the same Future, or is it a copy that's equivalent, or is it something else, like a pattern, called "equivalent Future"?

PR #3510, opened 2021-08-21
Clarify mention of equivalent future in library tour

parlough:
"Equivalent Future" may not have been easily understand, so I tried to elaborate on what that meant. I may have been too verbose in doing so, but I tried to follow what's present in the Dartdoc as it seemed to explain it well. 

Closes https://github.com/dart-lang/site-www/issues/1265

### DIFF
--- a/src/_guides/libraries/library-tour.md
+++ b/src/_guides/libraries/library-tour.md
@@ -1177,9 +1177,11 @@ The `then().catchError()` pattern is the asynchronous version of
 #### Chaining multiple asynchronous methods
 
 The `then()` method returns a Future, providing a useful way to run
-multiple asynchronous functions in a certain order. If the callback
-registered with `then()` returns a Future, `then()` returns an
-equivalent Future. If the callback returns a value of any other type,
+multiple asynchronous functions in a certain order. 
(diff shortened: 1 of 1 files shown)


src/_guides/libraries/library-tour.md
```diff
@@ -1177,9 +1177,11 @@ The `then().catchError()` pattern is the asynchronous version of
 #### Chaining multiple asynchronous methods
 
 The `then()` method returns a Future, providing a useful way to run
-multiple asynchronous functions in a certain order. If the callback
-registered with `then()` returns a Future, `then()` returns an
-equivalent Future. If the callback returns a value of any other type,
+multiple asynchronous functions in a certain order. 
+If the callback registered with `then()` returns a Future, 
+`then()` returns a Future that will be completed
```
kwalrath, line 1182: The explanation looks good to me. I just would like to avoid "will be completed". This seems like the rare example of a valid use of _will_, and we use "Future completes" elsewhere, thus... "will complete."

```suggestion
`then()` returns a Future that will complete
```

kwalrath (approved):

PR #3512, opened 2021-08-22
Replace old custom dart font with material icons

### LINKED ISSUE #1099: Replace custom icons in the "dart" font by standard ones
There are a few font files (and an SVG) under https://github.com/dart-lang/site-www/tree/master/src/_assets/fonts. The custom "dart" font provides these 8 glyphs (the first four appear blank):

screen shot 2018-08-21 at 08 39 10

Replace uses of these custom glyphs with more conventional ones such as Material icons and Font Awesome (both of which we are already using).

cc @Sfshaza @kwalrath

parlough:
Fixes #1099


johnpryan (approved):
LGTM - assuming there's not supposed to be a noticeable difference? I deployed to https://jr-dartlang-site.web.app/ and things are working as I expect

kwalrath (approved):
LGTM. Great to see unnecessary files/complexity going away!

Issue #920, opened 2018-06-12
Consider adding copy-heading-link-to-clipboard functionality


chalin:
On site-www and site-webdev our page headings are dynamically enhanced so include a link to self:

> <img width="418" alt="screen shot 2018-06-12 at 17 43 22" src="https://user-images.githubusercontent.com/4140793/41318969-3a54da58-6e68-11e8-8d69-c05426cddf0f.png">

Clicking on the link visits the heading. This is a good way to anchor a page at a particular heading.

We've had a request to include a copy-heading-link-to-clipboard functionality. This would be particularly useful if we could also put rich text in the clipboard (i.e., an anchor element with both the URL and the heading text).

Accessing the clipboard in a safe way that works across browsers can be tricky though. It might be nice if such functionality were always visible so that mobile users could access it.

chalin:
Candidate packages:

- https://www.npmjs.com/package/clipboard

parlough:
I'm going to close this as I personally think the current functionality to scroll to that header is more beneficial, especially since copying to clipboard can already be done with the heading anchor by using the browser context menu and/or key bindings.

If anyone feels strongly about this though, please let me know :)

PR #3513, opened 2021-08-22
Use material font for anchor icon and add style to indicate it is hovered

parlough:
We already have the font, no need to use an extra icon file that can get out of date.


src/_assets/css/main.scss
```diff
@@ -752,21 +752,26 @@ main .content {
...
-        display: block;
-        height: $icon-size;
+        @include md-icon-before("link");
+        font-size: 24px;
+        color: $site-color-body;
         opacity: 0;
-        width: $icon-size;
+
+        &:hover {
+          color: lighten($site-color-body, 25%);
```
kwalrath, line 769: Nice touch! What do you think of making the hover color match the one we use for link text?

To me, graying the icon makes it look like it's inactive. Also, some headings with similar UIs copy the link URL, instead of going there, so the blue is perhaps a little clearer. Related issue: #920

parlough: I like this idea, thanks. Do you think I should just change the hover state to the same one as links or should the base state use the link color as well?

kwalrath: I hadn't thought about that... It might be less distracting to keep the base state color the dark gray. But I can see it both ways. One question is whether having the icon blue in base state would look better or worse when the icon is misbehaving. (Iirc, it's partially cut off sometimes.)

parlough: That makes sense, I think I'll stick with the darker grey for the base state.

As for when it's cut off, I wanted to fix that as well, either providing more space or preferably moving it to the other end of the header(like [web.dev](https://web.dev/minify-css/#loading-unminified-css) has theirs). It's not as straightforward as it seems jekyll-toc generates the anchor without allowing much configuration but I'll likely tackle it in the future.

parlough: Okay, I've switched the hover color to match link text when hovered. Let me know if that fits what you're thinking. Thanks!

A showcase that the anchor icon is a light blue when hovered

kwalrath (approved):
LGTM. Thanks, @parlough!

PR #3515, opened 2021-08-23
Migrate the web get started page

parlough:


kwalrath (commented):
We need at least one change, but otherwise LGTM! (I tested the custom code instructions, and they work.)

kwalrath (approved):

PR #3530, opened 2021-08-31
Update version mentions on SDK tool version page

parlough:
- Updates the listed SDK version since it's best for users to use the latest stable release and its what the CI uses.
- Updates the note about tools still being included to 2.13 as they still are present.


kwalrath (approved):
LGTM, thanks!

Issue #3518, opened 2021-08-25
[PAGE ISSUE]: 'Creating packages'


jakemac53:
### Page URL

https://dart.dev/guides/libraries/create-library-packages

### Page source

https://github.com/dart-lang/site-www/tree/master/src/_guides/libraries/create-library-packages.md

### Describe the problem

The organization section (https://dart.dev/guides/libraries/create-library-packages#organizing-a-library-package) should suggest exporting explicitly the symbols you want to expose (using `show`).

This helps you to not accidentally make things public that you didn't intend to - and it also gives users an overview of the entire public api of the package when looking at this file. This especially helps non-IDE users, they can no exactly which implementation import to go to in order to find a certain API and don't have to guess based on file names etc. 

### Expected fix

Update the shelf package to use this best practice, and update the page to reflect the new code.

PR #3531, opened 2021-08-31
Update shelf export example to use show

parlough:
Updates the example to match [the updated shelf exports](https://github.com/dart-lang/shelf/pull/200) and adds some related guidance.

Closes https://github.com/dart-lang/site-www/issues/3518

### DIFF
--- a/src/_guides/libraries/create-library-packages.md
+++ b/src/_guides/libraries/create-library-packages.md
@@ -83,16 +83,18 @@ Directly under lib, the main library file,
 `shelf.dart`, exports several files from lib/src:
 
 ```dart
-export 'src/cascade.dart';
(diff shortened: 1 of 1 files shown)


kwalrath (changes requested):
I think it's confusing to update the code without updating the guidance. We could say something like this:

--------

Directly under lib, the main library file,
`shelf.dart`, exports API from several files in lib/src.
To avoid exposing more API than intended —
and to give developers an overview of the entire public API of the package —
`shelf.dart` uses `show` to specify exactly which symbols to export:

kwalrath (approved):
I love the new text. ;) Let's see what @jakemac53 has to say.

jakemac53 (approved):
LGTM!

PR #3536, opened 2021-08-31
Enable Firebase Preview Deployments

mit-mit:


mit-mit:
@athomas can you help review this? This is based on:
https://github.com/FirebaseExtended/action-hosting-deploy/blob/main/README.md#deploy-to-a-new-preview-channel-for-every-pr

.github/workflows/build.yml
```diff
@@ -114,3 +114,37 @@ jobs:
       - run: ./tool/shared/deploy.sh --robots ok default
         env:
           FIREBASE_TOKEN: ${{ secrets.FIREBASE_TOKEN }}
+
+  deploy_preview:
+    needs: test
+    runs-on: ubuntu-latest
+    if: '${{ github.event.pull_request.head.repo.full_name == github.repository }}'
```
parlough, line 121: Does this limit the previews to PRs from branches created on this repository?

mit-mit: Yes, that is my understanding

athomas: To illustrate this a bit, the source for this PR is mit-mit:firebasepreview so the steps here weren't actually run. So I would suggest to re-create the PR from a branch in this repo so the effects can be seen before landing this (but read the other comment before).

.github/workflows/build.yml
```diff
@@ -114,3 +114,37 @@ jobs:
...
+        # TOOD: caching requires a lockfile for node deps, so does npm ci.
+        run: npm install
+      - name: Install dart dependencies
+        run: dart pub get
+      - run: ./tool/shared/write-ci-info.sh -v
+      - run: ./tool/build.sh
+      - uses: FirebaseExtended/action-hosting-deploy@0e16177d01e10c4560245404f959e75a9c633533
+        with:
+          repoToken: '${{ secrets.GITHUB_TOKEN }}'
+          firebaseServiceAccount: '${{ secrets.FIREBASE_SERVICE_ACCOUNT_DART_DEV }}'
```
athomas, line 149: This seems to require an exported service account key which will need an exemption from go/cute-sake.

.github/workflows/build.yml
```diff
@@ -114,3 +114,37 @@ jobs:
       - run: ./tool/shared/deploy.sh --robots ok default
         env:
           FIREBASE_TOKEN: ${{ secrets.FIREBASE_TOKEN }}
+
+  deploy_preview:
```
athomas, line 118: It would be possible to avoid the duplication by merging the if: expression here and then have the specific ifs further down in the deployment steps. The ifs make it harder to read the workflow, so I'm also OK with it as is.

.github/workflows/build.yml
mit-mit: This was completed earlier

.github/workflows/build.yml
mit-mit: I think that would mean adding if's to each and every step here, as we don't want to run the build, etc. unless we're doing a push to master or a PR from a non-fork?

.github/workflows/build.yml
athomas: > I think that would mean adding if's to each and every step here, as we don't want to run the build, etc. unless we're doing a push to master or a PR from a non-fork?

(copied comment from the other thread)

No:
"if master or PR" on `deploy:`
"if master" on the step needed post submit.
"if PR" on the step needed for PRs.

The remaining steps are the same.

athomas (approved):

PR #3539, opened 2021-09-01
Updates wording around 2.12 and restructures publishing portion of migration guide

### LINKED ISSUE #2955: Follow-up to null safety migration guide update
Follow-ups to #2929:

* Remove `-nullsafety` from images.
* Be consistent with https://dart.dev/tools/pub/publishing#publishing-previews.
  * Maybe link to/from it.
  * Maybe remove some content from /null-safety/migration-guide if it isn't essential.
* Be consistent about release vs. version, especially in the final paragraph. 
* "Switch to the latest beta release" instead of "Switch to the latest 2.12 beta release".

/cc @kevmoo @mit-mit

parlough:
Considering previous fixes that have already been made, this should cover most of the remaining concerns https://github.com/dart-lang/site-www/issues/2955. Anything else can be made into a separate, new issue as it arises.

Closes #2955

### DIFF
--- a/src/null-safety/migration-guide.md
+++ b/src/null-safety/migration-guide.md
@@ -433,7 +433,7 @@ as soon as you migrate:
 
 * [Set the package version to indicate a breaking change.](#package-version)
 * [Update the SDK constraints and package dependencies.](#check-your-pubspec)
-* [Publish the package][publish].
+* [Publish the package](/tools/pub/publishing).
   If you don't consider this version to be a stable release, 
   then [publish the package as a prerelease][].
 
--- a/src/null-safety/migration-guide.md
+++ b/src/null-safety/migration-guide.md
@@ -427,11 +427,14 @@ If so, revert your code changes before using the migration tool again.
 
 ## 5. Publish {#step5-publish}
 
-We encourage you to publish packages as prereleases
-as soon as you migrate:
+We encourage you to publish packages —
+possibly as prereleases —
+as soon as you migrate.
 
 * [Set the package version to indicate a breaking change.](#package-version)
-* [Publish the package as a prerelease if development is still in progress.][]
+* [Publish the package][publish].
+  If you don't consider the migrated package to be final quality, then
+  [publish it as a prerelease][publish-prerelease].
 
 [Publish the package as a prerelease if development is still in progress.]: /tools/pub/publishing#publishing-prereleases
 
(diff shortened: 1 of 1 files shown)


src/null-safety/migration-guide.md
```diff
@@ -430,34 +430,23 @@ If so, revert your code changes before using the migration tool again.
 We encourage you to publish packages as prereleases
 as soon as you migrate:
 
-* [Set the SDK constraints to the tested beta version.](#sdk-constraints)
 * [Set the package version to indicate a breaking change.](#package-version)
+* [Publish the package as a prerelease if development is still in progress.][]
 
-### SDK constraints
```
parlough, line 436: This section was confusing and I'm not sure it's necessary now that 2.12 is out of beta.

kwalrath: Perhaps we should still say to set the SDK constraint to the lowest tested SDK version, though?

src/null-safety/migration-guide.md
```diff
@@ -430,34 +430,23 @@ If so, revert your code changes before using the migration tool again.
 We encourage you to publish packages as prereleases
 as soon as you migrate:
 
-* [Set the SDK constraints to the tested beta version.](#sdk-constraints)
 * [Set the package version to indicate a breaking change.](#package-version)
+* [Publish the package as a prerelease if development is still in progress.][]
```
kwalrath, line 434: We should put the "if" part first (https://developers.google.com/style/clause-order). Also... maybe feature publishing first, and prereleases next?

```suggestion
* [Publish the package][publish].
  If you don't consider the migrated package to be final quality, then
  [publish it as a prerelease][publish-prerelease].
```

Of course you'd need to add/adjust the link definitions accordingly.

src/null-safety/migration-guide.md
```diff
@@ -430,34 +430,23 @@ If so, revert your code changes before using the migration tool again.
 We encourage you to publish packages as prereleases
 as soon as you migrate:
```
kwalrath, line 431:
```suggestion
We encourage you to publish packages —
possibly as prereleases —
as soon as you migrate.
```

kwalrath (commented):
Good thinking! I think we probably do want to mention the lower SDK constraint... and looking at the text, it appears that info is already buried in the **Package version** section. I think we might want to have a separate **Check your pubspec** section, since most of this info is already in the page but is spread out.

parlough:
Will hopefully get back to this tomorrow, thanks for the great suggestions :)

kwalrath (approved):
LGTM. Thank you, @parlough!

PR #3540, opened 2021-09-01
Add security headers and policies

### LINKED ISSUE #1521: Apply security headers
See: https://securityheaders.com/?q=https%3A%2F%2Fdart.dev

Credits @sortie :)

Even though this site is static content and low impact for XSS exploits, we should always apply security headers for defense in depth.

parlough:
Closes #1521

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -12,6 +12,7 @@
       {
         "source": "**",
         "headers": [
+          { "key": "Content-Security-Policy", "value":  "default-src https:; script-src 'self' 'unsafe-inline' https://www.googletagmanager.com https://www.google-analytics.com https://ssl.google-analytics.com https://fonts.gstatic.com https://dartpad.dev https://*.dartpad.dev https://youtube.com https://fonts.googleapis.com https://doubleclick.net https://*.doubleclick.net https://www.google.com https://www.gstatic.com https://cse.google.com https://apis.google.com/ https://storage.googleapis.com; object-src 'none'; base-uri 'none'"},
           { "key": "Referrer-Policy", "value":  "strict-origin-when-cross-origin"},
           { "key": "X-Frame-Options", "value":  "SAMEORIGIN"},
           { "key": "X-Content-Type-Options", "value":  "nosniff"}
--- a/firebase.json
+++ b/firebase.json
@@ -7,11 +7,16 @@
         "source": "**/*.@(txt|json)",
         "headers": [
           { "key": "Access-Control-Allow-Origin", "value": "*" },
+        ]
+    },
+      {
+        "source": "**",
+        "headers": [
           { "key": "Referrer-Policy", "value":  "strict-origin-when-cross-origin"},
           { "key": "X-Frame-Options", "value":  "SAMEORIGIN"},
           { "key": "X-Content-Type-Options", "value":  "nosniff"}
         ]
-    }],
+      }],
     "redirects": [
       { "source": "/\\)", "destination": "/", "type": 301 },
       { "source": "/+", "destination": "https://plus.sandbox.google.com/+dartlang", "type": 301 },
--- a/firebase.json
+++ b/firebase.json
@@ -12,7 +12,7 @@
       {
         "source": "**",
         "headers": [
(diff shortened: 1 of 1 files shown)


firebase.json
```diff
@@ -5,7 +5,12 @@
     "trailingSlash": false,
     "headers": [{
         "source": "**/*.@(txt|json)",
-        "headers": [ { "key": "Access-Control-Allow-Origin", "value": "*" } ]
+        "headers": [
+          { "key": "Access-Control-Allow-Origin", "value": "*" },
+          { "key": "Referrer-Policy", "value":  "strict-origin-when-cross-origin"},
+          { "key": "X-Frame-Options", "value":  "SAMEORIGIN"},
+          { "key": "X-Content-Type-Options", "value":  "nosniff"}
+        ]
```
jonasfj, line 13: Why only apply this to `txt` and `json` files?

That is what this does, right?

jonasfj: I see no reason not to enforce:
```js
          { "key": "Referrer-Policy", "value":  "strict-origin-when-cross-origin"},
          { "key": "X-Frame-Options", "value":  "SAMEORIGIN"},
          { "key": "X-Content-Type-Options", "value":  "nosniff"}
```

on all requests..

If viable, we might also want to consider CSP policies along the lines of:
```
Content-Security-Policy: default-src https:; script-src 'nonce-{random}' 'unsafe-inline'; object-src 'none'; base-uri 'none'
```

Though, we might need to use hash instead of nonce... or allow list specific domains... since this is a static site.

jonasfj (commented):
This seems reasonable..

firebase.json
parlough: You're correct, I didn't mean to limit this to just `txt` and `json` files, I've expanded it to a wildcard.

parlough: As for the specific domains, I tried going the route of listing specific domains. I haven't set a CSP up before, so if you want to take a closer look, that'd be great! Thanks.

firebase.json
```diff
@@ -5,8 +5,20 @@
...
-        "headers": [ { "key": "Access-Control-Allow-Origin", "value": "*" } ]
-    }],
+        "headers": [
+          { "key": "Access-Control-Allow-Origin", "value": "*" }
+        ]
+    },
+      {
+        "source": "**",
+        "headers": [
+          { "key": "Content-Security-Policy", "value":  "default-src https:; script-src 'self' 'unsafe-inline' 'unsafe-eval' https://www.googletagmanager.com https://www.google-analytics.com https://ssl.google-analytics.com https://gstatic.com https://*.gstatic.com https://dartpad.dev https://*.dartpad.dev https://youtube.com https://fonts.googleapis.com https://doubleclick.net https://*.doubleclick.net https://google.com https://*.google.com https://storage.googleapis.com; object-src 'none'; base-uri 'none'; style-src https: 'unsafe-inline'"},
```
jonasfj, line 15: Why do we need: `'unsafe-inline' 'unsafe-eval'`

I'm guess there is an actual reason :D

Perhaps we file a follow-up to fix this.

jonasfj (approved):
This looks fine.

@kwalrath you'll want to check if this creates issues when rolling it out, and then revert if it does.

firebase.json
parlough: The [search page](https://dart.dev/search) and the Google Tag Manager script on every page relied on `unsafe-inline`. Adding the hashes, I was able to remove `unsafe-inline`, and as far as I can there are no new problems popping up from that change.

`unsafe-eval` seems trickier as it is currently used by Google CSP for the search page as well as on the [Dart SDK Archive](https://dart.dev/tools/sdk/archive) page.

kwalrath (commented):
LGTM, but let's wait just a bit before merging this.

firebase.json
jonasfj: @parlough another option, instead of adding hashes is to embed the Google Tag Manager script from a `.js` file.

See:
https://github.com/dart-lang/pub-dev/blob/master/static/js/gtm.js

That's how we did on pub.dev

jonasfj:
It's worth shipping even with `unsafe-eval`, because it points out something that can be fixed later :D

kwalrath (approved):

PR #3541, opened 2021-09-01
Add >>> operator (new in 2.14)

### LINKED ISSUE #3455: Document the triple-shift (>>>) operator

2.14 will support the `>>>` operator. I need to update the language tour to reflect that. Resources:

* [Feature specification](https://github.com/dart-lang/language/blob/master/accepted/future-releases/triple-shift-operator/feature-specification.md)
* https://github.com/dart-lang/language/issues/120
* https://github.com/dart-lang/language/issues/478
* https://github.com/dart-lang/sdk/issues/30890

The JS version is different because JS numbers are different:
* [The feature spec says](https://github.com/dart-lang/language/blob/master/accepted/future-releases/triple-shift-operator/feature-specification.md#background), "The JavaScript implementation of int's >>> operator must be decided and implemented by the JavaScript platforms. It likely works like >> except that it doesn't "sign-extend" the most significant bit."
* [Lasse's comment says](https://github.com/dart-lang/sdk/issues/30890#issuecomment-487969421), "The int.>>> operator should probably do non-sign-extending shift of Int32 values, then convert to Uint32. That is where the Dart x >> y is implemented in JS as (x >> y) >>> 0, then x >>> y in Dart can be implemented as just x >>> y in JS."

Language tour, where `>>` is listed/discussed.

kwalrath:
Closes #3455

Staged: https://kw-staging-dartlang-2.web.app/guides/language/language-tour

### DIFF
--- a/examples/misc/pubspec.yaml
+++ b/examples/misc/pubspec.yaml
@@ -2,7 +2,7 @@ name: examples
 description: dart.dev example code.
 
 environment:
-  sdk: ">=2.14.0 <3.0.0"
+  sdk: ">=2.14.0-0 <3.0.0"
 
 dev_dependencies:
   args: ^2.2.0
--- a/examples/misc/pubspec.yaml
+++ b/examples/misc/pubspec.yaml
@@ -2,7 +2,7 @@ name: examples
 description: dart.dev example code.
 
 environment:
-  sdk: ">=2.13.0 <3.0.0"
+  sdk: ">=2.14.0-0 <3.0.0"
 
 dev_dependencies:
   args: ^2.2.0
--- a/examples/misc/pubspec.yaml
+++ b/examples/misc/pubspec.yaml
@@ -2,7 +2,7 @@ name: examples
 description: dart.dev example code.
 
 environment:
-  sdk: ">=2.14.0-0 <3.0.0"
+  sdk: ">=2.14.0 <3.0.0"
 
 dev_dependencies:
   args: ^2.2.0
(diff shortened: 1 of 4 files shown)


parlough (approved):
This all looks great and makes sense to me :)

I'm not sure all users will understand exactly what `>>>` does from just the examples, but the other bitwise operators are lacking similar details as well. Perhaps that's a location for improvement in the future or a separate tutorial/article.

kwalrath:
> I'm not sure all users will understand exactly what `>>>` does from just the examples, but the other bitwise operators are lacking similar details as well. Perhaps that's a location for improvement in the future or a separate tutorial/article.

Yeah, I went with the same minimalism as before. I tried to come up with a better example, but we figured that `>>>` (and possibly all bitwise operators) is something most people don't need, and if they do need it, they already know how to use it. So I concentrated on having a working test and putting in the words people might search for.

kwalrath:
I think the build break for stable tests is due to requiring 2.14 (for `>>>`). Also, once a stable 2.14 is available, I should change examples/misc/pubspec.yaml to have 2.14.0 instead of 2.14.0-0.

kwalrath:
@lrhn or @mit-mit is this good to go once the feature's in a stable release?

lrhn:
I'd say so, yes. LGTM.

parlough:
@googlebot I consent.

parlough:
I rebased to fix the `stable` build but it seems updating the constraint to `2.14.0` didn't fit the constraint of the beta build.

We can either keep `2.14.0-0` or temporarily disable the `beta` analysis job.

parlough:
When I was taking another look before I merged this in,  but I noticed we have a section on compound assignment operators, and it seems `>>>=` is valid. Do we want to add it to that section as well?

mit-mit:
I asked @lrhn, and it sounds like we should indeed be listing `>>>=` in the table here:
https://kw-staging-dartlang-2.web.app/guides/language/language-tour#assignment-operators

kwalrath:
I'll add `>>>=` a little later.

PR #3549, opened 2021-09-07
Show M1 SDKs in stable

mit-mit:


parlough:
@mit-mit Next time, can you target changes to a site-www branch next time now that automatic previews are enabled. Or perhaps we can expand to allow previews from collaborator branches as well?

mit-mit:
This change doesn't show anything in preview (it needs the stable SDK to actually be deployed), so I deliberately didn't target site-www...

mit-mit:
But it would actually be useful to have it in preview come to thing of it, so that we can validate before we merge it... Will create a new one.

mit-mit:
replaced by https://github.com/dart-lang/site-www/pull/3550

PR #3559, opened 2021-09-08
Update evolution and "what's new" pages for 2.14

kwalrath:
Also removed a reference to the 2.13 SDK

### DIFF
--- a/src/_guides/language/evolution.md
+++ b/src/_guides/language/evolution.md
@@ -204,16 +204,16 @@ For more information, see the
 [bitwise and shift operator]: /guides/language/language-tour#bitwise-and-shift-operators
 
 Dart 2.14 also removed some restrictions on type arguments.
-You can now use a generic function type as a type argument,
-and you can pass type arguments to annotations.
+You can now pass type arguments to annotations,
+and you can use a generic function type as a type argument.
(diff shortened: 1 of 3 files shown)


src/_guides/language/evolution.md
```diff
@@ -192,6 +192,30 @@ anywhere the original type could be used.
...
+You can now use a generic function type as a type argument,
+and you can pass type arguments to annotations.
+All of the following code was invalid before 2.14,
+but is now allowed:
+
+```dart
+late List<T Function<T>(T)> idFunctions;
+var callback = [<T>(T value) => value];
+late S Function<S extends T Function<T>(T)>(S) f;
+@TypeHelper<int>(42, "The meaning")
```
parlough, line 216: This may be slightly confusing for readers as the other snippets are valid, but doesn't the annotation need some sort of declaration or identifier.

Perhaps even just putting it above one of the others would make it valid.

parlough (approved):
Looks good to me, one minor question which works either way.

mit-mit (approved):

src/_guides/language/evolution.md
kwalrath: I switched the order of the sentence and the examples, so I think it works now.

PR #3562, opened 2021-09-09
Add note about unawaited futures

### LINKED ISSUE #1250: Example shows Future without await
There's a sample on the end of the Handling Futures section, showing how to make `main` `async`.

That sample starts by calling `checkVersion()`, which returns a `Future`, as declared before, but there's no `await` there.

So for a beginner, the expected behavior isn't obvious.... will it `await`, because there's an `await` inside `checkVersion()`? Or will it just ignore it all, executing a function that returns a `Future`, and discards that `Future`?

I think there should be another paragraph explaining what would happen on that particular case.

kwalrath:
Fixes #1250

Staged: https://dart-dev--pr3562-kw-unawait-yitvub00.web.app/guides/language/language-tour#handling-futures

I opted not to explain the behavior, but to warn about it and to point people (twice!) to the codelab. I wonder: should the note be a warning or tip?

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -4073,7 +4073,8 @@ looks similar to synchronous code.
 When you need the result of a completed Future,
 you have two options:
 
-* Use `async` and `await`.
+* Use `async` and `await`, as described here and in the
+  [asynchronous programming codelab](/codelabs/async-await).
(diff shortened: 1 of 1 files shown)


parlough (approved):
I think just a note is fine, but a warning works as well. I just am not a big fan of the current warning color.

PR #3564, opened 2021-09-10
Flesh out the "Overriding members" part of the tour

### LINKED ISSUE #1038: A few words on override constraints might be helpful in the language tour.
I noted that the language tour section on 'Overriding members' does not mention any of the requirements that an overriding instance member declaration must satisfy. Given that every Dart developer who writes an overriding member declaration must satisfy those requirements, and given that Dart 1 was a lot more permissive in this area, I think it would be helpful to mention this topic already in the language tour.

The section is here: https://www.dartlang.org/guides/language/language-tour#overriding-members. I think the following snippet could be taken as a rough indication of the kind of information I'm thinking about:

```
Subclasses can override instance methods, getters, and setters.
You can use the `@override` annotation to indicate that you are
intentionally overriding a member:

<?code-excerpt "misc/lib/language_tour/metadata/television.dart (override)" replace="/@override/[!$&!]/g"?>
{% prettify dart %}
class Television {
  // ...
  set contrast(int value) {...}
}
class SmartTelevision extends Television {
  [!@override!]
  set contrast(num value) {...}
  // ···
}
{% endprettify %}

An overriding method declaration has to match the declaration(s)
that it is overrides in several ways. For instance, it must have a return
type which is a subtype of the overridden ones (it can be the same type),
and argument types which are supertypes (typically it will be the same
type); if the overridden method accepts _k_ positional parameters then
so must the overriding method; finally, a generic method cannot override
a non-generic one, nor vice versa.

In some situations you may wish to narrow the type of a method
parameter or instance variable. This violates the normal rules, and it
is similar to a downcast in that it incurs a possible type error at run time,
but it may still work well in a context where the code has been written
in such a way that the type error will not occur. You can use the 
[`covariant` keyword](/guides/language/sound-problems#the-covariant-keyword)
in a parameter declaration in order to use this feature.

For further details, please consult the 
[Dart Language Specification](/guides/language/spec).
```

Given that covariant parameters are a somewhat esoteric feature we might
want to remove the reference to that feature from the language tour entirely,
but in the above snippet I just modified the explanation a bit such that it's
explicitly marked as an exception, and as an unsafe feature.

kwalrath:
Fixes #1038

Staged: https://dart-dev--pr3564-kw-override-6ye3zc3b.web.app/guides/language/language-tour#overriding-members

### DIFF
--- a/examples/misc/lib/language_tour/metadata/television.dart
+++ b/examples/misc/lib/language_tour/metadata/television.dart
@@ -1,5 +1,7 @@
 // #docregion deprecated
+// #docregion override
 class Television {
+  // #enddocregion override
   /// Use [turnOn] to turn the power on instead.
   @Deprecated('Use turnOn instead')
   void activate() {
@@ -8,13 +10,18 @@ class Television {
 
   /// Turns the TV's power on.
   void turnOn() {/*...*/}
+  // #enddocregion deprecated
+
+  // #docregion override
+  set contrast(int value) {/*...*/}
+  // #docregion deprecated
 }
 // #enddocregion deprecated
 
-// #docregion override
 class SmartTelevision extends Television {
   @override
-  void turnOn() {/*...*/}
+  set contrast(num value) {/*...*/}
   // #enddocregion override
   // #docregion override
 }
+// #enddocregion override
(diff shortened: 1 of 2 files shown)


kwalrath:
Only 2 years later... Does this text work for you, @eernstg?

parlough (approved):
The example changes and the new additions make sense to me and look good.

kwalrath:
I'm going to commit this, since we can always iterate if need be.

eernstg (commented):
Looks very good, thanks!

(I somehow missed the request for input from me, sorry about that.)

PR #3565, opened 2021-09-10
Remove Dart 1 books from resources page

### LINKED ISSUE #3543: [PAGE ISSUE]: 'Books about Dart'
### Page URL

https://dart.dev/resources/books.html

### Page source

https://github.com/dart-lang/site-www/tree/master/src/resources/books.md

### Describe the problem

I think we should consider removing the Dart 1 books from this page. 

There are two Dart 2 books listed; the Wenderlich one is pretty good (I haven't reviewed the other). 

I can't think of a scenario when I'd recommend the other books: moreover, I can only imagine that most of them are sufficiently outdated that they will have a negative impact on folks' ability to learn modern idiomatic Dart. 

Curious on others' thoughts, @mit-mit and @kwalrath in particular? 

### Expected fix

_No response_

### Additional context

_No response_

timsneath:
Closes #3543


parlough:
The general idea looks good to me, I'm curious if a warning alert box would make sense, as readers of the website would be used to that structure identifying things to look out for.

It looks like this:
Screenshot from 2021-09-12 15-05-43


Implemented like:

```markdown
{{site.alert.warning}}
  Operator precedence is an approximation of the behavior of a Dart parser.
{{site.alert.end}}
```

src/resources/books.md
```diff
@@ -9,6 +9,15 @@ also cover Dart.
 If you find another Dart book that might be helpful, please
 [let us know.](https://github.com/dart-lang/site-www/issues)
 
+{{site.alert.warning}}
+  You may find other books published on the Dart programming 
+  language. We recommend that you check the date of publication, since 
+  Dart has evolved a great deal, and older books lack coverage of many 
+  important topics that are relevant to the modern Dart 2 language, 
+  including strong typing, null safety, FFI, the `dart` command line
+  utility, and our new developer tools.
```
kwalrath, line 18: Would it be more helpful to list the releases for important features, rather than publication dates? Maybe something like this:

```suggestion
  If you find a Dart book that isn't listed here,
  check which Dart SDK release it reflects.
  Older books lack coverage of Dart 2 topics such as
  strong typing (2.0), null safety (2.12),
  FFI, the `dart` command-line utility,
  and new developer tools. 
```

If not that, then maybe we should list the dates where each feature was introduced (e.g. 2.0 is Feb 2018, 2.12 is Mar 2021)... but books published after those dates might still have old information.

kwalrath (commented):
Thanks for the PR! I wonder if we shouldn't concentrate on releases, rather than a nebulous publication date.

src/resources/books.md
timsneath: For what it's worth, my feeling was that it essentially doesn't matter when those features were added. I picked them as examples of why any Dart 1 books are wildly out-of-date at this point. I like your other changes though!

kwalrath: Maybe we should just say "make sure it was published recently"? Then keep my changes, but remove the release numbers.

timsneath: Oh, sure... It's your call, either way!

kwalrath (approved):

parlough (approved):

PR #3570, opened 2021-09-13
Try using Google Sans in banner

parlough:
I noticed on my computer, particularly on Linux with a 1440p display, the banner text seemed a bit blurry. Improving the font-weight to 400 helped but then "2.14" looked strange. Switching to Google Sans fixed that issue and improved it overall even more.

It's not as large of a problem on my MacBook Pro. As a side note, flutter.dev also uses Google Sans for its banner.

I'm curious to see what you think. Open the [preview link](https://dart-dev--pr3570-misc-use-google-sans-rzpejl53.web.app/) below and the [live site](https://dart.dev/) next to each other on your computer. Thanks!

### DIFF
--- a/src/_assets/css/_dash.scss
+++ b/src/_assets/css/_dash.scss
@@ -357,6 +357,7 @@ $dash-dartpad-border: #293542;
 
     a {
       color: $site-color-card-link;
+      font-family: $site-font-family-gsans;
 
       &:hover, &:focus, &:active {
         color: darken($site-color-card-link, 20%);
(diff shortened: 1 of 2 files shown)


mit-mit:
On my screen, the new text looks overly saturated. Perhaps change just the font, not the weight?

parlough:
Yeah I took a look on my MacBook again, and it does seem to be a bit saturated. It looks fine on the Flutter website, but that's likely due to the larger font size, which I'm not sure would fit here.

Unfortunately it seems Google Sans is not available at a `300` font weight so I'll consider other ways to improve the legibility.

PR #3571, opened 2021-09-13
Clarify import path recommendations

### LINKED ISSUE #2606: Effective Dart - which import style is preferred for package's own imports?

jonasfj:
Bring import documentation in line with Effective Dart:
https://dart.dev/guides/language/effective-dart/usage#dont-allow-an-import-path-to-reach-into-or-out-of-lib

### DIFF
--- a/src/tools/pub/package-layout.md
+++ b/src/tools/pub/package-layout.md
@@ -313,10 +313,17 @@ tests) but you should never import from another package's `lib/src` directory.
 Those files are not part of the package's public API, and they might change in
 ways that could break your code.
 
-When you use libraries from within your own package, even code in `src`, you
-can (and should) still use `package:` to import them. For example:
+When you use libraries from within your own package, you should use:
+ * `package:` when [reaching inside or outside `lib/`](https://dart.dev/guides/language/effective-dart/usage#dont-allow-an-import-path-to-reach-into-or-out-of-lib) (lint: [_avoid_relative_lib_imports_](https://dart.dev/tools/linter-rules#avoid_relative_lib_imports))
+ * [Prefer relative imports](https://dart.dev/guides/language/effective-dart/usage#prefer-relative-import-paths) otherwise.
+
+For example:
 
 {% prettify dart tag=pre+code %}
+// When importing from lib/beans.dart
+import 'src/beans.dart';
+
+// When importing from test/beans_test.dart
 import 'package:enchilada/src/beans.dart';
 {% endprettify %}
 
--- a/src/tools/pub/package-layout.md
+++ b/src/tools/pub/package-layout.md
@@ -313,9 +313,15 @@ tests) but you should never import from another package's `lib/src` directory.
 Those files are not part of the package's public API, and they might change in
 ways that could break your code.
 
-When you use libraries from within your own package, you should use:
- * `package:` when [reaching inside or outside `lib/`](https://dart.dev/guides/language/effective-dart/usage#dont-allow-an-import-path-to-reach-into-or-out-of-lib) (lint: [_avoid_relative_lib_imports_](https://dart.dev/tools/linter-rules#avoid_relative_lib_imports))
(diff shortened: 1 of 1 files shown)


src/tools/pub/package-layout.md
```diff
@@ -313,10 +313,17 @@ tests) but you should never import from another package's `lib/src` directory.
 Those files are not part of the package's public API, and they might change in
 ways that could break your code.
 
-When you use libraries from within your own package, even code in `src`, you
-can (and should) still use `package:` to import them. For example:
+When you use libraries from within your own package, you should use:
+ * `package:` when [reaching inside or outside `lib/`](https://dart.dev/guides/language/effective-dart/usage#dont-allow-an-import-path-to-reach-into-or-out-of-lib) (lint: [_avoid_relative_lib_imports_](https://dart.dev/tools/linter-rules#avoid_relative_lib_imports))
+ * [Prefer relative imports](https://dart.dev/guides/language/effective-dart/usage#prefer-relative-import-paths) otherwise.
```
parlough, line 318: Switches to use relative links.

```suggestion
When you use libraries from within your own package, you should use:

 * `package:` when [reaching inside or outside `lib/`][] (lint: [_avoid_relative_lib_imports_][]).
 * [Prefer relative imports][] otherwise.
 
 [reaching inside or outside `lib/`]: /guides/language/effective-dart/usage#dont-allow-an-import-path-to-reach-into-or-out-of-lib
 [_avoid_relative_lib_imports_]: /tools/linter-rules#avoid_relative_lib_imports
 [Prefer relative imports]: /guides/language/effective-dart/usage#prefer-relative-import-paths
```

lrhn (approved):

src/tools/pub/package-layout.md
kwalrath: One tweak: The [intro to the list should be a sentence](https://developers.google.com/style/lists#introductory-sentences-for-lists), and the [list items should be parallel](https://developers.google.com/style/lists#parallel-syntax). Something like this:

How you import libraries from within your own package depends on the locations of the libraries:

* When [reaching...] (lint: ...), use `package:`.
* Otherwise, [prefer relative imports].

kwalrath (approved):
LGTM if our changes LGT @jonasfj

parlough (approved):

kwalrath:
I'm going to merge this. We can iterate if necessary.

PR #3578, opened 2021-09-15
Clarify "extends" in generics

### LINKED ISSUE #2659: Difference between <T extends SomeBaseClass> to just <SomeBaseClass>
Since Dart generics are very limited compared to other languages, I wonder what's the difference between `<T extends SomeBaseClass>` to just `<SomeBaseClass>`. This is the kind of use case the [language tour shows][1], but I don't see the point of using it.

Am I missing something? I feel the docs should at least point out why `<T extends SomeBaseClass>` would be desirable over the simpler `<SomeBaseClass>`.

  [1]: https://dart.dev/guides/language/language-tour#restricting-the-parameterized-type

kwalrath:
Inspired by #2659, but language changes made that no longer an issue.

### DIFF
--- a/examples/misc/lib/language_tour/generics/misc.dart
+++ b/examples/misc/lib/language_tour/generics/misc.dart
@@ -12,3 +12,9 @@ void miscDeclAnalyzedButNotTested() {
     // #enddocregion collection-literals
   }
 }
+
+// #docregion non-nullable
+class Foo<T extends Object> {
+  // Foo works on objects with the non-nullable type T.
+}
+// #enddocregion non-nullable
--- a/examples/misc/lib/language_tour/generics/misc.dart
+++ b/examples/misc/lib/language_tour/generics/misc.dart
@@ -15,6 +15,6 @@ void miscDeclAnalyzedButNotTested() {
 
 // #docregion non-nullable
 class Foo<T extends Object> {
-  // Foo works on objects with the non-nullable type T.
+  // Any type provided to Foo for T must be non-nullable.
 }
 // #enddocregion non-nullable
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -182,22 +182,22 @@ The following table lists the words that the Dart language treats specially.
 {% assign bii = '&nbsp;<sup title="built-in-identifier" alt="built-in-identifier">2</sup>' %}
 {% assign lrw = '&nbsp;<sup title="limited reserved word" alt="limited reserved word">3</sup>' %}
 <div class="table-wrapper" markdown="1">
-| [abstract][]{{bii}}   | [else][]              | [import][]{{bii}}     | [show][]{{ckw}}   |
-| [as][]{{bii}}         | [enum][]              | [in][]                | [static][]{{bii}} |
-| [assert][]            | [export][]{{bii}}     | [interface][]{{bii}}  | [super][]         |
-| [async][]{{ckw}}      | [extends][]           | [is][]                | [switch][]        |
-| [await][]{{lrw}}      | [extension][]{{bii}}  | [late][]{{bii}}       | [sync][]{{ckw}}   |
-| [break][]             | [external][]{{bii}}   | [library][]{{bii}}    | [this][]          |
-| [case][]              | [factory][]{{bii}}    | [mixin][]{{bii}}      | [throw][]         |
-| [catch][]             | [false][]             | [new][]               | [true][]          |
-| [class][]             | [final][]             | [null][]              | [try][]           |
-| [const][]             | [finally][]           | [on][]{{ckw}}         | [typedef][]{{bii}}|
-| [continue][]          | [for][]               | [operator][]{{bii}}   | [var][]           |
-| [covariant][]{{bii}}  | [Function][]{{bii}}   | [part][]{{bii}}       | [void][]          |
-| [default][]           | [get][]{{bii}}        | [required][]{{bii}}   | [while][]         |
-| [deferred][]{{bii}}   | [hide][]{{ckw}}       | [rethrow][]           | [with][]          |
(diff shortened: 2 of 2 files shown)


kwalrath:
@parlough, I'm not sure about my keyword table change (visible at /guides/language/language-tour#keywords). I wanted to point to both major discussions of `extends`.

@leafpetersen, the site should be auto-staged soon (a comment will appear in this PR with the URL). The part I'd like you to review will be at /guides/language/language-tour#restricting-the-parameterized-type.

parlough:
I agree the (a, b) is a little jarring and I'm not sure readers would know what they mean. Perhaps it's just better to point to one section and have that section point to the other early on in its text. I'll give some more thought to possible formatting options though.

examples/misc/lib/language_tour/generics/misc.dart
```diff
@@ -12,3 +12,9 @@ void miscDeclAnalyzedButNotTested() {
     // #enddocregion collection-literals
   }
 }
+
+// #docregion non-nullable
+class Foo<T extends Object> {
+  // Foo works on objects with the non-nullable type T.
```
leafpetersen, line 18: This is a little off to me.  It's true, but the sentence is also true for `T extends Object?`.  I think you want to say something closer to "// Any type provided to Foo for T must be non-nullable"

src/_guides/language/language-tour.md
```diff
@@ -3862,9 +3863,30 @@ print(names is List<String>); // true
 ### Restricting the parameterized type
 
 When implementing a generic type,
-you might want to limit the types of its parameters.
+you might want to limit the types of its parameters
```
leafpetersen, line 3866: This kind of mixes categories.  I don't want to be too technical, but it's probably worth trying for as much accuracy as possible here.  I'd say something like "limit the types that can be provided as arguments to be a subtype of a particular type".  The point being that the constraint is on arguments (not parameters).

src/_guides/language/language-tour.md
```diff
@@ -3862,9 +3863,30 @@ print(names is List<String>); // true
...
+(instead of the default, [`Object?`][top-and-bottom]).
+
+<?code-excerpt "misc/lib/language_tour/generics/misc.dart (non-nullable)"?>
+```dart
+class Foo<T extends Object> {
+  // Foo works on objects with the non-nullable type T.
+}
+```
+
+How does `Foo<T extends Object>` differ from `Foo<Object>`?
```
leafpetersen, line 3881: This sentence doesn't really make sense.  It's mixing categories.  Do you mean `Foo<T>` for the latter?  I actually have no idea what this sentence is trying to say, to be honest.

src/_guides/language/language-tour.md
```diff
@@ -3862,9 +3863,30 @@ print(names is List<String>); // true
...
+If you use `<Object>`, then objects used in
+`Foo` members can have _any_ subtype of `Object`.
+Adding `T extends` lets you restrict the type
+to subtypes of `T`.
+
+You can use `extends` with other types besides `Object`.
```
leafpetersen, line 3887: Maybe worth pointing out *why* you would do this?  The main use case is that it allows members of `SomeBaseClass` to be called on things of type `T`.

src/_guides/language/language-tour.md
```diff
@@ -3875,7 +3897,7 @@ class Foo<T [!extends SomeBaseClass!]> {
 class Extender extends SomeBaseClass {...}
 {% endprettify %}
 
-It's OK to use `SomeBaseClass` or any of its subclasses as generic argument:
+It's OK to use `SomeBaseClass` or any of its subclasses as the generic argument:
```
leafpetersen, line 3900: Not sure if it's worth going into, but it's not just subclasses - it's subtypes in general (i.e. also things that implement SomeBaseClass).

src/_guides/language/language-tour.md
kwalrath: This is an attempt to fix #2659.

kwalrath:
> I agree the (a, b) is a little jarring and I'm not sure readers would know what they mean. Perhaps it's just better to point to one section and have that section point to the other early on in its text. I'll give some more thought to possible formatting options though.

I think you're right that linking to the first occurrence — which then links to any other occurrences — is the way to go.

src/_guides/language/language-tour.md
leafpetersen: I still don't know what this is trying to say.  Let me try to expand out a bit.  The syntax `Foo<T extends Object>` only makes sense as part of a class declaration: `class Foo<T extends Object>`.  That is, I can use it as a type: `int f(Foo<T extends Object> parameter) => 3)` is not valid syntax.  The syntax `Foo<Object>` on the other hand, is *technically* valid in both a declaration and a use, but almost certainly is not what you want in a declaration.  Writing `class Foo<Object> {}` is exactly the same as writing `class Foo<T> {}`.  That is, `Object` here does *not* refer to the core `Object` type, but is just a bound type variable, just like `T`.  The syntax `int f(Foo<Object> parameter) => 3` on the other hand means "f is a function which takes a parameter of type `Foo<Object>`.  That is, `Object` is the type *argument* passed to `Foo`.  

So my confusion is that this sentence, I think, is mixing declarations and uses.  The discussion in the #2659, after the initial comment (which seemed to share the same confusion), became more about the difference between subtype and generic polymorphism.  If the goal here is to elaborate on that, I think a different structure is warranted.

parlough (approved):
These changes all make sense to me and read as expected.

src/_guides/language/language-tour.md
kwalrath: OK, I've removed that note.

kwalrath:
@leafpetersen I've removed the note since it's moot with the language changes, and #2659 is now closed.

PR #3579, opened 2021-09-15
Mention that top-level and class vars are lazily inited

### LINKED ISSUE #2493: Add information about lazy initialization of global and static variables
It would be nice to add some words to the language tour about the fact that global and static variables are initialized lazy,

kwalrath:
Fixes #2493

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -377,7 +377,10 @@ if (weLikeToCount) {
 print(lineCount);
 ```
 
-<!-- TODO: Point to coverage of ! and late -->
+Top-level and class variables are lazily initialized;
+the initialization code runs
+the first time the variable is used.
+
 
 ### Late variables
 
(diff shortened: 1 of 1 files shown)


src/_guides/language/language-tour.md
```diff
@@ -377,7 +377,10 @@ if (weLikeToCount) {
 print(lineCount);
 ```
 
-<!-- TODO: Point to coverage of ! and late -->
+Top-level and class variables are lazily initialized;
+the initialization code runs
+the first time the variable is used.
```
parlough, line 382: I'm not 100% sure or I may be misunderstanding, but I believe only static class variables are lazily initialized. Unless they are marked `late`. 

For example(https://dartpad.dev/?null_safety=true&id=88b7f2681551b90bfedfd15bcfc403c9):

```dart
class Test {
  final test = print(3);
}

void main() {
  Test();
}
```

prints 3.

kwalrath: I've been using "class variables" to mean "static variables" (as opposed to instance variables), but maybe that language is olde fashioned.

parlough: Ah yeah, that makes sense, I was mixing up my terminology. I think this should be fine then as it's actually the term used in the spec.

parlough (approved):
Looks good to me, thanks :D

PR #3580, opened 2021-09-17
LangTour: computing finals, factory constructors, and late final

### LINKED ISSUE #267: Mention computing finals as a use case for `factory`
https://github.com/dart-lang/site-www/blob/68367eb5a3d8cd821db46d432a8563f250e6972f/src/_guides/language/language-tour.md#factory-constructors

We often get confusion from devs who want to assign to a final in the constructor like they can in Java. Usually we point to factory constructors as an alternative. Our documentation introducing factory constructor should also call this out as a possible use case.

kwalrath:
Fixes #267
(maybe)

But we should perhaps update Effective Dart if we make this change.

/cc @natebosch @munificent 

Staged:
https://dart-dev--pr3580-kw-finals-rwmq5v76.web.app/guides/language/language-tour#instance-variables
https://dart-dev--pr3580-kw-finals-rwmq5v76.web.app/guides/language/language-tour#factory-constructors

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -2846,8 +2846,9 @@ class ProfileMark {
 If you need to assign the value of a `final` instance variable
 after the constructor body starts, you can use one of the following:
 
-* Use `late final`, but [_be careful_][late-final-ivar].
 * Use a [factory constructor](#factory-constructors).
+* Use `late final`, but [_be careful_][late-final-ivar];
+  a `late final` without an initializer adds a setter to the API.
 
 
 ### Constructors
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -2844,7 +2844,7 @@ class ProfileMark {
 }
 ```
 
-If you want to assign the value of a `final` instance variable
+If you need to assign the value of a `final` instance variable
 after the constructor body starts, you can use one of the following:
 
 * Use `late final`, but [_be careful_][late-final-ivar].
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -2847,7 +2847,7 @@ If you need to assign the value of a `final` instance variable
 after the constructor body starts, you can use one of the following:
 
 * Use a [factory constructor](#factory-constructors).
-* Use `late final`, but [_be careful_][late-final-ivar];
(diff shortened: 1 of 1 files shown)


src/_guides/language/language-tour.md
```diff
@@ -2845,10 +2845,10 @@ class ProfileMark {
 ```
 
 If you want to assign the value of a `final` instance variable
```
parlough, line 2847: What do you think about changing `want` to `need`, to make it more explicit these options aren't necessarily the best choice?

```suggestion
If you need to assign the value of a `final` instance variable
```

src/_guides/language/language-tour.md
```diff
@@ -2845,10 +2845,10 @@ class ProfileMark {
...
-after the constructor body starts, you can use `late final`,
-[but _be careful_][late-final-ivar].
+after the constructor body starts, you can use one of the following:
 
-[late-final-ivar]: /guides/language/effective-dart/design#avoid-public-late-final-fields-without-initializers
+* Use `late final`, but [_be careful_][late-final-ivar].
+* Use a [factory constructor](#factory-constructors).
```
parlough, line 2851: I think a factory constructor may be a better first alternative to suggest as `late final` can result in some tricky to identify errors. What do you think?

```suggestion
* Use a [factory constructor](#factory-constructors).
* Use `late final`, but [_be careful_][late-final-ivar].
```

parlough (approved):
This change looks good to me. Two minor suggestions.

src/_guides/language/language-tour.md
kwalrath: sgtm, but I'd like to get the opinions of @natebosch and/or @munificent

natebosch: +1 on swapping the order of the suggestions.

Can we also inline a brief bit of the reason why these are dangerous?

```
* Use `late final`, but [_be careful_][late-final-ivar], this pattern introduces a setter to the API.
```

kwalrath: Thanks, both! I've added a variation on Nate's text:

```
* Use a [factory constructor](#factory-constructors).
* Use `late final`, but [_be careful_][late-final-ivar];
  a `late final` without an initializer adds a setter to the API.
```

PR #3585, opened 2021-09-19
Standardize site alert inclusions and formatting

parlough:


kwalrath:
Nice! Source LGTM, but I'm going to check the staged version to be sure.

Here's one weird thing I see:

image

Do you have any idea why GitHub couldn't show the changes to that file?

PR #3586, opened 2021-09-20
Reference 2.14 as SDK version for the site

parlough:
We already were using 2.14 for the misc examples, so update the documented version as well as the minimum SDK for other examples.


kwalrath (approved):
Thanks for doing this!

I keep forgetting to update pkg-vers.json (as documented in https://github.com/dart-lang/site-www/wiki/Updating-to-new-SDK-releases), so I'm going to put this in the Dart team's release checklist.

PR #3589, opened 2021-09-21
fix: close info site alert properly

albertms10:
Fixes a missing closing site alert in the standardization process #3585.


parlough (approved):
Thanks for catching this!

PR #3593, opened 2021-09-23
Switch to ignore rather than ignore_for_file for tracked analysis issues

### LINKED ISSUE #3591: Switch examples ignoring analysis errors and hints on purpose to use ignore
### Describe the problem

Currently, there are some `ignore_for_file`s used to ignore a lot of errors and hints for issues we want to keep for demonstration purposes. This can cause issues in case there are other errors in the file we don't want to ignore, making them easy to miss.

### Expected fix

Update those situations to use a single line `ignore` instead and fix any problems with the resulting code excerpts.

### Additional context

_No response_

parlough:
Adds the previously used replaces for `ignore` to the affected files. They are quite large but I wanted to stay consistent in handling all cases for these files. Since in the analysis options explanation, we want some of the ignore statements, I've added a specifier to look for one of the SDK channels.

Fixes #3591

### DIFF
--- a/examples/analysis/lib/assignment.dart
+++ b/examples/analysis/lib/assignment.dart
@@ -7,7 +7,7 @@ import 'package:examples_util/ellipsis.dart';
 String downcastExample() {
   // #docregion implicit-downcast
   dynamic o = ellipsis<String>();
-  // ignore_for_file: stable, beta, dev, invalid_assignment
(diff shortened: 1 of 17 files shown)


kwalrath (approved):
LGTM, thanks!

I didn't run a diff, but a spot check shows no ill effects from the markdown changes.

Staged, fwiw: https://kw-www-dartlang-1.web.app

PR #3594, opened 2021-09-23
Add some more detail about our compilation formats

mit-mit:


src/overview.md
```diff
@@ -240,6 +240,22 @@ More information:
...
+Regardless of how you compile your Dart code, a Dart runtime is needed to
+execute the code. This runtime is responsible for a number of critical tasks:
+
+* Managing memory: Dart uses a managed memory model, where unused memory is
+  reclaimed by a garbage collector (GC). 
+
+* Enforcing the Dart type system: While most type checks are static (compile
+  time), some type checks are dynamic (run time), e.g. the Dart
+  [type check and cast operators](/guides/language/language-tour#type-test-operators).
+
```
mraleph, line 254: I'd add also a note about isolates which are also part of the runtime system.

src/overview.md
```diff
@@ -240,6 +240,22 @@ More information:
...
+On the native platforms, the Dart runtime is automatically included inside
+self-contained executables, and is part of the Dart VM invoked with [`dart
+run`](https://dart.dev/tools/dart-run). On the web platforms, the runtime is
```
mraleph, line 257: The part about runtime on a Web is a bit confusing, e.g. GC is provided by the JavaScript runtime. I'd remove the last sentence to avoid confusing people too much.

mraleph (approved):

src/overview.md
mit-mit: But isn't there still a Dart runtime when running on the web?

There was a confusion recently where someone asked why they could not run the output of `dart compile js` in node, which I _think_ boiled down to having to include some JS code that essentially is the runtime?

src/overview.md
mit-mit: How about something like this?

```
* Managing concurrency through [isolates](/guides/language/language-tour#isolates),
  the isolation containers that all Dart code runs within.
```

src/overview.md
mraleph: Well, there are some parallel, but calling it runtime carries some connotations that I think we would like to avoid (slow, heavy, foreign). Yes, there is some code that needs to run to bootstrap things, some code that implements type checks and some stuff for deferred loading. That's about it.

There are differences between how code runs in node vs browser (differences in global object, module system, etc) hence some challenges in running `dart2js` code in node... But you would have similar issues with any other JavaScript which is not entirely platform agnostic.

src/overview.md
mraleph: Isolates are not units of isolation - that's too strong of a wording. Maybe just "Managing concurrently executing [isolates](...)".

src/tools/dart-compile.md
```diff
@@ -74,39 +75,43 @@ The following table shows the subcommands of `dart compile`.
   <tr>
     <td> <code>exe</code> </td>
     <td> <span style="white-space: nowrap">Self-contained</span> executable </td>
-    <td> A standalone, architecture-specific executable file.
-      <a href="#exe">Learn more.</a>
+    <td> A standalone, architecture-specific executable file containing the source code
```
kevmoo, line 78: LOVE THIS!

src/overview.md
mit-mit: OK, I'll remove the runtime part from web than.

src/tools/dart-compile.md
mit-mit: 👍

mit-mit:
OK, I think this is ready for you @kwalrath

src/overview.md
```diff
@@ -240,6 +240,23 @@ More information:
...
+
+* Enforcing the Dart type system: While most type checks are static (compile
+  time), some type checks are dynamic (run time), e.g. the Dart
+  [type check and cast operators](/guides/language/language-tour#type-test-operators).
+
+* Managing [isolates](/guides/language/language-tour#isolates) running concurrently.
+
+On the native platforms, the Dart runtime is automatically included inside
+self-contained executables, and is part of the Dart VM invoked with [`dart
+run`](https://dart.dev/tools/dart-run).
```
parlough, line 259: Splits up the phrases a bit and updates the `dart run` link to a relative reference.

```suggestion
On the native platforms, the Dart runtime is automatically
included inside self-contained executables, 
and is part of the Dart VM invoked with [`dart run`](/tools/dart-run).
```

src/tools/dart-compile.md
```diff
@@ -74,39 +75,43 @@ The following table shows the subcommands of `dart compile`.
...
+      compiled to machine code and a small <a href="/overview#runtime">Dart runtime</a>.
+      <br><i><a href="#exe">Learn more.</a></i>
```
parlough, line 80:
```suggestion
      <br><em><a href="#exe">Learn more.</a></em>
```

src/tools/dart-compile.md
```diff
@@ -74,39 +75,43 @@ The following table shows the subcommands of `dart compile`.
...
   </tr>
   <tr>
     <td style="white-space: nowrap"> <code>aot-snapshot</code> </td>
-    <td style="white-space: nowrap"> AOT snapshot </td>
-    <td> An architecture-specific file with <b>no Dart runtime</b>.
-      <a href="#aot-snapshot">Learn more.</a>
+    <td style="white-space: nowrap"> AOT module </td>
+    <td> An architecture-specific file, containing the source code
+      compiled to machine code, and <b>no Dart runtime</b>.
+      <br><i><a href="#aot-snapshot">Learn more.</a></i>
```
parlough, line 88:
```suggestion
      <br><em><a href="#aot-snapshot">Learn more.</a></em>
```

src/tools/dart-compile.md
```diff
@@ -74,39 +75,43 @@ The following table shows the subcommands of `dart compile`.
...
-    <td> JIT snapshot </td>
+    <td> JIT module </td>
     <td> An architecture-specific file with
-      parsed classes and compiled code that's generated during
-      a training run of the program.
-      <a href="#jit-snapshot">Learn more.</a>
+      an intermediate representation of all source code,
+      and an optimized representation of the source code
+      exercised during a training run of the program.
+      <br><i><a href="#jit-snapshot">Learn more.</a></i>
```
parlough, line 98:
```suggestion
      <br><em><a href="#jit-snapshot">Learn more.</a></em>
```

src/tools/dart-compile.md
```diff
@@ -74,39 +75,43 @@ The following table shows the subcommands of `dart compile`.
...
     <td> <code>kernel</code> </td>
-    <td> Kernel snapshot </td>
-    <td> A portable
-      <a href="https://github.com/dart-lang/sdk/blob/main/pkg/kernel/binary.md">binary snapshot</a>.
-      <a href="#kernel">Learn more.</a>
+    <td> Kernel module </td>
+    <td> A portable,
+      <a href="https://github.com/dart-lang/sdk/blob/main/pkg/kernel/binary.md">intermediate representation</a>
+      of the source code.
+      <br><i><a href="#kernel">Learn more.</a></i>
```
parlough, line 107:
```suggestion
      <br><em><a href="#kernel">Learn more.</a></em>
```

src/tools/dart-compile.md
```diff
@@ -74,39 +75,43 @@ The following table shows the subcommands of `dart compile`.
...
     </td>
   </tr>
   <tr>
     <td> <code>js</code> </td>
     <td> JavaScript </td>
     <td> A deployable JavaScript file.
-      <a href="#js">Learn more.</a>
+      <br><i><a href="#js">Learn more.</a></i>
```
parlough, line 114:
```suggestion
      <br><em><a href="#js">Learn more.</a></em>
```

src/tools/dart-compile.md
```diff
@@ -74,39 +75,43 @@ The following table shows the subcommands of `dart compile`.
...
     <td> <code>jit-snapshot</code> </td>
-    <td> JIT snapshot </td>
+    <td> JIT module </td>
     <td> An architecture-specific file with
-      parsed classes and compiled code that's generated during
-      a training run of the program.
-      <a href="#jit-snapshot">Learn more.</a>
+      an intermediate representation of all source code,
+      and an optimized representation of the source code
+      exercised during a training run of the program.
```
parlough, line 97: I'm not sure the use of "exercised" will be super clear here, could it simply say "**generated** during a training run..."?

```suggestion
      generated during a training run of the program.
```

src/overview.md
```diff
@@ -240,6 +240,23 @@ More information:
...
+#### The Dart Runtime {#runtime}
+
+Regardless of how you compile your Dart code, a Dart runtime is needed to
+execute the code. This runtime is responsible for a number of critical tasks:
+
+* Managing memory: Dart uses a managed memory model, where unused memory is
+  reclaimed by a garbage collector (GC). 
+
+* Enforcing the Dart type system: While most type checks are static (compile
+  time), some type checks are dynamic (run time), e.g. the Dart
```
parlough, line 252: Not exactly sure how to restructure this currently to make it sound best, but we may want to consider moving from the "e.g." to an alternative such as "For example," as suggested in [Abbreviations not to use](https://developers.google.com/style/abbreviations?hl=en#dont-use).

parlough (commented):
Some minor comments on wording and switching from `<i>` to `<em>` to be consistent with elsewhere on the website as `<i>` is occasionally used for icons.

src/tools/dart-compile.md
kevmoo: It's like @mit-mit is writing HTML from...1997 😜

mit-mit: I enrolled in my C.S. program in 1994, so probably more like 1995/1996 ;-)

src/tools/dart-compile.md
mit-mit: That changes the meaning quite a bit: The important part here is the only the code that is actually exercised/run by the training input is optimized. Got a better way to phrase that?

parlough: I see what you mean, maybe "reached" or "executed" instead of "exercised"?

mit-mit: Yeah, I used "exercised" as it seemed a good pairing with "training input".

src/tools/dart-compile.md
```diff
@@ -74,39 +75,43 @@ The following table shows the subcommands of `dart compile`.
...
+    <td style="white-space: nowrap"> AOT module </td>
+    <td> An architecture-specific file, containing the source code
+      compiled to machine code, and <b>no Dart runtime</b>.
+      <br><em><a href="#aot-snapshot">Learn more.</a></em>
     </td>
   </tr>
   <tr>
     <td> <code>jit-snapshot</code> </td>
-    <td> JIT snapshot </td>
+    <td> JIT module </td>
```
vsmenon, line 93: Are jit snapshots something we want to encourage people to use?  ( @mraleph @a-siva )

If so, when should they consider this over aot?

kwalrath:
Just fyi, I'm going to make an editing pass once the technical content has been settled.

src/tools/dart-compile.md
mraleph: I think we still have not settled on JIT vs AOT ourselves. As mentioned on the internal discussions: I do believe we might be leaving something like 30% performance on the table due team constraints which force us to have strong AOT focus.

People should consider JIT snapshots over AOT when they want peak performance and they have good training data.

mit-mit: OK, added `      JIT-compiled code can have faster peak performance than AOT code
      in some cases, when the training data is good.
`

mit-mit:
@kwalrath I think this is ready for you now

src/overview.md
```diff
@@ -240,6 +240,24 @@ More information:
 * [`webdev` tool](/tools/webdev)
 * [Web deployment tips](/web/deployment)
 
+#### The Dart Runtime {#runtime}
```
kwalrath, line 243: It seems a little weird to have this section under "Dart: The platforms", since the preceding sections are "Dart Native" and "Dart Web".

What do you think of having a new, second-level "Dart: The runtime" section? If that seems wrong to you, maybe we could change the intro to tie into the platforms. Something like "Regardless of which platform you use or how you compile your code, executing the code requires a Dart runtime."

kwalrath: Also, I'm dying to fix the white space ([semantic line breaks](https://github.com/dart-lang/site-shared/blob/master/doc/writing-for-dart-and-flutter-websites.md#semantic-line-breaks). I could do it in a followup PR, or I could do it now (while restraining myself from changing anything else).

Do you have a preference?

kwalrath: I directly edited dart-compile.md (as described in the commit) since it didn't need many changes. Could you please check my work, @mit-mit?

mit-mit: | Also, I'm dying to fix the white space

Sorry, all other repos in the org use automatically wrapped paragraphs at 80 cols, so I keep forgetting about that :-(

mit-mit: Edits LGTM!

parlough: I think the question about moving up these new "Dart: The runtime" details to a higher level still remains. Taking a look, I agree it's a bit strange to be with the platforms, so perhaps a new second-level heading like you mentioned would be good.

mit-mit: I think it's tolerable: the runtime itself is part of the platforms, and the platforms content here really is more of an architectural overview.

mit-mit:
@kwalrath I think we can merge this?

src/overview.md
kwalrath: Not quite, but if we change the intro sentence to "Regardless of which platform you use or how you compile your code, executing the code requires a Dart runtime" then I'm OK with it.

kwalrath: > @kwalrath I think we can merge this?

I'll change that intro sentence and (in a separate commit) fix the formatting. If those are OK by you, then I'll merge this.

mit-mit: SGTM!

parlough (approved):

PR #3604, opened 2021-10-01
Move up most important language resources

mit-mit:
The tour and type system docs are critical resources for any new Dart developer. This moves them to top of the left-nav to signal that.


parlough (approved):
This makes sense to me, particular moving the Language Tour to the top.

parlough (approved):

kwalrath (approved):

PR #3607, opened 2021-10-04
Update or remove old Firebase redirects

### LINKED ISSUE #3546: Verify Firebase declared redirects still lead to useful locations
The site has a lot of redirects, a few of which are quite old and potentially lead to dead or at least outdated sites.

We should verify each of them leads to a useful location, and if not try to point it to an actually useful/updated location. If there is no alternate, we should either remove the redirect so they get a 404 or add a new page for removed content.

parlough:
Closes #3546


firebase.json
```diff
@@ -23,7 +23,7 @@
       { "source": "/\\)", "destination": "/", "type": 301 },
       { "source": "/+", "destination": "https://plus.sandbox.google.com/+dartlang", "type": 301 },
       { "source": "/+dart2js", "destination": "/tools/dart2js", "type": 301 },
-      { "source": "/+isolates", "destination": "https://dart-lang.github.io/observatory/isolate.html", "type": 301 },
+      { "source": "/+isolates", "destination": "https://medium.com/dartlang/dart-asynchronous-programming-isolates-and-event-loops-bffc3e296a6a", "type": 301 },
```
domesticmouse, line 26: I'm curious where this link is referred from. I'm guessing @kwalrath isn't updating the medium article?

domesticmouse (approved):

firebase.json
parlough: Yeah I'm not sure of any references to the redirect, but the article seemed the best destination for now. I believe there may be new isolate related content in the future though which could perhaps replace it.

kwalrath (approved):
Changes LGTM. Nice work, @parlough!

PR #3603, opened 2021-10-01
Add some more detail to security policy

mit-mit:

PR #3608, opened 2021-10-04
Copy edit, reformat security page

kwalrath:
I don't think I changed the meaning of anything, but the most changed sentence is the second one under **Reporting vulnerabilities**, so please check that. I'd love to add ", respectively" to that sentence — but only if that's correct.

### DIFF
--- a/src/security.md
+++ b/src/security.md
@@ -3,33 +3,40 @@ title: Security
 ---
 
 The Dart team takes the security of Dart and the applications
-created with it seriously. This page describes how to report any
-vulnerabilities you may find, and lists best practices to minimize
-the risk of introducing a vulnerability.
+created with it seriously.
+This page describes how to report any vulnerabilities that you find,
+and lists best practices to minimize the risk of introducing a vulnerability.
 
 ## Reporting vulnerabilities
 
-To report a security issue, please use [https://g.co/vulnz](https://g.co/vulnz).
-We use g.co/vulnz for our intake, and do coordination and disclosure
-[on GitHub](https://github.com/dart-lang/)
-(including using GitHub Security Advisory). The Google Security
-Team will respond within 5 working days of your report on g.co/vulnz.
+To report a security issue, use [https://g.co/vulnz][].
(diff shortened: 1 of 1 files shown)


domesticmouse (approved):

src/security.md
```diff
@@ -3,33 +3,40 @@ title: Security
...
 ## Reporting vulnerabilities
 
-To report a security issue, please use [https://g.co/vulnz](https://g.co/vulnz).
-We use g.co/vulnz for our intake, and do coordination and disclosure
-[on GitHub](https://github.com/dart-lang/)
-(including using GitHub Security Advisory). The Google Security
-Team will respond within 5 working days of your report on g.co/vulnz.
+To report a security issue, use [https://g.co/vulnz][].
+Coordination and disclosure happen in the [dart-lang GitHub repos][repos]
+and the [GitHub Advisory Database][].
```
mit-mit, line 14: I don't think this is accurate. We publish Security Advisories:
https://github.com/dart-lang/sdk/security/advisories?state=published

But we don't use the central GitHub Advisory Database:
https://github.com/advisories

Is that correct @athomas ?

kwalrath: This text is just a slightly rearranged version of what was in #3603 (https://github.com/dart-lang/site-www/pull/3603/files).

parlough: I think the idea here is that Dart advisories are not added to https://github.com/advisories but instead are included in the repositories, such as https://github.com/dart-lang/sdk/security/advisories/GHSA-hfq3-v9pv-p627

Even though it has an ID it doesn't seem to be present in the central repository: https://github.com/advisories/GHSA-hfq3-v9pv-p627

kwalrath: Ah, so it's [security advisories](https://docs.github.com/en/code-security/security-advisories/about-github-security-advisories), not the GitHub Security Advisory Database. I'll update the link and link text.

parlough (approved):
These changes look great to me.

mit-mit:
LGTM

PR #3609, opened 2021-10-05
Update try_catch.dart

dash137:
Changed order of awaited function call and the announcement thereof, because the notification that something is being waited for should come beforehand

### DIFF
--- a/examples/async_await/test/async_await_test.dart
+++ b/examples/async_await/test/async_await_test.dart
@@ -53,7 +53,7 @@ void main() {
     final output = '''
       Awaiting user order...
       Caught error: Cannot locate user order
-    '''
+    ''';
     expect(try_catch.main, m.printsLines(output));
   });
 }


liasquil:
@googlebot I signed it!

parlough:
Hey @liasquil, thanks for opening a pull request!

I think this makes and it matches what we have elsewhere for this particular example. We also need to modify the examples which use the snippet in the actual codelab to get the tests to pass. Feel free to let me know if you'd like me to do this instead, but I've included the two locations I believe this snippet is included at:

- https://github.com/dart-lang/site-www/blob/master/src/codelabs/async-await.md?plain=1#L574-L577
- https://github.com/dart-lang/site-www/blob/master/src/codelabs/async-await.md?plain=1#L594-L595

Thanks again!

liasquil:
Hi @parlough, thanks for looking into my patch and listing the other occurrences. I've now adapted those occurrences correspondingly. I also couldn't find any other files by searching for "Awaiting user order" and "Cannot locate user order".
(By the way, can I somehow run the tests myself within my fork?)

liasquil:
Actually I just found some more.

PR #3618, opened 2021-10-08
feature/3615-jq-doc-body

### LINKED ISSUE #3615: Use simpler/faster `$(document.body)` instead of `$('body')` in main.js
### Describe the problem

We currently use an inconsistent mix of `$('body')` and `$("body")` within [main.js](https://github.com/dart-lang/site-www/blob/master/src/_assets/js/main.js). 

### Expected fix

We should switch these to the more consistent and slightly more performant `$(document.body)`. This will also very slightly simplify a transition away from jquery in the future.

### Additional context

_No response_

bstuff:
fixes #3615

### DIFF
--- a/src/_assets/js/main.js
+++ b/src/_assets/js/main.js
@@ -112,7 +112,7 @@ $(function () {
   // open - close mobile navigation
   $('#menu-toggle').on('click', function (e) {
     e.stopPropagation();
-    $("document.body").toggleClass('open_menu');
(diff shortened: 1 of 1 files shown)


bstuff:
@googlebot I signed it!

bstuff:
> Thanks for working on this but there shouldn't be quotes around `document.body`, so it should just be: `$(document.body)`

@parlough 🤦‍♂️😁 I fixed

Issue #3628, opened 2021-10-14
[PAGE ISSUE]: 'Write command-line apps'


yanadior9029:
### Page URL

https://dart.dev/tutorials/server/cmdline

### Page source

https://github.com/dart-lang/site-www/tree/master/src/_tutorials/server/cmdline.md

### Describe the problem

Please provide more guidance on which path file - quotes.txt will be picked up from. Through hit and trial, i figured out that 'quotes.txt' should be under root of the dart-project.
Is there any concept of classpath in Dart ?

domesticmouse:
No, Dart doesn't have a concept of a classpath, nor does it have WAR files. Dart's command line experience is closer to that of Python or Node.js than to that of Java. The current working directory for an application is the directory in which the executable is launched. 

With that said, I think we could be clearer [here](https://dart.dev/tutorials/server/cmdline#running-dcat) and spell out downloading `quotes.txt` using curl, say, before running the command.

PR #3633, opened 2021-10-18
Add context to command line tutorial

domesticmouse:
fixes: https://github.com/dart-lang/site-www/issues/3628


parlough:
I like the idea of adding more context and a quick way to download it, but I'm concerned this may confuse users who do not have access to `curl` on their shell. Or do you think this is a fair assumption to make?

domesticmouse:
I like the solution of running `dcat` against the `pubspec.yaml` as it removes any ambiguity as to where files go et al.

parlough:
@kwalrath Does this simplification seem good to you?

PR #3638, opened 2021-10-24
Switch from X-Frame-Options to CSP frame-ancestors to allow DartPad as an ancestor

parlough:
This does not fix all of the DartPad issues, but according to MDN: `X-Frame-Options` is obsoleted by  the `frame-ancestors` directive. Switching to that also allows us to specify DartPad as supported as an ancestor.

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -10,17 +10,10 @@
         ]
       },
       {
-      "source": "**",
-      "headers": [
-        { "key": "Access-Control-Allow-Origin", "value": "*" }
-      ]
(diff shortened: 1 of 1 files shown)


kwalrath:
@johnpryan could you please confirm that the latest changes fix the problem? Thanks!

parlough:
This shouldn't fix those problems, but just moves to the latest suggested standards for controlling frame behavior.

johnpryan (approved):

PR #3646, opened 2021-11-01
Validate Firebase redirect sources are sorted and unduplicated

### LINKED ISSUE #3644: Investigate verifying Firebase redirects are sorted and unduplicated
### Describe the problem

For readability and editing purposes, it is helpful for the firebase.json redirects to be sorted. This has become even more clear with https://github.com/flutter/website/pull/6397 which identified a repeated go link leading to different sources, which can be a larger problem that is easier to notice when sorted.

### Expected fix

Investigate adding a CI check to verify these are sorted and that there are no duplicates.

### Additional context

This should also be introduced on the Flutter website side.

parlough:
Closes #3644


tool/firebase_redirect_check.dart
```diff
@@ -0,0 +1,102 @@
...
+
+      sources.add(source);
+    }
+
+    if (duplicatesFound > 0) {
+      stderr.writeln(
+          'Error: $duplicatesFound duplicate sources found in the firebase.json redirects.');
+    }
+
+    final sourcesCopy = List.of(sources, growable: false);
```
johnpryan, line 76: nit: you could also do `sources.toList(growable: false)`

johnpryan (approved):
LGTM - this might be something we could add to [site-shared](https://github.com/dart-lang/site-shared) and use for the Flutter website too.

parlough:
@johnpryan That's my plan to investigate once the Flutter website freeze is over and it hasn't hit any snags here. I'll include any improvements then as well :)

PR #3647, opened 2021-11-01
Describe new syntax for hosted dependencies

sigurdm:


kwalrath:
Thanks, @sigurdm! I'll make a few minor copy edits directly to this PR.

We don't usually cover features before they're in stable, but this one seems advanced enough that we could make an exception.

src/tools/pub/dependencies.md
```diff
@@ -31,25 +31,24 @@ dependencies:
...
 This YAML code creates a dependency on the `transmogrify` package
-using the default source ([pub.dev]({{site.pub}})) and
+using the default package repository ([pub.dev]({{site.pub}})) and
 allowing any version from `1.0.0` to `2.0.0` (but not including `2.0.0`).
 See the [version constraints](#version-constraints)
 section of this page for syntax details.
 
-If you want to specify a source, the syntax looks a bit different:
+If you want to specify a custom package repository,
+use the following syntax:
```
kwalrath, line 40: I'm uneasy about showing the 2.15 syntax here before 2.15 has a stable channel release. One possible solution is to remove the example and do a forward link to #hosted-packages.

kwalrath: Do we know which syntax people are most likely to use? Which they're most likely to have trouble with?

kwalrath: Finally... Will the pre-2.15 syntax still work in 2.15+?

src/tools/pub/dependencies.md
```diff
@@ -114,15 +113,31 @@ This example specifies that your package depends on a hosted package named
...
+environment: 
+  sdk: >=2.15.0 < 3.0.0
+
+dependencies:
+  transmogrify:
+    hosted: http://some-package-server.com
+    version: ^1.4.0
+{% endprettify %}
+
+Before Dart language version 2.15, the format is different:
```
parlough, line 130: Depending on when this PR is merged, do we want to separate the old version with a version-note mentioning language version like we do in https://github.com/dart-lang/site-www/blob/master/src/_guides/language/analysis-options.md?plain=1#L195-L201? To be consistent with elsewhere on the website. Then in the currently supported version above, don't add the "# This syntax requires language version 2.15 or higher" comment?

sigurdm:
> We don't usually cover features before they're in stable, but this one seems advanced enough that we could make an exception.

It is absolutely fine to wait merging this until 2.15 goes live! I just wanted to make sure this was written up...

Given the response on https://github.com/dart-lang/pub/issues/1806#issuecomment-956069925 we might even backpedal and just introduce the new syntax in 2.15.0 without the language-versioning logic, so we should at least wait merging until that is settled.

src/tools/pub/dependencies.md
sigurdm: > I don't want to show the 2.15 syntax here before 2.15 has a stable channel release.

Good point. I guess the first example is mainly to explain the concept of a description. We could instead let that be a git or path dependency... Or just waiting merging this until the 2.15 release.

> Do we know which syntax people are most likely to use? Which they're most likely to have trouble with?

Hopefully the new syntax! We're adding much better support for third-party repositories in 2.15, so it gets much more interesting to specify.

> Finally... Will the pre-2.15 syntax still work in 2.15+?

Yes it will.

src/tools/pub/dependencies.md
```diff
@@ -31,25 +31,24 @@ dependencies:
...
 
 {% prettify yaml tag=pre+code %}
 dependencies:
   transmogrify:
-    hosted:
-      name: transmogrify
-      url: http://some-package-server.com
+    hosted: 'http://some-package-server.com'
     version: ^1.0.0
 {% endprettify %}
```
jonasfj, line 47:
```suggestion
    version: ^1.0.0
environment:
  sdk: >= 2.15.0 <3.0.0
{% endprettify %}
```

kwalrath:
Looking at this page a bit more deeply, I don't see why it bothers to define _description_, and the intro could be simpler and clearer. I've been working on an update, but it's not ready to post yet. Tomorrow, I hope!

kwalrath:
@sigurdm ptal. Your changes inspired me to simplify the overview and some formatting on the page. Hopefully it's more understandable now!

src/tools/pub/dependencies.md
```diff
@@ -78,54 +66,77 @@ Currently, Flutter is the only SDK that is supported.
...
 The identifier after `sdk:` indicates which SDK the package comes from.
 If it's `flutter`, the dependency is satisfiable as long as:
 
 * Pub is running in the context of the `flutter` executable
 * The Flutter SDK contains a package with the given name
-* That package's version matches the version constraint
 
 If it's an unknown identifier, the dependency is always considered unsatisfied.
 
 A package with an `sdk` dependency
```
sigurdm, line 83: When do we remove documentation for old versions? Version 1.19 starts to be pretty niche...

sigurdm (commented):
Nice updates! - I have a few extra suggestions to consider...

src/tools/pub/dependencies.md
sigurdm: Done

src/tools/pub/dependencies.md
```diff
@@ -78,54 +66,77 @@ Currently, Flutter is the only SDK that is supported.
 
 The syntax looks like this:
```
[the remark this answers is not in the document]
sigurdm: Done

src/tools/pub/dependencies.md
```diff
@@ -173,13 +184,13 @@ The ref can be anything that Git allows to [identify a commit.][commit]
 Pub assumes that the package is in the root of the Git repository.
 To specify a different location in the repo, use the `path` argument:
```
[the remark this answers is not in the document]
sigurdm: Done

src/tools/pub/dependencies.md
```diff
@@ -70,62 +58,58 @@ Pub can use the following sources to locate packages:
...
 {% prettify yaml tag=pre+code %}
+environment: 
+  sdk: >=[!2.15.0!] < 3.0.0
+
 dependencies:
   transmogrify:
-    hosted:
-      name: transmogrify
-      url: http://your-package-server.com
+    [!hosted: http://some-package-server.com!]
```
parlough, line 89: With the new `https://dart.dev/tools/linter-rules#secure_pubspec_urls` linter rule, perhaps we should use `https` links in the examples.

```suggestion
    [!hosted: https://some-package-server.com!]
```

src/tools/pub/dependencies.md
```diff
@@ -70,62 +58,58 @@ Pub can use the following sources to locate packages:
...
+
+{% prettify yaml tag=pre+code %}
+environment:
+  sdk: >=[!2.14.0!] < 3.0.0
+
+dependencies:
+  transmogrify:
+    [!hosted:!]
+      [!name: transmogrify!]
+      [!url: http://some-package-server.com!]
```
parlough, line 108: Same here.

```suggestion
      [!url: https://some-package-server.com!]
```

src/tools/pub/dependencies.md
```diff
@@ -149,40 +133,43 @@ Even if the package repo is private, if you can
...
   kittens:
     git:
       url: git@github.com:munificent/cats.git
       path: path/to/kittens
-{% endprettify %}
+```
 
 The path is relative to the Git repo's root.
 
+Git dependencies are not allowed on [pub.dev][pubsite].
```
parlough, line 171: I wonder if it would be more clear if we specify in what way they are not allowed. What do you think?

```suggestion
Git dependencies are not allowed as dependencies
for packages uploaded to [pub.dev][pubsite].
```

parlough (approved):
Looks great to me, thanks. Some possible minor suggestions.

src/tools/pub/dependencies.md
sigurdm: Yeah - that's better!

Dfurb87 (changes requested):
Blocked

kwalrath (approved):
I'm not crazy about having [code in a version note](https://dartlang-org-staging-1.web.app/tools/pub/dependencies#hosted-packages) (too much white space), but we will survive. :)

kwalrath:
Restaged: https://dartlang-org-staging-1.web.app/tools/pub/dependencies

kwalrath:
This page has conflicts; I'll work on resolving them.

kwalrath (approved):

PR #3655, opened 2021-11-03
Fix TypeError in its_all_about_you example.

### LINKED ISSUE #3654: TypeError in example
### Page URL

https://dart.dev/tutorials/web/fetch-data

### Page source

_No response_

### Describe the problem

I see the following error while running its_all_about_you app:
Uncaught Error: TypeError: Instance of 'JSArray<dynamic>': type 'JSArray<dynamic>' is not a subtype of type 'List<String>'

### Expected fix

Edit the example code to cast correctly.

### Additional context

_No response_

kshiitiz:
Closes #3654

### DIFF
--- a/examples/fetch_data/web/main.dart
+++ b/examples/fetch_data/web/main.dart
@@ -76,7 +76,7 @@ void _populateFromJson() {
   favoriteNumber.value = jsonData['favoriteNumber'].toString();
   valueOfPi.value = jsonData['valueOfPi'].toString();
   horoscope.value = jsonData['horoscope'].toString();
-  final favoriteThings = jsonData['favoriteThings'] as List<String>;
+  final favoriteThings = List<String>.from(jsonData['favoriteThings']);
   favOne.value = favoriteThings[0];
   favTwo.value = favoriteThings[1];
   favThree.value = favoriteThings[2];
--- a/src/_tutorials/web/fetch-data.md
+++ b/src/_tutorials/web/fetch-data.md
@@ -135,7 +135,7 @@ void _populateFromJson() {
   favoriteNumber.value = jsonData['favoriteNumber'].toString();
   valueOfPi.value = jsonData['valueOfPi'].toString();
   horoscope.value = jsonData['horoscope'].toString();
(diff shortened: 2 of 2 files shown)


examples/fetch_data/web/main.dart
```diff
@@ -76,7 +76,7 @@ void _populateFromJson() {
   favoriteNumber.value = jsonData['favoriteNumber'].toString();
   valueOfPi.value = jsonData['valueOfPi'].toString();
   horoscope.value = jsonData['horoscope'].toString();
-  final favoriteThings = jsonData['favoriteThings'] as List<String>;
+  final favoriteThings = List<String>.from(jsonData['favoriteThings']);
```
parlough, line 79: What do you think about changing this to the following which JSArray properly subtypes, so we can fix the current analysis error causing the build to fail?

```suggestion
  final favoriteThings = jsonData['favoriteThings'] as List;
```

src/_tutorials/web/fetch-data.md
```diff
@@ -135,7 +135,7 @@ void _populateFromJson() {
   favoriteNumber.value = jsonData['favoriteNumber'].toString();
   valueOfPi.value = jsonData['valueOfPi'].toString();
   horoscope.value = jsonData['horoscope'].toString();
-  final favoriteThings = jsonData['favoriteThings'] as List<String>;
+  final favoriteThings = List<String>.from(jsonData['favoriteThings']);
```
parlough, line 138: If that's fine, then we need to change this one as well.

```suggestion
  final favoriteThings = jsonData['favoriteThings'] as List;
```

parlough (changes requested):
Thanks so much for catching and fixing this! The fix works but causes analysis to fail:

```
error - web/main.dart:79:44 - The argument type 'dynamic' can't be assigned to the parameter type 'Iterable<dynamic>'. - argument_type_not_assignable
```

For now what do you think about just casting to a `List` as well? I believe this should fix this error.

kshiitiz:
Closing this pull request. Created a new one.

Issue #3222, opened 2021-05-11
Follow up additions for linter rule inclusion


parlough:
- [x] Include rule set(s) in the listings in some fashion, particularly the core recommended [lints](https://pub.dev/packages/lints): core, recommended, and flutter. (https://github.com/dart-lang/linter/pull/2691) (https://github.com/dart-lang/site-www/pull/3290) (https://github.com/dart-lang/site-www/pull/3498)
- [x] Include documentation about what each of the rule sets is useful for and when to choose them. (https://github.com/dart-lang/site-www/pull/3228)
- [x] Include incompatible rules in the rule listings. (https://github.com/dart-lang/linter/commit/bf8de841def00dd94837493ac327bb75f59aecf2) (https://github.com/dart-lang/site-www/pull/3290) (https://github.com/dart-lang/site-www/pull/3306)
- [ ] Update other locations to refer to this link where appropriate.  
    - [ ] The sidebar

This issue follows up on PR #3064, which added https://dart.dev/tools/linter-rules.

parlough:
Once the next linter release(1.5.0) has its documentation generated, it will now include incompatible sets:
https://github.com/dart-lang/linter/commit/bf8de841def00dd94837493ac327bb75f59aecf2

The goal is to finalize the work to include the new rule sets as well:
https://github.com/dart-lang/linter/pull/2691

parlough:
Another piece of information that may be worth including in the future: https://github.com/dart-lang/linter/issues/2650

Issue #3650, opened 2021-11-02
Add page about understanding error and warning messages


kwalrath:
### What information is missing?

@bwilkerson wrote a [blog post about diagnostic messages](https://medium.com/dartlang/dart-diagnostic-messages-ae302aa398e8). @mit-mit would like the post's content summarized in its own page on this site, under the **Static analysis** topic (probably above **Customizing static analysis**).

parlough:
I think the suggested position in the menu makes sense. :+1: 

Just as a note, this would be great to link to from https://dart.dev/tools/diagnostic-messages as well but that would have to be done on the SDK side.

atsansone:
@MaryaBelanger : I think this should be part of your diagnostic work, right?

parlough:
Yep!

antfitch:
Tentatively closing this one as it looks like @parlough has addressed this.

Issue #2579, opened 2020-08-14
Add links from static analysis messages to diagnostic-messages?


kwalrath:
https://dart.dev/guides/language/sound-problems lists some static analysis messages. We used to auto-include them, but as of 2.9, we can't ignore errors (https://github.com/dart-lang/sdk/issues/42977), so we now avoid testing known bad code (#2575) and can't auto-include the messages.

To compensate a bit, we could link from the message in sound-problems to the message's entry in https://dart.dev/tools/diagnostic-messages. That way we could provide some more info and (if the message name changes) provide a way to find the new message.

/cc @bwilkerson @theacodes @johnpryan

atsansone:
@MaryaBelanger : Is this on your plan for updating the diagnostic messages?

PR #3663, opened 2021-11-08
Add diagnostic messages and linter rules to sidebar

parlough:
What do you think about adding these to the sidebar under "Static analysis"?

Closes https://github.com/dart-lang/site-www/issues/3222

### DIFF
--- a/src/_data/side-nav.yml
+++ b/src/_data/side-nav.yml
@@ -253,6 +253,10 @@
         permalink: /guides/language/analysis-options
       - title: Fixing common type problems
         permalink: /guides/language/sound-problems
+      - title: Diagnostic messages
+        permalink: /tools/diagnostic-messages
+      - title: Linter rules
+        permalink: /tools/linter-rules
   - title: Testing & optimization
     children:
       - title: Testing


kwalrath:
It does seem about time to put our top page in the sidenav. :) And the linter page, as well.

/cc @bwilkerson @devoncarew @pq

bwilkerson:
The one counterargument that I can think of is one of the ones we originally had for not putting it there: it seems unlikely to be a page that anyone would want to navigate to other than as a result of wanting more information about a specific diagnostic, and in those cases they're likely to get there by either clicking in the IDE or the console (where they ran `dart analyze`). I don't know whether that's still a compelling reason.

kwalrath:
> The one counterargument that I can think of is one of the ones we originally had for not putting it there: it seems unlikely to be a page that anyone would want to navigate to other than as a result of wanting more information about a specific diagnostic, and in those cases they're likely to get there by either clicking in the IDE or the console (where they ran `dart analyze`). I don't know whether that's still a compelling reason.

Yes, that occurred to me too. But it seemed weird to have a page with such heavy usage that isn't in the sidenav. We could put a query parameter on the sidenav if we want to count how many people use it :)

parlough:
I can definitely understand your thinking. Anecdotally though I've found myself referring to the page quite a lot as there is a lot of great information present in the document. As we expand other analysis related documentation, it may be even more useful.

@kwalrath I've never set up UTM query parameters before. What should they look like in a situation like this?

Something like the following?

```
?utm_source=dartdev&utm_medium=sidenav&utm_campaign=diagnostics
```

pq:
> I say we add them and track the usage.

💯 

I'm in favor.

This all looks great to me. 👍

kwalrath:
> ```
> ?utm_source=dartdev&utm_medium=sidenav&utm_campaign=diagnostics
> ```

lgtm!

/cc @kevmoo

kevmoo:
Since these are not redirects, we can use normal referral logic to track hits. No need for utm stuff...I think...

kwalrath:
> Since these are not redirects, we can use normal referral logic to track hits. No need for utm stuff...I think...

How would we know which hits are from the sidenav?

kevmoo:
> > Since these are not redirects, we can use normal referral logic to track hits. No need for utm stuff...I think...
> 
> How would we know which hits are from the sidenav?

We see links from pages on the site. We can size page X -> page Y

That info doesn't show up with a redirect, hence why we need the utm things

kwalrath:
> > > Since these are not redirects, we can use normal referral logic to track hits. No need for utm stuff...I think...
> > 
> > 
> > How would we know which hits are from the sidenav?
> 
> We see links from pages on the site. We can size page X -> page Y

True, we can see that it isn't the landing page. And assume that it was the sidenav that resulted in the visit, although another possibility is in-site search.

parlough:
Also some pages may separately link to these pages, for example a [page explaining diagnostics](https://github.com/dart-lang/site-www/issues/3650) or links from [mentions of specific diagnostics](https://github.com/dart-lang/site-www/issues/2579).

kwalrath:
I'm going to merge this. We can always add utm stuff later.

kwalrath (approved):

PR #3664, opened 2021-11-08
Add a warning about Random not being cryptographically secure

parlough:
Closes #1271

PR #3665, opened 2021-11-08
Suggest using Object's convenience hashing functions for overriding hashCode

### LINKED ISSUE #3359: [2.14 Stable Release] Suggest using `Object` convenience hashing functions
Rolling your own hashing logic can potentially cause issues, so perhaps it would be a good idea for us to suggest the usage of the newly provided hash functions from https://github.com/dart-lang/sdk/commit/83376bf1ee81607a182f8558b242f1c9f4883246

This can help avoid potential implementation issues and consistency is always great.

https://github.com/dart-lang/sdk/blob/master/CHANGELOG.md#dartcore

parlough:
Fixes #3359

### DIFF
--- a/src/_guides/libraries/library-tour.md
+++ b/src/_guides/libraries/library-tour.md
@@ -919,6 +919,20 @@ also want to override the `==` operator. Objects that are equal (via
 `==`) must have identical hash codes. A hash code doesn’t have to be
 unique, but it should be well distributed.
 
+{{site.alert.tip}}
+  To consistently and easily implement the `hashCode` getter,
+  consider using the static hashing methods provided by the `Object` class.
+
+  For multiple objects, consider using [`Object.hash`][], 
+  while for collections
+  consider [`Object.hashAll`][] or [`Object.hashAllUnordered`][] 
+  according to your object's definition of equality. 
+{{site.alert.end}}
+
+[`Object.hash`]: {{site.dart_api}}/{{site.data.pkg-vers.SDK.channel}}/dart-core/Object/hash.html
+[`Object.hashAll`]: {{site.dart_api}}/{{site.data.pkg-vers.SDK.channel}}/dart-core/Object/hashAll.html
+[`Object.hashAllUnordered`]: {{site.dart_api}}/{{site.data.pkg-vers.SDK.channel}}/Object/hashAllUnordered.html
+
 {% comment %}
 Note: There’s disagreement over whether to include identical() in the ==
 implementation. It might improve speed, at least when you need to
@@ -933,15 +947,9 @@ class Person {
 
   Person(this.firstName, this.lastName);
 
-  // Override hashCode using strategy from Effective Java,
-  // Chapter 11.
+  // Override hashCode using Object's static hashing functions.
   @override
-  int get hashCode {
-    int result = 17;
-    result = 37 * result + firstName.hashCode;
-    result = 37 * result + lastName.hashCode;
-    return result;
-  }
+  int get hashCode => Object.hash(firstName, lastName);
 
   // You should generally implement operator == if you
   // override hashCode.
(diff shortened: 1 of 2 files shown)


src/_guides/libraries/library-tour.md
```diff
@@ -1511,6 +1519,12 @@ var random = Random();
 random.nextBool(); // true or false
 ```
 
+{{site.alert.warning}}
```
kevmoo, line 1522: Uh...avoid unrelated changeds in a PR. Split this out.

src/_guides/libraries/library-tour.md
```diff
@@ -919,6 +919,20 @@ also want to override the `==` operator. Objects that are equal (via
 `==`) must have identical hash codes. A hash code doesn’t have to be
 unique, but it should be well distributed.
 
+{{site.alert.tip}}
```
kevmoo, line 922: Good idea!

src/_guides/libraries/library-tour.md
parlough: Woops, included this by accident, thanks for catching that! You can find those changes in https://github.com/dart-lang/site-www/pull/3664 :)

src/_guides/libraries/library-tour.md
```diff
@@ -919,6 +919,20 @@ also want to override the `==` operator. Objects that are equal (via
...
+  To consistently and easily implement the `hashCode` getter,
+  consider using the static hashing methods provided by the `Object` class.
+
+  For multiple objects, consider using [`Object.hash`][], 
+  while for collections
+  consider [`Object.hashAll`][] or [`Object.hashAllUnordered`][] 
+  according to your object's definition of equality.
```
kwalrath, line 929: I believe we use `()` after function names, so I added those. (They'd need to be added to the link definitions as well, of course.)

Also, I didn't understand the multiple objects vs. collections thing, and I wanted to reduce sentence length. So... Here's my attempt:

```suggestion
  To generate a single hash code for multiple objects,
  you can use [`Object.hash()`][].
  To generate a hash code for a collection,
  you can use either [`Object.hashAll()`][] (if element order matters)
  or [`Object.hashAllUnordered()`][].
```

I vaguely remember us talking about collection member terminology recently (_element? item? member?_) but I don't remember what we decided.

kwalrath (commented):
Nice! I'd like to tweak the wording, but the code and gist of this lgtm.

src/_guides/libraries/library-tour.md
parlough: These changes look good to me, thanks. I do believe we discussed element as the best choice here and it makes sense in this case as well.

parlough:
@lrhn Do you have any thoughts on this mention?

src/_guides/libraries/library-tour.md
```diff
@@ -919,6 +919,21 @@ also want to override the `==` operator. Objects that are equal (via
 `==`) must have identical hash codes. A hash code doesn’t have to be
 unique, but it should be well distributed.
 
+{{site.alert.tip}}
+  To consistently and easily implement the `hashCode` getter,
+  consider using the static hashing methods provided by the `Object` class.
+
+  To generate a single hash code for multiple objects,
```
lrhn, line 926: I'd say: "objects" -> "properties of an object"

src/_guides/libraries/library-tour.md
```diff
@@ -933,15 +948,9 @@ class Person {
 
   Person(this.firstName, this.lastName);
 
-  // Override hashCode using strategy from Effective Java,
-  // Chapter 11.
+  // Override hashCode using Object's static hashing functions.
```
lrhn, line 951: Object -> \`Object\`

lrhn (approved):

kwalrath (approved):
I don't think single quotes are really necessary for `//` comments, but they're not hurting anyone.

PR #3666, opened 2021-11-08
Clarify when dependency overrides are used

sigurdm:


src/tools/pub/dependencies.md
```diff
@@ -396,6 +396,9 @@ dependency_overrides:
   may break your application.
 {{site.alert.end}}
 
+Only dependency overrides from he root pubspec are taken into account
+when doing package resolution.
```
jonasfj, line 400: I think what we want to clearly state is that `dependency_overrides` from the root package is considered.
But `dependency_overrides` inside a package that you depend on will be ignored.

Which means when you publish a package to pub.dev `dependency_overrides` will be ignored, by anyone who uses the package.

parlough: @sigurdm Any thoughts on the above comments?

I think including this mention is a great idea, Jonas's first sentence particularly makes sense.

sigurdm: Yeah - I took it as implied - but it is totally worth highlighting!

domesticmouse (approved):

src/tools/pub/dependencies.md
```diff
@@ -396,6 +396,13 @@ dependency_overrides:
   may break your application.
 {{site.alert.end}}
 
+Only dependency overrides from he root pubspec are considered
```
parlough, line 399:
```suggestion
Only dependency overrides from the root pubspec are considered
```

src/tools/pub/dependencies.md
```diff
@@ -396,6 +396,13 @@ dependency_overrides:
   may break your application.
 {{site.alert.end}}
 
+Only dependency overrides from the root pubspec are considered
+when doing package resolution. This means that dependency overrides
+inside a package that you depend on will be ignored.
+
+Therefore when you publish a package to pub.dev, dependency overrides 
+will be ignored by anyone depending on that package.
```
parlough, line 404:
```suggestion
Only dependency overrides from the root pubspec are considered
when doing package resolution.
This means that dependency overrides
inside a package that you depend on will be ignored.

Therefore, when you publish a package to pub.dev,
dependency overrides will be ignored by anyone depending on that package.
```

parlough (approved):
This clarification looks good to me and I can see it preventing some confusion. Thanks!

kwalrath (approved):

parlough:
Looks great!

PR #3669, opened 2021-11-09
Improve wording in cheatsheet code example

parlough:
This may not be the best improvement, but the previous text didn't really make sense. Open to suggestions!


kwalrath (approved):
I can't think of anything better, offhand

PR #3680, opened 2021-11-16
Add documentation on package retraction

szakarias:


src/tools/pub/publishing.md
```diff
@@ -326,6 +326,41 @@ then you can ignore the following warning from `dart pub publish`:
...
+
+Retracting a package causes churn and has a somewhat negativ impact on consumers. 
+Hence, before retracting a version you should consider whether it is in fact the 
+right course of action instead of just publishing a new version. For instance, 
+if your package version contains a minor bug then retraction is probably not 
+needed. On the other hand, if you accidentally published a major version or 
+published a version with wrong or missing dependency constraints, then retration 
+might be the right solution. 
+
+This feature is available from *Dart 2.15*.
```
jonasfj, line 362: Maybe, we need to clarify that this means Dart prior to 2.15 has no notion of retracted package versions and will use them regardless of whether or not they are in the `pubspec.lock`.

jonasfj (approved):

src/tools/pub/publishing.md
parlough: Just to verify, is this dependent on the user's Dart version or the pubspec's SDK constraint?

jonasfj: The retracted feature depends on the users Dart SDK version.

src/tools/pub/publishing.md
```diff
@@ -359,7 +359,11 @@ needed. On the other hand, if you accidentally published a major version or
 published a version with wrong or missing dependency constraints, then retration
```
sigurdm, line 359:
```suggestion
published a version with wrong or missing dependency constraints, then retraction 
```

src/tools/pub/publishing.md
```diff
@@ -359,7 +359,11 @@ needed. On the other hand, if you accidentally published a major version or
 published a version with wrong or missing dependency constraints, then retration 
 might be the right solution. 
 
-This feature is available from *Dart 2.15*.
+{{site.alert.version-note}}
+  The `retracted` feature debuted in Dart 2.15. Prior to Dart 2.15, pub 
+  will disregard any retracted status, i.e. the version solver will consider
```
sigurdm, line 364:
```suggestion
  will disregard the retracted status, i.e. the version solver will consider
```

src/tools/pub/publishing.md
```diff
@@ -359,7 +359,11 @@ needed. On the other hand, if you accidentally published a major version or
...
+  a retracted version as any other version.
```
sigurdm, line 365:
```suggestion
  a retracted version on par with any other version.
```

sigurdm (approved):
A few suggestions/typos/nits

src/tools/pub/publishing.md
```diff
@@ -326,6 +326,45 @@ then you can ignore the following warning from `dart pub publish`:
...
    instead."*
 
+
+## Rectracting a package version {#retract}
+
+If you wish to prevent new package consumers from adopting a recently
+published version of your package, you may **retract** that package version.   
+
+A retracted package version is not deleted, it will still show up in the version 
+list of the package on pub.dev, in the "Retracted versions" section. Also, the
```
sigurdm, line 336:
```suggestion
listing of the package on pub.dev, in the "Retracted versions" section. Also, the 
```

src/tools/pub/publishing.md
```diff
@@ -326,6 +326,45 @@ then you can ignore the following warning from `dart pub publish`:
...
+If you are a package consumer and depend on a version that later gets retracted,
+you can still use it as long as it is present in your `pubspec.lock` file. 
+If you want to depend on a specific version that is already retracted you need 
+to pin the version in the `dependency_overrides` section in the `pubspec.yaml` file. 
+In any other case the retracted package version will not be taken into account by 
+the version solver.
+
+### When does it make sense to retract a package version?
+
+Retracting a package causes churn and has a somewhat negativ impact on consumers.
```
sigurdm, line 354:
```suggestion
Retracting a package causes churn and has a somewhat negative impact on consumers. 
```

src/tools/pub/publishing.md
```diff
@@ -326,6 +326,45 @@ then you can ignore the following warning from `dart pub publish`:
...
+If you want to depend on a specific version that is already retracted you need 
+to pin the version in the `dependency_overrides` section in the `pubspec.yaml` file. 
+In any other case the retracted package version will not be taken into account by 
+the version solver.
+
+### When does it make sense to retract a package version?
+
+Retracting a package causes churn and has a somewhat negativ impact on consumers. 
+Hence, before retracting a version you should consider whether it is in fact the 
+right course of action instead of just publishing a new version. For instance,
```
sigurdm, line 356: Perhaps simplify the sentence a bit (and add a paragraph):
```suggestion
Hence, before retracting a version you should consider whether it would be better to just publish a new version.

For instance, 
```

sigurdm (approved):
A few suggestions/typos/nits

src/tools/pub/publishing.md
szakarias: done

src/tools/pub/publishing.md
szakarias: done

src/tools/pub/publishing.md
szakarias: Done

src/tools/pub/publishing.md
szakarias: Done

src/tools/pub/publishing.md
szakarias: done

src/tools/pub/publishing.md
szakarias: Done

szakarias:
Comments

kwalrath:
@szakarias I did a copy edit pass, which ended up rearranging the section in a kind of major way (for example, putting the "when to retract" stuff at the top and the actual directions at the bottom). Could you please take a look at my changes to make sure they make sense and are helpful? Thanks!

Staged version: https://kw-www-dartlang-1.web.app/tools/pub/publishing#retract

src/tools/pub/publishing.md
```diff
@@ -319,13 +319,59 @@ pub.dev updates the package listing to display
...
+within 7 days of publication.
+The retracted version can be restored again within 7 days of retraction.
+
+A retracted package version isn't deleted. It appears in the version
+listing of the package on pub.dev in the **Retracted versions** section. Also, the
+detailed view of that package version has a **RETRACTED** badge.
+
+Before retracting a package,
+consider publishing a new version instead.
+Retracting a package causes churn and can have a negative impact on package users.
```
parlough, line 343: Maybe it's just me, but seeing "churn" here is a bit confusing and doesn't seem to add much. I'm not sure if it's necessary to have this portion of the statement or it could be clarified further. I'm fine with no changes here as well!

parlough (approved):
Overall looks great to me and the functionality and its conditional behavior all make sense to me after reading this.

szakarias:
@kwalrath looks great, thanks! We might consider modifying the last paragraph in the "Marking packages as discontinued" section below so that it matches the wording in the "How to retract or restore a package version" section.

src/tools/pub/publishing.md
szakarias: Thanks for reading this through, good to hear that the description makes sense overall. 

I'll give it some thought to see if I can find a more clear phrasing instead of "churn", otherwise I might keep it as is. Thanks again.

kwalrath:
> @kwalrath looks great, thanks! We might consider modifying the last paragraph in the "Marking packages as discontinued" section below so that it matches the wording in the "How to retract or restore a package version" section.

Oh, good catch. Want me to do that in this PR?

szakarias:
> > @kwalrath looks great, thanks! We might consider modifying the last paragraph in the "Marking packages as discontinued" section below so that it matches the wording in the "How to retract or restore a package version" section.
> 
> Oh, good catch. Want me to do that in this PR?

If it is not too much work, I think it would be nice to do it in the same PR. I can also give it a shot, let me know what you prefer. :-)

kwalrath:
> > > @kwalrath looks great, thanks! We might consider modifying the last paragraph in the "Marking packages as discontinued" section below so that it matches the wording in the "How to retract or restore a package version" section.
> > 
> > 
> > Oh, good catch. Want me to do that in this PR?
> 
> If it is not too much work, I think it would be nice to do it in the same PR. I can also give it a shot, let me know what you prefer. :-)

I'll be happy to do it, probably tomorrow.

firebase.json
```diff
@@ -145,6 +145,7 @@
       { "source": "/go/non-promo-write", "destination": "/tools/non-promotion-reasons#write", "type": 301 },
       { "source": "/go/null-safety-migration", "destination": "/null-safety/migration-guide", "type": 301 },
       { "source": "/go/pubignore", "destination": "/tools/pub/publishing#what-files-are-published", "type": 301 },
+      { "source": "/go/pubretract", "destination": "/tools/pub/publishing#retract", "type": 301 },
```
jonasfj, line 148:
```suggestion
      { "source": "/go/package-retraction", "destination": "/tools/pub/publishing#retract", "type": 301 },
```

@szakarias thoughts?

sigurdm: I also like this better, so it does not get confused with a pub command

kwalrath: I like the clearer URL. Can we use it, @szakarias?

kwalrath (approved):
LGTM, although we might want to change the go link

src/tools/pub/publishing.md
```diff
@@ -319,13 +319,59 @@ pub.dev updates the package listing to display
...
    version 2.12.0-0, consider publishing the package as a pre-release
    instead."*
 
+
+## Retracting a package version {#retract}
+
+To prevent new package consumers from adopting a recently
+published version of your package, you can retract that package version
+within 7 days of publication.
+The retracted version can be restored again within 7 days of retraction.
```
kevmoo, line 335: @mit-mit to clarify if the the 7-day window keeps getting extended...or not.

src/tools/pub/publishing.md
```diff
@@ -319,13 +319,59 @@ pub.dev updates the package listing to display
...
+
+A retracted package version isn't deleted. It appears in the version
+listing of the package on pub.dev in the **Retracted versions** section. Also, the
+detailed view of that package version has a **RETRACTED** badge.
+
+Before retracting a package,
+consider publishing a new version instead.
+Retracting a package causes churn and can have a negative impact on package users.
+
+For example, if your package version contains a minor bug, then retraction is
```
kevmoo, line 345: I'd argue that the **best** example of when this should be used is a missing or overly weak dependency constraint that allows versions of SDKs or other packages that won't work. In this case, publishing a new version won't help because folks can still "solve" to the broken version with the too-loose constraints.

src/tools/pub/publishing.md
mit-mit: @szakarias @jonasfj the question is: if a developer toggles between retracted or not, does that A) add 7 days to the window, or B) doesn't change the window; i.e. the window is always 7 days after publication ?

jonasfj: The window extends so that you always have 7 days to undo the retraction. But 7 days after publishing you can't flip it to retracted again.

src/tools/pub/publishing.md
jonasfj: It actually says that for minor bugs, it's "probably not needed". And then later says: "if you pulish a major version, or a version with missing or wrong dependency constraints then retraction ...".

Should we start by stating the positive case, and focusing on the "too-loose constraints" problem.
(Missing dependency constraints is actually the extreme variant of that).

mit-mit: I like that idea. What would that edit look like?

mit-mit: Actually, I need this PR to land as the CVE I'm about to publish, so can we handle that re-write in a follow-up PR?

jonasfj: Filed in https://github.com/dart-lang/site-www/pull/3707

PR #3682, opened 2021-11-21
Modify clarification and minor cleanup for pub outdated page

### LINKED ISSUE #3115: 'dart pub outdated' page issue
Page URL: https://dart.dev/tools/pub/cmd/pub-outdated.html
Page source: https://github.com/dart-lang/site-www/tree/master/src/tools/pub/cmd/pub-outdated.md

**Description of issue:**

The output shown In the very first run of `dart pub outdated` command:

image

As you can see here, `path` is a direct dependency but later on in the article it says 

> The `path` and `meta` packages are still out-of-date, due to reasons that this package doesn’t control. 

If `path` is not a direct dependency but rather a transitive, it should be printed in `transitive dependencies` tab. 

**Solution:**

We need to update the screenshot of terminal.

parlough:
Closes #3115


kwalrath (approved):
LGTM!

I haven't confirmed the output, but it looks like maybe you should also add a `:` after `transitive dependencies` in the [Output columns section](https://dart-dev--pr3682-misc-pub-outdated-cl-ffctj5ls.web.app/tools/pub/cmd/pub-outdated#output-columns).

kwalrath:
> LGTM!
> 
> I haven't confirmed the output, but it looks like maybe you should also add a `:` after `transitive dependencies` in the [Output columns section](https://dart-dev--pr3682-misc-pub-outdated-cl-ffctj5ls.web.app/tools/pub/cmd/pub-outdated#output-columns).

And maybe also in the [Example section](https://dart-dev--pr3682-misc-pub-outdated-cl-ffctj5ls.web.app/tools/pub/cmd/pub-outdated#example)

parlough:
I missed these as they are in separate HTML files. I'll update them now, thanks!

PR #3695, opened 2021-12-01
add userId example to DO capitalize acronyms

lukepighetti:
I added an example to `DO capitalize acronyms and abbreviations longer than two letters like words.` that explicitly calls out the **GOOD** `userId`  **BAD** `userID` example, even though it is alluded to in the description above 

> On the other hand, two-letter *abbreviations* like
> ID (identification) are still capitalized like regular words: `Id`.


parlough (approved):
The next line does showcase that the same rules apply for the id abbreviation already, but still seems like a good idea to me to specifically showcase a multi-word example too, thanks!

kwalrath:
I like it! Thanks, @lukepighetti!

PR #3696, opened 2021-12-03
New page: Concurrency in Dart

kwalrath:
~~Still needs some finishing work, detailed at the bottom of the doc.~~

staged: https://kw-dartlang-3.web.app/guides/language/concurrency


src/_guides/language/concurrency/index.md
```diff
@@ -0,0 +1,403 @@
...
+[json]: https://docs.flutter.dev/cookbook/networking/background-parsing)
+ [Flutter `compute()` function]: https://docs.flutter.dev/cookbook/networking/background-parsing#4-move-this-work-to-a-separate-isolate
+
+![PENDING: alt text here](/guides/language/concurrency/images/isolate-bg-worker.png)
+
+A background worker can perform I/O
+(reading and writing files, for example), set timers, and more.
+It has its own memory (including its own copy of code) and
+doesn’t share any state with the main isolate.
+
```
mit-mit, line 226: Could we add sample code here for this "single-task background worker" use case ?

I find it confusing to read the sample code further down on the page as it follows a diagram that shows a worker isolate that is not single-task. I think the docs would be easier to read with a structure like this:

```
### Using a single-task background worker

Intro text.

Diagram.

Sample-code using Isolate.exit().

### Using a multi-task background worker

Intro text.

Diagram.

Sample-code using Isolate.send()

```

src/_guides/language/concurrency/index.md
```diff
@@ -0,0 +1,403 @@
...
+  }
+}
+
+</style>
+
+Dart supports concurrent programming with async-await, isolates, and
+classes such as `Future` and `Stream`.
+Because Dart code executes in a predictable sequence that
+can’t be interrupted by other Dart code,
+sometimes Dart is called a _single-threaded language._
```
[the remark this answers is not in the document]
kwalrath: It's wording we've used before, but I could move it to be later. (I think we might want it in the page, for SEO reasons, but it doesn't have to be at the top.)

src/_guides/language/concurrency/index.md
```diff
@@ -0,0 +1,403 @@
...
+  final fileData = [!await!] _readFileAsync();
+  final jsonData = jsonDecode(fileData);
+
+  // Use that data.
+  print('Number of JSON keys: ${jsonData.length}');
+}
+
+[!Future<String>!] _readFileAsync() [!async!] {
+  final file = File(filename);
+  return [!await!] file.[!readAsString()!];
```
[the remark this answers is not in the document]
kwalrath: @johnpryan what do you say?

src/_guides/language/concurrency/index.md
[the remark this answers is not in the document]
mit-mit: I don't understand what that wording means! I find it very misleading.

Why have we used that wording before? What are we trying to say with it?

src/_guides/language/concurrency/index.md
```diff
@@ -0,0 +1,403 @@
...
+* ***[PENDING: link to example 5 in Isolate snippets][]***,
+  which shows how to spawn a long-running isolate that
+  receives and sends multiple times.
+
+-------------------
+
+## NOTES/QUESTIONS/TODOs (TO BE DELETED)
+
+TODO:
+* Ask John to publish examples 4 & 5 in dart-lang/samples; link to them
```
johnpryan, line 382: https://github.com/dart-lang/samples/pull/128

src/_guides/language/concurrency/index.md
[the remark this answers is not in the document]
timsneath: I'm a bit confused by this too. Rightly or wrongly, developers typically think of "single-threaded" as meaning "one thing at a time". 

My understanding is that isolates enable Dart code to run across multiple threads (that are hosted in different processes for isolation). 

I've seen this cause a _lot_ of confusion (e.g. https://stackoverflow.com/questions/53019756/dart-is-single-threaded-but-why-it-uses-future-objects-and-perform-asynchronous, which contains a variety of answers that are contradictory). It might be better to not introduce the phrase here, and instead have an question in the FAQ ("is Dart single-threaded?") that gives us the space to address lingering confusion head on.

src/_guides/language/concurrency/index.md
```diff
@@ -0,0 +1,403 @@
...
+## How isolates work
+
+Most modern devices have multi-core CPUs.
+To take advantage of all those cores,
+developers sometimes use shared-memory threads running concurrently.
+However, shared-state concurrency is
+[error prone](https://en.wikipedia.org/wiki/Deadlock) and
+can lead to complicated code.
+
+Instead of threads, all Dart code runs inside of isolates.
```
timsneath, line 145: It would be good to describe this in terms of existing OS concepts. 

For example, are isolates in the same process? Or different processes? Or is that a per-architecture implementation decision? Do they communicate via IPC, or shared memory, or something else?

src/_guides/language/concurrency/index.md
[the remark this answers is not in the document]
johnpryan: After re-reading  @munificent's [explanation](https://groups.google.com/a/dartlang.org/g/misc/c/A9eOQoLsRQM/m/8GcUQMiJAwAJ), I think it's more correct to say that in any given Dart isolate there's a single thread of execution at any given time. Dart doesn't run in parallel unless additional isolates are used.

src/_guides/language/concurrency/index.md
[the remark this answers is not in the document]
johnpryan: Technically, no. But it illustrates async / await. We could maybe make this more clear by adding some additional processing:

```dart
Future<String> _readFileAsync() async {
  final file = File(filename);
  final contents = await file.readAsString();
  return contents.trim();
}
```

src/_guides/language/concurrency/index.md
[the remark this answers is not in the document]
aam: One could read it as saying that Dart language has no concepts of semaphores, mutexes or other tools to manage concurrent data access(that other languages like Java, C++, Python, etc provide). When writing Dart you don't have to worry about data being changed underneath you by some other Dart code that runs concurrently with you.  So that result of `bankAccount.balance += 20` will always be predictable. Perhaps having such tools available in a language makes the language "multi-threaded".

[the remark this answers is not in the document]
mit-mit: Dart *will* utilize multiple cores even if just running a single Dart isolate. For example things like GC and async IO will utilize other cores. I got this confirmed by Slava today, and i think its aligned with what Bob says in that link also.

Dart really isn't "single treaded".

src/_guides/language/concurrency/index.md
```diff
@@ -0,0 +1,403 @@
...
+  The `await` keyword works only in functions that
+  have `async` before the function body.
+{{site.alert.end}}
+
+As the following figure shows,
+the Dart code pauses while `readAsString()` executes non-Dart code,
+in either the Dart virtual machine (VM) or the operating system (OS).
+Once `readAsString()` returns a value, Dart code execution resumes.
+
+![Flowchart-like figure showing app code executing from start to exit, waiting for native I/O in between](/guides/language/concurrency/images/basics-await.png)
```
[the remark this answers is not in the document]
kwalrath: It was intended to be for the async version. It's before the event handling discussion, so it doesn't get into detail about exactly when/where the Dart code is running, just the order that it runs in.

We _could_ build on it later, in the main-isolate-only section.

src/_guides/language/concurrency/index.md
```diff
@@ -0,0 +1,403 @@
...
+
+// Spawns an isolate and waits for the first message
+Future<Map<String, dynamic>> _spawnIsolate() async {
+  final p = ReceivePort();
+  await Isolate.spawn(_readAndParseJson, p.sendPort);
+  return await p.first;
+}
+```
+
+The `_spawnIsolate()` function contains the code that
```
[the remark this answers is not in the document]
kwalrath: /cc @johnpryan

src/_guides/language/concurrency/index.md
```diff
@@ -0,0 +1,403 @@
...
+your Dart Native app needs any of the following:
+
+* A background worker like what `compute()` supports,
+  but that might not be running in a Flutter app
+* A background worker that can
+  handle multiple requests or send multiple replies,
+  as shown in the following figure
+* Some other custom isolate behavior
+
+![PENDING: alt text here](/guides/language/concurrency/images/isolate-custom-bg-worker.png)
```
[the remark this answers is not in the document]
kwalrath: I've been concentrating on _message passing_ rather than ports. For me, the whole port thing is a little harder to explain than the idea of message passing.

src/_guides/language/concurrency/index.md
```diff
@@ -0,0 +1,403 @@
...
+
+The `_spawnIsolate()` function contains the code that
+_spawns_ (creates and starts) the isolate for the background worker,
+and then returns the result:
+
+1. Before spawning the isolate, the code creates a [`ReceivePort`][],
+   which enables the background worker
+   to send messages to the main isolate.
+
+2. Next is the call to `Isolate.spawn()`,
```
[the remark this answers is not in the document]
kwalrath: Probably a good idea, but I might not get to it in the first published version.

src/_guides/language/concurrency/index.md
```diff
@@ -0,0 +1,403 @@
...
+If one of the objects is, say, a `Socket`, then
+the send fails because sockets are unsendable.
+
+For information on the kinds of objects that you can send in messages,
+see the API reference documentation for the [`send()` method][].
+
+[`send()` method]: https://api.dart.dev/stable/dart-isolate/SendPort/send.html
+
+{{site.alert.tip}}
+  The `send()` method can be slow.
```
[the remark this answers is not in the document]
kwalrath: I don't actually know the answer to this! Which is it?

src/_guides/language/concurrency/index.md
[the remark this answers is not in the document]
timsneath: To @johnpryan's comment: "Dart doesn't run in parallel unless additional isolates are used", couldn't you make an equal argument that "C code doesn't run in parallel unless additional forks are spawned"? That is, the language is only single threaded if you don't use threading features :) 

Maybe the confusion is that some of us are thinking of the phrase "single-threaded" as a positive (you don't have to worry about things changing under you, code executes in a predictable sequence). The reason I'm sensitive to the phrase is that a lot of folk mistakenly think of it as indicating a deficit in Dart today...

src/_guides/language/concurrency/index.md
kwalrath: I don't know the answers to all of these questions. Maybe in v2?

src/_guides/language/concurrency/index.md
```diff
@@ -0,0 +1,403 @@
...
+
+A background worker can perform I/O
+(reading and writing files, for example), set timers, and more.
+It has its own memory (including its own copy of code) and
+doesn’t share any state with the main isolate.
+
+
+### Using the Isolate API
+
+Sometimes you might need to use the `Isolate` API directly.
```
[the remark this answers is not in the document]
kwalrath: The Flutter `compute()` function is mentioned just above, though. And I don't want to encourage people to use the Isolate API if `compute()` will do the job. I'll think about this.

src/_guides/language/concurrency/index.md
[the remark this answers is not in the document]
aam: > Why have we used that wording before? What are we trying to say with it?

also page 9 of Gilad's and Lars's [presentation at Goto conference](https://gotocon.com/dl/goto-aarhus-2011/slides/GiladBracha_and_LarsBak_OpeningKeynoteDartANewProgrammingLanguageForStructuredWebProgramming.pdf) might give some context ```SO WHAT IS DART?
...
• Single-threaded
...
```

File: src/_guides/language/concurrency/index.md (hunk 8 above)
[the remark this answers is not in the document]
Comment (kwalrath) [reply src/_guides/language/concurrency/index.md:241]: I had a little something about messages just after the figure; I'm moving it up before the figure. We might want to add "message" to the Request & Reply arrows. Or add "request" and "reply" to the text in the custom background worker.

File: src/_guides/language/concurrency/index.md (hunk 5 above)
Comment (timsneath) [reply src/_guides/language/concurrency/index.md:148]: @mit-mit Is there someone in Aarhus who could draft a couple of sentences here while we're sleeping? :)

File: src/_guides/language/concurrency/index.md (hunk 3 above)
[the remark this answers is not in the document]
Comment (kwalrath) [reply src/_guides/language/concurrency/index.md:110]: That also fixes the problem that `fileData` (which I discuss in the next paragraph) had disappeared. :)

File: src/_guides/language/concurrency/index.md [hunk 12]
```diff
@@ -360,6 +362,20 @@ The following figure summarizes the communication between
 the main isolate and the second isolate:
 
 ![PENDING: alt text here](/guides/language/concurrency/images/isolate-api.png)
+<br>
+***[PENDING: Background worker -> Worker isolate]***
+
+{{site.alert.info}}
+  **API note:**
+  When an isolate calls [`Isolate.spawn()`][],
+  it sets up the new isolate with a copy of the calling isolate’s code.
```
aam, line 371: @kwalrath did you by any chance see my Friday's email reply to your question regarding this?

src/_guides/language/concurrency/index.md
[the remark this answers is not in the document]
aam: When used between isolates in different isolate groups(spawned using `Isolate.spawnUri`), both sender and receiver are doing similar O(number-of-objects-being-sent) work: serializing done by the sender and deserializing done by the receiver.

When used between isolates in the same isolate group(spawned using `Isolate.spawn`), the sender (normally) does more work. The sender validates the message being sent and copies mutable objects of the message for the receiver's consumption. Heap is shared between isolates, so the receiver just picks up a reference to the prepared copy that the sender give to the receiver. Only further processing that the receiver has to do is to canonicalize/recalculate hashes for certain objects. If no canonicalization is needed, receiver will do no additional work.
Sending immutable(or "shareable") data is done by passing a reference, so no copy is needed and is very fast.

Use of `Isolate.exit(port, object)` avoids creating a copy too because it is guaranteed that the sender exits and does not mutate sent data. Because of that `Isolate.exit` is faster for sender than regular `send`. Sender still has to validate that only allowed objects are being sent, so it still does O(number-of-objects) work.

`TransferableTypedData` is similar to regular `send`, in that its sender who pays higher price by creating Transferable wrapper around unstructured bytes. There is no object traversal happening though, it is just memory copy, so it is faster than sending structured objects. Receiver gets data in constant time too.

src/_guides/language/concurrency/index.md
kwalrath: I was just about to get to it!

kwalrath: I've just updated the doc, and I hope this version will stage successfully. ptal!

src/_guides/language/concurrency/index.md
```diff
@@ -0,0 +1,456 @@
...
+  * Add this page to the sidenav. Link to it appropriately.
+  * Figure out how to save an editable version of the source that has the right fonts. (The SVG files don't like the custom fonts; otherwise, I would've used SVGs instead of PNGs.)
+* Maybe:
+  * Add the following text somewhere in this page (or in the FAQ):
+    * Sometimes Dart is called a _single-threaded language._
+    * Dart code executes in a predictable sequence that can’t be interrupted by other Dart code.
+  * Add a figure up high in the doc? (if so, what?)
+
+Qs:
+* The figures are by the index file, under images. Should we instead use the asset system? E.g., see the source for the following figure:
```
parlough, line 453: I think this is fine for now. I believe the rearch integration and related updates will remove support for the `{% asset` system.

kwalrath: Good to know, thanks!

kwalrath:
ptal @mit-mit, @johnpryan, and anyone else who's interested. New staged location:

https://kw-dartlang-3.web.app/guides/language/concurrency#

src/_guides/language/concurrency/index.md
mit-mit: I discussed this a bit, but the VM folks are hesitant to be too explicit as the VM use of processes and threads is considered a implementation detail, and subject to change. I'm still waiting for confirmation, but I think something like this might be acceptable?

```
Within an app, all Dart code runs in an isolate, conceptually similar to a process. Each Dart isolate has a single thread of execution and shares no memory with other isolates. Communication between isolates happens via message passing.
```

src/_guides/language/concurrency/index.md
```diff
@@ -0,0 +1,456 @@
...
+[asynchronous programming codelab]: /codelabs/async-await
+
+
+## How isolates work
+
+Most modern devices have multi-core CPUs.
+To take advantage of all those cores,
+developers sometimes use shared-memory threads running concurrently.
+However, shared-state concurrency is
+[error prone](https://en.wikipedia.org/wiki/Deadlock) and
```
mit-mit, line 147: I was told in a review comment of the block post that this wikipedia link is misleading. The main problem isn't deadlocks, but data races. I'm still waiting for confirmation, but I think this link might be better:
https://en.wikipedia.org/wiki/Race_condition#In_software

src/_guides/language/concurrency/index.md
```diff
@@ -0,0 +1,456 @@
...
+## Code examples
+
+This section discusses some examples
+that use the `Isolate` API
+to implement isolates.
+
+{{site.alert.tip}}
+  If you're using Flutter on a non-web platform,
+  then instead of using the `Isolate` API directly,
+  consider using the [Flutter `compute()` function][].
```
mit-mit, line 269: Can we extend this with something like ", which is a convenience function for moving a single function call to a worker isolate" ?

src/_guides/language/concurrency/index.md
timsneath: I would **love it** if we could say something like that explicitly ("the VM use of processes and threads is considered a implementation detail, and subject to change"). Because it's hard to take the line that Dart doesn't use threads ("instead of threads") -- since of course the unit of execution in any modern OS is indeed a thread.

src/_guides/language/concurrency/index.md
```diff
@@ -0,0 +1,456 @@
...
+  final jsonData = jsonDecode(fileData);
+  Isolate.exit(p, jsonData);
+}
+```
+
+The relevant statement is the last one, which exits the isolate,
+sending `jsonData` to the passed-in `SendPort`.
+Message passing between isolates normally involves data copying,
+and thus can be slow and increases with the size of the message —
+O(n) in big O notation.
```
mit-mit, line 350: Link to https://en.wikipedia.org/wiki/Big_O_notation

src/_guides/language/concurrency/index.md
```diff
@@ -0,0 +1,456 @@
...
+When an isolate calls [`Isolate.spawn()`][],
+the two isolates have the same executable code
+and are in the same _isolate group_.
+Isolate groups enable performance optimizations such as sharing code;
+a new isolate immediately runs the code owned by the isolate group.
+Also, `Isolate.exit()` works only when the isolates
+are in the same isolate group.
+
+In some special cases —
+for example, when implementing tools like `dart pub run` —
```
mit-mit, line 414: I'm not sure the reference to "implementing tools like dart pub run" is all that helpful here. Could we just say something like "for example, when you need to spawn a new isolate from new source code read from a Uri"?

src/_guides/language/concurrency/index.md
parlough: I wonder is it even useful to talk about Big O notation here? If someone doesn't know it, I'm not sure they're likely to go read a Wikipedia article to understand it. Perhaps a slightly expanded explanation of how the run time scales linearly with the message size would be more useful. Similar for the reference to constant time below.

src/_guides/language/concurrency/index.md
johnpryan: This has landed now

src/_guides/language/concurrency/index.md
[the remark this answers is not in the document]
johnpryan: Maybe we could say each isolate has a single call stack, heap, and event loop?

src/_guides/language/concurrency/index.md
aam: > shares no memory with other isolates

This might be misleading. Isolates do share immutable data(strings, constants, etc) and code with isolates in the same isolate group: the point of lightweight isolates is that they share the same VM heap. Isolates do not share _mutable_ data (as in - changes to the data done by one isolate is not visible to another isolate).

src/_guides/language/concurrency/index.md
kwalrath: Speaking as someone who knew about big O notation but doesn't use it often, the page was helpful to me. :) It might be good to go into more detail on this page (or in a future page about performance), but for now big O is a handy shorthand.

src/_guides/language/concurrency/index.md
kwalrath: That explanation is already in the next sentence, so I think I'll just delete the example.

kwalrath:
Maybe ready to go? I'll take another look tomorrow.

I've updated the staged version: https://kw-dartlang-3.web.app/guides/language/concurrency

src/_guides/language/concurrency/index.md
```diff
@@ -0,0 +1,446 @@
...
+
+Why does it matter whether a method is synchronous or asynchronous?
+It matters because most apps need to do more than one thing at a time.
+For example, an app might start an HTTP request,
+but need to update its display or respond to user input
+before the HTTP request completes.
+Asynchronous code helps apps stay responsive.
+
+### The async-await syntax
+The `async` and `await` keywords provide a declarative way to define asynchronous functions and use their results.
```
parlough, line 80:
```suggestion
### The async-await syntax

The `async` and `await` keywords provide
a declarative way to define asynchronous functions
and use their results.
```

src/_guides/language/concurrency/index.md
```diff
@@ -0,0 +1,449 @@
...
+
+{{site.alert.tip}}
+  The `send()` method can be slow.
+  When possible,
+  avoid using `send()` to send large amounts of data.
+  Instead, send the data when the worker isolate exits,
+  as described in the previous section.
+{{site.alert.end}}
+
+## Performance and isolate groups
```
parlough, line 418: For future work, I'm not sure the best way to improve it yet, but I feel the discussion around performance and isolate groups could maybe be better consolidated. Or perhaps isolate groups and their implications could be introduced earlier instead of tacked on in the end. It seems the reader may need to reference back to previous sections to understand this section and modify their previously gained understanding. 

For example, there is talk about the performance characteristics of `Isolate.exit()` earlier, with these clarifications added later on, so it seems the discussion around that is a bit separated.

parlough (approved):
I think this looks in a pretty good place as an initial release, nice job! It's nice to see some fleshed out documentation around these concepts.

src/_guides/language/concurrency/index.md
timsneath: OK, so then, this should work? 

> Within an app, all Dart code runs in an isolate, conceptually similar to a process. Each Dart isolate has a single thread of execution and shares no mutable data with other isolates. Communication between isolates happens via message passing. Naturally, Dart's isolate model is built with underlying primitives provided by the operating system, but the Dart VM's use of processes and threads is considered an implementation detail.

src/_guides/language/concurrency/index.md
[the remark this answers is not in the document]
mit-mit: OK, I think the current text works

src/_guides/language/concurrency/index.md
mit-mit: I discussed with Slava, and he's fine with this text:

> Within an app, all Dart code runs in an isolate. Each Dart isolate has a single thread of execution and shares no mutable data with other isolates. Communication between isolates happens via message passing. Naturally, Dart's isolate model is built with underlying primitives provided by the operating system, but the Dart VM's use of processes and threads is considered an implementation detail.

This leaves out the "conceptually similar to a process" which he found potentially very misleading: they are like processes in some ways but very different in others, and processes are typically very expensive, where isolates are light.

kwalrath: How about this (_italicized changes_ are to clarify and to avoid via, naturally, some passive voice, and other writing anathemas):

> Within an app, all Dart code runs in an isolate. Each Dart isolate has a single thread of execution and shares no mutable objects with other isolates. _Instead of using shared memory, isolates communicate by passing messages._ Dart’s isolate model is built with underlying primitives, such as processes and threads, that the operating system provides; the Dart VM’s use of these primitives is an implementation detail _that this page doesn’t discuss._

kwalrath: I've pushed this change and restaged it:

https://kw-dartlang-3.web.app/guides/language/concurrency

src/_guides/language/concurrency/index.md
```diff
@@ -0,0 +1,454 @@
...
+a new isolate immediately runs the code owned by the isolate group.
+Also, `Isolate.exit()` works only when the isolates
+are in the same isolate group.
+
+In some special cases,
+you might need to use [`Isolate.spawnUri()`][],
+which sets up the new isolate with a copy of the code
+that's at the specified URI.
+However, `spawnUri()` is much slower than `spawn()`,
+and the new isolate isn't in its spawner's isolate group.
```
mkustermann, line 438: Might be worth mentioning that `Isolate.spawnUri` is not supported on Flutter.

src/_guides/language/concurrency/index.md
```diff
@@ -0,0 +1,454 @@
...
+  receives and sends multiple times.
+
+{% assign samples = "https://github.com/dart-lang/samples/tree/master/isolates" %}
+
+[isolate samples]: {{ samples }}
+[send_and_receive.dart]: {{ samples }}/bin/send_and_receive.dart
+[long_running_isolate.dart]: {{ samples }}/bin/long_running_isolate.dart
+
+{{site.alert.tip}}
+  The `send()` method can be slow.
```
mkustermann, line 416: It's one of the main APIs that's used to communicate between isolates (it also got much faster recently). The performance of `Isolate.exit()` and `SendPort.send()` isn't that much different actually.

=> I would really not discourage users by saying `can be slow. avoid when possible`.

Instead I'd focus on the actual differences between the two:

Using `SendPort.send()` will transitively copy the message being sent (except for immutable objects such as strings which are shared). The receiver will receive this copy. To avoid such a memory copy one can use `Isolate.exit()` which will hand the existing object graph to the receiver without copying it, while exiting the current isolate.

mkustermann (commented):
In general looks good.

src/_guides/language/concurrency/index.md
kwalrath: After further discussion out of band, I'll delete "Instead of using shared memory."

kwalrath: New change is up and restaged

mit-mit: https://github.com/dart-lang/site-www/pull/3696/commits/09b735f2aa773e26c074572cf32212959666e732 LGTM, resolving this thread!

src/_guides/language/concurrency/index.md
mit-mit: I learned earlier today that Isolate.send is fast for isolates in a group.

Should we perhaps remove the "`send()` method can be slow." part from here, and then in the section below for Isolate.spawnUri add something like: "Also note that Isolate.send() is slower between isolates not in an isolate group" ?

src/_guides/language/concurrency/index.md
kwalrath: I had no idea! Will do.

src/_guides/language/concurrency/index.md
```diff
@@ -0,0 +1,454 @@
...
+  final file = File(filename);
+  final contents = file.readAsStringSync();
+  return contents.trim();
+}
+```
+
+Here’s similar code, but with changes (highlighted) to make it asynchronous:
+
+{% prettify dart tag=pre+code %}
+[!Future<void>!] main() [!async!] {
```
johnpryan, line 112: This can stay `void`, since no one is going to call `main`: https://github.com/dart-lang/samples/pull/128#discussion_r764096958

```suggestion
void main() [!async!] {
```

src/_guides/language/concurrency/index.md
```diff
@@ -0,0 +1,454 @@
...
+
+[`Isolate.exit()`]: {{site.dart_api}}/{{site.data.pkg-vers.SDK.channel}}/dart-isolate/Isolate/exit.html
+[`Isolate.spawn()`]: {{site.dart_api}}/{{site.data.pkg-vers.SDK.channel}}/dart-isolate/Isolate/spawn.html
+[`ReceivePort`]: {{site.dart_api}}/{{site.data.pkg-vers.SDK.channel}}/dart-isolate/ReceivePort-class.html
+[`SendPort`]: {{site.dart_api}}/{{site.data.pkg-vers.SDK.channel}}/dart-isolate/SendPort-class.html
+
+Here’s the code for the main isolate:
+
+```dart
+Future<void> main() async {
```
johnpryan, line 314:
```suggestion
void main() async {
```

src/_guides/language/concurrency/index.md
kwalrath: Wouldn't sending any message between isolate groups be slower? (No need to single out send(), is there?)

kwalrath: Of course, Isolate.exit() isn't supported between different isolate groups, so it's kind of equivalent.

johnpryan (approved):

src/_guides/language/concurrency/index.md
aam: Sending `TransferableTypedData` will be similarly fast regardless of whether isolates are in the same group or in different groups.

src/_guides/language/concurrency/index.md
mit-mit: Ported to https://github.com/dart-lang/site-www/pull/3711

src/_guides/language/concurrency/index.md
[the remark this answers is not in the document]
kwalrath: Noted in followup issue (#3720)

src/_guides/language/concurrency/index.md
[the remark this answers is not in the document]
kwalrath: Included in followup issue (#3720)

kwalrath:
> If you have a chance, creating an issue with some of the follow-up work you want done would be helpful. A comment may be hard to track.

I've created a followup issue: #3720

PR #3700, opened 2021-12-07
Standardize on call sites instead of callsites

parlough:
I noticed we were not consistent about which spelling. I went with "call sites" to be consistent with Wikipedia.

_Created this to test build, but still a fine change._


kwalrath (approved):

kwalrath:
Just confirmed that this change made it out (phew!)... but we have 2 more occurrences of _callsite_ in https://dart.dev/guides/language/effective-dart/design.

PR #3707, opened 2021-12-08
Attempt at refactoring retraction example

jonasfj:
This attempts to put the positive case before the negative case, in example of when to do and not to do package retraction.

_Staged: https://kw-www-dartlang-1.web.app/tools/pub/publishing#retract_

I'm not sure it's better... it's also possible that we should just write out a few examples in expandable `<details><summary>` fields... or possible we should write a separate page explaining example of when we think retraction is a good idea.
I'm not sure it's useful, ultimately it's up to the package author anyways, all we make here are guidelines -- if you've shipped a sufficiently bad bug, well then maybe you do want to pull out all of the tools at your disposal such as retraction.

/cc @mit-mit, @szakarias, @kevmoo following [discussion here](https://github.com/dart-lang/site-www/pull/3680#discussion_r764381584).

### DIFF
--- a/src/tools/pub/publishing.md
+++ b/src/tools/pub/publishing.md
@@ -343,10 +343,19 @@ Before retracting a package,
 consider publishing a new version instead.
 Retracting a package causes churn and can have a negative impact on package users.
 
-For example, if your package version contains a minor bug, then retraction is
-probably not needed. On the other hand, if you accidentally publish a major
-version or a version with wrong or missing dependency constraints,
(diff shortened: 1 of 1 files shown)


kevmoo (approved):
👏 👏 👏 👏 👏

kwalrath:
My changes might look large, but the visible effect is fairly small:

image

kwalrath (approved):

parlough (approved):

PR #3711, opened 2021-12-08
Tweak main example in concurrency

mit-mit:
Based on comment from here: https://github.com/dart-lang/samples/pull/128#discussion_r764096958

cc @johnpryan


kwalrath (approved):
I thought I'd already made this change :(

PR #3716, opened 2021-12-09
Document ignoring all linter rules

### LINKED ISSUE #3576: Static analysis page should be updated for new `ignore: type=lint` feature
### Page URL

https://dart.dev/guides/language/analysis-options#suppressing-rules-for-a-file

### Page source

_No response_

### Describe the problem

Recently (as of last week, post-2.14), you can ignore static analysis reports inline in a new way:

`// ignore_for_file: type=lint` (or `// ignore: type=lint`)

This (first one) will suppress _all lint_ in a file (library?). So exciting! I don't know of why anyone would do the latter example, suppressing a single line; the feature was designed for the former example.

In any case, I think this should be documented.

### Expected fix

Documentation! A single sentence, like what exists for the other examples of inline ignores, is all I would expect.

### Additional context

_No response_

parlough:
Closes #3576


kwalrath (approved):
LGTM!

src/_guides/language/analysis-options.md
```diff
@@ -407,6 +419,18 @@ To suppress more than one rule, supply a comma-separated list:
 const x = y;
 ```
 
+To suppress all linter rules, add a `type=lint` specifier:
```
srawlins, line 422: I personally would probably not document in this position; The `type=lint` support use case was really just for `// ignore_for_file`, but of course it does work with `// ignore:` for consistency. I think this might just be unnecessary text here. Just my 2c.

srawlins (approved):
Thanks!

PR #3726, opened 2021-12-14
Deploy to live using GitHub Action

mit-mit:
Stop using `deploy.sh` as that requires a token we'd rather not have.

This uses the same GitHub Action we use for deploying to Firebase Preview for PRs.

Based on:
https://github.com/FirebaseExtended/action-hosting-deploy/blob/main/README.md#deploy-to-your-live-channel-on-merge

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -121,9 +121,12 @@ jobs:
         run: dart pub get
       - run: ./tool/shared/write-ci-info.sh -v
       - run: ./tool/build.sh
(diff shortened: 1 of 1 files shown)


parlough (approved):
Looks good to me according to the documentation. Will verify after merged as well.

May want to wait for the infrastructure updates unless you want to test this deploy strategy soon. Up to you.

kwalrath:
Should we file an issue to remove deploy.sh?

PR #3731, opened 2021-12-14
2.15 change summaries

kwalrath:
Creating a PR even though we're frozen so that @mit-mit can review the contents. (I can create a new PR later if necessary.)

### DIFF
--- a/src/_guides/language/evolution.md
+++ b/src/_guides/language/evolution.md
@@ -220,8 +220,7 @@ late S Function<S extends T Function<T>(T)>(S) f;
 
 Dart 2.15 improved support for function pointers,
 known as _tear-offs._
-In particular, you can now pass constructors as arguments.
-
+In particular, constructor tear-offs are now supported.
 For details, see the [Dart 2.15 announcement][].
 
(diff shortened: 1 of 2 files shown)


kwalrath:
On second thought, I'm not sure the isolate improvements belong in the language evolution doc. What do you think, @mit-mit?

parlough:
That's tough, but taking a look at the previous entries, they only include language features so isolates may be a bit out of place.

src/_guides/language/evolution.md
```diff
@@ -216,6 +216,16 @@ var callback = [<T>(T value) => value];
 late S Function<S extends T Function<T>(T)>(S) f;
 ```
 
+### Dart 2.15
+
+Dart 2.15 improved support for function pointers,
+known as _tear-offs._
+In particular, you can now pass constructors as arguments.
```
mit-mit, line 223: I suggest "In particular, constructor tear-offs are now supported."

mit-mit (changes requested):

mit-mit (approved):

PR #3739, opened 2021-12-28
Fix concurrency according to `avoid_void_async`

AlexV525:
*Holding since the pull requests window is closed.*

According to the [`avoid_void_async` rule](https://dart-lang.github.io/linter/lints/avoid_void_async.html), we might expect the function to be written as `Future<void>`.

### DIFF
--- a/src/_guides/language/concurrency/index.md
+++ b/src/_guides/language/concurrency/index.md
@@ -109,7 +109,7 @@ String _readFileSync() {
 Here’s similar code, but with changes (highlighted) to make it asynchronous:
 
 {% prettify dart tag=pre+code %}
-void main() [!async!] {
+[!Future<void>!] main() [!async!] {
   // Read some data.
   final fileData = [!await!] _readFileAsync();
   final jsonData = jsonDecode(fileData);
@@ -312,7 +312,7 @@ This example uses the following isolate-related API:
 Here’s the code for the main isolate:
 
 ```dart
-void main() async {
+Future<void> main() async {
   // Read some data.
   final jsonData = await _parseInBackground();
 


parlough:
Thanks for opening the PR. This was changed to `void` due to https://github.com/dart-lang/samples/pull/128#discussion_r764096958 

I'm a bit torn as having the `Future` does indicate someone may see the result of the `main` function as significant. This makes sense to avoid but consistency is nice as well, especially as someone can technically call `main` like any other function.

Currently the docs are rather inconsistent, with some declarations using `Future<void>` and others just using `void`. I'm thinking for brevity and the fact developers generally shouldn't call the `main` method anyway, it may make sense to standardize on just `void` for at least the `main` method. I need to give it some more thought though.

---

There also has been [some discourse](https://github.com/dart-lang/linter/issues/2440#issuecomment-775842069) around the current usefulness of this lint as it does make sense for there to be functions which you should "fire and forget" and `void` can be the appropriate return type in those case. I have a feeling that `main` is one of those function as if you _do_ call it, you shouldn't really expect it to have a specific behavior or even return at all.

AlexV525:
I agree with the name-changing part, since `main()` shouldn't be called manually. But let's consider this:

When we have multiple entry-point in the main file, you'll need to write multiple "main" methods for different entry points. This might cause methods to be a mess when developers read them. As for consistency, when we apply this rule, all methods whether it's `main()` or not, should be written in `Future<void>`.

Regarding the `void` function in this article, they should be carried this much information, which makes readers need to consider whether the async-void is reasonable here, so a new name for those methods is better in this scenario.

For me in the overview, I don't want any of my code collaborators to consider the difference to use async-void and and `Future<void>`, and things like `await Future<void>.delayed(duration)` still happen in a high frequency.

parlough:
I appreciate for your thoughts, I agree that the benefits and resulting consistency of using `Future` probably win out. Particularly in this case, for illustrating necessary changes switching back to `Future<void>` is probably a good idea. 

I think this change looks good to me, thanks so much! I'll leave this open for others to comment on though and revisit it once the infrastructure changes have been merged.

_I also appreciate your mention of having a different name to specify if something shouldn't be awaited. It would be nice if we had a separate static way to suggest functions shouldn't be awaited like `@fireAndForget` rather than just a name, but perhaps a mention of naming to better indicate this would be useful somewhere. Perhaps on the linter rule itself, I'll give that some thought!_

kwalrath:
As @parlough implied, we originally had `Future<void>` as the return type for `main()`, but changed it to `void` due to [this review comment](https://github.com/dart-lang/site-www/pull/3696/commits/36680fd2d1484d5549022e2deb12e97c4eb47ad7#r765081064) by @johnpryan, which was presumably based on [the samples feedback](https://github.com/dart-lang/samples/pull/128#pullrequestreview-825421530) from @lrhn.

@munificent do you have any thoughts on `Future(void)` vs. `void` as a return type for functions like `main()` that should never be called?

johnpryan:
@kwalrath I used `void main() async {` is instead of `Future<void> main() async {` based off @lrhn's [comment](https://github.com/dart-lang/samples/pull/128#discussion_r764096958). @pq Maybe the `avoid_void_async` lint should ignore `main`?

johnpryan:
@AlexV525 @parlough  I'm guessing that dart.dev doesn't use the `avoid_void_async` lint rule right now because it's not part of the core, recommended, Flutter or pedantic categories in the [linter](https://github.com/dart-lang/linter) package.

AlexV525:
For the `main()` method, I think it's good enough to use `void async`. What if I changed the method's name here?

lrhn:
Personally, I don't mind `void` returning `async` functions. They have their places, and I *think* they happen often enough for me that I wouldn't want the `avoid_async_void` lint for my code. (I haven't actually checked, because I never needed to).

Returning `void` means something *different* from returning `Future<void>`. The latter suggests that you wait for that, the former does not.

Combining forcing a `Future<void>` return type with `unawaited_futures` will make returning a future that *no-one should care about* even more annoying.

I admit it's error-prone to forget returning or awaiting a future that you *are* supposed to wait for, or forgetting to handle errors in a deliberate `void`+`async` function.

sfshaza2 (commented):
This looks good. Why is it in draft?

sfshaza2:
Oh, I see. (I missed the convo at first.) I see the different thoughts on this and the consideration of handling main differently. @mit-mit, can you weigh in on best practice here?

mit-mit:
We discussed this in the language meeting, and we don't want to introduce any further complexity on the signature if `main`. Closing as by-design.

PR #3754, opened 2022-01-10
Add /concurrency redirect

### LINKED ISSUE #2022: Create destination page for isolates
It's time to give isolates a homepage.

I'm thinking of something similar to https://dart.dev/guides/json, at least to start with.

/cc @filiph

parlough:
Closes #2022


kwalrath:
@mit-mit we thought maybe a shortcut for isolates might be in order. What do you think?

kwalrath:
@mit-mit wdyt?

mit-mit:
I like adding a short-link for this, but deep linking that deep into the page seems a bit confusing. Could we just link to https://dart.dev/guides/language/concurrency ?

parlough:
I introduced this due to the lack of a good destination for isolates. This whole page is about more than isolates though. Do you think a /concurrency redirect would be more useful?

mit-mit:
Yeah, that migth be better.

kwalrath (approved):

PR #3758, opened 2022-01-13
Allow robots on release deploy

### LINKED ISSUE #3755: dart.dev/robots.txt excludes all robots
### Describe the problem

https://dart.dev/robots.txt is blocking most of the site content, as if it were a staging site. Looking at dart-lang/site-shared/pull/50, I'm guessing that we've made a change that prevents robots.txt from being overwritten to be more permissive.

@kevmoo found this issue; I don't know how long it's existed.

### Expected fix

_No response_

### Additional context

_No response_

parlough:
I made the change in a new file here on site-www as to not require larger changes in site-shared and then here as well. 

With the infrastructure update, which reworks deploys, we likely won't need this file anymore and can address this in a more consistent fashion to match flutter/website.

Closes #3755

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -121,6 +121,7 @@ jobs:
         run: dart pub get
       - run: ./tool/shared/write-ci-info.sh -v
       - run: ./tool/build.sh
+      - run: ./tool/before-release.sh
       - uses: FirebaseExtended/action-hosting-deploy@v0
         with:
           repoToken: '${{ secrets.GITHUB_TOKEN }}'
--- a/tool/before-release.sh
+++ b/tool/before-release.sh
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -e -o pipefail
+
+source ./tool/shared/env-set-check.sh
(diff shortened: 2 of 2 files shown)


kevmoo:
So I looked at the preview URL and `robots.txt` still seems "wrong".

kwalrath:
LGTM, and I confirmed that the staged version is correctly locked down. @parlough I assume you've tested the code locally and it produces something like https://flutter.dev/robots.txt?

We should have someone else engineer look at this, just in case. (Could be @kevmoo or @mit-mit.)

kwalrath:
> So I looked at the preview URL and `robots.txt` still seems "wrong".

The preview version _should_ be "wrong" — no preview versions should be indexed by search engines.

kevmoo (approved):

kevmoo:
It'd be hard to be more broken than we are now. Let's get this deployed and see?

kwalrath:
> It'd be hard to be more broken than we are now. Let's get this deployed and see?

Famous last words 😆 

But I've merged it anyway, because I think you're right.

PR #3759, opened 2022-01-13
Fix dead null safety DartPad links

parlough:
The https://nullsafety.dartpad.dev website was turned down a few weeks ago so these links lead nowhere.


parlough:
@kwalrath Hmm, not entirely sure why, but sometimes they don't start if GitHub actions is having issues at the time of push. 

I repushed the changes and the actions are running :)

PR #3765, opened 2022-01-14
Rearchitecture

btobin:
- Upgrade Ruby, all Gems
- Refactor html/markdown for removal of `jekyll-assets`
- Create mutli-stage Dockerfile and docker-compose service
- Write a Makefile for local development, testing, building and deployment
- Refactor javascript placement and inclusion
- Refactor css/sass shared, placement, inclusion, build
- Investigate continuing with build when non-zero exit code for usability
- Refactor tool bash scripts, including code refresh, formatting check, link check
- Remove symlinks for unused or unneeded site-shared files in both tool and sass
- Refactor Github action, tasks and how tests are run
- Add staging task for testing
- Remove unnecessary line length formatting for select files
- Remove all unnecessary files e.g. ./deploy/**

https://dash-site-dev--dart-r4q6o37f.web.app

Fixes https://github.com/dart-lang/site-www/issues/2918

PR #3766, opened 2022-01-15
Update Default parameter values information

myronrezzo:
The current text implies that it is possible to define default values for both types of positional parameters (required and optional), as it says: "Your function can use = to define default values for both named and positional parameters.". However, positional parameters can be of 2 types, required and optional, and only the optional ones allow the definition of a default value.

This change is intended to make it clear that only named and optional positional parameters are likely to have default values set.

**IMPORTANT:** Due to year-end holidays and work on the dart.dev infrastructure, this repository is accepting only simple or crucial pull requests.

Instead of creating a pull request, consider creating an issue (https://github.com/dart-lang/site-www/issues/new/choose) about the desired change.

We expect to return to normal PR handling in mid-January.

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -1398,8 +1398,8 @@ assert(say('Bob', 'Howdy', 'smoke signal') ==
 <a id="default-parameters"></a>
 #### Default parameter values
 
-Your function can use `=` to define default values for both named and positional
-parameters. The default values must be compile-time constants.
+Your function can use `=` to define default values for both named and optional
+positional parameters. The default values must be compile-time constants.
 If no default value is provided, the default value is `null`.
 
 Here's an example of setting default values for named parameters:
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -1398,8 +1398,8 @@ assert(say('Bob', 'Howdy', 'smoke signal') ==
 <a id="default-parameters"></a>
 #### Default parameter values
 
-Your function can use `=` to define default values for both named and optional
-positional parameters. The default values must be compile-time constants.
+Your function can use `=` to define default values for optional parameters,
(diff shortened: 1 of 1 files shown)


src/_guides/language/language-tour.md
```diff
@@ -1398,8 +1398,8 @@ assert(say('Bob', 'Howdy', 'smoke signal') ==
 <a id="default-parameters"></a>
 #### Default parameter values
 
-Your function can use `=` to define default values for both named and positional
-parameters. The default values must be compile-time constants.
+Your function can use `=` to define default values for both named and optional
+positional parameters. The default values must be compile-time constants.
```
parlough, line 1402: What are your thoughts on something like the following? Since technically required named arguments [cannot have a default value defined](https://dart.dev/tools/diagnostic-messages#default_value_on_required_parameter) either. Happy to consider other wordings as well. 

```suggestion
Your function can use `=` to define default values for optional parameters,
both named and positional. The default values must be compile-time constants.
```

parlough (commented):
This looks like a great change, thanks @myronrezzo! 

I have one potential suggestion to expand the "optional" specifier to both types of parameters.

As for merging this change, we will likely need to wait until the [infrastructure updates](https://github.com/dart-lang/site-www/pull/3765) are finalized or else this change may be accidentally lost. In the mean time, can you sign the Google CLA as outlined by the failed check: https://github.com/dart-lang/site-www/pull/3766/checks?check_run_id=4828607048

Thanks again!

src/_guides/language/language-tour.md
myronrezzo: Thanks for the complement, @parlough. The description is clear enough now.

parlough (approved):
Thanks again, looks great!

I will merge this despite the upcoming infrastructure changes as it affects lines which won't require changes, so it should allow for a clean merge. If there is any problem, I will make sure this change gets reinstated.

Issue #3760, opened 2022-01-13
[2.16] Update `dartdoc` command documentation to `dart doc`


parlough:
### Page URL

https://dart.dev/tools/dartdoc.html

### Page source

https://github.com/dart-lang/site-www/tree/master/src/tools/dartdoc.md

### Describe the problem

In Dart 2.16, the previous `dartdoc` command will be replaced by `dart doc`, with plans for an eventually removal.

### Expected fix

Switch usages and documentation of `dartdoc` as a command to `dart doc`. Still point out the previous command while it still exists, just like was done with `dartanalyzer` and `dartfmt`.

PR #3778, opened 2022-01-24
Update docs to use dart doc

mit-mit:
Documentation update related to deprecation of `dartdoc` (see https://github.com/dart-lang/sdk/issues/46100 for context)

Closes https://github.com/dart-lang/site-www/issues/3760

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -222,7 +222,7 @@
       { "source": "/tools/dart2native", "destination": "/tools/dart-compile", "type": 301 },
       { "source": "/tools/dartanalyzer", "destination": "/tools/dart-analyze", "type": 301 },
       { "source": "/tools/dartdoc", "destination": "/tools/dart-doc", "type": 301 },
-      { "source": "/tools/dartdocgen{,/**}", "destination": "https://github.com/dart-lang/dartdoc#dartdoc", "type": 301 },
+      { "source": "/tools/dartdocgen{,/**}", "destination": "/tools/dart-doc", "type": 301 },
       { "source": "/tools/dartfmt", "destination": "/tools/dart-format", "type": 301 },
       { "source": "/tools/dartium?(.html)", "destination": "/web/dart-2#tools", "type": 301 },
       { "source": "/tools/debian?(.html)", "destination": "/get-dart", "type": 301 },
--- a/firebase.json
+++ b/firebase.json
@@ -221,6 +221,7 @@
       { "source": "/tools/dart2aot", "destination": "/tools/dart-compile", "type": 301 },
       { "source": "/tools/dart2native", "destination": "/tools/dart-compile", "type": 301 },
       { "source": "/tools/dartanalyzer", "destination": "/tools/dart-analyze", "type": 301 },
+      { "source": "/tools/dartdoc", "destination": "/tools/dart-doc", "type": 301 },
       { "source": "/tools/dartdocgen{,/**}", "destination": "https://github.com/dart-lang/dartdoc#dartdoc", "type": 301 },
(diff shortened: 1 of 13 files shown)


src/tools/dart-doc.md
```diff
@@ -17,36 +18,29 @@ see the [documentation part of Effective Dart][effective doc].
...
 
 By default, the documentation files are static HTML files,
-placed in the `doc/api` directory.
-To view the rendered documentation, you need an HTTP server.
+placed in the `doc/api` directory. You can create the
+files in a different directory with the `--output-dir` flag.
 
-Although `dartdoc` is in the Dart SDK,
-it's also available in the [dartdoc package.][]
-You might use the package when you need a programmatic interface
```
parlough, line 36: Is it desired to deemphasize or not support the programmatic use case? I think it's quite useful to call out how `dart doc` is implemented and indicate its underlying implementation can also be used as a library. 

Perhaps just remove the mention of trying new features?

src/tools/index.md
```diff
@@ -96,7 +96,7 @@ The Dart SDK includes the following general-purpose tools:
   testing, compiling, and running Dart code,
   as well as working with the [pub package manager](/guides/packages).
 
-[`dartdoc`](/tools/dartdoc)
+[`dartdoc`](/tools/dart-doc)
```
parlough, line 99: Do you want to change these and similar mentions to `dart doc` as well?

```suggestion
[`dart doc`](/tools/dart-doc)
```

kwalrath: I suspect that we want to change every single mention of `dartdoc` to `dart doc`, but also (in the canonical discussions) mention that it used to be called `dartdoc`.

parlough:
I've gone ahead and updated the various locations we have previously changed with the other tool updates and updated the remaining user facing mentions I can find.

src/tools/dart-doc.md
mit-mit: I don't think this should be needed except in cases like validating a bug fix (in which case we can mention it in the bug).

When have you found it useful?

parlough: I was originally thinking more about using the `dartdoc` API/library as a dependency. But taking a look at [packages using it as a dependency on pub](https://pub.dev/packages?q=dependency%3Adartdoc) there are very few users of it not as a dev dependency. 

So I suppose it is a rather niche and advanced functionality that likely doesn't need to be mentioned on the page.

firebase.json
```diff
@@ -221,6 +221,7 @@
       { "source": "/tools/dart2aot", "destination": "/tools/dart-compile", "type": 301 },
       { "source": "/tools/dart2native", "destination": "/tools/dart-compile", "type": 301 },
       { "source": "/tools/dartanalyzer", "destination": "/tools/dart-analyze", "type": 301 },
+      { "source": "/tools/dartdoc", "destination": "/tools/dart-doc", "type": 301 },
       { "source": "/tools/dartdocgen{,/**}", "destination": "https://github.com/dart-lang/dartdoc#dartdoc", "type": 301 },
```
parlough, line 225: What do you think about updating this redirect as well? 

```suggestion
      { "source": "/tools/dartdocgen{,/**}", "destination": "/tools/dart-doc", "type": 301 },
```

With the recent changes to the dartdoc README, the old link wouldn't jump to the README anymore. If you still want it to point to the GitHub repo, then you can point it to: https://github.com/dart-lang/dartdoc#dart-documentation-generator

kwalrath:
Marked as a draft just so we don't accidentally merge it

PR #3783, opened 2022-01-27
Remove unnecessary null safety argument on embedded DartPads

parlough:
DartPad no longer supports pre-null safe code, so the `null_safety` parameter was removed and no longer does anything. This PR removes all unnecessary usages of it.


domesticmouse (approved):

kwalrath (approved):
Thanks, @parlough!

PR #3788, opened 2022-01-29
Clarify interpolation guideline doesn't apply to single objects

### LINKED ISSUE #3272: Clarify that string interpolation guideline doesn't apply to single objects
The guideline [PREFER using interpolation to compose strings and values](https://dart.dev/guides/language/effective-dart/usage#prefer-using-interpolation-to-compose-strings-and-values) is intended to only apply when building up a larger string out of multiple pieces. But some misread it and interpret to mean that you also should not call `toString()` when all you need to do is convert a single object to a string.

We should add a sentence clarifying that the latter is allowed.

kkreine:
Closes #3272

### DIFF
--- a/src/_guides/language/effective-dart/usage.md
+++ b/src/_guides/language/effective-dart/usage.md
@@ -460,6 +460,9 @@ it's almost always cleaner and shorter to use interpolation:
 'Hello, ' + name + '! You are ' + (year - birth).toString() + ' y...';
 {% endprettify %}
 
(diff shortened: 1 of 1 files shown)


domesticmouse (approved):

parlough (approved):

munificent (approved):

parlough:
Thanks so much for handling this @kkreine, looks great!

PR #3793, opened 2022-02-01
Documentation of the `platforms` key in `pubspec.yaml`

jonasfj:


src/tools/pub/pubspec.md
```diff
@@ -269,6 +273,65 @@ For more information, see
...
+  Some packages do import `dart:io` and avoid using it when they are compiled to
+  Javascript. This technically works, and but prevents automatic platform
+  detection from correctly determining that the package supports web.
+
+  These packages are encouraged to [conditionally import][] `dart:io` instead of
+  explicitly declaring supported platforms. Using conditional import reduces the
+  risk of runtime errors due to accidental usage of `dart:io` when compiled to
+  Javascript, and it ensures that packages with dependency upon this package
+  doesn't also have to explicitly declare supported platforms.
+{{site.alert.end}}
```
jonasfj, line 329: This note is a bit long... I'm not sure if we can phrase it shorter... if we should omit it?

Or if we should wrap it in a [`<details><summary>`](https://gist.github.com/scmx/eca72d44afee0113ceb0349dd54a84a2) to make it expandable.

The goal is really just to explain one of the common pitfalls, because `dart:io` is weird... compared to `dart:html` which cannot be imported when running on the VM.

jonasfj: I could also move this to a page dedicated to "platform detection". Just because we have such a page doesn't mean we have to include in the right hand menu... we could just reference from here, and leave it for people to find it that way or through google :D

kwalrath:
> \cc @mit-mit Since you created something similar at #3794

Yes, mit let me know about this :D. He said we could choose one or just create a new one.

src/tools/pub/pubspec.md
```diff
@@ -67,6 +67,10 @@ A pubspec can have the following fields:
 : Optional. Used to put a package's executables on your PATH.
   [_Learn more._](#executables)
 
+`platforms`
+: Optional. Used to explicitly declare supported platforms on the [pub.dev site.]({{site.pub}}).
+  [_Learn more._](#platforms)
```
parlough, line 72:
```suggestion
`platforms`
: Optional. Used to explicitly declare supported platforms
  on the [pub.dev site]({{site.pub}}).
  [_Learn more._](#platforms)
```

src/tools/pub/pubspec.md
```diff
@@ -269,6 +273,65 @@ For more information, see
 [pub global](/tools/pub/cmd/pub-global#running-a-script-from-your-path).
 
 
+### Platforms
+
+When publishing packages to the [pub.dev site.]({{site.pub}}) the set of
+_supported platforms_ will be automatically detected based on transitively
+imported `dart:...` libraries and [Flutter plugin declarations][]. The set of
+supported platforms is displayed on the package page, and used for filtering
+when searching for packages on the [pub.dev site.]({{site.pub}}).
```
parlough, line 282:
```suggestion
When publishing packages to the [pub.dev site]({{site.pub}})
the set of _supported platforms_ will be automatically detected based on 
transitively imported `dart:...` libraries and [Flutter plugin declarations][]. 
The set of supported platforms is displayed on the package page, 
and used for filtering when searching for packages
on the [pub.dev site]({{site.pub}}).
```

src/tools/pub/pubspec.md
```diff
@@ -269,6 +273,65 @@ For more information, see
...
+
+If the supported platforms automatically detected is incorrect, the list of
+_supported platforms_ can be explicitly declared using the `platforms:` key
+as follows:
```
parlough, line 286:
```suggestion
If the supported platforms automatically detected is incorrect,
the list of _supported platforms_ can be explicitly declared
using the `platforms:` key as follows:
```

src/tools/pub/pubspec.md
```diff
@@ -269,6 +273,65 @@ For more information, see
...
+{% prettify yaml tag=pre+code %}
+# This package supports all platforms listed below
+platforms:
+  android:
+  ios:
+  linux:
+  macos:
+  web:
+  windows:
+{% endprettify %}
```
parlough, line 297:
````suggestion
```
# This package supports all platforms listed below
platforms:
  android:
  ios:
  linux:
  macos:
  web:
  windows:
```
````

File: src/tools/pub/pubspec.md [hunk 6]
```diff
@@ -269,6 +273,65 @@ For more information, see
...
+
+The automatic detection of supported platforms is not perfect. For example, a
+package using `chmod` through `Process.run` from `dart:io` won't work on
+Windows, even though `dart:io` is available on Windows. Hence, it is reasonable
+to explicitly declare that the package only supports Linux and MacOS, as
+illustrated below:
```
parlough, line 303:
```suggestion
The automatic detection of supported platforms is not perfect. 
For example, a package using `chmod` through `Process.run` from `dart:io`
won't work on Windows, even though `dart:io` is available on Windows. 
Hence, it is reasonable to explicitly declare
that the package only supports Linux and macOS, as illustrated below:
```

src/tools/pub/pubspec.md
```diff
@@ -269,6 +273,65 @@ For more information, see
...
+
+{% prettify yaml tag=pre+code %}
+# This package only supports Linux and MacOS
+platforms:
+  linux:
+  macos:
+{% endprettify %}
```
parlough, line 310:
````suggestion
```yaml
# This package only supports Linux and macOS
platforms:
  linux:
  macos:
```
````

parlough (commented):
Some initial wording, naming, and formatting fixes (I've committed them as they don't change the contents themselves). Can you also add a small section to `publishing.md` similar to how mit did? https://github.com/dart-lang/site-www/pull/3794/files

src/tools/pub/pubspec.md
parlough: I think I like the idea of moving it to a separate page as this seems like a pretty advanced piece of functionality and we can link to it from other locations discussing conditional imports. @mit-mit What do you think?

parlough:
Also what version of the SDK was support for this introduced?

kwalrath:
Staged: https://kw-www-dartlang-1.web.app/tools/pub/pubspec

kwalrath:
Closing in favor of #3801

PR #3794, opened 2022-02-01
Docs for platform tags in pubspecs

mit-mit:

PR #3801, opened 2022-02-03
Document the platforms pubspec field

kwalrath:
Builds on #3793 & #3794.

### DIFF
--- a/src/tools/pub/publishing.md
+++ b/src/tools/pub/publishing.md
@@ -236,6 +236,38 @@ follow these practices:
   Cancel the upload if any undesired files appear in that list.
 
 
+## Platform support
+
+The [pub.dev site]({{site.pub}}) detects which platforms a package supports,
+displaying these platforms on the package page.
+Users of pub.dev can filter searches by platform.
+
+Automated platform-support detection relies on the package's
+transitively imported `dart:*` libraries and
+on its Flutter plugin declarations.
+For example, a package that requires `dart:html`
+works only on the web platform.
+A package that requires `dart:io`
+can't work on the web platform.
+
+{{site.alert.tip}}
+  If your package supports the web but uses `dart:io` for
+  code that isn't compiled to JavaScript,
+  [conditionally import][] `dart:io`.
+{{site.alert.end}}
+
+Platform-support detection isn't perfect.
+For example, although `dart:io` is available on Windows,
+a package that uses `chmod` (through `Process.run` from `dart:io`)
+can't work on Windows. 
+
+To change the list of platforms that a package supports,
+[specify supported platforms][] in the pubspec.
+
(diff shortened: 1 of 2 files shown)


src/tools/pub/publishing.md
```diff
@@ -236,6 +236,38 @@ follow these practices:
...
 
+## Platform support
+
+The [pub.dev site]({{site.pub}}) detects which platforms a package supports,
+displaying these platforms on the package page.
+Users of pub.dev can filter searches by platform.
+
+Automated platform-support detection relies on the package's
+transitively imported `dart:*` libraries and
+on its Flutter plugin declarations.
```
jonasfj, line 247: nit:

```suggestion
on its [Flutter plugin declarations](https://docs.flutter.dev/development/packages-and-plugins/developing-packages#plugin-platforms).
```

jonasfj (approved):

src/tools/pub/pubspec.md
```diff
@@ -269,6 +274,43 @@ For more information, see
 [pub global](/tools/pub/cmd/pub-global#running-a-script-from-your-path).
 
 
+### Platforms
```
mit-mit, line 277: I like this overall, but I think we need the "For Flutter packages you can use the [Flutter platforms tag][]." part from https://github.com/dart-lang/site-www/pull/3794/files

src/tools/pub/publishing.md
mit-mit, line 247: I find it a bit odd to mention Flutter this early, especially given we're on Dart.dev. I suggest moving that part much lower in the page.

mit-mit (changes requested):

src/tools/pub/publishing.md
kwalrath: Thanks!

kwalrath: This is the one-sentence overview of how pub.dev detects platform support, so it seems appropriate to mention Flutter here. If we move the Flutter mention to somewhere else, then we'll need more general words here like "and other, SDK-specific fields". We could do that in the future if pub.dev adds platform-detection support for other SDKs, but it seems a shame to make it less clear for now.

mit-mit: Right, but my issue is that there is a large difference between the general "guesswork" that is the library analysis, and the "declarations" that are in Flutter's platform tag. The latter is a real declaration just as the new top-level platforms tag is.

I also don't understand why we need to mention Flutter's specific platform tags here in the overview, but don't mention the new Dart tags until the fourth paragraph (just before the "Uploaders" heading.

kwalrath:
Comments addressed, ptal

mit-mit (approved):

mit-mit:
lgtm

parlough (approved):

PR #3822, opened 2022-02-08
A few tweaks

sfshaza2:
@parlough, @mit-mit, @itsjustkevin, I am very sorry that I didn't review parlough's original PR! I'll be more on top of it in future.  :)


parlough (approved):
Thanks for these improvements!

Issue #2338, opened 2020-04-08
Analysis options page should document strict modes


srawlins:
There are three strict analysis modes, `strict-raw-types`, `strict-inference`, and `strict-casts`. These should be documented at https://dart.dev/guides/language/analysis-options.

Summary:
* `strict-raw-types`:
  * since: 2.9.0
  * spec: https://github.com/dart-lang/language/blob/master/resources/type-system/strict-raw-types.md
* `strict-inference`:
  * since: 2.9.0
  * spec: https://github.com/dart-lang/language/blob/master/resources/type-system/strict-inference.md
* `strict-casts`:
  * since: 2.17.0
  * spec: https://github.com/dart-lang/language/blob/master/resources/type-system/strict-casts.md

I'd love to collaborate with someone on how the page should be updated.

These three modes are probably recommended over `implicit-dynamic: false`, so we could deprecate that (or stop documenting it) at the same time. I'll confer with my team on this deprecation idea.

bwilkerson:
I agree. I believe that we're now at a point where we can remove the documentation of the `strong-mode` option / section and replace it with the `language` options Sam mentioned.

@srawlins Just to confirm, are those the only two language options we plan to add at this point?

When we do this, we should be sure to update the characterization of the language options, which currently reads

> Another tag you might see is `language:`, which is used for experimental language features.

I don't think we want to characterize these two modes as experiments.

srawlins:
> I agree. I believe that we're now at a point where we can remove the documentation of the strong-mode option / section and replace it with the language options Sam mentioned.

Unfortunately, I got strong pushback from prominent voices inside google that `implicit-casts: false` is still useful and should not be deprecated. We had a plan to replace this mode with `strict-casts`, which would be almost identical, except that casts from dynamic would be allowed, to match the Null Safety type system. Again, users were saying that `implicit-casts: false` is useful as is. Perhaps we could create `strict-casts` to replace it, and with the _exact same_ functionality.

But this work has not been done yet (I'd be in favor of prioritizing it). Maybe leave `strong-mode` alone for now?

@srawlins Just to confirm, are those the only two language options we plan to add at this point?

Correct. We have plans for `strict-casts` but it has not been prioritized.

> I don't think we want to characterize these two modes as experiments.

Agree. They've been vetted by internal google and by the sass packages.

bwilkerson:
[...]

PR #3823, opened 2022-02-08
Introduce documentation for new strict modes

parlough:
Fixes https://github.com/dart-lang/site-www/issues/2338

Staged: https://dart-dev--pr3823-feature-document-new-rvn1lj8u.web.app/guides/language/analysis-options

### DIFF
--- a/examples/analysis/lib/strict_modes.dart
+++ b/examples/analysis/lib/strict_modes.dart
@@ -6,11 +6,13 @@ import 'dart:convert';
 import 'package:examples_util/ellipsis.dart';
 
 // #docregion strict-casts
-void foo(List<String> lines) => ellipsis<String>();
+void foo(List<String> lines) {
+  ellipsis<String>();
+}
 
 void bar(String jsonText) {
   // ignore: stable, beta, dev, argument_type_not_assignable
-  foo(jsonDecode(jsonText));
+  foo(jsonDecode(jsonText)); // Implicit cast
 }
 // #enddocregion strict-casts
 
--- a/examples/analysis/lib/strict_modes.dart
+++ b/examples/analysis/lib/strict_modes.dart
@@ -1,0 +1,32 @@
+// ignore_for_file: dead_code, prefer_typing_uninitialized_variables
+
+import 'package:examples_util/ellipsis.dart';
+
+String strictCasts() {
+  // #docregion strict-casts
+  dynamic o = ellipsis<String>();
+  // ignore: stable, beta, dev, invalid_assignment
+  String s = o; // Implicit downcast
+  String s2 = s.substring(1);
+  // #enddocregion strict-casts
+  return s2;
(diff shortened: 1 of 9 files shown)


examples/analysis/lib/strict_modes.dart
```diff
@@ -0,0 +1,32 @@
+// ignore_for_file: dead_code, prefer_typing_uninitialized_variables
+
+import 'package:examples_util/ellipsis.dart';
+
+String strictCasts() {
+  // #docregion strict-casts
+  dynamic o = ellipsis<String>();
```
srawlins, line 7: IANATW, but my 2c:

This example is perfectly correct, but if we want to highlight a more common or nefarious case, then parameters and JSON are a great example:

```dart
// a function in one file
void foo(List<String> p) { ... }

// another function far away, different file
void bar(String jsonText) {
  foo(jsonDecode(jsonText));
}
```

This example highlights 1) how code not on your screen can declare the type to which you are casting, and 2) how the code looks fine at compile-time, but could 100% crash at runtime, because of the implicit cast.

[1] I am not a technical writer :)

examples/analysis/lib/strict_modes.dart
```diff
@@ -0,0 +1,32 @@
...
+  dynamicValue.add(2);
+  // #enddocregion strict-inference
+}
+
+void strictRawTypes() {
+  // #docregion strict-raw-types
+  // ignore: stable, beta, dev, strict_raw_type
+  List numbers = [1, 2, 3]; // List with raw type
+  for (final n in numbers) {
+    print(n.length); // Runtime error
```
srawlins, line 29: Oof, so juicy. great example.

examples/analysis/lib/strict_modes.dart
```diff
@@ -0,0 +1,32 @@
...
+  String s = o; // Implicit downcast
+  String s2 = s.substring(1);
+  // #enddocregion strict-casts
+  return s2;
+}
+
+void strictInference() {
+  // #docregion strict-inference
+  // ignore: stable, beta, dev, inference_failure_on_uninitialized_variable
+  var dynamicValue;
```
srawlins, line 18: The collection literal examples might be more interesting, and I think are more common errors than this example.

https://github.com/dart-lang/language/blob/master/resources/type-system/strict-inference.md#collection-literal

parlough: I had a little trouble with this as I see the potential errors here having a lot of similarity to `strict-raw-types` but let me know what you think of the newest example?

srawlins (approved):

srawlins:
Wow I love it!

domesticmouse (approved):

sfshaza2 (approved):
Looks good to me and, more importantly, @srawlins likes it!

PR #3824, opened 2022-02-09
Add missing DevTools link

mit-mit:
Screen Shot 2022-02-09 at 15 17 29


parlough:
Sorry I meant to remove the the entry from resources as the package is discontinued and has no extra information. Does that seem fine instead?

PR #3825, opened 2022-02-09
Update check sums to fix build

parlough:


mit-mit:
This is fine. But I do wonder if there is some way of layering this on top of the official Dart stable image rather than hard-coding SHAs? cc @athomas who might have ideas?

mit-mit (approved):

PR #3826, opened 2022-02-09
Simplify devtools page links

parlough:
Staged: https://dart-dev--pr3826-misc-simplify-devtoo-ise5pie9.web.app/tools/dart-devtools

- Remove resources as it links to the same place as the other Flutter mentions on the page and the now discontinued package which we haven't documented how to use since https://github.com/dart-lang/site-www/commit/ccf58273b9f83acc4d566499469a936a23899a67
- Use the same link shortcut for the links pointing to the same overview link

Happy to hear lots of suggestions here such as introducing a version note linking to the package instead. I decided against it for now, as we haven't documented the package method since October and it has seemed fine so far.


mit-mit (approved):
Nice, thanks

Issue #3810, opened 2022-02-06
Code excerpts being out of date does not fail GitHub actions checks


parlough:
### Describe the problem

After the infrastructure updates, code excerpts being out of date or failing to run does not cause the GitHub actions tests to fail.

### Expected fix

If the code excerpter fails, the GitHub tests should also fail.

PR #3828, opened 2022-02-09
Fix tests as well as excerpt, formatting, and analysis checks

parlough:
Also makes the necessary fixes to fix the newly working tests.

Closes https://github.com/dart-lang/site-www/issues/3810


sfshaza2 (approved):
Wow, @parlough, thanks for this Herculean effort! lgtm

PR #3829, opened 2022-02-10
Simplify async main return types to not return a Future

parlough:
@mit-mit I'm not sure what you discussed at the language team meeting, but it seems the desire is to not suggest using `Future<void>` for the main method. How do these changes look?

### DIFF
--- a/examples/async_await/bin/async_example.dart
+++ b/examples/async_await/bin/async_example.dart
@@ -12,7 +12,7 @@ Future<String> fetchUserOrder() {
   return Future.delayed(const Duration(seconds: 4), () => 'Large Latte');
 }
 
-Future<void> main() async {
+void main() async {
   countSeconds(4);
   await printOrderMessage();
 }
--- a/examples/async_await/bin/get_order.dart
+++ b/examples/async_await/bin/get_order.dart
@@ -11,7 +11,7 @@ Future<String> fetchUserOrder() =>
     );
 
 // #docregion main-sig
-Future<void> main() async {
+void main() async {
   // #enddocregion main-sig
   print('Fetching user order...');
   // #docregion print-order
--- a/examples/async_await/bin/try_catch.dart
+++ b/examples/async_await/bin/try_catch.dart
@@ -19,6 +19,6 @@ Future<String> fetchUserOrder() {
   return str;
 }
 
-Future<void> main() async {
+void main() async {
   await printOrderMessage();
 }
--- a/examples/html/test/html_test.dart
+++ b/examples/html/test/html_test.dart
@@ -62,7 +62,7 @@ void main() {
   test('getString', () async {
     final url = 'https://httpbin.org';
     // #docregion getString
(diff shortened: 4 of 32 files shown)


examples/async_await/bin/async_example.dart
```diff
@@ -12,7 +12,7 @@ Future<String> fetchUserOrder() {
   return Future.delayed(const Duration(seconds: 4), () => 'Large Latte');
 }
 
-Future<void> main() async {
+void main() async {
```
domesticmouse, line 15: This should trigger a lint warning. An `async` function always returns a `Future`.

domesticmouse (commented):
I'm surprised this change isn't triggering lint warnings somewhere.

examples/async_await/bin/async_example.dart
parlough: That's the [`avoid_void_async`](https://dart.dev/tools/linter-rules#avoid_void_async) rule which we don't have enabled.

I am missing some context but it seems the language team decided it's best to discourage `Future<void>` for the `main` function or at least not showcase it in examples. Perhaps because it suggests the result is useful. There is [a linter issue](https://github.com/dart-lang/linter/issues/3141) considering adding it as an exception to the lint.

Hoping @mit-mit can provide some more thoughts.

domesticmouse: Oh, that's new to me. If @mit-mit is happy with this PR, I'm happy.

domesticmouse (approved):
LGTM if @mit-mit is happy

examples/async_await/bin/async_example.dart
mit-mit: You are right generally @domesticmouse but we discussed making an exception for `main` specifically. @pq @lrhn should we adjust the lint to include that exception?

lrhn: I'd be happy with changing the lint to not trigger on `main` functions. 
There is no official rules to not have `void` `async` functions, the lint is not part of the Dart team's official `lints` package, probably because there is too high a risk of false positives (there are reasonable cases for having a `void`/`async` function, they are just not easily distinguishable from doing it by mistake). 
Not triggering on `main` might reduce the false positives.

(Personally, I'll likely never enable the lint, because I am likely to get false positives. I write the kind of code that gets those false positives. I know Google's internal code style is stricter about disallowing things that can *potentially* be an error, and that's fine too. That's why it's a lint, but not a recommended one, so people can opt-in if they want it.)

pq: I'm in favor of special-casing `main` in `avoid_void_async ` and have updated https://github.com/dart-lang/linter/issues/3141 to a "proposal".  Unless there are objections, I'll go for it.

sfshaza2:
Should we defer this PR until the linter is updated, @mit-mit?

sfshaza2:
Thanks, @mit-mit and @parlough!

PR #3830, opened 2022-02-11
add css styles for material icons in head.html

hrk2023:


src/_includes/head.html
```diff
@@ -66,6 +66,15 @@
...
 
+  <!-- Material Icon Styles -->
+  <style>
+    .material-icons.md-18 { font-size: 18px; }
+    .material-icons.md-24 { font-size: 24px; }
+    .material-icons.md-36 { font-size: 36px; }
+    .material-icons.md-48 { font-size: 48px; }
+  </style>
+
+
```
parlough, line 77: So actually we don't want to include styles here in the header and this won't be consistent with our current external link styles.

```suggestion
```

parlough (changes requested):
Thanks so much for this!

I was thinking we could more closely align with our styles from the other [external link icon](https://github.com/dart-lang/site-www/blob/main/src/_sass/core/_base.scss#L57-L61) used elsewhere.

So we could add an entry to the sidebar external link styles here: https://github.com/dart-lang/site-www/blob/main/src/_sass/components/_sidebar.scss#L177

That looks something like:

```css
span {
  vertical-align: middle;
  margin-left: 2px;
  font-size: 12px;
  color: inherit;
  opacity: 0.8;
}
```

How does something like that sound? That way our external link styles can be more consistent across the entire site.

src/_includes/head.html
hrk2023: Ok I will make the changes

sfshaza2:
What a great idea, @parlough! And, @hrk2023, thanks for this PR!

parlough:
If you need any extra help implementing the changes @hrk2023, let me know. Thanks :)

hrk2023:
I am working on the changes. Was a little occupied with some other work.

src/_includes/head.html
hrk2023: @parlough I have made the changes that you have requested. 
```
span {
  vertical-align: middle;
  margin-left: 2px;
  font-size: 12px;
  color: inherit;
  opacity: 0.8;
}
```

Instead of using styles in the above way, I have applied the styles to the` .material-icons` class as applying styles in span would result in rendering styles for every span tag that uses the `_sidebar.css` file.

src/_includes/head.html
```diff
@@ -66,6 +66,7 @@
     <link href="{{css}}" rel="stylesheet" type="text/css">
   {% endfor -%}
 
+
```
domesticmouse, line 69: I'm curious why the additional blank line

domesticmouse (approved):

parlough (approved):
I made some final minor changes, particularly to limit the scope of the styles since these changes are required to allow other site-shared changes to land.

Thanks so much for your hard work on this!

PR #3831, opened 2022-02-13
Fix installation method for Debian

hongquan:
Since Ubuntu 20.10, setting up APT repo key with `apt-key` is deprecated. It is recommended to specify separate key for 3rd-party repo. This PR is to update the guide to follow newer method.

This updated guide also avoid running shell with `sudo` to reduce security concern.

### DIFF
--- a/src/get-dart/_linux.md
+++ b/src/get-dart/_linux.md
@@ -12,8 +12,8 @@ Perform the following **one-time setup**:
 ```terminal
 $ sudo apt-get update
 $ sudo apt-get install apt-transport-https
-$ sudo sh -c 'wget -qO- https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add -'
-$ sudo sh -c 'wget -qO- https://storage.googleapis.com/download.dartlang.org/linux/debian/dart_stable.list > /etc/apt/sources.list.d/dart_stable.list'
+$ wget -qO- https://dl-ssl.google.com/linux/linux_signing_key.pub | sudo gpg --dearmor -o /usr/share/keyrings/dart.gpg
+$ echo 'deb [signed-by=/usr/share/keyrings/dart.gpg arch=amd64] https://storage.googleapis.com/download.dartlang.org/linux/debian stable main' | sudo tee /etc/apt/sources.list.d/dart_stable.list
 ```
 Then install the Dart SDK:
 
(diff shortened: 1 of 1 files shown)


parlough:
Hi @hongquan, thanks so much for the PR! You mentioned the old method is deprecated. What versions support this new strategy?

parlough:
Thanks for the clarification, just making sure we can just document the one strategy :)

sfshaza2:
Thanks, @hongquan, @athomas, and @parlough, for your contributions to this PR! I don't know much about debian but, on reading through this PR, I wonder if we should add a note stating that, as of Debian release xxx, that the apt approach is deprecated.

@thomas, as our official Google reviewer here, I defer to you on this. It sounds like, at the very least, we should add info about the SecureApt option.

hongquan:
> we wonder if we could use SecureApt 
@athomas : We already use SecureApt, because:

- We "dearmour" the GPG key:
 ```
 wget -qO- https://dl-ssl.google.com/linux/linux_signing_key.pub | sudo gpg --dearmor -o /usr/share/keyrings/dart.gpg
```
- We use HTTPS:
 ```
 https://storage.googleapis.com/download.dartlang.org/linux/debian stable main
 ```
 
 > I wonder if we should add a note stating that, as of Debian release xxx, that the apt approach is deprecated
 
 @sfshaza2 No need, because every time user runs `apt-key` on that release, user will see a warning about deperecation.

sfshaza2 (approved):
OK, I'm going to merge this, because it sounds like @hongquan has addressed the open issues.. Thanks for the PR, @hongquan!

parlough:
Hi @hongquan,

Thanks again for working on this. Looking back on these changes, one piece I noticed is we now specify `arch=amd64` in the `deb` command. Since these instructions should work across supported architectures, could you elaborate on what this does? Could we simply not specify the architecture somehow (this would be nice)? Or should I add a note to modify this depending on their current architecture. 

Thanks!

hongquan:
Hi @parlough 

That `arch=amd64` is not my change. It is inside the file https://storage.googleapis.com/download.dartlang.org/linux/debian/dart_stable.list, which was used in old method.

parlough:
@hongquan Ah, I see. Thanks so much for the clarification, I didn't think to check that file.

Issue #3785, opened 2022-01-28
Add a key binding to focus search bar


parlough:
### Describe the problem

A lot of websites have a key binding to focus (and open if needed) the search bar. It's convenient, particularly for keyboard users.

### Expected fix

Add a key binding to focus the search bar.

Google Search and GitHub both use `/` so I think it would be a good choice.

Consider bringing over the implementation to docs.flutter.dev as well.

parlough:
Any thoughts?

kwalrath:
@khanhnwin or @johnpryan what do you think?

johnpryan:
Sounds great, it looks like [api.flutter.dev](https://api.flutter.dev/) uses `/` too.

khanhnwin:
Sgtm! That'd be great!

PR #3844, opened 2022-02-20
Add a key binding to focus search input

parlough:
Adds a key binding ("/") to open and target the appropriate search bar:

- If the side menu is open, always targets its search bar.
- If the 404 page is open and the side menu is closed, targets its inline search bar.
- If the side menu is not open and it's not the 404 page, target the main nav search bar.

Closes https://github.com/dart-lang/site-www/issues/3785


sfshaza2 (commented):
FYI, John is out until Thursday.  @domesticmouse

johnpryan (approved):

PR #3847, opened 2022-02-20
Fix a typo in async-await.md

thezjy:


domesticmouse:
PTAL @parlough

src/codelabs/async-await.md
```diff
@@ -232,7 +232,7 @@ If the function doesn't explicitly return a value, then the return type is
 
 <?code-excerpt "async_await/bin/get_order.dart (main-sig)" replace="/Future<\w+\W/[!$&!]/g;  /$/ ··· }/g"?>
 {% prettify dart tag=pre+code %}
-void main() async { ··· }
+Future<void> main() async { ··· }
```
parlough, line 235: We want to highlight the `Future` here :)

```suggestion
[!Future<void>!] main() async { ··· }
```

parlough (changes requested):
Thanks for fixing this! I accidentally messed this up when the tests weren't working properly.

You'll also need to switch the `void` back to `Future<void>` in the example code as well: https://github.com/dart-lang/site-www/blob/main/examples/async_await/bin/get_order.dart#L14

Let me know if you have any issues :)

domesticmouse (approved):

parlough (approved):
Thanks again for catching that we accidentally changed this and fixing it!

PR #3850, opened 2022-02-22
Add concurrency to the sidenav

mit-mit:


parlough (approved):
Looks good to me, thanks for adding this.

src/_data/side-nav.yml
```diff
@@ -49,6 +49,8 @@
       permalink: /guides/language/extension-methods
     - title: Numbers
       permalink: /guides/language/numbers
+    - title: Concurrency
+      permalink: /guides/language/concurrency
```
parlough, line 53: (Optionally), what do you think about flipping the order of these? I feel the concurrency discussion is generally more useful and it seems to look a bit better.

```suggestion
    - title: Concurrency
      permalink: /guides/language/concurrency
    - title: Numbers
      permalink: /guides/language/numbers
```

mit-mit: sgtm, done~

PR #3851, opened 2022-02-23
Update template names for `dart create`

mit-mit:
Related to:
https://dart-review.googlesource.com/c/sdk/+/232800

PR #3853, opened 2022-02-23
Revert "Update template names for `dart create`"

parlough:
As far as I can tell, these new names won't be supported until at least Dart 2.17, so I don't think we should merge these changes yet. Users will be confused if the specified templates don't work with their recent version of Dart.

Reverts dart-lang/site-www#3851

### DIFF
--- a/src/_tutorials/libraries/shared-pkgs.md
+++ b/src/_tutorials/libraries/shared-pkgs.md
@@ -65,13 +65,13 @@ $ dart create --help
 ```
 
 You'll see a list of templates, including various web and server-side apps.
-One of the templates is named **console**.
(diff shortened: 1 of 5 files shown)


sfshaza2:
Ahhh, I apologize for jumping the gun!! I'll wait for @mit-mit to review and merge.

sfshaza2 (approved):
lgtm, but waiting on @mit-mit!

domesticmouse (approved):
This should be landed, I believe

mit-mit:
Yeah, reverting the land is fine. I created the original PR to prepare for the upcoming release, so it should be tagged as dependent on the stable release.

PR #3866, opened 2022-03-03
description typo fixes.

chenglu:
Updated the description to "Learn how to remove the elements from DOM",
please check it out, thanks!

I can almost imagine the issue is from where we:
first we added three documents on the same topic, 
and then forgot to change the description when adding them by copying and pasting ;-)

### DIFF
--- a/src/_tutorials/web/low-level-html/remove-elements.md
+++ b/src/_tutorials/web/low-level-html/remove-elements.md
@@ -1,6 +1,6 @@
 ---
 title: Remove DOM elements
-description: Learn how to remove the elements from DOM.
+description: Learn how to remove children elements from the DOM.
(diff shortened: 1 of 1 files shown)


src/_tutorials/web/low-level-html/remove-elements.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: Remove DOM elements
-description: Learn how to set up basic scaffolding of a Dart web app.
+description: Learn how to remove the elements from DOM.
```
parlough, line 3:
```suggestion
description: Learn how to remove children elements from the DOM.
```

parlough (approved):
Oops! Thanks for catching this :)

PR #3871, opened 2022-03-05
Try using Google Sans Mono to improve code font style

parlough:
Roboto Mono hasn't been the most readable looking in a lot of places, so I thought we should give Google Sans Mono a shot instead. So far, it seems like it is an improvement to readability overall, open to other's thoughts though!

We use some of the other Google Sans derivatives elsewhere on the site so it doesn't have any problems fitting in.

I recommend comparing the language tour:

**Staged:** https://dart-dev--pr3871-feature-use-google-s-eykjao42.web.app/guides/language/language-tour
**Live:** https://dart.dev/guides/language/language-tour

### DIFF
--- a/src/_includes/head.html
+++ b/src/_includes/head.html
@@ -51,10 +51,10 @@
 
   <link rel="preconnect" href="https://fonts.googleapis.com">
   <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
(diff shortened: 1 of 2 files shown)


domesticmouse (approved):

sfshaza2:
When I try to view your staged site, @parlough, I get the following screen:
Screen Shot 2022-03-07 at 9 56 28 AM

So, I will trust @domesticmouse's review.  :)

PR #3875, opened 2022-03-06
Revert "Remove jquery usage from os tabs"

parlough:
This is causing issues preventing some site interactive elements to fail to a NPE.

Reverts dart-lang/site-www#3869

PR #3876, opened 2022-03-06
Fix os-tabs error breaking interactive elements and CSS

parlough:
I accidentally introduced this bug when converting from jquery to vanilla JS, as jquery gracefully handled not finding the elements. These elements don't exist on every page, causing a NPE and breaking the loading of some interactive elements (menu button).

Supersedes https://github.com/dart-lang/site-www/pull/3875


domesticmouse (approved):

parlough:
Thank so much @domesticmouse!

PR #3901, opened 2022-03-14
user-data in PUB_HOSTED_URL is no longer supported

jonasfj:
Never realized that this was an advertised feature. In any case, we removed user-data support in the URL a while ago.

Having username and password in the URL is unfortunate for many reasons:
 * Using hosted-dependency syntax in `pubspec.yaml` would put secrets in `pubspec.yaml`.
 * The URL and embedded user-data is cached in `PUB_CACHE` (potentially leaking secrets and minimizing cache reuse).
 * The URL and embedded user-data is cached in `pubspec.lock` (I'm guessing, I didn't test it).

In any case, using `dart pub token add` is a much better solution, see:
https://dart.dev/tools/pub/custom-package-repositories

### DIFF
--- a/src/tools/pub/environment-variables.md
+++ b/src/tools/pub/environment-variables.md
@@ -23,8 +23,10 @@ Environment variables allow you to customize pub to suit your needs.
   use the `PUB_HOSTED_URL` environment variable. For example:
 
 {% prettify sh tag=pre+code %}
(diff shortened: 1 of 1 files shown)


parlough (approved):
Thanks for clarifying this and adjusting the example!

PR #3912, opened 2022-03-14
Mention that Future-based APIs are usually backed by code on other threads

Hixie:


src/codelabs/async-await.md
```diff
@@ -42,6 +42,17 @@ another operation to finish. Here are some common asynchronous operations:
 To perform asynchronous operations in Dart, you can use the `Future` class
 and the `async` and `await` keywords.
 
+Typically, APIs that return `Future`s fall into one of the following categories:
```
parlough, line 45: Instead of [making a code element plural](https://developers.google.com/style/code-in-text?hl=en#grammatical-treatment-of-code-elements), we add a noun:

```suggestion
Typically, APIs that return `Future` objects
fall into one of the following categories:
```

src/codelabs/async-await.md
```diff
@@ -42,6 +42,17 @@ another operation to finish. Here are some common asynchronous operations:
...
+
+* APIs that are backed by logic running on a different thread. For example,
+  `File.readAsString` (and most of the other I/O APIs) call operating system
+  functions from threads on a dedicated thread pool, rather than from the
+  thread on which Dart code is exectured.
```
parlough, line 50: Just adjusting the semantic line breaks and fixing the spelling of "executed".

```suggestion
* APIs that are backed by logic running on a different thread. 
  For example, `File.readAsString` (and most of the other I/O APIs) 
  call operating system functions from threads on a dedicated thread pool, 
  rather than from the thread on which Dart code is executed.
```

src/codelabs/async-await.md
```diff
@@ -42,6 +42,17 @@ another operation to finish. Here are some common asynchronous operations:
...
+
+* APIs that wait for an event of some kind and perform no work in the meantime.
+  For example, `Future.delayed` waits for time to have elapsed, and `Stream.first`
+  waits for an event to be posted on the stream.
```
parlough, line 54: Adjusting line breaks to fit in width.

```suggestion
* APIs that wait for an event of some kind and perform no work in the meantime.
  For example, `Future.delayed` waits for time to have elapsed, 
  and `Stream.first` waits for an event to be posted on the stream.
```

parlough (approved):
These extra mentions look great to me. A few minor formatting comments.

What do you (optionally) think about including this under the [What is a future?](https://dart.dev/codelabs/async-await#what-is-a-future) section instead? Perhaps before the note: https://github.com/dart-lang/site-www/blob/main/src/codelabs/async-await.md?plain=1#L111

parlough:
@lrhn Could you also take a quick look at these new mentions?

src/codelabs/async-await.md
```diff
@@ -42,6 +42,18 @@ another operation to finish. Here are some common asynchronous operations:
 To perform asynchronous operations in Dart, you can use the `Future` class
 and the `async` and `await` keywords.
 
+Typically, APIs that return `Future` objects
+fall into one of the following categories:
+
+* APIs that are backed by logic running on a different thread. 
+  For example, `File.readAsString` (and most of the other I/O APIs) 
+  call operating system functions from threads on a dedicated thread pool,
```
lrhn, line 50: Non-native English speaker here: Should `call` be `calls` since it binds to `File.readAsString`, or does the parenthesis affect it?

lrhn, line 50: I wouldn't mention a thread pool, it is not important to the point:
Some futures are waiting for operations running outside of the Dart isolate, other futures are waiting for asynchronous operations inside the Dart isolate.

It's not so much "the thread on which Dart code is executed" as it's the thread that the current isolate is being executed on. Other isolates, also running Dart code, counts as other threads too.

(Those futures completed inside the same isolate are usually completed using microtasks, those from outside the isolate are usually completed by top-level events, but exceptions can occur.)

src/codelabs/async-await.md
```diff
@@ -42,6 +42,18 @@ another operation to finish. Here are some common asynchronous operations:
...
+fall into one of the following categories:
+
+* APIs that are backed by logic running on a different thread. 
+  For example, `File.readAsString` (and most of the other I/O APIs) 
+  call operating system functions from threads on a dedicated thread pool, 
+  rather than from the thread on which Dart code is executed.
+
+* APIs that wait for an event of some kind and perform no work in the meantime.
+  For example, `Future.delayed` waits for time to have elapsed, 
+  and `Stream.first` waits for an event to be posted on the stream.
```
lrhn, line 55: I'd add an "asynchronouse" here
> ... that wait for an asynchronous event of some kind ...

and possibly append

> This includes all user-written `async` functions.

src/codelabs/async-await.md
```diff
@@ -42,6 +42,18 @@ another operation to finish. Here are some common asynchronous operations:
...
+
```
lrhn, line 56: The distinction seems a little artificial. The "perform no work in the meantime" applies to both categories. Both spin off an asynchronous computation, then wait for it to complete, and does nothing until it completes. The only difference is whether the asynchronous computation is running in the same thread/isolate or another thread. I'm not sure why that matters to the client of the `Future` returning function. They don't need to know, the `Future` is an abstraction over the run-time characteristics of the computation.

If the computation is running in a separate thread, it needs to somehow communicate the result back to the isolate containing the future. When it does that, it usually triggers an asynchronous (top-level, not microtask) event, at which point the future is really just waiting for an asynchronous event in the current isolate.

lrhn (commented):
I'm not sure the distinction is really significant.

What is the goal of introducing this text. What is the reader supposed to do with the information?

If it's just to introduce the concept that a future can be completed by a result computed in a separate thread or isolate, just say that directly.

src/codelabs/async-await.md
parlough: Nice catch. I believe technically the content inside the parentheses should not affect the surrounding sentence, so it should be `calls`. 

It seems a bit weird if you read the parenthetical statement though, so if we keep this text, perhaps we should just remove the parentheses.

Hixie:
Context is https://www.reddit.com/r/FlutterDev/comments/tcp8kt/has_anyone_ever_had_a_need_to_use_isolates/

Hixie:
I have no particular opinion on the comments above and am happy to defer to whatever y'all think is best; my goal is just to address the issue that some of our users don't understand that async functions are actually waiting on code that happens on another thread as opposed to just scheduling synchronous work on the same thread, as we don't seem to document that anywhere currently.

See also https://dart-review.googlesource.com/c/sdk/+/237241 which tries to address this for API docs also.

lrhn:
That CL is fine. It says that "futures are often used for comptuatations in other threads/isolates", which is vague enough to be correct: For such computations, futures are often used - or stream. (Most futures are not directly backed by other-thread computations, they just propagate asynchrony inside the current isolate, but at the root of that asynchrony is usually something inherently external)

We can say the same thing about streams.

lrhn:
This is a hard concept to describe because it's one that isn't actually very visible. 

Futures do not promise anything about where the computation happens. Futures do not represent computations at all, they represent the *results* of computations.
We can't say that a future represents something happening in another thread, because it might not. 
And you *shouldn't care*. Not at the future level.

Functions which spawn new isolates should sometimes say so. 
Functions in `dart:io` which are doing non-blocking I/O should *not* say anything. It's an implementation detail how that's implemented (it might just be using non-blocking OS functions and OS signals to know when data is ready, or it can run blocking I/O functions in a separate isolate and communicate the result back through a send-port, you shouldn't need to know.)

It's a case where the author should consider what the user needs to know.

That's why I think saying these things about *futures* is not necessarily the right place. It's not a bout the futures.
It's OK to introduce the notion of things happening on other threads.

Maybe an alternative phrasing:

> Asynchronous computations are often the result of doing some computation outside of the current Dart code,
> a computation which doesn't complete immediately, and where you are not willing to block all your Dart code waiting for
> the result.
> Those computations can be operating system calls like non-blocking I/O, performing an HTTP request
> or communicating with a browser, it can be waiting for computations performed in another Dart isolate,
> or maybe just waiting for a timer to trigger. All of these either run in a different thread, or are handled 
> by the operating system or the Dart runtime, so that it allows Dart code to run concurrently with the computation.
>
> Such asynchronous computations usually provide their result as a `Future` (or, if the result has multiple parts or is just
> so big that it needs to be cut into multiple parts, as a `Stream`). The computations are *inherently* asynchronous
> and these are the functions which introduce asynchrony into a program.
> Then, to accommodate that initial asynchrony, other plain Dart functions also need to become asynchronous.
> Most asynchronous functions are just `async` Dart functions which depend, possibly deep down, on an inherently 
> asynchronous computation.
> 
> (There is one reason to introduce asynchrony into a computation that is running entirely in the same Dart isolate:
> To allow other code to interrupt an otherwise long-running computation.)

This isn't wrong. I'm not sure it's the right level of detail at this point in the codelab. 
I'm also not sure the original was.
In either case, it's getting woefully close to describing the philosophical reasoning behind asynchronous code, not just teaching people to code it. (Which I'm all for doing!)

telenieko:
> Nice catch. I believe technically the content inside the parentheses should not affect the surrounding sentence, so it should be `calls`.
> 
> It seems a bit weird if you read the parenthetical statement though, so if we keep this text, perhaps we should just remove the parentheses.

If correct grammar is part of the documentation style guide then it has to be written as `calls`. See [this post on stackexchange just about that](https://ell.stackexchange.com/q/310333) which refers to [the punctuation guide](https://www.thepunctuationguide.com/parentheses.html) (1st example). 

Best to rephrase it:

- For example, most I/O APIs like `File.readAsString` call operating system functions
- For example, `File.readAsString` (among most of the other I/O APIs) calls operating system functions

The second one keeps the phrase almost intact, though...

IMO, the 1st one is better as "most I/O APIs" becomes the subject and `File.readAsString` an example among I/O APIs; not the other way around. Which makes sense.

Regarding how authoritative "The Punctuation Guide" is as a source for this see [it's about page](https://www.thepunctuationguide.com/about-this-guide.html).

parlough:
I took some of the core ideas mentioned here and included them in the codelab and concurrency article in https://github.com/dart-lang/site-www/pull/4014

I limited the amount to include in the async-await codelab as it was too much detail for the purposes of the codelab. 

Thanks for your thoughts and feel free to open a new issue or PR if you'd like to see any further changes.

PR #3914, opened 2022-03-15
Mark Windows 11 as supported

parlough:
Windows 11 is growing to almost 10% of the Windows market share according to some sites, and I'm assuming among developers it is even larger. 

Once testing is added on the SDK side, we should signify support here.


parlough:
@athomas Is it okay to mark Windows 11 as officially supported here?

domesticmouse (approved):

mit-mit:
@athomas do we do any testing on 11?

athomas:
> @athomas do we do any testing on 11?

Not currently.

domesticmouse:
I'm developing Flutter desktop apps on Win11. We should probably add it to the test pool.

sfshaza2 (commented):
lgtm but I defer to @mit-mit.

mit-mit:
I'd prefer not to add this until we get the testing added.

But at 10% of share we really should get that done. @athomas is that something we can do fairly soon?

mit-mit (approved):

sfshaza2 (approved):
lgtm

parlough:
Thanks for the followup!

PR #3918, opened 2022-03-17
Link to the blog post on private package repositories

jonasfj:


domesticmouse (approved):

parlough:
I think this article fine to link to, but it seems to duplicate a lot of the same content from this page (except in more of a news form), especially with the addition of https://github.com/dart-lang/site-www/pull/3919. 

Is there anything new in particular from the article that you'd like to see gained by the reader? If so, perhaps we could add that to the page instead of linking to it. That way users don't read the same thing twice and we don't have to worry about the blog post becoming out of date.

sfshaza2:
Actually, @parlough, it was my idea that @jonasfj add this link. I would also be ok with adding info to the page on dart.dev, but, after reading through the page on dart.dev, it seems like a non-trivial task right now. In fact, I wonder if Kathy ever reviewed this page, as it could use some tech editing.  :)

So, for now, this is LGTM for me.

parlough:
Sounds good to me, thanks!

I’ll file an issue to track some improvements to the page.

sfshaza2:
Thanks!!!

PR #3919, opened 2022-03-17
Link to Cloudsmith and JFrog

jonasfj:
I'm not sure if we could some custom HTML, embed the logos in a table or something.


domesticmouse (approved):

parlough:
We have something like you mentioned for IDEs: https://dart.dev/tools#ides-and-editors

Would you like me to setup something like that for this as well?

sfshaza2 (approved):
lgtm!

PR #3947, opened 2022-03-24
Update dockerfile for Dart 2.16.2 release

parlough:


domesticmouse (approved):

AlexV525:
Can we skip the sum check instead? AFAICT flutter/website doesn't apply the check with their build steps.

PR #3948, opened 2022-03-25
fix missing path in `--set-exit-if-changed` example

DetachHead:


parlough (approved):
Thanks for fixing this! It's a good idea to be consistent here since a file (or directory) is required :)

PR #3949, opened 2022-03-26
Effective Dart: Add back guidance about linking to the unnamed constructor

jamesderlin:
Effective Dart: Add back guidance about linking to the unnamed constructor

Now that Dart has constructor tear-offs and a consistent way to refer
to unnamed constructors, undo https://github.com/dart-lang/site-www/pull/3006
with the new `.new` syntax.

### DIFF
--- a/examples/misc/lib/effective_dart/docs_good.dart
+++ b/examples/misc/lib/effective_dart/docs_good.dart
@@ -85,7 +85,7 @@ void miscDeclAnalyzedButNotTested() {
     void method1() {}
 
     // #docregion ctor
-    /// To create a point, call [Point.new()] or use [Point.polar()] to ...
+    /// To create a point, call [Point.new] or use [Point.polar()] to ...
     // #enddocregion ctor
     void method2() {}
   };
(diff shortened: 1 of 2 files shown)


parlough:
Looks mostly good to me, but I wonder if just `Point.new` should be preferred here since it matches the tear-off syntax. 

\cc @srawlins @munificent

jamesderlin:
Oh, good point.  Also, per https://github.com/dart-lang/dartdoc/issues/1850#issuecomment-1079605071, I've realized that `[Point()]` does now seem to work as intended, so we also could revert to that.  That seems tempting since that's how most people actually would invoke the constructor, and it'd look visually consistent alongside `Point.polar()`.

OTOH, using `[Class.new]` in documentation would help advertise how to create tear-offs for the unnamed constructor.

parlough:
@jamesderlin I think we should change `[Point.polar()]` to `[Point.polar]` as well. That is consistent with tear-offs and is what the analysis server supporters.

jamesderlin:
> @jamesderlin I thought about it some more, and I think we should change `[Point.polar()]` to `[Point.polar]` as well. That is consistent with tear-offs and is what the analysis server supports. What do you think?

Sounds good to me.

parlough (approved):
Looks good to me, thanks for bringing this piece back and up to date with the modern dartdoc standards.

Issue #3874, opened 2022-03-06
"CONSIDER copying a nullable field..." recommendation should be clearer


jamesderlin:
### Page URL

https://dart.dev/guides/language/effective-dart/usage#consider-copying-a-nullable-field-to-a-local-variable-to-enable-type-promotion

### Page source

https://github.com/dart-lang/site-www/tree/main/src/_guides/language/effective-dart/usage.md

### Describe the problem

The "CONSIDER copying a nullable field to a local variable to enable type promotion." recommendation uses the term "copy" in the recommendation title and in several places in the explanation. "Copy" often has a different meaning (a different object that compares equal), so the usage is potentially misleading.

At least one person on StackOverflow has been confused by this, and convincing them that no object was being copied unfortunately took more effort than it should have.

### Expected fix

I think that the recommendation should be adjusted to instead use "assign" or "alias" instead of "copy" everywhere to make it clear that the referenced object is not being copied and that the member variable and the local variable both refer to the same object.

I also would recommend using `final` for the local variable to disallow writing back to the field accidentally.

parlough:
I can definitely see where the confusion can come in, thanks for the explanation and detailed context @jamesderlin.

In [Fixing type promotion issues](https://dart.dev/tools/non-promotion-reasons) we are pretty direct and say "Create a local variable that holds the value of X", but I can see some word like "aliases" instead of "hold" working well.

@munificent Do you have any thoughts on how we can clarify the wording here?

munificent:
Yeah, I agree. "Copy" is totally the wrong word to use here. I like "assign":

> CONSIDER assigning a nullable field to a local variable to enable type promotion.

PR #3952, opened 2022-03-29
Effective Dart: Adjust the recommendation about copying nullable fields

jamesderlin:
Adjust the "CONSIDER copying a nullable field..." recommendation to
use "assign" instead of "copy". "Copy" usually means creating a new
object.

Also recommend using `final` for the local variable. (This is already
mentioned by https://dart.dev/tools/non-promotion-reasons#property-or-this,
but I think it should be mentioned here too, especially since I expect
Effective Dart to have higher visibility.) Update the examples to match.

Fixes https://github.com/dart-lang/site-www/issues/3874.

### DIFF
--- a/examples/misc/lib/effective_dart/usage_bad.dart
+++ b/examples/misc/lib/effective_dart/usage_bad.dart
@@ -322,7 +322,7 @@ Item? bestDeal(List<Item> cart) {
 
 //----------------------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


domesticmouse (approved):

parlough (approved):
These changes are great, thanks so much.

Nice updating the excerpt IDs to be consistent too.

PR #3955, opened 2022-03-30
Update redirects in firebase.json

mit-mit:
The repo https://github.com/dart-lang/dart-samples was archived, so removing all redirects to that.


parlough (approved):
Removing these seems fine to me with the redirect to the new samples repository, which worked as expected when testing locally. .

I fixed the build in https://github.com/dart-lang/site-www/pull/3953

mit-mit:
@parlough tests are passing now

PR #3957, opened 2022-03-31
Add I/O mention to banner

parlough:
I wasn't exactly sure what to include for the text on the banner. Let me know if you'd like to see any changes, thanks!


domesticmouse (commented):
Is this going to be too much content for the banner?

parlough:
It is similar to the docs.flutter.dev one and seems to handle different screen sizes just fine, so I think it should be okay.

sfshaza2 (approved):
Looks great!!

PR #3965, opened 2022-04-04
Adjust migration article link to a tip

parlough:
This seems better as a tip than info, especially right next to another info box.


domesticmouse (approved):
PTAL @sfshaza2

sfshaza2 (approved):
lgtm

Issue #3910, opened 2022-03-14
[2.17] Use new `dart` subcommands 


johnpryan:
We should ensure we are using `dart` subcommands if the previous command is deprecated in 2.17: https://github.com/dart-lang/sdk/issues/46100

PR #3973, opened 2022-04-10
Remove now unnecessary dart tool version note

parlough:
This note is no longer necessary with `dart run` being present for a long time and the other commands now being migrated.

Contributes to https://github.com/dart-lang/site-www/issues/3910


mit-mit (approved):
Yeah, this can definitely go now

PR #3978, opened 2022-04-10
Extend `dart fix` tool documentation

parlough:
- Adjusts redirect to point to dart.dev article since the content is repeated in both locations
- Expands on documentation of issues the tool fixes
- Remove mention of Dart 2.12 now that it's been more than a year
- Link to Flutter wiki's [Data driven fixes](https://github.com/flutter/flutter/wiki/Data-driven-Fixes) document which covers the migration type of fixes in detail

**Staged:** https://dart-dev--pr3978-misc-dart-fix-extend-ej3oy6to.web.app/tools/dart-fix

### DIFF
--- a/src/tools/dart-fix.md
+++ b/src/tools/dart-fix.md
@@ -1,15 +1,22 @@
 ---
 title: dart fix
-description: Help for bringing your code up-to-date.
+description: Command-line tool for applying analysis fixes and migrating API usages.
 toc: false
 ---
 
-The `dart fix` command (added in Dart 2.12)
+The `dart fix` command
 finds and fixes two types of issues:
 
-* Analysis issues that have associated automated fixes
+* Analysis issues identified by [`dart analyze`][]
+  that have associated automated fixes
   (sometimes called _quick-fixes_ or _code actions_)
+
+  To learn more about customizing the analysis issues identified,
+  see [Customizing static analysis](/guides/language/analysis-options).
 * Issues that have associated package API migration information
+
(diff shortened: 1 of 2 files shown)


domesticmouse (approved):

src/tools/dart-fix.md
```diff
@@ -1,16 +1,23 @@
...
+* Analysis issues identified by [`dart analyze`][]
+  that have associated automated fixes
   (sometimes called _quick-fixes_ or _code actions_)
+
+  To learn more about customizing the analysis issues identified,
+  see [Customizing static analysis](/guides/language/analysis-options).
 * Issues that have associated package API migration information
 
+  For information about how automated package API changes work,
+  see [Data driven fixes][] on the Flutter wiki.
```
bwilkerson, line 19: The most recent decision that I'm aware of is that we don't want to advertise the data-driven fixes support. We are supporting it for Flutter because they're already using it, but we're not supporting it for other users. Please remove this reference to the documentation.

parlough: Should I not mention that `dart fix` will fix these at all or just not document how they work/how to add them?

bwilkerson: Just don't document the mechanism that makes them work. It's good to mention that it also helps migrate your code to new versions of the Dart and Flutter SDKs.

parlough: Thanks for the context. Adjusting the documentation in https://github.com/dart-lang/site-www/pull/3987

PR #3984, opened 2022-04-11
Update CoC e-mail

Hixie:
I'm just guessing but this seems out of date.


parlough:
I’m not sure a dart.dev version of this email exists. Referring to @mit-mit who may know more.

sfshaza2 (approved):
lgtm

Hixie:
Yeah looks like conduct@dart.dev doesn't go anywhere. We should fix that...

kevmoo:
perhaps we should start by reverting this PR – DNS + email takes a bit to set up 😁

PR #3987, opened 2022-04-13
Adjust mention of data driven fixes in dart fix doc

parlough:
@bwilkerson mentioned in https://github.com/dart-lang/site-www/pull/3978#discussion_r848873454 that the current stance is to not advertise support for data driven fixes and simply mention to users the functionality of helping with Dart and Flutter specific updates.


bwilkerson (approved):
Thanks!

PR #3988, opened 2022-04-16
Remove some unnecessary old tool mentions

parlough:
With the release of Dart 2.17, these commands will have been replaced and removed for enough releases the extra mentions of the old functionality don't add much value and distract from the primary content.

Fixes https://github.com/dart-lang/site-www/issues/3910


sfshaza2 (approved):
lgtm! Thanks, @parlough!

PR #3997, opened 2022-04-21
Add archive support for Windows ARM64 builds

parlough:
This won't work yet as no dev build has been published yet, only mainline builds.

Waiting on the next dev build to be published and making sure the ARM64 toolchain support remains.

### DIFF
--- a/src/get-dart/archive/out/web/download_archive.dart.js
+++ b/src/get-dart/archive/out/web/download_archive.dart.js
@@ -3823,6 +3823,8 @@ Kh(a,b){var s,r=b.length
 if(r===0)return
 if(a===b)throw A.J(A.a4(a))
 for(s=0;s<r;++s)a.push(b[s])},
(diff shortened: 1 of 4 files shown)


mit-mit:
This is not a supported platform, so we should not add this, I think. I'm curious, where did you hear about this?

parlough:
I follow every SDK commit so I had been tracking the Windows ARM64 progress. With the toolchain support added in https://github.com/dart-lang/sdk/commit/8778e322cc813cdaaf1b521bd9e15ced0c5d8cf9 resulting mainline builds now available at https://storage.googleapis.com/dart-archive/channels/be/raw/latest/sdk/dartsdk-windows-arm64-release.zip, I assumed it would be in a similar experimental state to riscv64 with the next dev release.

If we don't want to highlight these builds as available, we can revisit this in the future, just let me know if/when that is desired.

mit-mit:
Got it. Yeah, this is pre-experiment really so not something we're ready to highlight yet.

PR #4022, opened 2022-05-04
Mark actions workflows read only to address code scanning alerts

parlough:
As suggested by this security alert: https://github.com/dart-lang/site-www/security/code-scanning/32

\cc @sealesj

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -75,7 +75,8 @@ jobs:
 
   stage:
     permissions:
-      contents: read
+      checks: write
+      contents: write
       pull-requests: write
     if: '${{ github.event.pull_request.head.repo.full_name == github.repository }}'
     needs: test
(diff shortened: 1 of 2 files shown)


domesticmouse (approved):

domesticmouse:
PTAL @godofredoc

.github/workflows/build.yml
```diff
@@ -71,6 +77,9 @@ jobs:
           channelId: live
 
   stage:
+    permissions:
+      checks: write
```
godofredoc, line 81: How were this permissions determined?

parlough: For `pull-requests: write`, the Firebase action adds comments to PRs as seen by the above comment (this knowledge base also included it https://github.com/step-security/secure-workflows/blob/main/knowledge-base/firebaseextended/action-hosting-deploy/action-security.yml).

As for `checks: write`, the build failed without it, I'm assuming since it adjusts the results of the `build/deploy` job.

godofredoc: Thanks for the details!

godofredoc (approved):

PR #4027, opened 2022-05-08
Remove some JS mentions from the language tour

parlough:
The language tour is not meant just for those that know JavaScript, especially when they can use the [Learning Dart as a JavaScript developer](https://dart.dev/guides/language/coming-from/js-to-dart) article. This PR removes some of those JavaScript mentions to make the documentation more generic for other developers.


domesticmouse (approved):

munificent (commented):
LGTM!

PR #4040, opened 2022-05-16
Add an example of configuring dart fix

mit-mit:


src/tools/dart-fix.md
```diff
@@ -31,4 +28,61 @@ To apply the proposed changes, use the `--apply` flag:
...
+
+  Vector3d(super.x, super.y, this.z);
+}
+```
+
+To enable the dart fix to upgrade existing code to use this feature,
+and to ensure that the analyzer warns you when you later forget to use it,
+configure your `analysis_options.yaml` file like this:
+
+```
```
parlough, line 70:
````suggestion
```yaml
````

File: src/tools/dart-fix.md [hunk 2]
```diff
@@ -31,4 +28,61 @@ To apply the proposed changes, use the `--apply` flag:
...
+linter:
+  rules:
+    - use_super_parameters
+```
+
+You should then see:
+
+```console
```
parlough, line 78:
````suggestion
```terminal
````

File: src/tools/dart-fix.md [hunk 3]
```diff
@@ -31,4 +28,61 @@ To apply the proposed changes, use the `--apply` flag:
...
+mit-macbookpro5:si mit$ dart fix --dry-run
```
parlough, line 79:
```suggestion
$ dart fix --dry-run
```

parlough (approved):
Looks like a great addition, thanks. Just some minor comments:

src/tools/dart-fix.md
```diff
@@ -31,4 +28,61 @@ To apply the proposed changes, use the `--apply` flag:
...
+and to ensure that the analyzer warns you when you later forget to use it,
+configure your `analysis_options.yaml` file like this:
+
+```
+linter:
+  rules:
+    - use_super_parameters
+```
+
+You should then see:
```
parlough, line 76: Maybe be more specific about what the following is doing:

```suggestion
You should then see the following when viewing the proposed changes:
```

src/tools/dart-fix.md
```diff
@@ -31,4 +28,61 @@ To apply the proposed changes, use the `--apply` flag:
 $ dart fix --apply
 ```
 
+## Customizing dart fix
+
+The `dart fix` command applies it's fixes based on the analysis configuration
+of the code it us run on. To learn more about customizing the analysis
+see [Customizing static analysis](/guides/language/analysis-options).
```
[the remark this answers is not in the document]
parlough: I didn't even notice the typo. Thanks!

mit-mit:
Direct staging link: https://dart-dev--pr4040-mit-mit-patch-1-5jjzsagl.web.app/tools/dart-fix

domesticmouse (approved):

src/tools/dart-fix.md
```diff
@@ -31,4 +28,62 @@ To apply the proposed changes, use the `--apply` flag:
 $ dart fix --apply
 ```
 
+## Customizing dart fix
+
+The `dart fix` command applies its fixes
+based on the analysis configuration of the code it runs on.
```
bwilkerson, line 34: While this is accurate, and better than what we had before, it might still be a little vague in that it begs the question of how the "analysis configuration" impacts fixes. (Is "analysis configuration" a concept that we use elsewhere?)

It might be useful to state explicitly that

> The `dart fix` command only applies fixes when there is a problem identified by a diagnostic. If no diagnostic is being reported, then no fix will be applied. Some diagnostics, such as compilation errors, are implicitly enabled, while others, such as lints, must be explicitly enabled in the analysis options file.
>
> You can sometimes increase the number of fixes that can be applied by enabling additional lints. It should be noted, however, that not all diagnostics have associated fixes that can be applied, so enabling new lints might not change the behavior of `dart fix`.

I'd also move lines 35-36 to after the example.

src/tools/dart-fix.md
```diff
@@ -31,4 +28,62 @@ To apply the proposed changes, use the `--apply` flag:
...
+see [Customizing static analysis](/guides/language/analysis-options).
+
+### Example
+
+Imagine you have code like this:
+
+```dart
+class Vector2d {
+  final double x, y;
+  Vector2d(this.x, this.y);
```
bwilkerson, line 45: Consider adding a blank line between members to be consistent with the rest of the example code in this section.

src/tools/dart-fix.md
mit-mit: Thanks, incorporated!

PR #4043, opened 2022-05-17
Add weak references and finalizers section to library tour

### LINKED ISSUE #3903: documentation for weak refs and finalizers
https://github.com/dart-lang/language/blob/master/accepted/future-releases/1847%20-%20FinalizationRegistry/proposal.md

johnpryan:
This adds a section on weak references and finalizers to the library tour. I added it to the `dart:core` section, but we might want to add a `dart:ffi` section section to this page also.

closes #3903

### DIFF
--- a/src/_guides/libraries/library-tour.md
+++ b/src/_guides/libraries/library-tour.md
@@ -1047,11 +1047,10 @@ if the target object can't be subclassed.
 
 A [WeakReference][]
 stores a reference to the target object
-without increasing the reference count
-considered by the garbage collector.
+that does not affect how it is 
+collected by the garbage collector.
 Another option is to use an [Expando][]
-to add properties to an object
-without affecting how it's garbage collected.
+to add properties to an object.
 
 A [Finalizer][] can be used to execute a callback function
 after an object is no longer referenced.
@@ -1065,8 +1064,8 @@ after the object is no longer referenced.
 Also, it can be used to close native resources
 such as a database connection or open files.
 
-To ensure that an object won't be finalized
-or garbage collected too early,
+To ensure that an object won't be
+garbage collected and finalized too early,
 classes can implement the [Finalizable][] interface.
 
 ## dart:async - asynchronous programming
--- a/src/_guides/libraries/library-tour.md
+++ b/src/_guides/libraries/library-tour.md
@@ -1040,7 +1040,7 @@ For more information, see
 Dart is a [garbage-collected][] language,
 which means that any Dart object
 that isn't referenced by at least one other object
-will eventually be disposed by the garbage collector. 
+can be disposed by the garbage collector. 
 This default behavior might not be desirable in
 some scenarios involving native resources or 
 if the target object can't be subclassed.
--- a/src/_guides/libraries/library-tour.md
+++ b/src/_guides/libraries/library-tour.md
@@ -1035,6 +1035,38 @@ For more information, see
 [Exceptions](/guides/language/language-tour#exceptions)
 (in the language tour) and the [Exception API reference.][Exception]
 
+### Weak references and finalizers
+Dart is a [garbage-collected][] language,
+which means that any Dart object
+that isn't referenced by at least one other object
+will eventually be disposed by the garbage collector. 
+This default behavior may not be desireable in
+some scenarios involving native resources or 
+if the target object cannot be sub-classed.
+
+A [WeakReference][]
+stores a reference to the target object
+without increasing the reference count
+considered by the garbage collector.
+Another option is to use an [Expando][]
+to add properties to an object
+without affecting how they are garbage-collected.
+
+A [Finalizer][] can be used to execute a callback function
+after an object is no longer referenced.
+It is not guaranteed to execute this callback, however.
+
+A [NativeFinalizer][]
+provides stronger guarantees
+for interacting with native code using [dart:ffi][].
+Its callback will be called at least once
+after the object is no longer referenced.
+It can be used to close native resources
+such as a database connection or open files.
+
+To ensure that an object won't be finalized
+or garbage collected too early,
+classes can implement the [Finalizable][] interface.
 
 ## dart:async - asynchronous programming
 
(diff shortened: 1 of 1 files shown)


src/_guides/libraries/library-tour.md
```diff
@@ -1035,6 +1035,38 @@ For more information, see
 [Exceptions](/guides/language/language-tour#exceptions)
 (in the language tour) and the [Exception API reference.][Exception]
 
+### Weak references and finalizers
```
parlough, line 1038:
```suggestion
### Weak references and finalizers

```

src/_guides/libraries/library-tour.md
```diff
@@ -1035,6 +1035,38 @@ For more information, see
...
+provides stronger guarantees
+for interacting with native code using [dart:ffi][].
+Its callback will be called at least once
+after the object is no longer referenced.
+It can be used to close native resources
+such as a database connection or open files.
+
+To ensure that an object won't be finalized
+or garbage collected too early,
+classes can implement the [Finalizable][] interface.
```
parlough, line 1070: What do you think about adding a version-note at the end of this section?

parlough:
Did you have @dcharkes take a look at this already? If not, @dcharkes a technical review here would be greatly appreciated! Thanks :)

src/_guides/libraries/library-tour.md
```diff
@@ -1035,6 +1035,38 @@ For more information, see
...
 (in the language tour) and the [Exception API reference.][Exception]
 
+### Weak references and finalizers
+Dart is a [garbage-collected][] language,
+which means that any Dart object
+that isn't referenced by at least one other object
+will eventually be disposed by the garbage collector. 
+This default behavior may not be desireable in
+some scenarios involving native resources or 
+if the target object cannot be sub-classed.
```
parlough, line 1045: According to https://developers.google.com/style/word-list#subclass:

```suggestion
if the target object cannot be subclassed.
```

src/_guides/libraries/library-tour.md
```diff
@@ -1035,6 +1035,38 @@ For more information, see
...
+
+A [WeakReference][]
+stores a reference to the target object
+without increasing the reference count
+considered by the garbage collector.
+Another option is to use an [Expando][]
+to add properties to an object
+without affecting how they are garbage-collected.
```
parlough, line 1053: I think this should only be [hyphenated](https://developers.google.com/style/hyphens?hl=en) when used before a noun:

```suggestion
without affecting how they are garbage collected.
```

src/_guides/libraries/library-tour.md
```diff
@@ -1035,6 +1035,38 @@ For more information, see
...
+
+A [Finalizer][] can be used to execute a callback function
+after an object is no longer referenced.
+It is not guaranteed to execute this callback, however.
```
parlough, line 1057: If we're not going to expand on this yet, perhaps put however in the beginning.

```suggestion
However, it is not guaranteed to execute this callback.
```

src/_guides/libraries/library-tour.md
```diff
@@ -1035,6 +1035,38 @@ For more information, see
...
+provides stronger guarantees
+for interacting with native code using [dart:ffi][].
+Its callback will be called at least once
+after the object is no longer referenced.
+It can be used to close native resources
+such as a database connection or open files.
+
+To ensure that an object won't be finalized
+or garbage collected too early,
+classes can implement the [Finalizable][] interface.
```
parlough, line 1069: I'm not sure if it is the right level of detail for this entry, but as a reader I'm a little confused on what "too early" means. Maybe we should expand more explicitly on what guarantee implementing `Finalizable` provides? Something like being alive the entire scope of the variable (or whatever it guarantees).

src/_guides/libraries/library-tour.md
```diff
@@ -1035,6 +1035,38 @@ For more information, see
...
+
+A [WeakReference][]
+stores a reference to the target object
+without increasing the reference count
+considered by the garbage collector.
+Another option is to use an [Expando][]
+to add properties to an object
+without affecting how they are garbage-collected.
+
+A [Finalizer][] can be used to execute a callback function
```
parlough, line 1055: Also, question for @dcharkes: Is it important enough to reference the callback must not throw in this section? Seems like an important detail to keep in mind even in this high level discussion.

src/_guides/libraries/library-tour.md
parlough, line 1053: Seems the plurality of "an object" is not consistent with "how they are".

src/_guides/libraries/library-tour.md
```diff
@@ -1035,6 +1035,38 @@ For more information, see
 [Exceptions](/guides/language/language-tour#exceptions)
 (in the language tour) and the [Exception API reference.][Exception]
 
+### Weak references and finalizers
+Dart is a [garbage-collected][] language,
+which means that any Dart object
+that isn't referenced by at least one other object
```
parlough, line 1041: Not sure if it's important, but should this include something like "or from the current scope"?

parlough (commented):
Thanks for adding this. I've seen lots of developers excited about these library additions so adding some website documentation is great.

Looks mostly good to me, I just have a few initial considerations and questions.

src/_guides/libraries/library-tour.md
sfshaza2: This looks the same?

src/_guides/libraries/library-tour.md
sfshaza2: Also cannot => can't

src/_guides/libraries/library-tour.md
sfshaza2: +1. Should be:
without affecting how it's garbage collected.

src/_guides/libraries/library-tour.md
```diff
@@ -1035,6 +1035,38 @@ For more information, see
...
+A [Finalizer][] can be used to execute a callback function
+after an object is no longer referenced.
+It is not guaranteed to execute this callback, however.
+
+A [NativeFinalizer][]
+provides stronger guarantees
+for interacting with native code using [dart:ffi][].
+Its callback will be called at least once
+after the object is no longer referenced.
+It can be used to close native resources
```
sfshaza2, line 1064:
```suggestion
for interacting with native code using [dart:ffi][];
its callback is invoked at least once
after the object is no longer referenced.
Also, it can be used to close native resources
```

src/_guides/libraries/library-tour.md
```diff
@@ -1035,6 +1035,38 @@ For more information, see
 [Exceptions](/guides/language/language-tour#exceptions)
 (in the language tour) and the [Exception API reference.][Exception]
 
+### Weak references and finalizers
+Dart is a [garbage-collected][] language,
+which means that any Dart object
+that isn't referenced by at least one other object
+will eventually be disposed by the garbage collector. 
+This default behavior may not be desireable in
```
sfshaza2, line 1043:
```suggestion
This default behavior might not be desirable in
```

sfshaza2 (commented):
some small updates/clarifications

src/_guides/libraries/library-tour.md
parlough: Just adding some white space before the text of the section :)

src/_guides/libraries/library-tour.md
```diff
@@ -1035,6 +1035,38 @@ For more information, see
 [Exceptions](/guides/language/language-tour#exceptions)
 (in the language tour) and the [Exception API reference.][Exception]
 
+### Weak references and finalizers
+Dart is a [garbage-collected][] language,
+which means that any Dart object
+that isn't referenced by at least one other object
+will eventually be disposed by the garbage collector.
```
dcharkes, line 1042: "will eventually be" -> "can be"

(We can dive into a deep hole of definitions here. Integers are objects, but small integers are never disposed. Neither are Strings which happen to be const in the program.)

src/_guides/libraries/library-tour.md
dcharkes, line 1045: can't be subclassed? What is this referring to?

src/_guides/libraries/library-tour.md
```diff
@@ -1035,6 +1035,38 @@ For more information, see
...
+which means that any Dart object
+that isn't referenced by at least one other object
+will eventually be disposed by the garbage collector. 
+This default behavior may not be desireable in
+some scenarios involving native resources or 
+if the target object cannot be sub-classed.
+
+A [WeakReference][]
+stores a reference to the target object
+without increasing the reference count
```
dcharkes, line 1049: We don't do reference counting in Dart.

I believe what you're trying to say is: 

> Being the target of a weak reference does not keep an object from being garbage collected.

https://api.dart.dev/stable/2.17.0/dart-core/WeakReference-class.html

src/_guides/libraries/library-tour.md
```diff
@@ -1035,6 +1035,38 @@ For more information, see
...
+A [NativeFinalizer][]
+provides stronger guarantees
+for interacting with native code using [dart:ffi][].
+Its callback will be called at least once
+after the object is no longer referenced.
+It can be used to close native resources
+such as a database connection or open files.
+
+To ensure that an object won't be finalized
+or garbage collected too early,
```
dcharkes, line 1068: > finalized or garbage collected

garbage collected and finalized (either both happen or neither happen)

dcharkes (approved):
Looks good!

src/_guides/libraries/library-tour.md
johnpryan: I'm aware of some scenarios where an Expando / WeakMap is useful for adding properties to objects. Maybe this should be more specific...

johnpryan: My understanding is that WeakMap is used for tools that work alongside an app

src/_guides/libraries/library-tour.md
johnpryan: I think that's correct, but there's probably more situations where an object would be kept alive. Maybe we should stick with "that isn't referenced"

src/_guides/libraries/library-tour.md
dcharkes: That's a bit of a strange thing to say under `### Weak references and finalizers`. If you want to cover the use cases for Expandos as well, then maybe change the header to include expandos.

src/_guides/libraries/library-tour.md
johnpryan: That's a good point, I added another sentence to clarify.

src/_guides/libraries/library-tour.md
johnpryan: I'll change this to "can't be modified"

src/_guides/libraries/library-tour.md
```diff
@@ -1035,6 +1035,45 @@ For more information, see
...
+
+To ensure that an object won't be
+garbage collected and finalized too early,
+classes can implement the [Finalizable][] interface.
+When a local variable is Finalizable, 
+it won't be garbage collected
+until the code block where it is declared has exited.
+
+{{site.alert.version-note}}
+Support for weak references and finalizers was added in Dart 2.17.
```
sfshaza2, line 1075:
```suggestion
  Support for weak references and finalizers was added in Dart 2.17.
```

sfshaza2 (approved):
LGTM! (Though there is that one question that @parlough is asking of @dcharkes.)

parlough:
Looks good to me, thanks!

PR #4046, opened 2022-05-18
Standardize em dash spacing to match style guide

### LINKED ISSUE #3768: Standardize usage of em dash
### Describe the problem

Currently we are inconsistent in our usage of em-dash(and perhaps other dashes) and not all usages follow the [Google developer documentation style guide](https://developers.google.com/style/dashes).

### Expected fix

We should consistently follow the style guide.

### Additional context

_No response_

parlough:
The Google style guide's section covering [dashes](https://developers.google.com/style/dashes?hl=en) states em dashes should not be preceded or followed by spaces. This standardizes our usage to follow that guideline.

This PR also standardizes to the unicode em dash (`—`) from `&mdash;`. This is because it is easier to read while editing, articles are often copied from word processors like Google Docs which would use the character, but mostly because it is useful to standardize on one choice.

The PR also fixes and slightly modifies some surrounding wordings and formatting issues.

Fixes #3768


domesticmouse (approved):

sfshaza2 (approved):
I guess this is ok, though GitHub displays the emdash such that it looks like a regular dash, until you go to edit the line. I find this confusing. But I 100% agree with removing the spaces surrounding the emdash!!!

PR #4047, opened 2022-05-18
Expand documentation on nullable field promotion issues

### LINKED ISSUE #4037: "if (... is int)" does not "promote to the tested type", as you have written here:
### Page URL

https://dart.dev/null-safety/understanding-null-safety

### Page source

_No response_

### Describe the problem

Before null-safety, I had a Map with "pings" that would count down for different players in a game. As players would come and go, their ping would be added and removed from the Map, but the countdown might try to count down after an entry was removed, so I added an "if" statement which worked fine.

But after upgrading to null-safety, I can't get the "--" operator to work! According to this page, https://dart.dev/null-safety/understanding-null-safety , I should be able to write "if (x is int) {}", and inside the body of that "if", x should be treated as an actual int. But it doesn't work in my case! Look:

```dart
    while (pingCountDownMap.containsKey(follower) && this.mounted) {
      await Future.delayed(Duration(seconds: 1));
      if (pingCountDownMap.containsKey(follower) && pingCountDownMap[follower] != null) {
        // Original:
        // pingCountDownMap[follower]--;
        
        // Monkey-patch:
        // int localPing = pingCountDownMap[follower] as int;
        // localPing--;
        // pingCountDownMap[follower] = localPing;
        
        // Should work, according to documents, but doesn't:
        if (pingCountDownMap[follower] is int) {
          pingCountDownMap[follower]--;
        }
    }
```

Shouldn't that last "if" statement work? Because it doesn't. I get the compile error that "--" can't be used on a value that might be null.

Is there any other check or correction that I could do so that I can still use the "--" operator, without having to go through that monkey-patch I show above?

### Expected fix

Please tell me how to use the "--" operator on an int in a Map<String, int>, where the int in question may have been removed at the time the "--" operator is supposed to fire.

And please update the page in the provided link to include this situation.

### Additional context

_No response_

parlough:
Fixes #4037

### DIFF
--- a/src/null-safety/understanding-null-safety/index.md
+++ b/src/null-safety/understanding-null-safety/index.md
@@ -647,8 +647,8 @@ the non-nullable side so you can call methods on it. Flow analysis is one of the
 primary ways to do this for local variables and parameters. We've extended type
 promotion to also look at `== null` and `!= null` expressions.
 
-If you check a variable with nullable type to see if it is not `null`, Dart then
-promotes the variable to the underlying non-nullable type:
+If you check a local variable with nullable type to see if it is not `null`, 
+Dart then promotes the variable to the underlying non-nullable type:
 
 ```dart
 // Using null safety:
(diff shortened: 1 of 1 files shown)


sfshaza2 (approved):
lgtm

PR #4049, opened 2022-05-18
Fix accidental en dashes which should be hyphen-minus

parlough:
If someone copied these into Dart code they wouldn't have worked. I'm surprised no one noticed before.


domesticmouse (approved):

sfshaza2 (approved):
lgtm

PR #4053, opened 2022-05-19
Adjust homepage Try Dart styles to avoid distracting white bar

parlough:
The "Try Dart" navigation element is meant to navigate to a portion of the page which fills the viewport and looks dedicated to trying Dart in the embedded DartPad. Currently though there is a distracting white bar on the bottom due to the margin added to enable this sizing (https://github.com/dart-lang/site-www/pull/2232). 

This adjusts the row it is in to have a matching background color. I tried switching the margin to padding instead but due to the desire to fill the viewport, this reduces the space available to DartPad. 

The images make the change hard to visualize, I recommend viewing the staged #try-dart on a browser window wide enough to show the navbar items.

**Previously:**

Previous Try Dart portion with white bar

**After:**

New Try Dart portion without white bar

### DIFF
--- a/src/_sass/_dash.scss
+++ b/src/_sass/_dash.scss
@@ -298,6 +298,10 @@ $dash-dartpad-border: #293542;
   }
 }
 
+.dash-dartpad-row {
+  background-color: $dash-callout;
+}
+
 .dash-dartpad {
   flex: 1;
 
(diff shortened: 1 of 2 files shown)


johnpryan:
Could we include this in https://github.com/dart-lang/site-www/pull/4054 ?

Issue #292, opened 2017-06-02
Apps shouldn't HAVE to commit pubspec.lock


kwalrath:
https://www.dartlang.org/guides/libraries/private-files says that apps should commit their pubspec.lock files, but for example apps (and maybe others), that's not necessarily a good idea.

We should probably say it's OK to commit it.

See the discussion of https://github.com/dart-lang/one-hour-codelab/pull/41.

kwalrath:
Any objection to this change, @kevmoo?

kwalrath:
It seems this is still a bit controversial:

#2207
https://github.com/dart-lang/site-www/pull/2332/files/b55ada735c7bcea6c16a0c9385460e74a9d224d8#diff-56ed1d9a5e898e17130ab3002180103eR437
...

kwalrath:
I'm going to close this, because example apps are an exception to the rule

PR #4058, opened 2022-05-23
Make the reason of commit .lock file clearer

aboutmydreams:
It can't `ensures that everyone working on the app uses the exact same versions` because the member of the team are working in different branches, and when they add some new libs or run `flutter clean && flutter pub get` , `pubspec.lock` file will update.

I recommend that you commit the`pubspec.lock` file just because it will help you locate problems in production or development due to dependency changes.

And some old discussions:
https://github.com/dart-archive/one-hour-codelab/pull/41
https://github.com/dart-lang/site-www/pull/2332
https://github.com/dart-lang/site-www/issues/292

### DIFF
--- a/src/_guides/libraries/private-files.md
+++ b/src/_guides/libraries/private-files.md
@@ -103,7 +103,10 @@ similar to Ruby's `Gemfile.lock`.
 Regenerating the `pubspec.lock` file lets you test your package
 against the latest compatible versions of its dependencies.
 
-**For application packages**, we recommend that you commit
-the `pubspec.lock` file.
-Saving `pubspec.lock` will track the dependencies in production or
-development, it can help you locate problems due to dependency changes.
+**For application packages**, 
+we recommend that you commit the `pubspec.lock` file.
+Versioning the `pubspeck.lock` file
+ensures changes to transitive dependencies are explicit.
+Each time the dependencies change due to `dart pub upgrade`
+or a change in `pubspec.yaml` 
+the difference will be apparent in the lock file.
--- a/src/_guides/libraries/private-files.md
+++ b/src/_guides/libraries/private-files.md
@@ -103,7 +103,7 @@ similar to Ruby's `Gemfile.lock`.
 Regenerating the `pubspec.lock` file lets you test your package
 against the latest compatible versions of its dependencies.
 
-**For application packages**, we recommend that you
-**commit** the `pubspec.lock` file.
-Saving `pubspec.lock` ensures that everyone working on the app
-uses the exact same versions.
+**For application packages**, we recommend that you commit
+the `pubspec.lock` file.
+Saving `pubspec.lock` will track the dependencies in production or
+development, it can help you locate problems due to dependency changes.


src/_guides/libraries/private-files.md
```diff
@@ -103,7 +103,7 @@ similar to Ruby's `Gemfile.lock`.
...
 against the latest compatible versions of its dependencies.
 
-**For application packages**, we recommend that you
-**commit** the `pubspec.lock` file.
-Saving `pubspec.lock` ensures that everyone working on the app
-uses the exact same versions.
+**For application packages**, we recommend that you commit
+the `pubspec.lock` file.
+Saving `pubspec.lock` will track the dependencies in production or
+development, it can help you locate problems due to dependency changes.
```
parlough, line 109: Is your concern that the previous statement accidentally suggested `pubspec.lock` won't ever be affected by `pub get` and we just need to be more explicit about the benefits of checking in the lockfile?

How does something like the following sound?

```suggestion
**For application packages**, we recommend that you
**commit** the `pubspec.lock` file.
Saving `pubspec.lock` ensures dependency updates are explicit
and don't unexpectedly change during the development process
or when working with other developers.
```

kevmoo:
@natebosch @devoncarew @mit-mit @lrhn – thoughts here?

aboutmydreams:
@parlough Glad you can reply! Maybe we need a more precise wording to reduce the possibility of misunderstanding. I listed some common situations for developers:

When build for application(after `flutter pub get`, dependencies will be different from other developers, and resulting in an update of the .lock file)  If there is a problem which due to the update of the dependency version, he might think it's caused by other collaborators, but it's just because he updated the local cache.

so,  it is can't ensure that the versions of packages cached locally by different developers are consistent.

For developers, there are more common situations：
- when switch application repo branch locally. The two branch .lock files are the same, but when execute the switch branch command, the dart plugin of vscode automatically helps execute `flutter pub get` 
- when developers add a new package in his application, they will run `flutter pub get` (maybe run automatically by dart plugin). And then, the lock file changed and the dependencies will be different from others.

src/_guides/libraries/private-files.md
```diff
@@ -103,7 +103,7 @@ similar to Ruby's `Gemfile.lock`.
 Regenerating the `pubspec.lock` file lets you test your package
 against the latest compatible versions of its dependencies.
 
-**For application packages**, we recommend that you
-**commit** the `pubspec.lock` file.
-Saving `pubspec.lock` ensures that everyone working on the app
-uses the exact same versions.
+**For application packages**, we recommend that you commit
+the `pubspec.lock` file.
+Saving `pubspec.lock` will track the dependencies in production or
```
mit-mit, line 108: The word "saving" here is confusing (it sounds like save as in File>Save, but I think you mean something different?)

mit-mit: Did you mean "versioning"?

mit-mit:
@aboutmydreams regarding:

When build for application(after flutter pub get, dependencies will be different from other developers, and resulting in an update of the .lock file) 

When `dart pub get` / `flutter pub get` runs, and there is both a `pubspec.yaml` and `pubspec.lock` file, pub will check if there is a valid resolution of the constraint in `pubspec.yaml` that matches the versions listed in `pubspec.lock`. If that is the case, the `pubspec.lock` file will not be updated. If that is not what you are seeing, please file a bug with detailed repro steps.

src/_guides/libraries/private-files.md
lrhn: I never understood the `.lock` file to begin with, so bear with me.

Is the idea that for an *application* package, it's important to be able to recreate the the released code.
That means knowing the precise versions of all the packages that are included in the final program, and to that end, you can commit the `pubspec.lock` file used to build the release version as part of the versioning of your development environment.

If so, you don't need to commit `pubspec.lock` on your development branch, but can, if you want to make sure you have a stable environment for everybody working on the code. 

You can get the same effect by precisely versioning your dependencies in the pubspec.yaml, so `test: 1.16.2` instead of a allowing a range like `test: ^1.10`. The advantage of using the `.lock` file is that you can use tools to update it when you want it to be updated, where you have to manually modify `pubspec.yaml`.

Am I understanding it correctly?

natebosch: > You can get the same effect by precisely versioning your dependencies in the pubspec.yaml,

You'd need to use `dependency_overrides` to pin a version for transitive dependencies. This is not a pattern we recommend.

I like the framing that using version control makes changes to dependencies explicit.

"Using version control on `pubspec.lock` makes changes in transitive dependencies explicit. Each time the dependencies change from a `dart pub upgrade` or a change in `pubspec.yaml` the difference will apparent in the lock file."

I don't know if it's important for us to say why an author may want dependencies to change or not change - that feels outside the scope here.

natebosch:
> When build for application(after `flutter pub get`, dependencies will be different from other developers, and resulting in an update of the .lock file

I don't think we expect a `flutter pub get` to change the lock file, unless the `pubspec.yaml` has also changed.

src/_guides/libraries/private-files.md
parlough: Sounds good to me, thanks! Something like the following:

```suggestion
**For application packages**, 
we recommend that you commit the `pubspec.lock` file.
Versioning the `pubspeck.lock` file
ensures changes to transitive dependencies are explicit.
Each time the dependencies change due to `dart pub upgrade`
or a change in `pubspec.yaml` 
the difference will be apparent in the lock file.
```

src/_guides/libraries/private-files.md
aboutmydreams: oh, yes, my mistake. Thanks for pointing out and I will fix it.

src/_guides/libraries/private-files.md
aboutmydreams: Refine the description about the suggestion

src/_guides/libraries/private-files.md
```diff
@@ -103,7 +103,10 @@ similar to Ruby's `Gemfile.lock`.
...
-**commit** the `pubspec.lock` file.
-Saving `pubspec.lock` ensures that everyone working on the app
-uses the exact same versions.
+**For application packages**, 
+we recommend that you commit the `pubspec.lock` file.
+Versioning the `pubspeck.lock` file
+ensures changes to transitive dependencies are explicit.
+Each time the dependencies change due to `dart pub upgrade`
+or a change in `pubspec.yaml` 
+the difference will be apparent in the lock file.
```
aboutmydreams, line 112: Thanks to everyone, this is the change to this pr now. suggestion form @parlough

aboutmydreams:
Thanks @mit-mit and @natebosch 
I try to reproduce the problem in https://github.com/aboutmydreams/flutter_pub_get_test/pull/1 
This may not be in the scope of this pr's discussion, but thanks for your help again if you can.

I hit this button sometime when switching branches, but then I reset the code to keep the .lock file consistent with the remote unupdated branch code (which might give some clues)
image

mit-mit:
> This may not be in the scope of this pr's discussion

Yes, please file a separate issue in the SDK repo if you have a repro for this issue.

parlough (approved):
Thanks for this PR and the follow up discussion. I think with everyone’s help we have gotten to a description that provides a more clear context and reasoning. If anyone would like to see further changes, feel free to open a new issue or PR.

Thanks again!

PR #4063, opened 2022-05-23
Cookie Notice

btobin:
- Adds cookie notice / consent (html, scss, js, external lib)
- Cleans up `main.js` file
- Removes `--offline` flag from `pub get` in Dockerfile so fresh install does not fail


src/assets/js/main.js
```diff
@@ -187,27 +77,158 @@ function initVideoModal() {
...
+    banner = $(".banner"),
+    bb = banner.length > 0 ? banner[0].getBoundingClientRect().bottom : hh,
+    headerHeight = Math.max(hh, bb),
+    h = window.innerHeight,
+    // space between scroll position and top of the footer
+    whenAtBottom = f - t,
+    mh = Math.min(h, whenAtBottom) - hh;
+  $("#sidenav").css({ top: headerHeight, maxHeight: mh });
+  $("#site-toc--side").css({ top: headerHeight, maxHeight: mh });
+}
```
johnpryan, line 93: What changes did you make to this file? There's a lot of changed lines due to the functions being moved around.

src/_includes/cookie-notice.html
```diff
@@ -0,0 +1,10 @@
+<section class="cookie-notice">
+  <div class="container">
+    <p>Google uses cookies to deliver its services, to personalize ads, and to 
+      analyze traffic. You can adjust your privacy controls anytime in your 
+      <a href="https://myaccount.google.com/data-and-personalization" target="_blank" rel="noopener">Google settings</a>. 
+      <a href="https://policies.google.com/technologies/cookies" target="_blank" rel="noopener">Learn more</a>.
```
parlough, line 6: @sfshaza2  How about something like "To learn more, see [How Google uses cookies](https://policies.google.com/technologies/cookies)."

parlough, line 6: I don't think we need the external icons in these links:

```suggestion
      <a href="https://myaccount.google.com/data-and-personalization" target="_blank" rel="noopener" class="no-automatic-external">Google settings</a>. 
      <a href="https://policies.google.com/technologies/cookies" target="_blank" rel="noopener" class="no-automatic-external">Learn more</a>.
```

src/_sass/components/_cookie-notice.scss
```diff
@@ -0,0 +1,63 @@
...
+  .container {
+    p {
+      color: white;
+      
+      a {
+        color: #40C4FF;
+      }
+    }
+  }
+}
```
parlough, line 63: Add a new line at the end:

```suggestion
}

```

src/_sass/components/_cookie-notice.scss
```diff
@@ -0,0 +1,63 @@
...
+      font-size: 1rem;
+      line-height: 1.6;
+      margin-right: 2rem;
+      margin-bottom: 0;
+    }
+  }
+}
+
+body.homepage .cookie-notice {
+  background-color: #1C2834;
```
parlough, line 52: You should be able to import the variables file and use the `$site-color-dark-background` variable so we don't lose track.

```suggestion
  background-color: $site-color-dark-background;
```

src/_sass/components/_cookie-notice.scss
```diff
@@ -0,0 +1,63 @@
...
+
+  .container {
+    p {
+      color: white;
+      
+      a {
+        color: #40C4FF;
```
parlough, line 59: Same here:

```suggestion
        color: $site-color-card-link;
```

parlough (changes requested):
Thanks for working on these changes! Some comments:

On the homepage there is not focus/hover/active style for the links. Perhaps you can match what the banner uses? I think it is `darken($site-color-card-link, 20%);` for those states currently. 

Also there seems to be no way to focus the "Okay" button. Can you make it keyboard interact-able?

src/assets/js/main.js
btobin: Hey @johnpryan. I felt like it was a good time to clean this file up and make it have a consistent style, something we left out of the last cleanup. I did the following:

- I consolidated two document ready jQuery functions into 1 i.e. `$(function() { ... })`
- I moved all inner functions outside of that scope, like the majority of them
- I moved loose code into the ready function
- I moved 2 vars inside the only function that uses them
- I put the jQuery ready function at the bottom of the file
- Added the cookie function

I did not change any logic inside existing functions.

src/_includes/cookie-notice.html
btobin: It was requested by @RedBrogdon that we carry over the same cookie notice module from Flutter.dev, which has the approved messaging and linkage from legal. Happy to change messaging if that is the consensus. 

I can add aria, etc to the button. I changed the styles of the banner on the homepage only so that it matches those particular styles, so it makes sense to be complete here. I'll add that.

src/assets/js/main.js
johnpryan: Thanks, LGTM

johnpryan (approved):

src/_includes/cookie-notice.html
parlough: If that is the message from Andrew and legal, then that is perfectly fine! Removing the external link icons will make it much more manageable.

src/assets/js/main.js
parlough: Thanks for the clarification and the cleanup!

johnpryan (approved):

src/_includes/cookie-notice.html
btobin: @parlough I have pushed up those style/accessibility changes.

src/_sass/components/_cookie-notice.scss
```diff
@@ -0,0 +1,67 @@
...
+  background-color: $site-color-dark-background;
+
+  .container {
+    p {
+      color: white;
+      
+      a {
+        color: $site-color-card-link;
+
+        &:hover {
```
parlough, line 61:
```suggestion
        &:hover, &:focus, &:active {
```

src/_includes/cookie-notice.html
```diff
@@ -0,0 +1,11 @@
+<section class="cookie-notice">
+  <div class="container">
+    <p>Google uses cookies to deliver its services, to personalize ads, and to 
+      analyze traffic. You can adjust your privacy controls anytime in your 
+      <a href="https://myaccount.google.com/data-and-personalization" target="_blank" rel="noopener" class="no-automatic-external">Google settings</a>. 
+      <a href="https://policies.google.com/technologies/cookies" target="_blank" rel="noopener" class="no-automatic-external">Learn more</a>.
+    </p>
+    <a id="cookie-consent" class="btn btn-primary"
+       tabindex="1" role="button">Okay</a>
```
parlough, line 9: I believe this should be "0" to be focusable, but it still will not be usable by a keyboard user due to lacking an "href" attribute. 

Is it possible to use a normal `button` which will be inherently interactive.

btobin: No problem. Button is great, that is what we have on flutter.dev. Changed it for docs, but you are right though. And you do have the compatible button styles so no issue on the dart side (changed, but not committed). 

A tabindex of 1 gives it first tab focus, and then the focus heads back to the header. The issue I am seeing is that Dartpad steals the focus permanently (in Chrome). I can also add this banner to the top of the html body so it's in a more ideal stacking order. I think it makes sense for it to get immediate focus though so it can be removed. 

Let me know! Thanks.

parlough: I like the idea of it getting immediate focus, but it feels weird not being able to directly tab to the links with more information directly from it. That might get confusing.

I think either keep it last (so people can access it with shift-tab) or move the cookie notice to the top of the html body but keep it fixed to the bottom so all those links are at the beginning of the focus order together. Whichever you prefer is fine.

btobin: I just tested on Flutter docs, moving the notice to the top of the `<body />` works great without setting a tabindex. That's much nicer for a screen reader as well.

parlough: Sounds great to me!

parlough (approved):
Looks good to me! Thanks for adding this and making all these adjustments!

PR #4074, opened 2022-05-30
Add an explicit example of iterating over an iterable to JS article

### LINKED ISSUE #4038: Suggestion: 'Learning Dart as a JavaScript developer' should mention `for` loops
### Page URL

https://dart.dev/guides/language/coming-from/js-to-dart

### Page source

https://github.com/dart-lang/site-www/tree/main/src/_guides/language/coming-from/js-to-dart.md

### Describe the problem

I suspect that one of the reasons why I see `Iterable.forEach` used so much in Dart questions on StackOverflow is because many people have been institutionalized by JavaScript where `for ... in` unintuitively iterates over an object's properties.  JavaScript developers who want to avoid potential indexing/bounds errors therefore are recommended to use `Array.forEach` instead.

Dart doesn't have that problem, so `for ... in` should be the preferred way to loop.  Maybe this would be worth mentioning?

### Expected fix

_No response_

### Additional context

_No response_

parlough:
Fixes #4038

### DIFF
--- a/src/_guides/language/coming-from/js-to-dart.md
+++ b/src/_guides/language/coming-from/js-to-dart.md
@@ -883,6 +883,35 @@ although some have additional uses when it comes
 to collections
 (discussed in more detail in the [Collections](#collections) section).
 
+#### Loops
+
+While both JavaScript and Dart have `for-in` loops,
+their behavior is different.
+
+JavaScript's `for-in` loop iterates over an object's properties,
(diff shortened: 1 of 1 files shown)


domesticmouse (approved):

sfshaza2 (approved):
lgtm!  Thanks, @parlough!!

PR #4090, opened 2022-06-07
Document multiple dependencies in pub-add

sigurdm:
Fix of: https://github.com/dart-lang/pub/issues/3450

### DIFF
--- a/src/tools/pub/cmd/pub-add.md
+++ b/src/tools/pub/cmd/pub-add.md
@@ -9,11 +9,12 @@ _Add_ is one of the commands of the [pub tool](/tools/pub/cmd).
 $ dart pub add <package>[:<constraint>] [<package2>[:<constraint2>]... ] [options]
 ```
 
-This command adds the specified packages to the pubspec as dependencies, and
-then gets the dependencies.
+This command adds the specified packages to the pubspec as dependencies, 
+and then gets the dependencies.
 
-For example, the following command is equivalent to editing `pubspec.yaml` to
-add the `http` package, and then calling `dart pub get`:
+For example, the following command is equivalent to
+editing `pubspec.yaml` to add the `http` package, 
+and then calling `dart pub get`:
 
 ```terminal
 $ dart pub add http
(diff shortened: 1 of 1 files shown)


sigurdm:
Yeah - really we should fix it. It feels so stupid to write it out...

Added an explanation.

src/tools/pub/cmd/pub-add.md
```diff
@@ -38,6 +38,10 @@ $ dart pub add --dev test
 For options that apply to all pub commands, see
 [Global options](/tools/pub/cmd#global-options).
 
+The options given will apply to all the packages added in an invocation. For
+example `dart pub add test http --dev` will add both package:test and
+package:http as dev-dependencies.
```
parlough, line 43: Some minor adjustments to be consistent with the rest of the page and make it a note to better fit with the flow of the page:

```suggestion
{{site.alert.note}}
  Any specified options will apply to all the packages
  included in an invocation of the command.
  For example, `dart pub add test http --dev` 
  will add both the `test` and `http` packages 
  as dev dependencies.
{{site.alert.end}}
```

src/tools/pub/cmd/pub-add.md
```diff
@@ -6,14 +6,14 @@ description: Use dart pub add to add a dependency.
...
-This command adds the specified package to the pubspec as a dependency,
-and then gets the dependency.
-For example, the following command is equivalent to
-editing `pubspec.yaml` to add the `http` package,
-and then calling `dart pub get`:
+This command adds the specified packages to the pubspec as dependencies, and
+then gets the dependencies.
+
+For example, the following command is equivalent to editing `pubspec.yaml` to
+add the `http` package, and then calling `dart pub get`:
```
parlough, line 16: Adjusting the semantic breaks:

```suggestion
This command adds the specified packages to the pubspec as dependencies, 
and then gets the dependencies.

For example, the following command is equivalent to
editing `pubspec.yaml` to add the `http` package, 
and then calling `dart pub get`:
```

parlough (approved):
Some minor formatting adjustments, but looks good to me. Thanks!

PR #4092, opened 2022-06-07
Pin ruby dockerfile to a hash, also pin versions of firebase-tools

drewroengoogle:
This PR pins the dockerfile ruby image hash, along with explicitly declaring the version of firebase-tools in order to ensure consistent builds and increase security.
This PR resolves a couple of security vulnerabilities:
* https://github.com/dart-lang/site-www/security/code-scanning/16
* https://github.com/dart-lang/site-www/security/code-scanning/46
* https://github.com/dart-lang/site-www/security/code-scanning/45


parlough (approved):
I tested the full docker setup locally and this works great, thanks for pinning these!

PR #4093, opened 2022-06-07
Drop reference to (long deprecated) dump info visualizer

kevmoo:
Replace with link to tools in the SDK


kevmoo:
@devoncarew @sigmundch – added a go link we can use other places!

PR #4094, opened 2022-06-08
Update create-library-packages.md

cnmade:
To tell user how to create a package via dart create command

### DIFF
--- a/src/_guides/libraries/create-library-packages.md
+++ b/src/_guides/libraries/create-library-packages.md
@@ -9,6 +9,9 @@ This page tells you how to create a package,
 with a focus on the most common kind of package,
 [_library_ packages](/tools/pub/glossary#library-package).
 
+Create library package is easy to run `dart create -t package [YOUR_PACKAGE_NAME]`,
+For more details you can read [dart create](/tools/dart-create)
+
 {% comment %}
 TODO: Add coverage of packages that contain tools.
 {% endcomment %}


src/_guides/libraries/create-library-packages.md
```diff
@@ -9,6 +9,9 @@ This page tells you how to create a package,
 with a focus on the most common kind of package,
 [_library_ packages](/tools/pub/glossary#library-package).
 
+Create library package is easy to run `dart create -t package [YOUR_PACKAGE_NAME]`,
+For more details you can read [dart create](/tools/dart-create)
+
```
parlough, line 14: Overall, I think it's a good idea to add something like this, thanks so much! Some questions and suggestions:

It seems a bit awkward to introduce this in the same section as the opening for the page. What do you think about creating a new starting section?

````suggestion

## Creating a new package

To create the initial project and structure for a package,
use the [`dart create`](/tools/dart-create) command
and the `package` template:

```terminal
$ dart create -t package <PACKAGE_NAME>
```

````

I also included some other suggestions/adjustments:

- Switch the command to a standalone codelab as they can be long and look confusing within a block of text
- Use arrow brackets as directory/package_name is a required argument (compared to optional which would use square brackets)
- Some grammar adjustments

cnmade:
I am ok , any improve should be good .

parlough (approved):
Looks good to me. Thanks for thinking of adding this!

Issue #3093, opened 2021-03-22
Make updates null safety codelab


kwalrath:
#2983 created the null safety codelab. Here's how we should follow up, in rough order of most urgent/easy to less urgent/easy:

- [x] Add /codelabs/null-safety to the list of codelabs in /codelabs (#3112)
- [x] Listof -> ListOf: https://github.com/dart-lang/site-www/pull/2983#discussion_r598166849 (#3112)
- [x] Add link to this bug from the null safety codelab, and perhaps explicitly mention features that aren't covered (yet) (#3112)
- [x] Retire the [null safety tour](https://dart.dev/null-safety/tour), in favor of the codelab (#3112)
- [x] Redo the late variable example: https://github.com/dart-lang/site-www/pull/3167
- [ ] Add hints & tests to the examples
- [ ] Add coverage of other features, such as `required`, `?.`, and `?..`

parlough:
If we want to keep the codelab, they are still relevant.

PR #4103, opened 2022-06-15
Update index.md, cheatsheet is more important to understand to deal with null check.

cnmade:


parlough:
Thanks for opening a pull request!

I gave it some thought and I'm not sure we want to link to this portion of the cheatsheet as it doesn't cover all of the null-aware operators and is pretty barebones, not really a place to use as a documentation destination. 

Do you think linking to somewhere covering the null aware operators quicker is important? Perhaps we could open a new issue to add a new section to the null safety codelab covering more of them? Then the codelab would handle that case as well.

parlough:
Thanks for opening a pull request!

I gave it some thought and I'm not sure we want to link to this portion of the cheatsheet as it doesn't cover all of the null-aware operators and is pretty barebones, not really a place to use as a documentation destination. 

Do you think linking to somewhere covering the null aware operators quicker is important? Perhaps we could open a new issue to add a new section to the null safety codelab covering more of them? Then the codelab would handle that case as well.

cnmade:
The problem is the sequence to read the document. 

codelab is the less read , i have never read it before. so i have hard problem to understand with Null handle.
 
`null-safety/index.md` This document what user may first read, but it did not  tell us really best practice deal with Null .

I got very hard painful with Null. I  don't know how to do with it .


Until they know Null aware operators. 

The life be easy to understand, coding is peaceful.

cnmade:
image
Tell me please, tell me how to deal with Null.

Like the document of Null Aware Operators.

I stuck on dig out how to handle Null, for spent  serval month.

May because me stupid. or the document does not tell me full story.

parlough:
I see your concern, thanks for explaining further.

The original goal of these articles was to discuss the introduction of the null safety feature, not the pre-existing operators, so I agree it could be highlighted a bit better.

The [Working with nullable types](https://dart.dev/null-safety/understanding-null-safety#working-with-nullable-types) sections covers dealing with nullable values with those operators and other solutions in detail. Would linking directly to this section somewhere help?

I've also modified https://github.com/dart-lang/site-www/issues/3093 to track adding the null-aware operators to the codelab.

cnmade:
@parlough as you suggest, i guess we can put 
1. the cheatsheet may be a quick start
2. if user has some experience, and want know more ,they can read detail explain.

parlough:
I'm still not confident the cheatsheet is a great resource for this, as the related sections aren't currently grouped together and don't give a lot of information.

I'm going to go ahead and add a new section to the null safety codelab today or tomorrow. You can let me know if that is sufficient and we can then keep the current links.

cnmade:
The link `codelab` add to `null-safety/index.md` just like a quick link. make clear how to use Null aware operator, is good start for user want to deal with code.

```
talk is cheap, show me the code
```

Code is better self explain than just document.

parlough:
I agree @cnmade, thanks for pushing this forward!

I don't want to link to too many resources here, especially since there is no grouping of related topics on the cheatsheet codelab (and it's not really a resource with headers meant to be linked to). So it's better to standardize on the null-safety codelab which already covers related topics.

I'll add two new sections to the null-safety codelab covering `??`, `??=`, and `?.` (hopefully will open a PR tomorrow) that are shorter with code examples just like the cheatsheet has. I'm tracking this work in https://github.com/dart-lang/site-www/issues/4114 and will link the PR to that issue. Let me know if that doesn't work for you.

Thanks again :)

PR #4110, opened 2022-06-21
Update design.md

frmrgr:
Mention `_` in https://dart.dev/guides/language/effective-dart/design#do-annotate-with-dynamic-instead-of-letting-inference-fail


src/_guides/language/effective-dart/design.md
```diff
@@ -1518,6 +1518,8 @@ mergeJson(original, changes) => ...
 
 Note that it's OK to omit the type when Dart *successfully* infers `dynamic`.
 
+An exception to this is skipping the type annotation on an unused symbol (`_`), such as on an argument in a lambda callback.
```
sfshaza2, line 1521:
```suggestion
An exception to this is skipping the type annotation on an unused symbol (`_`),
such as on an argument in a lambda callback.
```

sfshaza2 (commented):
I, too, defer to @munificent here, but I did want to break that overly long line into 2 lines.

munificent:
It's good to call this out as an exception. Instead of "argument on a lambda callback", I would just say "an unused parameter".

src/_guides/language/effective-dart/design.md
```diff
@@ -1548,6 +1548,7 @@ code did not specify one.
   safety and performance of static types.
 {{site.alert.end}}
 
+**Exception**: type annotation on an unused parameter (`_`) can be skipped.
```
parlough, line 1551:
```suggestion
**Exception**: Type annotations on unused parameters (`_`) can be omitted.
```

frmrgr:
Rephrased and moved per suggestins.

parlough (approved):
Thanks so much for being catching this, being patient, and adjusting these changes. Looks great!

PR #4137, opened 2022-07-11
Formalize .dart_tool/

jonasfj:
I don't have the best words, so maybe there is a way to write this more concisely, or maybe there is a better place we document this.

I just think it's important that we do document it, even if it's mostly for ourselves.

### DIFF
--- a/src/tools/pub/package-layout.md
+++ b/src/tools/pub/package-layout.md
@@ -514,9 +514,12 @@ When running these build steps, files are cached in `.dart_tool/build/`.
 This helps speed-up future re-runs of the build steps.
 
 {{site.alert.warning}}
-When developing a tool that wants to cache files in `.dart_tool/`, ensure that:
-  * You are using a subdirectory named after a package you own (`.dart_tool/<my_tool_package_name>/`).
-  * Never store files that should be under source control in `.dart_tool/`.
+  When developing a tool that wants to cache files in `.dart_tool/`,
+  ensure the following:
+    * You are using a subdirectory named after a package you own
+       (`.dart_tool/<my_tool_package_name>/`)
+    * Your files don't belong under source control, 
+       as `.dart_tools` is generally listed in `.gitignore`
 {{site.alert.warning}}
 
 
--- a/src/tools/pub/package-layout.md
+++ b/src/tools/pub/package-layout.md
@@ -84,12 +84,13 @@ Every package has a [_pubspec_](/tools/pub/pubspec), a file named
 `pubspec.yaml`, in the root directory of the package. That's what *makes* it a
 package.
 
-Running [`pub get`](/tools/pub/cmd/pub-get),
-[`pub upgrade`](/tools/pub/cmd/pub-upgrade), or
-[`pub downgrade`](/tools/pub/cmd/pub-downgrade) on the package creates a
-**lockfile**, named `pubspec.lock`. If your package is an [application
-package](/tools/pub/glossary#application-package), check the lockfile into source
-control. Otherwise, don't.
+Running [`dart pub get`](/tools/pub/cmd/pub-get),
+[`dart pub upgrade`](/tools/pub/cmd/pub-upgrade), or
+[`dart pub downgrade`](/tools/pub/cmd/pub-downgrade) on the package
+creates a **lockfile**, named `pubspec.lock`. 
+If your package is an 
+[application package](/tools/pub/glossary#application-package), 
+check the lockfile into source control. Otherwise, don't.
 
 For more information, see the [pubspec page](/tools/pub/pubspec).
 
--- a/src/tools/pub/package-layout.md
+++ b/src/tools/pub/package-layout.md
@@ -480,20 +480,33 @@ If you have any of these, place them in a directory called `tool`.
   Instead keep `.dart_tool/` in `.gitignore`.
 {{site.alert.end}}
 
-The `.dart_tool/` is created when you've run `dart pub get`.
-The directory used by various tooling for caching files specific to your project and/or local machine.
-Thus, the `.dart_tool/` directory should never be checked into source control, or copied between machines.
+The `.dart_tool/` directory is created when you run `dart pub get`.
+The directory is used by various tools for caching files specific to your project and/or local machine.
+The `.dart_tool/` directory should never be checked into source control, or copied between machines.
 
 It is also generally safe to delete the `.dart_tool/` directory, though some tools might need recompute the cached information before they work again. 
 
-**Example:** The [dart pub get](https://dart.dev/tools/pub/cmd/pub-get) tool will dowload and extract dependencies to a global `$PUB_CACHE` directory, and then write a `.dart_tool/package_config.json` file mapping _package names_ to directories in the global `$PUB_CACHE` directory. The `.dart_tool/package_config.json` file is used by other tools, such as analyzer, compiler, etc. when they need to resolve statments such as `import "package:foo/foo.dart"`.
+**Example:** The [dart pub get](https://dart.dev/tools/pub/cmd/pub-get) tool
+will dowload and extract dependencies to a global `$PUB_CACHE` directory,
+and then write a `.dart_tool/package_config.json` file mapping _package names_
+to directories in the global `$PUB_CACHE` directory.
+The `.dart_tool/package_config.json` file is used by other tools,
+such as analyzer, compiler, etc, when they need to resolve statments
+such as `import "package:foo/foo.dart"`.
 
-When developing a tool that needs project specific caching, it can be advantagous to use the `.dart_tool/` directory because most users already have it ignored in `.gitignore`.
-When caching files for your tool in a user's `.dart_tool/` directory, you should always use a unique subdirectory.
-To avoid collisions, subdirectories like `.dart_tool/<tool_package_name>/` are reserved for the package named `<tool_package_name>`.
+When developing a tool that needs project specific caching, it can be
+advantagous to use the `.dart_tool/` directory because most users already
+ignore the directory using `.gitignore`.
+When caching files for your tool in a user's `.dart_tool/` directory,
+you should always use a unique subdirectory.
+To avoid collisions, subdirectories of the form `.dart_tool/<tool_package_name>/`
+are reserved for the package named `<tool_package_name>`.
 If your tool is not distributed through [pub.dev](https://pub.dev), you may publish a placeholder package in order to reserve the reserve a unique name.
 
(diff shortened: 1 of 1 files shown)


jonasfj:
@dcharkes, @devoncarew, @natebosch, @jakemac53, thoughts?

We can also make it `.dart_tool/pkg/<packageName>`, but I'm not sure it matters much either way.

dcharkes:
Yeah, we should document and specify this.

> We can also make it `.dart_tool/pkg/<packageName>`, but I'm not sure it matters much either way.

The upside is that we have all subfolders available for dart tooling, in case we don't want(or can) claim a package with the folder name we want.

The downside is that it is not what we're currently doing in various packages, for example `package:build`. These packages would then have to migrate. (These migrations should be relatively pain-free.)

So, I don't have a strong opinion about it.


One thing that might be useful to add to the documentation is which `<packageName>` to use. For example if `package:build` is used in `package:foobar` and generates code for `foobar`, it should still use `.dart_tool/build` (for it's own convenience it can use `.dart_tool/build/foobar`). If it would use `.dart_tool/foobar` it could start colliding with another package also used in `package:foobar`.

jakemac53:
Fwiw for `.dart_tool/build` there are some other tools (such as the analyzer) which rely on the current location of the directory. So the migration wouldn't be nearly as simple for that one.

src/tools/pub/package-layout.md
```diff
@@ -476,4 +473,33 @@ documentation generators, or other bits of automation.
...
 If you have any of these, place them in a directory called `tool`.
 
+## Project specific caching for tools
+
+{{site.alert.info}}
+  Do not check the `.dart_tool/` directory into source control.
+  Instead keep `.dart_tool/` in `.gitignore`.
+{{site.alert.end}}
+
+The `.dart_tool/` is created when you've run `dart pub get`.
```
lrhn, line 483: "The `.dart_tool/` directory is created when you run `dart pub get`."

 (Add "directory", "when you've" -> "when you" - otherwise it should be "have been created when you've run")

src/tools/pub/package-layout.md
```diff
@@ -476,4 +473,33 @@ documentation generators, or other bits of automation.
...
+The directory used by various tooling for caching files specific to your project and/or local machine.
```
lrhn, line 484: ... directory is used by various tools ...

"Various" suggests a number of things, but "tooling" is not countable.

src/tools/pub/package-layout.md
```diff
@@ -476,4 +473,33 @@ documentation generators, or other bits of automation.
...
+Thus, the `.dart_tool/` directory should never be checked into source control, or copied between machines.
+
+It is also generally safe to delete the `.dart_tool/` directory, though some tools might need recompute the cached information before they work again. 
+
+**Example:** The [dart pub get](https://dart.dev/tools/pub/cmd/pub-get) tool will dowload and extract dependencies to a global `$PUB_CACHE` directory, and then write a `.dart_tool/package_config.json` file mapping _package names_ to directories in the global `$PUB_CACHE` directory. The `.dart_tool/package_config.json` file is used by other tools, such as analyzer, compiler, etc. when they need to resolve statments such as `import "package:foo/foo.dart"`.
```
lrhn, line 489: (Break long lines like this. It makes it easier to edit in the future.)

Add comma after "etc.".

lrhn (approved):
LGTM, only a few phrasing and punctuation nits.

"Punctuation, is? fun!"

lrhn:
The code in `package:build` (or other packages owned by the same people) should use `build/`, even when building things for `package:foobar`. They cannot know whether `package:foobar` uses `foobar/` itself. They do know that they use `build/`, and nobody else should.

So yes, it should say that. I think it does now, using `<tool_package_name>`.

lrhn:
@jakemac53 Can't use `sdk/` since there is already a [package called `sdk`](https://pub.dev/packages/sdk) that we don't own.

I'd just use `.sdk/` if I needed a subdirectory. Or use any filename which is not a valid Pub package name.
(Although, not all packages are Pub packages, and some could have names that a are not valid Pub package names. I guess we should just tell them to pick a valid pub package name instead).

jakemac53:
> The code in `package:build` (or other packages owned by the same people) should use `build/`, even when building things for `package:foobar`.

Correct, and this is what `package:build` does.

> @jakemac53 Can't use `sdk/` since there is already a [package called `sdk`](https://pub.dev/packages/sdk) that we don't own.

@matanlurey owns it and I am sure would happily hand it over

jakemac53:
👍  I think it is quite reasonable to suggest using domain names instead of package names, if you don't have a package. That also matches up nicely with the publishers on pub.dev.

We could explicitly also allow using your publisher domain if you want a single canonical dir for all your packages, with whatever layout you choose.

natebosch:
I do not see value in a `.dart_tool/pkg` extra level.

I do think it makes sense to also allow a domain name - no package can have a name that looks like a domain, and no domain can look like a package since it must have a `.`.

src/tools/pub/package-layout.md
```diff
@@ -476,4 +473,33 @@ documentation generators, or other bits of automation.
...
+
+When developing a tool that needs project specific caching, it can be advantagous to use the `.dart_tool/` directory because most users already have it ignored in `.gitignore`.
+When caching files for your tool in a user's `.dart_tool/` directory, you should always use a unique subdirectory.
+To avoid collisions, subdirectories like `.dart_tool/<tool_package_name>/` are reserved for the package named `<tool_package_name>`.
```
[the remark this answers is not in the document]
jonasfj: Package naming conventions are documented elsewhere (if they are documented at all). We could add it, but I doubt it's necessary, and we just risk failing to keep it updated.

src/tools/pub/package-layout.md
```diff
@@ -476,4 +473,46 @@ documentation generators, or other bits of automation.
...
+{{site.alert.info}}
+  Do not check the `.dart_tool/` directory into source control.
+  Instead keep `.dart_tool/` in `.gitignore`.
+{{site.alert.end}}
+
+The `.dart_tool/` directory is created when you run `dart pub get`.
+The directory is used by various tools for caching files specific to your project and/or local machine.
+The `.dart_tool/` directory should never be checked into source control, or copied between machines.
+
+It is also generally safe to delete the `.dart_tool/` directory, though some tools might need recompute the cached information before they work again.
```
dcharkes, line 487: Should we strengthen this, so people using the `.dart_tool` directory know what the contract is: "Your files might get deleted at any time".

It is safe to delete the `.dart_tool/` directory, though some tools might need recompute the cached information before they work again.

dcharkes (approved):
lgtm

src/tools/pub/package-layout.md
parlough:
```suggestion
It is safe to delete the `.dart_tool/` directory,
but some tools might need recompute the cached files
before they work again. 
```

sfshaza2 (approved):
I have some minor edits, so please double check that I haven't made any errors. Once these are handled, lgtm

src/tools/pub/package-layout.md
```diff
@@ -58,14 +58,11 @@ enchilada/
     style.css
 {% endprettify %}
 
-\* The `.dart_tool/` directory exists after you've run `pub get`.
```
parlough, line 61: I think it would be nice to keep this quick reference here as readers may not go further down the page.

Could add something like to reference that later section as well:

```markdown
To learn more, 
see [Project specific caching for tools](#project-specific-caching-for-tools).
```

src/tools/pub/package-layout.md
```diff
@@ -476,4 +473,46 @@ documentation generators, or other bits of automation.
...
+
+It is also generally safe to delete the `.dart_tool/` directory, though some tools might need recompute the cached information before they work again. 
+
+**Example:** The [dart pub get](https://dart.dev/tools/pub/cmd/pub-get) tool
+will dowload and extract dependencies to a global `$PUB_CACHE` directory,
+and then write a `.dart_tool/package_config.json` file mapping _package names_
+to directories in the global `$PUB_CACHE` directory.
+The `.dart_tool/package_config.json` file is used by other tools,
+such as analyzer, compiler, etc, when they need to resolve statments
+such as `import "package:foo/foo.dart"`.
```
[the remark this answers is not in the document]
parlough: These changes look good. I've also adjusted the format and link to `dart pub get` in the suggestion.

src/tools/pub/package-layout.md
```diff
@@ -476,4 +473,46 @@ documentation generators, or other bits of automation.
...
+
+When developing a tool that needs project specific caching, it can be
+advantagous to use the `.dart_tool/` directory because most users already
+ignore the directory using `.gitignore`.
+When caching files for your tool in a user's `.dart_tool/` directory,
+you should always use a unique subdirectory.
+To avoid collisions, subdirectories of the form `.dart_tool/<tool_package_name>/`
+are reserved for the package named `<tool_package_name>`.
+If your tool is not distributed through [pub.dev](https://pub.dev), you may publish a placeholder package in order to reserve the reserve a unique name.
```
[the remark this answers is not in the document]
parlough: Thanks for cleaning this up. I've also adjusted the pub link.

parlough (commented):
Thanks for adding this to better formalize the directory and how it is used.

Looks good to me once the question from Daco is resolved and feedback from Shams is addressed.

src/tools/pub/package-layout.md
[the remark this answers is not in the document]
parlough: Thanks for cleaning this up. I've adjusted some line breaks and the pub link in the provided suggestion.

src/tools/pub/package-layout.md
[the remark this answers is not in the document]
parlough: These changes look great, thanks. I've also adjusted the format and link to `dart pub get` in the suggestion.

src/tools/pub/package-layout.md
sfshaza2, line 61: Running `pub get` creates the `.dart_tool` directory if none exists.
To learn more, 
see [Project specific caching for tools](#project-specific-caching-for-tools).

sfshaza2 (approved):
lgtm

src/tools/pub/package-layout.md
sfshaza2: I believe this is addressed.

src/tools/pub/package-layout.md
sfshaza2: Ahh, shoot, it's not letting me commit this suggestion.

parlough: I've manually pushed a fix adding this back.

sfshaza2 (approved):
lgtm

Issue #4113, opened 2022-06-22
[PAGE ISSUE]: 'Asynchronous programming: futures, async, await'


magickibbles:
### Page URL

https://dart.dev/codelabs/async-await.html

### Page source

https://github.com/dart-lang/site-www/tree/main/src/codelabs/async-await.md

### Describe the problem

The test solution for [practice using async and await](https://dart.dev/codelabs/async-await#exercise-practice-using-async-and-await) doesn't work because of null-safety

### Expected fix

please update this so it functions

parlough:
Hi @magickibbles, thanks for opening an issue.

Could you provide some more information on the issue you're facing? Are there any specific errors? The tests and solution provided for this exercise seemed to work for me.

Thanks again :)

Tohusk:
I'm also getting this issue on firefox,
![Screenshot 2022-07-11 at 10-51-37 Asynchronous programming futures async await](https://user-images.githubusercontent.com/33451949/178169654-c2992586-57c8-4597-9b36-1b4eb7c04b68.png)
same issue with the two exercises below:
![Screenshot 2022-07-11 at 10-52-11 Asynchronous programming futures async await](https://user-images.githubusercontent.com/33451949/178169687-b1bb624b-26e9-4c4b-bdde-ed71c0f0dbe9.png)
![Screenshot 2022-07-11 at 10-52-24 Asynchronous programming futures async await](https://user-images.githubusercontent.com/33451949/178169697-62435e74-e6cd-4c25-91f9-83350d084a64.png)

parlough:
Thanks for catching this @Tohusk, fixing it now :)

magickibbles:
> Hi @magickibbles, thanks for opening an issue.
> 
> Could you provide some more information on the issue you're facing? Are there any specific errors? The tests and solution provided for this exercise seemed to work for me.
> 
> Thanks again :)

honestly forgot why I submitted this. sorry. ty tho :^)

PR #4138, opened 2022-07-11
Add code excerpts for async-await codelab

parlough:
Staged: https://dart-dev--pr4138-misc-async-await-cod-jnz6avub.web.app/codelabs/async-await

Fixes https://github.com/dart-lang/site-www/issues/4113


parlough:
The code excerpt processor/framer is causing some problems, perhaps due to the file name. I haven't looked at that Ruby code before, but I'll take a look.

PR #4146, opened 2022-07-16
Fix typo in language tour

nownabe:
Hi, I fixed a tiny typo in the language tour.
https://dart.dev/guides/language/language-tour#late-variables


parlough (approved):
Thanks for catching and fixing this, looks good to me!

PR #4152, opened 2022-07-23
[Effective Dart] Updated docs guidance for boolean variables

samlythemanly:
The current guidance for documentation adequately handles the cases for non-boolean variables, but doesn't quite properly cover booleans. The most common documentation I've seen for booleans (and also what I believe is the most concise) is to start it with "Whether"  followed by some phrase describing the states for the variable. "Whether" is a conjunction and is _technically_ not a part of a noun phrase that follows it. I'm aware of the pedantry here but it _has_ led to inconsistencies in code reviews due to misunderstandings. Additionally, the current guidance doesn't afford the use of a gerund phrase in this sort of comment which is also rather common.

I've updated the guidance to recommend starting the comments with "Whether" and then either followed by a noun phrase or a gerund phrase. Let me know if this works!


src/_guides/language/effective-dart/documentation.md
```diff
@@ -267,11 +267,56 @@ int weekday;
...
+<?code-excerpt "docs_good.dart (getter-and-setter)"?>
+{% prettify dart tag=pre+code %}
+/// The pH level of the water in the pool.
+///
+/// Ranges from 0-14, representing acidic to basic, with 7 being neutral.
+int get phLevel => ...
+set phLevel(int level) => ...
+
+{:.bad}
+<?code-excerpt "docs_bad.dart (getter-and-setter)"?>
```
parlough, line 312: You'll need to open and close the prettify block for each excerpt.

```suggestion
set phLevel(int level) => ...
{% endprettify %}

{:.bad}
<?code-excerpt "docs_bad.dart (getter-and-setter)"?>
{% prettify dart tag=pre+code %}
```

src/_guides/language/effective-dart/documentation.md
```diff
@@ -267,11 +267,56 @@ int weekday;
 int get checkedCount => ...
 {% endprettify %}
 
+### PREFER starting boolean variables or properties with "Whether" followed by noun or gerund phrases.
+
+The doc comment should clarify the states this variable represents. This is
+true even for getters which may do calculation or other work. What the caller
+cares about is the *result* of that work, not the work itself.
```
parlough, line 274:
```suggestion
The doc comment should clarify the states this variable represents. 
This is true even for getters which may do calculation or other work. 
What the caller cares about is the *result* of that work, not the work itself.
```

src/_guides/language/effective-dart/documentation.md
```diff
@@ -267,11 +267,56 @@ int weekday;
...
+bool get canResize => ...
+{% endprettify %}
+
+{{site.alert.note}}
+  This guideline intentionally doesn't include using "Whether or not". In many
+  cases, usage of "or not" with "whether" is superfluous and can be omitted,
+  especially when used in this context.
+{{site.alert.end}}
+
+### DON'T write documentation for both the getter and setter of a property.
```
parlough, line 295: I like adding this as a standalone guideline, makes it easier to discover and point to. Thanks :)

parlough (commented):
Thanks for working on these additions and adjustments. They make sense to me, but since it's adding a new guideline, I'll defer to @munificent. 

You'll need to fix the code block to satisfy the excerpt checker and perhaps other issues once that is fixed.

src/_guides/language/effective-dart/documentation.md
```diff
@@ -267,11 +267,58 @@ int weekday;
...
 {% endprettify %}
 
+### PREFER starting boolean variables or properties with "Whether" followed by noun or gerund phrases.
+
+The doc comment should clarify the states this variable represents. 
+This is true even for getters which may do calculation or other work. 
+What the caller cares about is the *result* of that work, not the work itself.
+
+{:.good}
+<?code-excerpt "docs_good.dart (noun-phrases-for-non-boolean-var-etc)"?>
```
parlough, line 277: Seems this is the boolean excerpt, which is probably why the excerpt check is failing now:

```suggestion
<?code-excerpt "docs_good.dart (noun-phrases-for-boolean-var-etc)"?>
```

src/_guides/language/effective-dart/documentation.md
```diff
@@ -251,14 +251,14 @@ void start() {
 }
 {% endprettify %}
 
-### PREFER starting variable, getter, or setter comments with noun phrases.
+### PREFER starting non-boolean variables or properties with noun phrases.
```
munificent, line 254: I think it's important to keep "comment" in here, because it's the comment this applies to and not the member itself. How about:

```
### PREFER starting a non-boolean variable or property comment with a noun phrase.
```

src/_guides/language/effective-dart/documentation.md
```diff
@@ -267,11 +267,58 @@ int weekday;
 int get checkedCount => ...
 {% endprettify %}
 
+### PREFER starting boolean variables or properties with "Whether" followed by noun or gerund phrases.
```
munificent, line 270: And here:

```
### PREFER starting a boolean variable or property comment with "Whether" followed by noun or gerund phrase.
```

munificent (approved):
I like this, thanks!

src/_guides/language/effective-dart/documentation.md
samlythemanly: Woops, didn't mean to remove it! Fixed.

parlough (approved):

parlough:
Thanks for working on this and being patient! Looks good to me (after regenerating the TOC).

PR #4155, opened 2022-07-27
Fix: dart installing reference

vpstarucka:
When trying to install dart from the latest Homebrew version with the command "brew install dart" it gives a warning:

"
Warning: No available formula with the name "dart". Did you mean dar or dirt?
==> Searching for similarly named formulae...
These similarly named formulae were found:
dart-sdk            dartsim             dar                 dirt
"

Update dart reference to one that works, dart-sdk

### DIFF
--- a/src/get-dart/_mac.md
+++ b/src/get-dart/_mac.md
@@ -3,7 +3,7 @@ and then run the following commands:
 
 ```terminal
 $ brew tap dart-lang/dart
-$ brew install dart
+$ brew install dart-sdk
 ```
 
 {{site.alert.important}}


domesticmouse:
PTAL @mit-mit

mit-mit:
cc @athomas

parlough:
It looks like dart-sdk is a [new formula](https://github.com/Homebrew/homebrew-core/blob/master/Formula/dart-sdk.rb) but it works a bit differently than the Dart team’s official formula and doesn’t support as much version or channel control.

As far as I know, the recommended way is to still use the Dart team’s Homebrew tap to install Dart as the install page outlines:

```console
brew tap dart-lang/dart
brew install dart
```

If `dart` doesn’t work after adding the tap though, let us know. Thanks!

mit-mit:
Ahh, interesting @parlough!

@athomas I wonder if we could/should contribute what's missing to the core one, and then make that the official way?

also cc @kevmoo for thoughts

athomas:
The dart-sdk formula is a very different beast. It's creating a local build. I don't know how that affects signing and notarization but in theory it's more homebrewy. I suspect that user didn't run `brew tap dart-lang/dart`.

parlough:
I'm going to close this for now as it sounds like for the time being, the `dart-lang/dart` tap will be preferred destination.

If it doesn't work after running `brew tap dart-lang/dart` though, let us know. Thanks!

PR #4168, opened 2022-08-11
Fixes typo in creating-streams.md

cofirazak:
Fixed a typo at:
Stream<int> timedCounter(Duration interval, [int? maxCount]) async* {
...
If the c̶o̶u̶n̶t̶ maxCount parameter is omitted


parlough (approved):
Thanks for catching this, looks good to me!

PR #4179, opened 2022-08-19
Fix broken build due to dart format changes

parlough:
Also updates Node checksum to fix check run during build and staging.


atsansone (approved):
I'll assume those are valid unless there is another way to validate.

PR #4193, opened 2022-08-25
Update JS compiler content for v2.18

atsansone:
### Removed mentions of `dart2js` and `dartdevc`

* [FAQ](https://dart--staging.web.app/faq)
* [Null Safety FAQ](https://dart--staging.web.app/null-safety/faq)
* [Overview](https://dart--staging.web.app/overview)
* [`dart compile`](https://dart--staging.web.app/tools/dart-compile)
* [`dart`](https://dart--staging.web.app/tools/dart-tool)
* [tools](https://dart--staging.web.app/tools)
* [webdev](https://dart--staging.web.app/tools/webdev)
* [debugging](https://dart--staging.web.app/web/debugging)
* [deployment](https://dart--staging.web.app/web/deployment)

### Added production debugging section from `dart2js`
* [debugging](https://dart--staging.web.app/web/debugging)

### Removed all content
* [`dart2js`](https://dart--staging.web.app/tools/dart2js)
* [`dartdevc`](https://dart--staging.web.app/tools/dartdevc)


src/_guides/language/language-tour.md
```diff
@@ -2470,9 +2470,9 @@ When exactly do assertions work?
 That depends on the tools and framework you're using:
 
 * Flutter enables assertions in [debug mode.][Flutter debug mode]
-* Development-only tools such as [dartdevc][]
+* Development-only tools such as [`webdev serve`][]
   typically enable assertions by default.
-* Some tools, such as [`dart run`][] and [dart2js][]
+* Some tools, such as [`dart run`][] and [`dart compile`][]
   support assertions through a command-line flag: `--enable-asserts`.
```
mit-mit, line 2476: I don't think `dart compile` has the `--enable-asserts` flag?

src/tools/dart-compile.md
```diff
@@ -136,11 +136,21 @@ plus a small [Dart runtime][] that handles
...
 ```terminal
 $ dart compile exe bin/myapp.dart -o /tmp/myapp
+```
+
+When successful, this command returns the following.
+```terminal
 Generated: /tmp/myapp
+```
+
+To run your compiled app, change to the `/tmp` directory and run.
```
mit-mit, line 152: I think this would be simpler if it was just `/tmp/myapp` without having to change directory.

src/tools/dartdevc.md
```diff
@@ -4,209 +4,15 @@ short-title: dartdevc
...
 {{site.alert.note}}
-  dartdevc is used for _development_ only.
-  Continue to use [dart2js](/tools/dart2js)
-  through [`dart compile js`](/tools/dart-compile#js) 
-  or the [`webdev`][] tool
-  to compile for deployment.
+  Dart 2.18 removes the `dartdevc` command-line tool from the Dart
+  package, but retains the dartdevc compiler.
+  
+  Use [`dart compile js`](/tools/dart-compile#js) to compile Dart code
```
mit-mit, line 15: I don't think this is true. `compile js` will produce non-debuggable JS. I think maybe we need to point to webdev instead, @sigmundch ?

mit-mit: It actually looks like the removed content below might all still work, and just need to be retained?

mit-mit (changes requested):

src/_guides/language/language-tour.md
atsansone: Maybe? See [dart2js code](https://github.com/dart-lang/sdk/blob/main/sdk/bin/dart2js#L40-L44).

src/tools/dartdevc.md
atsansone: This all falls under the `dartdevc` command. I can excise some parts and put them elsewhere, like `webdev`. I thought you were looking for removing anything other than `dart compile js`, so I tried to stay in that lane. Can you indicate which lines I should move? All of them, changing `dartdevc` to `webdev serve`?

src/tools/dart-compile.md
atsansone: Not a problem here.

src/_guides/language/language-tour.md
sigmundch: We have it for the dart2js subcommand (`dart compile js`), but I don't know about all the other subcommands under `dart compile`. For example `dart compile exe` seems to have it, but jit-snapshot and kernel don't seem to have it.

src/_guides/language/language-tour.md
```diff
@@ -4202,9 +4202,9 @@ Here are some cases when you might use deferred loading:
 * To load rarely used functionality, such as optional screens and dialogs.
 
 {{site.alert.warn}}
-  **Only dart2js supports deferred loading.**
-  Flutter, the Dart VM, and dartdevc don't support deferred loading.
-  For more information, see
+  **Only `dart compile js` supports deferred loading.**
```
sigmundch, line 4205: FYI - I believe some support for deferred loading has been added to the VM and we are working towards adding it to the dartdevc workflow (webdev). That's tangential to this change, so I wouldn't change anything here, just wanted to give you a heads up :)

src/_guides/whats-new.md
```diff
@@ -17,6 +17,17 @@ and follow the [Dart blog][].
...
+## August 30, 2022: 2.18 release
+
+This section lists notable changes made from May 12, 2022,
+through August 30, 2022.
+For details about the 2.18 release,
+see [Dart 2.18: ][], and the [change log][https://github.com/dart-lang/sdk/blob/main/CHANGELOG.md#2180].
+
+In addition to bug fixes and incremental improvements,
+we made the following changes to this site:
+
```
sigmundch, line 29: is your intent to also add here some bullet points about these recent changes?

src/faq.md
```diff
@@ -238,16 +238,16 @@ long enough for runtime optimizations to be applied.
...
 
   * Chrome
   * Edge
   * Firefox
   * Safari
 
-The _development_ compiler ([dartdevc][]) supports only Chrome.
-You _might_ be able to use other modern browsers
+The _development_ web compiler supports Chrome only.
+You _might_ be able to use other modern browsers based on Chromium
```
sigmundch, line 250: I'd recommend reverting this line change - only Edge is based on chromium, FF and Safari are not.

src/faq.md
```diff
@@ -238,16 +238,16 @@ long enough for runtime optimizations to be applied.
...
 (Chromium Edge, Firefox, and Safari).
```
sigmundch, line 251: Consider removing there "Chromium", since now the default is that Edge is a chromium based browser?

src/tools/dartdevc.md
sigmundch: My recommendation here is that we are not downplaying one command over others. Both commands are serving different purposes and each is valid for the journey the developer is at the moment (I provided additional details in the planning doc if that helps).

That means that we'd keep both references to web development and web production compilers, but change the sample commands. If a command used `dart2js`, we turn it into `dart compile js`, if a command used `dartdevc`, we turn it into `webdev serve`.

In the past, calling `dartdevc` directly was not recommended anyways, and not something I expect we had said to do in the documentation (we used the term to refer to the development compiler, but I'd be surprised if we had examples to show people how to call it directly). So I expect for the most part we used `webdev serve` to describe those use cases, and we can simply preserve good chunks of the old documentation.

src/null-safety/faq.md
```diff
@@ -319,72 +319,75 @@ and situations where a null value is really expected. So the tool tells you what
 it knows ("it looks like this condition will always be false!") and lets you
 decide what to do.
 
-## What should I know about dart2js and null safety?
+## What should I know about compiling and null safety?
 
-For a long time, the dart2js compiler has had optimizations specifically
-targeting null values and null checks. Because of that, null safety is not
-expected to affect much the output of the compiler.
+The compiler optimizes null values and null checks. So null safety
```
sigmundch, line 324: I think there is some missing pieces in this paragraph that it's worth adding.  I realize now that when I first wrote this I didn't word this properly to really convey what I was trying to say, so while your edits make sense looking at this with fresh eyes, I feel that some of the nuance that I was trying to convey here gets lost in the process.

To provide additional context. Developers know that null safety bring many benefits. One such benefit is that it will reduce code-size and improve performance of apps compiled natively (e.g. flutter and AOT targets). Many developers believe that the same benefits will come to JavaScript apps, but in reality the benefits will be much smaller. This Q/A was added because we wanted to set expectations appropriately.

My goal with the first sentence "For a long time ..." was to provide the reason behind this. That is, while our compiler learns from null-safety and it makes the compiler job easier, it's just that in the past there were many optimizations added to the production web compiler to specifically recover the cost of null checks, and as a result the benefits of null safety are not as easily observable when looking at the output of the compiler.

There is probably a better way to convey all of this :). I know this is tangential to your edits, so I'd be fine keeping the original wording for now.

src/null-safety/faq.md
```diff
@@ -319,72 +319,75 @@ and situations where a null value is really expected. So the tool tells you what
...
-  unnecessary null checks when it knows the value is not null.
-
-* By default dart2js would generate parameter subtype checks (runtime checks
-  used to ensure covariant virtual calls are given appropriate arguments).
-  These are elided with the `--omit-implicit-checks` flag just as before.
-  Recall that this flag can make programs have unexpected behavior if types
-  are invalid, so we continue to recommend that the code has strong test
-  coverage to avoid any surprises. In particular, dart2js optimizes code based
+* The compiler generates `!` null assertions, but you might not notice
+  them. Pre null-safety, the compiler generated null checks that still
```
sigmundch, line 330: nit: reading this with fresh eyes - this is not very easy to follow. Maybe the subject here is to implicit (my fault when I first added this text, I think). Suggestion:

* The production web compile generates `!` null assertions, but you might not notice them when comparing the output of the compiler before and after adding null assertions. That's because the compiler already emitted null checks in programs that weren't null-safe.

src/null-safety/faq.md
```diff
@@ -319,72 +319,75 @@ and situations where a null value is really expected. So the tool tells you what
...
+  remain in existing programs).
+
+* The compiler generates these null assertions regardless of the
+  soundness of null safety or optimization level. In fact, the compiler
+  doesn't remove `!` when using `-O3` or `--omit-implicit-checks`.
+
+* The compiler might remove unnecessary null checks. Previous compiler
```
sigmundch, line 337: I have a similar concern here as with the the intro paragraph: I think the context is getting lost. Seeing your edit reveal to me that the original wording was vague, so it highlights that we need to be more explicit about that context.

A suggestion here may be to clarify that when we used to say "optimizations done in the past" we meant "optimizations done in the compiler prior to null-safety".

src/null-safety/faq.md
```diff
@@ -319,72 +319,75 @@ and situations where a null value is really expected. So the tool tells you what
...
-  types before, and is true with inconsistent nullabilities now with sound
-  null-safety.
+  the code provides arguments of an invalid type, those optimizations
+  would be wrong and the program could misbehave. This was true for
+  inconsistent types before, and is true with inconsistent 
+  nullabilities now with sound null-safety.
 
-* You may notice that DDC and the Dart VM have special error
-  messages for null checks, but to keep applications small dart2js does not.
+* You may notice that the development mode of the compiler and the Dart
```
sigmundch, line 354: nit: replace "the development mode of the compiler" with "the web development compiler".

src/overview.md
```diff
@@ -187,9 +187,8 @@ Dart's compiler technology lets you run code in different ways:
   an ahead-of-time (AOT) compiler for producing machine code.
 
 * **Web platform**: For apps targeting the web,
-  Dart includes both a development time compiler (dartdevc) and
-  a production time compiler (dart2js).
-  Both compilers translate Dart into JavaScript.
+  Dart can compile for debugging or production.
```
sigmundch, line 190: nit: debugging/development ? (to keep things consistent?)

src/overview.md
```diff
@@ -187,9 +187,8 @@ Dart's compiler technology lets you run code in different ways:
...
+  Its web compiler translates Dart into JavaScript.
```
sigmundch, line 191: Technically there are 2 ocmpilers though... another thought here is to keep the original wording, but remove the words in parenthesis from the original sentence?

src/tools/dart-compile.md
```diff
@@ -259,24 +254,45 @@ they can have much slower startup than architecture-specific AOT output formats.
...
 
-Here's an example of compiling a Dart application to JavaScript
-with many optimizations enabled:
+{{site.alert.note}}
+  Use the [`webdev` tool][webdev] rather than running the 
+  Dart-to-JavaScript compiler.
+  The [`webdev build`][] command, by default,
+  also produces minified, deployable JavaScript.
+  The [`webdev serve`][] command, by default,
+  produces JavaScript modules to compile web apps
```
sigmundch, line 263: nit: delete "to compile web apps" ?

src/_includes/tools/dart-compile-js-options.md
```diff
@@ -0,0 +1,112 @@
...
+  run `dart compile js -hv`.
+
+  * `-O0`: Disables many optimizations.
+  * `-O1`: Enables default optimizations.
+  * `-O2`: Enables `-O1` optimizations, plus additional ones
+    (such as minification) that respect the language semantics and
+    are safe for all programs.
+    {{site.alert.note}}
+      With `-O2`, string representations of types are no longer the same as
+      those in the Dart VM and when compiled with [dartdevc][].
```
sigmundch, line 34: note: one more reference to dartdevc to remove here.

src/tools/dartdevc.md
sigmundch: Correct - in this case we'd point them at `webdev serve`

There is still a page for webdev below, so rather than retaining the content of the dartdevc.md file, I'd recommend selecting the relevant pieces to preserve that make sense to show in the context of webdev. From looking at the contents on this file, thought, I don't think there is much I'd keep (all content about module format, etc, are not really that relevant to our developers).

src/web/deployment.md
```diff
@@ -3,67 +3,31 @@ title: Web deployment
...
-  If some code paths aren't tested,
-  your app might run with dartdevc but
-  misbehave when compiled using dart2js.
-{{site.alert.end}}
-
-You can specify dart2js options in a build config file,
-as described in the [build_web_compilers page.][build_web_compilers]
+[Use the `webdev build` command][build] to create a deployable version
+of your app. This command converts your code to JavaScript and saves
+the result as `build/web/main.dart.js`.
```
sigmundch, line 25: I agree with most of the deletions here, but it may be worth including also a link to the dart2js options (I wouldn't document them individually here, but refer to the fact that under the hood `webdev build` invokes `dart compile js`, so all options available there can also be used to configure a `webdev build`.

src/tools/dart-compile.md
sfshaza2: Agree with mit.

src/_guides/whats-new.md
parlough: I guess this was accidentally included as that is being handled in https://github.com/dart-lang/site-www/pull/4195

src/overview.md
atsansone: I made a different update. Let me know what you think.

src/tools/dartdevc.md
```diff
@@ -4,209 +4,15 @@ short-title: dartdevc
...
-  dartdevc is used for _development_ only.
-  Continue to use [dart2js](/tools/dart2js)
-  through [`dart compile js`](/tools/dart-compile#js) 
-  or the [`webdev`][] tool
-  to compile for deployment.
+  Dart 2.18 removes the `dartdevc` command-line tool from the Dart
+  package, but retains the dartdevc compiler.
+  
+  Use [`webdev serve`](/tools/webdev#serve) to compile Dart code
+  to modular JavaScript. You can debug these modules more easily.
```
johnpryan, line 16:
```suggestion
  Use [`webdev serve`](/tools/webdev#serve) to compile Dart code
  to modular JavaScript. This makes compilation faster and debugging easier.
```

src/faq.md
```diff
@@ -238,17 +238,17 @@ long enough for runtime optimizations to be applied.
...
 
   * Chrome
   * Edge
   * Firefox
   * Safari
 
-The _development_ compiler ([dartdevc][]) supports only Chrome.
-You _might_ be able to use other modern browsers
-(Chromium Edge, Firefox, and Safari).
+The _development_ web compiler supports Chrome only. You _might_ be
```
johnpryan, line 249: Should we link to webdev here?

src/faq.md
```diff
@@ -373,34 +370,38 @@ For example, the `dart:io` library
...
+right compiler for your use case. 
 
-The two compilers have different use cases. You don't usually have to worry
-about which compiler you're using, because the [webdev][] tool
-chooses the right compiler for your use case. When you're developing your app,
-webdev chooses [dartdevc][], which supports incremental compilation so
-you can quickly see the results of your edits.
-When you're building your app for deployment, webdev chooses [dart2js][],
-which uses techniques such as tree shaking to produce optimized code.
+* When you're developing your app, `webdev` chooses a development web
```
johnpryan, line 380:
```suggestion
* When you're developing your app, `webdev` uses Dart's development
```

src/faq.md
```diff
@@ -435,19 +436,15 @@ staying within smi range is still good practice.
...
-Float32List becomes a Float32Array.
-The one exception today is that dart2js does not support 64-bit integers
-and thus does not support Int64List or Uint64List.
-Dart code compiled via dart2js results in a runtime exception
-if either of those lists is used.
+JavaScript offers typed arrays compatible with Dart’s typed lists.
+The mapping is trivial: `Float32List` becomes a `Float32Array`.
+One exception exists: the production web compiler does not support
+64-bit integers: `Int64List` or `Uint64List`. Compiling Dart code with
+either of those lists results in a runtime exception.
```
johnpryan, line 443: Just FYI, [semantic line breaks](https://sembr.org/) like these are a convention the Dart and Flutter sites often use

src/null-safety/faq.md
```diff
@@ -172,13 +172,16 @@ The assert will be unnecessary when everything is fully migrated, but for now it
...
-But, the check *is* still needed if the program is a mixed-version one.  Until
-everything is fully migrated and the code switches to running with sound null
-safety, it will be possible for `arg` to be null.
+```dart
+if (arg == null) throw ArgumentError(...)`
+```
+
+You must include this check if the program is a mixed-version one.
+Until everything is fully migrated and the code switches to running
+with sound null safety, `arg` might be set to `null`.
```
johnpryan, line 184: This is unrelated to js compilation, not sure if this was committed by accident

src/tools/index.md
```diff
@@ -103,23 +102,12 @@ The Dart SDK includes the following general-purpose `dart` tool:
 : A suite of debugging and performance tools.
 
 
-## Tools for developing web apps {#web}
+## Tool for developing web apps {#web}
 
 The following tools support developing web apps:
```
johnpryan, line 107: This section might need a bigger update, but for consistency: 

```suggestion
The following tool support developing web apps:
```

src/tools/webdev.md
```diff
@@ -4,68 +4,76 @@ description: Command-line tools for Dart web development.
...
+`build_runner` to test your app.
 
 ## Setup
 
-The easiest way to get `webdev` is to globally install it,
-so that it can be [in your PATH.][PATH]
-Before you can use `webdev`,
-your web app must depend on the
-build_runner and build_web_compilers packages.
+Install `webdev` for all users. This adds the tool [to your PATH][PATH].
```
johnpryan, line 12:
```suggestion
Follow these instructions to get started using `webdev`.
```

src/tools/webdev.md
```diff
@@ -4,68 +4,76 @@ description: Command-line tools for Dart web development.
...
-to your app's `pubspec.yaml` file:
+To depend on these packages, add the following [dev dependencies][] to
+your app's `pubspec.yaml` file:
 
 <!--?code-excerpt "quickstart/pubspec.yaml (build dependencies)"?-->
-```
+```yaml
   dev_dependencies:
-    # ···
+    ···
```
johnpryan, line 48: The hash is a YAML comment, I'm not sure if we validate the yaml snippets or not, but if we ever do, we should avoid causing trouble:

```suggestion
    # ···
```

src/tools/dartdevc.md
sigmundch: Is it a requirement that we keep old pages around? or is it OK to delete them altogether?

After looking at this more closely, I'm no longer sure I see the value in keeping around both the dartdevc.md and the dart2js.md pages. So I'd be inclined in deleting them both. If we have to keep them, then I would suggest to trim them down further. For example, by deleting anything that is describing the compilers beyond production/development (e.g. I'd suggest deletion the whole explanation of _modules_ and _AMD_).

One suggestion:
* _"Version note: Since Dart 2.18 the `dart2js` web production compiler is only available from `dart compile js`(link)"_, and
* _"Version note: Since Dart 2.18 the `dartdevc` web development compiler is only available from `webdev serve`(link)."_ 

Another suggestion (I slightly prefer this one):
* _"dart2js is Dart's web production compiler. Version note: Since Dart 2.18 the `dart2js` compiler can only be invoked from `dart compile js`. It is also used internally by deployment tools like `webdev build`"_, and
* _"dartdevc is Dart's web development compiler. Version note: Since Dart 2.18 the `dartdevc` compiler cannot be invoked directly, but it is used internally by development tools like  `webdev serve`."_

src/tools/webdev.md
parlough: Yeah I suppose we added the hash in-case the user wanted to copy this snippet directly. 

Perhaps we should switch these instructions to `dart pub add` instead anyway though as those mentioned versions get out of date.

src/overview.md
sigmundch: How about "Its web compilers translate Dart into JavaScript" (plural compilers vs singluar compiler?)

Alternatively, combining both sentences:
Web platform: For apps targeting the web, Dart can compile to JavaScript, both for development and production purposes.

src/tools/index.md
```diff
@@ -103,23 +102,12 @@ The Dart SDK includes the following general-purpose `dart` tool:
...
-[dart2js](/tools/dart2js)
-: The original Dart-to-JavaScript compiler, with tree shaking.
-  IDEs, `dart compile js`, and the `webdev` CLI 
-  use dart2js when building web apps for deployment.
-
-[dartdevc](/tools/dartdevc)
-: The Dart dev compiler, a modular Dart-to-JavaScript compiler.
-  IDEs and the `webdev` CLI use dartdevc when running a development server.
-
+: A CLI to build and serve Dart web apps.
```
sigmundch, line 110: Should we consider adding back a bullet point equivalent to `dart2js` (maybe it would be `dart compile js` instead?).

It seems these were the primary links we had to steer readers towards the more detailed articles about how to develop and deploy apps to JavaScript, so without those, this index page feels incomplete.

Also worth nothing something for a follow up (clearly out of scope for this change): it's strange that have a couple workflow articles like web/deployment and web/debugging, but no article on web/development. It would be be consistent and have an article per user journey. My guess the reason is that it overlaps entirely with the content of the webdev.md article, so it would require a bit of editing to figure out the right organization of that content.

sigmundch (commented):
Overall the changes look great, thanks @atsansone!

I defer to the proper editors here to give the final stamp of approval 😊

mit-mit:
@atsansone is this ready for a last review? If so, could you please stage it somewhere? I'm having a bit of trouble visualizing how the pages end up linking together from just the source.

atsansone:
> 

@mit-mit : The links to the staged pages are in the description at the top of this PR.

src/_includes/known-issues/compile-ki.md
```diff
@@ -0,0 +1,20 @@
+No cross-compilation support ([issue 28617][])
```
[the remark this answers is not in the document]
atsansone: I yanked out this block from the `dart compile` file to cover `exe` and `aot-snapshot` more cleanly. It is unrelated. Apologies for the extra change.

src/_includes/tools/dart-compile-js-options.md
```diff
@@ -0,0 +1,112 @@
...
+
+Some other handy options include:
+
+`--packages=<path>`
+: Specifies the path to the package resolution configuration file.
+  For more information, see
+  [Dart Package Configuration File.][]
+
+`-D<flag>=<value>`
+: Defines an environment variable.
```
kevmoo, line 66: We have to have a doc on this that can be linked to.

I'm annoyed this page https://api.dart.dev/stable/2.18.0/dart-core/int/int.fromEnvironment.html – and others don't link to something like that!

src/faq.md
atsansone: Just learned what this is. Great point. I'll fix some here and others later.

src/null-safety/faq.md
atsansone: This was to clean up the language and format around this check. It's the same content as what was there originally.

src/tools/index.md
atsansone, line 110:
```suggestion
: A CLI to [build](/tools/webdev#build) and [serve](/tools/webdev#serve) Dart JavaScript apps.
```

atsansone (commented):
Made corrections per @johnpryan .

src/_guides/language/language-tour.md
```diff
@@ -2470,9 +2470,9 @@ When exactly do assertions work?
 That depends on the tools and framework you're using:
 
 * Flutter enables assertions in [debug mode.][Flutter debug mode]
-* Development-only tools such as [dartdevc][]
+* Development-only tools such as [`webdev serve`][]
```
mit-mit, line 2473: This like and the compile js one below don't render correctly. I think you are missing definitions of them.

Screen Shot 2022-08-30 at 10 49 31

mit-mit (approved):
Just one more comment (left above)

kevmoo (approved):

src/_includes/tools/dart-compile-js-options.md
```diff
@@ -0,0 +1,113 @@
+#### Options {#prod-compile-options}
```
johnpryan, line 1: includes are typically not used for sections like this. This can make things harder to maintain (the heading type might change depending on the context where it's included.) IMO it's better to just inline things and use includes for banners and other smaller things.

johnpryan (approved):

PR #4196, opened 2022-08-26
Add 2.18 type inference workaround from change log to sound-problems.md

MaryaBelanger:
Thought it might be a good idea to add the ["workaround"](https://github.com/dart-lang/sdk/blob/master/CHANGELOG.md#language-1) from the type inference enhancement included in 2.18 _somewhere_ in the docs, and this seemed like a good spot.

I'm not sure how the code excerpts work yet though, so, still a draft.

TODO
- [x]  Error message
- [ ]  Complete example code excerpts

I also moved a summary paragraph from the bottom of the page under the **Runtime errors** section. To me, it seems like a summary of the entire page and read strange at the bottom. So it's at the top now.

### DIFF
--- a/examples/type_system/lib/common_fixes_analysis.dart
+++ b/examples/type_system/lib/common_fixes_analysis.dart
@@ -134,14 +134,18 @@ void funcCast() {
 
 //-----------------------------------------------
 
-// #docregion type-inf-null
+void infNull() {
+  // #docregion type-inf-null
   List<int> ints = [1, 2, 3];
-  var maximumOrNull = ints.fold(null,
-      (a, b) => a == null || a < b ? b : a);
-// #enddocregion type-inf-null
+  // ignore: non_bool_operand, undefined_operator
+  var maximumOrNull = ints.fold(null, (a, b) => a == null || a < b ? b : a);
+  // #enddocregion type-inf-null    
+} 
 
-// #docregion type-inf-fix
+void infFix() {
+  // #docregion type-inf-fix
   List<int> ints = [1, 2, 3];
-  var maximumOrNull = ints.fold<int?>(null,
-      (a, b) => a == null || a < b ? b : a);
-// #enddocregion type-inf-fix
+  var maximumOrNull =
+      ints.fold<int?>(null, (a, b) => a == null || a < b ? b : a);
+  // #enddocregion type-inf-fix
+}
--- a/examples/type_system/lib/common_fixes_analysis.dart
+++ b/examples/type_system/lib/common_fixes_analysis.dart
@@ -131,3 +131,17 @@ void funcCast() {
   filterValues((x) => (x as String).contains('Hello'));
   // #enddocregion func-cast
 }
+
+//-----------------------------------------------
+
+// #docregion type-inf-null
+  List<int> ints = [1, 2, 3];
+  var maximumOrNull = ints.fold(null,
+      (a, b) => a == null || a < b ? b : a);
+// #enddocregion type-inf-null
+
+// #docregion type-inf-fix
+  List<int> ints = [1, 2, 3];
+  var maximumOrNull = ints.fold<int?>(null,
+      (a, b) => a == null || a < b ? b : a);
+// #enddocregion type-inf-fix
--- a/src/_guides/language/sound-problems.md
+++ b/src/_guides/language/sound-problems.md
@@ -37,7 +37,8 @@ List<String> [!string = numbers!];
 Since neither `List<int>` nor `List<String>` is a subtype of the other,
 Dart rules this out statically. 
 
-You can see other examples of static analysis errors, as well as other error types, in the sections below.
+You can see other examples of static analysis errors,
+as well as other error types, in the following sections.
 
 
 ## No type errors {#no-type-errors}
@@ -489,24 +490,24 @@ filterValues((x) => (x [!as String!]).contains('Hello'));
 ### Incorrect type inference
 
 ```nocode
-error - ...
+error - The return type '...' isn't a '...', as required by the closure's context - return_of_invalid_type_from_closure
 ```
 
-On rare occasions, Dart's enhaced type inference may infer the wrong type for function literal arguments in a generic constructor invocation.
+On rare occasions, Dart's enhanced type inference might infer
+the wrong type for function literal arguments in a generic constructor invocation.
 This primarily affects `Iterable.fold`
 
 #### Example
 
-In this code, type inference will infer that `a` has a type of `Null`:
+In the following code,
+type inference will infer that `a` has a type of `Null`:
 
 {:.fails-sa}
 <?code-excerpt "lib/... ()" replace=""?>
 {% prettify dart tag=pre+code %}
-void main() {
   List<int> ints = [1, 2, 3];
   var maximumOrNull = ints.fold(null,
       (a, b) => a == null || a < b ? b : a);
-}
 {% endprettify %}
 
 #### Fix: Supply appropriate type as explicit type argument
@@ -514,11 +515,9 @@ void main() {
 {:.passes-sa}
 <?code-excerpt "lib/... ()" replace=""?>
 {% prettify dart tag=pre+code %}
-void main() {
   List<int> ints = [1, 2, 3];
   var maximumOrNull = ints.fold<int?>(null,
       (a, b) => a == null || a < b ? b : a);
-}
 {% endprettify %}
 
 <hr>
--- a/src/_guides/language/sound-problems.md
+++ b/src/_guides/language/sound-problems.md
@@ -495,7 +495,7 @@ error - The return type '...' isn't a '...', as required by the closure's contex
 
 On rare occasions, Dart's enhanced type inference might infer
 the wrong type for function literal arguments in a generic constructor invocation.
-This primarily affects `Iterable.fold`
+This primarily affects `Iterable.fold`.
 
 #### Example
 
@@ -503,7 +503,7 @@ In the following code,
 type inference will infer that `a` has a type of `Null`:
 
 {:.fails-sa}
-<?code-excerpt "lib/... ()" replace=""?>
+<?code-excerpt "lib/common_fixes_analysis.dart (type-inf-null)"?>
 {% prettify dart tag=pre+code %}
   List<int> ints = [1, 2, 3];
   var maximumOrNull = ints.fold(null,
@@ -513,7 +513,7 @@ type inference will infer that `a` has a type of `Null`:
 #### Fix: Supply appropriate type as explicit type argument
 
 {:.passes-sa}
-<?code-excerpt "lib/... ()" replace=""?>
+<?code-excerpt "lib/common_fixes_analysis.dart  (type-inf-fix)"?>
 {% prettify dart tag=pre+code %}
   List<int> ints = [1, 2, 3];
   var maximumOrNull = ints.fold<int?>(null,
--- a/src/_guides/language/sound-problems.md
+++ b/src/_guides/language/sound-problems.md
@@ -508,8 +508,7 @@ type inference will infer that `a` has a type of `Null`:
 <?code-excerpt "lib/common_fixes_analysis.dart (type-inf-null)"?>
 {% prettify dart tag=pre+code %}
   List<int> ints = [1, 2, 3];
-  var maximumOrNull = ints.fold(null,
-      (a, b) => a == null || a < b ? b : a);
+  var maximumOrNull = ints.fold(null, (a, b) => a == null || a < b ? b : a);
 {% endprettify %}
 
 #### Fix: Supply appropriate type as explicit type argument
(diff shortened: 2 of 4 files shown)


src/_guides/language/sound-problems.md
```diff
@@ -20,7 +20,27 @@ and see [these other resources](/guides/language/type-system#other-resources).
 
 ## Troubleshooting
```
sfshaza2, line 21: It looks like it was this way before, but our style guidelines want at least one sentence between headers.

src/_guides/language/sound-problems.md
```diff
@@ -20,7 +20,27 @@ and see [these other resources](/guides/language/type-system#other-resources).
...
+{:.fails-sa}
+{% prettify dart tag=pre+code %}
+List<int> numbers = [1, 2, 3];
+List<String> [!string = numbers!];
+{% endprettify %}
+
+Since neither `List<int>` nor `List<String>` is a subtype of the other,
+Dart rules this out statically. 
+
+You can see other examples of static analysis errors, as well as other error types, in the sections below.
```
sfshaza2, line 40:
```suggestion
You can see other examples of static analysis errors,
as well as other error types, in the following sections.
```

It's a holdover from the old days, but I do try to avoid "above" and "below" as, if this is translated to Chinese or whatever, that isn't relevant.

src/_guides/language/sound-problems.md
```diff
@@ -466,29 +486,48 @@ filterValues((x) => (x [!as String!]).contains('Hello'));
...
-you can't get into a situation where the value stored in a variable is
-different from the variable's static type. Like most modern statically
-typed languages, Dart accomplishes this with a combination of static
-(compile-time) and dynamic (runtime) checking.
+```nocode
+error - ...
+```
 
-For example, the following type error is detected at compile-time:
+On rare occasions, Dart's enhaced type inference may infer the wrong type for function literal arguments in a generic constructor invocation.
```
sfshaza2, line 495:
```suggestion
On rare occasions, Dart's enhanced type inference might infer
the wrong type for function literal arguments in a generic constructor invocation.
```

Line length!!!  :D. We don't break at exactly 80 chars, we try to break on phrasing.

src/_guides/language/sound-problems.md
```diff
@@ -466,29 +486,48 @@ filterValues((x) => (x [!as String!]).contains('Hello'));
...
+This primarily affects `Iterable.fold`
+
+#### Example
+
+In this code, type inference will infer that `a` has a type of `Null`:
```
sfshaza2, line 500:
```suggestion
In the following code,
type inference infers that `a` has a type of `Null`:
```

src/_guides/language/sound-problems.md
```diff
@@ -466,29 +486,48 @@ filterValues((x) => (x [!as String!]).contains('Hello'));
...
 
 {:.fails-sa}
+<?code-excerpt "lib/... ()" replace=""?>
 {% prettify dart tag=pre+code %}
-List<int> numbers = [1, 2, 3];
-List<String> [!string = numbers!];
+void main() {
```
sfshaza2, line 505: Yeah, once you modify code in an excerpt, you have to run "./tool/refresh-excerpts" in your branch. But, if you have any questions about code excerpts, Khanh is our expert!

sfshaza2 (commented):
A few comments. And you will have to run the code_excerpts tool.

src/_guides/language/sound-problems.md
```diff
@@ -466,29 +486,48 @@ filterValues((x) => (x [!as String!]).contains('Hello'));
...
+### Incorrect type inference
 
-Dart enforces a sound type system. Roughly, this means
-you can't get into a situation where the value stored in a variable is
-different from the variable's static type. Like most modern statically
-typed languages, Dart accomplishes this with a combination of static
-(compile-time) and dynamic (runtime) checking.
+```nocode
+error - ...
+```
```
MaryaBelanger, line 493: @mit-mit @kevmoo I would like to add the actual error message that would show up with the example mentioned in the [change log](https://github.com/dart-lang/sdk/blob/master/CHANGELOG.md#language-1) to match the rest of the page. I could not find it though, so I'm wondering the best way to find relevant prs when starting from the change log. 

I followed the link in the change log to the issue, but couldn't locate the actual PR where the error message was added to the code (if that error message was actually added as part of 2.18, as it being mentioned in the change log implies).

src/_guides/language/sound-problems.md
MaryaBelanger: I definitely have a lot of questions about the excerpts haha. I'll reach out!

src/_guides/language/sound-problems.md
parlough: https://github.com/dart-lang/site-shared/blob/main/doc/writing-for-dart-and-flutter-websites.md#semantic-line-breaks

src/_guides/language/sound-problems.md
parlough: My recommendation with these is to include the code in a project and then run `dart analyze` to see the resulting error. It's a bit hard to follow the code implemented in the SDK, as errors touch a lot of pieces.

In this case, with the code you included below, I get an error like the following:

```
error - The return type '...' isn't a '...', as required by the closure's context - return_of_invalid_type_from_closure
```

Eventually, the goal is to get the code in to an example(https://github.com/dart-lang/site-www/tree/main/examples/type_system) so we can keep track of the actual error in an analysis results file(https://github.com/dart-lang/site-www/blob/main/examples/type_system/analyzer-results-stable.txt) and use the code excerpt tool when referencing it. That way if the error message updates we'll know to update this. This is a bit of a pain to set up currently though for error messages, especially ones not yet in stable, so it's fine if we add that later.

src/_guides/language/sound-problems.md
parlough, line 505: Perhaps we don't need to include the `main()` function scaffolding in these example, so it's easier to focus on the relevant code.

MaryaBelanger:
I ran the refresh excerpts script and dart analyze, but am not really sure if they worked/what I'm supposed to be looking for to indicate if they did. I haven't had a chance to study dart _at all_ yet, so this is probably as "complete" as the code snippets are going to get on my end for now. As far as I can see they're rendering acceptably on the page, hopefully that's enough to merge

sfshaza2:
@MaryaBelanger, this PR looks mostly good but, have you staged this (so that I can see it, not on localhost)? I'd like to see the staged version. Also, the CI seems happy so I guess that running code excerpts was sufficient.  :D

src/_guides/language/sound-problems.md
```diff
@@ -20,7 +20,28 @@ and see [these other resources](/guides/language/type-system#other-resources).
 
 ## Troubleshooting
 
-### No type errors {#no-type-errors}
+Dart enforces a sound type system. Roughly, this means
+you can't get into a situation where the value stored in a variable is
+different from the variable's static type. Like most modern statically
+typed languages, Dart accomplishes this with a combination of
+[static (compile-time)](#static-errors-and-warnings) and [dynamic (runtime)](#runtime-errors) checking.
```
atsansone, line 27: **suggestion:** This could be trimmed down.
```suggestion
Dart enforces a sound type system. In general terms, this means you can't write code where a variable's value differs from its static type. An `int` type can't store a number with a decimal place. Dart checks variable values against their types at [compile-time](#static-errors-and-warnings) and [runtime](#runtime-errors).
```

These changes drop readability from Grade 11 to Grade 5. Sure, it's eight more words, but they're small.

examples/type_system/lib/common_fixes_analysis.dart
```diff
@@ -131,3 +131,17 @@ void funcCast() {
...
   // #enddocregion func-cast
 }
+
+//-----------------------------------------------
+
+// #docregion type-inf-null
+  List<int> ints = [1, 2, 3];
+  var maximumOrNull = ints.fold(null,
+      (a, b) => a == null || a < b ? b : a);
+// #enddocregion type-inf-null
```
parlough, line 141: You should put these in methods as they both can't be in the global scope as their names will collide.

You also need to ignore the resulting errors of the failed type inference since they will cause code analysis to fail.

We currently enforce [`dart format`](https://dart.dev/tools/dart-format) being run on files, so they will end up looking like this too:

```suggestion
void infNull() {
  // #docregion type-inf-null
  List<int> ints = [1, 2, 3];
  // ignore: non_bool_operand, undefined_operator
  var maximumOrNull = ints.fold(null, (a, b) => a == null || a < b ? b : a);
  // #enddocregion type-inf-null    
} 
```

examples/type_system/lib/common_fixes_analysis.dart
```diff
@@ -131,3 +131,17 @@ void funcCast() {
...
+
+// #docregion type-inf-fix
+  List<int> ints = [1, 2, 3];
+  var maximumOrNull = ints.fold<int?>(null,
+      (a, b) => a == null || a < b ? b : a);
+// #enddocregion type-inf-fix
```
parlough, line 147: Same thing here: adding a method so names aren't in global scope, running `dart format`, and adding a new line at the end of the file (which we do in all files).

```suggestion
void infFix() {
  // #docregion type-inf-fix
  List<int> ints = [1, 2, 3];
  var maximumOrNull =
      ints.fold<int?>(null, (a, b) => a == null || a < b ? b : a);
  // #enddocregion type-inf-fix
}

```

src/_guides/language/sound-problems.md
```diff
@@ -466,29 +487,46 @@ filterValues((x) => (x [!as String!]).contains('Hello'));
...
+### Incorrect type inference
 
-Dart enforces a sound type system. Roughly, this means
-you can't get into a situation where the value stored in a variable is
-different from the variable's static type. Like most modern statically
-typed languages, Dart accomplishes this with a combination of static
-(compile-time) and dynamic (runtime) checking.
+```nocode
+error - The return type '...' isn't a '...', as required by the closure's context - return_of_invalid_type_from_closure
+```
```
parlough, line 495: Okay I ran the analyzer again with Dart 2.18 and 2.19 and I'm getting different errors. Perhaps I was using an older development version of Dart on that machine.

Since this has to do with type inference, it results in many different errors depending on the context, due to types flowing through the code in different ways according to each situation.

Perhaps it doesn't make a lot of sense to include a specific error due to this, and your explanation of "incorrect type inference" should be sufficient. What do you think?

```suggestion
```

src/_guides/language/sound-problems.md
```diff
@@ -466,29 +487,46 @@ filterValues((x) => (x [!as String!]).contains('Hello'));
...
+the wrong type for function literal arguments in a generic constructor invocation.
+This primarily affects `Iterable.fold`.
+
+#### Example
+
+In the following code,
+type inference will infer that `a` has a type of `Null`:
 
 {:.fails-sa}
+<?code-excerpt "lib/common_fixes_analysis.dart (type-inf-null)"?>
```
parlough, line 506: You'll need to rerun the code excerpt tool again after making changes to the excerpt files.

parlough (changes requested):
Thanks for working on this, looks like a nice addition to the page.

The build does look to be passing, but for some reason it didn't actually run any of the checks. It will fail due to analysis and formatting issues in the example file. When you push again, they should run. GitHub Actions are sometimes finicky :)

I've included some suggested fixes and some other comments. Let me know what you think or if you have any questions. Thanks!

src/_guides/language/sound-problems.md
parlough: As this PR didn't write this text and there are some particularities of these adjustments that may need further review, perhaps best to leave these changes for a separate PR. 

The whole article could likely use some readability improvements, so perhaps that could be done in a follow-up PR?

src/_guides/language/sound-problems.md
MaryaBelanger: Definitely agree, especially now when we just need to get the update out for 2.18. Thanks so much!

src/_guides/language/sound-problems.md
```diff
@@ -20,7 +20,33 @@ and see [these other resources](/guides/language/type-system#other-resources).
 
 ## Troubleshooting
 
-### No type errors {#no-type-errors}
+Dart enforces a sound type system. This means you can't write code where a
+variable's value differs from its static type.
+An `int` type can't store a number with a decimal place.
+Dart checks variable values against their types at
+[compile-time](#static-errors-and-warnings) and [runtime](#runtime-errors).
```
parlough, line 27: Clarifies it's not the `int` type that stores a value, but rather a variable.

```suggestion
Dart enforces a sound type system. 
This means you can't write code where a
variable's value differs from its static type.
A variable with an `int` type can't store a number with a decimal place.
Dart checks variable values against their types at
[compile-time](#static-errors-and-warnings) and [runtime](#runtime-errors).
```

parlough:
Oh, I see why the checks aren't running. This is targeting the v2.18 branch but we currently only have checks setup to run against the `main` branch.

If you want to test the changes locally for now, you should be able to run `make run` and then whatever test you want to run, such as `./tool/test.sh` which is the primary command we use for testing. 

You may need to merge the latest changes from the 2.18 branch into yours first.

MaryaBelanger:
I manually updated the code excerpt in the article because of issues with running the tools, but I will run the tools once I'm sure they're working correctly to confirm

PR #4198, opened 2022-08-29
2.18 .packages update

MaryaBelanger:
- Removed mention of `.packages` from "What not to commit" page *
- Added work around for third-party legacy dependencies to "`dart pub get`" page

*from my observation, the pattern is to leave notes and mentions of deprecations in the release doc updates where the concept is deprecated, then completely remove any mention once fully de-supported. So I completely removed notes/mentions from "What not to commit"


src/tools/pub/cmd/pub-get.md
```diff
@@ -145,6 +145,13 @@ By default, pub precompiles executables
 in immediate dependencies (`--precompile`).
 To prevent precompilation, use `--no-precompile`.
 
+### `--legacy-packages-file`
+
+If you utilize any third-party tools with a legacy dependency on `.packages` (desupported in [2.18](/guides/whats-new)),
+use `--legacy-packages-file` to generate a `.packages` file.
```
parlough, line 151: I would rather not link to the "What's new" page as that is more about what is new on the website, not the Dart SDK itself. Support is being removed in Dart 2.19 so perhaps we can just include the version this is useful for as a main part of the sentence?

Perhaps something like:

```suggestion
If you are using Dart 2.18 and have any third-party tools
which still use the discontinued `.packages` file,
use `--legacy-packages-file` to generate a `.packages` file.
```

src/tools/pub/cmd/pub-get.md
```diff
@@ -145,6 +145,13 @@ By default, pub precompiles executables
...
+This support will be removed in an upcoming stable release.
```
parlough, line 152: This work has now landed in Dart 2.19 so perhaps we should be explicit:

```suggestion
Support for this flag will be removed in Dart 2.19.
```

parlough (commented):
Thanks for this, looks good! I have a few small comments to consider.

src/tools/pub/cmd/pub-get.md
```diff
@@ -145,6 +145,14 @@ By default, pub precompiles executables
 in immediate dependencies (`--precompile`).
 To prevent precompilation, use `--no-precompile`.
 
+### `--legacy-packages-file`
+
+If you are using Dart 2.18 and have any third-party tools
+which still use the discontinued `.packages` file,
+use `--legacy-packages-file` to generate a `.packages` file.
+Support for this flag will be removed in Dart 2.19.
```
sfshaza2, line 153:
```suggestion
If you're using Dart 2.18 (or later) and use any third-party tools
that rely on the discontinued `.packages` file,
use `--legacy-packages-file` to generate a `.packages` file.
Support for this flag will be removed in Dart 2.19.
```

sfshaza2 (approved):
Some minor changes, otherwise, lgtm

src/tools/pub/cmd/pub-get.md
parlough: It seems the flag was removed in Dart 2.19 already, so do we need to add the "(or later)"?

sfshaza2:
Good point! Nevermind. :D

MaryaBelanger:
If everyone's happy I'll merge soon. Thanks @parlough and @sfshaza2!

mit-mit (approved):

parlough (approved):
Thanks @MaryaBelanger, looks good to me after those adjustments!

PR #4199, opened 2022-08-30
Fix link to "Dart Language Versioning" proposal

julemand101:
The link was broken after the proposal was moved from `future-releases` to `2.8` in the Dart language specification repository.


parlough (approved):
Nice catch. Thanks for fixing this!

PR #4206, opened 2022-08-30
Issue with Dart tool/test.sh changes

atsansone:
When someone runs `make build`, these files are altered and flagged as updated in `git`. This causes issues with the current commit in `git`. Can we see why this is occurring and how to fix it?


parlough:
I can look into this in a few days. 

Converted this PR to a draft to guarantee it's not merged.

PR #4208, opened 2022-09-02
Update linter rules to v1.28.0

parlough:


domesticmouse (approved):

julemand101:
The rules uploaded here is for Linter 1.28 which is going to be part of Dart 2.19.0.

But that means the overview over the available lints on: https://dart.dev/tools/linter-rules

Now shows lints that is actually not available for the latest stable version of Dart and it does not state in the lints when they become available. E.g. the following two lints that is not part of Dart 2.18:

https://dart.dev/tools/linter-rules#use_string_in_part_of_directives
https://dart.dev/tools/linter-rules#unreachable_from_main

It is possible to get more clear information if we go to:
https://dart-lang.github.io/linter/lints/use_string_in_part_of_directives.html
https://dart-lang.github.io/linter/lints/unreachable_from_main.html

But it just tells us what version of `linter` where the lint became available and it is not really expected of a user to know what version of `linter` is part of a given version of Dart (or Flutter).

So where should a user go if they want to know the exact list of available lints for the latest stable release of Dart?

parlough:
You're right, this is definitely a source of confusion for readers. 

The SDK information used to be included on the standalone linter website but had some maintenance issues (https://github.com/dart-lang/linter/issues/2824). It's on my to-do list to bring back for both sites and hopefully I'll get to it soon or someone else will and I can incorporate that information on this end.

PR #4212, opened 2022-09-06
Fixed grammar

dark-chocolate:


parlough (approved):
Thanks for this!

This makes it consistent with the other codelabs as well as usages on flutter/website.

PR #4233, opened 2022-09-25
Added equality operator to fix assertion error

iLoveDocs:
It's equally important to override `==` operator for the code to work otherwise it will give assertion failed error.

### DIFF
--- a/examples/misc/lib/language_tour/classes/vector.dart
+++ b/examples/misc/lib/language_tour/classes/vector.dart
@@ -7,18 +7,17 @@ class Vector {
   Vector operator +(Vector v) => Vector(x + v.x, y + v.y);
   Vector operator -(Vector v) => Vector(x - v.x, y - v.y);
 
-  // Operator == and hashCode not shown.
-  // #enddocregion ''
   @override
   bool operator ==(Object other) =>
       other is Vector && x == other.x && y == other.y;
 
   @override
-  int get hashCode => 37 * (629 + x.hashCode) + y.hashCode;
+  int get hashCode => Object.hash(x, y);
 
+  // #enddocregion ''
   @override
   String toString() => '($x, $y)';
-  // #docregion ''
+// #docregion ''
 }
 
 void main() {
--- a/examples/misc/lib/language_tour/classes/vector.dart
+++ b/examples/misc/lib/language_tour/classes/vector.dart
@@ -1,4 +1,3 @@
-// #docregion ''
 class Vector {
   final int x, y;
 
@@ -13,11 +12,6 @@ class Vector {
 
   @override
   int get hashCode => Object.hash(x, y);
-
-  // #enddocregion ''
-  @override
-  String toString() => '($x, $y)';
-// #docregion ''
 }
 
 void main() {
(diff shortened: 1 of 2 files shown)


parlough (changes requested):
Thanks for opening a pull request. I can definitely see where the problem could arise if someone tries to run this snippet.

Since this example is focused on the `-` and `+` operators, these implementations were excluded to keep them focused on the topic at hand. Since a `main` method is included though, it may make it seem runnable, so having an extra few lines shouldn't hurt. It's nice to show an example of overriding `==` as well.

You'll also have to adjust the source of this sample in https://github.com/dart-lang/site-www/blob/main/examples/misc/lib/language_tour/classes/vector.dart

```dart
class Vector {
  final int x, y;

  Vector(this.x, this.y);

  Vector operator +(Vector v) => Vector(x + v.x, y + v.y);
  Vector operator -(Vector v) => Vector(x - v.x, y - v.y);

  @override
  bool operator ==(Object other) =>
      other is Vector && x == other.x && y == other.y;

  @override
  int get hashCode => Object.hash(x, y);

  // #enddocregion ''
  @override
  String toString() => '($x, $y)';
  // #docregion ''
}
```

Perhaps can switch the hashCode to `Object.hash` at the same time to keep consistent with the recommendation to use that.

Then you can update the snippet in the Markdown file to match or run the [refresh-code-excerpts tool](https://github.com/dart-lang/site-www/blob/main/tool/refresh-code-excerpts.sh).

Let me know if you need any help :)

iLoveDocs:
@parlough 

Thanks, I made the changes in that file and created a new [pull request](https://github.com/dart-lang/site-www/pull/4243) for the same. 

> Then you can update the snippet in the Markdown file to match or run the [refresh-code-excerpts tool](https://github.com/dart-lang/site-www/blob/main/tool/refresh-code-excerpts.sh).

But now I don't know what else to do. I made the changes in this pull request's file but I don't know where they are, I still see the original `hashCode` but when I click the `Edit File` button right next to the file, I see the updated code but not sure how to create second pull request of this pull request. Nor I have any idea about the `refresh code excerpts` thing you shared.  I am too dumb for Github.

parlough:
Thanks for looking into that @iLoveDocs! I'm not completely sure about the in-browser GitHub workflow, but the goal would be to add the other changes to the same branch ([`patch-2`](https://github.com/iLoveDocs/site-www/tree/patch-2)) as this one but with a second commit rather than pull request. This would allow the changes to be in the same pull request, so that they can pass the checks.

I'll go ahead and locally push the other changes to this branch and refresh code excerpts for you. If you're curious, refresh code excerpts is a custom tool we have that allows the source files for samples and their representations in the Markdown files be guaranteed to be up to date. It can be a bit of a pain to set up our site infrastructure locally though, so I can handle that for you. For smaller changes, you can always just make the changes in both locations instead of running the tool.

Thanks again for your help and adjusting this to make it clear for everyone!

parlough (approved):
Thanks for noticing the potential confusion here and fixing it. Looks good to me after some code-excerpt adjustments :)

iLoveDocs:
@parlough 

> the goal would be to add the other changes to the same branch ([patch-2](https://github.com/iLoveDocs/site-www/tree/patch-2)) as this one but with a second commit rather than pull request.

Ahh, now I understood what you meant. I didn't know I can make changes in the same pull request and do a second commit. And thanks for explaining the refresh code excerpt part.

PR #4251, opened 2022-09-30
[WIP] Show release date on archive

### LINKED ISSUE #2389: Add date to SDK archive page?
Page URL: https://dart.dev/tools/sdk/archive/
Page source: https://github.com/dart-lang/site-www/tree/master/src/tools/sdk/archive/index.md

https://flutter.dev/docs/development/tools/sdk/releases?tab=macos has dates for each release, which is handy (e.g. when updating wikipedia, or when deciding whether you should download the SDK again). It'd be nice to have release dates on the Dart page. We might consider aligning the pages in other ways, too.

/cc @sfshaza2 @johnpryan @domesticmouse @mit-mit

parlough:
Closes #2389

### DIFF
--- a/src/get-dart/archive/_archives_table.html
+++ b/src/get-dart/archive/_archives_table.html
@@ -17,8 +17,8 @@
   <thead>
     <tr>
       <th>Version</th>
-      <th>OS</th>
       <th>Architecture</th>
+      <th>Release date</th>
       <th>Downloads</th>
     </tr>
   </thead>
@@ -31,8 +31,8 @@
       {% else %}
       <td>0.0.0-0.0.dev (rev 00000)</td>
       {% endif %}
-      <td>Windows</td>
       <td>64-width</td>
+      <td>01/01/1970</td>
       <td>JSON-formatted API documentation</td>
     </tr>
   </tbody>
--- a/src/get-dart/archive/assets/download_archive.dart.js
+++ b/src/get-dart/archive/assets/download_archive.dart.js
@@ -179,7 +179,7 @@ c.Y5(a3,q,o)}q=m
 break}}A.ZE(a3,r,q,a6)}else A.ZE(a3,r,q,a6)},
 ix:function ix(a,b){this.a=a
 this.$ti=b},
-pg:function pg(a,b,c){var _=this
+rK:function rK(a,b,c){var _=this
 _.a=a
 _.b=b
 _.d=_.c=null
@@ -251,7 +251,7 @@ this.$ti=b},
 JB:function JB(a,b){this.a=a
 this.$ti=b},
 SU:function SU(){},
-Re:function Re(){},
+Ja:function Ja(){},
 w2:function w2(){},
 iK:function iK(a,b){this.a=a
 this.$ti=b},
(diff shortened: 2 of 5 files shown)


domesticmouse (approved):

mit-mit:
@athomas can you say a little about how the creation date on these archives compares to the actual release date? My concern is that they may actually on occasion be days apart?

athomas:
> @athomas can you say a little about how the creation date on these archives compares to the actual release date? My concern is that they may actually on occasion be days apart?

We have three folders for each release: `raw` (just built, the CI infra puts binaries here), `signed` (signing tooling puts signed binaries there) and `release`.

So the binaries in there will be available on these days:
`raw`: On the day we merge/CP as soon as the builds finish.
`signed`: On the day we publish (though we want to automate this process and then they will be ready soon after the `raw` builds).
`release`: On the day we publish, we move binaries from `signed` (and until we have signing on all platforms from `raw`) to `release`.

We don't really have a "creation time" in GCS, but because we don't overwrite releases modification time is essentially a creation time:
https://cloud.google.com/storage/docs/metadata#modification-time

The biggest risk is that we modify the metadata somehow (e.g. set `Cache-Control`) and we end up trashing all the dates. We'd have to use `Custom-Time` or something not built into GCS to avoid that.

Btw, I see dates on https://dart-dev--pr4251-feature-release-date-xseot5xf.web.app/get-dart/archive.

mit-mit:
> As for the missing release dates, it looks to just be the listing for the API docs, but I could add the release date to that row too if desired.

@parlough  no, it's more fundamentally broken for me. This is e.g. the full table for 2.18.2:

Screen Shot 2022-10-03 at 10 14 20

I looked in the error console in devtools, and it has:

```
download_archive.dart.js:2147 Uncaught LocaleDataException: Locale data has not been initialized, call initializeDateFormatting(<locale>).
    at Object.J (https://dart-dev--pr4251-feature-release-date-xseot5xf.web.app/get-dart/archive/assets/download_archive.dart.js:350:3)
    at kH.tl (https://dart-dev--pr4251-feature-release-date-xseot5xf.web.app/get-dart/archive/assets/download_archive.dart.js:7657:14)
    at z.t2 (https://dart-dev--pr4251-feature-release-date-xseot5xf.web.app/get-dart/archive/assets/download_archive.dart.js:3297:24)
    at Object.fm (https://dart-dev--pr4251-feature-release-date-xseot5xf.web.app/get-dart/archive/assets/download_archive.dart.js:3334:25)
    at Object.fm (https://dart-dev--pr4251-feature-release-date-xseot5xf.web.app/get-dart/archive/assets/download_archive.dart.js:3334:10)
    at Object.FI (https://dart-dev--pr4251-feature-release-date-xseot5xf.web.app/get-dart/archive/assets/download_archive.dart.js:3290:14)
    at C.PS (https://dart-dev--pr4251-feature-release-date-xseot5xf.web.app/get-dart/archive/assets/download_archive.dart.js:7193:23)
    at https://dart-dev--pr4251-feature-release-date-xseot5xf.web.app/get-dart/archive/assets/download_archive.dart.js:7119:10
    at Gs.a (https://dart-dev--pr4251-feature-release-date-xseot5xf.web.app/get-dart/archive/assets/download_archive.dart.js:1462:62)
    at Gs.$2 (https://dart-dev--pr4251-feature-release-date-xseot5xf.web.app/get-dart/archive/assets/download_archive.dart.js:4781:14)
```

I'm using Chrome 105.0.5195.125 on macOS.

parlough:
Thanks for including the error. Looks like I need to set a backup locale.

parlough:
Thanks for the heads up. I pushed a fix so non-default locales are also initialized. You'll likely need to disable the browser's cache to test it for now. Let me know if you have any other issues.

Alternative locale formatted dates

mit-mit:
Yay, working now after a shift-reload!

mit-mit:
> It may not match the release in terms of the blog post, but it did for most releases I looked at.

Yeah, I checked back to Dart 2.12 and they all matches except Dart 2.13 that had one day earlier in the archive.  I think this is close enough!

parlough:
@mit-mit I made the date format change, how does that look?

mit-mit:
Looks great, thanks!

lgtm_stamp_by_nushe4ka_d5w539d-fullview

PR #4272, opened 2022-10-07
Effective Dart: update type promotion suggestion for private final

MaryaBelanger:
- Added private final fields to the "promotion is only sound for..." sentence
- Expanded on the "...so fields and top-level variables aren't promoted" clause 
  - To me it's a little more nuanced than just "so fields _that aren't private and final_ and top-level variables aren't promoted", hence the longer explanation of what's not promoted and _why_. 
- Added caveat before the main example that if your field is private and final, which can be concluded as the recommended default from the Effective Dart: Design page, that is also a technically a workaround to nullable fields not being type promoted. 

I don't want to disrupt the purpose of that page, so please let me know if the drawn out explanation there distracts from the recommendation in that section!

------
_Note for future me:_
I admit this is a weird spot to essentially define type promotion, but it's kind of the only/clearest place that already does so in the docs... (maybe I'm wrong). 

There's [Type promotion on null checks](https://dart.dev/null-safety/understanding-null-safety#type-promotion-on-null-checks) and [Working with nullable fields](https://dart.dev/null-safety/understanding-null-safety#working-with-nullable-fields) in Understanding null safety. Based on the page title it seems like that'd be a better spot for a type promotion definition. 

But the actual definition of what is and isn't possible with type promotion, in both those sections, is buried in the narrative style of the examples in each section. I.e. just clicking one of those links and arriving on the page at that section doesn't define type definition clearly for a reader who lands there without context. I will be updating those sections as well, though.

### DIFF
--- a/src/_guides/language/effective-dart/usage.md
+++ b/src/_guides/language/effective-dart/usage.md
@@ -354,13 +354,20 @@ then it probably does make sense to have a separate boolean field.
 
 Checking that a nullable variable is not equal to `null` promotes the variable
 to a non-nullable type. That lets you access members on the variable and pass it
-to functions expecting a non-nullable type. Unfortunately, promotion is only
-sound for local variables and parameters, so fields and top-level variables
-aren't promoted.
+to functions expecting a non-nullable type. 
 
-One pattern to work around this is to assign the field's value to a local
-variable. Null checks on that variable do promote, so you can safely treat
-it as non-nullable.
+Type promotion is only sound, however, for local variables, parameters, and
+private final fields. Instances that are open to manipulation, whether
+explicitly, like top-level variables, or implicitly, like fields with a concrete
+getter of the same name in the same library, cannot be type promoted.
+
+Declaring members [private] and [final] as we generally recommend would bypass
+these limitations, but that's not always an option. One pattern to work around
+this is to assign the field's value to a local variable. Null checks on that
+variable will promote, so you can safely treat it as non-nullable.
+
+[private]: design#prefer-making-declarations-private
+[final]: design#prefer-making-fields-and-top-level-variables-final
 
 {:.good}
 <?code-excerpt "usage_good.dart (shadow-nullable-field)"?>
@@ -384,7 +391,7 @@ class UploadException {
 {% endprettify %}
 
 Assigning to a local variable can be cleaner and safer than using `!` every
-place the field or top-level variable is used:
+place the instance is used:
 
 {:.bad}
 <?code-excerpt "usage_bad.dart (shadow-nullable-field)" replace="/!\./[!!!]./g"?>


stereotype441 (commented):
Let's hold off on this for a bit.  I need to talk to folks about whether this feature is going to be fully turned on in 2.19 or just behind an experimental flag.  If it's just going to be behind an experimental flag, then we shouldn't update documentation yet.

munificent (approved):
Looks great!

src/_guides/language/effective-dart/usage.md
```diff
@@ -354,13 +354,20 @@ then it probably does make sense to have a separate boolean field.
...
-it as non-nullable.
+Type promotion is only sound, however, for local variables, parameters, and
+private final fields. Instances that are open to manipulation, whether
+explicitly, like top-level variables, or implicitly, like fields with a concrete
+getter of the same name in the same library, cannot be type promoted.
+
+Declaring members [private] and [final] as we generally recommend would bypass
+these limitations, but that's not always an option. One pattern to work around
+this is to assign the field's value to a local variable. Null checks on that
+variable will promote, so you can safely treat it as non-nullable.
```
atsansone, line 367: **suggestion:** Tighten some of the language. The sentences run a little long.

```suggestion
You can _soundly_ promote the types of local variables, parameters, and private final fields only.
You cannot promote instances that can be manipulated.
This potential manipulation can be explicit, like top-level variables, or implicit,
like fields with a concrete getter of the same name in the same library.

To bypass these limitations, try one of the following options:

- Declare members [private] and [final].
- Assign the field's value to a local variable. Null checks on that
variable promote, so you can safely treat it as non-nullable.
```

src/_guides/language/effective-dart/usage.md
```diff
@@ -384,7 +391,7 @@ class UploadException {
 {% endprettify %}
 
 Assigning to a local variable can be cleaner and safer than using `!` every
-place the field or top-level variable is used:
+place the instance is used:
```
atsansone, line 394: **suggestion:** Clean up the language some to keep it concise.

```suggestion
Assign the value to a local scope variable. This reduces the chance for accidental errors and the length of the code.
```

atsansone (changes requested):
@MaryaBelanger : Added some suggestions to tighten the language.

src/_guides/language/effective-dart/usage.md
MaryaBelanger: I'm gonna skip direct-committing this because there's a lot of nuance here that gets lost in these suggestions.

For example, "You can soundly promote..." and "You cannot promote..."; type promotion is handled by the compiler, so it's odd to attribute it to the reader/user. I guess you can technically write your code in a way that enables type promotion, but "you" cannot promote instances. 

Also, the paragraph starting on line 364 intentionally separates the points you combined into a bullet list. While they both technically make type promotion "work", they are conceptually different. The first point, declaring members private and final, 
1. is not covered in the example that immediately follows, nor is it the point of this section at all, and 
2. is what we want users to be doing all the time as a best practice. 

What I intended this paragraph to say was "if there's some reason you _can't_ use the best practice we want you to be using all the time, here's a different way to make type promotion work", not "here's two ways to make type promotion work". I think I'll just completely remove the first sentence though, seeing now how it can be misconstrued. 

I definitely agree the long sentence starting on line 360 is not very concise though (I spent a long, _long_ time getting these concepts compressed into a few sentences, so I do understand the immediate reaction to the wordiness). I think I'll end up moving that level of detail on type promotion constraints to "Understanding null safety" and just get rid of it all together here.

src/_guides/language/effective-dart/usage.md
MaryaBelanger: I think the original sentence is as concise as possible already, so I'm gonna skip this suggestion too.
- "Local scope variable" is redundant; "local" is intrinsic to scope. 
- "Accidental errors" also feels redundant; errors are typically unintentional. 

And again, some nuance is lost. "Length of code" is not the concern when it says using the `!` operator is not clean; if you compare the good and bad example, the length of the code is basically not affected at all. The "cleanness" concern here is, imo, about the excessive complexity of comprehending the null assertion operator so many times in a single statement.

I also don't think this line needs to be written in the imperative ("assigning" to "assign"). The instruction is already given on line 366: "_...assign the field's value to a local variable_". Line 393 is just commenting on that; basically saying "doing this is better than <bad example> because..."

MaryaBelanger:
Closing this to re-do on a branch right off of main, plus a lot has changed since I wrote this so there will probably be significant changes

PR #4274, opened 2022-10-08
Mention super parameters with initializing formals

### LINKED ISSUE #4262: Effective Dart: Don't type annotate initializing super formals
### Page URL

https://dart.dev/guides/language/effective-dart/design#dont-type-annotate-initializing-formals

### Page source

_No response_

### Describe the problem

Effective Dart has a rule, "DON’T type annotate initializing formals." which should be extended to cover super parameters.

### Expected fix

I don't know if a super parameter is an "initializing formal" or a new construct. So either we need a new rule, change the name of this rule, or just add some text in this rule.

### Additional context

_No response_

natebosch:
Fixes #4262

Dart 2.17.0 introduced a syntax for implicitly forwarding constructor arguments to the super constructor. The type is inferred in the same way as initializing formals, so expand the section to include mention of `super.` syntax, and an example.

### DIFF
--- a/examples/misc/lib/effective_dart/design_bad.dart
+++ b/examples/misc/lib/effective_dart/design_bad.dart
@@ -8,6 +8,13 @@ import 'package:examples_util/ellipsis.dart';
 
 import 'design_good.dart';
 
+class Key {}
+
+class StatelessWidget {
+  final Key? key;
+  StatelessWidget({this.key});
+}
+
 void miscDeclAnalyzedButNotTested() {
   (errors, monsters, subscription) {
     // #docregion code-like-prose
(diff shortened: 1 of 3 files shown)


parlough (approved):
Looks great to me, thanks!

I pushed a change adding the new snippets to the example files they get pulled from, which our excerpt tool checks during the build.

natebosch:
@srawlins - do you think it would be better to separate this into it's own rule for alignment with the different lints? Or do we think it's OK for two lints to both reference this rule?

parlough:
I think it should be fine to have them as the same rule. We have other rules which correspond to multiple lints, such as [DON’T use `.length` to see if a collection is empty](https://dart.dev/guides/language/effective-dart/usage#dont-use-length-to-see-if-a-collection-is-empty).

parlough:
I'm going to merge this in. If there are other requested changes we can make those later. I'll also add the lint to the new lint (if it is separate) once it's added. Thanks!

munificent (commented):
LGTM!

PR #4277, opened 2022-10-12
Remove link to unmaintained and outdated translated site

parlough:
The site is unmaintained, outdated, and doesn't have much translated currently.

We can always add it back in the future if the site or another equivalent is maintained again.

### DIFF
--- a/src/community/index.md
+++ b/src/community/index.md
@@ -72,7 +72,7 @@ Learn how to
 
 ## Additional community resources
 
-Our wonderful community has provided these resources.
+Our wonderful community has provided these resources:
 
 * [Chinese version of this site (此网站的中文版)](https://dart.cn)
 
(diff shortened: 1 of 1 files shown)


src/community/index.md
```diff
@@ -75,7 +75,6 @@ Learn how to
 Our wonderful community has provided these resources.
```
MaryaBelanger, line 75:
```suggestion
Our wonderful community has provided these resources:
```
Idk if it's bad practice to add random punctuation changes to a PR with a specific purpose like this, but I just noticed a colon could be better here. Feel free to include or not!

MaryaBelanger (approved):

src/community/index.md
parlough: Doesn't hurt with such a small (and slightly related change)!

MaryaBelanger (approved):
Thanks!

PR #4280, opened 2022-10-13
Update some necessary package 

ragu-1997:
nano - editor for developer use
linconfig-dev -dev tool
wget - getting some information via websites
clean - clean cache 

all are some necessary packages

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,6 +15,7 @@ RUN apt update && apt install -yq --no-install-recommends \
       libconfig-dev \
       wget \
     && rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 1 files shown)


parlough:
I agree with @domesticmouse. We want the Docker image to be as slim as possible. If someone needs one of these extra packages in that environment, which seems unlikely for our use cases, they can install it themselves.

As for the `apt-get clean`, official Debian images which our Ruby image depends on, already run this automatically, so we don't need to run it explicitly. 

If you feel any of these are needed for everyone working on the site in the future, let us know though, thanks again for the PR!

PR #4284, opened 2022-10-15
Fix broken code excerpts showing up on library tour

parlough:


domesticmouse (approved):

domesticmouse:
Any idea why CI is failing?

parlough:
I forgot to enable unicode in the regex, but I couldn't seem to be able to with the syntax code-excerpt requires, so I went another direction for now.

Issue #4288, opened 2022-10-17
[PAGE ISSUE]: 'The Dart type system'


iLoveDocs:
### Page URL

https://dart.dev/guides/language/type-system

### Page source

https://github.com/dart-lang/site-www/tree/main/src/_guides/language/type-system.md

### Describe the problem

In the last paragraph of [Generic type assignment](https://dart.dev/guides/language/type-system#generic-type-assignment) 

There's a line

> To make this code pass static analysis, use an explicit cast, which might fail at runtime.
> `List<Cat> myCats = <Animal>[] as List<Cat>;`

Even if you provide `Cat` in the list of `Animal`, the code will still fail at runtime. So, there is no such thing as 
> might fail at runtime

### Expected fix

We need to either change the following sentence. 
> which might fail at runtime

or provide a solution of using casting, like: 

```dart
List<Cat> myCats = (<Animal>[]).cast(); // or List.from or List.castFrom
```

parlough:
Hey @iLoveDocs, thanks for opening an issue. I believe this section is saying how this cast enables the code to pass static analysis but this type of explicit cast (not this specific situation) may fail at runtime, rather than about how to avoid potential runtime errors.

Perhaps we could be more explicit it's talking about this general case?

iLoveDocs:
@parlough 

Yes, I understand the context but I think your this line makes more sense than the the current line in the docs: 

> but this type of explicit cast may fail at runtime

PR #4289, opened 2022-10-18
Make explicit downcast section more releastic

parlough:
Updates the example and wording to more clearly specify this is just an example situation of when a runtime exception may occur, and add ambiguity as to the actual value of the casted list to showcase that it may or may not fail.

Staged: https://dart-dev--pr4289-fix-4288-rx9byyx1.web.app/guides/language/type-system#generic-type-assignment

Fixes https://github.com/dart-lang/site-www/issues/4288

### DIFF
--- a/src/_guides/language/type-system.md
+++ b/src/_guides/language/type-system.md
@@ -448,9 +448,10 @@ In the following example, you can assign a `MaineCoon` list to `myCats` because
 `List<MaineCoon>` is a subtype of `List<Cat>`:
 
 {:.passes-sa}
-<?code-excerpt "lib/strong_analysis.dart (generic-type-assignment-MaineCoon)" replace="/MaineCoon/[!$&!]/g"?>
+<?code-excerpt "lib/strong_analysis.dart (generic-type-assignment-MaineCoon)" replace="/<MaineCoon/<[!MaineCoon!]/g"?>
 {% prettify dart tag=pre+code %}
-List<Cat> myCats = <[!MaineCoon!]>[];
+List<[!MaineCoon!]> myMaineCoons = ...
+List<Cat> myCats = myMaineCoons;
 {% endprettify %}
 
 What about going in the other direction? 
--- a/src/_guides/language/type-system.md
+++ b/src/_guides/language/type-system.md
@@ -457,9 +457,9 @@ What about going in the other direction?
 Can you assign an `Animal` list to a `List<Cat>`?
 
 {:.fails-sa}
-<?code-excerpt "lib/strong_analysis.dart (generic-type-assignment-Animal)" replace="/Animal/[!$&!]/g"?>
+<?code-excerpt "lib/strong_analysis.dart (generic-type-assignment-Animal)" replace="/<Animal/<[!Animal!]/g"?>
 {% prettify dart tag=pre+code %}
-List<[!Animal!]> myAnimals = ...;
+List<[!Animal!]> myAnimals = ...
 List<Cat> myCats = myAnimals;
 {% endprettify %}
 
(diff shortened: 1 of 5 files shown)


parlough:
@iLoveDocs What do you think about these changes?

iLoveDocs:
@parlough First time writing this  

**LGTM** 🥇

But there's some code which went over my head, for example: 

```dart
List<Animal> myAnimals = ellipsis();
```

parlough:
Thanks for checking @iLoveDocs! We use `ellipsis()` in the `.dart` file so it can be proper code, but our code excerpt tool converts that to `...` in the actual Markdown file.

src/_guides/language/type-system.md
```diff
@@ -475,13 +476,14 @@ which is disallowed from non-`dynamic` types such as `Animal`.
   in the [analysis options file.][analysis]
 {{site.alert.end}}
 
-To make this code pass static analysis, 
-use an explicit cast, 
-which might fail at runtime.
+To make this type of code pass static analysis, 
+you can use an explicit cast, 
+but it might fail at runtime.
```
MaryaBelanger, line 481:
```suggestion
To make this type of code pass static analysis, 
you can use an explicit cast. 
```
I think the example code should come directly after this line, so the reader can directly pair the syntax they're seeing with the phrase "explicit cast".  Why it might fail still is irrelevant to understanding the code sample.

src/_guides/language/type-system.md
```diff
@@ -475,13 +476,14 @@ which is disallowed from non-`dynamic` types such as `Animal`.
...
 
 <?code-excerpt "lib/strong_analysis.dart (generic-type-assignment-implied-cast)" replace="/as.*(?=;)/[!$&!]/g"?>
 {% prettify dart tag=pre+code %}
-List<Cat> myCats = <Animal>[] [!as List<Cat>!];
+List<Animal> myAnimals = ...
+List<Cat> myCats = myAnimals [!as List<Cat>!];
 {% endprettify %}
```
MaryaBelanger, line 487:
```suggestion
List<Animal> myAnimals = ...
List<Cat> myCats = myAnimals [!as List<Cat>!];
{% endprettify %}

An explicit cast might still fail at runtime, though,
depending on the actual value of the casted list (`myAnimals`).
```

- "why" this might fail is unclear, should be elaborated on a little more, basically how you said it in the PR description
- the fact that it might still fail at run time is kind of an kind of an aside from the point of the section (generic type, downcasting) and from the code itself (passing static analysis with implicit downcast), so it can come last

src/_guides/language/type-system.md
```diff
@@ -457,9 +457,10 @@ What about going in the other direction?
 Can you assign an `Animal` list to a `List<Cat>`?
 
 {:.fails-sa}
-<?code-excerpt "lib/strong_analysis.dart (generic-type-assignment-Animal)" replace="/Animal/[!$&!]/g"?>
+<?code-excerpt "lib/strong_analysis.dart (generic-type-assignment-Animal)" replace="/<Animal/<[!Animal!]/g"?>
 {% prettify dart tag=pre+code %}
-List<Cat> myCats = <[!Animal!]>[];
+List<[!Animal!]> myAnimals = ...
```
MaryaBelanger, line 462: Maybe not following the example, but should you add a similar declaration for the `MaineCoon` list above line 453? Does that help with clarity also?

parlough:
Thanks for the great feedback @MaryaBelanger! I like the idea of putting the warning after the example. How does it look now?

MaryaBelanger (approved):
LGTM!

Issue #4267, opened 2022-10-06
Expand on the documentation for callabe classes


parlough:
### Page URL

https://dart.dev/guides/language/language-tour#callable-classes

### Page source

https://github.com/dart-lang/site-www/tree/main/src/_guides/language/language-tour.md

### Describe the problem

We used to have an article called [Emulating functions in Dart](https://web.archive.org/web/20190402210311/www.dartlang.org/articles/language/emulating-functions), but now have very limited documentation.

### Expected fix

We should analyze relevant portions and incorporate what's still relevant, including use cases for the construct.

### Additional context

Related issues:
- https://github.com/dart-lang/site-www/issues/2071
- https://github.com/dart-lang/site-www/issues/1252

MahmoudShawkyy:
Hello @parlough I am working on this Part in Docs, and I will submit a PR after finishing it.

Issue #2071, opened 2019-11-05
Language tour should clarify relationship between callable classes and functions


mraleph:
In Dart 2 we changed how callable classes work

```dart
class WannabeFunction {
  call(String a, String b, String c) => '$a $b $c!';
} 

typedef void RealFunction(String a, String b, String c);

main() {
  var wf = new WannabeFunction();
  var out = wf("Hi","there,","gang");
  RealFunction rf = wf;  // since Dart 2 this means: rf = wf.call
  print('$out');
}
```

This is not mentioned in the language tour and might be confusing. 

We used to have the whole article about emulating functions on the site - but it seems to have been deleted?

kwalrath:
Yep, I should fix that part of the language tour.

About the emulating functions article... It looks like the latest version is saved here: https://web.archive.org/web/20190402210311/www.dartlang.org/articles/language/emulating-functions

When we did the dart.dev work we removed a bunch of low-readership articles that had no owners and/or source code tests. (Having everything be tested will be important as we switch to non-nullable types.) Also, taking a quick look at the article, I suspect it makes some claims about the design philosophy of Dart that are no longer true.

@mraleph if you think the topics in this article still belong on dart.dev, and you're willing to mark up the article and provide test files for the code, I'd be happy to apply those changes and publish the content (somewhere) on dart.dev.

kwalrath:
In case we decide to reinstate the article's content, the discussion/links here might be handy: #600

atsansone:
This work will be handled in https://github.com/dart-lang/site-www/issues/4267. Closing.

PR #4301, opened 2022-10-20
 Expanding on the documentation for callable classes

MahmoudShawkyy:
I have added explanation and some examples with showing expected errors when dealing with callable classes


parlough:
Thanks for opening a PR @MahmoudShawkyy! Actually, I think the previous example and surrounding text was already good for what's covered here for a few reasons:

1. We want the text to read more as a reference than a tutorial that gets right to the point.
2. I don't think we need to call out examples with and without parameters. We want the section to be pretty focused since the language tour is long already, so I think having the first example showcase this should be sufficient since the normal function documentation covers all the functionality functions support. If we want to be more explicit in how it supports return types, parameters, etc, we should just say that and link to the the "Functions" section where they can learn more about that functionality. Especially since this a more advance topic and readers likely are already familiar with functions.
3. The same goes for the added note, since the same limitation applies to all functions in Dart.
4. Also before, this was a runnable DartPad which allowed users to run it and see what was happening. That's super useful and not something we want to lose.

In https://github.com/dart-lang/site-www/issues/4267, there are three main goals I was thinking:

1. Make the implicit tear-off behavior clear, that is discussed more here (https://github.com/dart-lang/site-www/issues/2071).
2. Document what use cases this is useful/important in, rather than just what it does. 
3. Perhaps more explicitly link to the [Functions](https://dart.dev/guides/language/language-tour#functions) section with an indication that the same functionality is supported.

Perhaps point 3 is the most important to handle and should cover the changes you made here. It could look something like this with just some slight modifications from the original text:

```
To allow an instance of your Dart class to be called like a function,
implement the `call()` method.

The `call()` method is a special method
that allows any class that defines ones to emulate a function,
while supporting the same functionality as normal [functions](#functions)
such as parameters and return types.

In the following example, the `WannabeFunction` class defines a `call()` function
that takes three strings and concatenates them, separating each with a space,
and appending an exclamation. Click **Run** to execute the code.
```

What do you think about keeping the current combined example and just modifying the text like above to make it more explicit how it supports that other functionality? Thanks again!

MahmoudShawkyy:
Thanks @parlough I have reviewed the text and examples and I have added your modified suggested code to the the callable classes section

src/_guides/language/language-tour.md
```diff
@@ -4544,13 +4544,17 @@ Iterable<int> naturalsDownFrom(int n) sync* {
...
 ## Callable classes
 
 To allow an instance of your Dart class to be called like a function,
 implement the `call()` method.
 
-In the following example, the `WannabeFunction` class defines a call() function
+The `call()` method is a special method
+that allows any class that defines it to emulate a function,
+while supporting the same functionality as normal [functions](#functions)
+such as parameters and return types.
```
atsansone, line 4555: **issue:** Tighten the language. The `...is a special method that...` does not aid in reader understanding. I would also cut this into shorter sentences.

```suggestion
The `call()` method allows any class that defines it to emulate a function.
This method supports the same functionality as normal [functions](#functions)
such as parameters and return types.
```

atsansone (dismissed):
@MahmoudShawkyy : I would tighten up the language some in this change.

parlough (approved):
Thanks @MahmoudShawkyy! I committed the suggested changes and this looks good to me.

PR #4302, opened 2022-10-20
Adjust anonymous function examples to not trigger lint

### LINKED ISSUE #3667: Tour page has examples that get lint warning: Avoid using `forEach` with a function literal.dartavoid_function_literals_in_foreach_calls
### Page URL

https://dart.dev/guides/language/language-tour

### Page source

_No response_

### Describe the problem

I use the recommended lints on VSCode for Dart (default setting) and I see this lint `Avoid using 'forEach' with a function literal.dartavoid_function_literals_in_foreach_calls` for the following code:

> const list = ['apples', 'bananas', 'oranges'];
> list.forEach((item) {
>   print('${list.indexOf(item)}: $item');
> });

And in the same section, the following code gets the same warning:

>  list.forEach(
>       (item) => print('${list.indexOf(item)}: $item'));

### Expected fix

I think recommended code should pass the recommended linter.  For these examples, maybe change to for..in?

### Additional context

_No response_

parlough:
This is especially important as we show the example in DartPad which warns about the triggered lint.

Also expands on the example by offering a second use case beyond forEach.

**Staged:** https://dart-dev--pr4302-fix-3667-pe6ip9wh.web.app/guides/language/language-tour#anonymous-functions

Fixes #3667

### DIFF
--- a/examples/misc/test/language_tour/control_flow_test.dart
+++ b/examples/misc/test/language_tour/control_flow_test.dart
@@ -1,5 +1,3 @@
-// ignore_for_file: avoid_function_literals_in_foreach_calls
-
 import 'package:test/test.dart';
 import 'package:examples_util/print_matcher.dart' as m;
 
@@ -21,7 +19,10 @@ void main() {
       for (var i = 0; i < 2; i++) {
         callbacks.add(() => print(i));
       }
-      callbacks.forEach((c) => c());
+
+      for (final c in callbacks) {
+        c();
+      }
       // #enddocregion for-and-closures
     }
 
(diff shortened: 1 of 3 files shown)


domesticmouse (approved):

atsansone:
@parlough : For those starting out, can you explain what triggers automatic linting in Dart?

domesticmouse:
> @parlough : For those starting out, can you explain what triggers automatic linting in Dart?

Your IDE usually. On the command line `dart analyze` and `flutter analyze` are the manual approaches.

parlough:
Like @domesticmouse mentioned, usually through your IDE, but [linter rules](https://dart.dev/tools/linter-rules) have to be specifically [enabled](https://dart.dev/guides/language/analysis-options#individual-rules). 

In the case of the rule being fixed here, [`https://dart.dev/tools/linter-rules#avoid_function_literals_in_foreach_calls`](https://dart.dev/tools/linter-rules#avoid_function_literals_in_foreach_calls), it is part of the [`package:lints` recommended set](https://dart.dev/tools/linter-rules#lints) which is included by default in projects by `dart create` and used in DartPad, so it's best to avoid our example code triggering it.

PR #4304, opened 2022-10-20
Remove now unnecessary glibc version mention

parlough:
Our Linux description marks Ubuntu LTS under standard support as the earliest supported release, but Ubuntu 18 (the current oldest LTS in standard support) already has glibc 2.27 according to https://launchpad.net/ubuntu/bionic/+source/glibc.

### DIFF
--- a/src/get-dart/index.md
+++ b/src/get-dart/index.md
@@ -67,11 +67,6 @@ The Dart SDK is supported on Windows, Linux, and macOS.
   Support for RISC-V is experimental, 
   and is available only in the beta and dev channels.
 
-{{site.alert.note}}
-  The arm support requires glibc 2.23 or newer due to a
-  [dynamic linker bug](https://sourceware.org/bugzilla/show_bug.cgi?id=14341).
-{{site.alert.end}}
-
 ### macOS
 
 * **Supported versions:** Latest three major versions.


atsansone (approved):
I reviewed the supporting docs, but leave final to @mit-mit .

mit-mit:
@mraleph any concerns with removing this note given Ubuntu LTS/18 now has a recent enough version?

mit-mit:
Ah, yes that was the origin bug! I suspect that there is still a non-trivial chance that folks might be on older versions, so I think we should keep the note for now. But we could bring it into the text right after the RISC-V note so that it doesn't stand out so much?

parlough:
If it's still potentially relevant, I think it's fine to keep the note as is for now. The text may get lost.

timsneath:
I was wondering about this today. I did a little research, and it looks like glibc 2.23 (the bugfixed version) was released in 2016 (https://sourceware.org/glibc/wiki/Release/2.23). All downloadable versions of RPi are fixed, and have been since 2020. Even the legacy archive no longer contains broken versions: https://downloads.raspberrypi.org/raspios_oldstable_armhf/images/

I think it might now be safe to remove this, since the affected audience is primarily Raspberry Pi users who haven't upgraded for 2+ years (so they're unsupported by RPi), and there's a trivial upgrade for anyone affected.

PR #4305, opened 2022-10-20
Adjust pub credential file locations

### LINKED ISSUE #4281: [PAGE ISSUE]: 'Troubleshooting pub'
### Page URL

https://dart.dev/tools/pub/troubleshoot

### Page source

https://github.com/dart-lang/site-www/tree/main/src/tools/pub/troubleshoot.md

### Describe the problem

This page needs to be updated to reference the new location of the pub `credentials.json` file. I believe the location is now OS dependent

- `~/Library/Application Support/dart/pub-credentials.json` on the mac
- ~/.config/dart/pub-credentials.json` in linux
- I'm not sure where on windows

### Expected fix

_No response_

### Additional context

_No response_

parlough:
I wasn't 100% sure the best format for this, particularly the Linux entry, so I'm open to suggestions!

Fixes #4281

**Staged:** https://dart-dev--pr4305-fix-4281-46ynnntg.web.app/tools/pub/troubleshoot#pub-publish-403

### DIFF
--- a/src/tools/pub/troubleshoot.md
+++ b/src/tools/pub/troubleshoot.md
@@ -18,19 +18,42 @@ You aren't an uploader for package '<foo>'
 This problem can occur if one of your accounts was granted permission to
 publish a package, but the pub client registers you with another account.
 
-You can reset pub's authentication process by deleting the pub credentials file,
-which can be found in the following locations:
+You can reset pub's authentication process
+by deleting the pub credentials file:
 
-Linux
-: If `$XDG_CONFIG_HOME` is defined
-  then `$XDG_CONFIG_HOME/dart/pub-credentials.json`,
-  otherwise `$HOME/.config/dart/pub-credentials.json`.
+### Linux
 
-macOS
-: `$HOME/Library/Application Support/dart/pub-credentials.json`
+If `$XDG_CONFIG_HOME` is defined:
 
-Windows
-: `%APPDATA%/dart/pub-credentials.json`
+```terminal
+$ rm $XDG_CONFIG_HOME/dart/pub-credentials.json
+```
+
+Otherwise:
+
+```terminal
+$ rm $HOME/.config/dart/pub-credentials.json
+```
+
+### macOS
+
+```terminal
+$ rm $HOME/Library/Application Support/dart/pub-credentials.json
+```
+
+### Windows
+
+If you're using Command Prompt:
+
+```terminal
+$ del "%APPDATA%\dart\pub-credentials.json"
+```
+
+If you're using PowerShell:
+
+```terminal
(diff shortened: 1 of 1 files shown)


src/tools/pub/troubleshoot.md
```diff
@@ -18,11 +18,27 @@ You aren't an uploader for package '<foo>'
...
+Linux
+: If `$XDG_CONFIG_HOME` is defined
+  then `$XDG_CONFIG_HOME/dart/pub-credentials.json`,
+  otherwise `$HOME/.config/dart/pub-credentials.json`.
+
+macOS
+: `$HOME/Library/Application Support/dart/pub-credentials.json`
+
+Windows
+: `%APPDATA%/dart/pub-credentials.json`
```
MaryaBelanger, line 33: I think it's important to keep the actual `rm` command still, since the instruction is "deleting the credentials file" like the original example (quick copy-paste fix, no cognitive overhead, albeit with filling in your own HOME I guess?). If `HOME` is variable, could we also put it in carrot brackets `<HOME>`

Idk what the correct syntax for comments in the terminal is though, I'm guessing it's not `//` haha

```
### Linux
```terminal
    // If `$XDG_CONFIG_HOME` is defined:
    $ rm $XDG_CONFIG_HOME/dart/pub-credentials.json
    // Otherwise:
    $ rm $HOME/.config/dart/pub-credentials.json
```

```
### macOS
```terminal
    $ rm $HOME/Library/Application Support/dart/pub-credentials.json
```

```
### Windows
```terminal
    $ rm %APPDATA%/dart/pub-credentials.json
```

All the backticks are not rendering right, and I'm sure I butchered the syntax with `$` here but does the idea make sense?

parlough: I thought about keeping the commands, but it got a little messy with the different commands, including Windows which has different possible commands as well.

But you're right that the ease of copy-pasting is nice here. I'm going to go ahead and try it and see what it ends up looking like. Thanks!

As for `$HOME`, it is an environment variable, so they shouldn't need to fill it in themselves I believe.

parlough: Okay, I made some changes. We can see how that looks when staged.

I'm not 100% on the Windows commands as I don't have a Windows machine to test them on, but from reading some docs they look right.

MaryaBelanger (approved):
LGTM

It is clunky though, like you said. I think this is a good example of where **on-page tabs** could be a big improvement.  I created issue #4308 to investigate

parlough:
I agree, thanks for opening the issue and its discussion. I'll merge this for now, but once I can design a more generic tab solution, I can switch to that. Will just track that work alongside https://github.com/dart-lang/site-www/issues/4308.

PR #4315, opened 2022-10-24
Add pub.dev to pub troubleshooting solution

parlough:
As of Dart 2.19, pub will use pub.dev instead of pub.dartlang.org. Mentioning both URLs here covers users using future or old versions.


domesticmouse (approved):

atsansone (approved):
LGTM % one update and one nit. Make the update and I don't need another look.

jonasfj (commented):
Nice catch,

PR #4332, opened 2022-10-31
Provide better clarity around operator precedence and associativity

### LINKED ISSUE #4287: Operator precedence table

I would expect that when I search for the phrase `dart operator precedence` on Google to receive a first page result for the dart operator precedence table pointing to a `dart.dev` page.

Currently the latest operator precedence table is only accessible by tapping on the link in the following warning in the Language Tour:
Screen Shot 2022-10-15 at 2 44 57 PM
or going to the Language / Specification and opening the formal specification PDF and searching in there for the table (located on page 240 in the [latest published version](https://dart.dev/guides/language/specifications/DartLangSpec-v2.10.pdf)).

It is fairly easy to miss and can be quite helpful to have.
I noticed that the operator table has the same order in the Language Tour as in the precedence table, but has the warning instead of displaying the order explicitly, not sure why.

I would love to see the operator precedence table displayed on the `dart.dev` website directly so it can be indexed.
One can always find the information quickly by doing the search, it is available here: https://www.woolha.com/articles/dart-operator-precedence-and-associativity ; but I think the community would have much more confidence in the validity if it would come from the official source.

parlough:
While the precedence outlined by the table should only be used as a guideline, users are instead finding third-party sources when googling which do not highlight potential issues and are not able to identify/search that this table still approximates the behavior even if they navigate to the language tour.

This PR makes the table more searchable and identifiable as outlining the approximate precedence, while also clarifying the warning which was a bit confusing, focusing on that it can just be used as a "helpful guide". Beyond that, this adds the associativity labels present on the language spec's "helpful guide".

Fixes #4287

**Staged:** https://dart-dev--pr4332-feature-document-ass-ys8cfbyn.web.app/guides/language/language-tour#operators


parlough:
@eernstg What do you think about these clarifications? I know we want to avoid highlighting this table as outlining operator precedence as the concept has some flaws, but users seem to be finding alternate sites which do not highlight this concern. I tried to make the table a bit more searchable in terms of this, while also making the warning more clear.

src/_guides/language/language-tour.md
```diff
@@ -1723,32 +1723,37 @@ assert(foo() == null);
...
+| bitwise XOR              | `^`                                            | Left |
+| bitwise OR               | `|`                                            | Left |
+| relational&nbsp;and&nbsp;type&nbsp;test | `>=`    `>`    `<=`    `<`    `as`    `is`    `is!` | None |
+| equality                 | `==`    `!=`                                   | None|
+| logical AND              | `&&`                                           | Left |
+| logical OR               | `||`                                           | Left |
+| if null                  | `??`                                           | Left |
+| conditional              | <code><em>expr1</em> ? <em>expr2</em> : <em>expr3</em></code> | Right |
+| cascade                  | `..` &nbsp;&nbsp; `?..`                        | Right |
+| assignment               | `=`    `*=`    `/=`    `+=`    `-=`    `&=`    `^=`    <em>etc.</em> | Right
```
domesticmouse, line 1749: Missing trailing `|`?

src/_guides/language/language-tour.md
```diff
@@ -1723,32 +1723,37 @@ assert(foo() == null);
...
+|--------------------------+------------------------------------------------+----------------------|
+| unary postfix            | <code><em>expr</em>++</code>    <code><em>expr</em>--</code>    `()`    `[]`    `?[]`    `.`    `?.`    `!` | None |
+| unary prefix             | <code>-<em>expr</em></code>    <code>!<em>expr</em></code>    <code>~<em>expr</em></code>    <code>++<em>expr</em></code>    <code>--<em>expr</em></code>      <code>await <em>expr</em></code>    | None |
+| multiplicative           | `*`    `/`    `%`    `~/`                      | Left |
+| additive                 | `+`    `-`                                     | Left |
+| shift                    | `<<`    `>>`    `>>>`                          | Left |
+| bitwise AND              | `&`                                            | Left |
+| bitwise XOR              | `^`                                            | Left |
+| bitwise OR               | `|`                                            | Left |
+| relational&nbsp;and&nbsp;type&nbsp;test | `>=`    `>`    `<=`    `<`    `as`    `is`    `is!` | None |
```
domesticmouse, line 1742: nit: should all the columns be lined up?

parlough: The non-breaking spaces make it a bit awkward in some editors, but I'll try my best.

domesticmouse: Feel free to say no here. There is a reason I asked it as a question, and marked it as a nit. +)

domesticmouse (approved):

src/_guides/language/language-tour.md
```diff
@@ -1723,32 +1723,38 @@ assert(foo() == null);
...
 {:.table .table-striped}
 
 {{site.alert.warning}}
-  Operator precedence is an approximation of the behavior of a Dart parser.
+  The operator precedence outlined here
+  is only an approximation of the precedence and associativity
+  given implicitly by Dart's grammar
+  and should only be used as a helpful guide.
   For definitive answers, consult the grammar in the
   [Dart language specification][].
```
atsansone, line 1759: **question:** Why does this only approximate precedence? Is there a difficulty in providing the specific precedence? What is that difficulty?

**suggestion:** Tighten the language and use active voice.
```suggestion
  The operator precedence that the previous table outline provides 
  a rough guide. You can find the precedence and associativity for Dart
  in the [Dart language specification][].
```

src/_guides/language/language-tour.md
```diff
@@ -1723,32 +1723,38 @@ assert(foo() == null);
 
 ## Operators
 
-Dart supports the operators shown in the following table.
+Dart supports the operators shown in the following table,
+which **approximates** Dart's operator associativity 
+as well as operator precedence from highest to lowest.
```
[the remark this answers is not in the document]
atsansone: **comment:** I like where @eernstg is going with his comment. I would recommend more shorter sentences than longer sentences.

atsansone (changes requested):
@parlough : I have a few comments for your review.

src/_guides/language/language-tour.md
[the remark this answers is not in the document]
parlough: Thanks for the feedback! I had some trouble incorporating the full "source of truth" portion in to the primary text as many readers may not be familiar with what a "language grammar" is and I'm worried it may scare them away from this section. I added a little elaboration to the primary text and added the "source of truth" text to the warning which is rather prominent. How does that seem?

As for the example and explanation of precedence, we have one right after the table. I added a link to there instead. We don't mention the underlying grammar, but I think that is probably good as many readers of this may not know how to read it, and it could just confuse them. We also don't want too much text/description in the language tour.

src/_guides/language/language-tour.md
parlough: It's not that you can find any authoritative precedence and associativity in the language specification as they aren't really defined (or able to be exactly), but rather you need to look at the whole grammar to understand the relationships between operators.

I've took some of your changes and Erik's, while also being clear about it only be used as a guide (which was the source of confusion that caused the introduction of the note a long time ago).

eernstg: I could comment a bit more on why operator precedence is not a precise concept.

If you settle on a specific class of parsers (say, table driven LALR(1) parsers) then you _can_ have a technically precise notion of precedence. In case of an ambiguity, the parser will then use the assigned precedence values to make the decision about whether to shift or reduce, and which rule to reduce. This makes sense if we have a grammar where a lot of different expressions use the same nonterminal:

```
...
expr ::= expr ('+' | '-') expr @precedence(5)
expr ::= expr ('*' | '/') expr @precedence(6)
...
```

Precedence has a precise meaning in this context, and it is a tool outside the underlying context free grammar formalism (it is defined in terms of the chosen parser algorithm, it's not a property of the grammar &mdash; the grammar is simply ambiguous).

If we are using a different approach to parsing (for instance, Dart has always used an enhanced recursive descent parser, which is more like ANTLR) then there is no reason to assume that there exists any assignment of precedence to any set of operators that will describe the behavior of the parser correctly. In other words, the whole concept of precedence doesn't necessarily fit that approach to parsing. When using that approach, the ambiguities simply do not arise, because we use different nonterminals to establish a similar property:

```
...
additiveExpression ::= multiplicativeExpression (('+' | '-') multiplicativeExpression)*
multiplicativeExpression ::= someOtherExpression (('*' | '/') someOtherExpression)*
...
```

Looking at the grammar you can see that an additive expression can consist of a sequence of multiplicative expressions, separated by '+' or '-' (and not the other way around), and that makes it tempting to say that '*' has a higher precedence than '+'. (Of course, there would be more rules about `additiveExpression`, because it can also derive other things than a sum of products, but this is the core part with respect to precedence.)

However, the source of truth is the grammar, and if we settle on a particular set of precedence values, it takes a non-trivial proof to verify that this precedence table describes the derivable programs with that grammar precisely (and maybe that proof doesn't exist, because it just isn't true).

src/_guides/language/language-tour.md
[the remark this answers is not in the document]
eernstg: Sounds good. You are probably right that it is unreasonable to try to describe any subtle details here.

src/_guides/language/language-tour.md
atsansone: @parlough : I'd still recommend the tighter language.

@eernstg : Thanks for the explanation!

atsansone (approved):
LGTM % 1 suggestion.

src/_guides/language/language-tour.md
```diff
@@ -1723,33 +1723,38 @@ assert(foo() == null);
 ## Operators
 
 Dart supports the operators shown in the following table.
+The table shows Dart's operator associativity 
+and [operator precedence](#operator-precedence-example) from highest to lowest,
+which are an **approximation** of Dart's operator relationships.
 You can implement many of these [operators as class members](#_operators).
```
atsansone, line 1729: **suggestion:** To tighten language, sometimes you need to write more, shorter sentences.
```suggestion
The following table lists the operators that Dart supports.
The table sorts the operators from highest to lowest precedence.
When determining which operator to apply first, Dart compares operators.
When comparing two different operators, 
Dart applies the operator with greater precedence first. 
When comparing two of the same operator, 
Dart applies the operator from the associative direction first. 
Complex expressions may apply operators in a recursive fashion. 
Due to the complexity of [operator-precedence grammar](https://en.wikipedia.org/wiki/Operator-precedence_grammar), use this table as a guideline only.  

You can use many of these operators as class members.
```

- [x] How it should only be used as a "helpful guide" not definitive
- [x] Briefly explaining why it can only be used as a helpful guide
- [x] Where/how to find and understand the authoritative behavior

eernstg: I can't help commenting here! ;-)

This doesn't tell the reader that the whole concept of operator precedence is imprecise. It's not a matter of complexity, there's just no reason to assume that there exists an operator precedence table which is a correct description of the behavior of a Dart parser. It would certainly be a non-trivial effort to establish any firm results in this area.

parlough: I'm also a bit worried about providing so much information in the opening here, when most readers are just looking for the table. We expand on the the precedence right after the table already, and I like your comments around that, so I'll try to incorporate some of that there. I'm going to go ahead and open a new PR which does a larger overhaul of this operator section, incorporating some of everyone's thoughts.

eernstg (approved):
We talked about it, this makes sense, thanks!

PR #4336, opened 2022-11-02
Use code block tags instead of plain text to identify language in Coming from article

parlough:
- Adds a class which allows the Dart/Js language to show up in the codeblock
- Replaces manual Dart/JS tags with the class
- Uses the primary brand color for Dart and a darker yellow than GitHub uses (to preserve contrast) for JS
- Adds font-weight 500 for Google Sans Mono as it is what this codeblock extra text is intended to use

**Before:**
image

**After:**
image

**Staged:** https://dart-dev--pr4336-feature-language-tag-418hoc9h.web.app/guides/language/coming-from/js-to-dart

### DIFF
--- a/src/_guides/language/coming-from/js-to-dart.md
+++ b/src/_guides/language/coming-from/js-to-dart.md
@@ -139,8 +139,7 @@ For example, the equals operator `==` and the `identical()`
 method are guaranteed to return true for the
 same values of these types, as shown in the following code:
 
-<!--skip-->
-_Dart_
+{:.include-lang}
 ```dart
 var a = 1;
 var b = 1;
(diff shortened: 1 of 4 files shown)


domesticmouse (approved):

atsansone (approved):
@parlough : Love this. The JavaScript text is a bit light; if that can be fixed, please do that, otherwise LGTM. I have a change that follows this, so I'm happy to see this merge ASAP.

parlough:
I made it a little darker while still trying to be relatively close to JavaScript's common colors.

The build is breaking due to new security warnings by CodeQL. I created a fix in https://github.com/dart-lang/site-www/pull/4339

johnpryan (approved):

PR #4341, opened 2022-11-02
Be more specific about dart compile JS environment declaration option

parlough:
Moves to the term "environment declaration" as these are different from environment variables.

The environment declaration isn't necessarily just an `int` too, so if we want to link to API docs here, we should likely link to all of them. 

Also use site variables for the API docs links, so we don't point to outdated API links.


domesticmouse (approved):

src/_includes/tools/dart-compile-js-options.md
```diff
@@ -64,11 +64,18 @@ Some other handy options include:
   [Dart Package Configuration File.][]
 
 `-D<flag>=<value>`
-: Defines an [environment variable](https://api.dart.dev/stable/2.18.0/dart-core/int/int.fromEnvironment.html).
+: Defines an environment declaration and value pair
+  which can be accessed with 
+  [`String.fromEnvironment`][], [`int.fromEnvironment`][], 
+  [`bool.fromEnvironment`][], or [`bool.hasEnvironment`][].
```
atsansone, line 70: **nit:** Consider breaking into two sentences.
```suggestion
: Defines an environment declaration and value pair.
  Access this pair with 
  [`String.fromEnvironment`][], [`int.fromEnvironment`][], 
  [`bool.fromEnvironment`][], or [`bool.hasEnvironment`][].
```

atsansone (approved):
@parlough : LGTM % one nit. You can ignore the nit if you want.

PR #4346, opened 2022-11-06
Only add a tag to JS code where include-lang is specified

parlough:


atsansone (approved):
LGTM. Would like @khanhnwin view, too.

PR #4354, opened 2022-11-08
Update JS to Dart

atsansone:
[Staged page](https://dart--staging.web.app/guides/language/coming-from/js-to-dart).

Original comments are in a [Google Doc](https://docs.google.com/document/d/1h7okBLXoKX2UuG7bXJSVcAQZiuV9AcRfoDKsxG5gMdg/edit#).

### DIFF
--- a/src/_guides/language/coming-from/js-to-dart.md
+++ b/src/_guides/language/coming-from/js-to-dart.md
@@ -3,25 +3,25 @@ title: Learning Dart as a JavaScript developer
 description: Leverage your JavaScript knowledge when learning Dart.
 ---
 
-This guide aims to leverage your JavaScript (JS)
+This guide aims to leverage your JavaScript
 programming knowledge when learning Dart.
 It showcases key similarities and differences
 in both languages, and introduces Dart concepts
-that are not present in vanilla JavaScript.
+that are not present in JavaScript.
 As a JavaScript developer,
 Dart should feel quite familiar,
-as both languages share many concepts. 
+as both languages share many concepts.
 
 Like JavaScript, Dart runs on an event loop,
 so both languages execute code in a similar way.
 For example, asynchronous concepts like futures
-(promises in JS) and the `async/await` syntax are very similar.
+(promises in JavaScript) and the `async/await` syntax are very similar.
 
 Dart is strongly typed, unlike JavaScript.
 If you are also familiar with TypeScript or tooling
 that adds static typing (like Flow), then learning Dart
 probably won't be much of an extra step for you.
-However, if you've mostly worked with vanilla JavaScript,
+However, if you've mostly worked with JavaScript,
 it might take a bit more effort to adjust.
 The good news is that since Dart is strongly typed,
 many errors that might exist in JavaScript code are
@@ -34,7 +34,7 @@ but the trade-off is better protection against
 null reference exceptions that are detected even
 before compiling Dart code. (Thereby avoiding those
 dreaded `TypeError`s that occur when doing operations
-on a JS variable that turns out to be null.)
+on a JavaScript variable that turns out to be null.)
 
 ## Conventions and linting
 
@@ -48,56 +48,57 @@ a linter to make compliance effortless.
 To customize the lint rules for your project,
 follow the [Customizing static analysis][] instructions.
 
+Dart provides [`dart fix`][] to find and fix errors that the analyzer finds.
+
+Dart also provides a code formatter similar to JavaScript tools like [Prettier][].
+To format code in any Dart project, run [`dart format`](/tools/dart-format) on
+your command line. In Flutter, use `flutter format`.
+The IDE plugins for Dart and Flutter also provide this ability.
+
+Dart supports trailing commas for comma-separated lists of collections,
+parameters, or arguments. When you add the trailing comma,
+the formatter places each list item on its own line.
+When you believe your list may have more items at a later date,
+add the trailing comma. Avoid adding the trailing comma for the formatting
+benefit alone.
+
+JavaScript supports trailing commas in list and map literals only.
+
 {{site.alert.secondary}}
-  **Pro tip:** Dart provides [`dart fix`][],
-  which finds and fixes errors found by the analyzer.
+  To learn more about:
+
+  * Using commas to make your code read more like HTML, read
+    [Using trailing commas][] on flutter.dev.
+  * Linting Dart, read [Linter rules][].
+  * Writing good Dart code, read [Effective Dart][].
 {{site.alert.end}}
-
-Dart also provides a code formatter,
-which is similar to JS tools like [Prettier][].
-Automatically format any Dart project by running
-[`dart format`](/tools/dart-format) on the command line. 
-For Flutter, `flutter format` acts as an alias for this command.
-(Note that the IDE plugins for Dart and Flutter
-also provide this functionality.)
-
-{{site.alert.secondary}}
-  **Pro tip:** Dart supports optional trailing
-  commas for any comma-separated values,
-  such as function parameters or list items.
-  This causes the formatter to place each item
-  onto its own line, which helps with readability,
-  especially when you have a lot of nested code
-  (as can happen in Flutter layout code).
-{{site.alert.end}}
-
-For more information on using commas to make your code
-read more like HTML, check out
-[Using trailing commas][] on flutter.dev.
-
-For more information about Dart conventions and linting,
-check out [Effective Dart][] and [Linter rules][].
 
 
 [Customizing static analysis]: /guides/language/analysis-options
 [`dart fix`]: /tools/dart-fix
-[Effective Dart]: /guides/language/effective-dart 
+[Effective Dart]: /guides/language/effective-dart
 [Linter rules]: /tools/linter-rules
 [Prettier]: https://prettier.io/
 [Using trailing commas]: {{site.flutter-docs}}/development/tools/formatting#using-trailing-commas
 
-
 ## Built-in types
 
-While both JavaScript and Dart have types,
-only Dart is a strongly typed language.
-This means that all Dart types have to be
-inferrable by the analyzer, explicitly defined,
-or assigned as `dynamic`, which disables static type
-checking for that identifier.
+Both JavaScript and Dart categorize their data into _types_.
+Every variable has an associated type.
+The type determines the kind of value the variable can store and
+what operations can be performed on these values.
+Dart differs from JavaScript in that it assigns a static type to every
+variable and to every expression.
+In Dart, the static type predicts the runtime type of the values of a
+variable, or of the value of an expression.
+Dart assigns a static type to every expression and variable.
+In Dart, the static type predicts the runtime type of the value of an
+expression. This means that Dart apps have sound static typing.
 
-Dart supports nullable and non-nullable versions
-of the the following built-in types:
+JavaScript provides primitive types `num`, `string`, and `boolean`
+and the `null` value as well as *arrays* and a `Map` type.
+
+Dart supports built-in types include:
 
 * Numbers (`num`, `int`, `double`)
 * Strings (`String`)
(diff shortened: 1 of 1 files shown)


src/_guides/language/coming-from/js-to-dart.md
```diff
@@ -2356,27 +2519,68 @@ consumer = Cat();
...
+class Bat extends Animal {}
+class Goose extends Animal {}
+class Dog extends Animal {}
+
+// Composing the class instances with
+// their correct functionality.
+const bat =
+  Object.assign(
+    new Bat(),
+    new Flyer()
```
lrhn, line 2560: This works, but it relies on the `fly`/`walk` "methods" being declared as fields instead.
If you wrote it as 
```js
class Flyer {
  fly() => console.log('Flaps wings');
}
```
instead, which would be the canonical way to write a JS class, the `assign` would not copy the method (because it's on the prototype and is not enumerable).
Might be a caveat worth mentioning.

src/_guides/language/coming-from/js-to-dart.md
```diff
@@ -2607,21 +2793,19 @@ class Animal {
...
-beyond a single file using the `part` directive,
-it's generally advised to [avoid doing so][] and is usually
-reserved for code generators. 
+enforced by the compiler.
+
+Dart makes private members private to the library, not the class.
+This means that you can access private members from code in the same library.
+By default, this works anywhere in the same file.
+Private class members can be accessed from code in the same file.
+Using the `part` directive to expand the scope of a library scope a single file.
```
lrhn, line 2806: Word missing? Maybe "beyond a single files"?

src/_guides/language/coming-from/js-to-dart.md
```diff
@@ -2607,21 +2793,19 @@ class Animal {
...
+When possible, [avoid doing so using `part`][]. The common practice would be
```
lrhn, line 2807: Just ``[avoid using `part`]``. There are other reasons to use `part` files and you should avoid those too if you can.

src/_guides/language/coming-from/js-to-dart.md
```diff
@@ -2632,88 +2816,69 @@ indicate they are initialized at a later point,
...
+    if (capture) captures.add(foo);
+  }
 }
-```
 
-In this example, `_readThermometer()` only runs
-when the field is first accessed,
-rather than on initialization. 
+In the preceding example, the compiler does not know to assign
+`captures` if `capture` is true. In this case `capture` is true,
```
lrhn, line 2858: Drop "In this case `capture` is true,", we don't know that. The logic is sound whether it's true or not.

lrhn (commented):
Some comments. Most stylistic.
The initial section on "primitive types" really doesn't work for me. The original wasn't better.
I'll try doing a rewrite myself.

src/_guides/language/coming-from/js-to-dart.md
atsansone: As this describes a specific caveat for JS, I'll leave it as is.

atsansone (commented):
@lrhn : Made the requested changes.

src/_guides/language/coming-from/js-to-dart.md
```diff
@@ -48,56 +48,55 @@ a linter to make compliance effortless.
...
-  (as can happen in Flutter layout code).
-{{site.alert.end}}
+Dart supports trailing commas for comma-separated lists of collections,
+parameters, or arguments. When you add the trailing comma,
+the formatter places each list item on its own line.
+When you believe your list may have more items at a later date,
+add the trailing comma. Avoid adding the trailing comma for the formatting
+benefit alone.
+
+JavaScript supports trailing commas in list and map literals only.ß
```
lrhn, line 65: Spurious "ß"

src/_guides/language/coming-from/js-to-dart.md
```diff
@@ -1030,21 +1135,20 @@ assert(25 == 51.6 ~/ 2);
...
-| Greater than                  | `>`           | `>`           |
-| Less than                     | `<`           | `<`           |
-| Greater than or equal to      | `>=`          | `>=`          |
-| Less than or equal to         | `<=`          | `<=`          |
+| Meaning                   | JavaScript operator | Dart operator    |
+|---------------------------|---------------------|------------------|
+| Strict equal              | `===`               | `indentical(,)`  |
+| Abstract equal            | `==`                | `==`             |
+| Strict not equal          | `!==`               | `!indentical(,)` |
+| Abstract not equal        | `!=`                | `!=`             |
```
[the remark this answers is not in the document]
parlough: It seems like Dart and JavaScript's `==` operators shouldn't be equated, as mentioned earlier Dart's `==` doesn't do automatic type conversions.

Also it doesn't seem it's a good idea to equate `===` to Dart's `identical`] as it seems most modern JavaScript code defaults to `===` but `identical` generally shouldn't be the default choice for determining equality.

lrhn (approved):
Ah, had this late yesterday, but forgot to submit.

src/_guides/language/coming-from/js-to-dart.md
```diff
@@ -108,135 +107,140 @@ of the the following built-in types:
...
+
+#### Strings
+
+Strings in Dart work like strings in JavaScript.
+To write a string literal, enclose it in single (`'`) or double (`"`)
+quotation marks.
+The majority of Dart developers use single quotes,
+but the language enforces no standard.
+Use double quotation marks if you don't want to escape apostrophes or
+single quotes within the string.
```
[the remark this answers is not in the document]
atsansone: I just removed apostrophes. That's a confusing addition.

parlough:
I'll take a final look at this tomorrow morning, sorry about the delay.

src/_guides/language/coming-from/js-to-dart.md
```diff
@@ -1239,17 +1340,26 @@ print(animal.name); // "Bob"
...
-in JavaScript.
+This section covers some collection types in Dart and compare them
+to similar types in JavaScript.
 
 ### Lists
 
-List literals are defined the same way in Dart as
-arrays are defined in JavaScript,
-using square brackets and separated by commas: 
+Dart writes list literals as JavaScript arrays.
```
parlough, line 1361: This text is a bit confusing to me. It seems to imply Dart uses JavaScript arrays somehow.

Perhaps:

```suggestion
Dart writes list literals similarly to JavaScript arrays.
```

src/_guides/language/coming-from/js-to-dart.md
```diff
@@ -1515,6 +1629,8 @@ final gift = gifts['first'];
 
 {{site.alert.note}}
   The index operator (`[]`) returns a **nullable** value.
+  The operator returns `null` if the map does not include
+  the lookup key.
```
parlough, line 1633: Saying it returns a nullable value and it returns `null` seems confusing and a bit unnecessary. Can we just say one or the other, or connect them better somehow?

src/_guides/language/coming-from/js-to-dart.md
```diff
@@ -1695,15 +1840,28 @@ to their JavaScript counterparts.
 ### Futures
 
 `Future` is Dart's version of a `Promise`:
-an asynchronous operation that resolves at a later point. 
+an asynchronous operation that resolves at a later point.
```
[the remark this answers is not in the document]
parlough: This clarification makes sense and I think should still be addressed.

src/_guides/language/coming-from/js-to-dart.md
```diff
@@ -1713,42 +1871,45 @@ httpResponseBody.then((String value) {
...
-```js
+```dart
 httpResponseBody
   .then(...)
-  .catch(err => {
-    console.log("Promise encountered an error before resolving.");
+  .catchError((err) {
+    print(
+      'Future encountered an error
+       before resolving.'
```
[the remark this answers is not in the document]
parlough: @atsansone These should still be fixed.

src/_guides/language/coming-from/js-to-dart.md
```diff
@@ -1916,20 +2056,19 @@ you have several different ways to create a stream.
...
 
-Another common way to create streams is
-by using a [`StreamController`][],
-a utility class that builds streams.
-A `StreamController` contains a `stream` property
-that exposes the stream it controls,
-and multiple methods for controlling the stream,
-such as emitting new items using the `add` method,
-or completing the stream using the `close` method.
+The utility class [StreamController][] can create and control streams.
```
parlough, line 2065: Keep this in code syntax to be consistent with elsewhere:

```suggestion
The utility class [`StreamController`][] can create and control streams.
```

src/_guides/language/coming-from/js-to-dart.md
```diff
@@ -2555,26 +2706,25 @@ void main() {
...
-all class members are public by default. 
+all class members are public by default.
 
-While private class members are not yet officially
-part of JavaScript because they are not part of any
-published EcmaScript standard,
-a proposal for this has been completed and is ready
-to be included in the next publication of the standard.
+JavaScript will include private class members in the next
+practical revision of the EcmaScript standard.
```
parlough, line 2717: This whole talk about the EcmaScript spec for private members not being complete is confusing and I'm not sure it should be included. Either way, it looks like PrivateIdentifiers were part of EcmaScript 2022 which has been released.

Reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields

src/_guides/language/coming-from/js-to-dart.md
```diff
@@ -2607,113 +2756,87 @@ class Animal {
...
-Assign the `late` keyword to class fields to
-indicate they are initialized at a later point,
-while remaining non-nullable. This is useful
-for cases where a variable is never observed
-before being initialized, allowing it to be initialized later.
-This has several advantages over just labeling the field as nullable:
+To indicate that Dart initializes class fields at a later point,
+assign the `late` keyword to those class fields.
+Those class fields remain non-nullable.
+Do this when a variable cannot be observed before being initialized.
```
parlough, line 2775: This new sentence is confusing as every variable technically "cannot be observed before being initialized". I think the original text was trying to say something like "Do this when a variable does not need to be(or is not) observed/accessed right away and can be initialized later".

parlough (changes requested):
Some comments, suggestions, and issues. Feel free to push back on any. I noticed @lrhn has a few open comments that still seem relevant as well. I commented or reacted on them.

If possible, avoid squashing commits and rebasing as it made it hard to see what changed between/after reviews.

src/_guides/language/coming-from/js-to-dart.md
```diff
@@ -1167,69 +1243,94 @@ assert((value >>> 4) == 0x02); // Unsigned shift right
...
 ```
 
-#### Ternary operator
+#### Conditional operator
+
+Both Dart and JavaScript contain a conditional operator (`?:`)
+for evaluating expressions.
+Some developers refer to this operator as a ternary operator
+as it takes three operands.
+Dart has another operator (`[]=`) that takes three operands.
```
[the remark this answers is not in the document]
atsansone: I think I have this updated correctly.

src/_guides/language/coming-from/js-to-dart.md
atsansone: I checked the reference specifically; privates are in stage 4, which means they will be in the next practical revision. That's why I phrased it as I did.

src/_guides/language/coming-from/js-to-dart.md
atsansone: Fair. Fixed.

atsansone (commented):
@parlough : Back to you!

src/_guides/language/coming-from/js-to-dart.md
```diff
@@ -3,101 +3,95 @@ title: Learning Dart as a JavaScript developer
...
-to enforce standard conventions. However,
-while JavaScript has `ESLint` as a standalone tool
-(among other available tooling),
-and many different standards and configurations,
-Dart has official layout conventions and includes
-a linter to make compliance effortless.
+JavaScript and Dart both have linting tools to enforce standard conventions.
+While JavaScript offers many tools, standards, and configurations,
+Dart has one official layout conventions and includes a linter to make
+compliance effortless. Dart calls this linter the analyzer.
```
parlough, line 41: The analyzer is a lot more than just the linter, providing warnings and errors for misuse of language features,
code completion, actions like "Find usages", "Go to definition", "Fix X", etc.

```suggestion
Dart has official layout and styles conventions and includes
a linter to make compliance effortless. 
Support for this and other analysis functionality is provided by the analyzer.
```

parlough (approved):
Thanks for cleaning up this article. Looks good to me.

Issue #4166, opened 2022-08-09
[PAGE ISSUE]: 'dart pub global' 


GroovinChip:
### Page URL

https://dart.dev/tools/pub/cmd/pub-global#activating-a-package-with-git

### Page source

https://github.com/dart-lang/site-www/tree/main/src/tools/pub/cmd/pub-global.md

### Describe the problem

There should be instructions for activating a package from git that is contained in a monorepo or other grouped directory structure. 

For example, I want to use my forked version of [flutter_distributor](https://github.com/GroovinChip/flutter_distributor/tree/main/packages/flutter_distributor) since my PR to the origin repo has not yet been reviewed/merged. However, since it is contained in a monorepo I cannot simply run `dart pub global activate -sgit https://github.com/GroovinChip/flutter_distributor.git` because that is not specifically where the package is located.

For such cases, instructions for installing a package from a subdirectory of a git repository should be added to the page.

### Expected fix

Add instructions for installing a package from a subdirectory of a git repository to the page. It appears that there is a `--git-path` subcommand for `activate` - if this is the correct way to accomplish this, instructions for using it should be added.

PR #4356, opened 2022-11-08
Document specifying a file path when activating a package

parlough:
Fixes https://github.com/dart-lang/site-www/issues/4166

### DIFF
--- a/src/tools/pub/cmd/pub-global.md
+++ b/src/tools/pub/cmd/pub-global.md
@@ -73,10 +73,10 @@ $ dart pub global activate --source git https://github.com/dart-lang/async_await
 $ dart pub global activate -sgit https://github.com/dart-lang/async_await.git
 ```
 
-Pub assumes that the package is in the root of the Git repository. 
-To specify a different location in the repo,
-use the `--git-path` flag to
-specify a path relative to the repository root:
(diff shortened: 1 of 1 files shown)


src/tools/pub/cmd/pub-global.md
```diff
@@ -73,6 +73,15 @@ $ dart pub global activate --source git https://github.com/dart-lang/async_await
 $ dart pub global activate -sgit https://github.com/dart-lang/async_await.git
 ```
 
+Pub assumes that the package is in the root of the Git repository. 
+To specify a different location in the repo,
+use the `--git-path` flag to
+specify a path relative to the repository root:
```
atsansone, line 79: **question:** Do we use _flag_ or _option_? I would recommend option. (See [docopt standards](http://docopt.org/).

**suggestion:** 
```suggestion
Pub expects to find the package in the root of the Git repository.
To specify a different location, use the  `--git-path` option with
a path relative to the repo root.
```

atsansone (approved):
LGTM % one question and one suggestion.

src/tools/pub/cmd/pub-global.md
parlough: Using option makes sense and is consistent with the help menu. Thanks!

PR #4357, opened 2022-11-08
[2.19] Concurrency updates for isolate.run()

### LINKED ISSUE #3904: [2.19] `Isolate.run()` docs
https://dart-review.googlesource.com/c/sdk/+/217008

### Summary

"Runs [computation] in a new isolate and returns the result." - [API docs](https://github.com/dart-lang/sdk/commit/a5f25ee3ba5a7f2e22b7698590727b32c37204f9#diff-c9ae8fb1ff4a1b2edd5b4883ebb21d80f818511d5f90ff7e79993f2cb9f909fcR153)

Currently, you have to use sendport and receiveport to get a return from an isolate. Isolate.run() doesn't eliminate the need for send-and-receive ports (there are still cases where one might want to fine tune the configuration of using send-and-receive ports for some reason, and they still can), it just wraps their usage so users don't have to handroll that implementation themselves. Now you can just call isolate.run() and get a return without having to write anything else

### Evaluation

- Potential updates needed on [Concurrency in Dart](https://dart.dev/guides/language/concurrency)
  - The use of sendport/receiveport shouldn't be mentioned in the default examples. 
  - Use isolate.run by default, mention that you could use send/receive if for some reason you want to configure that more closely
    - This page might not be the place to go into that much detail
- Potential new content needed; needs more info to determine
  - I don't think there is a need for new content, just updating the existing mentions of isolate that go in detail enough to talk about sharing data between isolates.
- Not sure about the significance of the [flutter samples repo](https://github.com/flutter/samples/tree/main/isolate_example), but it might need updating or adding to  
  - Use of send and receive ports is the default
  - My suggestion: any use of send and receive ports needs to be changed to isolate.run
  - Then, maybe, create a new example that shows how to use send and receive if you really want to
  - Samples being updated [here](https://github.com/flutter/samples/issues/1487)

MaryaBelanger:
Figuring out code excerpt stuff before review

Closes #3904 

Staged here: https://dart-dev-marya.web.app/guides/language/concurrency
Specifically the [Code examples](https://dart-dev-marya.web.app/guides/language/concurrency#code-examples) section


examples/concurrency/lib/simple_isolate_closure.dart
```diff
@@ -0,0 +1,20 @@
+import 'dart:convert';
+import 'dart:io';
+import 'dart:isolate';
+
+const String filename = 'with_keys.json';
+
+// #docregion main
+void main() async {
+  // Read some data.
+  final jsonData = await Isolate.run(() {
```
lrhn, line 10: There should be an `async` at `() async {`, to make it possible to use `await` in the function.

examples/concurrency/lib/simple_isolate_closure.dart
```diff
@@ -0,0 +1,20 @@
...
+    final fileData = await File(filename).readAsString();
+    final jsonData = jsonDecode(fileData) as Map<String, dynamic>; 
+    return await jsonData;
```
lrhn, line 13: This `await` is unnecessary and should be removed.

(The code from `()` to `}` should be precisely the same as in `_readAndParseJason` below.)

examples/concurrency/lib/simple_isolate_closure.dart
```diff
@@ -0,0 +1,20 @@
...
+  });
+  
+  // Use that data
```
lrhn, line 16: Extreme nit: Put a `.` at the end (Dart style says format comments as sentences).

src/_guides/language/concurrency/index.md
```diff
@@ -295,6 +295,12 @@ The worker isolate can block without affecting other isolates.
 
 [`send()` method]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-isolate/SendPort/send.html
 
+Message passing normally involves data copying, and thus can be slow.
+However, when you send data using isolates,
+the memory that holds the message in the worker isolate isn’t copied,
+but instead is transferred to the receiving isolate.
```
lrhn, line 301: This is not generally true (and not sure what "send data using isolates" mean). It's only true for the message sent using `Isolate.exit`.
So, should it say: "When you send data using `Isolate.exit`,"?

src/_guides/language/concurrency/index.md
```diff
@@ -295,6 +295,12 @@ The worker isolate can block without affecting other isolates.
...
+The sender will nonetheless perform a verification pass to ensure
+the objects are allowed to be transferred.
```
lrhn, line 303: (That does seem to match `Isolate.exit` behavior.)

src/_guides/language/concurrency/index.md
```diff
@@ -315,104 +321,136 @@ to implement isolates.
...
 Here’s the code for the main isolate:
 
 <?code-excerpt "lib/simple_worker_isolate.dart (main)"?>
 ```dart
 void main() async {
   // Read some data.
-  final jsonData = await _parseInBackground();
+  final jsonData = await Isolate.run(_readAndParseJson);
 
   // Use that data
```
lrhn, line 349: (Final `.`).

src/_guides/language/concurrency/index.md
```diff
@@ -315,104 +321,136 @@ to implement isolates.
...
-   the single message that the worker isolate sends.
+In this example, `_readAndParseJson()` is an existing,
+asynchronous function that could just as easily
+run directly in the main isolate.
+But using `Isolate.run()` enables concurrency,
+so its computations are completely abstracted by the worker isolate
+and can complete without blocking the main isolate.
 
-[`first`]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-async/Stream/first.html
+An added benefit is that the result of
```
lrhn, line 382: Being asynchronous is not what people would normally call a "benefit" :)

I'd say rather just not value-load it, and say something like:

> Because `Isolate.run()` allows other code to run while the computation
> executed in another isolate, its result is always returned as a `Future`.
> To the original isolate, it doesn't matter whether the computation running in the 
> other isolate is synchronous or asynchronous, because it's being run concurrently
> either way.

src/_guides/language/concurrency/index.md
MaryaBelanger: That's correct, this paragraph was previously under the first example and specifically about `SendPort.send` vs. `Isolate.exit()`. I was trying to generalize it and make it a conceptual summary instead of the specific mention of those methods because they're abstracted now. But, I don't want to completely remove the info...

I'll see how I can incorporate the valuable parts of this more accurately and then request your review again, thanks!

MaryaBelanger:
Checks failing because `Isolate.run()` is undefined.
Will stay that way until 2.19 release.

parlough:
Can you also run `dart format` on the files and refresh code excerpts? That's what's causing beta and dev to fail.

MaryaBelanger:
> Can you also run `dart format` on the updated `.dart` files and refresh code excerpts? That's what's causing beta and dev to fail.

Oh, nice! It was `concurrency_test.dart` that I missed, `main()` needed a `Future`

src/_guides/language/concurrency/index.md
```diff
@@ -312,107 +311,144 @@ to implement isolates.
...
-This section shows the implementation for a
-main isolate and the simple worker isolate that it spawns.
-The worker isolate executes a function and then exits,
-sending the main isolate a single message as it exits.
-(The [Flutter `compute()` function][] works in a similar way.)
+These examples show the implementation of a
+main isolate that spawns a simple worker isolate.
+The examples use [`Isolate.run()`][], 
+which efficiently abstracts the complexity
+of setting up and managing worker isolates manually:
```
atsansone, line 320: **suggestion:** Remove the nominalizations and adverbs. Both make the copy wordy and distract some from the understanding.

```suggestion
These examples implement a main isolate that spawns
a simple worker isolate.
Using [`Isolate.run()`][] simplifies setting up and
managing worker isolates.
```

src/_guides/language/concurrency/index.md
```diff
@@ -312,107 +311,144 @@ to implement isolates.
...
 
-This example uses the following isolate-related API:
+* Spawns (starts and creates) an isolate
```
atsansone, line 322: **question:** Are these items in order? If so, should they not be an ordered list?

src/_guides/language/concurrency/index.md
```diff
@@ -312,107 +311,144 @@ to implement isolates.
...
-* [`ReceivePort`][] and [`SendPort`][]
+[`Isolate.run()`]: {{site.dart-api}}/dev/dart-isolate/Isolate/run.html
 
-[`Isolate.exit()`]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-isolate/Isolate/exit.html
-[`Isolate.spawn()`]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-isolate/Isolate/spawn.html
-[`ReceivePort`]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-isolate/ReceivePort-class.html
-[`SendPort`]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-isolate/SendPort-class.html
+#### Running an existing method in a new isolate
 
 Here’s the code for the main isolate:
```
atsansone, line 333: **suggestion:** Don't use "here" to indicate location. "Here" would be that exact line. It's not "here", the code _follows_ this block. I know that it's good to have copy before an example and after a heading, but make sure it adds value.
```suggestion
You can write a main isolate that resembles the following code:
```

**question:** What makes this the "main" isolate? Is it because it's in the `main` class? Does `main` = "first"? I would recommend clarifying this here and under **The main isolate**.

src/_guides/language/concurrency/index.md
```diff
@@ -312,107 +311,144 @@ to implement isolates.
...
-  final jsonData = await _parseInBackground();
+  final jsonData = await Isolate.run(_readAndParseJson);
 
-  // Use that data
+  // Use that data.
   print('Number of JSON keys: ${jsonData.length}');
 }
+```
+
+The spawned isolate executes the following code:
```
atsansone, line 346: **question:** Is this a spawned isolate or the main isolate? If it's the main, you may want to update this.

src/_guides/language/concurrency/index.md
```diff
@@ -312,107 +311,144 @@ to implement isolates.
...
-_spawns_ (creates and starts) the isolate for the background worker,
-and then returns the result:
+1. `Isolate.run()` spawns an isolate for the background worker,
+   while `main()` waits for the result.
 
-1. Before spawning the isolate, the code creates a `ReceivePort`,
-   which enables the worker isolate
-   to send messages to the main isolate.
+2. The new isolate executes the argument passed to `run()`:
+   the function `_readAndParseJson()`.
```
atsansone, line 361: **suggestion:** Recommend providing some consistency: "...spawns a new isolate..." and "The new isolate..." _or_ "...spawns an isolate..." and "The spawned isolate...".

atsansone: **question:** Reading the next step, when is the worker isolate created? Is that different than the main? It's unclear here.

src/_guides/language/concurrency/index.md
```diff
@@ -312,107 +311,144 @@ to implement isolates.
...
-   which creates and starts the isolate for the background worker.
-   The first argument to `Isolate.spawn()` is the function that
-   the worker isolate executes: `_readAndParseJson`.
-   The second argument is the `SendPort`
-   that the worker isolate can use to send messages to the main isolate.
-   The code doesn't _create_ a `SendPort`;
-   it uses the `sendPort` property of the `ReceivePort`.
+3. The `return` statement ends the computation,
+   shutting down the worker isolate and
+   sending `jsonData` back to the main isolate as the result.
```
atsansone, line 365: **suggestion:** Tighten the language somewhat.
```suggestion
3. The `return` statement shuts down the spawned isolate and
   sends `jsonData` back to the main isolate as the result.
```

src/_guides/language/concurrency/index.md
```diff
@@ -312,107 +311,144 @@ to implement isolates.
...
 
-3. Once the isolate is spawned, the main isolate waits for the result.
-   Because the `ReceivePort` class implements `Stream`,
-   the [`first`][] property is an easy way to get
-   the single message that the worker isolate sends.
+4. The memory holding the result in the worker isolate is *transferred*
+   to the main isolate, *not copied*.
+   The worker isolate still performs a verification pass to ensure
+   the objects are allowed to be transferred.
```
atsansone, line 370: **suggestion:** Clarify who has agency. Is it the worker isolate? If so, use the active voice here. Also, why 

Also, clarify what the work isolate is _doing_.
```suggestion
4. The worker isolate transfers the heap holding the result
   to the main isolate. It _does not copy_ the heap.
   The worker isolate verifies if it can transfer the objects.
```

src/_guides/language/concurrency/index.md
```diff
@@ -312,107 +311,144 @@ to implement isolates.
...
 
-[`first`]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-async/Stream/first.html
+`_readAndParseJson()` is an existing,
+asynchronous function that could just as easily
+run directly in the main isolate.
+But using `Isolate.run()` enables concurrency,
+so its computations are completely abstracted by the worker isolate
+and can complete without blocking the main isolate.
```
atsansone, line 377: **suggestion:** Tighten the language. Use active voice.
```suggestion
The existing `_readAndParseJson()` asynchronous function
could run within the main isolate.
Use `Isolate.run()` to enable concurrency.
This allows the worker isolate to abstract computations
and complete the work without blocking the main isolate.
```

src/_guides/language/concurrency/index.md
```diff
@@ -312,107 +311,144 @@ to implement isolates.
...
 
-The spawned isolate executes the following code:
+Because `Isolate.run()` allows code in the main isolate to continue running
+while its computations execute in a new one,
+its result is always returned as a `Future`.
+The original isolate doesn't care whether the computation
+running in the other isolate is synchronous or asynchronous,
+because it's being run concurrently either way.
```
atsansone, line 384: **suggestion:** This is a matter of tightening language again. Shorten the sentences, use active voice, avoid anthropomorphizing.
```suggestion
The `Isolate.run()` method returns a `Future`. 
This result returns because the main isolate continues to run
while computations execute in the new isolate.
The code running in the new isolate doesn't impact
 the original isolate.
```

src/_guides/language/concurrency/index.md
```diff
@@ -312,107 +311,144 @@ to implement isolates.
...
+  the main isolate and the worker isolate:
+  
+  ![A figure showing the previous snippets of code running in the main isolate and in the worker isolate](/guides/language/concurrency/images/isolate-api.png)
+{% endcomment %}
+
+#### Sending closures with isolates
+
+This section shows another simple worker isolate created with `run()`,
+this time written as a function literal, or closure,
+directly in the main isolate.
```
atsansone, line 401: **suggestion:** Avoid the "This section..." trap.
```suggestion
You can also write a simple worker isolate a
function literal, or closure, in the main isolate.
```

src/_guides/language/concurrency/index.md
```diff
@@ -312,107 +311,144 @@ to implement isolates.
...
-{{site.alert.version-note}}
-  `Isolate.exit()` was added in 2.15.
-  Previous releases support only explicit message passing,
-  using `Isolate.send()` as shown in the next section's example.
-{{site.alert.end}}
-
-The following figure illustrates the communication between
-the main isolate and the worker isolate:
+This example accomplishes the same as the previous:
+a new isolate spawns and executes a computation, sending back the result.
```
atsansone, line 419: **suggestion:** Avoid a compound sentence with a colon. If you can use a colon, you can split the sentence.
```suggestion
This example accomplishes the same as the previous.
A new isolate spawns, executes code, and returns the result.
```

src/_guides/language/concurrency/index.md
```diff
@@ -312,107 +311,144 @@ to implement isolates.
...
 
-![A figure showing the previous snippets of code running in the main isolate and in the worker isolate](/guides/language/concurrency/images/isolate-api.png)
+However, now the isolate sends a [closure][]. 
+This allows versatility not only for how the isolate can be utilized
+(as closures are less limited than typical named functions),
+but also for the written style of the code.
+Here, `Isolate.run()` executes what looks like local code, concurrently.
+In that sense, you can imagine `run()` almost like a [control flow operator][]
+for “run in parallel”.
```
atsansone, line 427: **suggestion:** This could also use tightening.

```suggestion
Now the isolate returns a [closure][]. 
Closures provide more options as to how to use the isolate
and how to write the code.
They have fewer limits than named functions.
In this example, `Isolate.run()` executes what looks like
local code in a concurrent manner.
In that sense, imagine `run()` to work like a [control flow operator][]
for “run in parallel”.
```

src/_guides/language/concurrency/index.md
```diff
@@ -312,107 +311,144 @@ to implement isolates.
...
+
+[`Isolate.exit()`]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-isolate/Isolate/exit.html
+[`Isolate.spawn()`]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-isolate/Isolate/spawn.html
+[`ReceivePort`]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-isolate/ReceivePort-class.html
+[`SendPort`]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-isolate/SendPort-class.html
+
+However, there are cases where it’s necessary
+to directly utilize the primitives that make up `run()`
+for more granular control over isolate management.
+Sending multiple messages between isolates is one of those cases.
```
atsansone, line 448: **suggestion:** Don't bury the lede. Also, simplify the language. (That is, "control" and "management" are synonyms.)
```suggestion
To provide more granular control of isolates might
require using the primitives that make up `run()`.
One example use case would be sending multiple
messages between isolates.
```

src/_guides/language/concurrency/index.md
```diff
@@ -312,107 +311,144 @@ to implement isolates.
...
+
+For more communication between isolates,
+you need to use the [`send()` method][] of `SendPort`.
 One common pattern, which the following figure shows,
 is for the main isolate to send a request message to the worker isolate,
 which then sends one or more reply messages.
```
atsansone, line 454: **question:** What does "for _more_ communication" mean? I would recommend finding a way to proffer an example as part of the explanation.

atsansone (changes requested):
@MaryaBelanger : There's a few changes here along the lines of agency, conciseness, and direct language. Happy to discuss.

src/_guides/language/concurrency/index.md
MaryaBelanger: Yes, they are ordered. Will do!

src/_guides/language/concurrency/index.md
MaryaBelanger: The main isolate is the isolate that every Dart program run in by default, so yes, the "first" or "starting" (defined higher up [on the same page](https://dart.dev/guides/language/concurrency#the-main-isolate)). I'll go through the page and make sure the terminology is consistent, thanks for putting it on my radar!

src/_guides/language/concurrency/index.md
MaryaBelanger: This is the spawned one, spawned by the main isolate.

src/_guides/language/concurrency/index.md
```diff
@@ -469,7 +506,6 @@ TODO:
 * Maybe:
   * Add a new macro & style for flutter notes?
   * Add the following text somewhere in this page (or in the FAQ):
-    * Sometimes Dart is called a _single-threaded language._
```
johnpryan, line 472: Are the rest of these TODOs still relevant?

src/_guides/language/concurrency/index.md
```diff
@@ -312,107 +311,144 @@ to implement isolates.
...
-(The [Flutter `compute()` function][] works in a similar way.)
+These examples show the implementation of a
+main isolate that spawns a simple worker isolate.
+The examples use [`Isolate.run()`][], 
+which efficiently abstracts the complexity
+of setting up and managing worker isolates manually:
 
-This example uses the following isolate-related API:
+* Spawns (starts and creates) an isolate
+* Takes a function argument and executes
```
johnpryan, line 323: This was hard for me to parse. These are steps that Isolate.run does for you right? It also sounds like we're saying it takes a function argument _after_ the isolate is spawned. 

Instead I'd probably say something like:
```suggestion
* Runs a function on the spawned isolate
```

src/_guides/language/concurrency/index.md
johnpryan, line 346: nitpick:
```suggestion
The spawned isolate executes the function passed as the first argument, `_readAndParseJson`:
```

src/_guides/language/concurrency/index.md
johnpryan: It's a little pedantic, but `return` doesn't shut down the isolate, Isolate.run() will take this result and call [exit](https://api.dart.dev/dev/2.19.0-388.0.dev/dart-isolate/Isolate/exit.html), which sends the value back to the main isolate.

johnpryan (commented):
looks good overall, Isolate.run() is much easier to use 👍

src/_guides/language/concurrency/index.md
MaryaBelanger: No, not for me at least. I'll incorporate the "Dart code executes..." line then remove the whole comment in my next edit.

parlough: Removing the TODOs and maybes seems good to me! The important ones are fixed or tracked already.

examples/concurrency/test/concurrency_test.dart
```diff
@@ -4,9 +4,11 @@ import 'package:concurrency_examples/async_number_of_keys.dart'
     as async_number_of_keys;
 import 'package:concurrency_examples/simple_worker_isolate.dart'
     as simple_worker_isolate;
+import 'package:concurrency_examples/simple_isolate_closure.dart'
+    as simple_isolate_closure;
 import 'package:test/test.dart';
 
-void main() {
+Future<void> main() async {
```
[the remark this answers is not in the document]
MaryaBelanger: I removed `async` here and got this error from `dart analyze`:

```
  error • test/concurrency_test.dart:11:14 • The body might complete normally, causing 'null' to be returned, but the return
          type, 'Future<void>', is a potentially non-nullable type. Try adding either a return or a throw statement at the
          end. • body_might_complete_normally
```

Nbd since it's just a nit, but any thoughts on that?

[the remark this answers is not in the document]
parlough: If the function is no longer `async`, you can just return `void` instead of `Future<void>`.

src/_guides/language/concurrency/index.md
MaryaBelanger: The section header is "Sending multiple messages between isolates", that's what "more communication" is in reference to. The context is that one message is the standard, more than one requires different configuration. I'll figure out how to include that

src/_guides/language/concurrency/index.md
MaryaBelanger: Is it a hard-must to get rid of every "However"? I do overuse it sometimes, but I feel like in this case it's somewhat warranted... 

Reading the line preceding this paragraph, and then following with "Now" feels abrupt and reads like an incomplete thought. 

> This example accomplishes the same as the previous. A new isolate spawns, computes something, and sends back the result.
> 
> Now the isolate sends a closure

Like, it doesn't sound like we're talking about the difference between the previous example and this one.
I want to keep the emphasis on the _difference_, not just stating what's happening for the sake of stating it. Is there another way to do that "feels" smoother?

src/_guides/language/concurrency/index.md
MaryaBelanger: I kept the first sentence as-is because the emphasis of the example is that you're just running normal functions you could use anywhere. I feel like the suggestion loses a bit of that emphasis. Tightened the rest though!

src/_guides/language/concurrency/index.md
MaryaBelanger: Incorporated these suggestions, except:
> Also, why

I think it's reasonable to assume the concept of _transferring_ memory rather than _copying_ data is an understood benefit in general software engineering (maybe I'm wrong?). I don't want to clutter the page by explaining it. Also, the rest of the examples don't get into the motivation behind the implementation details of isolates, so I'd rather avoid that here. That kind of background is more blog post material.

atsansone (approved):
LGTM. Tight copy.

johnpryan (approved):

MaryaBelanger:
@parlough could you help with / explain the failing check? I scanned the details but don't see exactly what the issue is

parlough:
@MaryaBelanger Hmm, the message doesn't show it well, but since `Isolate.run` doesn't exist, the tests for `examples/concurrency` will fail to run on stable.

I guess we can maybe configure it to show a compiler error, but usually analysis will catch any issues.

Issue #4361, opened 2022-11-11
[2.19] Document content-hashing in pubspec.lock


jonasfj:
Starting Dart 2.19 we're embedding a sha256 content-hashes in `pubspec.lock`.

We should document this, and create a link `dart.dev/go/content-hashes` pointing to the documentation. This link have already been added to messages in the pub tool and sdk CHANGELOG.

jonasfj:
Filed link in https://github.com/dart-lang/site-www/pull/4362

sigurdm:
The feature was implemented here: https://github.com/dart-lang/pub/pull/3482

parlough:
Without the new `pub get` option, what should be documented here? The fact that a sha256 hash of the package archive is included in the `pubspec.lock` description of each package that can be manually verified against?

We don't really have great centralized documentation of `pubspec.lock`. Perhaps the best place is the [Lockfile section of Package versioning](https://dart.dev/tools/pub/versioning#lockfiles).

johnpryan:
From a user perspective, how do you use this feature to verify the integrity of downloaded packages?

MaryaBelanger:
> We don't really have great centralized documentation of `pubspec.lock`. Perhaps the best place is the [Lockfile section of Package versioning](https://dart.dev/tools/pub/versioning#lockfiles).

That seems like the only decent place to put it at the moment. It might be a little too high level/conceptual if we want to document the "how" in @johnpryan's question (@jonasfj or @sigurdm it'd be wonderful to get that info to fill out the docs!)

_Another option:_
Currently the [pubspec page](https://dart.dev/tools/pub/pubspec) is just about `pubspec.yaml`, but maybe that'd be a good place to tack a `pubspec.lock` section at the end to cover the more technical/implementation side of `pubspec.lock`? That would also give us a good place to talk about the technical implementation of `--enforce-lockfile` that's also going into 2.19

_An aside:_
Isn't it odd that neither the [`dart pub get`](https://dart.dev/tools/pub/cmd/pub-get), [`upgrade`](https://dart.dev/tools/pub/cmd/pub-upgrade), or [`downgrade`](https://dart.dev/tools/pub/cmd/pub-downgrade) pages mention `pubspec.lock` at all, even though that's how the file gets created?

jonasfj:
> Currently the [pubspec page](https://dart.dev/tools/pub/pubspec) is just about pubspec.yaml, but maybe that'd be a good place to tack a pubspec.lock section

[...]

PR #4362, opened 2022-11-11
Add dart.dev/go/content-hashes

jonasfj:
Ensuring that the link doesn't go to 404, instead it'll go to the bug that we should write the page.

See https://github.com/dart-lang/site-www/issues/4361

Hopefully, this helps us not forget to write the page, and helps users highlight it if we do.
It's certainly better than having a 404.


sigurdm (approved):

parlough (approved):
Looks good to me. Thanks for the PR and issue!

PR #4365, opened 2022-11-12
Remove non-null-safe video playlist

parlough:
Many of these videos predate Dart 2 and all predate null safety. 

It would be nice to include an alternate, but better not to link to an outdated resource.


domesticmouse:
@craiglabenz do you want to take note of these videos, either for deletion or re-make?

domesticmouse (approved):

domesticmouse:
> @craiglabenz do you want to take note of these videos, either for deletion or re-make?

Cancel that request, these aren't our videos.

atsansone (approved):

Issue #3737, opened 2021-12-18
Clarify that `required` does not mean non-nullable


hexaron:
### Page URL

https://dart.dev/guides/language/language-tour#named-parameters

### Describe the problem

Using the `required` keyword does not force the parameter to be non `null`, it only forces a parameter to be specified.
Consider:
```dart
void main() {
  C c = C(x: null);
  
  print(c.x);
}

class C {
  dynamic x;
  
  C({required this.x});
}
```
This will compile and print out `null`.

### Expected fix

A warning that `required` is not the same as non-nullable.

### Additional context

In the last sentence of [How does @required compare to the new required keyword?](https://dart.dev/null-safety/faq#how-does-required-compare-to-the-new-required-keyword) it's hinted at the fact, that `null` is in fact valid:

> Instead, you could add a default or make the argument type nullable.

(To clarify: I believe this is expected behavior, as in the example above `dynamic` is actually a nullable type. I just think, that the documentation should be more explicit.)

parlough:
Thanks for opening an issue @hexaron! I can definitely see how this clarification and perhaps a nullable example as well would be useful to showcase.

Issue #4350, opened 2022-11-07
Improve discussion around `const` and `const` constructors in language tour


parlough:
### Page URL

https://dart.dev/guides/language/language-tour

### Page source

https://github.com/dart-lang/site-www/tree/main/src/_guides/language/language-tour.md

### Describe the problem

The documentation around `const` was never fully fleshed out and there are some questions remaining. 

Some current issues:

- [ ] When `const` should be used, potential trade-offs, etc (https://github.com/dart-lang/site-www/issues/1061)
- [ ] Limitations of `const` constructors (https://github.com/dart-lang/site-www/issues/2010)
- [ ] Better clarity on what happens when in a constant context vs not (https://github.com/dart-lang/site-www/issues/1859)
- [ ] Better clarity on when `const` isn't necessary (https://github.com/dart-lang/site-www/issues/1364)

### Expected fix

We should go through the language tour and make sure `const` is well explained as well as how and when to use it.

antfitch:
I think this was addressed as some point as I haven't seen any new asks since 2019 and we've got a section on it here that looks robust: https://dart.dev/language/variables#final-and-const

Tentatively closing.

PR #4366, opened 2022-11-12
Move default value for optional parameters discussion to each individual section

parlough:
Currently setting default values for optional named and positional parameters was separated, but it is pretty core to each topic, so this PR moves the discussion of default values to each of the parameter sections. This helps the examples make a bit more sense as they are contextualized.

I've also removed the mention of migrating from `=` to `:` because it was added over 6 years ago, most code already uses `:`, the extra mention is large and potentially confusing (and don't want to duplicate it in this new layout), and in Dart 2.19 the old syntax will become a warning anyway.

Also removes the mention of how list and maps can be passed. It doesn't talk about sets or other constant objects which can be passed as well and isn't super relevant to the goals of these sections. We should instead expand on our documentation of const (https://github.com/dart-lang/site-www/issues/4350) elsewhere and link to that.

Overall:
- Moves default value discussion to named and optional positional sections instead of isolated
- Better clarifies the nullable types and when they are needed, as well as how `null` is the default default value
- Moves discussion of named parameters anywhere to end
- Removes mention of old `=` syntax
- Removes mention of how list and maps can be passed as don't want to duplicate it and it is better documented by const documentation elsewhere

Fixes https://github.com/dart-lang/site-www/issues/3737

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -1316,7 +1316,7 @@ unless they're explicitly marked as `required`.
 When defining a function, use
 <code>{<em>param1</em>, <em>param2</em>, …}</code>
 to specify named parameters.
-If you don't provide a [default value](#default-parameter-values)
+If you don't provide a default value
 or mark a named parameter as `required`,
 their types must be nullable
 as their default value will be `null`:
@@ -1334,12 +1334,13 @@ For example:
 
 <?code-excerpt "misc/lib/language_tour/functions.dart (use-named-parameters)"?>
 ```dart
-enableFlags(bold: true);
+enableFlags(bold: true, hidden: false);
 ```
 
-To provide a default value to a named parameter besides `null`,
+<a id="default-parameters"></a>
+To define a default value for a named parameter besides `null`,
 use `=` to specify a default value.
-The specified default value must be a compile-time constant.
+The specified value must be a compile-time constant.
 For example:
 
 <?code-excerpt "misc/lib/language_tour/functions.dart (named-parameter-default-values)"?>
@@ -1368,7 +1369,9 @@ then the analyzer reports an issue.
   A parameter marked as `required`
   can still be nullable:
 
+  <?code-excerpt "misc/lib/language_tour/functions.dart (required-named-parameters)" replace="/required/[!$&!]/g; /ScrollbarTwo/Scrollbar/g;"?>
   ```dart
+  const Scrollbar({super.key, required Widget? child});
   ```
 {{site.alert.end}}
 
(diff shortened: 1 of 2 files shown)


src/_guides/language/language-tour.md
```diff
@@ -1315,7 +1315,11 @@ unless they're explicitly marked as `required`.
 
 When defining a function, use
 <code>{<em>param1</em>, <em>param2</em>, …}</code>
-to specify named parameters:
+to specify named parameters.
+If you don't provide a default value
+or mark a named parameter as `required`,
+their types must be nullable
+as their default value will be `null`:
```
atsansone, line 1322: **nit:** Little long, but it works.

src/_guides/language/language-tour.md
```diff
@@ -1333,45 +1337,62 @@ For example:
...
 If someone tries to create a `Scrollbar`
 without specifying the `child` argument,
 then the analyzer reports an issue.
 
-{% comment %}
-NULLSAFE: Rewrite this section.
-{% endcomment %}
+{{site.alert.note}}
+  A parameter marked as `required`
+  can still be nullable:
```
atsansone, line 1370: **question:** Is it best to say `is nullable` everywhere? Would it be more clear to say `...can be set to null`?

src/_guides/language/language-tour.md
```diff
@@ -1333,45 +1337,62 @@ For example:
...
+
+  <?code-excerpt "misc/lib/language_tour/functions.dart (required-named-parameters-nullable)" replace="/Widget\?/[!$&!]/g; /ScrollbarTwo/Scrollbar/g;"?>
+  ```dart
+  const Scrollbar({super.key, required [!Widget?!] child});
+  ```
+{{site.alert.end}}
+
+Although it often makes sense to place positional arguments first,
+named arguments can be placed anywhere in the argument list
+when it suits your API:
```
atsansone, line 1380: **suggestion:** This was a little hard to read.
```suggestion
You might want to place positional arguments first,
but Dart doesn't require it.
Dart allows named arguments to be placed anywhere in the
argument list when it suits your API.
```

atsansone (approved):
LGTM % a nit, suggestion, and a question. Change them if you want or not. Acknowledge what you're doing either way. I don't need a second pass.

johnpryan (approved):

src/_guides/language/language-tour.md
parlough: This is talking more about the type of the variable, not the variable itself, so it seems better to stick to nullable.

PR #4368, opened 2022-11-13
Mention generate constructor in null safety FAQ

### LINKED ISSUE #2837: FAQ (null safety) should mention `growable` optional parameter for `List.generate()`
With my understanding, `List<int>(10)` generates a fixed length list, ~~while `List<int>.filled(10,0)` generates a growable list which is very slow with subscripting.~~
~~The page should also mention `List<int>.generate(10, generator, growable: false)` replacing `List<int>(10)` in some case.~~
We sometime have to migrate unnamed `List` constructor to `generate()` to avoid initialize members twice.
But, the default of `growable` flag in `generate()` is `true`, and it is source of trouble, because subscripting on growable `List` is very slow.
I propose that the FAQ will introduce an additional example for migrating to `generate()` noting the default value of `growable`.

parlough:
Fixes #2837

### DIFF
--- a/src/null-safety/faq.md
+++ b/src/null-safety/faq.md
@@ -285,6 +285,21 @@ _jellyPoints = [
 ];
 {% endprettify %}
 
+If you need to generate a fixed-length list,
+you can instead use the [`List.generate`][] constructor
+and set the `growable` parameter to `false`:
+
+```dart
+_jellyPoints = List.generate(jellyMax, (_) => Vec2D(), growable: false);
+```
+
+[`List.generate`]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-core/List/List.generate.html
+
+{% comment %}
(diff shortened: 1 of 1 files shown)


domesticmouse (approved):

src/null-safety/faq.md
```diff
@@ -285,6 +285,21 @@ _jellyPoints = [
 ];
 {% endprettify %}
 
+If you need to generate a fixed-length list,
+you can instead use the [`List.generate`][] constructor
+and set the `growable` parameter to `false`:
```
atsansone, line 290: **suggestion:** Tighten this language. The conditional clause is not needed. You can just state this in the "why, how" construction. If they don't need it, they can move past it.
```suggestion
To generate a fixed-length list,
use the [`List.generate`][] constructor
with the `growable` parameter set to `false`:
```

atsansone (approved):
LGTM % a suggestion for more readability.

PR #4370, opened 2022-11-13
Update site-shared to remove now unused infrastructure files

parlough:
The goal is to reduce distractions while identifying or fixing infrastructure related issues.

Also reduces repository size as a small bonus.


parlough:
Superseded by https://github.com/dart-lang/site-www/pull/4372

Issue #4042, opened 2022-05-17
Dart is single-threaded


mit-mit:
I've seen statements on various social media a number of times stating that "Dart is single-thread". I've not before been able to understand where that impression comes from, but someone recently linked to this: https://dart.dev/articles/libraries/dart-io. This has in it: _"Dart is a single-threaded programming language"_.

I suggest we:

1. Add a new Q/A entry in https://dart.dev/faq under "Native execution" for "Is Dart single-threaded" and a good answer. The answer should link to https://dart.dev/guides/language/concurrency

1. Maybe also add a Q/A entry in the FAQ under web explaining that the limitations are

1. Determine if any content from https://dart.dev/articles/libraries/dart-io is still relevant, and if so move that elsewhere. Then take down that page entirely.

johnpryan:
This makes sense to me

johnpryan:
For 3, the [introduction to the dart:io](https://dart.dev/articles/libraries/dart-io) page is better covered in the [library tour](https://dart.dev/guides/libraries/library-tour#dartio) and the new [concurrency](https://dart.dev/guides/language/concurrency) docs.

sigmundch:
Agree, makes sense to me too!

atsansone:
@mit-mit : Would this issue be better served by splitting 1-4 as separate issues or should this all be done as one effort?

mit-mit:
I think a single issue is fine? I updated the top to have check-boxes!

mit-mit:
Making this P1 as I think it has the potential to cause confusion over Dart's architecture.

mit-mit:
cc'ing @craiglabenz who just did a talk on this topic and might have some ideas for how to phrase this?

MaryaBelanger:
_fyi [this blog](https://medium.com/dartlang/dart-asynchronous-programming-isolates-and-event-loops-bffc3e296a6a) also says Dart is single-threaded_

Questions about each of the work items:

### 1. FAQ

@mit-mit Dart _is_ single-threaded, isn't it? Would the answer here be something like: 

> Dart is single threaded, but supports [concurrent programming](https://dart.dev/guides/language/concurrency) in multiple ways

Or if we want to provide more information, this sentence (mostly copied right from the [language tour](https://dart.dev/guides/language/language-tour#isolates)): 

> Dart is single threaded; shared-state concurrency is error prone and can lead to complicated code.
>
[...]

PR #4374, opened 2022-11-14
Remove Intro to dart:io page

MaryaBelanger:
Intro to dart:io is outdated, and the important info it contained is available on other pages (the [library tour](https://dart.dev/guides/libraries/library-tour#dartio) and [Concurrency in Dart](https://dart.dev/guides/language/concurrency))

Part of #4042


parlough:
Makes sense to me! Are you able to remove the source of the excerpts in https://github.com/dart-lang/site-www/tree/main/examples/misc/lib/articles/io or are these used elsewhere?

MaryaBelanger:
@parlough I just checked (manually by searching the whole project for the file names, that should cover it right?) and nothing came up so I removed them. I ran `refresh-code-excerpts` and `check-links` too. `check-links` locally doesn't show any issues but I noticed the check didn't pass here, something on the what's new page?

parlough (approved):
Removing this makes sense to me and everything looks set, including the redirect, so looks good to me. Thanks!

PR #4375, opened 2022-11-14
Include link check in build to speed up actions

parlough:
This should hopefully be faster since it doesn't need to setup Dart, Node, and firebase-tools an extra time during staging or deploying.


domesticmouse (approved):

parlough:
This isn't seeing the speedup I'm looking for, perhaps due to caching the docker image doesn't have. Will reinvestigate this later.

PR #4376, opened 2022-11-15
Adding a doc outlining dart vs swift concurrency

leighajarett:
Creating a doc that outlines some of the differences in concurrency for Swift developers that are new to Dart. 

This was created in collaboration with @vandadnp and the content was reviewed by members of the Dart VM and Flutter iOS teams.

Let me know if there's a better place in the nav bar for this material

### DIFF
--- a/examples/concurrency/lib/async_weather.dart
+++ b/examples/concurrency/lib/async_weather.dart
@@ -1,0 +1,70 @@
+import 'package:flutter/cupertino.dart';
+
+void main() {
+  runApp(const MyApp());
+}
+
+class MyApp extends StatelessWidget {
+  const MyApp({super.key});
+
+  @override
+  Widget build(BuildContext context) {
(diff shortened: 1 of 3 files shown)


parlough:
Before I review this more thoroughly, is this only relevant for iOS app developers or is it relevant for Swift developers on other platforms? I ask because in its current state, it seems very Flutter specific, with mentions like "Flutter’s event loop is", using Flutter to display the results in a UI, "Flutter apps can run on a variety of multicore hardware", and the `compute` function.

We don't have a precedent for such Flutter focused documentation on the site, so I'm just making sure this is the best place for it.

\cc @mit-mit

mit-mit:
Yeah, this seems very Flutter specific to me, so I also think this should be on flutter.dev.

leighajarett:
Makes sense! Happy to move it over to the Flutter website instead, thanks!

PR #4395, opened 2022-11-22
[2.19] New restriction for noSuchMethod

### LINKED ISSUE #4263: [2.19] `noSuchMethod` docs
## This is an _evaluation_ ticket

Evaluating the _possible_ need for doc updates due to dart-lang/sdk#49687

### Summary

Behavior (implementing class using overriding private fields) that used to delegate to `noSuchMethod` now just throws an exception. Allowing this behavior to be handled by `noSuchMethod` previously posed a privacy gap, and made the recent implementation of field promotion unsound. 

### Evaluation

- This is closing a privacy gap that shouldn't have existed, as well as allowing implementation of something that didn't exist previously ([#2020](https://github.com/dart-lang/language/issues/2020)), so there's nothing documented already that needs to be updated. 

- I would like to know the message of the exception that's thrown, might fit into docs somewhere.

- https://dart.dev/guides/language/language-tour#nosuchmethod
  > You can’t invoke an unimplemented method unless one of the following is true:
  
  Maybe add "you can't invoke an unimplemented interface name that's private to a different library" (but i feel like that statement needs more elaboration...?). 

### References

- Implemented to support dart-lang/language#2020

MaryaBelanger:
Closes #4263 

I'm not sure if this makes sense @stereotype441, but there's this section that presents some caveats to using `noSuchMethod` and I thought the new restriction to invoking private members might be useful to add.

The wording is awkward and I can adjust that, but does the purpose make sense in light of [your recent changes](https://github.com/dart-lang/sdk/issues/49687)?

Another way I thought to add it was injected into the original caveat, like:

> You **can't invoke** an unimplemented method unless
the members you're invoking are **not private**, and
**one** of the following is true:

But that might be more awkward or might not even make sense. Let me know!

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -3564,6 +3564,8 @@ defines the unimplemented method (abstract is OK),
 and the dynamic type of the receiver has an implementation of `noSuchMethod()`
 that's different from the one in class `Object`.
 
+Additonally, you **can't invoke** private members of an implemented interface, **ever**.
+
 For more information, see the informal
 [noSuchMethod forwarding specification.](https://github.com/dart-lang/language/blob/master/archive/feature-specifications/nosuchmethod-forwarding.md)
 


src/_guides/language/language-tour.md
```diff
@@ -3564,6 +3564,8 @@ defines the unimplemented method (abstract is OK),
 and the dynamic type of the receiver has an implementation of `noSuchMethod()`
 that's different from the one in class `Object`.
 
+Additonally, you **can't invoke** private members of an implemented interface, **ever**.
```
stereotype441, line 3567: I don't think this is accurate.  You *can* invoke private members of an implemented interface, if you are in the same library (and hence you can mention the private name), e.g. this totally works:

```dart
class C {
  void _privateMember() {
    print('C._privateMember() called');
  }
}

class D implements C {
  @override
  dynamic noSuchMethod(Invocation invocation) {
    print('D._noSuchMethod() called');
  }
}

main() {
  C()._privateMember(); // Prints `C._privateMember() called`
  D()._privateMember(); // Prints `D._noSuchMethod() called`
}
```

The new behaviour from https://github.com/dart-lang/sdk/issues/49687 doesn't change whether you can invoke a private member of an implemented interface; it just changes what happens if you do, under one very specific circumstance: if you invoke a private member of a class, and that class has an implemention of `noSuchMethod` but doesn't implement that member, _and that member is private to a different library than the one the class is declared in_, then instead of the invocation being routed to your `noSuchMethod` member, an exception will be thrown.

That's quite a mouthful, and honestly it's quite a rare situation.  IMHO there's no need to change the language tour to mention this minor little detail, because the number of people are going to run across it is really quite low.

MaryaBelanger: Ok thanks for explaining! I figured it was unnecessary but just wanted to make sure. Closing this

Issue #4348, opened 2022-11-07
Update and improve the zones page


parlough:
Page URL: https://dart.dev/articles/archive/zones

Page source: https://github.com/dart-lang/site-www/tree/main/src/_articles/archive/zones.md

The [Dart Zones](https://dart.dev/articles/archive/zones) page is the only document covering zones but it is out of date, has issues, and is not following the website's current documentation and writing guidelines. We should update or completely replace it.

Some issues that the current page has for reference:

- [ ] Code does not follow Dart 2 standards, formatting, or suggested linter rules
- [ ] Some code/explanations are now incorrect
- [ ] Code is not tested
- [ ] Writing and structure does not match current site standards
- [ ] "A zone represents the asynchronous dynamic extent of a call" section is too technical and doesn't explain vocabulary (https://github.com/dart-lang/site-www/issues/2147)
- [ ] "Child zones don’t completely replace their parent zone" is not clear enough about what is replaced and what isn't (https://github.com/dart-lang/site-www/issues/2149)
- [ ] "A simple rule determines where errors are handled in a sequence" does not explain what the simple rule is (https://github.com/dart-lang/site-www/issues/2150)
- [ ] "the original zone must be the one that executes the microtask" has no explanation of why (https://github.com/dart-lang/site-www/issues/2151)
- [ ] "The run parameter also works for the initial, synchronous code that executes just after calling runZoned()." doesn't explain what the synchronous code is (https://github.com/dart-lang/site-www/issues/2153)
- [ ] `runZoned` will no longer have an `onError` parameter and must use `runZoneGuarded` (https://github.com/dart-lang/site-www/issues/4224)
- [ ] Error zones may be removed from zones (https://github.com/dart-lang/sdk/issues/49457)

PR #4403, opened 2022-11-29
[2.19] Add `runZonedGuarded` to Zones page

### LINKED ISSUE #4224: [2.19] runZoned docs
### Parent: 
https://github.com/dart-lang/sdk/issues/49679

### Evaluation: 
- Update needed:
    - https://dart.dev/articles/archive/zones : `runZoned` with `onError` used
    - ~Page is archived, what's the policy there? Update and take out of archive?~
      - Will take care of in a separate issue: https://github.com/dart-lang/site-www/issues/4348
- ~No entry in change log yet, add~
  - The only changes here are to the API docs, `onError` was already deprecated in a past release. I don't think cathcing up on API docs changes warrant an item in the change log 
  - But still warrants mention in 'What's new' page

### Reference:
- API docs already updated, use new syntax: https://github.com/dart-lang/sdk/commit/59cd692fb996f7b4f534a11ec9e270fcea07629e
- [This issue](https://github.com/dart-lang/sdk/issues/40681) better describes the change from `runZoned` with `onError` to `runZoned` and `runZonedGuarded`
- Some examples of `runZonedGuarded` in use:
  - [errorZone API](https://api.dart.dev/dev/2.19.0-437.0.dev/dart-async/Zone/errorZone.html)
  - [`(error, stackTrace)` example](https://github.com/privacybydesign/irmamobile/commit/577c26320a3ebdc7faa74cfe990efbdf70923d4a)
  - [`(dynamic error, StackTrace stackTrace)` example](https://github.com/aam/engine/commit/9a49932bf69b72c80eebb04fca22629ce7e7f2a8)
  - [other error handler output](https://github.com/abegehr/monsterjump/commit/12be8147386c6c9538fdd868eca8a13e880acef5)

MaryaBelanger:
This page obviously needs a lot of work, so I didn't update any of the conceptual content or any other dart syntax stuff _except_ what is directly related to changing the examples to using `runZonedGuarded`.

Full page update is tracked here: https://github.com/dart-lang/site-www/issues/4348

Closes #4224


parlough:
You're right, the code examples aren't currently linked to anything or tested. I'm assuming some of them don't even compile anymore or are incorrect. I haven't had a chance to add code excerpts yet but I'm tracking the work in https://github.com/dart-lang/site-www/issues/4348. Adjusting the code in the Markdown should be sufficient for this change. If you want me to prioritize the code excerpts, updating the code, and testing, let me know.

src/_articles/archive/zones.md
```diff
@@ -25,24 +25,28 @@ _Written by Florian Loitsch and Kathy Walrath<br>
...
 [Futures and Error Handling](/guides/libraries/futures-error-handling).
 
+[`runZoned()`]: ({{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-async/runZoned.html)
+[`runZonedGuarded()`]: ({{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-async/runZonedGuarded.html)
+
 Currently, the most common use of zones is
 to handle errors raised in asynchronously executed code.
 For example,
 a simple HTTP server
 might use the following code:
```
atsansone, line 41: **suggestion:** This reads longer than needed. Consider tightening the language. 
```suggestion
Many developers use zones to handle errors that asynchronously executed code raises.
For example, a simple HTTP server might use the following code.
```

src/_articles/archive/zones.md
```diff
@@ -165,16 +171,24 @@ Another way to produce an uncaught error is
 to call `new Future.error()` or
 a Completer's `completeError()` method.
 
-Use the `onError` argument to `runZoned()` to
-install a _zoned error handler_—an asynchronous error handler
+`runZoned()` does not handle errors.
+Its error-handling parameter `onError` 
+was deprecated because it requires the
+function's return type to always be nullable,
+which is unneccesarily cumbersome for [Null Safe code][].
```
atsansone, line 178: **question:** When was `onError` deprecated and by whom? If this deprecation occurred recently, call that out.

**suggestion:** Provide context and tighten up the writing.

```suggestion
`runZoned()` deprecated its error-handling parameter `onError`.
That parameter required the function's return type to be nullable.
That requirement proves cumbersome for [Null Safe code][]. 
```

src/_articles/archive/zones.md
```diff
@@ -193,9 +207,9 @@ they still execute.
...
 
-Also note that an _error zone_—a zone that has an error handler—might
-handle errors that originate in a child (or other descendant)
-of that zone.
+Also note that an _error zone_ — a zone that has an error handler
+— might handle errors that originate in a child
+(or other descendant) of that zone.
 A simple rule determines where
 errors are handled in a sequence of future transformations
 (using `then()` or `catchError()`):
```
atsansone, line 215: **suggestion:** Rephrase to tie this back to the last sentence. Also, the callout is the rule, so you don't need to preface it with a sentence, then a callout. It's too much. 

```markdown
A zoned error handler exists in an _error zone_.
This error zone might handle errors that originate
in a descendant of that zone.
```

atsansone: **suggestion:** Maybe include the sentence after the note within the note. That would tie the thoughts together better.

src/_articles/archive/zones.md
```diff
@@ -271,20 +285,20 @@ Consider this example:
...
 zoneFuture.catchError((e) { print('Never reached'); });
 completer.complete(499);
 {% endprettify %}
 
 Even though the future chain ends in a `catchError()`,
 the asynchronous error can't leave the error zone.
-Instead, the zoned error handler (the one specified in `onError`)
+Instead, the zoned error handler
+(the one specified in `runZonedGuarded()`)
 handles the error.
```
atsansone, line 300: **suggestion:** You don't need the instead or the parenthetical. Both imply that some of this text can be ignored. It really can't. Also, the parenthetical disrupted your flow, because otherwise "The zoned error handler handles the error" would have stood out more.

```suggestion
The zoned error handler found in `runZonedGuarded()`
responds to an error.
```

src/_articles/archive/zones.md
```diff
@@ -302,21 +316,21 @@ streams should have no side effect until listened to.
 A similar situation in synchronous code is the behavior of Iterables,
 which aren't evaluated until you ask for values.
 
-### Example: Using a stream with runZoned()
+### Example: Using a stream with runZonedGuarded()
 
-Here is an example of using a stream with `runZoned()`:
+Here is an example of using a stream with `runZonedGuarded()`:
```
atsansone, line 321: **issue:** I understand needing a sentence between the title and the code, but this reads like a rehashing of the title. Is there any way to rephrase or expand upon this so it reads like less of a "I had to do this" sentence?

atsansone (changes requested):
@MaryaBelanger : I noted some readability issues. Have a look. I leave most as suggestions in that what I added is a possibility, but consider working with the language in each comment.

src/_articles/archive/zones.md
lrhn: It happened as part of Dart 2.12 (null safety), because the existing API just didn't work very well with null safety.

src/_articles/archive/zones.md
```diff
@@ -165,16 +171,24 @@ Another way to produce an uncaught error is
...
+`runZoned()` does not handle errors.
+Its error-handling parameter `onError` 
+was deprecated because it requires the
+function's return type to always be nullable,
+which is unneccesarily cumbersome for [Null Safe code][]. 
+
+[Null Safe code]: /null-safety/index
+
+Instead, use `runZonedGuarded()` to install a
+_zoned error handler_ — an asynchronous error handler
```
lrhn, line 183: Might be worth saying that it catches both uncaught async errors *and* uncaught synchronous errors (which is why it needs the nullability in the return type, something throwing synchronously doesn't have value to return).

In comparison `runZoned` *can* have an error handler (because you can install one as part of the  `ZoneSpecification`), but that only works for asynchronous errors.

Let's take a step back.

The title of this section is "handling async errors".

The text here is mainly about *uncaught* async errors and zones handling those, which makes sense since the file is about zones.
Maybe the title should include "uncaught", otherwise it seems like simple error handling like `try`/`catch` in an `async` function would also be relevant. (And even if only looking at zones, `Zone.errorCallback` might be relevant too, and I don't think we need to burden readers with that. It's too specialized.)

Focusing on uncaught errors, and handling them, the section can then either show *all* the ways to do that, or it can focus on the most common one.
Which one do we want?

"All the ways" include `runZoned`, `runZonedGuarded`, and `Zone.fork`, and mentions `ZoneSpecification.uncaughtErrorHandler`.

I think I'd just focus on `runZonedGuarded`, as the easiest and most convenient way, and refer only to the others in a "see also" section.

In which case, we don't need to compare `runZoned` and `runZonedGuarded`, just say how `runZonedGuarded` works. (Even if we do mention the others, having the section above starting out telling you what *not* to do seems like wasted prime real-estate, it's better to start saying what you _should_ do. Then the reader can stop at any point after that and still come away with the answer they need.)

So, maybe:

> One of the most used features of zones is
> their ability to catch and handle uncaught errors.
>
>  An _uncaught error_ is often caused by some code using `throw`
> to raise an exception that is not handled by a `catch` statement.
> In an `async` function, such an error usually ends up as the result of a `Future`,
> which allows you to handle the error where you `await` the future.
> Sometimes an error escapes, and sometimes a future is not `await`ed, 
> in which case it becomes an uncaught error.
>
> Such an error can crash your program.
>
> A zone can have an _uncaught error handler_ installed when the zone is created.
> If an uncaught error occurs in code which runs in that zone, 
> the uncaught error handler is called instead of just crashing the program.
> (In fact, the code which crashes the program on an uncaught error
> is just the uncaught error handler of the root zone, 
> which any zone inherits if it doesn’t specify another handler).
>
> The easiest way to introduce a new zone with an uncaught error handler
> is to run code using `runZoneGuarded`. It’s `onError` callback becomes the
> uncaught error handler of a new zone, and it’s *also* used to handle any
> synchronous errors thrown by the call.
>
> Example:
>
> ... the timer example ...

In any case, don't focus on what doesn't work, just say what does, and how to do it.
Then maybe add a section saying

> See also `Zone.fork`, `Zone.runGuarded` and `ZoneSpecification.uncaughtErrorHandler`.

MaryaBelanger:
> Visit the preview URL for this PR (updated for commit [cf8f4d5](https://github.com/dart-lang/site-www/commit/cf8f4d58040610c5aae0328e8b5d56c0a4651984)):
> 
> https://dart-dev--pr4403-zones-onerror-0aw72vy3.web.app
> 
> (expires Tue, 06 Dec 2022 19:42:38 GMT)
> 
> 🔥 via [Firebase Hosting GitHub Action](https://github.com/marketplace/actions/deploy-to-firebase-hosting) 🌎
> 
> Sign: d851bc446d3c4d7394c5406c6f07255afc7075f3

Unrelated to the actual topic. but wondering about this. @parlough I don't think I explicitly deployed to a firebase project of mine, does this just happen automatically?

parlough:
If the build and tests pass, changes on a branch of dart-lang/site-www will be automatically staged on a temporary Firebase preview channel. Helps you avoid having to rebuild and deploy for every commit and makes it so some contributors don't necessarily need a local setup, which can be cumbersome.

src/_articles/archive/zones.md
MaryaBelanger: going to rewrite this whole section per Lasse's comment below

src/_articles/archive/zones.md
MaryaBelanger: @lrhn I only slightly reworded some parts of this. It's not necessary because I don't think I changed any meaning, but if you have a minute to confirm [this section](https://github.com/dart-lang/site-www/pull/4403/commits/5c1173dfdc70060289eb1b280be6991f3b48cfed#diff-2b7e547663296c3749c4b011b531d9163e186954912d493b1ecaa60063b013abR157) is still accurate that would be great thank you!

src/_articles/archive/zones.md
```diff
@@ -152,33 +154,51 @@ Code might execute in other nested child zones as well,
...
+rather than just crashing the program,
+it calls its uncaught error handler.
+(In fact, any zone that doesn't specify a handler
+inherits the uncaught error handler of the root zone,
+which just crashes the program on an uncaught error).
+
+The simplest way to introduce a new zone with an uncaught error handler
+is to use `runZoneGuarded`. Its `onError` callback becomes the
+uncaught error handler of a new zone, and handles any
+synchronous errors thrown by the call as well.
```
atsansone, line 183: **question:** Are there other ways to introduce a new zone with UEH? If so, do they matter in the context of this section? If they don't, be direct.

```markdown
To create a new zone with an ... , add a `runZoneGuarded` ...
```

src/_articles/archive/zones.md
```diff
@@ -152,33 +154,51 @@ Code might execute in other nested child zones as well,
...
   Timer.run(() { throw 'Would normally kill the program'; });
-}, onError: (error, stackTrace) {
+}, (error, stackTrace) {
   print('Uncaught error: $error');
 });
 {% endprettify %}
 
+_See also [`Zone.fork`][], [`Zone.runGuarded`][]
+and [`ZoneSpecification.uncaughtErrorHandler`][]
+for more uncaught error handling methods in zones._
```
atsansone, line 196: **issue:** Why, then how.
```suggestion
To learn more about uncaught error handling methods in zones,
see [`Zone.fork`][], [`Zone.runGuarded`][]
and [`ZoneSpecification.uncaughtErrorHandler`][]
```

**nit:** I really dislike `For more...`. It lacks agency.

src/_articles/archive/zones.md
```diff
@@ -193,16 +213,13 @@ they still execute.
...
+An error zone might handle errors that originate
+in a descendant of that zone.
 A simple rule determines where
 errors are handled in a sequence of future transformations
 (using `then()` or `catchError()`):
-
-{{site.alert.note}}
-  Errors on Future chains never cross the boundaries of error zones.
-{{site.alert.end}}
+Errors on Future chains never cross the boundaries of error zones.
```
atsansone, line 222: **suggestion:** Consider framing this as to why the reader would need an error zone, then define an error zone.

src/_articles/archive/zones.md
```diff
@@ -302,21 +318,22 @@ streams should have no side effect until listened to.
 A similar situation in synchronous code is the behavior of Iterables,
 which aren't evaluated until you ask for values.
 
-### Example: Using a stream with runZoned()
+### Example: Using a stream with `runZonedGuarded()`
 
-Here is an example of using a stream with `runZoned()`:
+Here's an example that sets up a stream with a callback,
+and then executes that stream in a new zone with `runZonedGuarded()`:
```
atsansone, line 324: **nit:** Recommend indicating where the example is in the page.
```markdown
The following example sets...
```

src/_articles/archive/zones.md
```diff
@@ -302,21 +318,22 @@ streams should have no side effect until listened to.
...
-runZoned(() { stream.listen(print); },
-         onError: (e) { print('Caught error: $e'); });
+runZonedGuarded(() { stream.listen(print); },
+         (e) { print('Caught error: $e'); });
 {% endprettify %}
 
 The exception thrown by the callback
-is caught by the error handler of `runZoned()`.
+is caught by the error handler of `runZonedGuarded()`.
 Here's the output:
```
atsansone, line 337: **suggestion:** Use active voice.
```markdown
The error handler in `runZonedGuarded()` catches the error that the callback threw.
This results in the following response.
```

atsansone (changes requested):
@MaryaBelanger : Solid content! Just needs some tightening. I explain why instead of just offering alternatives. See what you think?

src/_articles/archive/zones.md
```diff
@@ -21,28 +21,28 @@ obsolete: true
...
 
-Currently, the most common use of zones is
-to handle errors raised in asynchronously executed code.
+[`runZoned()`]: ({{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-async/runZoned.html)
+[`runZonedGuarded()`]: ({{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-async/runZonedGuarded.html)
+
+Developers commonly use zones to handle uncaught errors.
 For example,
 a simple HTTP server
 might use the following code:
```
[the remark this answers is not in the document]
MaryaBelanger: Aha, I see. So my issue with the last suggestion AND the way I rewrote it here is because I feel like making it active makes the statement about _developers_' personal preference more than about zones, so it...reads weird to me. 

I'm not sure if I'm explaining that well (?), but basically I don't know how to make it more zone-centric. What do you think of these?
> Zones handle uncaught errors

or

> Zones are useful for handling uncaught errors

or to stick to the original sentence the most

> The most common use of zones is to handle uncaught errors

src/_articles/archive/zones.md
MaryaBelanger: This part isn't about why you'd need error zones; "error zone" is just a zone that has an error handler, which is defined way earlier (it's basically the point of the whole page up to here). 

This is just to introduce the term "error zone" because it helps conceptualize the fact that a handler might end up handling errors from descendent zones, and how to assess the chain of handling.

src/_articles/archive/zones.md
MaryaBelanger: These are all the other ways to handle uncaught errors with zones, so not just "To learn more about..." but.... idk how to phrase it otherwise.

Maybe one of these?
> You can handle uncaught errors using other zone APIs, including ...

or

> Other methods of handling uncaught errors with zones are ...

src/_articles/archive/zones.md
MaryaBelanger: Yeah there are, I mentioned in the other comment about "See also...". `runZonedGuarded` is the best/most convenient

src/_articles/archive/zones.md
```diff
@@ -152,33 +154,51 @@ Code might execute in other nested child zones as well,
...
-
-Use the `onError` argument to `runZoned()` to
-install a _zoned error handler_—an asynchronous error handler
-that's invoked for every uncaught error in the zone.
-For example:
+their ability to catch and handle uncaught errors,
+both synchronous and asynchronous.
+An _uncaught error_ is often caused by code using `throw`
+to raise an exception without an accompanying `catch`
+statement to handle it.
```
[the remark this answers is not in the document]
MaryaBelanger: > Consider the asynchronous and synchronous clause. Wouldn't that be all errors? If so, exclude this clause.

Yeah I also just realized that sync and async is only true of `runZoneGuarded` anyway 

> Consider providing more context and chunking to get to an uncaught error. Explain it, then define it.

The text in the suggestion spends a lot of effort basically just explaining how error handling works, which is basic cs, not unique to Dart or zones, so doesn't need to be described. 

> Consider moving how zones help to the end. The prior sentences could build up to why it matters.

I don't really understand this. I'm in the school of main-point-first, details/why-following, because people are almost always skimming. Is this a principle we're using in practice? Can you show me other examples? This seems like a pretty obfuscating method for writing technical content.

src/_articles/archive/zones.md
```diff
@@ -152,33 +154,51 @@ Code might execute in other nested child zones as well,
...
+
+In `async` functions, errors usually end up as the result of a Future.
+Futures allow you to handle errors where you `await` the future.
+Sometimes an error escapes, or a future is not `await`-ed,
+in which case an error becomes an uncaught error.
+Uncaught errors can crash your program.
```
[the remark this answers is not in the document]
MaryaBelanger: Oh I think this is unclear. "errors usually end up as the result of a Future." doesn't mean the Future is causing errors, it means that when an error happens in async code, it is returned as a result _to_ a Future. Is that what you meant by "qualifier"?

I could reword it:
> In `async` functions, errors are usually returned as a result to a Future.

src/_articles/archive/zones.md
```diff
@@ -152,33 +154,51 @@ Code might execute in other nested child zones as well,
...
+
+Zones can install an _uncaught error handler_ upon creation.
+When an uncaught error occurs in that zone,
+rather than just crashing the program,
+it calls its uncaught error handler.
+(In fact, any zone that doesn't specify a handler
+inherits the uncaught error handler of the root zone,
+which just crashes the program on an uncaught error).
```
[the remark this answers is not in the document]
MaryaBelanger: > That last sentence didn't make sense. If the zone inherits the root zone's UEH, then why would it crash at all? Does it crash if the root zone doesn't have an UEH?

This was a little confusing for me to wrap my head around too, and they way I got it out definitely isn't ideal. 

Root zones have an implicit (or maybe "default" or "built-in" is a better term) uncaught error handler. That UEH also handles uncaught errors: it handles them by crashing the program. So it's saying that user-created zones inherit that implicit UEH from the root, and to explicitly define one is to override the root's implicit one, to do something other than crash (is that correct @lrhn ?)

To me, it's like a neat little implementation detail of the language (which is why it's in parenthesis, it actually can probably be ignored). Like saying that even crashing is a form of error handling because Dart had to be designed to react, that reaction doesn't come out of no where.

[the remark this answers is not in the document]
MaryaBelanger: Looking again, idk why I rewrote it the way I did because Lasse's original wording might make it more clear, what do you think?

> (In fact, the code which crashes the program on an uncaught error
is just the uncaught error handler of the root zone,
which any zone inherits if it doesn’t specify another handler).

src/_articles/archive/zones.md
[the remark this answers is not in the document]
atsansone: **reply:** That's definitely better. I would adjust, though, to clarify what has agency here. What returns an error? The function, then maybe:
```markdown
Async functions tend to return errors due to a Future.
```

I would get away from `usually` or adverbs in general. Find a better verb, because something like `usually` begs for `what about unusually`?

src/_articles/archive/zones.md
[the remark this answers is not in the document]
atsansone: **reply:** This makes sense, but reads like a meandering path. Maybe:
```markdown
The root zone handles any uncaught errors when another zone
lacks an explicit uncaught error handler.
```

I will cop to the fact that I despise parentheticals in technical documentation. Asides make little sense in these contexts. Explain what needs to be understood. If it's worth writing, it's worth writing as if you intend it to be read.

src/_articles/archive/zones.md
atsansone: **reply:** OK, that's a bit different. I like the first one or maybe:
```markdown
Zones can also handle uncaught errors using...
```

src/_articles/archive/zones.md
atsansone: **reply:** Maybe consider defining this with a little more clarity.
```
An _error zone_ covers any zone with an error handler.
```

...that's... not great on my part. But do you see the point?

src/_articles/archive/zones.md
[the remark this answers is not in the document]
atsansone: **reply:** OK, I get the distinction. How about...
```markdown
Dart needs a way to handle uncaught errors. 
Zones provide the most common solution to this problem.
```

src/_articles/archive/zones.md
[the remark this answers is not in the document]
atsansone: **reply:** I see your point. I tend to be in the "why should I care about this feature / item / topic" and like to lead folks a _little_ bit. If you just state what it is, I feel that conveys a finality to the text. It feels like a closed door to the face. I admit some of this comes from my background in ops where something tacitly known by devs doesn't get explained to others sometimes. I think that a little build up, a little context, makes the point more clear and better understood.

That said, I don't think that this is a hill for me to die on here and now. I think we can and should discuss it further later. If you're ok with it, let it ride. We can always come back to it.

**comment:** The original text does explain throw and catch, too, but "after the fact", which is why I recommended the switch.

atsansone (changes requested):
@MaryaBelanger : Great work and great points! I updated with a few questions and comments.

src/_articles/archive/zones.md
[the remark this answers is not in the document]
lrhn: Being pedantic ... Dart is a programming language. It doesn't need anything. (And it doesn't like being anthropomorphized ;).
Also, `try`/`catch` is the most common solution to catching errors. We haven't defined what an "uncaught error" is, so the reader should assume that it's any error which isn't caught (by a `catch`).

So, maybe:

> Uncaught errors are thrown errors which reach the event loop, 
> or error results of futures that nobody waits for.
> An uncaught error is always reported to the current zone of the code which failed to catch it.
> The default behavior is to terminate a native program, or show error messages in the browser console.
> 
> By running code in a custom zone, with a custom _uncaught error handler_ installed,
> you can intercept and handle the uncaught errors in the way you prefer.

src/_articles/archive/zones.md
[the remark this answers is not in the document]
lrhn: Nit: Thrown `Error`s do not *need* a corresponding `catch`. Errors shouldn't happen, so you shouldn't try to catch them. The person writing the `throw` should definitely never worry about that error being caught.

(You can have a global "catch all errors to prevent the app from crashing" error handler.)

Also, we don't throw "another exception" (the `_uncaught error_`). That's just a message printed when an uncaught error is reported.

Exceptions should always be caught, close to where they are thrown. But again, it's not the author writing the `throw` who writes the `catch`, because exceptions are really intended to be used across API boundaries.

So, the phrasing here, which links "when your code throws" to "needs related `catch`", conflates the responsibilities of two different authors, working on different sides of an API boundary.


It's tricky to say something general and decisive, because our recommendations are to not catch errors, but application developers really do like their apps to not crash, so they will catch errors and try to recover. The problem with errors is that it's not always clear whether you *can* recover.


I'd probably go back to what I wrote above, defining an uncaught error as:

> _Uncaught errors_ are `throw`s which reach the event loop without being caught by a `catch`.
> The can also come from a future completed with an error result, when nobody waits for that future.

That defines what the problem is, and gives it a name. The reader should agree to, and understand why, an uncaught error is a problem.
It doesn't say how to avoid the problem (well, `catch` and waiting for the future, but not in any detail).

> The current zone handles all uncaught errors. The default is to crash a native application.
> You can add your own custom _uncaught error handler_ to a new zone, handling them in your own way,
> to avoid uncaught errors crashing your program.

src/_articles/archive/zones.md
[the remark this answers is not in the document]
MaryaBelanger: I think your suggestion is missing the meaning again, because of `due to`. The future is not causing anything here, it's just what async function results _are_. It's like a return type almost. 

The sentence is also not about async functions, it's about the errors: what happens with uncaught errors in aysnc functions. 

I actually think the original way the sentence is written is kind of the only good way to capture the point, so I'm going to change `usually` to `tend to` but otherwise leave it as is. I think the confusing part here is that I separated a continuous thought with a new paragraph, so I undid that. Maybe with the previous sentence connected, it's clearer:
> An _uncaught error_ is often caused by code using `throw`
to raise an exception without an accompanying `catch`
statement to handle it.
In `async` functions, these errors tend to end up as the result of a Future.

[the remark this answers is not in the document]
MaryaBelanger: Ahh wait I get it now, my interpretation was incorrect. It's "Uncaught errors in async functions tend to happen when attempting to handle errors with Futures" ..... I think. 

I think I can actually rewrite that in the sentence form you suggested. And it's not about the paragraphs being separated, they actually should be separate.... I'll re-request your review on the next commit

src/_articles/archive/zones.md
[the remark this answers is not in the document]
atsansone: **reply:** The GSG covers [anthropomorphism](https://developers.google.com/style/anthropomorphism), but I'd argue "need" is not a human quality per se. Maybe `require` would be better. 

Another important distinction: we can't use code elements as nouns or verbs. See the [GSG](https://developers.google.com/style/code-in-text#grammatical-treatment-of-code-elements) on that point.

That said, I would give a bit more context, even though it is basic CS, because:

1. It helps with flow.
2. Not everyone for whom we write understands all CS concepts.

May I suggest:

```markdown
Good programming practice includes providing ways to handle errors in code.
When code tries to handle an error, this process is called a `throw`. The target of that `throw`
would be the process that handles the error which is called a `catch`.
If a `throw` lacks a related `catch`, the result is called an _uncaught error_.
Uncaught errors can also occur when a future completes with an error and no
means exist to wait for that future.

An uncaught error reports to the current zone of the code that failed to catch it.
By default, the program would stop or display the error message in a console.

If you run your code within a custom zone with its own uncaught error handler,
you can intercept uncaught errors and choose how to handle them.
```

src/_articles/archive/zones.md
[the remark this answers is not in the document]
atsansone: **reply:** One important distinction: we can't use code elements as nouns or verbs. See the [GSG](https://developers.google.com/style/code-in-text#grammatical-treatment-of-code-elements) on that point.

I would also note my previous suggestion for some language for this point.

src/_articles/archive/zones.md
```diff
@@ -152,33 +143,47 @@ Code might execute in other nested child zones as well,
...
+Uncaught errors can also arise in `async` functions
+when a Future completes with an error result,
+but is missing a corresponding `await` to handle the error.
+
+An uncaught error always reports to the current zone which 
+failed to catch it. Zones' default behavior is to handle uncaught
+errors by crashing the program.
+Instead of letting your program terminate,
+you can install your own custom _uncaught error handler_ to a new zone
+to intercept and handle uncaught errors however you prefer.
```
MaryaBelanger, line 163: I think these were the contentious paragraphs, so I incorporated your comments @lrhn @atsansone if you could double check hopefully we can wrap this up!

src/_articles/archive/zones.md
```diff
@@ -21,58 +21,49 @@ obsolete: true
...
-* Performing an operation—such as
+* **Performing an operation**—such as
   starting or stopping a timer,
-  or saving a stack trace—each time that
-  code enters or exits a zone.
+  or saving a stack trace—**each time that
+  code enters or exits a zone**.
 
 You might have encountered something similar to zones in other languages.
 _Domains_ in Node.js were an inspiration for Dart’s zones.
```
atsansone, line 69: **question:** Why does this page explain the basics of zones this far down in the page (line 77)? Could and should this not appear further up?

src/_articles/archive/zones.md
atsansone, line 163: **issue:** Tighten and simplify this copy. Use `that` instead of `which`. (`Which` clauses are disposable, `that` clauses are not.)
```suggestion
An uncaught error reports to the current zone that failed to catch it.
By default, a zone crashes an app in a response to uncaught errors.
You can write your own custom _uncaught error handler_ to a new zone.
This _zoned error handler_ can manage uncaught errors in any way you prefer.
```

src/_articles/archive/zones.md
```diff
@@ -152,33 +143,47 @@ Code might execute in other nested child zones as well,
...
+failed to catch it. Zones' default behavior is to handle uncaught
+errors by crashing the program.
+Instead of letting your program terminate,
+you can install your own custom _uncaught error handler_ to a new zone
+to intercept and handle uncaught errors however you prefer.
+
+The simplest way to introduce a new zone with an uncaught error handler
+is to use `runZoneGuarded`. Its `onError` callback becomes the
+uncaught error handler of a new zone, and handles any
+synchronous errors thrown by the call as well.
```
atsansone, line 168: **issue:** This could be simplified.
```suggestion
To introduce a new zone with an uncaught error handler,
use the `runZoneGuarded` method. Its `onError` callback becomes the
uncaught error handler of a new zone. This callback handles any
synchronous errors that the call throws.
```

src/_articles/archive/zones.md
```diff
@@ -193,20 +198,25 @@ they still execute.
 As a consequence a zoned error handler might
 be invoked multiple times.
 
-Also note that an _error zone_—a zone that has an error handler—might
-handle errors that originate in a child (or other descendant)
-of that zone.
+Any zone that includes a zoned error handler is an _error zone_.
+An error zone might handle errors that originate
```
atsansone, line 202: **question:** Can you not use a less complex word? Would `start` or `begin`?

atsansone (changes requested):
@parlough : Could use some tightened language.

src/_articles/archive/zones.md
```diff
@@ -21,58 +21,49 @@ obsolete: true
...
-  **API note:**
-  This use case won't always require zones.
-  We expect that isolates will have an API enabling you to
-  listen for uncaught errors.
-{{site.alert.end}}
+[dart:async]: ({{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-async/dart-async-library.html)
+[`runZoned()`]: ({{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-async/runZoned.html)
+[`runZonedGuarded()`]: ({{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-async/runZonedGuarded.html)
 
 Zones make the following tasks possible:
```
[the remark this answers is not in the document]
MaryaBelanger: Skipping because it reads like an instruction, this article is purely informational. And the tasks listed aren't only possible by using zones, in fact, zones aren't even recommended for some of them.

src/_articles/archive/zones.md
MaryaBelanger: I don't think this is really _that_ far down. It's just at the bottom of the intro section. It's still introduction material, but not more important than the two parts the precede it: first paragraph stating what the article covers and prereqs, and second paragraph/block stating what zones are capable of.

src/_articles/archive/zones.md
```diff
@@ -152,33 +143,47 @@ Code might execute in other nested child zones as well,
...
-to raise an exception that is not handled by a `catch` statement.
-Another way to produce an uncaught error is
-to call `new Future.error()` or
-a Completer's `completeError()` method.
-
-Use the `onError` argument to `runZoned()` to
-install a _zoned error handler_—an asynchronous error handler
-that's invoked for every uncaught error in the zone.
-For example:
+their ability to catch and handle uncaught errors.
```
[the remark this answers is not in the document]
MaryaBelanger: > contradict themselves (..._catch_ and handle _uncaught_ errors.).

Uncaught errors can still be caught, and need to be caught to be handled. They become "uncaught" the first time they bypass a catch or whatever. Zones are able to catch them after that, which is the benefit to using them. This isn't really necessary to dissect.

src/_articles/archive/zones.md
```diff
@@ -152,33 +143,47 @@ Code might execute in other nested child zones as well,
...
+
+_Uncaught errors_ often occur because of code using `throw`
+to raise an exception without an accompanying `catch`
+statement to handle it.
+Uncaught errors can also arise in `async` functions
+when a Future completes with an error result,
+but is missing a corresponding `await` to handle the error.
```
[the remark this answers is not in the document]
MaryaBelanger: Skipping, not necessary

src/_articles/archive/zones.md
MaryaBelanger: Skipping, "originate" captures the nuance, "start" or "begin" implies some error lifecycle which isn't really the case. Also "originate" is not so complex that it's jargon or derailing.

atsansone: OK. Non-blocking, so makes sense.

atsansone (approved):

PR #4406, opened 2022-11-30
[2.19] Remove `.packages` workaround

MaryaBelanger:
`--legacy-packages-file` completely de-supported in 2.19


parlough (approved):
Normally I would say let's keep this around for a while and just mention it was removed, but since it was only around for 1 version as a temporary (and hopefully rare) workaround, I think it's fine to remove it.

atsansone (approved):

PR #4412, opened 2022-12-03
Add Traditional Chinese version of this site

doggy8088:


parlough:
It's really cool to provide this resource, thanks for translating, hosting, and for the pull request!

There are a few things I'd like to see fixed if possible:

The [archive page](https://dart.tw.gh.miniasp.com/get-dart/archive/) and related scripts should likely use the `dart-lang/site-www` versions: [`install.js`](https://github.com/dart-lang/site-www/blob/main/src/get-dart/archive/assets/install.js) and [`download_archive.dart.js`](https://github.com/dart-lang/site-www/blob/main/src/get-dart/archive/assets/download_archive.dart.js) as the dart.cn site uses custom handling due to Google servers not being accessible for many in China.

You'll likely want to update the variables and links in the [`config.yml`](https://github.com/doggy8088/dart.tw/blob/zh-tw/_config.yml) and [`page-footer.html`](https://github.com/doggy8088/dart.tw/blob/zh-tw/src/_includes/page-footer.html) as well, to those relevant for your site or back to the defaults of `dart-lang/site-www`. Some of the Chinese sources may not be relevant, functional, or necessary and this will prevent confusion and slowness for visitors of your site and so issue reporters don't report to the maintainers of dart.cn. Perhaps other links need to be updated as well.

Also, a small question. When traditional Chinese is referenced, I usually see it called 繁體字 but here you called it 正體字. What's the difference?

doggy8088:
When we use 繁體 means Traditional Chinese that opposite to "Simplified Chinese". The "terms" may be or may not correct in some ways. Most of the people are simply translated "Simplified Chinese" into "Traditional Chinese" because they don't know what's difference. More importantly, the "terms" usage some different between China, Taiwan and Hong Kong.

When we use 正體 means more formal and more precisely use the "terms" in Taiwan and some other area.

doggy8088:
I changed `_config.yml` file. It seems no effect. Anyway, I still fixed all the links another way.

atsansone (approved):

parlough (approved):
Looks mostly good to me, thanks so much for working on this! Also thanks for the clarification around 正體. Let us know if you need any changes to make translation or other functionality easier.

Some recommendations if you get a chance to take a look:

- Adjust the remaining dartpad.cn links to dartpad.dev
- Adjust the remaining api.dart.cn links to api.dart.dev
- Adjust pub.flutter.dev to just pub.dev 
- Adjust the GitHub link in `page-footer.html` to your own repository
- Adjust the link to https://policies.google.cn/privacy to https://policies.google.com/privacy?hl=zh-TW instead
- If possible update the [archive page](https://dart.tw.gh.miniasp.com/get-dart/archive/) and related scripts should to use the `dart-lang/site-www` versions: [`install.js`](https://github.com/dart-lang/site-www/blob/main/src/get-dart/archive/assets/install.js) and [`download_archive.dart.js`](https://github.com/dart-lang/site-www/blob/main/src/get-dart/archive/assets/download_archive.dart.js) as the dart.cn site uses custom handling due to Google servers not being accessible for many in mainland China, and it looks like it breaks on your site.

chenglu:
Hi Parker:

The `.tw` site is the auto-translation version of the `.cn` site, some of the broken page may caused by the hotlink protection on some of the services, I will check with Will for the further issue fixes.

Let me know if you have any questions. @parlough 

Thanks!

parlough:
Looks to be working for me as well. Thanks!

PR #4416, opened 2022-12-06
Add null safety warning about Dart 3

mit-mit:
This is similar to what we already added on https://dart.dev/null-safety

### DIFF
--- a/src/null-safety/unsound-null-safety.md
+++ b/src/null-safety/unsound-null-safety.md
@@ -12,11 +12,14 @@ rely on **unsound null safety**.
 Dart 3--planned for a mid-2023 release--
 will require sound null safety. It will prevent code from running without
 null safety, or with unsound null safety.
-All existing code must be [migrated](#migrate) to sound null safety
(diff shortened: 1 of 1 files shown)


src/null-safety/unsound-null-safety.md
```diff
@@ -3,12 +3,23 @@ title: Unsound null safety
...
-A Dart program can contain some libraries that
+A Dart program may contain some libraries that
 are [null safe][] and some that aren't.
 These **mixed-version programs**
-execute with **unsound null safety**.
+rely on **unsound null safety**.
+
+{{site.alert.warn}}
+Dart 3--planned for a mid-2023 release--
+will require sound null safety. It will prevent code from running without
```
parlough, line 13: Use three dashes for an [em dash](https://developers.google.com/style/dashes) here instead:

```suggestion
Dart 3---planned for a mid-2023 release---will require sound null safety. 
It will prevent code from running without
```

parlough (approved):

PR #4423, opened 2022-12-09
Replace ?? guideline

### LINKED ISSUE #4388: Replace "PREFER using ?? to convert null to a boolean value." with more precise guideline
### Page URL

https://dart.dev/guides/language/effective-dart/usage#prefer-using--to-convert-null-to-a-boolean-value

### Page source

_No response_

### Describe the problem

The guideline currently suggests using `?? <bool>` instead of `== <bool>` or `!= <bool>` when dealing with a nullable Boolean. But the `??` operator often doesn't type promote in ways that you want, which is why the guideline also has a large exception that mostly cancels out the guidance.

Also, the examples all use `?.`, which makes it seem like the rule is advocating or at least has something to do with null-aware operators.

Really, the key point of the guideline is that equality operators are all very confusing when you have a Boolean *literal* on one side and a *nullable* Boolean expression on the other. If you see:

```dart
if (foo == true) { ... }
```

It appears to have nothing to do with `null` and that the `== true` is pointless and the code should instead be:

```dart
if (foo) { ... }
```

But that chances the behavior if `foo` happens to be nullable. You could improve this code with:

```dart
if (foo ?? false) { ... }
```

Now the `??` operator makes it clear that something to do with `null` is happening. But it's equally valid (and often better because of type promotion) to do:

```dart
if (foo != null && foo) { ... }
```

This guideline shouldn't be that `??` is good, it's more that `== true` etc. are bad.

### Expected fix

We should clearly target the guideline to the behavior we want to avoid. Something like: "DON'T use `true` or `false` in equality operators."

If the other operand isn't nullable, then it's always simpler to eliminate the equality operator and add `!` if needed. If the other operand is nullable, you should use `??` or an explicit null check to make that clear.

### Additional context

_No response_

MaryaBelanger:
Fixes #4388 

I'll add the code excerpts properly once the content is confirmed. 

@munificent have a few questions, first draft is _very_ rough

First of all, there's [a lint](https://github.com/dart-lang/linter/blob/main/lib/src/rules/use_if_null_to_convert_nulls_to_bools.dart) that points to this section with a name that's very specific to the whole "conversion" aspect, which I feel like the rewrite doesn't 100% correspond to. I redirected the link to the old section to this section (@parlough not sure if I did that right) in the mean time, but the example in the lint is basically saying what the new section is recommending about explicit null checks is "bad" so we'll have to do something about that.

The idea of converting null to a bool is kind of lost, I squeezed it in towards the end but not sure if we want to keep any remnant of that idea or toss it completely. 

Let me know what you think!


src/_guides/language/effective-dart/usage.md
```diff
@@ -245,87 +245,65 @@ void error([String? message = null]) {
...
+if (nullableBool == true) { ... }
 {% endprettify %}
 
+`if (nullableBool == true)` is a viable expression, but shouldn't be used for several reasons:
+
+* It doesn't indicate the code has anything to do with `null`.
+
+* Because it's not evidently `null` related, it can easily be mistaken for the non-nullable case,
+where the equality operator is redundant and can be removed.
+That’s only true when the boolean expression on the left has no chance of producing null, but not when it can.
```
atsansone, line 292: **suggestion:** This can go as is, but it could be tighter.

Consider:
* Removing the adverbs. You could say `...it's not clear that it relates to `null` (values? expressions?)...` and `...it can be mistaken...`
* Cutting the sentence into two: `You can remove the equality operator as it is redundant.`
* Simplify the language in the second sentence: `It's true when the left-side boolean expression can't produce a null value. It's false when it can't.`

src/_guides/language/effective-dart/usage.md
```diff
@@ -245,87 +245,65 @@ void error([String? message = null]) {
...
-### PREFER using `??` to convert `null` to a boolean value.
-
-This rule applies when an expression can evaluate to `true`, `false`, or `null`,
-and you need to pass the result to something that expects a non-nullable boolean
-value. A common case is using the result of a null-aware method call as a
-condition. You can "convert" `null` to `true` or `false` using `==`, but we
-recommend using `??`:
+Using the equality operator to evaluate a *non-nullable* Boolean expression 
+against a Boolean literal is redundant. 
+It's always simpler to eliminate the equality operator, and add `!` if necessary:
```
munificent, line 252: This paragraph is great.

src/_guides/language/effective-dart/usage.md
```diff
@@ -245,87 +245,65 @@ void error([String? message = null]) {
...
+That’s only true when the boolean expression on the left has no chance of producing null, but not when it can.
+
+* The boolean logic is confusing. If `nullableBool` is null, then `nullableBool == true`
+means the condition evaluates as false.
+
+The `??` operator makes it clear that something to do with null is happening,
+so it won't be mistaken for a redundant operation. The logic is much clearer too; 
+the result of the expression being null is the same as the Boolean literal.
+This is commonly considered "converting" `null` to `true` or `false` to be able to
+pass the result of the evaluation to something that expects a non-nullable Boolean value.
```
munificent, line 301: I'd probably just remove the sentence on "converting". I don't think it adds much value. That was how I framed it originally, but I don't know if that's how anyone actually thinks about it.

munificent (approved):
This is so much better!

src/_guides/language/effective-dart/usage.md
MaryaBelanger: @munificent The reason I kept that point from the original section is because of the lint `use_if_null_to_convert_nulls_to_bools` that currently points to the section I rewrote/removed. 

I redirected that lint to point to this new section, but the new content doesn't mention anything about "converting" so it's...odd. 

I will raise this with the analyzer team (and @ you there) to change the lint if no one thinks about  it like "converting" anyway.

src/_guides/language/effective-dart/usage.md
```diff
@@ -245,87 +245,66 @@ void error([String? message = null]) {
...
+// Error if null 
+if (nullableBool) { ... }
 
-  return 0;
-}
+// If you want null to be false:
+if (nullableBool == true) { ... }
 {% endprettify %}
 
+`if (nullableBool == true)` is a viable expression, but shouldn't be used for several reasons:
```
parlough, line 287: I believe just the `nullableBool = true` portion of this is the expression. Maybe just stick to that as it should be avoided in other conditional situations as well (`nullableBool = true ? oneThing : SecondThing`):

```suggestion
`nullableBool == true` is a viable expression, 
but shouldn't be used for several reasons:
```

src/_guides/language/effective-dart/usage.md
```diff
@@ -245,87 +245,66 @@ void error([String? message = null]) {
...
+means the condition evaluates as false.
+
+The `??` operator makes it clear that something to do with null is happening,
+so it won't be mistaken for a redundant operation. The logic is much clearer too; 
+the result of the expression being null is the same as the Boolean literal.
+This is commonly considered "converting" `null` to `true` or `false` to be able to
+pass the result of the evaluation to something that expects a non-nullable Boolean value.
+
+Using a null-aware operator on a variable inside a condition
+doesn’t promote the variable to a non-nullable type.
```
parlough, line 305: Maybe conditional statement?

```suggestion
Using a null-aware operator on a variable inside a conditional statement
doesn’t promote the variable to a non-nullable type. 
```

src/_guides/language/effective-dart/usage.md
```diff
@@ -245,87 +245,66 @@ void error([String? message = null]) {
...
-### PREFER using `??` to convert `null` to a boolean value.
-
-This rule applies when an expression can evaluate to `true`, `false`, or `null`,
-and you need to pass the result to something that expects a non-nullable boolean
-value. A common case is using the result of a null-aware method call as a
-condition. You can "convert" `null` to `true` or `false` using `==`, but we
-recommend using `??`:
+Using the equality operator to evaluate a *non-nullable* Boolean expression 
+against a Boolean literal is redundant. 
+It's always simpler to eliminate the equality operator, and add `!` if necessary:
```
parlough, line 253: This mention of adding "`!` if necessary" is a bit confusing to me, because it makes it seem like it will be part of the solution, but the guideline goes on to say "you should only use `??` or an explicit `!= null` check." and doesn't mention `!`. 

Depending on the goal of this mention, perhaps just remove the mention of `!` as I think the other two solutions are generally a better choice anyway.

parlough (commented):
Cool! I like the ideas outlined here. I have a few comments and questions though.

src/_guides/language/effective-dart/usage.md
MaryaBelanger: That's meant to capture the case of negating the expression, depending on what you're trying to accomplish. These are both "good" for non-nullable boolean expressions:
```
if (nonNullableBool) { ... }

if (!nonNullableBool) { ... }
```

So maybe removing ` and add `!` if necessary` and just adding the second "good" example is clearer. Or even keeping the clause and adding the second example, that'd probably be clearest

src/_guides/language/effective-dart/usage.md
MaryaBelanger: Ok based on https://github.com/dart-lang/linter/issues/3908 it seems safe to remove that part, thanks!

src/_guides/language/effective-dart/usage.md
parlough: Oh my, sorry! I was so focused on nullability, I thought this was talking about the null assertion operator and was pretty confused.

I think adding a second example would be great, but I'd love to also modify the clause so people don't read the text incorrectly as being about the null assertion operator. 

```suggestion
It's always simpler to eliminate the equality operator, 
and use the unary negation operator (`!`) if necessary:
```

src/_guides/language/effective-dart/usage.md
MaryaBelanger: Skipping because:
> You could say ...it's not clear that it relates to null (values? expressions?)... and ...it can be mistaken...

"`null` related" is sufficient, really just related to null in any way is what's important, nothing specific about values/expressions/etc. 

"_easily_ mistaken", not just mistaken, is the issue here and why we're going out of the way to advise against using it even though it technically works. 

> Cutting the sentence into two: You can remove the equality operator as it is redundant.

The sentence is saying "it can be mistaken for the non-nullable case, _i.e. the case where the equality operator is redundant_". It's defining the clause before it, so cutting the sentence like that wouldn't make sense. 

> It's true when the left-side boolean expression can't produce a null value. It's false when it can't.

This reads much more awkward to me. Starting with "That's" indicates I'm talking about the condition I just described, while "It's" makes it sound like we're talking about the expression in question since it's referred to as "it" in every sentence prior. "It's false when it can't" is super confusing.

MaryaBelanger:
@parlough So I was going to make the good/bad examples actual code excerpts, but.... is that necessary? In this case specifically all the `if` bodies are just `{...}`, which I'm imagining wouldn't work in excerpts, or would just take a lot of `ignore` methods. 

What do you think? I leaning towards skip it. I can't imagine the syntax changing so much that basic stuff like equality operators and `if` statements need to be checked regularly for accuracy.

parlough:
I think it's not a huge deal if there are no excerpts, but they can be nice for keeping up with formatting changes, new lints, etc. Perhaps most importantly in this case, it makes sure it still compiles if anyone makes any new changes or additions to these snippets in the future.

In the `.dart` file you should be able to use `{/* ... */}` and the refresh snippet tool should remove the block comments to get output like you want.

MaryaBelanger:
I'm going to skip incorporating real code snippets for such simple blocks. @atsansone this is ready to merge, needs an approval.

src/_guides/language/effective-dart/usage.md
```diff
@@ -245,87 +245,68 @@ void error([String? message = null]) {
...
+* Because it's not evidently `null` related, it can easily be mistaken for the non-nullable case,
+where the equality operator is redundant and can be removed.
+That’s only true when the boolean expression on the left has no chance of producing null, but not when it can.
+
+* The boolean logic is confusing. If `nullableBool` is null, then `nullableBool == true`
+means the condition evaluates as false.
+
+The `??` operator makes it clear that something to do with null is happening,
+so it won't be mistaken for a redundant operation. The logic is much clearer too; 
+the result of the expression being null is the same as the Boolean literal.
```
parlough, line 304: Split up a bit and make boolean lowercase:

```suggestion
The `??` operator makes it clear that something to do with null is happening,
so it won't be mistaken for a redundant operation. 
The logic is much clearer too; 
the result of the expression being null is the same as the boolean literal.
```

parlough (approved):
Thanks for those other adjustments. A few final minor fixes:

parlough:
I've created a PR targeting this branch with code excerpts (and removing the old ones replaced by this) as I still think they have value: https://github.com/dart-lang/site-www/pull/4458

PR #4428, opened 2022-12-11
Fix broken API doc link and add more ignores to linkcheck

parlough:
- Fixes the broken link to `bool.hasEnvironment`
- Adds some ignores to the link checker tool we use which cause issues when checking external links:
  - GitHub since it rate limits requests
  - /go/ links to match flutter/website since these are redirects to external sites, not part of the site
  - /jobs for a similar reason - the linkcheck tool seems to think it is part of the site
  
In CI we currently don't check external links, so these issues don't pop up, but when I run external links locally the problems arise.


atsansone (approved):
LGTM!

PR #4429, opened 2022-12-12
Typographic correction

zanovis:
include -> including


domesticmouse (approved):

src/_guides/language/coming-from/js-to-dart.md
```diff
@@ -92,7 +92,7 @@ expression. This means that Dart apps have sound static typing.
 JavaScript provides primitive types `num`, `string`, and `boolean`
 and the `null` value as well as _arrays_ and a `Map` type.
 
-Dart supports built-in types include:
+Dart supports built-in types, including:
```
parlough, line 95: Nice catch! It seems a bit weird for me to say Dart supports built-in types though. What do you think about the following?

```suggestion
Dart's supported built-in types include:
```

zanovis: Works for me :)

PR #4430, opened 2022-12-12
Use colored language tags in Swift to Dart article

parlough:
Instead of including the language tags in bold text, it's now part of the code snippet:

Example of language tags


sfshaza2 (approved):
lgtm. I didn't know about this!

PR #4445, opened 2022-12-14
Update some old version mentions

parlough:
- Remove mention of the `.analysis_options` file which has been removed for over 4 years.
- Update mention of standalone `pub` tool to reference removal

### DIFF
--- a/src/tools/pub/cmd/index.md
+++ b/src/tools/pub/cmd/index.md
@@ -30,7 +30,7 @@ on the [Flutter website]({{site.flutter}}).
   Although you might still find examples of
   using the standalone `pub` command instead of
   `dart pub` or `flutter pub`,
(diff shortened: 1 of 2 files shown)


domesticmouse (approved):

src/tools/pub/cmd/index.md
```diff
@@ -30,7 +30,7 @@ on the [Flutter website]({{site.flutter}}).
   Although you might still find examples of
   using the standalone `pub` command instead of
   `dart pub` or `flutter pub`,
-  the standalone `pub` command is deprecated.
+  the standalone `pub` command has been removed.
```
atsansone, line 33: **comment:** This runs a bit long, but it helps as is. We should return to tighten up the language here.

atsansone (approved):
LGTM % 1 comment.

PR #4442, opened 2022-12-14
Fixes bad subscript error on macOS

atsansone:

PR #4446, opened 2022-12-15
Enable fetching checksums while not in dev container

parlough:
Allows running `make fetch-sums` even if not in the dev container and without needing to specify anything else.

- Allow defaulting to `/tool` directory if `TOOL_DIR` is not specified
- Support older versions of bash which do not support negative indices into arrays
- Keep using colors from `utils.sh` so we don't replicate them
- Fix the format of `fetch-node-ppa-sum.sh` to match what is used in Dockerfile

### DIFF
--- a/tool/fetch-dart-sdk-sums.sh
+++ b/tool/fetch-dart-sdk-sums.sh
@@ -3,6 +3,7 @@
 # Prints output similar to cases in Dockerfile for easy composition 
 # when having to update checksum values for updates dart SDK.
 set -eu -o pipefail
+TOOL_DIR="${TOOL_DIR:=$(dirname "$0")}"
 source $TOOL_DIR/utils.sh
 
 VERSION="latest"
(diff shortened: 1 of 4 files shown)


tool/fetch-dart-sdk-sums.sh
```diff
@@ -28,12 +29,15 @@ done
 BASEURL="https://storage.googleapis.com/dart-archive/channels"
 CHANNELS="stable beta dev"
 ARCHS="amd64 arm64"
+ENDING='\\\n'
+
+printf "\n$(blue "Copy the following output and replace the existing code in the Dockerfile")\n"
+printf "$(blue "inside the 'set -eu' run statement:")\n\n"
```
atsansone, line 35: **issue:** These colors don't work on the macOS. All they return is the color codes, which do nothing here. Can we find another solution?

atsansone (changes requested):
@parlough : The color commands return the color code text and do not set text into the given color. Can you verify that it works on your Mac? If so, can you also see if the changes from #4442 work on your Mac? If the latter, then let's use those going forward.

tool/fetch-dart-sdk-sums.sh
atsansone: Also, see the comments in [this article](https://dev.to/ifenna__/adding-colors-to-bash-scripts-48g4).

parlough: I wanted to avoid specifying the colors again as then we would be repeating functionality across the scripts and would have to remember to include `ENDCOLOR` on each one. 

The color commands were being used incorrectly in the script before, but I thought these changes fixed that. This PR works for me in the Docker container, macOS terminal with ZSH and bash shell, as well as in IntelliJ's terminal emulator.

Terminal output

Could you show me the output you get with these new changes?

atsansone (approved):
@parlough : Works great! Thanks!

PR #4454, opened 2022-12-18
Incorrect lint link

MaryaBelanger:
[PREFER using a final field to make a read-only property](https://dart.dev/guides/language/effective-dart/usage#prefer-using-a-final-field-to-make-a-read-only-property) points to [`unnecessary_getters_setters`](https://dart.dev/tools/linter-rules#unnecessary_getters_setters), but the actual corresponding lint is [`unnecessary_getters`](https://github.com/dart-lang/linter/blob/main/lib/src/rules/unnecessary_getters.dart).

However, `unnecessary_getters` doesn't seem to have an entry on the [Linter rules](https://dart.dev/tools/linter-rules) page? This will stay a draft until there's something to link to.

### DIFF
--- a/src/_guides/language/effective-dart/usage.md
+++ b/src/_guides/language/effective-dart/usage.md
@@ -1115,7 +1115,7 @@ class Box {
 
 ### PREFER using a `final` field to make a read-only property.
 
-{% include linter-rule-mention.md rule="unnecessary_getters_setters" %}
+{% include linter-rule-mention.md rule="unnecessary_getters" %}
 
 If you have a field that outside code should be able to see but not assign to, a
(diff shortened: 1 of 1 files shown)


parlough:
The lint likely won’t be fixed soon, so perhaps just best to remove any link for now.

MaryaBelanger:
Thanks @parlough, I guess just no lint reference to be safe then? Lmk if you think keeping it the way it was is better than nothing.

atsansone:
@MaryaBelanger : Sounds like removing it would be the best move. Best not include information that is incongruent with user experience.

atsansone (approved):

PR #4458, opened 2022-12-20
Add code excerpts to new boolean equality guideline

parlough:
This PR is targeting https://github.com/dart-lang/site-www/pull/4423, both adding code excerpts and removing the ones replaced by the new guideline. This also adds the bad example of comparing to `false`.

I'd like to add code excerpts as while they are simple snippets, it guarantees any future updates or additions to the snippets still are correct. It also helps guarantee formatting is correct.


MaryaBelanger (approved):
**Thank you** it would've taken me all day to do that, really appreciate it

parlough:
Always happy to work on excerpts and tests :)

PR #4463, opened 2022-12-22
Fix the example code in the Exceptions section on the Language samples page

hibix43:
I found the wrong in the Exceptions section on the Language samples page. 
So, I created this pull request.
If this pull request isn’t correct, please reject this.

I only write simple English, sorry.

### DIFF
--- a/src/samples/index.md
+++ b/src/samples/index.md
@@ -447,7 +447,7 @@ To catch an exception, use a `try` statement with `on` or `catch` (or both):
 <?code-excerpt "misc/test/samples_test.dart (try)"?>
 ```
 try {
-  for (final object in flybyObjects) {
+  for (final object in flybyObjects) async {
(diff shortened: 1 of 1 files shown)


parlough:
Thanks so much for the pull request, but the `async` keyword can only be added to functions, not statements like for loops.

We explain this code is `async` in a note right after the snippet, but do you think it would be helpful if we explicitly include a method declaration in the code? I've proposed a potential solution in https://github.com/dart-lang/site-www/pull/4464, what do you think of that?

hibix43:
Thank your comment. I saw the draft. I need that draft to understand it is the async function. I think the draft should be merged.

parlough:
Sounds good, thanks for taking a look and for the pull request @hibix43 :)

I'll close this PR in favor of https://github.com/dart-lang/site-www/pull/4464

PR #4468, opened 2022-12-28
[2.19] Adjust Flutter compute function mention

parlough:
Move the mention of Flutter's `compute` function to after introducing `Isolate.run` and adjust the guidance around when to use it in terms of `Isolate.run`.

### DIFF
--- a/src/_guides/language/concurrency/index.md
+++ b/src/_guides/language/concurrency/index.md
@@ -298,16 +298,6 @@ This section discusses some examples
 that use the `Isolate` API
 to implement isolates.
 
-{{site.alert.flutter-note}}
-  If you're using Flutter on a non-web platform,
-  then instead of using the `Isolate` API directly,
-  consider using the [Flutter `compute()` function][].
-  The `compute()` function is a simple way to
-  move a single function call to a worker isolate.
-{{site.alert.end}}
-
- [Flutter `compute()` function]: {{site.flutter-docs}}/cookbook/networking/background-parsing#4-move-this-work-to-a-separate-isolate
-
 ### Implementing a simple worker isolate
 
 These examples implement a main isolate
@@ -323,6 +313,19 @@ setting up and managing worker isolates:
 6. Checks, captures, and throws exceptions and errors back to the main isolate
 
 [`Isolate.run()`]: {{site.dart-api}}/dev/dart-isolate/Isolate/run.html
+
+{{site.alert.flutter-note}}
+  If you're using Flutter,
+  then instead of using `Isolate.run()`,
+  consider using the [Flutter `compute()` function][].
+  On native platforms, 
+  the `compute()` function is implemented using `Isolate.run()`,
+  while on the web platform, 
+  it falls back to running the passed-in callback
+  on the current event loop.
+{{site.alert.end}}
+
+[Flutter `compute()` function]: {{site.flutter-docs}}/cookbook/networking/background-parsing#4-move-this-work-to-a-separate-isolate
 
 #### Running an existing method in a new isolate
 
--- a/src/_guides/language/concurrency/index.md
+++ b/src/_guides/language/concurrency/index.md
@@ -316,14 +316,14 @@ setting up and managing worker isolates:
 
 {{site.alert.flutter-note}}
   If you're using Flutter,
-  then instead of using `Isolate.run()`,
-  consider using the [Flutter `compute()` function][].
-  On native platforms, 
-  the `compute()` function is implemented using `Isolate.run()`,
-  while on the web platform, 
-  it falls back to running the passed-in callback
-  on the current event loop.
+  consider using the [Flutter `compute()` function][]
+  instead of `Isolate.run()` to allow your code to work
+  on both [native and non-native platforms][].
+  `Isolate.run()` is the better API choice
+  for Flutter on native platforms only. 
 {{site.alert.end}}
+
+[native and non-native platforms]: overview#platform
 
 [Flutter `compute()` function]: {{site.flutter-api}}/flutter/foundation/compute.html
 
--- a/src/_guides/language/concurrency/index.md
+++ b/src/_guides/language/concurrency/index.md
@@ -317,10 +317,12 @@ setting up and managing worker isolates:
 {{site.alert.flutter-note}}
   If you're using Flutter,
   consider using [Flutter's `compute()` function][]
-  instead of `Isolate.run()` to allow your code to work
-  on both [native and non-native platforms][].
-  `Isolate.run()` is the more ergonomic API choice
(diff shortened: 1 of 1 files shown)


domesticmouse (approved):
I like the way this has been worded

src/_guides/language/concurrency/index.md
```diff
@@ -324,6 +314,19 @@ setting up and managing worker isolates:
...
+{{site.alert.flutter-note}}
+  If you're using Flutter,
+  then instead of using `Isolate.run()`,
+  consider using the [Flutter `compute()` function][].
+  On native platforms, 
+  the `compute()` function is implemented using `Isolate.run()`,
+  while on the web platform, 
+  it falls back to running the passed-in callback
+  on the current event loop.
+{{site.alert.end}}
```
MaryaBelanger, line 326: **Yes** to moving it to the section that mentions `run` for the first time.

I think getting into the details of what Flutter does on the web (lines 321+) in this case is out of scope for the dart docs. 

More of a hot take (so, arguable), is that I think even mentioning what `compute` does or how it works, like the original note did, is out of scope too. I think providing a link to the Flutter docs and saying it's Flutter's `run` is sufficient, the Flutter docs can explain the nuance and context. 

```suggestion
{{site.alert.flutter-note}}
  The [Flutter `compute()` function][]
  is implemented using `Isolate.run()`,
  and should be used when developing with 
  Flutter for non-web platforms.
{{site.alert.end}}
```

MaryaBelanger: What I don't know is if the whole "should be used when developing with Flutter for non-web platforms" part of my suggestion is a technically accurate statement, so please tweak that part as you see fit

parlough: I'm happy to have as little detail as possible. It looks like technically the `compute` function supports the web platform but does not move that work off the event loop, so users have to be careful about using it as it's not always equivalent to `Isolate.run`.

MaryaBelanger: Hmm... yeah that is a little more detail. 

So if you use `compute` for Flutter on the web it just... doesn't execute concurrently? Is it like a single function, but overridden for native and web? Like, if you call it on native platforms, it uses a `compute` that's implemented with `Isolate.run`? But if you call it on web it uses a `compute` that _isn't_ implemented with `run`? (I didn't see any mention of the web/native functionality difference in the cookbook or the API docs for compute, could you point me to where you learned that from?).

parlough: Yeah it's a single function which [delegates](https://github.com/flutter/flutter/blob/9bb4ffe152/packages/flutter/lib/src/foundation/isolates.dart#L7-L8) to two (or more) implementations depending on the platform. Since you can't access `dart:isolate` and it can't execute concurrently on the web, it simply runs on the current event loop, I'm assuming similarly to any `async` function. 

The API docs were updated recently with the distinction, you can find the `master` branch docs for `compute` at https://master-api.flutter.dev/flutter/foundation/compute.html

MaryaBelanger: If that's all the case, then I'd keep all the context of the original, maybe making it a little clearer (but not sure if this is really that much of an improvement over your original):
```suggestion
{{site.alert.flutter-note}}
  Consider using the [Flutter `compute()` function][]
  instead of `Isolate.run()`
  if you're using Flutter.
  On native platforms, 
  the `compute()` function is implemented with `Isolate.run()`
  and functions the same.
  `compute()` functions differently on web platforms, though,
  so be careful equating it to `Isolate.run()`
{{site.alert.end}}
```
I still think it's something that should be specified in the `compute` docs instead, but since it's not we can say it here.

Also, should it link straight to the [API docs for compute](https://api.flutter.dev/flutter/foundation/compute-constant.html) instead of that cookbook? Might be a better general resource

parlough: Perhaps now that the new `compute` API docs cover the difference we can just keep it simple and link to the API docs instead (already done in newest commit)?

```suggestion
{{site.alert.flutter-note}}
  If you're using Flutter,
  consider using the [Flutter `compute()` function][]
  instead of `Isolate.run()`.
{{site.alert.end}}
```

Right now the link will point to outdated docs, but with the release of 2.19 it will point to the updated docs with the extra information.

mkustermann: IMHO it would be good to describe **why** one should do that. If one doesn't care about web, using `Isolate.run()` is actually a more ergonomic/better API.

Consider making this

> If you're using Flutter consider using Flutter's `compute()` function, doing so will allow your code to also work on non-native platforms (such as web). See more information <link>

mkustermann (approved):

src/_guides/language/concurrency/index.md
MaryaBelanger: Ok so...
- `Isolate.run()` is better than `compute` for native platforms, _even if you're using flutter_
- For flutter on non-native (web) platforms, `compute` is the best option

So the guidance here is use `compute` for Flutter web _and_ native projects. Otherwise `run` is better.

Suggestion just for wording/writing:
```suggestion
{{site.alert.flutter-note}}
  If you're using Flutter,
  consider using the [Flutter `compute()` function][]
  instead of `Isolate.run()` to allow your code to work
  on both [native and non-native platforms][].
  `Isolate.run()` is the better API choice
  for Flutter on native platforms only. 
{{site.alert.end}}

[native and non-native platforms]: overview#platform
```

I don't love the wording in my suggestion, so @parlough I approved and you can tweak however you like, but you get the idea. Thanks for adding this!

MaryaBelanger (approved):

src/_guides/language/concurrency/index.md
```diff
@@ -324,6 +314,18 @@ setting up and managing worker isolates:
 
 [`Isolate.run()`]: {{site.dart-api}}/dev/dart-isolate/Isolate/run.html
 
+{{site.alert.flutter-note}}
+  If you're using Flutter,
+  consider using [Flutter's `compute()` function][]
+  instead of `Isolate.run()` to allow your code to work
+  on both [native and non-native platforms][].
+  `Isolate.run()` is the more ergonomic API choice
+  for Flutter apps targeting native platforms only.
```
atsansone, line 323: **suggestion:** This is fine as is, but I would suggest making this more concise.
```suggestion
  If you're using Flutter,
  consider using [Flutter's `compute()` function][]
  instead of `Isolate.run()`. The `compute` function
  allows your code to work  on both
  [native and non-native platforms][].
  Use `Isolate.run()` when targeting native platforms only.
  This Isolate function serves as a more ergonomic
  API choice for native Flutter apps. 
```

atsansone (approved):
LGTM % one suggestion

PR #4469, opened 2022-12-28
Switch to opt-out system for code block language tags

parlough:
This works by adding an extra class to the body on pages in which language tags should be included. Code blocks can add the `no-lang` class with `{:no-lang}` to avoid the language tag being added.

Classes which want the language tags to show up need to add `body_class: highlight-languages` to the front matter.

### DIFF
--- a/src/_guides/language/coming-from/js-to-dart.md
+++ b/src/_guides/language/coming-from/js-to-dart.md
@@ -1,6 +1,7 @@
 ---
 title: Learning Dart as a JavaScript developer
 description: Leverage your JavaScript knowledge when learning Dart.
+body_class: highlight-languages
 ---
 
 This guide aims to leverage your JavaScript programming knowledge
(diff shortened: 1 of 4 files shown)


atsansone (approved):
LGTM. I see where it applies on the **Coming from...** pages, but don't see it on [other](https://dart-dev--pr4469-feature-opt-out-syst-xilb6i87.web.app/guides/language/language-tour#a-basic-dart-program) [pages](https://dart-dev--pr4469-feature-opt-out-syst-xilb6i87.web.app/guides/language/concurrency#the-async-await-syntax). Do we need to make further changes on those pages to get this benefit?

domesticmouse (approved):

parlough:
@atsansone The pages have to opt in by adding `body_class: highlight-languages` to the front matter since it doesn't make sense for most pages to include these tags when they are only documenting Dart.

PR #4490, opened 2023-01-06
Change the data type in the text to match the example

DemX86:
Problem: There is a `list` data type in the description text but the following example has the type of a `set`

### DIFF
--- a/src/_guides/language/coming-from/js-to-dart.md
+++ b/src/_guides/language/coming-from/js-to-dart.md
@@ -1585,7 +1585,7 @@ Pure JavaScript doesn't support immutability.
 Dart offers multiple ways to make collections like arrays, sets, or
 dictionaries immutable.
 
-* If the list is a compile-time constant and shouldn't
+* If the set is a compile-time constant and shouldn't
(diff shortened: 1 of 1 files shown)


domesticmouse (approved):

parlough (approved):
Nice catch, looks good to me! Thanks :)

src/_guides/language/coming-from/js-to-dart.md
```diff
@@ -1585,7 +1585,7 @@ Pure JavaScript doesn't support immutability.
 Dart offers multiple ways to make collections like arrays, sets, or
 dictionaries immutable.
 
-* If the list is a compile-time constant and shouldn't
+* If the set is a compile-time constant and shouldn't
```
parlough, line 1588: Actually this section might be better served with a more generic word like collection:

```suggestion
* If the collection is a compile-time constant and shouldn't
```

DemX86: Agree, that will be better — thanks for the correction!

parlough:
Looks good to me :)

PR #4494, opened 2023-01-06
Scroll active sidenav entry into view on load

parlough:
Right now the sidenav always resets to scrolling to the top when loading the page, interrupting the user flow, especially when the active entry is below their view port. 

This PR adds JS code to automatically scroll the sidenav on load so that deepest active entry is in view.

Test by reducing browser height and opening a link further down the sidebar on the staged site.

Similar to the flutter/website change at https://github.com/flutter/website/pull/8050

### DIFF
--- a/src/assets/js/main.js
+++ b/src/assets/js/main.js
@@ -226,7 +226,6 @@ $(function() {
   }
 
   $(window).smartresize(fixNav);
-
   scrollSidebarIntoView();
 
   // Add external link indicators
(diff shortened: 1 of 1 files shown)


domesticmouse (approved):

src/assets/js/main.js
```diff
@@ -211,6 +227,8 @@ $(function() {
 
   $(window).smartresize(fixNav);
```
domesticmouse, line 229: nit: drop this newline?

atsansone (approved):
Works consistently. Great work! LGTM.

Issue #4500, opened 2023-01-11
~/.pub-cache/README.md point to https://dart.dev/go/pub-cache - which 404s!


kevmoo:
### Page URL

https://dart.dev/go/pub-cache

### Describe the problem

The generated `~/.pub-cache/README.md` files points to https://dart.dev/go/pub-cache – which doesn't exist!

### Expected fix

Redirects to "the right page"

### Additional context

CC @mit-mit

PR #4501, opened 2023-01-11
Add go link for pub cache

mit-mit:
This is link to from `~/.pub-cache/README.md

Fixes https://github.com/dart-lang/site-www/issues/4500


kevmoo:
The source really doesn't talk about the `pub cache` command, right? It's more about what the cache directory IS – right?

kevmoo:
Hrm...okay maybe https://dart.dev/tools/pub/cmd/pub-cache

kevmoo (approved):

parlough (approved):
Thanks for fixing this! Looks good to me since the README talks about "how to manage" the cache.

jonasfj:
It could also be: https://dart.dev/tools/pub/glossary#system-cache

But https://dart.dev/tools/pub/cmd/pub-cache is probably what you need for modifying..

Issue #4297, opened 2022-10-20
Don't break build when a new Node or Dart release is made


parlough:
### Describe the problem

It can impact work, breaking deploys and staging on PRs until someone updates the checksums, which not everyone may have the local setup to do so easily. Even once it's updated, PRs need to be updated by rebasing or merging to pass the build.

### Expected fix

I understand the checksum checks are important, but could we just pin and check a specific version and then update those pinned versions as desired?

MaryaBelanger:
> could we just pin and check a specific version and then update those pinned versions as desired?

I don't have anything useful to add to whether it'd be ok to do this. Just emphatically voting to not have have the build break so often! Thanks Parker.

PR #4506, opened 2023-01-12
False positive security vulnerabilities

Hixie:
We occasionally get reports of security vulnerabilities. This document attempts to explain them so that we can just point people to this document instead of having to explain it each time.


Hixie:
I can't figure out why `build / stage (pull_request)` is failing.

parlough:
Sorry about that. I’ve submitted a fix in https://github.com/dart-lang/site-www/pull/4507 and am tracking not breaking the build on a release in https://github.com/dart-lang/site-www/issues/4297.

src/resources/security-false-positives.md
```diff
@@ -0,0 +1,157 @@
+---
+title: Security False Positives
```
mit-mit, line 2: nit: please use sentence casing

mit-mit (commented):
A high-level question before I review the details: Should this perhaps be on flutter.dev, not dart.dev? Yes, there are quite a few Dart details, but most of the content is about the Dart embedding in Flutter, not Dart itself. For example, Dart per se doesn't have a `libapp.so`, that's a Flutter thing.

mit-mit:
cc @mraleph

mraleph:
Yeah, this probably should be on `flutter.dev`. I did review the original doc so I am not sure I have anything else to say :)

Hixie:
I'll let @sfshaza2 and y'all debate which site it should be on. I have no strong opinion. Let me know when you come to a conclusion. :-)

sfshaza2:
I thought it should go on dart.dev because it was primarily dart-focused. But if these errors occur mostly with Flutter apps, then, yes, let's do that. Sorry, Ian!  We can put it under src/testing

sfshaza2 (commented):
Closing this PR because I'm putting it on the docs.flutter.dev site.

PR #4516, opened 2023-01-15
Switch to tag/hash based anchor icon and fix spacing

parlough:
Switches the anchor icon to use a tag/hash symbol similar to other sites (and matching flutter/website as of https://github.com/flutter/website/pull/7890). Also increases content padding to allow the page to fit the anchor icon and prevent crowding. We still are slightly more dense than other sites even with this change, but this seemed like a good in-between. While doing this, removes a few unused styles as well.

**After:**
Anchor icon after change

**Before:** _(note the icon overlapping the sidenav)_
Anchor icon before change


domesticmouse (approved):

atsansone (approved):
LGTM. I wonder if the icon is the best choice, but let's get this in.

parlough:
Thanks! I'm open to changes to the icon. 

Went with the hash/tag as other sites like web.dev, MDN, and Chrome developers site all use it.

PR #4527, opened 2023-01-18
Say cascade operator is left associative in language tour

whesse:
The update adding associativity to the operator table in the language tour marked it as right associative.


parlough (approved):
Nice catch, thanks!

PR #4528, opened 2023-01-18
[2.19] Add next-generation preview section to js-interop page

### LINKED ISSUE #4450: [2.19] Document future-proof JS-interop

JS interop is currently evolving, partially to prepare for `dart2wasm`. It looks like there is a subset and suggested set of functionality to prepare for the future state of JS-interop and support all three web compilers.

While we don't really have any JS-interop docs, we should consider documenting what is "future-proof" to best prepare for the future.

Context: https://github.com/dart-lang/sdk/issues/50721 has some more context and a list of things that currently work.

Related: https://github.com/dart-lang/site-www/issues/4331

MaryaBelanger:
Staged: https://dart-dev--pr4528-future-js-interop-zfqketfc.web.app/web/js-interop

Fixes #4450

### DIFF
--- a/src/web/js-interop.md
+++ b/src/web/js-interop.md
@@ -68,7 +68,7 @@ expected to work across compilers for JS interop.
 
 The key feature of next-generation JS interop is [static interop][].
 We recommend using static interop as the default for `package:js`,
-as it is more declaritive, more likely to be optimized,
+as it is more declarative, more likely to be optimized,
 more likely to perform better, and required for `dart2wasm`.
 Static interop addresses several gaps in the existing JS interop story:
 
@@ -83,7 +83,7 @@ as before.
 * **Clarity:** Static interop takes a step towards making JS interop
 more idiomatic in Dart, and making the boundary between the two languages more visible.
 For example, it enforces that JS classes are not meant to be mixed with Dart
-(e.g. no dynamic calls, JS interop types cannot be implemented as a Dart class).
+(dynamic calls aren't allowed, and JS interop types cannot be implemented as a Dart class).
 We expect upcoming integration with inline classes
 will help make interop even more idiomatic.
 
--- a/src/web/js-interop.md
+++ b/src/web/js-interop.md
@@ -103,7 +103,7 @@ The set of features for future static interop currently includes:
 To learn how to implement static interop and see examples,
 visit the [static interop][] specification.
 
-[`@staticInterop`]: https://pub.dev/documentation/js/latest/js/staticInterop-constant.html
+[`@staticInterop`]: {{site.pub-api}}/js/latest/js/staticInterop-constant.html
 [static interop]: {{site.pub-pkg}}/js#staticinterop
 [`@JSExport`]: {{site.pub-pkg}}/js#jsexport-and-js_utilcreatedartexport
 
--- a/src/web/js-interop.md
+++ b/src/web/js-interop.md
@@ -22,3 +22,77 @@ For help using the `js` package, see the following:
 [js-api]: {{site.pub-api}}/js
 [sass]: {{site.pub-pkg}}/sass
 
+## Future-proof JS interop
+
+Dart's JS interop story is currently evolving. 
+Many of the features that enable future JS interop,
+and support of all three compilers,
+are ready to use (Dart version 2.19). 
+To ease the transition to future JS interop,
+you can refactor your code to conform to the
+syntax and semantics of the new features now.
+
+{{site.alert.note}}
+  Future-proofing is for those *very* eager to
+  prepare their code for [Dart 3][], or to try
+  integrating with [`dart2wasm`][] before Dart 3.
+  None of this necessarily works *performantly* on
+  `dart2wasm` yet, but refactoring now means you
+  won't need to later, if that's important to you.
+{{site.alert.end}}
+
+The following sections are the set of features
+expected to work across compilers for JS interop,
+per library.
+
+*Requirements:*
+* Dart SDK constraint: `>= 2.19`
+* [`package:js`][] constraint: `>= 0.6.6`
+
+[Dart 3]: https://medium.com/dartlang/the-road-to-dart-3-afdd580fbefa
+[`dart2wasm`]: https://github.com/dart-lang/sdk/blob/main/pkg/dart2wasm/dart2wasm.md#running-dart2wasm
+[`package:js`]: https://pub.dev/packages/js
+
+### `package:js`
+
+If you're interested in interop that will work with
+`dart2wasm` when Dart 3 is released, you can implement
+[static interop][] using the `package:js` class `@staticInterop`.
+Static interop will be continually eveloving until Dart 3.
+
+The set of features for future static interop currently includes:
+* `@staticInterop` interfaces
+  * External factory constructors with and without `@anonymous`
+  * External `static` class members
+  * External non-`static` extension members on a `@staticInterop`
+  class (fields, getters, setters, methods)
+  * Non-external extension members on a `@staticInterop` class
+* Top-level external members
+* [`@JSExport`][] for mocking and exports
+
+[static interop]: https://github.com/dart-lang/sdk/tree/main/pkg/js#staticinterop
+[`@JSExport`]: https://github.com/dart-lang/sdk/tree/main/pkg/js#jsexport-and-js_utilcreatedartexport
+
+### `dart:js_util`
+
+If you're interested in refactoring your [`js_util`][]
+interop code in preparation for Dart 3, the following
+list of features for the library currently work:
+* jsify
+* dartify
+* getProperty
+* hasProperty
+* setProperty
+* callMethod
+* callConstructor
+* instanceof
+* promiseToFuture
+* globalThis
+* newObject
+* createStaticInteropMock
+* createDartExport
+* allowInterop
+
+The library may continue to evolve before Dart 3.
+
+[`js_util`]: https://pub.dev/packages/js_util
--- a/src/web/js-interop.md
+++ b/src/web/js-interop.md
@@ -33,30 +33,31 @@ Dart's JS interop story is currently evolving.
 Many of the features that enable future JS interop
 are ready to experiment with as of Dart version 2.19.
 These features support the existing production
-and development web compilers, as well as Dart's
-in-progress Wasm compiler ([`dart2wasm`][]).
+and development web compilers, 
+as well as Dart's in-progress Wasm compiler ([`dart2wasm`][]).
 
 For a glimpse into the next generation of JS interop,
-you can refactor your code to conform to the new
-syntax and semantics now. Doing so will
-likely not prevent the need to refactor again once
-[Dart 3][] lands, as the features are still in development. 
-However, the features available for preview are much
-closer to future JS interop than any pattern supported today.
+you can refactor your code to conform to
+the new syntax and semantics now. 
+Doing so will likely not prevent the need to refactor again
+once [Dart 3][] lands, as the features are still in development. 
+However, the features available for preview are much closer
+to future JS interop than any pattern supported today.
 So, there are a few reasons to try them out now:
 
 * New JS interop developers can learn and build with future JS interop
-so they won't have to unlearn obsolete patterns in a few months.
+  so they won't have to unlearn obsolete patterns in a few months.
 * Existing JS interop developers eager to experiment with
-the latest features in JS interop
-or with `dart2wasm` when it becomes available.
-* Potentially ease transition of existing JS
-interop code once migration becomes necessary.
+  the latest features in JS interop 
+  or with `dart2wasm` when it becomes available.
+* Potentially ease transition of existing JS interop code
+  once migration becomes necessary.
 
 The following sections are the set of features
 expected to work across compilers for JS interop.
 
 *Requirements:*
+
 * Dart SDK constraint: `>= 2.19`
 * [`package:js`][] constraint: `>= 0.6.6`
 
@@ -73,28 +74,31 @@ more likely to perform better, and required for `dart2wasm`.
 Static interop addresses several gaps in the existing JS interop story:
 
 * **Missing features:** Static interop enables previously
-unavailable features, like easily wrapping and transforming APIs,
-renaming members, and static checking.
+  unavailable features, like easily wrapping and transforming APIs,
+  renaming members, and static checking.
 
 * **Inconsistencies:** Static interop makes backends more consistent,
-so development and production web compilers won't behave as differently
-as before.
+  so development and production web compilers
+  won't behave as differently as before.
 
-* **Clarity:** Static interop takes a step towards making JS interop
-more idiomatic in Dart, and making the boundary between the two languages more visible.
-For example, it enforces that JS classes are not meant to be mixed with Dart
-(dynamic calls aren't allowed, and JS interop types cannot be implemented as a Dart class).
-We expect upcoming integration with inline classes
-will help make interop even more idiomatic.
+* **Clarity:** Static interop takes a step towards making 
+  JS interop more idiomatic in Dart, 
+  and making the boundary between the two languages more visible.
+  For example, it enforces that JS classes are not meant to be mixed with Dart
+  (dynamic calls aren't allowed, 
+  and JS interop types cannot be implemented as a Dart class).
+  We expect upcoming integration with inline classes
+  will help make interop even more idiomatic.
 
-You can implement static interop using the `package:js`
-annotation `@staticInterop`.
+You can implement static interop using
+the `package:js` annotation `@staticInterop`.
 The set of features for future static interop currently includes:
+
 * `@staticInterop` interfaces
   * External factory constructors with and without `@anonymous`
   * External `static` class members
   * External non-`static` extension members on a `@staticInterop`
-  class (fields, getters, setters, methods)
+    class (fields, getters, setters, methods)
   * Non-external extension members on a `@staticInterop` class
 * Top-level external members
 * [`@JSExport`][] for mocking and exports
--- a/src/web/js-interop.md
+++ b/src/web/js-interop.md
@@ -51,13 +51,13 @@ per library.
 
 [Dart 3]: https://medium.com/dartlang/the-road-to-dart-3-afdd580fbefa
 [`dart2wasm`]: https://github.com/dart-lang/sdk/blob/main/pkg/dart2wasm/dart2wasm.md#running-dart2wasm
-[`package:js`]: https://pub.dev/packages/js
+[`package:js`]: {{site.pub-pkg}}/js
 
 ### `package:js`
 
 If you're interested in interop that will work with
 `dart2wasm` when Dart 3 is released, you can implement
-[static interop][] using the `package:js` class `@staticInterop`.
+[static interop][] using the `package:js` annotation `@staticInterop`.
 Static interop will be continually eveloving until Dart 3.
 
 The set of features for future static interop currently includes:
@@ -70,8 +70,8 @@ The set of features for future static interop currently includes:
 * Top-level external members
 * [`@JSExport`][] for mocking and exports
 
-[static interop]: https://github.com/dart-lang/sdk/tree/main/pkg/js#staticinterop
-[`@JSExport`]: https://github.com/dart-lang/sdk/tree/main/pkg/js#jsexport-and-js_utilcreatedartexport
+[static interop]: https://pub.dev/packages/js#staticinterop
+[`@JSExport`]: https://pub.dev/packages/js#jsexport-and-js_utilcreatedartexport
 
 ### `dart:js_util`
 
--- a/src/web/js-interop.md
+++ b/src/web/js-interop.md
@@ -2,7 +2,6 @@
 title: JavaScript interoperability
 short-title: JS interop
 description: "Use package:js to integrate JavaScript code into your Dart web app."
-toc: false
 ---
 
 The [Dart web platform](/overview#web-platform) supports calling
@@ -87,11 +86,9 @@ Static interop addresses several gaps in the existing JS interop story:
   For example, it enforces that JS classes are not meant to be mixed with Dart
   (dynamic calls aren't allowed, 
   and JS interop types cannot be implemented as a Dart class).
-  We expect upcoming integration with inline classes
-  will help make interop even more idiomatic.
 
 You can implement static interop using
-the `package:js` annotation `@staticInterop`.
+the `package:js` annotation [`@staticInterop`][].
 The set of features for future static interop currently includes:
 
 * `@staticInterop` interfaces
@@ -106,8 +103,9 @@ The set of features for future static interop currently includes:
 To learn how to implement static interop and see examples,
 visit the [static interop][] specification.
 
-[static interop]: https://pub.dev/packages/js#staticinterop
-[`@JSExport`]: https://pub.dev/packages/js#jsexport-and-js_utilcreatedartexport
+[`@staticInterop`]: https://pub.dev/documentation/js/latest/js/staticInterop-constant.html
+[static interop]: {{site.pub-pkg}}/js#staticinterop
+[`@JSExport`]: {{site.pub-pkg}}/js#jsexport-and-js_utilcreatedartexport
 
 ### `dart:js_util`
 
--- a/src/web/js-interop.md
+++ b/src/web/js-interop.md
@@ -22,44 +22,69 @@ For help using the `js` package, see the following:
 [js-api]: {{site.pub-api}}/js
 [sass]: {{site.pub-pkg}}/sass
 
-## Future-proof JS interop
-
-Dart's JS interop story is currently evolving. 
-Many of the features that enable future JS interop,
-and support of all three compilers,
-are ready to use (Dart version 2.19). 
-To ease the transition to future JS interop,
-you can refactor your code to conform to the
-syntax and semantics of the new features now.
+## Next-generation JS interop preview
 
 {{site.alert.note}}
-  Future-proofing is for those *very* eager to
-  prepare their code for [Dart 3][], or to try
-  integrating with [`dart2wasm`][] before Dart 3.
-  None of this necessarily works *performantly* on
-  `dart2wasm` yet, but refactoring now means you
-  won't need to later, if that's important to you.
+  This interop feature is **experimental**,
+  and [in active development](https://github.com/dart-lang/sdk/issues/35084).
 {{site.alert.end}}
 
+Dart's JS interop story is currently evolving. 
+Many of the features that enable future JS interop
+are ready to experiment with as of Dart version 2.19.
+These features support the existing production
+and development web compilers, as well as Dart's
+in-progress Wasm compiler ([`dart2wasm`][]).
+
+For a glimpse into the next generation of JS interop,
+you can refactor your code to conform to the new
+syntax and semantics now. Doing so will
+likely not prevent the need to refactor again once
+[Dart 3][] lands, as the features are still in development. 
+However, the features available for preview are much
+closer to future JS interop than any pattern supported today,
+so there are a few reasons to try them out now:
+* Potentially ease transition of existing JS
+interop code once migration becomes necessary.
+* Existing JS interop developers eager to prepare their code for Dart 3.
+* Existing JS interop developers eager to integrate with `dart2wasm`.
+* New JS interop developers learning future JS interop
+so they won't have to unlearn obsolete patterns in a few months.
+
 The following sections are the set of features
-expected to work across compilers for JS interop,
-per library.
+expected to work across compilers for JS interop.
 
 *Requirements:*
 * Dart SDK constraint: `>= 2.19`
 * [`package:js`][] constraint: `>= 0.6.6`
 
-[Dart 3]: https://medium.com/dartlang/the-road-to-dart-3-afdd580fbefa
 [`dart2wasm`]: https://github.com/dart-lang/sdk/blob/main/pkg/dart2wasm/dart2wasm.md#running-dart2wasm
+[Dart 3]: https://medium.com/dartlang/the-road-to-dart-3-afdd580fbefa
 [`package:js`]: {{site.pub-pkg}}/js
 
 ### `package:js`
 
-If you're interested in interop that will work with
-`dart2wasm` when Dart 3 is released, you can implement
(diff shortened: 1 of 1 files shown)


srujzs:
Thanks for adding this! I think it's what I had in mind for a future-proof interop, but I had a couple of comments.

As for landing this, I can go either way. @kevmoo, do you think the users you had in mind who want to use future-proof interop will find it valuable to have this here versus pointing them to the issue?

src/web/js-interop.md
```diff
@@ -22,3 +22,77 @@ For help using the `js` package, see the following:
...
+  * External non-`static` extension members on a `@staticInterop`
+  class (fields, getters, setters, methods)
+  * Non-external extension members on a `@staticInterop` class
+* Top-level external members
+* [`@JSExport`][] for mocking and exports
+
+[static interop]: https://github.com/dart-lang/sdk/tree/main/pkg/js#staticinterop
+[`@JSExport`]: https://github.com/dart-lang/sdk/tree/main/pkg/js#jsexport-and-js_utilcreatedartexport
+
+### `dart:js_util`
```
parlough, line 76: @srujzs One thing that isn't clear to me, should people be using `dart:js_util` or the rexported version in `package:js`?

src/web/js-interop.md
```diff
@@ -22,3 +22,77 @@ For help using the `js` package, see the following:
...
+
+[Dart 3]: https://medium.com/dartlang/the-road-to-dart-3-afdd580fbefa
+[`dart2wasm`]: https://github.com/dart-lang/sdk/blob/main/pkg/dart2wasm/dart2wasm.md#running-dart2wasm
+[`package:js`]: https://pub.dev/packages/js
+
+### `package:js`
+
+If you're interested in interop that will work with
+`dart2wasm` when Dart 3 is released, you can implement
+[static interop][] using the `package:js` class `@staticInterop`.
```
parlough, line 60: When you add the `@`, it's not really a class, more so an annotation.

```suggestion
[static interop][] using the `package:js` annotation `@staticInterop`.
```

parlough, line 60: Since we aren't including an inline example, I would love for it to be more clear that this "static interop" link links to useful documentation and an example.

Perhaps an additional mention after the list of supported features saying something about "To learn more about how to implement static interop and see an example, see ..."?

src/web/js-interop.md
```diff
@@ -22,3 +22,77 @@ For help using the `js` package, see the following:
...
+  * External factory constructors with and without `@anonymous`
+  * External `static` class members
+  * External non-`static` extension members on a `@staticInterop`
+  class (fields, getters, setters, methods)
+  * Non-external extension members on a `@staticInterop` class
+* Top-level external members
+* [`@JSExport`][] for mocking and exports
+
+[static interop]: https://github.com/dart-lang/sdk/tree/main/pkg/js#staticinterop
+[`@JSExport`]: https://github.com/dart-lang/sdk/tree/main/pkg/js#jsexport-and-js_utilcreatedartexport
```
parlough, line 74: I think it might be better to link to the pub page. I know the links don't work yet, but once Dart 2.19 is released, I believe pub will no longer consider 0.6.6 a preview release and they should work.

```suggestion
[static interop]: https://pub.dev/packages/js#staticinterop
[`@JSExport`]: https://pub.dev/packages/js#jsexport-and-js_utilcreatedartexport
```

src/web/js-interop.md
```diff
@@ -22,3 +22,77 @@ For help using the `js` package, see the following:
...
+
+### `dart:js_util`
+
+If you're interested in refactoring your [`js_util`][]
+interop code in preparation for Dart 3, the following
+list of features for the library currently work:
```
parlough, line 80: "currently work" in the context of what? Work at all? Across web compilers including `dart2wasm`?

\cc @srujzs

src/web/js-interop.md
```diff
@@ -22,3 +22,77 @@ For help using the `js` package, see the following:
...
+* promiseToFuture
+* globalThis
+* newObject
+* createStaticInteropMock
+* createDartExport
+* allowInterop
+
+The library may continue to evolve before Dart 3.
+
+[`js_util`]: https://pub.dev/packages/js_util
```
parlough, line 98: I think this link should point elsewhere, but I'm not exactly sure where. `js_util` has moved around a bit.

- `package:js_util`: A long discontinued package which is linked to here
- `dart:js_util`: A first-class library in the SDK with the maintained util/interop functionality
- 'package:js/js_util.dart': A library in `package:js` which seems to reexport `dart:js_util`

Not sure if it should point to `dart:js_util` API docs or `package:js/js_util.dart` docs. I asked in https://github.com/dart-lang/site-www/pull/4528#discussion_r1080688553

src/web/js-interop.md
```diff
@@ -22,3 +22,77 @@ For help using the `js` package, see the following:
 [js-api]: {{site.pub-api}}/js
 [sass]: {{site.pub-pkg}}/sass
 
+## Future-proof JS interop
+
+Dart's JS interop story is currently evolving. 
+Many of the features that enable future JS interop,
+and support of all three compilers,
```
parlough, line 29: For someone first reading this "all three compilers" may be a bit confusing, as they may think of non-web compilers or not know the difference or existence of the different web compilers.

Maybe something "all web compilers, including Dart's in-progress Wasm compiler," as that seems to be the important part of this section.

src/web/js-interop.md
```diff
@@ -22,3 +22,77 @@ For help using the `js` package, see the following:
...
+expected to work across compilers for JS interop,
+per library.
+
+*Requirements:*
+* Dart SDK constraint: `>= 2.19`
+* [`package:js`][] constraint: `>= 0.6.6`
+
+[Dart 3]: https://medium.com/dartlang/the-road-to-dart-3-afdd580fbefa
+[`dart2wasm`]: https://github.com/dart-lang/sdk/blob/main/pkg/dart2wasm/dart2wasm.md#running-dart2wasm
+[`package:js`]: https://pub.dev/packages/js
```
parlough, line 54:
```suggestion
[`package:js`]: {{site.pub-pkg}}/js
```

parlough (commented):
Thanks for working on this! I have some minor suggestions, comments, and questions.

I definitely do think there is value in landing this since if the more future-proof "way" is already supported, this helps increase that awareness. In the best case this helps new JS-interop developers start by building this way rather than other ways while also helping existing JS-interop developers prepare or just know to keep an eye on this area.

This page is also linked to from the [`dart:js`](https://api.dart.dev/stable/dart-js/dart-js-library.html) API docs as a destination to learn more about how developers should do things now. Sidenote: we should consider making the notice on that page more prominent.

src/web/js-interop.md
MaryaBelanger: I was hoping someone would catch this, because _I_ don't know what "all three compilers" means, I think I got it from one of the related issues or issue comments. What are all three? 

I would rather say something like 
> and support of web compilers, including ________, _________, _and_ Dart's in-progress Wasm compiler"

MaryaBelanger:
Thanks @parlough , I'll incorporate the rest of your suggestions (the ones I'll have to update manually) once @srujzs responds with the details.

> In the best case this helps new JS-interop developers start by building this way rather than other ways 

This is a good point and I think it changes the nature of the note saying this is only for people eager to prepare. I'll change the wording slightly to emphasize people who are just starting to implement js interop should do it this way.

src/web/js-interop.md
```diff
@@ -22,3 +22,77 @@ For help using the `js` package, see the following:
...
+you can refactor your code to conform to the
+syntax and semantics of the new features now.
+
+{{site.alert.note}}
+  Future-proofing is for those *very* eager to
+  prepare their code for [Dart 3][], or to try
+  integrating with [`dart2wasm`][] before Dart 3.
+  None of this necessarily works *performantly* on
+  `dart2wasm` yet, but refactoring now means you
+  won't need to later, if that's important to you.
```
kevmoo, line 41: @srujzs – don't we get a bit of size/speed improvement using the new stuff NOW?

parlough: I seemed to get a little one in https://github.com/dart-lang/dartdoc/pull/3299 :)

src/web/js-interop.md
```diff
@@ -22,3 +22,77 @@ For help using the `js` package, see the following:
 [js-api]: {{site.pub-api}}/js
 [sass]: {{site.pub-pkg}}/sass
 
+## Future-proof JS interop
```
mit-mit, line 25: "Future-proof" is a pretty big label; and one I suspect most software projects can't really promise...

Can I suggest something like "next-generation"?

src/web/js-interop.md
```diff
@@ -22,3 +22,77 @@ For help using the `js` package, see the following:
...
+* [`package:js`][] constraint: `>= 0.6.6`
+
+[Dart 3]: https://medium.com/dartlang/the-road-to-dart-3-afdd580fbefa
+[`dart2wasm`]: https://github.com/dart-lang/sdk/blob/main/pkg/dart2wasm/dart2wasm.md#running-dart2wasm
+[`package:js`]: {{site.pub-pkg}}/js
+
+### `package:js`
+
+If you're interested in interop that will work with
+`dart2wasm` when Dart 3 is released, you can implement
```
mit-mit, line 59: It's too early to state confidently that `dart2wasm` will 100% be in Dart 3.

src/web/js-interop.md
```diff
@@ -22,3 +22,77 @@ For help using the `js` package, see the following:
...
+
+The following sections are the set of features
+expected to work across compilers for JS interop,
+per library.
+
+*Requirements:*
+* Dart SDK constraint: `>= 2.19`
+* [`package:js`][] constraint: `>= 0.6.6`
+
+[Dart 3]: https://medium.com/dartlang/the-road-to-dart-3-afdd580fbefa
```
mit-mit, line 52: In a few days this should probably be the new blog post

src/web/js-interop.md
```diff
@@ -22,3 +22,77 @@ For help using the `js` package, see the following:
 [js-api]: {{site.pub-api}}/js
 [sass]: {{site.pub-pkg}}/sass
 
+## Future-proof JS interop
+
```
mit-mit, line 26: We need a note here clearly stating that this work is experimental, similar to e.g. the note here: https://dart.dev/guides/libraries/objective-c-interop

src/web/js-interop.md
```diff
@@ -22,3 +22,77 @@ For help using the `js` package, see the following:
...
+Dart's JS interop story is currently evolving. 
+Many of the features that enable future JS interop,
+and support of all three compilers,
+are ready to use (Dart version 2.19). 
+To ease the transition to future JS interop,
+you can refactor your code to conform to the
+syntax and semantics of the new features now.
```
sigmundch, line 33: I fear this may be too strong as a promise.

The syntax that is available today via `@staticInterop` works semantically similar to the future JSInterop we will have with inline-classes, but both syntax and semantics are expected to change (syntax is expected to change more than semantics at this point, but some changes are expected on both).

That said, the `@staticInterop` approach is much much much closer to that future JSInteorp than any other pattern we support today (non-`@staticInterop` package:js, or the old and deprecated `dart:js`)

So my TLDR: migrating now will not save you from a migration later, but will let you experiment and experience something similar to the next-gen JSInterop

src/web/js-interop.md
sigmundch: Similar note as I mentioned above: refactoring now means simpler refactoring later, but we can't promise no changes later.

Re performance, the current statement seems OK to me:
* dart2wasm has known overheads that are not polished yet (e.g. cost of a trampoline between JS and wasm is high right now)
* dart2js performance can be observably better or worse depending on the scenario. Calls are more direct, code-size may be smaller, but there are new soundness checks at the boundary that could potentially make some APIs slower. 

Circling back to the intent of this note, I'd say this preview of the next gen interop is meant mostly for experimentation: `package:js` interop as it exist today (w/o `@staticInterop`) will not be supported by dart2wasm, `dart:js_util` is supported, but it's not as ergonomic and will not be optimized. So we expect that anyone interested in trying dart2wasm that also needs jsinterop, will likely be interested in this newer form of JSInterop, even though more changes are expected as we introduce inline-classes and make additional tweaks in the next few months.

src/web/js-interop.md
sigmundch: :heavy_plus_sign: Also consider adding the word "preview" or "approximation" - it may be good to highlight this is not what the finished product looks like, but a glimpse of the semantics we'll have in the future.

src/web/js-interop.md
MaryaBelanger: I can remove that clause and just say "Refactoring now means...."

src/web/js-interop.md
sigmundch: Also - I am not sure how much emphasis we are putting on dart2wasm on other documentation. I see that we are staring to mention it in the blog post, but I wasn't sure how comfortable we are emphasizing it here.

Circling back to the sentence here. My take is that "Next generation interop" addresses several gaps we had in JSInterop in the past:

* missing features: before we couldn't easily wrap and transform APIs, rename members, and we missed static checking. With static interop we can now do all of these.
* inconsistencies: before development vs production web compilers sometimes behaved differently. With static interop this is now better, backends are more consistent. **
* clarity: JSInterop adds cognitive challenges because it is using Dart syntax in a non-Dart way. Static interop doesn't solve this yet, but inline-classes will help us make this feature be more idiomatic.

** Since our investment in dart2wasm, we discovered that details above like the lack of static checking made existing JSInterop approaches prohibitively expensive to support in a non JavaScript environment (like wasm). Static interop enables us to have an interop solution that we can support better in dart2wasm, and as a result will enable experimentation on that backend whenever it is available as a preview.

src/web/js-interop.md
sigmundch: We recommend dart:js_util directly. This is because we hope in the future to move away from `package:js` having the annotations you use for JSInteop. Those will move into `dart:js_interop` once we are ready to publish the new form of JSInterop that uses inline-classes.

src/web/js-interop.md
sigmundch: I'd point to dart:js_util

src/web/js-interop.md
sigmundch: I'm not sure if it's worth highlighting the list of APIs here. This list is practically the entire set of APIs exposed in `dart:js_util`.

Long term our recommendation is to use inline-classes with static JSInterop whenever that's possible. And only use js_util when it's absolutely necessary. For example, we have a bug and you need a workaround, or there is something that can't be expressed with static interop (possibly an omission we'd like to fix as well).

It may come up that some migrations are easier if js_util is available. For example, preexisting uses of `dart:js` (now deprecated) can be replaced by a combination of static interop and dart:js_util. The former is what we recommend, but it may be easier to automate migrations using the latter :shrug:

src/web/js-interop.md
MaryaBelanger: Correct me if I'm wrong: changing link to this https://api.dart.dev/stable/dart-js_util/dart-js_util-library.html

parlough: Link makes sense to me :)

src/web/js-interop.md
MaryaBelanger: > {{site.alert.note}}
>       This interop feature is **experimental**,
>       and [in active development]().
> {{site.alert.end}}

What should I link "in active development" to? This issue: https://github.com/dart-lang/sdk/issues/50721 ?

parlough: Maybe https://github.com/dart-lang/sdk/issues/35084?

src/web/js-interop.md
MaryaBelanger: Great, I'll focus the section on addressing the general improvements instead of directly relating it to dar2wasm

src/web/js-interop.md
MaryaBelanger: Ok, so I'll remove the list of APIs, and just describe the general long-term goal of only using js_util for edge cases that static interop isn't designed to handle?

src/web/js-interop.md
MaryaBelanger: Oh wow, _there's_ the giant background issue I've been dreaming of! 😅 Thanks @parlough that'll be very helpful for all the upcoming JSinterop docs work

src/web/js-interop.md
MaryaBelanger: @sigmundch One more question about this: are you basically saying that `js_util` provides lower-level functionality that can be configured in a more customized way than static interop? So you're keeping it around just in case some unknown cases go unaccounted for in static interop's design?

sigmundch: Exactly. But we expect that to be rare. For context, in the past it was helpful to address some of the limitations of non-static-interop `package:js` that couldn't be done before we added static-interop.

One thing we want to make clear in our messaging is that we recommend static-interop, that static-interop is more declarative, and as a result more likely to be optimized and more likely to perform better. We may optimize js_util patterns, but that is not a guarantee we will promise in our backends.

MaryaBelanger:
Wooo... ok, that was a big change, thanks for the input everyone.

I don't really expect anyone to review this right now, but _it would be really nice_ to roll this in with the [v2.19 branch](https://github.com/dart-lang/site-www/pull/4541#pullrequestreview-1266524190), which is targeted for around 10am PST tomorrow.

@sigmundch if you could verify the extra context I added based on your comments, that would be so helpful! @mit-mit @parlough , your feedback or approval is greatly appreciated too if you can squeeze that in! Thank you

src/web/js-interop.md
```diff
@@ -22,3 +22,124 @@ For help using the `js` package, see the following:
...
+that static interop won't be able to  address.
+
+For example, we may discover that some migrations are
+easier to accomplish with `js_util`. Preexisting uses of
+`dart:js` (now deprecated) can be replaced by a combination
+of static interop and `dart:js_util`.
+The former is what we recommend, but it may be easier
+to automate migrations using the latter.
+
+The following subset of `dart:js_util` features
```
sigmundch, line 124: I'm still inclined to remove this paragraph, since this is listing almost everything in the dart:js_util library.

src/web/js-interop.md
```diff
@@ -22,3 +22,124 @@ For help using the `js` package, see the following:
...
+likely not prevent the need to refactor again once
+[Dart 3][] lands, as the features are still in development. 
+However, the features available for preview are much
+closer to future JS interop than any pattern supported today,
+so there are a few reasons to try them out now:
+* Potentially ease transition of existing JS
+interop code once migration becomes necessary.
+* Existing JS interop developers eager to prepare their code for Dart 3.
+* Existing JS interop developers eager to integrate with `dart2wasm`.
+* New JS interop developers learning future JS interop
```
sigmundch, line 51: Consider moving this to the top of the list?

src/web/js-interop.md
```diff
@@ -22,3 +22,124 @@ For help using the `js` package, see the following:
...
+you can refactor your code to conform to the new
+syntax and semantics now. Doing so will
+likely not prevent the need to refactor again once
+[Dart 3][] lands, as the features are still in development. 
+However, the features available for preview are much
+closer to future JS interop than any pattern supported today,
+so there are a few reasons to try them out now:
+* Potentially ease transition of existing JS
+interop code once migration becomes necessary.
+* Existing JS interop developers eager to prepare their code for Dart 3.
```
sigmundch, line 49: Technically we don't plan to break anything by Dart 3. Consider combining this and the bullet below to present it as all the future exicting things happening? For example:

* Existing JS interop developers eager to experiment with the latest features in JS interop or with `dart2wasm` when it becomes available.

sigmundch (commented):
Thanks Marya!

I know you are short on time, so just a few comments below. Will further changes be possible after the 10 AM time, or is the site going to be frozen for a while?

MaryaBelanger:
@sigmundch Actually not short on time anymore! We re-targeted this to `main` so the rest of the 2.19 updates are not waiting on it (they're [already in](https://github.com/dart-lang/site-www/pull/4541))

> Will further changes be possible after the 10 AM time, or is the site going to be frozen for a while?

We can keep making changes, no deadline, and as far as I'm aware we don't freeze dart.dev ever anyway. So take your time! Thanks for your concern :)

src/web/js-interop.md
MaryaBelanger: Oops I think I just forgot to do that!

src/web/js-interop.md
```diff
@@ -22,3 +22,110 @@ For help using the `js` package, see the following:
...
+visit the [static interop][] specification.
+
+[static interop]: https://pub.dev/packages/js#staticinterop
+[`@JSExport`]: https://pub.dev/packages/js#jsexport-and-js_utilcreatedartexport
+
+
+### `dart:js_util`
+
+We recommend using static interop over `js_util`
+in future JS interop.
```
sigmundch, line 112: consider removing "in future JS interop" altogether.

src/web/js-interop.md
```diff
@@ -22,3 +22,110 @@ For help using the `js` package, see the following:
...
+that static interop won't be able to  address.
+
+For example, we may discover that some migrations are
+easier to accomplish with `js_util`. Preexisting uses of
+`dart:js` (now deprecated) can be replaced by a combination
+of static interop and `dart:js_util`.
+The former is what we recommend, but it may be easier
+to automate migrations using the latter.
+
+While `dart:js_util` is supported by `dart2wasm` in future interop,
```
sigmundch, line 128: similarly here, consider dropping "in future interop". Maybe replace it by "in the future".

src/web/js-interop.md
```diff
@@ -22,3 +22,110 @@ For help using the `js` package, see the following:
...
+[static interop]: https://pub.dev/packages/js#staticinterop
+[`@JSExport`]: https://pub.dev/packages/js#jsexport-and-js_utilcreatedartexport
+
+
+### `dart:js_util`
+
+We recommend using static interop over `js_util`
+in future JS interop. 
+
+However, since [`js_util`][] provides lower-level
```
sigmundch, line 114: Reading this again, I am starting to feel the discussion here around why we still have js_util and it's use for migrations could be inadvertently encouraging more usage than we want. So I'm reconsidering whether we should highlight it as much.

For example, how do you feel about replacing this section entirely with a more direct paragraph along these lines:

> `dart:js_util` provides a low-level interop API and is supported by the JS and `dart2wasm` backends. `dart:js_util` can provide more flexibility, for example in rare edge cases we haven't accounted for where static interop is not expressive enough. That said, it is not as ergonomic, and it is not planned to be optimized in the same way as static interop. As a result, we highly recommend using static interop over `dart:js_util` whenever it's possible.

sigmundch (approved):
Thank you so much @MaryaBelanger!

A few small suggestions below.

src/web/js-interop.md
MaryaBelanger: That's much better, I agree the last version was a little too encouraging for js_util. Thanks for the text!

src/web/js-interop.md
```diff
@@ -22,3 +22,100 @@ For help using the `js` package, see the following:
...
+
+[`dart2wasm`]: https://github.com/dart-lang/sdk/blob/main/pkg/dart2wasm/dart2wasm.md#running-dart2wasm
+[Dart 3]: https://medium.com/dartlang/dart-3-alpha-f1458fb9d232
+[`package:js`]: {{site.pub-pkg}}/js
+
+### `package:js`
+
+The key feature of next-generation JS interop is [static interop][].
+We recommend using static interop as the default for `package:js`,
+as it is more declaritive, more likely to be optimized,
```
atsansone, line 71: **typo:** declarative.
```suggestion
as it is more declarative, more likely to be optimized,
```

src/web/js-interop.md
```diff
@@ -22,3 +22,100 @@ For help using the `js` package, see the following:
...
+renaming members, and static checking.
+
+* **Inconsistencies:** Static interop makes backends more consistent,
+so development and production web compilers won't behave as differently
+as before.
+
+* **Clarity:** Static interop takes a step towards making JS interop
+more idiomatic in Dart, and making the boundary between the two languages more visible.
+For example, it enforces that JS classes are not meant to be mixed with Dart
+(e.g. no dynamic calls, JS interop types cannot be implemented as a Dart class).
```
atsansone, line 86: **issue:** Don't use e.g. per the [Style Guide](https://developers.google.com/style/word-list#eg).

Also, e.g. means _exempli gratia_ or _for example_. Don't use the same phrase twice in the same sentence.

Consider:
```markdown
This reinforces that the developer should not mix JavaScript and Dart classes.
Developers can't implement JavaScript interop types as a Dart class.
```

atsansone (commented):
@MaryaBelanger : Gave a light review. There's a spelling mistake in the copy and a few suggestions.

src/web/js-interop.md
```diff
@@ -22,3 +22,100 @@ For help using the `js` package, see the following:
...
+`dart:js_util` can provide more flexibility,
+for example, in potential, rare edge cases we haven't yet
+accounted for where static interop is not expressive enough.
+
+However, it is not as ergonomic, and we do not plan
+to optimize it in the same way as static interop.
+As a result, we highly recommend using static interop over
+`dart:js_util` whenever it's possible.
+
+[`dart:js_util`]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-js_util/dart-js_util-library.html
```
domesticmouse, line 121: Add trailing new line

domesticmouse (approved):
Excited for these changes!

src/web/js-interop.md
```diff
@@ -22,3 +22,104 @@ For help using the `js` package, see the following:
...
+  so development and production web compilers
+  won't behave as differently as before.
+
+* **Clarity:** Static interop takes a step towards making 
+  JS interop more idiomatic in Dart, 
+  and making the boundary between the two languages more visible.
+  For example, it enforces that JS classes are not meant to be mixed with Dart
+  (dynamic calls aren't allowed, 
+  and JS interop types cannot be implemented as a Dart class).
+  We expect upcoming integration with inline classes
```
parlough, line 90: Readers may not (or likely will not) know what inline classes are. Consider cross linking this to: https://github.com/dart-lang/language/blob/master/accepted/future-releases/inline-classes/feature-specification.md

src/web/js-interop.md
```diff
@@ -22,3 +22,104 @@ For help using the `js` package, see the following:
...
+  will help make interop even more idiomatic.
+
+You can implement static interop using
+the `package:js` annotation `@staticInterop`.
```
parlough, line 94: Optional: Consider linking this first `@staticInterop` to its API docs.

https://pub.dev/documentation/js/latest/js/staticInterop-constant.html

src/web/js-interop.md
```diff
@@ -22,3 +22,104 @@ For help using the `js` package, see the following:
...
+  * External non-`static` extension members on a `@staticInterop`
+    class (fields, getters, setters, methods)
+  * Non-external extension members on a `@staticInterop` class
+* Top-level external members
+* [`@JSExport`][] for mocking and exports
+
+To learn how to implement static interop and see examples,
+visit the [static interop][] specification.
+
+[static interop]: https://pub.dev/packages/js#staticinterop
```
parlough, line 109:
```suggestion
[static interop]: {{site.pub-pkg}}/js#staticinterop
```

src/web/js-interop.md
```diff
@@ -22,3 +22,104 @@ For help using the `js` package, see the following:
...
+[`@JSExport`]: https://pub.dev/packages/js#jsexport-and-js_utilcreatedartexport
```
parlough, line 110:
```suggestion
[`@JSExport`]: {{site.pub-pkg}}/js#jsexport-and-js_utilcreatedartexport
```

parlough (approved):
Really great job on this and adjusting for everyone's comments. Sorry I didn't respond earlier, but somehow I accidentally(?) unsubscribed to this PR (oops) and didn't realize Sigmund reviewed it again already. I have a few minor suggestions.

Also consider trying to remove `toc: false` from the frontmatter(top) of this page now as well and seeing how that looks, since there is more content on the page now. When there were no headings the TOC wasn't useful so it was added to remove it.

I've pushed some minor formatting fixes as well. The main thing was that in a list we indent the content of a list entry to match the text on its first line (https://github.com/dart-lang/site-shared/blob/main/doc/markdown.md#do-properly-indent-markdown-lists).

srujzs (approved):

srujzs:
I can't tell if my comments have been accidentally duplicated or something - GitHub UI is confusing to me.

Anyways, they've all been addressed so you can ignore them. This looks great, thanks for working on this!

parlough:
@srujzs Looks like the comments just weren't published until now for some reason. We appreciate all of your insights and comments :D

srujzs:
> @srujzs Looks like the comments just weren't published until now for some reason, but I can see them now :)

!!! Ugh, sorry about that, I guess I needed to click the "Start Review" button to publish. Luckily, Siggi's comments encapsulated all of mine.

src/web/js-interop.md
parlough: Or maybe this issue would be better: https://github.com/dart-lang/language/issues/2727

MaryaBelanger: Yeah, now that you mention int I actually think just removing that line would be best

src/web/js-interop.md
MaryaBelanger: Idk why I can't make that link work with `{{site.pub-api}}/` syntax. I've tried a few ways but get a 404 every time
```
{{site.pub-api}}/staticInterop-constant
{{site.pub-api}}/js/staticInterop-constant
{{site.pub-api}}/js/js/staticInterop-constant
{{site.pub-api}}/js/latest/js/staticInterop-constant
```

parlough: @MaryaBelanger Currently pub.dev and api.dart.dev aren't set up to handle a missing `.html`, I believe the last one should work:

```
{{site.pub-api}}/js/latest/js/staticInterop-constant.html
```

MaryaBelanger:
Added a couple minor edits just now. I'll wait a little while to merge just in case any final concerns are lingering, but I think this is done (@srujzs your comment conundrum confused me a little too 😅 I think everything you talked about is incorporated? But let me know if there's anything specific you want to reword/remove/link to/etc.)

Thanks everyone!

PR #4536, opened 2023-01-22
[alpha] Mention alpha build availability for testing

mit-mit:


mit-mit:
@MaryaBelanger can you please review this?

src/null-safety/index.md
```diff
@@ -115,6 +115,16 @@ To resolve these issues, check for
 of any packages you installed from pub.dev, and [migrate](#migrate) all
 of your own source code to use sound null safety.
 
+As of late January 2023, Dart 3 alpha is available from the Dart dev channel
+and the Flutter master channel, and we recommend you test your code
+for Dart 3 compatibility using that release:
+
+```terminal
+$ dart --version                     # make sure this reports 3.0.0-151.0.dev or higher
```
MaryaBelanger, line 123: I'm wondering, if `dart --version` _doesn't_ report that version back, or I already know I _don't_ have the dart dev channel or flutter master channel set up, how can I fix that? Maybe the steps are described elsewhere on this page, I didn't check.

But just in case, can we anchor link on "the Dart dev channel"/"the Flutter master" channel to instructions for setting up those channels, even if it's just a couple lines somewhere? And/or where it says "make sure" `dart --version` reports <...>, point them to what they need to do to get it to report that.

src/null-safety/index.md
```diff
@@ -115,6 +115,16 @@ To resolve these issues, check for
...
+$ dart pub get                       # this should resolve without issues
+$ dart analyze / flutter analyze     # this should pass without errors
```
MaryaBelanger, line 125: Kind of the same as the previous comment, "this _should_ ....": if it should do something, can we point them to some troubleshooting or something to do if it _doesn't_ do what it should? 

If the first comment I left about pointing them to how to set up those channels would ensure that these lines will behave as expected, then nothing else needs to be added here.

src/null-safety/index.md
mit-mit: Good point. I'm fixing that by adding instructions on the download page:
https://github.com/dart-lang/site-www/pull/4548

src/null-safety/index.md
```diff
@@ -115,6 +115,24 @@ To resolve these issues, check for
 of any packages you installed from pub.dev, and [migrate](#migrate) all
 of your own source code to use sound null safety.
 
+As of late January 2023, **Dart 3 alpha** is available from the Dart dev channel
+and the Flutter master channel; see [the download page][] for details.
+We recommend you test your code for Dart 3 compatibility using that release:
+
+```terminal
+$ dart --version                     # make sure this reports 3.0.0-151.0.dev or higher
+$ dart pub get                       # this should resolve without issues
```
parlough, line 124: If you're going to specify the `flutter` equivalent command for `analyze`, should probably do so here to:

```suggestion
$ dart pub get / flutter pub get     # this should resolve without issues
```

src/null-safety/index.md
```diff
@@ -115,6 +115,24 @@ To resolve these issues, check for
...
+$ dart analyze / flutter analyze     # this should pass without errors
+```
+
+If the `pub get` step fails, check the [status of the dependencies][].
```
parlough, line 128: Maybe "migration status of your dependencies"?

src/null-safety/index.md
```diff
@@ -115,6 +115,24 @@ To resolve these issues, check for
 of any packages you installed from pub.dev, and [migrate](#migrate) all
 of your own source code to use sound null safety.
 
+As of late January 2023, **Dart 3 alpha** is available from the Dart dev channel
```
parlough, line 118: This section got pretty large. Consider putting it in a subheader.

```suggestion
### Dart 3 alpha

As of late January 2023, **Dart 3 alpha** is available from the Dart dev channel
```

mit-mit: I want this to be prominent, so keeping. It's only for some months anyway

mit-mit:
@MaryaBelanger @parlough I'm going to merge this now, as it's needed for the blog post which is going out imminently.

Please fell free to tweak (even without waiting for my approval) in follow-up PRs.

parlough (approved):
Sounds good to me :)

PR #4540, opened 2023-01-23
Update banner for Flutter Forward

parlough:
Wasn't sure on the wording, such as including any Dart specific message, so I went with something generic.


mit-mit:
This is great for today!

On wednesday I think we might want to link to https://flutter.dev/events/flutter-forward#schedule

mit-mit (approved):

parlough:
Sounds good. I'll make sure it gets updated :)

PR #4543, opened 2023-01-24
Update various SDK constraints and constraint examples

parlough:
Most are updated to 2.19 while others to 2.12, since we should avoid having ones below 2.12 as examples.


parlough:
This is waiting on a 3.0 build to be released to the beta channel.

atsansone (approved):
@parlough : Great docs maintenance! Did you check that no errors occur with these changes?

PR #4550, opened 2023-01-25
New banner post-Flutter forward

MaryaBelanger:
image


parlough (approved):
Looks good to me, thanks for adding this!

MaryaBelanger:
@sfshaza2 What do you think about the anchor-linking here? Do we need to do it on a separate action item like "Watch now" or "Learn more", or is way ok?

sfshaza2 (commented):
lgtm

PR #4560, opened 2023-01-31
[Ignore] Individual page diagnostic proof of concept

parlough:
This a really rough and hacky proof of concept exploring the possibility of each diagnostic being on its own page to see if it's a direction we want to go. If it is a direction we want to go, I will rewrite this setup and code to be much cleaner and performant, and also make the necessary changes in analyzer to make this process easier.

- The home diagnostic page is now located at [`tools/diagnostics`](https://dart-dev--pr4560-feature-individual-p-qnvfdadd.web.app/tools/diagnostics).
  - This page no longer includes individual details, only the glossary and context definitions, as well as a definition list of the diagnostics themselves.
- Each diagnostic now has its own page in the format of `/tools/diagnostics/diagnostic_name` such as [`tools/diagnostics/conflicting_constructor_and_static_member](https://dart-dev--pr4560-feature-individual-p-qnvfdadd.web.app/tools/diagnostics/conflicting_constructor_and_static_member).
  - Currently the old diagnostic documentation is kept in its current state, but now that they are on their own page, we can make additions more easily
  - I added a new Jekyll/Liquid filter to insert the word-break opportunity element [`<wbr />`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/wbr) after underscores in the page h1 header, as the diagnostic names were too long and going off the page.
- The old diagnostic page still exists at [`tools/diagnostic-messages`](https://dart-dev--pr4560-feature-individual-p-qnvfdadd.web.app/tools/diagnostic-messages), but has the following modifications:
  - Has a JS script which reads the user's fragment and attempts to redirect them to the diagnostic's new page if it exists
  - If the user doesn't have JS enabled, they will see text which explains the documentation for that diagnostic has moved with a link to the new diagnostic documentation.
  - See [`/tools/diagnostic-messages#conflicting_constructor_and_static_member`](https://dart-dev--pr4560-feature-individual-p-qnvfdadd.web.app/tools/diagnostic-messages#conflicting_constructor_and_static_member) for an example of the redirect working
  - [Disable JS](https://developer.chrome.com/docs/devtools/javascript/disable/) temporarily and visit the previous link to see the existing page with redirect text
  - The page also has a yellow border indicating it's out of date
  - The page is hidden from the sitemap to prevent it from showing up on search
  - The page has a warning at the top linking to the new diagnostics article


RedBrogdon:
@parlough, do you need to keep this PR open, or has it served its exploratory purpose for now and can be closed?

I don't want to interrupt work if this is still a useful space, but I'm going through all the PRs and wanted to check.

PR #4566, opened 2023-02-01
Incorporate Dart 3.0 changes into `main` branch

parlough:
Incorporates all Dart 3 documentation changes into the `main` branch.


MaryaBelanger (approved):
Let's goooooo 🎉  (when it's time 🙃)

src/language/patterns.md
```diff
@@ -0,0 +1,362 @@
...
+```
+
+Many patterns make use of subpatterns, sometimes called _outer_ and _inner_
+patterns, respectively. Patterns match recursively on their subpatterns.
+For example, the individual fields of any [collection-type][] pattern could be 
+[variable patterns][variable] or [constant patterns][constant]:
+
+```dart
+switch (obj) {
+  // List pattern [a,b] matches obj first if obj is a list wth two fields,
```
pedromassango, line 60: Missing `i`

```suggestion
  // List pattern [a, b] matches obj first if obj is a list with two fields,
```

src/language/patterns.md
```diff
@@ -0,0 +1,362 @@
...
+
+### Destructuring
+
+When an object and pattern match, the pattern can then access the object's data 
+and extract it in parts. In other words, the pattern _destructures_ the object:
+
+```dart
+var numList = [1, 2, 3];
+// List pattern [a, b, c] destructures the three elements from numList...
+var [a, b, c] = numList;
```
pedromassango, line 77: We can use final keyword instead.

```suggestion
final numList = [1, 2, 3];
// List pattern [a, b, c] destructures the three elements from numList...
final [a, b, c] = numList;
```

src/language/patterns.md
```diff
@@ -0,0 +1,362 @@
...
+
+```dart
+switch (obj) {
+  case 1: ...
+    // Matches if 1 == obj.
+  case >= first && <= last: ...  
+    // Matches if the value of obj is between the constant values of 'first' and 'last'.
+  case (var a, var b): ...
+    // Matches if obj is a record with two fields, then assigns the fields to 'a' and 'b'.
+  case Rect(width: var w, height: var h): ...
```
pedromassango, line 155: Shouldn't those `var` properties be `final` as it is a Dart's best practice to declare final whenever the is no point change a property? Am I missing something? 👀

pedromassango (commented):
Found some small issues and things to improve. Would be great to be able to run the code snippets :)

src/language/patterns.md
parlough: There's a balance in documenting Dart, as both ways are supported and `var` is shorter, which is especially beneficial in this case. 

Unlike on Flutter, different Dart terms prefer different ways, as long as you are consistent.

_I do prefer `final` personally, but sometimes `var` is more approachable for the docs._

src/language/patterns.md
parlough: Thanks for catching this! Could you open a separate PR targeting this branch with a fix? Or I can. We just don't want to commit directly to this tracking branch.

PR #4574, opened 2023-02-03
Note on mono-repos

jonasfj:
I'm open to resolving this by better comments..

Context: https://github.com/dart-lang/setup-dart/issues/67

I think the simplest advice is to create a workflow for each package.
But really, it's up to package authors what they want to do.

I did a fairly complex setup in: https://github.com/google/dart-neats/blob/master/.github/workflows/publish.yml
This does avoid multiple files, but it's not something I'd suggest we write a guide about how to replicate. If users want that then perhaps they have to do it themselves :D


src/tools/pub/automated-publishing.md
```diff
@@ -116,6 +116,9 @@ jobs:
 
 Make sure to match the pattern in `on.push.tags` with the _tag pattern_
 specified on pub.dev. Otherwise, the GitHub Action workflow won’t work.
+If publishing multiple packages from the same repository, use a per-package
+_tag pattern_ like `my_package_name-v{{version}}` can create a workflow file
```
parlough, line 120: I'm assuming the tag pattern and the "workflow file for each package" are two separate ideas, but this text seems to combine them. Did you intend to separate these?

jonasfj: You'll need a workflow file for each tag-pattern.

There many ways to configure it, but that's the simplest one.

jonasfj: You'll configure a separate tag-pattern for each package on pub.dev, then create a workflow file for each package.
The workflow files will need to have a trigger pattern that aligns with the tag pattern for the package referenced in working-directory

jonasfj:
I think this looks fine to me. Feel free to merge.

parlough (approved):
Thanks for the extra context, looks good to me. 

I've adjusted the text slightly two make the two steps more defined, as I was little confused originally.

src/tools/pub/automated-publishing.md
parlough: Thanks for the extra context!

PR #4580, opened 2023-02-06
Added the onepub logo/url to the list of custom package repositories.

bsutton:
@sigurdm suggested raising a PR was the correct way to have OnePub added as an alternate custom package repository.

Here is the logo.

ufo - on white


src/tools/pub/custom-package-repositories.md
```diff
@@ -220,6 +220,10 @@ your own custom package repository:
   <img src="/assets/img/tools/jfrog.svg" width="48" alt="JFrog logo">
   <a href="https://www.jfrog.com/confluence/display/JFROG/Pub+Repositories"><b>JFrog Artifactory</b></a>
 </li>
+<li>
+  <img src="onepub-logo.svg" width="48" alt="OnePub logo">
```
parlough, line 224: Can you add the logo to [assets/img/tools/onepub-logo](https://github.com/dart-lang/site-www/tree/main/src/assets/img/tools) and then update this reference:

```suggestion
  <img src="/assets/img/tools/onepub-logo.svg" width="48" alt="OnePub logo">
```

parlough (changes requested):
Thanks for the PR! Always nice to see more package repositories :)

Can you add the logo into the repository next to the other logos used here? Also on line 210, can you update the ul's `class="col3"` to use `col2` instead now that we have four entries.

bsutton:
update the path, changed col3 to col2 and added the asset as requested.

src/tools/pub/custom-package-repositories.md
```diff
@@ -220,6 +220,10 @@ your own custom package repository:
   <img src="/assets/img/tools/jfrog.svg" width="48" alt="JFrog logo">
   <a href="https://www.jfrog.com/confluence/display/JFROG/Pub+Repositories"><b>JFrog Artifactory</b></a>
 </li>
+<li>
+  <img src="/assets/img/tools/onepub.svg" width="48" alt="OnePub logo">
+  <a href="https://onepub.dev"><b>OnePub Dart Repository</b></a>
```
parlough, line 225: One question. Is "Dart Repository" part of the name or is it simply "OnePub"? 

If it's just "OnePub", "Dart Repository" may be unnecessary since its apparent from the context, similar to Cloudsmith's entry.

parlough (approved):
Thanks for making those adjustments, looks good to me. One last question though:

bsutton:
It's is part of the name.

PR #4593, opened 2023-02-13
Standardize and adjust 'What's the point' boxes

parlough:
The primary goal here is to remove duplication of styles as well as unused functionality, with a few minor style improvements.

Removes the old `.panel` styles only used on the web tutorial pages and reuses the existing `.mini-toc` class. Also removes the `tutorial` layout as its functionality is not used anywhere.

Then makes adjustments to the `.mini-toc` so there is less unnecessary padding at the top and bottom, as well as a little less padding on the left of the `ul`.

After example:
After example with less padding

Before example:
Before example with unnecessary padding


domesticmouse (approved):

atsansone (dismissed):
@parlough : Check with @MaryaBelanger on this. She was discussing some standardization around content presentation, including a "what's the point" type setup and differentiating between true codelabs, internal codelabs, and tutorials. I'd ask you to pause this until her meeting on Thursday.

parlough:
Sure. As context, the main goal here is to standardize to one set of styles as currently we have two different styles/setups for these boxes with minor differences. This standardizes on the newer and preferred style, making any future changes easier as we'll only need to change one. There's not really any functional difference in terms of what content is presented.

MaryaBelanger:
I was literally just looking at these! 😅 There's a very similar "[What you'll learn](https://docs.flutter.dev/development/ui/layout/tutorial#:~:text=6%3A%20Final%20touch-,What%20you%E2%80%99ll%20learn,-How%20Flutter%E2%80%99s%20layout)" section used in some flutter docs too, which I think is a little less abrasive than "What's the point?"

But yes, I agree with Parker. This isn't about standardizing the purpose or templates of all future docs, this is just cleaning up the presentation. Whether we decide to call it "What's the point" or "What you'll learn", or whether we use it on every page or just tutorial pages (and what even is a "tutorial", while we're at it?), are all way more esoteric problems that aren't blocking this improvement. 

So this can definitely land as far as I'm concerned. If anything, having a cleaner backend for the UI will make it easier to implement our content standardization decisions once we get around to making them.

MaryaBelanger (approved):

atsansone:
@MaryaBelanger . @parlough : Perfect! Just wanted all on the same page.

Issue #4588, opened 2023-02-10
The "Language" umbrella


MaryaBelanger:
This issue is about disassembling the Language tour page (see #4565). To have a clean slate to unpack the language tour into, this also covers moving some of the existing content to more suitable locations.

### New structure proposal
Top level bullets are what's visible when you first expand "Language" in the left nav.

Second level bullets will either just be individual pages or combined with other topics on a single page. 
(Will decide per topic how to organize, but for the sake of time, second level bullets will all be on the parent page to begin with (for example, there's one page named "Types" and that has sections for all the built-in types, records, and typedefs)).

> Language tour _(previously "Language")_
> - Basics 
>   - [A basic Dart program](https://dart.dev/guides/language/language-tour#a-basic-dart-program)
>   - [Important concepts](https://dart.dev/guides/language/language-tour#important-concepts)
>   - [Keywords](https://dart.dev/guides/language/language-tour#keywords)
>   - [Variables](https://dart.dev/guides/language/language-tour#variables)
>   - [Operators](https://dart.dev/guides/language/language-tour#operators)
>   - [Metadata](https://dart.dev/guides/language/language-tour#metadata) _I moved this up from the bottom of the language tour to consolidate_
>   - [Comments](https://dart.dev/guides/language/language-tour#comments) _I moved this up from the bottom of the language tour to consolidate_
> - Types 
>   - [Built-in types](https://dart.dev/guides/language/language-tour#built-in-types)
>   - Collection types (will include [lists](https://dart.dev/guides/language/language-tour#lists), [maps](https://dart.dev/guides/language/language-tour#maps), and records (new))
>   - [Typedefs](https://dart.dev/guides/language/language-tour#typedefs) _Not originally under "Types", I added it here to consolidate_
>   - [Generics](https://dart.dev/guides/language/language-tour#generics) _Not originally under "Types", I added it here to consolidate_
> - [Control flow](https://dart.dev/guides/language/language-tour#control-flow-statements)
> - [Exceptions](https://dart.dev/guides/language/language-tour#exceptions)
> - Functions
>   - [Functions](https://dart.dev/guides/language/language-tour#functions)
>   - [Parameters](https://dart.dev/guides/language/language-tour#parameters)
>   - [The main() function](https://dart.dev/guides/language/language-tour#the-main-function)
>   - [Functions as first-class objects](https://dart.dev/guides/language/language-tour#functions-as-first-class-objects)
[...]

PR #4598, opened 2023-02-13
Remove async tutorials from Development side nav

MaryaBelanger:
These are conceptually unrelated to the rest of the development section


parlough:
I think while they don't necessarily make sense as top-level entries under "Development", it's a bit of shame the streams tutorial won't have a home in the sidenav with this change though.

We have other tutorials under Development, such as interop and `dart:html` ones, so I wonder if an "Asynchronous programming" dropdown would make sense under "Development". It could contain these two pages and in the future an updated Zones guide. Or we could place the Streams article somewhere else in the sidenav for now until we have a better idea of our desired IA. 

Async concepts are tricky to place as they are heavily defined in the SDK libraries but are deeply linked to some language features as well.

MaryaBelanger:
I guess this is actually a little premature, and I did it based on some context/assumptions about the entire sitemap plan that I haven't really confirmed with anyone yet. 

I'm leaning heavily towards asynchronous programming topics all being under Language, like I proposed in the [language umbrella issue](https://github.com/dart-lang/site-www/issues/4588#:~:text=Futures%2C%20async%2C%20await-,Streams,-Isolates%20previously%20%22Concurrency). I feel like anything that comes up in the logic of the actual code you're writing is a language topic. Development, packages, tools and techniques are all topics that come after, and are for the most part agnostic to, the logic the code itself.

So maybe the right thing to do is just hold off on this until that Async section gets created under Language, then these can move there. 

Side note, though, that we really shouldn't have the same pages showing up twice in the side nav in different sections. But that's more dependent on how the Samples & tutorials section gets reworked. I think, like [flutter.dev](https://docs.flutter.dev/codelabs), tutorials and codelabs like these two pages won't expand in the side nav, there will just be a page that summarizes all of them. But they will show up in the side nav under the topic umbrella they're related to (e.g. this [tutorial](https://docs.flutter.dev/development/ui/layout/tutorial))

parlough:
Sounds like it might be better to make this change alongside that then, so these aren't disappearing or moving around too much. Staying in this spot a little longer shouldn't hurt.

I'll make a note so we can further discuss language vs library more and where/how that distinction should be made in the docs.

MaryaBelanger:
Ok, let's leave the on hold label up for now, but I might just end up closing it later on in favor just doing this move in the PR where I move the other async content out of the language tour as well. And yeah @parlough lets make sure we go over language vs. library features in our next meeting! Thanks everyone

MaryaBelanger:
Done in #5367

PR #4605, opened 2023-02-18
Update link in native extension docs

fuzzybinary:
I've deprecated the old embedding example for one that utilizes .dlls / .sos with a simpler interface. Updating the link to point to that repo instead.


parlough (approved):
Makes sense to me, cool tooling and sample! Thanks :)

PR #4606, opened 2023-02-20
Add section for extension without name

adithyank:


parlough:
Thanks so much for opening a pull request!

I do wonder if we need a whole usage example of an unnamed extension, since it is used in the same way as normal extensions which is already documented elsewhere on the page. We also already document the library-private behavior in the [Implementing extension methods](https://dart.dev/guides/language/extension-methods#implementing-extension-methods) section on [L215-L218](https://github.com/dart-lang/site-www/blob/main/src/_guides/language/extension-methods.md?plain=1#L215-L218) of the doc.

Do you think making that text more noticable would be sufficient? Or we could also add a small example in that section so both types of declarations are together in that section, avoiding duplication.

adithyank:
> Thanks so much for opening a pull request!
> 
> I do wonder if we need a whole usage example of an unnamed extension, since it is used in the same way as normal extensions which is already documented elsewhere on the page. We also already document the library-private behavior in the [Implementing extension methods](https://dart.dev/guides/language/extension-methods#implementing-extension-methods) section on [L215-L218](https://github.com/dart-lang/site-www/blob/main/src/_guides/language/extension-methods.md?plain=1#L215-L218) of the doc.
> 
> Do you think making that text more noticable would be sufficient? Or we could also add a small example in that section so both types of declarations are together in that section, avoiding duplication.



Yes. Making that text more noticeable will be sufficient. Just adding an example will make it noticeable.

parlough:
Take a look at https://github.com/dart-lang/site-www/pull/4616 where I've modified the existing content around unnamed extensions, included a simple declaration example, and documented some potential concerns. I've included you as a co-author. Thanks again!

PR #4607, opened 2023-02-21
Make migration version support more clear

parlough:
Now that the alpha release is advertised, adding this note to make the lack of the `dart migrate` tool more clear in case someone tries to use it.


mit-mit (approved):
Nice catch!

Issue #2337, opened 2020-04-07
'Extension methods' implies "name" is required, but it is not


matanlurey:
Page URL: https://dart.dev/guides/language/extension-methods
Page source: https://github.com/dart-lang/site-www/tree/master/src/_guides/language/extension-methods.md

Based on this page, it looks like the `<name>` is required, but based on the specification it is not.

For example, this is legal:

```dart
extension on String {
  String get first => this[0];
}

void main() {
  print('Hello'.first); // H
}
```

I wonder if this is worth noting, and if it is, if it worth noting if this implicit extension name is public or private (the spec says private, but @munificent and others seem to have thought it was public). 

/cc @leafpetersen

kwalrath:
iirc, we decided to always show the name because it's usually a bad idea not to include the name (even though that's allowed). @leafpetersen, do you recall?

leafpetersen:
I don't think it's a particularly bad idea to omit the name - we added that ability for a reason.  An extension without a name is private to the library it is defined in and is not visible anywhere else.  If you're using library private extensions, most of the time you don't need a name for the extension at all.   The only reasons you need a name for a local extension are if 1) you wanted to put a static method on the extension, or 2) you needed to use the name to explicitly call the extension (either to override the default resolution or because the invocation is ambiguous).

kwalrath:
In my hazy recollection, the desire to only show the named case might have come from making autocomplete work. Maybe. @filiph do you remember?

filiph:
Sorry, I can't remember. FWIW, in the video, I start with unnamed extensions, and only mention named extensions at the very end.

kwalrath:
@mit-mit might remember. I looked at the PR (#2155) but there's nothing there.

iirc, if you don't specify a name, a default one is assigned, so technically no extension is unnamed — it just might not be explicitly named.

Anyhoo, let's wait until mit gets back from Danish holidays and can weigh in.

mit-mit:
Using a name *seems* more idiomatic to me, but not sure... @munificent is this something the style guide should cover?

munificent:
Probably at some point, yes. But I don't think we have enough real-world usage to distill what the best practice is yet. Until then, it's reasonable to let people try different styles so we have enough genetic diversity to evolve to the optimal practice.

PR #4616, opened 2023-02-23
Make support for unnamed extensions more clear and explicit

parlough:
Closes https://github.com/dart-lang/site-www/issues/2337


src/_guides/language/extension-methods.md
```diff
@@ -212,14 +213,28 @@ extension NumberParsing on String {
...
+<?code-excerpt "extension_methods/lib/string_extensions/string_apis_unnamed.dart (unnamed)"?>
+```dart
+extension on String {
+  bool get isBlank => trim().isEmpty;
+}
+```
+
+Note that without a name,
+an extension's static members can only be invoked
+within the extension declaration.
```
atsansone, line 236: **nit:** Starting a sentence with note should either not be done or included in a note callout. If this doesn't rise to that occasion, exclude it.
```suggestion
Without a name, you can invoke an extension's
static members only within the extension declaration.
```

**question:** Is this not redundant with what's written in [lines 221 to 225](https://github.com/dart-lang/site-www/pull/4616/files#diff-3d07c88e0a0b6c36d93bad9dccd5211a89f70f70f94bedd205a61413a6b932d1R221-R225)?

atsansone (commented):
@parlough : Two small issues. Should result a quick update.

Great work!

src/_guides/language/extension-methods.md
parlough: As for its relation to L221-225, it's a different class of limitation that is less likely to be come across.

src/_guides/language/extension-methods.md
```diff
@@ -212,14 +213,28 @@ extension NumberParsing on String {
...
+The members of an extension can be methods, getters, setters, or operators.
 Extensions can also have static fields and static helper methods.
 
+### Unnamed extensions
+
+You can optionally omit the name when declaring an extension.
+An unnamed extension is visible only
+in the library where it's declared.
+You also can't apply it explicitly
+to resolve [API conflicts](#api-conflicts).
```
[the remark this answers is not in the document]
parlough: Thanks for feedback. I've decided to rewrite the API conflict sentence to provide some extra context.

I needed to keep the word "explicitly" since its a specific type of usage mentioned above, in explicitly using the name to apply an extension rather than applying it directly through the underlying type. 

Let me know what you think of the adjusted text.

parlough:
Staged: https://dart-dev--pr4616-feature-unnamed-exte-w5el6qxg.web.app/guides/language/extension-methods#unnamed-extensions

domesticmouse (approved):

src/_guides/language/extension-methods.md
```diff
@@ -212,14 +213,30 @@ extension NumberParsing on String {
...
 Extensions can also have static fields and static helper methods.
 
+### Unnamed extensions
+
+When declaring an extension, you can omit the name.
+Unnamed extensions are visible only
+in the library where they're declared.
+Since they don't have a name,
+there is no way to explicitly apply them
+to resolve [API conflicts](#api-conflicts).
```
atsansone, line 226: **cleanup:** This still needs active voice to clean it up.
```suggestion
When declaring an extension, you can omit the name.
An unnamed extension can be used only in the library
where it's declared.
Since they don't have a name,
they cannot be explicitly applied
to resolve [API conflicts](#api-conflicts).
```

**note:** If you say _visible_, you need to say visible to _whom_.

atsansone (approved):
LGTM % one tweak. Change as needed minding active voice. No need for another pass.

src/_guides/language/extension-methods.md
parlough: I'm rather hesitant to lose the word visible here, "visible inside" or "visible in" a library, package, module, etc. is a common construct in programming languages and their docs, including ours.

atsansone: @parlough : What does "visible" mean? Visible to whom? What does it mean to not be visible? That's what I'm trying to understand.

parlough: @atsansone By "visible in a library", I mean visible to other code (at least) within the same library, visible to the analyzer and compiler(s) when determining what is accessible in the library, in some-cases visible to reflective accessors of the library through `dart:mirrors`.

If it's visible only to that library (private), it won't show up in the `dart doc` output for that library, other libraries can't access the extension, the analyzer won't suggest it when coding in other libraries.

If it's not private, that means other libraries can import it making its members will be visible and accessible, `dart doc` can include the member in its rendered output, etc.

Essentially, I do think "can be used only in the library where it's declared" covers the primary use case of a developer coding, but it's also about whether its discoverable, included in various outputs, etc.

PR #4618, opened 2023-02-25
Annotate Dart code block as Dart, not Swift

alurm:
Dart code block is annotated as Swift code, which is incorrect. Fix this by annotating Dart code block as Dart code


parlough (approved):
Nice catch, thanks!

Issue #4573, opened 2023-02-03
[PAGE ISSUE]: 'Asynchronous programming: futures, async, await'


DemX86:
### Page URL

https://dart.dev/codelabs/async-await#example-introducing-futures

### Page source

https://github.com/dart-lang/site-www/tree/main/src/codelabs/async-await.md

### Describe the problem

This explanation of the printing order can be confusing:
> In the preceding example, even though `fetchUserOrder()` executes before the print() call on line 8, the console shows the output from line 8(“Fetching user order…”) before the output from `fetchUserOrder()` (“Large Latte”). **This is because `fetchUserOrder()` delays before it prints “Large Latte”.**

The beginner reading this text can think that the key is in that `Future.delayed()` method but even if we remove the obvious delay we still will have the same behavior and output:
```
Future<void> fetchUserOrder() {
  return Future(() => print('Large Latte'));
}

void main() {
  fetchUserOrder();
  print('Fetching user order...');
}
```

We know that the real reason why `print('Fetching user order...')` executes before `print('Large Latte')` from the Future is that first one in synchronous (so it runs first) and the second one firstly goes to the Event Queue and then run after all synchronous code is executed.

### Expected fix

I don't know how to make it better — maybe it is too much for beginner to met the event loop mechanism from very beginning of that codelab, but for sure the current explanation can lead to misunderstanding of the principles how asynchrony works

atsansone:
Schedule after https://github.com/dart-lang/site-www/issues/4588 completes.

PR #4626, opened 2023-03-02
doc: update future example explanation

jasimameen:
issue: #4573


parlough:
I'm so sorry about the super late response. I thought I responded to this PR already, but I must have mixed it up with the other one.

Thanks for wanting to improve this, this is a challenging issue to solve without going into too many technical details. I have some comments that might need some extra discussion.

From the original issue, I think the primary concern is that the use of `Future.delayed` is distracting from the reason why "Fetching user order..." is printed out first. As mentioned in the issue, one way to solve this is by switching to the default `Future` constructor. I think that's part of the change that can be made. 

As for explaining it, I think you have some great improvements, but it would be useful for readers if it was more brief or was separated into steps, similar to an earlier explanation on the page (https://github.com/dart-lang/site-www/blob/main/src/codelabs/async-await.md?plain=1#L88-L99). 

Here's a possible structure and potential text which assumes that `fetchUserOrder` is modified to not use `Future.delayed`:

```
In the preceding example, even though `fetchUserOrder()` 
is called before the `print()` call on line 8, 
the console shows output from line 8 ("Fetching user order...") 
before the output from `fetchUserOrder()` ("Large Latte").

This occurs in the following steps:

1. The fetchUserOrder function creates a `Future` containing
   a function that prints "Large Latte" when executed.
2. Dart doesn't execute the function immediately upon creating the Future.
   Instead, it schedules the function for
   asynchronous execution at a later time.
3. Control returns to the main function, which doesn't
   wait for the `Future` returned by `fetchUserOrder()` to complete.
   It proceeds to the next statement and
   immediately prints "Fetching user order...".
4. As soon as possible after the `print()` call in the main function, 
   Dart executes the scheduled function from the `Future`, 
   and prints "Large Latte" to the console.
```

Some notes about this text:

1. It tries to maintain the introduction of what actually happens so the reader can keep that in mind while they build their understanding following the steps, rather than pushing some of it to the end. This also keeps symmetry with the opening paragraph. 
2. It breaks the logic flow into a few steps, hoping to make the text easier to follow.
3. It tries to be specific about the idea of the task being "scheduled". We're trying to avoid introducing the event loop here, as it's out of the scope of this codelab, but my understanding is the `Future()` constructor is essentially scheduling the specified callback on the event loop to be ran later.
4. Avoids [passive voice](https://developers.google.com/style/voice) where possible, in favor of active voice.
5. It attempts to introduce some semantic line breaks (for future editing and easier code review). You can read more about this practice [in our docs](https://github.com/dart-lang/site-shared/blob/main/doc/writing-for-dart-and-flutter-websites.md#semantic-line-breaks).

Let me know what you think or if you have any ideas, happy to discuss more and create a good explanation together! Also if you have any questions about working with the site, such as [updating the code samples](https://developers.google.com/style/voice), let me know. Thanks again for working on this and sorry again about the delay! I'll make sure to not lose track next time.

parlough:
I'm going to close this for now, but if you'd like to revisit it, please feel free to reopen or open a new pull request.

Thanks so much!

PR #4628, opened 2023-03-03
Improve the archives build tool

AlexV525:
* The change to `tool/get-dart/archive/compile.sh` is intended to `cd` into the script directory to make paths relatively stable regardless of where the tool has been run from.
* Changes to Dart files are making the storage base can be configured easily.

This also helps to build [the automatic compilation of dart.cn](https://github.com/cfug/dart.cn/blob/master/.github/workflows/compile_archive_js.yml).

### DIFF
--- a/tool/get-dart/archive/compile.sh
+++ b/tool/get-dart/archive/compile.sh
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+cd -P -- "$(dirname -- "${BASH_SOURCE[0]}")"
 
 : ${TMP:=tmp}
 OUT_DIR="$TMP/download_archive"
--- a/tool/get-dart/archive/compile.sh
+++ b/tool/get-dart/archive/compile.sh
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+# Enter the tool's directory as the working directory to avoid path conflicts under different environments.
 cd -P -- "$(dirname -- "${BASH_SOURCE[0]}")"
 
 : ${TMP:=tmp}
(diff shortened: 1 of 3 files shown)


tool/get-dart/archive/compile.sh
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+cd -P -- "$(dirname -- "${BASH_SOURCE[0]}")"
```
parlough, line 2: Can you add a short comment above this explaining what the line does?

AlexV525:
```suggestion
# Enter the tool's directory as the working directory to avoid path conflicts under different environments.
cd -P -- "$(dirname -- "${BASH_SOURCE[0]}")"
```

tool/get-dart/dart_sdk_archive/lib/src/version_selector.dart
```diff
@@ -9,7 +9,7 @@ import 'package:sdk_builds/sdk_builds.dart';
 
 import 'operating_system.dart';
 
-const _storageBase = 'https://storage.googleapis.com/dart-archive';
+const _storageBase = '${storageBaseURL}dart-archive';
```
parlough, line 12: Fix this to match your declaration:

```suggestion
const _storageBase = '${storageBaseUrl}dart-archive';
```

tool/get-dart/archive/compile.sh
```diff
@@ -1,4 +1,6 @@
 #!/usr/bin/env bash
+# Enter the tool's directory as the working directory to avoid path conflicts under different environments.
```
parlough, line 2:
```suggestion
# Enter the tool's directory as the working directory
# to avoid path conflicts under different environments.
```

parlough (changes requested):
Thanks! One fix and one minor suggestion:

Issue #4622, opened 2023-03-01
Create site-wide "Glossary"


MaryaBelanger:
### Summary

dart.dev needs a resource for definitions of frequently used terminology. In many cases, it doesn't make sense to define a term in the context it's being mentioned in, because it's derailing or doesn't fit the purpose of that doc. Content like what we have in the language tour faces this issue a lot, since it usually conveys simplified and straightforward "quickstart" types of information per topic; definitions of peripheral terms don't have a place. 

This is why the diagnostics page has it's own glossary section, because many of it's messages need to refer to terms that they don't have the space to define in a diagnostic message (nor should they have to do so, the purpose of a diagnostic message doesn't lend to defining every term inline). 

This is coming up now because there are new Dart 3 terms the diagnostic messages need defined and linked to in the main docs. 

### Evaluation

1. The new page will go under Resources >.Glossary

2. The contents of the Glossary section of the diagnostic messages page will be removed from that page and be the basis of the new Glossary page.

3. Over time, more terminology will be added to the new Glossary page, so other pages can link to its definitions. New Dart 3 definitions will be placed there.

4. The goal is that anywhere in the docs where a user can benefit from having the definition of a term handy, we will add links on the terms to the glossary.

5. A concern is that the glossary might become a dumping ground for information that should really be covered elsewhere in the documentation. Every new entry in the glossary needs to be heavily scrutinized to ensure
    - It's truly just a _definition_, not a full blown reference or example
    - There isn't a better existing location for that information. 

6. Need to go through the diagnostic messages (`analyzer/tool/diagnostics/generate.dart`) and add links to the new page anywhere one of the terms currently in the diagnostics glossary is used, so that the diagnostic messages page doesn't suffer from not having the definitions on the same page anymore

#### TBD

7. The Packages section has its own fully-formed, very involved, glossary. Currently I don't have any plans to remove that content from the packages section and incorporate it into the new glossary, but that might make sense in the future.

8. (stretch) @parlough has an idea to develop some tooling where you can hover over a word in the docs that's linked to a definition in the glossary, and a small window/blurb will open over the word showing it's definition, plus a link to it in the glossary in case the full definition is too long for the tiny pop up.  

### References

- [the diagnostic messages glossary](https://dart.dev/tools/diagnostic-messages#glossary)
- [the package content's glossary](https://dart.dev/tools/pub/glossary)

### Related/ToDo

[...]

PR #4630, opened 2023-03-03
Create "Glossary" page for diagnostics to link to 

MaryaBelanger:
Part of #4622 

This is a first PR to add a glossary page, specifically with definitions for the new patterns-related terms "refutable" and "irrefutable". New Dart 3 diagnostic messages need to point to these definitions, so I'm adding them now so the diagnostics can be committed: [`286524`](https://dart-review.googlesource.com/c/sdk/+/286524)

The page will remain hidden from the sitemap, though, until the remainder of the glossary content from https://dart.dev/tools/diagnostic-messages can move there too (it will eventually be added under "Resources" in the side nav).

That should only happen after the new diagnostics are complete, so we can simultaneously change all the pointers within the diagnostics to point to the new page.

### DIFF
--- a/src/resources/glossary.md
+++ b/src/resources/glossary.md
@@ -1,0 +1,19 @@
+---
+title: Glossary
+description: A glossary reference for terminology used across dart.dev
+sitemap: false
+---
+
+The following are definitions of terminology used across the Dart documentation.
+
+## Irrefutable patterns
+
+_Irrefutable patterns_ are patterns that always match. 
+Irrefutable patterns appear in _irrefutable contexts_: 
+_declaration context_ and _assignment context_.
+
+## Refutable patterns
+
+A _refutable pattern_ is a pattern that can be tested against a value to
+determine if the pattern matches the value. If not, the pattern _refutes_ the value.
+Refutable patterns appear in _matching contexts_.


src/resources/glossary.md
```diff
@@ -0,0 +1,19 @@
+---
+title: Glossary
+description: A glossary reference for terminology used across dart.dev
```
parlough, line 3: We tend to end descriptions with a period.

```suggestion
description: A glossary reference for terminology used across dart.dev.
```

src/resources/glossary.md
```diff
@@ -0,0 +1,19 @@
...
+sitemap: false
+---
+
+The following are definitions of terminology used across the Dart documentation.
+
+## Irrefutable patterns
```
parlough, line 9: I wonder if these entries should be singular? Seems most of the ones in the [Fuchsia Glossary](https://fuchsia.dev/fuchsia-src/glossary#terms) are.

```suggestion
## Irrefutable patterns
```

src/resources/glossary.md
```diff
@@ -0,0 +1,19 @@
...
+
+_Irrefutable patterns_ are patterns that always match. 
+Irrefutable patterns appear in _irrefutable contexts_: 
+_declaration context_ and _assignment context_.
+
+## Refutable patterns
+
+A _refutable pattern_ is a pattern that can be tested against a value to
+determine if the pattern matches the value. If not, the pattern _refutes_ the value.
```
parlough, line 18: It doesn't necessarily need to be addressed now, but defining "Refutable patterns" with "refute" seems a bit tricky to me. I'm not sure most readers will know what refute means.

parlough (commented):
It begins!

I plan to make this page generated in the future, but since the diagnostics want to point to here, this should be good for now. A few comments :)

src/resources/glossary.md
```diff
@@ -0,0 +1,19 @@
...
+description: A glossary reference for terminology used across dart.dev
+sitemap: false
+---
+
+The following are definitions of terminology used across the Dart documentation.
+
+## Irrefutable patterns
+
+_Irrefutable patterns_ are patterns that always match. 
+Irrefutable patterns appear in _irrefutable contexts_:
```
bwilkerson, line 12: To me, "appear" implies "can only appear", but irrefutable patterns can also appear in a matching context. For example
```
void f((int, int) r) {
  switch (r) {
    case (0, var y):
      // `y` is bound to the second field if the first field is zero (0).
  }
}
```

Consider "Irrefutable patterns appear" --> something like "Only irrefutable patterns can appear"

src/resources/glossary.md
```diff
@@ -0,0 +1,19 @@
...
+
+_Irrefutable patterns_ are patterns that always match. 
+Irrefutable patterns appear in _irrefutable contexts_: 
+_declaration context_ and _assignment context_.
+
+## Refutable patterns
+
+A _refutable pattern_ is a pattern that can be tested against a value to
+determine if the pattern matches the value. If not, the pattern _refutes_ the value.
+Refutable patterns appear in _matching contexts_.
```
bwilkerson, line 20: Do we want to enumerate the matching contexts?

Do we want to provide examples of a subset of such contexts (both here and for irrefutable contexts)?

bwilkerson:
> I plan to make this page generated in the future ...

Completely unrelated to the PR, but out of curiosity, generated from where / what?

parlough:
@bwilkerson The Fuchsia docs have a glossary system which can be filtered, link to more details/other resources, have optional longer definitions, and can be used across the site for clickable links with tooltips. This is generated from this [glossary.yml](https://fuchsia.googlesource.com/fuchsia/+/refs/heads/main/docs/glossary/glossary.yaml) file.

I'm planning on implementing some sort of similar system in the future, but I have some other work to finish first.

The current anchor references here should continue to work fine once we make that transition.

parlough (approved):
Thanks for those adjustments!

Looks good to me if @bwilkerson is happy with the definitions :)

MaryaBelanger:
_I'm_ not thrilled with these definitions myself 😅 right now they're place holders for the links. I'm hoping as I progress in my work on patterns I'll find a more useful way to present these concepts.

parlough:
That's what I imagined! I expect your future patterns work will result in good related areas these can link to as well.

bwilkerson (approved):

src/resources/glossary.md
```diff
@@ -0,0 +1,19 @@
+---
+title: Glossary
+description: A glossary reference for terminology used across dart.dev
+sitemap: false
+---
+
+The following are definitions of terminology used across the Dart documentation.
```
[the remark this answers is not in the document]
parlough: Good point. The [Google style guide](https://developers.google.com/style/word-list) uses "terms" extensively.

bwilkerson:
Yes, I'm fine with it landing as is. Not sure what happened, but I tried to say that earlier but while my review comments appear to have been posted, the approval didn't get recorded the first time around.

src/resources/glossary.md
MaryaBelanger: I was thinking maybe "Patter contexts" could be a glossary section too, or maybe each individual pattern context. That kind of section would be nice to enumerate and provide examples. I'll make  a note with the ret of my patterns plan to return to this.

PR #4612, opened 2023-02-22
Break up the Language Tour

MaryaBelanger:
Staged: https://dart-dev--pr4612-language-tour-2-koullabo.web.app/

---

_Waiting for #4604 to merge first -- Fixes #4588_

---

This PR breaks sections of the language tour into separate pages and lists them directly under "Language" in the side nav, so the features of the Dart language are visible and navigable more easily than the previous single page. 

_(see #4588 and #4565 for background; see [Kotlin](https://kotlinlang.org/docs/basic-types.html) and [Swift](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/thebasics/) docs for inspiration/direction)_

## What this PR _is_:

- **Addressing the side nav and new organization.** Please comment and review the new layout of the side nav under "Language".
  
  image

- **Addressing the grouping of info on the new pages.** Please let me know if the new page groupings make sense. They almost entirely follow the structure of original Language Tour -- I didn't write that, I won't be improving that structure in this PR beyond what's necessary.
- Fixing links to the old language tour page. 
- Adding manual redirect links to old language tour page

### TODO: 
In this PR...
- [x] Manually add links to old language tour page, in case external links end up there
  - Link on every section to its new page/location  
- [x] Search the entire site source for internal links to `/language-tour`/`#<section>` to change to new page/location
- [x] Add ignore thing from language-tour for code excerpts
- [ ] Figure out "<< Previous" / "Next >>" buttons for bottom of each page
  - Parkers going to write a script after this lands
- [x] Prevent old `/language-tour` page from showing up in search results 
- [ ] Add a little introduction about the language tour to intro page
- [ ] Firebase redirects after this PR

---

## What this PR _is not_:

This PR is not...
- **Addressing content.** 
  - Please do not review the contents of any of the new pages, I didn't write it. Open a separate issue for... 
    - redundancy
    - sub-optimal information flow _within a page_ (page-to-page is in scope for the PR)
    - inaccurate/outdated content
    - etc.
- **Major rewrite.** 
  - Splitting up or rearranging certain content requires major rewrites or at minimum analysis of the existing content; that will not happen as part of this PR.
  - Referring to the [Type System](https://dart.dev/guides/language/type-system) page and all the [Null Safety content](https://dart.dev/null-safety), which can and should be split up into more granular topic by page, but it's not straight forward (see TODO below)
- **Creating a "Getting started"** or de facto intro page.
[...]

PR #4641, opened 2023-03-10
Revert "Break up the Language Tour (#4612)"

parlough:
Reverts dart-lang/site-www#4612

I didn't have a chance to review changes beyond the structure yet and there are a few broken changes, broken DartPads, missing redirects, etc.

I can begin work on fixing and implementing those missing changes tonight, but I don't think it's a good idea to leave links on search engines broken and broken or misleading docs in the mean time.


parlough:
Actually I’ve just reverted to an old deploy and will hold off on new changes until the fixes last.

PR #4645, opened 2023-03-11
Rename 'Collection types' to 'Collections'

parlough:
Blocked on https://github.com/dart-lang/site-www/pull/4647

"Collection types" makes me think of the generic type of a collection rather than collections themselves.

Overall, I think "Collections" is more intuitive. I suppose it could be "Built-in collections" but let me know what you think.

### DIFF
--- a/src/_guides/language/language-tour.md
+++ b/src/_guides/language/language-tour.md
@@ -59,15 +59,15 @@ This content has moved to [Collections](/language/collections#lists).
 <a id="trailing-comma"></a>
 #### Trailing commas
 
-This content has moved to [Collection types](/language/collection-types#trailing-commas).
+This content has moved to [Collections](/language/collections#trailing-commas).
 
 #### Spread operator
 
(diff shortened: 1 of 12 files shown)


src/_guides/language/language-tour.md
```diff
@@ -54,16 +54,30 @@ This content has moved to [Built-in types](/language/built-in-types#booleans).
 
 ### Lists
 
-This content has moved to [Collection types](/language/collection-types#lists).
+This content has moved to [Collections](/language/collections#lists).
+
+<a id="trailing-comma"></a>
```
domesticmouse, line 59: I feel like this change is from another PR...

parlough: Yep, I based this one on https://github.com/dart-lang/site-www/pull/4647 since I knew that would land first and they affect some of the same content.

MaryaBelanger (approved):
Thanks, "Collections" seems much better!

PR #4651, opened 2023-03-13
Implement client side redirection from old language tour

parlough:
Since fragments aren't sent to the server for redirection, this client-sidescript handles users landing on the old language tour, and redirects them based on their current fragment to the new location. The logic is as follows:

1. If the user isn't on the language tour page, do nothing.
2. If the user didn't specify a fragment, send them to the [Dart introduction page](https://dart.dev/language).
3. If the fragment they specified is empty or just whitespace, send them to the [Dart introduction page](https://dart.dev/language).
4. If the fragment they specified was not on the page previously, send them to the [Dart introduction page](https://dart.dev/language).
5. Check with an HTTP request to see if the fragment they specified maps to a real, existing page. If it does, redirect them to that page. **!!!! This step may not be necessary, but let me know what you think :)**
6. If the page mapped to does not exist or is not accessible, send them to the introduction page.

For users without JavaScript enabled, the textual-based redirection will still be present.

**TESTING NOTE:**

**Please** test that this works as expected for you:

1. Use the staged language tour link: https://dart-dev--pr4651-feature-old-language-9w6zvjjc.web.app/guides/language/language-tour
2. Append one of the headers/fragments from the language tour page and verify it redirects you to the correct page
3. Try for multiple different headers/fragments
4. Verify situations which don't pass the above situations go to the introduction page (no fragment, empty fragment, incorrect fragment)


domesticmouse (approved):

MaryaBelanger (approved):
This is perfect! Tested everything, except I wasn't sure how to test 5?

> Check with an HTTP request to see if the fragment they specified maps to a real, existing page. If it does, redirect them to that page.

But either way, approved!

parlough:
@MaryaBelanger If the others worked as expected, that means 5 did as well.

Thanks so much for helping me testing all of that!

MaryaBelanger:
Great, well it _sounds_ like a useful thing to have, so I vote keep it even if it's not necessary :)

PR #4653, opened 2023-03-13
Move 'Generators' to 'Functions' for now

parlough:
It's not currently linked to from the sidenav, so it isn't really accessible.

Since we have all the function content in one place, there isn't really a great place to put it in the sidenav. For now, we can put it at the bottom of the functions page since it is less commonly used. What do you think?


MaryaBelanger (approved):
Definitely agree. Looks like putting Generators at the end of Functions was part of the [original plan](https://github.com/dart-lang/site-www/issues/4588#:~:text=Return%20values-,Generators,-Not%20originally%20under); I must've lost track of it when actually doing the restructure. So, thanks for catching that!

PR #4655, opened 2023-03-13
Cleanup metadata article

parlough:
- Remove TOC since it is empty
- Update description to include "Annotations" since that's what most people will likely search for
- Add extra line to end of file


MaryaBelanger (approved):
🙏

PR #4656, opened 2023-03-13
Misc cleanup to new language pages

parlough:
- Adds missing new lines at end of files
- Removes TOC ("Contents") from pages with no TOC entries


MaryaBelanger (approved):
Thanks Parker!

PR #4658, opened 2023-03-13
Set canonical link for old language tour

parlough:
As suggested by https://developers.google.com/search/docs/crawling-indexing/consolidate-duplicate-urls when you want to direct search engines to a preferred version of some content.

Also uses `noindex` as suggested by https://developers.google.com/search/docs/crawling-indexing/block-indexing to block indexing of the language tour.

You can test by going to the [old language tour](https://dart-dev--pr4658-feature-old-language-10qfdojm.web.app/guides/language/language-tour) and verifying the new `<meta` and `<link` are present in the `<head>`. Then that they don't exist in other pages, such as the [new language introduction](https://dart.dev/language).


domesticmouse (approved):

MaryaBelanger:
Thanks so, so, sooooo much for taking care of this @parlough !

PR #4665, opened 2023-03-14
Update for latest gem releases

parlough:


atsansone (changes requested):
Two questions:

1. Why are we modifying the `Gemfile.lock` file? That's generated automatically from the `Gemfile`.
2. Do we have any reservations about updating any of these Gems? Do we count on the internal testing to make sure they work?

parlough:
@atsansone 

1. It's generated from the ranges allowed in `Gemfile`, yes, but we have to tell bundler to generate the lock file. The build only uses the dependencies specified in the `Gemfile.lock`, to prevent changes being included we don't want, such as those having undesired or breaking changes or even security vulnerabilities. It's generally a good practice to use the latest dependencies though, for latest security fixes, performance improvements, new functionality, etc. So occasionally, I go through and regenerate the `Gemfile.lock` and verify each dependencies changes and functionality. dependabot can handle some of these updates, but not all of them, and recently has had some issues in how it's updating native dependencies that I haven't had a chance to look into yet.
2. The Ruby dependencies are used during the build, so new build failures would be caught by CI, and other issues by some other checks. When updating dependencies though, I check their changelog or commit log as well as test the changes locally.

atsansone (approved):

PR #4667, opened 2023-03-14
Remove the function name firstWhereWithOrElse() from example

arunkumaryadav87:


src/codelabs/iterables.md
```diff
@@ -250,18 +250,6 @@ There is no right or wrong way.
 Use the way that works best for you, 
 and that makes your code easier to read and understand.
 
-In the example, `firstWhereWithOrElse()` calls `firstWhere()` with
```
parlough, line 253: I assume you're removing this because `firstWhereWithOrElse` doesn't exist in the code above, correct? Nice catch!

Despite that, it seems this text is still useful to explain the final example in the code block above. 

Perhaps instead of referencing `firstWhereWithOrElse` which doesn't exist, we could still refer to that example somehow. Perhaps:

```
The final use of `firstWhere` above specifies
....
```

Let me know what you think. Thanks again!

parlough (changes requested):
Hi @arunkumaryadav87, thanks for the PR. You'll have to [sign the CLA](https://github.com/dart-lang/site-www/pull/4667#issuecomment-1468916052) with both the account that created the PR and the account that created the commit before we can take a look or include this change.

I have a comment as well, let me know what you think or if you have any issues.

src/codelabs/iterables.md
arunkumaryadav87: It exists the issue is I have accidentally committed using my office git hub account and when pushed branched used personal account

PR #4669, opened 2023-03-14
Remove the function name firstWhereWithOrElse() from example

arunkumaryadav87:
firstWhereWithOrElse() function doesn't exist so I've removed the following example from site.

### DIFF
--- a/src/codelabs/iterables.md
+++ b/src/codelabs/iterables.md
@@ -250,6 +250,18 @@ There is no right or wrong way.
 Use the way that works best for you, 
 and that makes your code easier to read and understand.
 
+In the example, `firstWhere()` with
+the optional named parameter `orElse`,
+which provides an alternative when an element isn't found.
+In this case, the text `'None!'` is returned because
+no element satisfies the provided condition.
+
+{{site.alert.note}}
+  If no element satisfies the test predicate and
+  the `orElse` parameter isn't provided,
+  then `firstWhere()` throws a [StateError.][StateError class]
+{{site.alert.end}}
+
 {{site.alert.secondary}}
   **Quick review:**
   * The elements of an `Iterable` must be accessed sequentially.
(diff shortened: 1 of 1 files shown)


src/codelabs/iterables.md
```diff
@@ -250,18 +250,6 @@ There is no right or wrong way.
 Use the way that works best for you, 
 and that makes your code easier to read and understand.
 
-In the example, `firstWhereWithOrElse()` calls `firstWhere()` with
```
parlough, line 253: I assume you're removing this because `firstWhereWithOrElse` doesn't exist in the code above, correct? Nice catch!

Despite that, it seems this text is still useful to explain the final example in the code block above. 

Perhaps instead of referencing `firstWhereWithOrElse` which doesn't exist, we could still refer to that example somehow. Perhaps:

```
The final use of `firstWhere` above specifies
....
```

Let me know what you think. Thanks again!

parlough (commented):
Thanks for signing the CLA and again for opening the PR.

My comment for discussion still stands. Let me know what you think.

I also recommend in general, when opening a pull request, include a justification/explanation in the description. It helps maintainers know why the change is being made and how they should approach the review :)

Thanks again!

src/codelabs/iterables.md
arunkumaryadav87: firstWhere() function is already explained before firstWhereWithOrElse(). I feel no point of duplicating information. Thanks for your comment.

parlough: This text isn't explaining `firstWhere` though, it is specifically explaining the final example of it, and what the `orElse` optional-named parameter does and how it works. That doesn't seem to be explained earlier in the document.

```dart
var foundItem4 = items.firstWhere(
    (item) => item.length > 10,
    orElse: () => 'None!',
);
```

parlough (approved):
Thanks for noticing and fixing this!

Issue #4639, opened 2023-03-09
Update RISC-V statement


mit-mit:
### Page URL

https://dart.dev/get-dart

### Describe the problem

When 3.0 stable is available, we should update the RISC-V statement from:
_"Support for RISC-V is experimental, and is available only in the dev channel."_ 

to:
_"Support for RISC-V is in preview, and is available only in the dev and beta channels."_

PR #4672, opened 2023-03-15
Unmask RV64 beta channel downloads.

rmacnak-google:
Closes https://github.com/dart-lang/site-www/issues/4639


parlough (approved):
Thanks @rmacnak-google! Works as expected :)

PR #4674, opened 2023-03-15
Make '/resources' pages highlight 'Docs' and move some pages

parlough:
Make pages under `/resources` highlight **Docs** in top navbar. Also move the `/faq` and `/dart-team-packages` under `/resources`. We already include `/faq` under `/resources` in the sidenav, so it made sense. Let me know what you think about the new placement though :)

For example: https://dart-dev--pr4674-feature-resources-cl-9ima3feg.web.app/resources/faq will now highlight **Docs** at the top of the page, compared to https://dart.dev/faq which didn't.


MaryaBelanger (approved):
Great, randomly selected 10 or so to test and LGTM. Nice catch on the top nav bar highlighting too

PR #4684, opened 2023-03-17
Updated Language Evolution for chrono descending, plus clean up

atsansone:
This PR updates the Language Evolution page. I made some language updates, but re-sorted the versions as reverse chronological and added release dates and links to the release announcements.

### DIFF
--- a/src/_guides/language/evolution.md
+++ b/src/_guides/language/evolution.md
@@ -8,7 +8,7 @@ This page lists notable changes and additions to the
 Dart programming language.
 
 * To learn specific details about the most recent supported language version,
-check out the [language documentation][] or the [language specification][].
+  check out the [language documentation][] or the [language specification][].
 * For a full history of changes to the Dart SDK, see the [SDK changelog][].
 
 To use a language feature introduced after 2.0,
@@ -28,9 +28,10 @@ environment:
 [language versioning section]: #language-versioning
 
 {{site.alert.tip}}
-  To review the features being discussed, investigated, and added to
-  the Dart language,
-  check out the [language funnel][] tracker in the Dart language GitHub repo.
+  To review the features being discussed, investigated, and
+  added to the Dart language,
+  check out the [language funnel][] tracker
+  on the Dart language GitHub repo.
 {{site.alert.end}}
 
 
@@ -283,8 +284,8 @@ Widget build(BuildContext context) {
 ```
 
 The **[collection for][]** operator enables building repeated elements.
-The following example adds one `HeadingAction` element for each section
-in `sections`:
+The following example adds one `HeadingAction` element for 
+each section in `sections`:
 
 ```dart
 Widget build(BuildContext context) {
@@ -301,8 +302,7 @@ Widget build(BuildContext context) {
 _Released 26 February 2019_
 | [Dart 2.2 announcement](https://medium.com/dartlang/announcing-dart-2-2-faster-native-code-support-for-set-literals-7e2ab19cc86d)
 
-Dart supports literal lists and maps, but Dart 2.2 added support for
-**[set literals][]**:
+Dart 2.2 added support for **[set literals][]**:
 
 ```dart
 const Set<String> currencies = {'EUR', 'USD', 'JPY'};
@@ -330,9 +330,10 @@ padding: const EdgeInsets.symmetric(
 _Released 22 February 2018_
 | [Dart 2.0 announcement](https://medium.com/dartlang/announcing-dart-2-80ba01f43b6)
 
-Dart 2.0 implemented a new **[sound type system][]**. Before Dart 2.0,
-types weren't fully sound, and Dart relied heavily on runtime type
-checking. Dart 1.x code had to be [migrated to Dart 2][].
+Dart 2.0 implemented a new **[sound type system][]**. 
+Before Dart 2.0, types weren't fully sound, and 
+Dart relied heavily on runtime type checking. 
+Dart 1.x code had to be [migrated to Dart 2][].
 
 ## Language versioning
 
@@ -381,7 +382,7 @@ Deriving the language version from the SDK version
 implies the following:
 
 * Whenever a minor version of the SDK ships, a new language version appears.
-  In practice, many of these language versions function in very
+  In practice, many of these language versions function in a very
   similar manner to previous versions with full compatibility.
   For example, the Dart 2.9 language works much like the Dart 2.8
   language.
@@ -395,8 +396,8 @@ implies the following:
 ### Per-library language version selection
 
 By default, every Dart file in a package uses the same language version.
-The pubspec YAML file identifies this language version as the lower SDK
-constraint.
+Dart identifies the default language version as the 
+lower-bound of the SDK constraint specified in the `pubspec.yaml` file.
 Sometimes, a Dart file might need to use an older language version.
 For example, you might not be able to migrate all the files in a package
 to null safety at the same time.
--- a/src/_guides/language/evolution.md
+++ b/src/_guides/language/evolution.md
@@ -63,20 +63,19 @@ library;
 _Released 30 August 2022_
 | [Dart 2.18 announcement](https://medium.com/dartlang/dart-2-18-f4b3101f146c)
 
-Dart 2.18 enhanced type inference. 
-This change allows information flow between 
-arguments in generic function calls. 
-Before 2.18, if you didn't specify an
-argument's type in some methods, Dart reported errors. 
-These type errors cited potential null occurrences. 
-With 2.18, the compiler infers the argument type 
-from other values in an invocation. 
+Dart 2.18 enhanced type inference.
+This change allows information flow between arguments in generic function calls.
+Before 2.18, if you didn't specify an argument's type in some methods,
+Dart reported errors.
+These type errors cited potential null occurrences.
+With 2.18, the compiler infers the argument type
+from other values in an invocation.
 You don't need to specify the argument type inline.
 
 Dart 2.18 also discontinued support for mixin classes that don't extend
 `Object`.
 
-To learn more about Dart 2.18, check out:
+To learn more about these features, check out:
 
 * [Type argument inference][]
 * [Adding features to a class: mixins][]
@@ -85,7 +84,7 @@ To learn more about Dart 2.18, check out:
 [Adding features to a class: mixins]: /language/mixins
 
 ### Dart 2.17
-_Released 11 May 2022_ 
+_Released 11 May 2022_
 | [Dart 2.17 announcement](https://medium.com/dartlang/dart-2-17-b216bfc80c5d)
 
 Dart 2.17 expanded enum functionality with enhanced enums.
@@ -111,7 +110,7 @@ void main() {
 }
 ```
 
-To learn more about Dart 2.17, check out:
+To learn more about these features, check out:
 
 * [Enhanced enums][]
 * [Super parameters][]
@@ -125,8 +124,7 @@ To learn more about Dart 2.17, check out:
 _Released 3 February 2022_
 | [Dart 2.16 announcement](https://medium.com/dartlang/dart-2-15-7e7a598e508a)
 
-Dart 2.16 added no new features to the Dart language, but fixed one
-security vulnerability and made two breaking changes.
+Dart 2.16 added no new features to the Dart language.
 It did expand the Dart tools.
 
 ### Dart 2.15
@@ -144,6 +142,10 @@ Dart 2.14 added the unsigned shift (or _triple-shift_) operator (`>>>`).
 This new operator works like `>>`,
 except that it always fills the most significant bits with zeros.
 
+To learn more about these operators, check out [bitwise and shift operator][].
+
+[bitwise and shift operator]: /language/operators#bitwise-and-shift-operators
+
 Dart 2.14 removed some restrictions on type arguments.
 You can pass type arguments to annotations and use a generic function
 type as a type argument.
@@ -155,10 +157,6 @@ late List<T Function<T>(T)> idFunctions;
 var callback = [<T>(T value) => value];
 late S Function<S extends T Function<T>(T)>(S) f;
 ```
-
-To learn more about these operators, check out [bitwise and shift operator][].
-
-[bitwise and shift operator]: /language/operators#bitwise-and-shift-operators
 
 ### Dart 2.13
 _Released 19 May 2021_
@@ -193,6 +191,10 @@ Dart 2.10 added no new features to the Dart language.
 It did add an expanded [`dart` tool][dart-tool] similar to the
 Flutter SDK's [`flutter` tool][].
 
+### Dart 2.9
+
+Dart 2.9 added no new features to the Dart language.
+
 ### Dart 2.8
 _Released 6 May 2020_
 | [Dart 2.8 announcement](https://medium.com/dartlang/announcing-dart-2-8-7750918db0a)
@@ -209,11 +211,9 @@ _Released 11 December 2019_
 | [Dart 2.7 announcement](https://medium.com/dartlang/dart-2-7-a3710ec54e97)
 
 Dart 2.7 added support for **[extension methods][]**,
-enabling you to add functionality to any type—even types you don’t control—with
-the brevity and auto-complete experience of regular method calls.
-Because the tech preview for this feature was in 2.6,
-you can use extension methods without warnings if you
-specify 2.6.0 or a later release as the lower SDK constraint.
+enabling you to add functionality to any type
+—-even types you don’t control—-
+with the brevity and auto-complete experience of regular method calls.
 
 The following example extends the `String` class from
 `dart:core` with a new `parseInt()` method:
@@ -235,7 +235,23 @@ void main() {
 _Released 5 November 2019_
 | [Dart 2.6 announcement](https://medium.com/dartlang/dart2native-a76c815e6baf)
 
-Dart 2.6 added a new compiler that has since been retired.
+Dart 2.6 introduced a
+[breaking change (dart-lang/sdk#37985)](https://github.com/dart-lang/sdk/issues/37985).
+Inference changes when using Null values in a `FutureOr` context.
+Constraints of the forms similar to `Null <:FutureOr` now yield `Null`
+as the solution for `T`.
+
+For example: The following code now prints `Null`.
+Before Dart 2.6, it printed `dynamic`.
+The anonymous closure `() {}` returns the `Null` type.
+
+```dart
+import 'dart:async';
+
+void foo<T>(FutureOr<T> Function() f) { print(T); }
+
+main() { foo(() {}); }
+```
 
 ### Dart 2.5
 _Released 10 September 2019_
@@ -244,6 +260,24 @@ _Released 10 September 2019_
 Dart 2.5 didn't add any features to the Dart language, but it did add
 support for [calling native C code][] from Dart code
 using a new **core library, `dart:ffi`.**
+
+### Dart 2.4
+_Released 27 June 2019_
+
+
+Dart 2.4 introduces a breaking change
+[dart-lang/sdk#35097](https://github.com/dart-lang/sdk/issues/35097).
+
+Dart now enforces covariance of type variables used in super-interfaces.
+For example: Prior to this release Dart accepted, but now rejects,
+the following code:
+
+```dart
+class A<X> {};
+class B<X> extends A<void Function(X)> {};
+```
+
+You can now use the identifier async in asynchronous and generator functions.
 
 ### Dart 2.3
 _Released 8 May 2019_
@@ -282,7 +316,7 @@ Widget build(BuildContext context) {
 ```
 
 The **[collection for][]** operator enables building repeated elements.
-The following example adds one `HeadingAction` element for 
+The following example adds one `HeadingAction` element for
 each section in `sections`:
 
 ```dart
@@ -329,9 +363,9 @@ padding: const EdgeInsets.symmetric(
 _Released 22 February 2018_
 | [Dart 2.0 announcement](https://medium.com/dartlang/announcing-dart-2-80ba01f43b6)
 
-Dart 2.0 implemented a new **[sound type system][]**. 
-Before Dart 2.0, types weren't fully sound, and 
-Dart relied heavily on runtime type checking. 
+Dart 2.0 implemented a new **[sound type system][]**.
+Before Dart 2.0, types weren't fully sound, and
+Dart relied heavily on runtime type checking.
 Dart 1.x code had to be [migrated to Dart 2][].
 
 ## Language versioning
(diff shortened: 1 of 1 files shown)


atsansone:
Staged at https://dart--staging.web.app/guides/language/evolution

src/_guides/language/evolution.md
```diff
@@ -180,131 +182,157 @@ turn into edit-time analysis errors.
...
-Dart 2.18 also discontinued support for mixin classes that don't extend `Object`.
-To learn more about mixins,
-see [Adding features to a class: mixins][] in the language tour.
+### Dart 2.2
+_Released 26 February 2019_
+| [Dart 2.2 announcement](https://medium.com/dartlang/announcing-dart-2-2-faster-native-code-support-for-set-literals-7e2ab19cc86d)
 
-[Adding features to a class: mixins]: /language/mixins
+Dart supports literal lists and maps, but Dart 2.2 added support for
+**[set literals][]**:
```
parlough, line 305: I think losing the "always supported" makes this a bit more confusing, since Dart 2.2 has already been out for a long time. If this was a release note, this would make more sense. Perhaps just focus on what was added?

```suggestion
Dart 2.2 added support for **[set literals][]**:
```

src/_guides/language/evolution.md
```diff
@@ -180,131 +182,157 @@ turn into edit-time analysis errors.
...
+_Released 15 November 2018_
+| [Dart 2.1 announcement](https://medium.com/dartlang/announcing-dart-2-1-improved-performance-usability-9f55fca6f31a)
 
-* More flow analysis flags for unreachable code cases.
-* No longer delegate inaccessible private names to `noSuchMethod`.
-* Top-level type inference throws on cyclic dependencies.
+Dart 2.1 added support for **int-to-double conversion**, allowing developers to
+set `double` values using integer literals. This feature removed the annoyance
+of being forced to use a `double` literal (for example, `4.0`)
+when the value was conceptually an integer.
```
parlough, line 318: This text was missing some semantic breaks:

```suggestion
Dart 2.1 added support for **int-to-double conversion**, allowing
developers to set `double` values using integer literals. 
This feature removed the annoyance of 
being forced to use a `double` literal (for example, `4.0`)
when the value was conceptually an integer.
```

src/_guides/language/evolution.md
```diff
@@ -180,131 +182,157 @@ turn into edit-time analysis errors.
...
 ```
 
-[and should]: /guides/language/effective-dart/style#dont-explicitly-name-libraries
+### Dart 2.0
+_Released 22 February 2018_
+| [Dart 2.0 announcement](https://medium.com/dartlang/announcing-dart-2-80ba01f43b6)
+
+Dart 2.0 implemented a new **[sound type system][]**. Before Dart 2.0,
+types weren't fully sound, and Dart relied heavily on runtime type
+checking. Dart 1.x code had to be [migrated to Dart 2][].
```
parlough, line 335: Cleaning up the semantic breaks here:

```suggestion
Dart 2.0 implemented a new **[sound type system][]**. 
Before Dart 2.0, types weren't fully sound, and 
Dart relied heavily on runtime type checking. 
Dart 1.x code had to be [migrated to Dart 2][].
```

parlough (changes requested):
I've gone through and left some issues, suggestions, and nits. Please feel free to push back or discuss anything.

src/_guides/language/evolution.md
```diff
@@ -180,131 +182,157 @@ turn into edit-time analysis errors.
...
-For more information, see the
-[bitwise and shift operator][] section of the language tour.
+### Dart 2.8
+_Released 6 May 2020_
+| [Dart 2.8 announcement](https://medium.com/dartlang/announcing-dart-2-8-7750918db0a)
 
-[bitwise and shift operator]: /language/operators#bitwise-and-shift-operators
+Dart 2.8 didn't add any features to the Dart language, but it did
+contain a number of preparatory [breaking changes][2.8 breaking changes]
+to ensure great nullability-related usability and performance in the
```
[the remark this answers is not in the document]
atsansone: **comment:** Fixed.

src/_guides/language/evolution.md
```diff
@@ -313,23 +341,21 @@ multiple versions of the Dart language.
...
-null safety might take a while,
-Dart uses language versioning to support
-using non-null-safe code alongside null-safe code.
-For an example of how an app or package can migrate to
-a new language version
-with an incompatible feature (null safety, for example), see 
+Language versioning becomes important on the rare occasions when Dart
+introduces an incompatible feature like [null safety][].
+Before null safety, code could compile but might crash at runtime.
+With null safety, the code might no longer compile.
```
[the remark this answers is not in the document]
atsansone: **comment:** I think I have a good update. Have a look.

atsansone (commented):
@parlough : Merging my changes after this review.

src/_guides/language/evolution.md
```diff
@@ -313,74 +338,74 @@ multiple versions of the Dart language.
...
-For an example of how an app or package can migrate to
-a new language version
-with an incompatible feature (null safety, for example), see 
+Language versioning becomes important on the rare occasions when Dart
+introduces an incompatible feature like [null safety][].
+When Dart introduces a breaking change, code that did compile
+might no longer compile. Using language versions allows you to indicate
+the compatibility of the code using the `pubspec.yaml` file.
+
+In the case of null safety, Dart versions 2.12 through 2.19 allowed you
```
parlough, line 347: Let's be specific about what kind of version we're talking about. Language versions (which this article is focused on) don't allow you to choose. It's the SDKs themselves (and their included tooling) that do.

```suggestion
In the case of null safety, Dart SDKs 2.12 through 2.19 allowed you
```

src/_guides/language/evolution.md
```diff
@@ -313,74 +338,74 @@ multiple versions of the Dart language.
...
+to _choose_ to update your code to use null safety.
+Dart uses language versioning to permit non-null-safe code to run
+alongside null-safe code.
+This decision enabled migration from non-null-safe to null-safe code.
+To check out an example of how an app or package can migrate to a new
+language version with an incompatible feature, check out
```
parlough, line 353: question: Is it fine to say "check out" twice like this. Would it be better to change the second version to "visit"?

src/_guides/language/evolution.md
```diff
@@ -313,74 +338,74 @@ multiple versions of the Dart language.
...
 [Migrating to null safety](/null-safety/migration-guide).
 
-Each package has a default language version,
-equal to the `<major>.<minor>` part of the
-**lower SDK constraint** in the pubspec.
-For example, the following entry in a `pubspec.yaml` file
-indicates that this package uses the Dart 2.7 language version.
+Each package has a default language version, equal to the
+`<major>.<minor>` part of the **lower SDK constraint** in the pubspec.
```
parlough, line 357: Especially as we transition to using caret syntax in pubspec files in the feature, it might make more sense to say something like "the lower bound of the SDK constraint" since the upper and lower bounds won't be specified separately.

src/_guides/language/evolution.md
```diff
@@ -338,49 +364,47 @@ environment:
...
+  In practice, many of these language versions function in very
+  similar manner to previous versions with full compatibility.
+  For example, the Dart 2.9 language works much like the Dart 2.8
+  language.
 
 * When a patch version of the SDK ships,
   it cannot introduce any language features.
-  For example, because 2.7.2 is language version 2.7,
-  it must be completely compatible with 2.7.1 and 2.7.0.
+  For example, because 2.18.3 belongs to language version 2.18,
```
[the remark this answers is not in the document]
parlough: Reading the previous text, it seems the point of the previous text was that 2.18.3 it doesn't patch the 2.18 language version, it just still is 2.18 in terms of language version. So I'm not sure "patches" is the best choice either.

The feature specification also mentions "A language version is a major/minor version number pair" so it seems there is no patch language versions.

src/_guides/language/evolution.md
```diff
@@ -390,20 +415,19 @@ For example:
...
+The `@dart` string must be in a `//` comment (not `///` or `/*`),
 and it must appear before any Dart code in the file.
 Whitespace (tabs and spaces) doesn't matter,
 except within the `@dart` and version strings.
-As the example above shows,
+As the previous example shows,
 other comments can appear before the `@dart` comment.
 
-For more information about how language versioning works, see the
+To learn how language versioning works, check out the
```
parlough, line 430: Isn't this whole section about how language versioning works?

We could either drop this or be more specific about why they would want to visit the specification.

site-shared
parlough, line 1: This reverts to an older version of site-shared, please undo this change.

There's a few ways to do this. Two of these easiest might be:

- `git submodule foreach git pull origin main` - Updates each submodule to track their latest commit on `main`. If you named `dart-lang/site-www` something besides `origin`, replace it with that.
- Switch to a branch tracking the latest `origin/main` and run `git submodule update` there.

parlough (changes requested):
Thanks for making some adjustments and responding to my comments and questions!

I'm also happy to discuss synchronously any of the concerns I raised if that works better for you as well.

src/_guides/language/evolution.md
```diff
@@ -180,131 +182,157 @@ turn into edit-time analysis errors.
...
+contain a number of preparatory [breaking changes][2.8 breaking changes]
+to ensure great nullability-related usability and performance in the
+upcoming [null safety][] feature.
 
-Dart 2.14 also removed some restrictions on type arguments.
-You can now pass type arguments to annotations,
-and you can use a generic function type as a type argument.
-All of the following code was invalid before 2.14,
-but is now allowed:
+It also contained a faster pub tool, and a new [pub outdated][] command.
```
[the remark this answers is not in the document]
parlough: See [my comments on another issue](https://github.com/dart-lang/site-www/pull/4684#discussion_r1150007415). This page isn't for general release notes and we shouldn't pretend it is in just some situations.

If we want release notes outside the changelog, that likely should be separat.e

src/_guides/language/evolution.md
```diff
@@ -4,173 +4,175 @@ short-title: Language evolution
...
-[null safety][] feature.
+```dart
+@TypeHelper<int>(42, "The meaning")
+late List<T Function<T>(T)> idFunctions;
+var callback = [<T>(T value) => value];
+late S Function<S extends T Function<T>(T)>(S) f;
+```
 
-It also contained a faster pub tool, and a new [pub outdated][] command.
+To learn more about these operators, check out [bitwise and shift operator][].
```
parlough, line 156: Sorry I missed this one before I submitted my review, but the text preceding this is not about the operator. Perhaps move this to right after the discussion of the operator.

src/_guides/language/evolution.md
```diff
@@ -313,74 +338,74 @@ multiple versions of the Dart language.
...
-  a new language version appears.
-  In practice, many of these language versions are very similar to
-  and entirely compatible with previous versions.
-  For example, the Dart 2.9 language is essentially identical to
-  the Dart 2.8 language.
+* Whenever a minor version of the SDK ships, a new language version appears.
+  In practice, many of these language versions function in a very
+  similar manner to previous versions with full compatibility.
+  For example, the Dart 2.9 language works much like the Dart 2.8
+  language.
```
parlough, line 388: It seems the changes lost the connection to [my comment](https://github.com/dart-lang/site-www/pull/4684#discussion_r1142087005) about this section, but my thoughts still stand:

> I think in general, I prefer the old text here. I think this change makes it say that the "language versions function in a similar manner when they have full compatibility", but the old text is saying that "many language versions are entirely compatible with each other".

src/_guides/language/evolution.md
```diff
@@ -313,74 +338,74 @@ multiple versions of the Dart language.
...
 
-* Whenever a minor version of the SDK ships,
-  a new language version appears.
-  In practice, many of these language versions are very similar to
-  and entirely compatible with previous versions.
-  For example, the Dart 2.9 language is essentially identical to
-  the Dart 2.8 language.
+* Whenever a minor version of the SDK ships, a new language version appears.
+  In practice, many of these language versions function in a very
+  similar manner to previous versions with full compatibility.
```
parlough, line 386: It seems the changes lost the connection to [my comment](https://github.com/dart-lang/site-www/pull/4684#discussion_r1142087005) about this section, but my thoughts still stand:

> I think in general, I prefer the old text here. I think this change makes it say that the "language versions function in a similar manner when they have full compatibility", but the old text is saying that "many language versions are entirely compatible with each other".

src/_guides/language/evolution.md
atsansone: **comment:** Updated.

src/_guides/language/evolution.md
```diff
@@ -4,173 +4,175 @@ short-title: Language evolution
...
 
-### Dart 2.5
+To learn more about Dart 2.17, check out:
 
-Dart 2.5 didn't add any features to the Dart language, but it did add
-support for [calling native C code][] from Dart code
-using a new **core library, `dart:ffi`.**
+* [Enhanced enums][]
+* [Super parameters][]
+* [Named parameters][]
```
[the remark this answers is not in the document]
parlough: I do think naturally embedding them may be better overall, but this new header is good to. We can try this out, thanks!

src/_guides/language/evolution.md
```diff
@@ -4,173 +4,176 @@ short-title: Language evolution
...
+except that it always fills the most significant bits with zeros.
 
-### Dart 2.8
+To learn more about these operators, check out [bitwise and shift operator][].
 
-Dart 2.8 didn't add any features to the Dart language, but it did
-contain a number of preparatory [breaking changes][2.8 breaking changes] to ensure great
-nullability-related usability and performance in the upcoming
-[null safety][] feature.
+[bitwise and shift operator]: /language/operators#bitwise-and-shift-operators
```
parlough, line 147: These should have an "s" at the end, since its about multiple operators.

parlough, line 147: These should have an "s" at the end, since its about multiple operators.

src/_guides/language/evolution.md
```diff
@@ -180,131 +182,157 @@ turn into edit-time analysis errors.
...
-Dart 2.13 also improved the struct support in **[Dart FFI][]**,
-adding support for inline arrays and packed structs.
+### Dart 2.10
+_Released 1 October 2020_
+| [Dart 2.10 announcement](https://medium.com/dartlang/announcing-dart-2-10-350823952bd5)
 
-### Dart 2.14
+Dart 2.10 added no new features to the Dart language.
+It did add an expanded [`dart` tool][dart-tool] similar to the
+Flutter SDK's [`flutter` tool][].
```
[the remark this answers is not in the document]
parlough: Tools are part of the SDK and similar to library changes, they are not relevant when evaluating and understanding language versions.

To be consistent with this focused direction, please remove these tool mentions.

src/_guides/language/evolution.md
```diff
@@ -180,131 +182,157 @@ turn into edit-time analysis errors.
...
 
-Dart 2.14 added the unsigned shift operator (`>>>`),
-also known as _triple-shift_.
-This new operator is similar to `>>`,
-except that it always fills the most significant bits with zeros.
-For more information, see the
-[bitwise and shift operator][] section of the language tour.
+### Dart 2.8
```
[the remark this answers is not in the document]
parlough: Thanks for adding it back! You can always check the [SDK changelog](https://github.com/dart-lang/sdk/blob/main/CHANGELOG.md) or the [archive](https://dart.dev/get-dart/archive) for releases which don't have a release blog.

src/_guides/language/evolution.md
```diff
@@ -180,131 +183,190 @@ turn into edit-time analysis errors.
...
-but now work for any type.
-You can use the new name created with a type alias
-anywhere the original type could be used.
+Dart 2.10 added no new features to the Dart language.
+It did add an expanded [`dart` tool][dart-tool] similar to the
+Flutter SDK's [`flutter` tool][].
 
-Dart 2.13 also improved the struct support in **[Dart FFI][]**,
-adding support for inline arrays and packed structs.
+### Dart 2.9
```
parlough, line 194: Since you added release dates for all the other entries, consider adding one here.

The [archive](https://dart.dev/get-dart/archive) specifies the binaries were released on Aug 5, 2020.

```suggestion
### Dart 2.9
_Released 5 Aug 2020_
```

src/_guides/language/evolution.md
```diff
@@ -180,131 +183,190 @@ turn into edit-time analysis errors.
...
+to improve nullability-related usability and performance for [null safety][].
 
-Dart 2.14 also removed some restrictions on type arguments.
-You can now pass type arguments to annotations,
-and you can use a generic function type as a type argument.
-All of the following code was invalid before 2.14,
-but is now allowed:
+It improved the performance of the `pub` tool and
+added the [`pub outdated`][] command.
+
```
parlough, line 208: Remove this tool mention as well to be consistent with the focused direction I mentioned before.

We don't mention tool changes for the other releases.


```suggestion
```

src/_guides/language/evolution.md
```diff
@@ -180,131 +183,190 @@ turn into edit-time analysis errors.
...
+_Released 5 November 2019_
+| [Dart 2.6 announcement](https://medium.com/dartlang/dart2native-a76c815e6baf)
 
-Dart 2.15 improved support for function pointers,
-known as _tear-offs._
-In particular, constructor tear-offs are now supported.
-For details, see the [Dart 2.15 announcement][].
+Dart 2.6 introduced a
+[breaking change (dart-lang/sdk#37985)](https://github.com/dart-lang/sdk/issues/37985).
+Inference changes when using Null values in a `FutureOr` context.
```
parlough, line 240: I believe it's not that inferences changes when using `Null` now, it's that inference overall changed in this release.

```suggestion
Inference changed when using `Null` values in a `FutureOr` context.
```

src/_guides/language/evolution.md
```diff
@@ -180,131 +183,190 @@ turn into edit-time analysis errors.
...
+Constraints of the forms similar to `Null <:FutureOr` now yield `Null`
```
parlough, line 241: The `<:` should be outside of the code.

```suggestion
Constraints of forms similar to `Null` <: `FutureOr` now yield `Null`
```

Also we might want to avoid the symbol completely. It's something readers may not be familiar with. Perhaps something like?

```suggestion
Constraints of the form where `Null` is a subtype of `FutureOr` now yield `Null`
```

src/_guides/language/evolution.md
```diff
@@ -180,131 +183,190 @@ turn into edit-time analysis errors.
...
 ```dart
-void main() {
-  test(skip: true, 'A test description', () {
-    // Very long function body here...
-  });
+class A<X> {};
+class B<X> extends A<void Function(X)> {};
+```
+
+You can now use the identifier async in asynchronous and generator functions.
```
parlough, line 280: Something doesn't exist as an identifier until you use it as such, so the following may read better:

```suggestion
You can now use `async` as an identifier in asynchronous and generator functions.
```

src/_guides/language/evolution.md
```diff
@@ -313,74 +375,75 @@ multiple versions of the Dart language.
...
+language version with an incompatible feature, check out
 [Migrating to null safety](/null-safety/migration-guide).
 
-Each package has a default language version,
-equal to the `<major>.<minor>` part of the
-**lower SDK constraint** in the pubspec.
-For example, the following entry in a `pubspec.yaml` file
-indicates that this package uses the Dart 2.7 language version.
+Each package has a default language version
+equal to the **lower bound of the SDK constraint** in the `pubspec.yaml` file.
```
parlough, line 395: Consider moving the emphasis to "lower bound".

```suggestion
equal to the **lower bound** of the SDK constraint in the `pubspec.yaml` file.
```

src/_guides/language/evolution.md
```diff
@@ -313,74 +375,75 @@ multiple versions of the Dart language.
...
-a new language version
-with an incompatible feature (null safety, for example), see 
+Language versioning becomes important on the rare occasions when Dart
+introduces an incompatible feature like [null safety][].
+When Dart introduces a breaking change, code that
+did compile might no longer compile.
+Language versioning allows you to set each library's language version
+to maintain compatibility.
+
+In the case of null safety, Dart SDKs 2.12 through 2.19 allowed you
```
parlough, line 385: This general area of text has got to a great point. Thanks for all of the back and forth with me here!

src/_guides/language/evolution.md
```diff
@@ -313,74 +375,75 @@ multiple versions of the Dart language.
...
-that match the first two components of the Dart SDK.
-For example, the latest language version supported by
-the 2.7.3 Dart SDK is Dart 2.7.
-Each Dart SDK supports all of the language versions covered
-by its major version number.
-That means that the 2.7.3 Dart SDK supports language
-versions 2.7, 2.6, 2.5, and so on, down to 2.0.
+Dart formats its language versions as two numbers separated with a period.
+It reads as a major version number and a minor version number.
+Though this appears to follow semantic versioning, it does not.
```
parlough, line 409: Consider dropping this sentence, it doesn't add anything to the meaning and it just seems a bit scary.

```suggestion
```

parlough (commented):
It's getting super close. Thanks for all of those changes.

A few last comments and answers to questions:

src/_guides/language/evolution.md
atsansone: **comment:** I'll drop it, but I do think it makes any confusion unambiguous.

src/_guides/language/evolution.md
```diff
@@ -180,131 +183,185 @@ turn into edit-time analysis errors.
...
+| [Dart 2.6 announcement](https://medium.com/dartlang/dart2native-a76c815e6baf)
 
-Dart 2.15 improved support for function pointers,
-known as _tear-offs._
-In particular, constructor tear-offs are now supported.
-For details, see the [Dart 2.15 announcement][].
+Dart 2.6 introduced a
+[breaking change (dart-lang/sdk#37985)](https://github.com/dart-lang/sdk/issues/37985).
+Constraints of the forms similar to `Null` serves as a subtype of `FutureOr`
+now yield `Null` as the solution for `T`.
```
parlough, line 237: I think the "Constraints of the forms similar to" portion is pretty confusing, can we simplify it? I also didn't notice the missing generic on `FutureOr<T>`. That's important for knowing what the following `T` is referring to.

```suggestion
Constraints where `Null` serves as a subtype of `FutureOr<T>`
now yield `Null` as the solution for `T`.
```

src/_guides/language/evolution.md
```diff
@@ -313,74 +371,74 @@ multiple versions of the Dart language.
...
-Each package has a default language version,
-equal to the `<major>.<minor>` part of the
-**lower SDK constraint** in the pubspec.
-For example, the following entry in a `pubspec.yaml` file
-indicates that this package uses the Dart 2.7 language version.
+Each package has a default language version equal to the **lower bound**
+of the SDK constraint in the `pubspec.yaml` file.
+
+**For example:** The following entry in a `pubspec.yaml` file
+indicates that this package uses the Dart 2.18 language version or later.
```
parlough, line 394: This indicates it uses the 2.18 language version by default, not later. It can't use later language versions or else that would break the constraint.

```suggestion
indicates that this package uses the Dart 2.18 language version by default.
```

parlough (approved):
I had to rebase these changes to fix some issues with the PR and pushed some minor changes. There's **two last changes** that need to be addressed before merging but I want you to take a look. 

Otherwise, looks good to me. Thanks for doing this and all of the discussion. I think the page has got to a really great spot and has a clearer vision now.

Optionally get a second review from someone after making those fixes, but since it's mostly reorganized and rewritten existing content, it should be fine.

PR #4697, opened 2023-03-20
Remove This Week in Dart link for now

parlough:
As I've been busy with other work, I'm not able to update it as much as I would like currently. I hope to revive it in the future, but for now, let's not link to potentially outdated information.


domesticmouse (approved):
/me does sad face

PR #4699, opened 2023-03-20
Add Dart 3.0 diagnostic messages

parlough:
These docs have already been reviewed, just need to be updated here so the analyzer can link to them.

These are surfaced in https://dart-dev--pr4699-diagnostics-3-0-yj4g7j5c.web.app/tools/diagnostic-messages


src/tools/diagnostic-messages.md
```diff
@@ -3076,18 +3255,19 @@ Add an initializer:
 const c = 'c';
 {% endprettify %}
 
-### const_set_element_type_implements_equals
```
parlough, line 3079: @bwilkerson It seems this one loses a destination since it was renamed to `const_set_element_not_primitive_equality` with added functionality. Is that expected or should `const_set_element_not_primitive_equality` have an anchor generated for this so Dart 2.19 users don't lose a link destination?

parlough: Thanks! I'll hold off for your changes in https://dart-review.googlesource.com/c/sdk/+/290180 :)

bwilkerson (approved):

sfshaza2 (approved):
lgtm!

atsansone:
@parlough : Would these be merged into the `v3.0` branch instead?

parlough:
Yes, I plan to publish these now so the beta SDK can link to them. 

However, I [removed the added `final` modifiers](https://github.com/dart-lang/site-www/pull/4699/commits/7e9a83b2f8633b4c64a28f5825a98cc37c4829c8) from the `class` declarations in the pre-existing FFI diagnostics, since they will only cause confusion. I will make another update when 3.0 is released when we can add them back.

parlough:
Thanks for the great diagnostics Brian and Marya!

These changes will be deployed in ~10 minutes.

Issue #4681, opened 2023-03-16
Gitter.im link no longer working correctly on "Community and support" page


julemand101:
### Page URL

https://dart.dev/community

### Describe the problem

Since Gitter.im moved over to use their own Matrix server with an Element client, the link on the community page no longer works as expected. Instead of going to an active channel, it will give the error:

> #dart-lang_home:gitter.im does not exist.
> Are you sure you're at the right place?

### Expected fix

I am not entire sure what the link should be updated to, since the new Gitter.im does not seem to have any concept of "groups of channels" but instead the 4 Dart specific channels does have each an individual link:

- https://app.gitter.im/#/room/#dart-lang_build:gitter.im
- https://app.gitter.im/#/room/#dart-lang_server:gitter.im
- https://app.gitter.im/#/room/#dart-lang_TALK-general:gitter.im
- https://app.gitter.im/#/room/#dart-lang_sdk:gitter.im

So it gets a bit more convoluted to provide instructions into how to join the channels.

parlough:
We could always link to all of them, or at least the general one. 

I'm not active there, so I'm curious, does it see any traffic these days? Does it make sense to keep the links at all, especially now the channels aren't grouped? Or would it be better for most users to look elsewhere?

julemand101:
The Gitter.im channels are rather dead and mostly only used by a few members. Only community channel more dead are the #dart IRC channel on libera.chat.

I guess the most active chat service related to Dart are now Discord but that does face the problem that a lot of developers seems to not really like Discord and we have in the past a person who simply could not join Discord because of phone number requirement.

The issue then becomes if alternatives should be provided but where the alternatives does not provide any active community that can help in case of questions. I am personally logged in on most Dart social channels in case somebody are going to start a conversation. But I do recognize that we are only a very small group of people who does this.

So, personally, I don't really have an issue with removing the Gitter.im channels since I am mostly active on Discord. And I do also think it is a better user experience to not be meet with a dead chat as a beginner. But I don't know if there are any solution for people who seeks a community chat but cannot use Discord for some reason or another.

parlough:
Currently the Dart Discord is set up to only require an account with an email, not a phone number. The Gitter instances seem to require an account as well, so at least that's not a regression.

Like you mentioned, my main concern is leading users to a dead chat where they won't get help or have proper discussions. I also haven't heard any other reports of the link being dead since they made the transition in February. Perhaps StackOverflow, Reddit, and Discord are enough for now.

[...]

PR #4700, opened 2023-03-21
Remove outdated Gitter link

parlough:
The 4 Dart channels are no longer grouped with Gitter's full transition to Matrix, and the channels see little traffic. It doesn't seem like a great experience to direct users to 4 different channels where they may not get the support they need. Very few Dart team members are active there anymore as well, despite the description provided here.

Closes https://github.com/dart-lang/site-www/issues/4681


atsansone (approved):
LGTM.

Issue #4693, opened 2023-03-20
Updates for using caret syntax in pubspec SDK constraint


parlough:
As of Dart 2.19, pub now doesn't block specifying the SDK constraint with the caret syntax. We should document it, modify our mention of recommending against it, and consider recommending/preferring it in certain locations.

Commit reference: https://github.com/dart-lang/pub/commit/06ab0148b6b320b4bfc69b0adba028dd745268ef.

parlough:
`dart create` may use it as the default in the future too: https://dart-review.googlesource.com/c/sdk/+/289900

parlough:
Currently planning to change this alongside Dart 3 or closer to Dart 3, but let me know if you think we should adjust the warning now.

domesticmouse:
So totally lean into specifying SDK constraint with caret syntax!

sigurdm:
> but let me know if you think we should adjust the warning now or not change to this syntax at all :)

I think this can be adjusted already now. (No specific urgency though).

parlough:
Sounds good, I’ll make the adjustments then. Thanks!

atsansone:
@parlough : Is this complete?

parlough:
It's partially complete. I have some in-progress local work I can wrap up when I have a chance.

parlough:
Fixed by https://github.com/dart-lang/site-www/commit/605b86c52111dd530a7a0f64014e5d0a44fb8172

Will land into `main` with other Dart 3 changes soon.

PR #4701, opened 2023-03-21
Switch tooling SDK constraints to caret syntax

parlough:
Completing this as a first step to verify everything works as expected. Will follow up with doc changes and updates later on.

Contributes to https://github.com/dart-lang/site-www/issues/4693


atsansone (approved):
@parlough : Looks good.

PR #4702, opened 2023-03-21
Update site-shared for Dart 3 prettify updates

parlough:
This is necessary for https://github.com/dart-lang/site-www/pull/4699 which uses Dart 3 syntax in prettify code blocks.


atsansone (approved):
@parlough : LGTM.

PR #4708, opened 2023-03-22
Added Analyzer plugins section

atsansone:


src/_guides/language/analysis-options.md
```diff
@@ -381,6 +381,44 @@ linter:
...
+Each analyzer plugin that you use increases how much memory the analyzer uses.
+
+Don’t use analyzer plugins if your situation meets either of the following
+conditions:
+
+* Your development machine has less than 16 GB of memory
+* You use a mono-repo with more than 10 `pubspec.yaml` and
+  `analysis_options.yaml` files.
+  
+You can find a few analyzer plugins on [pub.dev]({{site.pub}}).
```
parlough, line 400: I haven't taken a look at the rest of this PR yet, but I wanted to ask if it would make sense to link to packages that depend on `package:analyzer_plugin` here?

https://pub.dev/packages?q=dependency%3Aanalyzer_plugin

Linking to just pub.dev may be confusing as they won't know how to find analyzer plugins from there.

jacob314 (approved):
lgtm

src/_guides/language/analysis-options.md
atsansone: @parlough : I agree with that. Let me do that.

src/_guides/language/analysis-options.md
```diff
@@ -381,6 +381,44 @@ linter:
...
+
+The Analyzer package offers an experimental feature called Analyzer plugins.
+These plugins allow you to add linter rules and quick fixes that third-party
+plugins specify.
+You can use no more than one Analyzer plugin in each
+`analysis_options.yaml` file.
+Each analyzer plugin that you use increases how much memory the analyzer uses.
+
+Don’t use analyzer plugins if your situation meets either of the following
+conditions:
```
parlough, line 394:
```suggestion
Don’t use analyzer plugins if your situation meets
either of the following conditions:
```

parlough (commented):
Sorry about the delay on the review and thanks for working on this. Originally the plan was to not document analyzer plugins at all, so it's interesting to see that has changed.

I had some questions and uncertainties, and struggled explaining some thoughts in text. So if something doesn't make sense, let me know. Thanks!

src/_guides/language/analysis-options.md
```diff
@@ -381,6 +381,44 @@ linter:
   You can, however, use the other syntax for rules in an included file.
 {{site.alert.end}}
 
+## Enabling analyzer plugins (experimental)
+
+The Analyzer package offers an experimental feature called Analyzer plugins.
+These plugins allow you to add linter rules and quick fixes that third-party
+plugins specify.
```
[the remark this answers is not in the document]
incendial: IMO, lints and linter in general are a broader term than the linter package (as part of the Dart analyzer) and "lint rules" sounds more familiar in terms of static analysis than "diagnostics", but I may be biased here. The term `rules` that we use is basically a short version of `lint rules` and the code metrics calculation is not based on the plugins API.

> Also, as far as I know, analyzer plugins can do a lot more than just add diagnostics and quick fixes. For example, provide quick assists, code completions, highlighting and folding regions, and much more.

If I'm not mistaken there were some discussions about changing the plugins API.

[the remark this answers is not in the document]
parlough: I agree these are definitely broader terms, I'm just worried about confusion with the current idea of the dart linter and its linter rules. At least right now on the site, we use "linter rules" to specifically refer to rules added to the analyzer's linter under `linter` in `analysis_options.yaml`. We could work to be more specific, but I thought a more generic term could cover more cases and avoid that confusion for now.

For example, couldn't a plugin also report a diagnostic with an `error` severity that is not explicitly enabled like "linter rules" (similar to the analyzer's errors and warnings)? I guess in my mind, I don't really consider a built-in error a lint.

Either way, if "lints" is the preferred term, I think that's fine. We'll just need to be more careful about how we position the built-in ones.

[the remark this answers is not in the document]
parlough: As for the API changing. If that's the case, we should definitely stick to whatever features are more guaranteed as part of that.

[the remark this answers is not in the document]
incendial: > At least right now on the site, we use "linter rules" to specifically refer to rules added to the analyzer's linter under linter in analysis_options.yaml. We could definitely work to be more specific, but I thought a more generic term could avoid that confusion for now while also covering more situations.

Ah, I see your point now, thank you for the details.

src/_guides/language/analysis-options.md
```diff
@@ -381,6 +381,44 @@ linter:
   You can, however, use the other syntax for rules in an included file.
 {{site.alert.end}}
 
+## Enabling analyzer plugins (experimental)
```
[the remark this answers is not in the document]
jacob314: Things have changed. While we are still not comfortable with the current support we need to document the experimental feature because users are confused and filing bugs when they hit corner cases such as only one plugin per package.

jacob314:
> 



> Sorry about the delay on the review and thanks for working on this. My understanding was that originally the plan was to not document analyzer plugins at all, since it hasn't been determined if they will stick around, so it's interesting to see that has changed. Could you provide some more context there?
> 
> I had some questions and uncertainties, and struggled explaining some thoughts in text. I also think I'm missing some context on this change, such as potential changes to the plugin API. So if something doesn't make sense, let me know. Thanks!

Discuss

src/_guides/language/analysis-options.md
```diff
@@ -376,11 +376,48 @@ linter:
 ```
 
 {{site.alert.note}}
-  Due to YAML restrictions, 
+  Due to YAML restrictions,
   **you can't mix list and key-value syntax in the same `rules` entry.**
-  You can, however, use the other syntax for rules in an included file.
+  You may use the other syntax for rules in an included file.
```
parlough, line 381: Consider reverting or changing this, since we [avoid "may"](https://developers.google.com/style/word-list#may).

src/_guides/language/analysis-options.md
```diff
@@ -376,11 +376,48 @@ linter:
...
+* You use a development machine with less than 16 GB of memory.
+* You use a mono-repo with more than 10 `pubspec.yaml` and
+  `analysis_options.yaml` files.
+  
+You can find a few analyzer plugins on
+[pub.dev]({{site.pub-pkg}}?q=dependency%3Aanalyzer_plugin).
+
+To enable a plugin:
+
+1. Add the package containing the plugin as a dev dependency.
```
parlough, line 404: We haven't been super consistent about this, but with ordered lists, the first line of each entry should begin with one space. This prevents additional space being added to the beginning of the code snippets.

We document this at https://github.com/dart-lang/site-shared/blob/main/doc/markdown.md#do-properly-indent-markdown-lists

```suggestion
 1. Add the package containing the plugin as a dev dependency.
```

src/_guides/language/analysis-options.md
```diff
@@ -376,11 +376,48 @@ linter:
...
+
+    ```terminal
+    dart pub add --dev your_favorite_analyzer_plugin_package
```
parlough, line 407: We begin our terminal examples with a `$` across the site:

```suggestion
    $ dart pub add --dev your_favorite_analyzer_plugin_package
```

src/_guides/language/analysis-options.md
```diff
@@ -376,11 +376,48 @@ linter:
...
+    ```
+
+2. Edit your `analysis_options.yaml` file to enable the plugin.
```
parlough, line 410: Same as above: 

> We haven't been super consistent about this, but with ordered lists, the first line of > each entry should begin with one space. This prevents additional space being added to > the beginning of the code snippets.
> 
> We document this at https://github.com/dart-lang/site-shared/blob/main/doc/markdown.md#do-properly-indent-markdown-lists

```suggestion
 2. Edit your `analysis_options.yaml` file to enable the plugin.
```

parlough (approved):
Looks good to me after a few final changes.

Thanks for making all of those adjustments, this looks like it's in a pretty good place!

src/_guides/language/analysis-options.md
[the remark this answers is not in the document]
parlough: Thanks for the extra context, definitely makes sense!

parlough:
Thanks for making those adjustments! Looks good to me :)

PR #4733, opened 2023-04-03
Updated the language index page for clarification

mitulamin:
Updated a bullet point to include a simple code snippet as previously it was referring to a non existent example.


parlough (approved):
Thanks for catching and fixing this! We recently restructured the page, replacing the sample it was referring to, and forgot to adjust this.

PR #4750, opened 2023-04-11
Remove mention of `spawn` sample on concurrency page

### LINKED ISSUE #4721: use `Isolate.spawn()` instead of `Isolate.run()` in sample send_and_receive
As section [Sending multiple messages between isolates](https://dart.dev/guides/language/concurrency#sending-multiple-messages-between-isolates) mentions that there is a [sample](https://github.com/dart-lang/samples/blob/master/isolates/bin/send_and_receive.dart) about using `Isolate.spawn()` instead of `Isolate.run()`, but i can't see any difference with previous code snippets.

image

image

MaryaBelanger:
fixes #4721

### DIFF
--- a/src/language/concurrency/index.md
+++ b/src/language/concurrency/index.md
@@ -381,6 +381,8 @@ Whether the computation the worker isolate executes
 is synchronous or asynchronous doesn't impact the
 main isolate, because it's running concurrently either way.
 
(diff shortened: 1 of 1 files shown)


parlough (approved):
Thanks for clearing up this confusion. I forgot to update this when I updated the sample as well.

PR #4752, opened 2023-04-12
docs: Use absolute path for PUB_CACHE env variable

Gustl22:
The description "home directory" can lead to confusion. It can be stated clearer where to find the cache directory by using the absolute path with `$HOME` environment variable. This also avoids issues with `relative path`, or miswriting the path e.g. with `~` when manually setting it in env variables (`.bashrc`, etc.)

Closes https://github.com/dart-lang/pub/issues/3861


src/tools/pub/environment-variables.md
```diff
@@ -9,8 +9,7 @@ Environment variables allow you to customize pub to suit your needs.
 
 `PUB_CACHE`
 : Some of pub's dependencies are downloaded to the pub cache.
-  By default, this directory is located under `.pub-cache`
-  in your home directory (on macOS and Linux),
+  By default, this directory is located under `$HOME/.pub-cache` (on macOS and Linux),
```
parlough, line 12: We try to keep lines under 80 characters:

```suggestion
  By default, this directory is located under
  `$HOME/.pub-cache` (on macOS and Linux),
```

parlough (approved):
Thanks for updating this to make it more clear! Looks good to me.

Issue #4730, opened 2023-03-31
[Dart 3] Document general use of class modifier by core Dart libraries


parlough:
A lot of Dart core libraries are introducing usage of class modifiers to limit various uses of them (extending, implementing, mixing in, etc). We should make users aware of this somehow, likely when introducing the core libraries themselves. It can link to other [class modifier docs](https://github.com/dart-lang/site-www/issues/4497) if necessary.

MaryaBelanger:
I think the communication plan for all of the Dart 3 migration / breaking changes info is to add a "Dart 3 migration" page to dart.dev somewhere. I've added notes about this to the planning doc for that communication and am waiting to see if others think it fits.

I'm not sure who's in charge of creating that content though, will update.

parlough:
That sounds good, thanks.

I definitely still see value in education users of the possibility of these limitations, whether they are migrating or not, while they approach core library usage.

MaryaBelanger:
List of affected APIs per library from the changelog:

- [dart:core](https://github.com/dart-lang/sdk/blob/main/CHANGELOG.md#:~:text=for%20moving%20backwards.-,Breaking%20change,-when%20migrating%20code)
- [dart:collection](https://github.com/dart-lang/sdk/blob/main/CHANGELOG.md#dartcollection)
- [dart:math](https://github.com/dart-lang/sdk/blob/main/CHANGELOG.md#:~:text=dart%3Amath-,Breaking%20change,-when%20migrating%20code)

MaryaBelanger:
Note: [the migration guide](https://dart.dev/resources/dart-3-migration#extends--implements) summarizes this a little (not a replacement for documenting in the library tour)

MaryaBelanger:
3.1.0 library changes dues to class modifiers:

- https://github.com/dart-lang/sdk/issues/52334

parlough:
Thanks for calling attention to that, looks like it was part of the 3.0.3 release. I opened a PR adding them to the migration guide - https://github.com/dart-lang/site-www/pull/5075

MaryaBelanger:
@parlough Thanks for adding those 3.0.3 changes to the migration guide. Just want to re-confirm: you still think those and all the other core library changes should be documented in their sections in the Library Tour, correct? 

I'm updating the title of this issue since documenting them in the migration guide could be misconstrued as solving the title as written, but lmk if you're thinking differently!

parlough:
Thanks for updating that.

[...]

Issue #4635, opened 2023-03-07
[Patterns] Control flow docs


MaryaBelanger:
### Summary

Patterns-related features mostly surrounding capabilities with control flow. Breakaway from main patterns plan: #4178 

#### If-case

1. Likely a new page with the control flow content, following "[if and else](https://dart.dev/guides/language/language-tour#if-and-else)"
    - Or, could change the existing if and else section to "If, case, and else" since you can do if-case-else 
3. Both [statement](https://github.com/dart-lang/language/blob/master/accepted/future-releases/0546-patterns/feature-specification.md#if-case-statement-and-element) and [element](https://github.com/dart-lang/language/blob/master/accepted/future-releases/0546-patterns/feature-specification.md#if-case-element) are new
4. if-case elements are allowed inside collection literals 
    - might be content for that section, or include in if-case content
6. if-case statements: can now follow if with case, followed by a pattern
7. `else` branch can follow if-case have an else branch
8. Use case: null check pattern as the pattern following `case` to see if the incoming value is not null, then declare a variable right inside the condition for the null-safe value 
    - kind of an alternative to typical null-check-and-declare/assign set up 

#### Guard clause

1. Both switch and if-case can utilize the new `when` guard clause
    - [this section ](https://github.com/dart-lang/language/blob/master/accepted/future-releases/0546-patterns/feature-specification.md#guard-clause)is under "switch"
    - guard mentioned under [if-case section](https://github.com/dart-lang/language/blob/master/accepted/future-releases/0546-patterns/feature-specification.md#if-case-statement-and-element)

2. Couple of options here
     - Guard has it's own section under control flow (level with switch and if-case)
     - Guard is mentioned in both sections 
     - Guard is mentioned in one topic's contents (probably switch), and pointed to from if-case

3. Guards are useful because "when the predicate evaluates to false, execution proceeds to the next case instead of exiting the entire switch like it would if you nested an if statement inside the switch case's body"

4. Ambiguity with `=>` after `when` (guard) : everything between `when` and the first `=>` is the guard clause, and after `=>` is the case body (even if it contains another `=>`) 
     - this really only occurs for switch _expressions_, but context-sensitive syntax is harder to learn, so the rule is applied anywhere guard - case happens (switch statements and expressions, and if-case)

#### Switch

1. Extend [Switch and case](https://dart.dev/guides/language/language-tour#switch-and-case) content from language tour.
    - Will separate into major sections or pages for switch statements (current content), expressions (new)

[...]

PR #4754, opened 2023-04-12
[Dart 3] Class modifiers

### LINKED ISSUE #4497: [Class modifiers] Language feature docs
### Summary

Class modifiers determine how a class can be implemented, extended, or mixed in relative to the library it's defined in. Invariants like this are up to the library author; they can make types easier to understand and use. 

### Evaluation

1. ~Class modifiers will be a stand-alone page under "Language > Classes & objects" in the side nav~

    There may also be coverage, or at least references to the core coverage, somewhere in the "Libraries" content.

2. ~The [table of combinations:impacts](https://github.com/dart-lang/language/blob/master/accepted/future-releases/class-modifiers/feature-specification.md#syntax) is NOT necessary~
    - just the use cases, start with important ones

4. ~[Mixin docs](https://dart.dev/guides/language/language-tour#adding-features-to-a-class-mixins) need to be updated: classes can't be used as a mixin by default anymore, must explicitly use the modifier `mixin class`~
  
  -  ~Class modifiers content needs to come after existing mixin content, because all the modifiers apply to mixins and classes.~ 
  - ~See [Mixin restrictions](https://github.com/dart-lang/language/blob/master/accepted/future-releases/class-modifiers/feature-specification.md#mixin-restrictions) section of spec for changes to mixins around the allowance of constructor declaration (currently can't declare a constructor, restrictions loosened)~

5. ~Potentially, for each new class modifier/class modifiers combination, briefly mention the use cases under "Motivation" in the spec to give users an idea of when to use this.~

    ~e.g. "The `final` modifier` ... disallowing implementation and extension of a type makes it entirely safe to add new members to"~

    ~Another way to organize this information would be the inverse, i.e. by motivation: "The modifiers disallow implementing or extending, which is good for.... <use case>"~

6. Make sure to add the page link to related diagnostics e.g. https://github.com/dart-lang/linter/pull/4101

7. See #4730  
- will open another PR to address this since the first one is pretty large

8. ~(maybe) add something to development or package content about how modifiers are for package maintainers / developers who want to limit the usage capabilities for others using their package~
- skip

9. ~"[extend a class](https://dart.dev/language/extend)" page and "[implicit interfaces](https://dart.dev/language/classes#implicit-interfaces)" section need to point to class modifiers page ("class modifiers restrict how classes can be extended/implemented")~
- think I'll skip this for now

10. ~(maybe) move "[abstract classes](https://dart.dev/language/classes#abstract-classes)" section to class modifiers page?~
 - ~need to find and fix links to it across the site, add redirect~

12. ~add mention + link to class modifiers in intro paragraph of main "[Classes](https://dart.dev/language/classes)" page~

13. ~Not sure if anything needs to change in "[nosuchmethod](https://dart.dev/language/extend#nosuchmethod)" section based on this note:~
    
    > ~If you have a class that is marked base or final, that means that you are ensured when an instance of that class is passed into your library that any access of a private member on it will never throw a nosuchmethod error~

15. ~#4662~  

### Resources

- https://github.com/dart-lang/language/blob/master/accepted/future-releases/class-modifiers/feature-specification.md
- https://dart.dev/guides/language/language-tour#adding-features-to-a-class-mixins

### LINKED ISSUE #4581: [Sealed class] Language feature docs
### Summary

Sealed classes are a class modifier. Declaring a supertype class `sealed` means it can't be directly extended, implemented, or mixed in from outside their originating library. 

From the [Class Modifiers spec](https://github.com/dart-lang/language/blob/master/accepted/future-releases/class-modifiers/feature-specification.md#restrictions-within-the-same-library): 

> "But it doesn't prevent you from subtyping within the same library. In fact, the whole point of sealed is to define subtypes within the same library so that you can pattern match on those to cover the supertype."

Their motivating use case is exhaustiveness checking, a new feature of switch cases. They are the only class modifier that touch on patterns. 

### Evaluation

1. Document semantics with new Class modifiers content. Show how to use (just `sealed` keyword), show example of sealed superclass and subclasses, and what affect that has on subclasses. 

     Content in class modifiers section shouldn't address exhaustiveness checking or patterns, but should link out to that content ("learn more about sealed classes use cases...).

    - **Addressed above two in #4754**     

     Class modifiers content might go under the "Libraries" umbrella (side nav), or under the "Classes" section of the "Language" umbrella (see #4497)

     - **Tracked in #4730** 

3. In exhaustiveness checking content (which will be part of new switch content), link to semantic definition of the `sealed` class modifier.
Then, go into detail on how sealed creates a closed hierarchy of classes, how that enables exhaustiveness checking that allows you to switch on object patterns.  

4. IF we get to a point where we're documenting the collection of patterns features that enable functional style, then sealed classes will come up there (but that might be sufficiently covered by the description of exhaustiveness checking and switch)

5. MAYBE point to sealed content from [enum content in language tour](https://dart.dev/guides/language/language-tour#enumerated-types).
    
    Something like "you can enumerate subclasses using the [`sealed` class modifier]()..."

### Resources

- https://github.com/dart-lang/language/blob/master/accepted/future-releases/sealed-types/feature-specification.md

### LINKED ISSUE #4662: Change link for interface on 'Keywords' page
### Page URL

https://dart.dev/language/keywords.html

### Page source

https://github.com/dart-lang/site-www/tree/main/src/language/keywords.md

### Describe the problem

The new [_class modifiers_](https://github.com/dart-lang/language/blob/master/accepted/future-releases/class-modifiers/feature-specification.md) feature which is about to be added to Dart will introduce the built-in identifier `interface` as a part of Dart source code (until now it has been unused, but it was still classified as a built-in identifier).

The 'Keywords' page shows `interface` as a built-in identifier (as it should), but it is linked to a section about 'implicit interfaces', which is misleading now.

### Expected fix

`interface` is going to be a class modifier such that `interface class C ...` declares a class where subtypes can be declared using `class D implements C {}` and `mixin M on C {}`, but not `class D extends C {}` (at least not in a different library). The link might actually just refer to the feature specification, https://github.com/dart-lang/language/blob/master/accepted/future-releases/class-modifiers/feature-specification.md, or it could refer to a new section about class modifiers.

### Additional context

Thanks to @e1b916a3 for reporting this in https://github.com/dart-lang/sdk/issues/51728.

MaryaBelanger:
Reviewers: 
- New class modifiers page [staged here](https://dart-dev--pr4754-class-modifiers-lrg9su20.web.app/language/class-modifiers)
- Minor changes to mixin page [here](https://dart-dev--pr4754-class-modifiers-lrg9su20.web.app/language/mixins)
- Added modifiers to [Keywords page](https://dart-dev--pr4754-class-modifiers-lrg9su20.web.app/language/keywords)
- Moved "Abstract classes" section off of "Classes" page and to "Class modifiers > `abstract`" section (didn't update text) [here](https://dart-dev--pr4754-class-modifiers-lrg9su20.web.app/language/class-modifiers#abstract)

---

Updates in progress after first review: 
- Need to change the examples so the code where `extends`, `implements`, etc is failing is clearly NOT in the same library where the modified class is declared.
- The way I have it written now makes it seem like it's in the same library, in which case the modifiers wouldn't restrict anything.
- Call out that the modifiers affect _outside libraries_ more often and explicitly.
- Need to change the first line [here](https://dart.dev/language#interfaces-and-abstract-classes): "Dart has no interface keyword." 

---

TODO (possibly separate PRs):
- Change code to actual code excerpts
- Address changes to core libraries in library tour (#4730)
- In `/language/class-modifiers#sealed`, need to change link on "exhaustiveness checking" to actual section once that's written

---

Fixes #4497 
Fixes #4662 
Fixes #4581


src/language/class-modifiers.md
```diff
@@ -0,0 +1,240 @@
...
+    - [Inheritance](#interface)?
+    - [Both](#final)?
+    - [Subtyping](#sealed) by outside libraries at all?
+3. Then, does it need to be [abstract](#abstract)?
+4. And lastly, should it be a [class, mixin, or mixin class](#class-mixin-or-mixin-class)?
+
+## No modifier
+
+Use a `class` or `mixin` declaration without a modifier to **allow unrestricted
+access** to outside libraries. The standard behaviors apply:
```
kallentu, line 27: Maybe we should specify "outside its library" more often, or earlier. It doesn't seem to be mentioned until this point which seems a little far in. Maybe something on line 6?

src/language/class-modifiers.md
```diff
@@ -0,0 +1,240 @@
...
+
+- [Construct][] new instances of a class
+- [Extend][] a class to create a new subtype
+- [Implement][] a class's interface
+- [Mix in][] a mixin or mixin class
+
+## `base`
+
+Use the `base` modifier to **disallow implementation** of a class or mixin.
+A base class can be extended and constructed. This applies trasitively to all
```
kallentu, line 37: Nit: Typo on `transitively`.
But also, I'm assuming you're talking about disallowing implementation. It feels like this sentence would flow better coming out of the first sentence.

src/language/class-modifiers.md
```diff
@@ -0,0 +1,240 @@
...
+- Closing the type hierarchy.
+- Ensuring that adding new members to the class won't risk breaking changes for 
+its consumers.
+- Ensuring access to private members on instances of the final class will never
+throw a [`noSuchMethod`][] error.
+
+## `sealed`
+
+Use the `sealed` modifier to define a class that **cannot be constructed, extended,
+or implementated** outside of the library where the class is defined.
```
kallentu, line 153: Nit: Typo with `implemented`

src/language/class-modifiers.md
```diff
@@ -0,0 +1,240 @@
...
+its consumers.
+- Ensuring access to private members on instances of the final class will never
+throw a [`noSuchMethod`][] error.
+
+## `sealed`
+
+Use the `sealed` modifier to define a class that **cannot be constructed, extended,
+or implementated** outside of the library where the class is defined. 
+
+The purpose of the `sealed` modifier is to enable [_exhaustiveness checking_][]
```
kallentu, line 155: I would maybe start with this information -- that the main point of `sealed` is for exhaustiveness. The fact that it can't be subtyped at all outside of its library comes after the exhaustiveness imo.

We prefer users to use `final` if they don't need the exhaustiveness feature.

src/language/class-modifiers.md
```diff
@@ -0,0 +1,240 @@
...
+
+  void updateChildren(); // Abstract method.
+}
+```
+
+## `class`, `mixin`, or `mixin class`?
+
+The `mixin` declaration defines a [mixin][]. The `mixin` _modifier_ can only be
+applied to a `class` declaration to create a _mixin class_: a class that can be
+treated both as a class and a mixed in. Regular classes cannot be mixed in.
```
kallentu, line 208: Nit: Typo on `treated both as a class and a mixed in`.

src/language/class-modifiers.md
```diff
@@ -0,0 +1,240 @@
...
+
```
kallentu, line 209: But also, maybe an example of a mixin class would be useful? Like how we do with the other modifiers.

src/language/class-modifiers.md
```diff
@@ -0,0 +1,240 @@
...
+    @override
+    void moveForward …
+}
+```
+
+Disallowing inheritance makes `interface` useful for:
+- Defining API interfaces.
+- Preventing users of the API from overriding a subset of methods.
+
+## `final`
```
kallentu, line 113: Regarding your major concerns in the PR:

> Having the code together like that in the same code block makes it seem like it's all in the same library.

I only read this after I commented and I agree. I think it would be useful to not have them in the same code block or to create some representation of different libraries.

> But, if it's all in the same library, would attempting to extend the final class actually be an error? Or would it only be an error if it was attempted in an external library?

It's only an error if attempted in an external library.
You have (almost) full capabilities if using a final/base/interface/sealed class in the same library.

For example, implementing a final class within the same library -- https://github.com/dart-lang/sdk/blob/main/tests/language/class_modifiers/final/final_class_implement_lib.dart
And then once you go outside that library and try to implement a final class -- https://github.com/dart-lang/sdk/blob/main/tests/language/class_modifiers/final/final_class_implement_outside_error_test.dart

src/language/class-modifiers.md
```diff
@@ -0,0 +1,240 @@
+---
+title: Class modifiers
+description: Modifier keywords for class declarations to control external library access.
+---
+
+Class modifiers control how a class or a mixin can be implemented, extended,
+or mixed in.
```
leafpetersen, line 7: maybe mention exhaustiveness?  "or mixed in, and whether they are considered for exhaustiveness checking in switches.".

src/language/class-modifiers.md
```diff
@@ -0,0 +1,240 @@
...
+Only the `base` modifier can appear before a mixin declaration. The modifiers do
+not apply to other declarations like `enum`, `typedef`, or `extension`.
+
+When deciding whether or not to use class modifiers, consider the following questions:
+1. Is [any access](#no-modifier) fine?
+2. If not, what do you want to restrict?
+    - [Implementation](#base)?
+    - [Inheritance](#interface)?
+    - [Both](#final)?
+    - [Subtyping](#sealed) by outside libraries at all?
```
leafpetersen, line 20: This feels a little off.  The reason for using sealed is not that you want to prevent subtyping by outside libraries (that's what final is for).  The reason is that you want to enable exhaustiveness.

src/language/class-modifiers.md
```diff
@@ -0,0 +1,240 @@
...
+access** to outside libraries. The standard behaviors apply:
+
+- [Construct][] new instances of a class
+- [Extend][] a class to create a new subtype
+- [Implement][] a class's interface
+- [Mix in][] a mixin or mixin class
+
+## `base`
+
+Use the `base` modifier to **disallow implementation** of a class or mixin.
```
leafpetersen, line 36: I don't know if you want to get it into this or not, but it technically only disallows implementation outside of the current library.

src/language/class-modifiers.md
```diff
@@ -0,0 +1,240 @@
...
+4. And lastly, should it be a [class, mixin, or mixin class](#class-mixin-or-mixin-class)?
+
+## No modifier
+
+Use a `class` or `mixin` declaration without a modifier to **allow unrestricted
+access** to outside libraries. The standard behaviors apply:
+
+- [Construct][] new instances of a class
+- [Extend][] a class to create a new subtype
+- [Implement][] a class's interface
```
leafpetersen, line 31: a class or mixin's interface.

src/language/class-modifiers.md
```diff
@@ -0,0 +1,240 @@
...
+{{site.alert.end}}
+
+```dart
+base class Vehicle {
+  String make;
+  String model;
+  void moveForward(int meters) { … }
+}
+
+var myCar = Vehicle(     // Can be constructed
```
leafpetersen, line 54: The class as defined doesn't have a constructor with this signature.  Is this intentional?  It feels better to me if someone could copy this code into an IDE and observe the results themselves.

src/language/class-modifiers.md
```diff
@@ -0,0 +1,240 @@
...
+class MockVehicle     // ERROR: Cannot be implemented
+  implements Vehicle {
+    @override
+    void moveForward …
+}
+
+```
+
+Disallowing implementation makes `base` useful for:
+- Preventing users of the API from overriding the constructor, for example, to
```
leafpetersen, line 74: I don't think overriding is the right term here, constructors aren't virtual.  Something like "Ensuring that the base class constructor is called whenever an instance of a subtype of the class is created"

src/language/class-modifiers.md
```diff
@@ -0,0 +1,240 @@
...
+bypass its validation.
+- Ensuring all private members are available to subtypes of the base class, and
```
leafpetersen, line 76: available to -> available on

src/language/class-modifiers.md
```diff
@@ -0,0 +1,240 @@
...
+An interface class can be implemented and constructed.
+
+```dart
+interface class Vehicle {
+  String make;
+  String model;
+  void moveForward(int meters) { … }
+}
+
+var myCar = Vehicle(     // Can be constructed
```
leafpetersen, line 91: As above, no such constructor.

src/language/class-modifiers.md
```diff
@@ -0,0 +1,240 @@
...
+class MockVehicle     // Can be implemented     
+  implements Vehicle {
+    @override
+    void moveForward …
+}
+```
+
+Disallowing inheritance makes `interface` useful for:
+- Defining API interfaces.
+- Preventing users of the API from overriding a subset of methods.
```
leafpetersen, line 111: You could mention the fragile base class problem here (for which interfaces are one solution).

src/language/class-modifiers.md
```diff
@@ -0,0 +1,240 @@
...
+Final classes can be constructed.
+
+```dart
+final class Vehicle {
+  String make;
+  String model;
+  void moveForward(int meters) { … }
+}
+
+var myCar = Vehicle(     // Can be constructed
```
leafpetersen, line 125: As above.

src/language/class-modifiers.md
```diff
@@ -0,0 +1,240 @@
...
+}
+```
+
+Disallowing inheritance makes `interface` useful for:
+- Defining API interfaces.
+- Preventing users of the API from overriding a subset of methods.
+
+## `final` 
+
+Use the `final` modifier to **disallow extending _and_ implementing** a class.
```
leafpetersen, line 115: Might be worth saying that the transitive base property applies here as well?  That is, something like "within the same library, a final class can be extended or implemented, but the subtype is required to be marked base".  Or are you avoiding talking about the "same library" exceptions entirely?

src/language/class-modifiers.md
```diff
@@ -0,0 +1,240 @@
...
+Disallowing inheritance and implementation makes `final` useful for:
+- Closing the type hierarchy.
+- Ensuring that adding new members to the class won't risk breaking changes for 
+its consumers.
+- Ensuring access to private members on instances of the final class will never
+throw a [`noSuchMethod`][] error.
+
+## `sealed`
+
+Use the `sealed` modifier to define a class that **cannot be constructed, extended,
```
leafpetersen, line 152: This feels like the wrong lead.  I would switch the order of the first two paragraphs. That is, say something like "Use the `sealed` modifier to define a family of classes within a single library for which switches will be checked for exhaustiveness.  In order to allow the compiler to check exhaustiveness, applying `sealed` to a class will prevent it from being extended or implemented outside of the library where the class is defined."

src/language/class-modifiers.md
```diff
@@ -0,0 +1,240 @@
...
+    - [Subtyping](#sealed) by outside libraries at all?
+3. Then, does it need to be [abstract](#abstract)?
+4. And lastly, should it be a [class, mixin, or mixin class](#class-mixin-or-mixin-class)?
+
+## No modifier
+
+Use a `class` or `mixin` declaration without a modifier to **allow unrestricted
+access** to outside libraries. The standard behaviors apply:
+
+- [Construct][] new instances of a class
```
fishythefish, line 29: I found it a little confusing to have these written in the imperative. This is a list of default *capabilities*, not *obligations*. Similarly, I wouldn't say these "behaviors apply". Maybe change that line to something like "By default, you can:" or "This allows you to:"?

Grammar nit: As written, these bullets look like sentences with missing periods.

src/language/mixins.md
```diff
@@ -26,10 +26,17 @@ class Maestro extends Person [!with Musical, Aggressive, Demented!] {
 }
 {% endprettify %}
 
-To _implement_ a mixin, create a class that extends Object and
-declares no constructors.
-Unless you want your mixin to be usable as a regular class,
-use the `mixin` keyword instead of `class`.
+To _implement_ a mixin, use the `mixin` declaration. Or, if you want your mixin
```
mit-mit, line 29: "Implementing a mixin" sounds a bit confusing to me (as "implementing" to me usually suggests implementing an interface). Should we just say "define a mixin"?

I wonder if we should be a bit more prescriptive about "mixin class" and say something like "In the rare case where you need to define both a mixin and a class, you can use the `mixin class` decl..."

src/language/mixins.md
```diff
@@ -26,10 +26,17 @@ class Maestro extends Person [!with Musical, Aggressive, Demented!] {
 }
```
mit-mit, line 26: I think we also have to update the lead-in up in lines 9-10 to something like "Mixins are a way of defining code that can be reused in multiple class hierarchies" (i.e. not talk about "a class's code")? cc @leafpetersen

src/language/class-modifiers.md
eernstg, line 208: Again, I don't think it's helpful to introduce the notion of "a class that can be mixed in". The mixin which is introduced by `... mixin class MC ...` is a perfectly normal mixin, and the class is a perfectly normal class, and we don't need to give rise to questions about "how mixin application works when we apply a class that can be mixed in" etc.etc.

Perhaps:
```suggestion
A `mixin` declaration defines a [mixin][]. A `class` declaration defines a [class][]. A
`mixin class` declaration defines both a mixin and a class, with the same name and
the same type.
```

src/language/mixins.md
```diff
@@ -26,10 +26,17 @@ class Maestro extends Person [!with Musical, Aggressive, Demented!] {
...
-use the `mixin` keyword instead of `class`.
+To _implement_ a mixin, use the `mixin` declaration. Or, if you want your mixin
+to be usable as both a regular class and a mixin, use the `mixin class` declaration.
+
+{{site.alert.version-note}}
+  The `mixin class` declaration requires a language version of at least 3.0.
+  See [`class`, `mixin`, or `mixin class`?][] for more information.
+{{site.alert.end}}
+
+Mixins and mixin classes must extend `Object`
```
eernstg, line 37: It's actually a compile-time error for a mixin class declaration to have an extends clause: https://github.com/dart-lang/language/blob/master/accepted/future-releases/class-modifiers/feature-specification.md#:~:text=has%20an%20extends%20clause%2C.

Perhaps:
```suggestion
Mixins and mixin classes cannot have an extends clause. It does not make
sense for a mixin, and for the mixin class it means that its superclass is `Object`.
```

eernstg (commented):
Looks good! Added some comments here and there, many of them about mixins or about terminology.

src/language/class-modifiers.md
lrhn: Agree.
Rather than focus on the restrictions, I'd focus on intended use-cases and promises (to yourself or others).

So, when deciding whether to use class modifiers, consider the intended uses of the class, and which behavior the class needs to be able to rely on.


If the class is just an interface, and has only abstract members, it should probably be marked `abstract` and `interface`.
Examples are classes like `Sink` or `Future`, which are pure interfaces that others are free to implement.

(The `interface` modifier *can* be used to prevent access to an actual implementation, and force others to implement it from scratch, but there are other ways to achieve that too, so it's will not be a common use-case.)

With `base` you *ensure* that all subclasses inherit the implementation of this class. That allows you to safely call private members on every object of that type, something which allowing `implements` makes unsafe. To ensure that, no subclass is allowed to be implemented, and all subclasses must themselves be at least marked `base` to ensure that.

You should use `base` when you _need_ to rely subclasses inheriting behavior. You shouldn't add `base` just because extending is the *intended* use, and there is no good reason to implement them, because `base` forces subclasses to also be `base`, and subclasses may not want to prevent implementation.

Example of `base` classes are rare.

The `final` prevents subclassing entirely, allowing the author to safely add incremental changes to the API, and to call instance methods knowing that they haven't been overwritten in a third-party subclass.
(The `base` modifier gives most of the same guarantees, except for subclasses overriding instance members, and a `base` class with no public generative constructor could just as well have been `final`. The `final` modifier includes the effect of `base` and therefore all subclasses must again be marked at least `base`.)

The `sealed` modifier is only used to create a known (enumarable) set of subtypes, such that a switch over those types can be statically ensured to be exhaustive. It does not add any new restrictions on the subtypes of those direct subtypes. The sealed class itself cannot be instantiated.

src/language/class-modifiers.md
lrhn, line 36: I'd focus on the other direction: It *enforces inheritance* of the class or mixin's implementation.

Preventing implementation is a means toward that goal, not a goal in itself.
It's probably mainly a philosophical distinction, but it underlines *why* you'd add `base`. It's not to prevent implementation. If there were other ways to create a subclass which doesn't inherit the superclass implementation, preventing implementation wouldn't give you anything useful by itself.
There isn't, which is why `base` only prevents implementation. That's enough to reach its goal of ensuring that all subclasses inherit the implementation.

src/language/class-modifiers.md
lrhn, line 37: This also sounds like a `base` class cannot be `abstract`, but that's an orthogonal modifier. An `abstract base class` cannot be constructed.

I'd not discuss construction in the section about `base`, because it's mostly unrelated.

src/language/class-modifiers.md
```diff
@@ -0,0 +1,240 @@
...
+);
+
+class Car extends     // Can be extended
+  Vehicle {
+    int passengers;
+…
+}
+
+class MockVehicle     // ERROR: Cannot be implemented
+  implements Vehicle {
```
lrhn, line 66: This class is (apparently) in the same library as the `Vehicle` class, so it *can* implement `Vehicle` (but it must still be marked `base`, `final` or `sealed`, like all subtypes of a `base` declaration must.)

src/language/class-modifiers.md
lrhn: And it's needed since `String make;` is not nullable, and therefore needs to be initialized during construction.

src/language/class-modifiers.md
```diff
@@ -0,0 +1,240 @@
...
+This could be useful to define a "pure contract", or a class that doesn't update
+any preexisting state.
+
+Modifiers that are mutually exclusive cannot be combined. For example,
+`base` allows extending and disallows implementing, and `interface` allows
+implementing and disallows extending.
+
+Modifiers that are redundant cannot be combined. For example, mixins and sealed
+classes already cannot be constructed, so combining them with `abstract`, which
+disallows construction, is redundant.
```
lrhn, line 226: This way of describing it, as if all modifiers are the same and the rules for how they can be combined are complicated, gets me very confused about something that's fairly simple (to me).

A class declaration can be, in order:
* Optionally an `abstract`, describing whether the class can contain abstract members (and it also prevents instantiation).
* Optionally one of `base`, `interface`, `final` or `sealed`, describing restrictions on other libraries subtyping the class.
* Optionally a `mixin`, describing whether the declaration also introduces a mixin (aka. can be mixed in).
* and then the `class` keyword itself.

Some combinations are disallowed:
* `abstract` with `sealed` (because a `sealed` class is always implicitly `abstract`).
* `interface`, `final` or `sealed` with `mixin` (because those access modifiers prevent mixing in).

The phrasing above doesn't say that `abstract` has to come first, and it sounds like combining `abstract` and `base` is at the same level as combining `base` and `interface`. To me it isn't. There are different groups of modifiers that can be combined (almost) orthogonally.

src/language/class-modifiers.md
```diff
@@ -0,0 +1,240 @@
+---
+title: Class modifiers
+description: Modifier keywords for class declarations to control external library access.
+---
+
+Class modifiers control how a class or a mixin can be implemented, extended,
+or mixed in. 
+
+The full set of modifiers that can appear before a class declaration are `base`,
+`final`, `interface`, `sealed`, `abstract`, and `mixin`.
```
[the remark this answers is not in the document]
MaryaBelanger: I think I took the spec a little too literally by including `mixin` here, but here's my reasoning (which might just be a misunderstanding on my part):

The spec says the new terminology `mixin class` is the _modifier_ `mixin` _on_ `class`, whereas your `mixin M {}` example is a pure mixin declaration. So this page talks about the former, the `mixin` modifier on `class`, while the latter has its own dedicated page for mixin as a standalone construct. 

But even if that is correct(?), I can see that the distinction between `mixin` as a modifier and a pure mixin is too much of an implementation detail on our side and doesn't matter to users, so I'll fix that here.

src/language/class-modifiers.md
MaryaBelanger: Hmm... Considering your comments, I'm actually leaning towards making this statement even more generic, something like: "Class modifiers control how a class or mixin can be _utilized_ by libraries other than the one they're defined in."

And then letting each modifier's section define what that "utilization" means to it. Or maybe that sounds terrible, or is that technically inaccurate? 

> maybe mention exhaustiveness?

This is a tricky one. My idea is that exhaustiveness checking is technically a product of the fact that `sealed` classes' ability to be utilized by outside classes is restricted, right? So, the intro statement is technically not excluding exhaustiveness checking, even though it is omitting it.

I wanted the purpose of this page to be kept really tight and just define the class modifiers, as agnostic from other topics as possible. Which is tough, considering the purpose of `sealed`. But in the end that decision was made specifically _because_ of `sealed`: because it would be easy for the definition of exhaustiveness checking to spread (and ultimately, be weakened because there'd be no single source of truth) across three different docs (this one, switch, patterns). 

[There will be dedicated content on exhaustiveness checking](https://github.com/dart-lang/site-www/issues/4635) added to the control flow section of the language tour that'll link to and from the `sealed` section here. I'll definitely be incorporating your reviews on emphasizing exhaustiveness checking in the `sealed` section on this page. But, I'm not sure I want to mention anything that specific at the top of this page, muddying what's technically the purpose statement of the page. I might change my mind as I go through the rest of these reviews though.

src/language/class-modifiers.md
MaryaBelanger: Thanks @lrhn, this is great. I'll be incorporating your points throughout the page, especially the angle of use cases over restrictions.

src/language/class-modifiers.md
MaryaBelanger: I like "By default...", thanks for advising!

src/language/class-modifiers.md
MaryaBelanger: I really like the phrase "permission to create subtype relationships" as a summary for what all the modifiers, in general, are doing. I'll try to use that wherever it makes sense.

src/language/class-modifiers.md
MaryaBelanger: Do you think _all_ the statements about what _isn't_ restricted for each modifier should just be removed? Like this line, or "An interface class _can_ be implemented and constructed.", etc.

src/language/class-modifiers.md
MaryaBelanger: What would be the simplest and shortest way to correct this for all the examples? Ideally not distracting from the can/can't be constructed point. Something like this:
```dart
interface class Vehicle {
  String make?;
  String model?;
  void moveForward(int meters) { … }
}
// ...
var myCar = Vehicle();        // Can be constructed

```
?

src/language/class-modifiers.md
MaryaBelanger: > Is this intentional?

I just copied these examples from some presentation slides; is there a reason that it'd be intentionally written incorrectly? Maybe to prove some other point? I'm not sure. I wrote this in another reply, but would the correct way be:
```dart
interface class Vehicle {
  String make?;
  String model?;
  void moveForward(int meters) { … }
}
// ...
var myCar = Vehicle();        // Can be constructed
```

src/language/class-modifiers.md
MaryaBelanger: Sorry, should be in a different class. Does that mean that the preceding example (`class Car extends ...`) should actually be (`base class Car extends ...`) to be accurate?

src/language/class-modifiers.md
MaryaBelanger: That's perfect, thank you for simplifying it

MaryaBelanger: I think I will end up doing that, but I'm also going to add some mixin examples to this page.

src/language/class-modifiers.md
MaryaBelanger: That's a lot clearer, I think I'll rewrite the entire section to more or less match this comment

src/language/class-modifiers.md
leafpetersen: To be honest, I don't think there's a lot of value in showing an example of construction.  Construction is controlled orthogonally - it has nothing to do with `interface` or `base` etc.  It's controlled by `abstract`.  So I personally would just leave the constructor part off.  But if you feel that it's important to call out that in the absence of `abstract` an `interface class` can be constructed, then yes, what you have now works.  Or you could just leave off the `make` and `model` fields - they don't add much.

src/language/class-modifiers.md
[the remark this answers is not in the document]
leafpetersen: FWIW, I think of `mixin` as a modifier when used on a class, and I expect our users will too.

src/language/class-modifiers.md
leafpetersen: Hmm, I don't agree with the suggested rewrite here.  I don't know what it means to have a class and a mixin in scope with the same name and the same type.  I think the original phrasing is both more accurate and less confusing.

src/language/mixins.md
MaryaBelanger: That's kind of what I meant, as in extending Object is what classes do implicitly so having an extends clause would mean it's not extending Object. But yes, this is much, much more clear 😅

src/language/mixins.md
MaryaBelanger: > I wonder if we should be a bit more prescriptive about "mixin class" 

I'm going to move the section I added on the topic to the class modifiers page, so that'll cover it

src/language/class-modifiers.md
```diff
@@ -0,0 +1,240 @@
...
+```
+
+## `class`, `mixin`, or `mixin class`?
+
+The `mixin` declaration defines a [mixin][]. The `mixin` _modifier_ can only be
+applied to a `class` declaration to create a _mixin class_: a class that can be
+treated both as a class and a mixed in. Regular classes cannot be mixed in.
+
+Mixins and mixin classes are meant to be mixed in, so they cannot be combined
+with the modifiers that prevent inheritance: `interface`, `final`, and `sealed`.
```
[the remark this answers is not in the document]
MaryaBelanger: Thanks for this, especially the definition of what mixins are intended for. Unrelated to the rest of this PR, I felt that statement was missing from the original mixins page, so I added it there.

src/language/class-modifiers.md
MaryaBelanger: I unintentionally avoided talking about the same library, because I was unclear on whether that functionality was something worth talking about, but now I've got it! I'll add this, thanks

src/language/class-modifiers.md
MaryaBelanger: @leafpetersen what is the "fragile base class problem"?

src/language/class-modifiers.md
```diff
@@ -0,0 +1,240 @@
...
+
+## `sealed`
+
+Use the `sealed` modifier to define a class that **cannot be constructed, extended,
+or implementated** outside of the library where the class is defined. 
+
+The purpose of the `sealed` modifier is to enable [_exhaustiveness checking_][]
+for switch cases.
+Any possible direct subtypes of a sealed type can only exist in the same library,
+so the compiler is always aware of the extent of a sealed type's hierarchy.
```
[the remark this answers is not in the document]
MaryaBelanger: Thank you for explaining that!

src/language/class-modifiers.md
```diff
@@ -0,0 +1,245 @@
...
+## `final` 
+
+Use the `final` modifier to prevent subtyping from a class outside of the current
+library, closing the type hierarchy. 
+
+Disallowing both inheritance and implementation means `final` classes provide all
+the same guarantees of [`base`](#base) and [`interface`](#interface) classes.
+
+Final class can be extended or implemented within the
+same library, but like the `base` modifier, any subtypes must be marked `base`.
```
eernstg, line 154: Perhaps:
```suggestion
same library, but like the `base` modifier, any subtypes must be marked `base`,
`final`, or `sealed`.
```
The point would be that it is misleading to tell the reader that subtypes of a final class must be `base`. For instance, it makes sense that they would declare every subtype of a given final class `final` as well, thus creating an entirely `final` subtype hierarchy. That hierarchy is completely closed, in the sense that there will never be any subtypes that we don't know about (for instance, if the classes aren't generic then we can use `o.runtimeType == something` to deal with each of those types precisely, which is even stronger than `sealed`). So we shouldn't imply that this is impossible.

eernstg (approved):
Looks very good to me! I added a couple of extra comments where I thought the text could be clarified a bit.

src/language/class-modifiers.md
```diff
@@ -0,0 +1,240 @@
...
+
+Modifiers that are mutually exclusive cannot be combined. For example,
+`base` allows extending and disallows implementing, and `interface` allows
+implementing and disallows extending.
+
+Modifiers that are redundant cannot be combined. For example, mixins and sealed
+classes already cannot be constructed, so combining them with `abstract`, which
+disallows construction, is redundant. 
+
+The [syntax specification] documents the valid combinations and their capabilities.
```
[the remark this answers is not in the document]
lrhn: As with all writing, it depends on who the intended audience is.
I'd consider these documents as introductory. They are not exhaustive in defining or describing the language, so I would focus in the common use-cases, for the syntax, to *show* how it's used, not *specify* it.

(We have a secondary problem that we don't have a good, complete and readable exhaustive description of the language. The specification fails, at least, the "readable" part, because it mixes intricate semantic details with the "what does this language construct really do" text, and is very short on examples.)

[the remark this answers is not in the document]
MaryaBelanger: > We have a secondary problem that we don't have a good, complete and readable exhaustive description of the language

Yeah that issue is definitely on my radar. I'd like to start developing a "reference" type of documentation, especially now with patterns and class modifiers. Hopefully after this release 🤞

src/language/class-modifiers.md
```diff
@@ -0,0 +1,245 @@
...
+Only the `base` modifier can appear before a mixin declaration. The modifiers do
+not apply to other declarations like `enum`, `typedef`, or `extension`.
+
+When deciding whether to use class modifiers, consider the intended uses of the
+class, and what behaviors the class needs to be able to rely on.
+
+## No modifier
+
+Use a `class` or `mixin` declaration without a modifier to allow unrestricted
+permission to create subtype relationships from outside libraries.
```
johnpryan, line 20: To me, it's a little odd to say "from outside libraries" here, since `abstract` doesn't place restriction based on the library, it applies to all code that tries to subclass it.

```suggestion
permission to create subtype relationships from any library.
```

src/language/class-modifiers.md
```diff
@@ -0,0 +1,245 @@
+---
+title: Class modifiers
+description: Modifier keywords for class declarations to control external library access.
+---
+
+Class modifiers control how a class or a mixin can be subtyped in libraries
```
johnpryan, line 6: except for `abstract`, right?

src/language/class-modifiers.md
```diff
@@ -0,0 +1,245 @@
...
+outside of the library where it's defined.
```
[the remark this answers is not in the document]
johnpryan: Ah, thanks. It would be great to add this definition to the docs if we haven't already.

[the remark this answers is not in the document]
MaryaBelanger: We have a new Glossary page that this definition seems suited for! Thanks, I'll add that

src/language/class-modifiers.md
MaryaBelanger: Yeah... maybe moving that here was a mistake because it makes it harder to summarize the whole page. I'll fix it, but need to make sure that "outside libraries" is explicitly stated for all of the ones that it does affect

src/language/class-modifiers.md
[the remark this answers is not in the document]
MaryaBelanger: I made an issue to track the need for further reference for this topic: https://github.com/dart-lang/site-www/issues/4783

src/language/class-modifiers.md
```diff
@@ -0,0 +1,245 @@
...
+Use the `base` modifier to enforce inheritance of a class or mixin's implementation.
+
+The `base` modifier disallows implementation outside of the current library, which
+guarantees:
+
+- The base class constructor is called whenever an instance of a subtype of the
+class is created.
+- That all implemented private members exist in subtypes.
+- Adding new members to a class reduces the risk of breaking subtypes, since all
+subtypes will inherit the new member.
```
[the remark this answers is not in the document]
MaryaBelanger: Ah, I see. Thanks for catching that, very subtle!

src/language/class-modifiers.md
```diff
@@ -0,0 +1,245 @@
...
+
+class MockVehicle          // Can be implemented     
+  implements Vehicle {
+    @override
+    void moveForward ...
+}
+```
+
+To define a pure interface that others are free to implement, that cannot be
+inherited and only has abstract members, define an `abstract interface class`.
```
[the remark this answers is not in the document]
MaryaBelanger: I added the line "All members of the class must have an implementation." to the list of guarantees at the beginning of the `interface` section, if that makes sense?

src/language/class-modifiers.md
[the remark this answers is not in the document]
MaryaBelanger: Created https://github.com/dart-lang/site-www/issues/4784 to make sure this gets added and linked from here.

src/language/class-modifiers.md
johnpryan: I'm not sure... maybe the most clear explanation is to explicitly call out that `abstract` applies to all classes regardless of the library, and the rest of the modifiers control how a class or mixin can be subtyped.

src/language/class-modifiers.md
```diff
@@ -0,0 +1,258 @@
...
+
+- [Construct][] new instances of a class.
+- [Extend][] a class to create a new subtype.
+- [Implement][] a class or mixin's interface.
+- [Mix in][mixin] a mixin or mixin class.
+
+## `abstract`
+
+Use the `abstract` modifier to define a class that can’t be constructed from any
+library (it's own, as well as outside libraries).
```
Hixie, line 41: i would say the reason you use abstract isn't to prevent construction so much as to allow you to not provide a full concrete implementation of the entire interface.

src/language/class-modifiers.md
```diff
@@ -0,0 +1,258 @@
...
+Use the `base` modifier to enforce inheritance of a class or mixin's implementation.
+The `base` modifier disallows implementation outside of the current library, which
+guarantees:
+
+- The base class constructor is called whenever an instance of a subtype of the
+class is created.
+- That all implemented private members exist in subtypes.
+- Adding a new implemented member to a `base` class will not break subtypes (except
+  when they already declare a member with the same name and an incompatible
+  signature), since all subtypes will inherit the new member.
```
Hixie, line 85: that parenthetical is doing a lot of work!

src/language/class-modifiers.md
```diff
@@ -0,0 +1,258 @@
...
+
+The `base` modifier must be applied transitively to all the subtypes of a class;
+any class which implements or extends a base class must be marked `base`, `final`,
+or `sealed`.
```
Hixie, line 89: we should say why

src/language/class-modifiers.md
```diff
@@ -0,0 +1,258 @@
...
+    void moveForward ...
+}
+```
+
+## `interface`
+
+Use the `interface` modifier to define an interface. Interface classes can be
+implemented outside of the current library, but not extended. This guarantees:
+
+- All members of the class must have an implementation.
```
Hixie, line 122: that's actually quite surprising, if you're coming from other languages with an `interface` keyword. (Most other languages' `interface` keyword means what we mean by `abstract interface class`.) Might be worth acknowledging this weirdness.

src/language/class-modifiers.md
```diff
@@ -0,0 +1,258 @@
...
+- Users of the API are prevented from overriding a subset of methods.
```
Hixie, line 123: not sure what this line means

src/language/class-modifiers.md
```diff
@@ -0,0 +1,258 @@
...
+- A solution to the [fragile base class problem][], or the freedom to change the
+interface's implementation, because there are no inherited member implementations
+outside of the current library whose behavior needs to be preserved.
```
Hixie, line 126: this doesn't sound like a guarantee, it sounds like a reason to use this setup.

src/language/class-modifiers.md
```diff
@@ -0,0 +1,258 @@
...
+class MockVehicle          // Can be implemented     
+  implements Vehicle {
+    @override
+    void moveForward ...
+}
+```
+
+To define a pure interface that others are free to implement, that cannot be
+inherited but is allowed to have abstract members, define an
+`abstract interface class`.
```
Hixie, line 156: i feel like we should emphasize this combination more. maybe even having this as a top-level section before the existing `interface class`?

src/language/class-modifiers.md
```diff
@@ -0,0 +1,258 @@
...
+
+## `final` 
+
+Use the `final` modifier to prevent subtyping from a class outside of the current
+library, closing the type hierarchy. 
+Disallowing both inheritance and implementation means `final` classes provide all
+the same guarantees of [`base`](#base) and [`interface`](#interface) classes.
```
Hixie, line 163: "the same guarantees of [`base`](#base) and [`interface`](#interface) classes" is a weird framing. I mean, it's true, but I feel it's misleading. When I think of the "guarantees of base" I think "allows me to extend it" and when I think of the "guarantees of interface" I think "allows me to implement it". But "final" means neither of those.

src/language/class-modifiers.md
```diff
@@ -0,0 +1,258 @@
...
+
+Final classes can be extended or implemented within the
+same library, but like the `base` modifier, any subtypes must be marked `base`, 
+`final`, or `sealed`.
```
Hixie, line 167: we should say why

src/language/class-modifiers.md
```diff
@@ -0,0 +1,258 @@
...
+// ...
+
+return switch (Vehicle vehicle) {       // ERROR: The switch is missing a subtype of Vehicle
+  Car => 'vroom',
+  Truck => 'VROOOOMM'
+};
+```
+
+If [exhaustiveness checking][exhaustive] is not necessary for your use case, use
+[`final`](#final) instead.
```
Hixie, line 227: we should talk about how `sealed` implies `abstract` and why that is

src/language/class-modifiers.md
```diff
@@ -0,0 +1,258 @@
...
+
+[class, mixin, or mixin class]: /language/mixins#class-mixin-or-mixin-class
+[mixin]: /language/mixins
+[fragile base class problem]: https://en.wikipedia.org/wiki/Fragile_base_class
+[`noSuchMethod`]: /language/extend#nosuchmethod
+[construct]: /language/constructors
+[extend]: /language/extend
+[implement]: /language/classes#implicit-interfaces
+[factory constructor]: /language/constructors#factory-constructors
+[exhaustive]: /
```
Hixie, line 256: TODO?

Hixie: (if it was intended as a joke, it's a funny joke, but maybe one too subtle for people to get when what they're looking for is information on exhaustiveness checking!)

src/language/classes.md
```diff
@@ -14,6 +14,7 @@ has exactly one superclass, a class body can be reused in
 multiple class hierarchies.
 [Extension methods][] are a way to
 add functionality to a class without changing the class or creating a subclass.
+[Class modifiers][] are a way to restrict use of a class by libraries outside of its own.
```
Hixie, line 17: this is where we define `abstract` now, but that's not "a way to restrict use of a class by libraries outside of its own"

src/language/class-modifiers.md
eernstg: There are some sources of ambiguity here. 'All members of the class' would generate the question 'Which class?'.

You can certainly have an `abstract interface class A ...` where some members do not have an implementation. With a plain `interface class C` it is required that `C` is a concrete class (simply because we didn't say `abstract`), and in that case all its members must have an implementation.

But this is all about `abstract` or not, which means that it is confusing to mention it as if it had something to do with `interface`.

I guess the point is that we're allowed to have implementations of members in an `interface class` and we're still guaranteed (assuming no reopening) that nobody can inherit those member implementations. This means that we have more freedom to change that implementation.

In other languages you'd probably write an `interface I` and then `private class C implements I`, and that's how you would ensure that nobody can inherit the instance member implementations. With a Dart `interface class` we can do both in the same declaration. This may be seen as ugly or not, but you will at least be able to avoid duplicating a bunch of member signatures (that is: the parameter types, return type and name, for each member).

eernstg (commented):
Added a couple of extra comments.

src/language/class-modifiers.md
MaryaBelanger: Great point, some of the guarantees cancel each other out. I'll reword this

src/language/class-modifiers.md
```diff
@@ -0,0 +1,258 @@
...
+};
+```
+
+If [exhaustiveness checking][exhaustive] is not necessary for your use case, use
+[`final`](#final) instead.
+
+## Combining modifiers
+
+It is possible to combine some modifiers for layered restrictions. 
+A class declaration can be, in order:
```
atsansone, line 232: **question:** If this is in order, why is the list not an ordered list?

src/language/class-modifiers.md
```diff
@@ -0,0 +1,258 @@
...
+
+* (Optional) `abstract`, describing whether the class can contain abstract members
+and prevents instantiation.
+* (Optional) One of `base`, `interface`, `final` or `sealed`, describing
+restrictions on other libraries subtyping the class.
+* (Optional) `mixin`, describing whether the declaration can be mixed in.
+* The `class` keyword itself.
+
+Some combinations are disallowed beacause they are contradictory, redundant, or
+otherwise mutually exclusive:
```
atsansone, line 242: **issues:** Write in active voice per [GSG](https://developers.google.com/style/voice). 
Write as shorter sentences per [GSG](https://developers.google.com/style/accessibility#ease-of-reading). (Reading: 19.4 > 7.1)
**typo:** `because`, not `beacause`.
```suggestion
You can't combine some modifiers. They include the following:
```

src/language/class-modifiers.md
```diff
@@ -0,0 +1,258 @@
...
+
+* `abstract` with `sealed` (because a sealed class is always implicitly abstract).
+* `interface`, `final` or `sealed` with `mixin` (because those access modifiers
+prevent mixing in).
```
atsansone, line 246: **issue:** Avoid parentheses per [GSG](https://developers.google.com/style/parentheses). 
Write in active voice per [GSG](https://developers.google.com/style/voice).

```suggestion
* `abstract` with `sealed`. A sealed class is always abstract.
* `interface`, `final` or `sealed` with `mixin`.
    These access modifiers prevent mixing in.
```

**suggestion:** Link `sealed class` to the `sealed` anchor earlier in the page. Also, link to `mixin` if the modifiers are mentioned on that page.

atsansone (changes requested):
@MaryaBelanger : Solid work. Issues are:
* Condition, then instruction per [GSG](https://developers.google.com/style/sentence-structure).
* Write in active voice per [GSG](https://developers.google.com/style/voice).
* Avoid parentheses per [GSG](https://developers.google.com/style/parentheses). 

I provided suggestions throughout.

src/language/class-modifiers.md
munificent: The `mixin` modifier applies to code within the same library too.

src/language/classes.md
MaryaBelanger: Changed to : 
> Class modifiers allow you to control how libraries can subtype a class.

src/language/class-modifiers.md
```diff
@@ -0,0 +1,258 @@
+---
+title: Class modifiers
+description: Modifier keywords for class declarations to control external library access.
+---
+
+Class modifiers control how a class or a mixin can be subtyped, both
```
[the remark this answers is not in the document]
MaryaBelanger: This is a tough one because it's typically outside libraries (so, not "you", probably someone else) attempting the subtyping. So you'd have to say "how _other libraries_ can subtype...". BUT, it's not exclusively about _other_ libraries, because it could also be from within the same library... Maybe:

> Class modifiers control _how a class or a mixin allows subtyping,_ both
[from within its own library](#abstract), and from outside of the library where
it's defined.

At this point I'm scared to change anything that hasn't undergone technical review, but I think this is safe... 😬

src/language/class-modifiers.md
MaryaBelanger: Changed to:
> Use the `abstract` modifier to define a class that doesn't require a full, concrete
implementation of the entire interface.

src/language/class-modifiers.md
```diff
@@ -0,0 +1,258 @@
...
+[from within its own library](#abstract), and from outside of the library where
+it's defined.
+
+Modifiers prepend a class or mixin declaration.
```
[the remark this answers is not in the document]
MaryaBelanger: I don't want to make it passive, but I'm also hesitant to say "_You_ can prepend class modifiers...". I'm not sure exactly what you mean by the subject and object being backwards so I can't think of another way to say it besides "Modifiers _come before_ a class or mixin declaration"?

src/language/class-modifiers.md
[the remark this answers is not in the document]
MaryaBelanger: Shortened the sentence, not sure what the condition/instruction is here

src/language/class-modifiers.md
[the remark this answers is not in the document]
MaryaBelanger: Can't really active-voice the first bullet again since it could be anyone, most likely not "you", when we're talking about restricting external libraries. Other notes incorporated though

MaryaBelanger: Made it a sub bullt

src/language/class-modifiers.md
```diff
@@ -0,0 +1,258 @@
...
+  Truck => 'VROOOOMM'
+};
+```
+
+If [exhaustiveness checking][exhaustive] is not necessary for your use case, use
+[`final`](#final) instead.
+
+## Combining modifiers
+
+It is possible to combine some modifiers for layered restrictions.
```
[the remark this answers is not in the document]
MaryaBelanger: Didn't know second person is preferred, that'll come in handy

src/language/class-modifiers.md
MaryaBelanger: Changed to:

> Disallowing both inheritance and implementation prevents subtyping entirely.
> This guarantees:
> 
> - You can safely add incremental changes to the API.
> - You can call instance methods knowing that they haven't been overwritten in a
> third-party subclass.

src/language/class-modifiers.md
```diff
@@ -73,20 +73,20 @@ define a [factory constructor][].
...
+since all subtypes inherit the new member.
+  - This is true unless the subtype already declares a member with the same name
+  and an incompatible signature.
 
-The `base` modifier must be applied transitively to all the subtypes of a class;
-any class which implements or extends a base class must be marked `base`, `final`,
-or `sealed`. 
+Any class which implements or extends a base class must transitively be marked
+`base`, `final`, or `sealed`. This is necessary to prevent outside libraries from
+breaking the base class guarantees.
```
atsansone, line 89: **issue:** This is a little hard to grok.
```suggestion
Mark all classes derived from a base class as `base`, `final`, or `sealed`.
This prevents outside libraries from breaking the base class guarantees.
```

src/language/class-modifiers.md
```diff
@@ -116,14 +116,14 @@ base class MockVehicle          // ERROR: Cannot be implemented
...
 
-- All members of the class must have an implementation.
-- Users of the API are prevented from overriding a subset of methods.
-- A solution to the [fragile base class problem][], or the freedom to change the
-interface's implementation, because there are no inherited member implementations
-outside of the current library whose behavior needs to be preserved. 
+- When an instance method in the class calls another instance method on `this`,
+it will always invoke a known implementation of the method from the same library.
+- Other libraries can't override methods which may then be called by the interface
+class's own methods in unexpected ways. This reduces the [fragile base class problem][].
```
atsansone, line 126: **issue:** Write as shorter sentences per [GSG](https://developers.google.com/style/accessibility#ease-of-reading).
```suggestion
- When a class instance method calls another instance method on `this`,
the first method invokes a known implementation of the method from the same library.
- Other libraries can't override methods that the interface class's own
methods might call later in unexpected ways. This reduces the [fragile base class problem][].
```

src/language/class-modifiers.md
```diff
@@ -151,20 +151,31 @@ class MockVehicle          // Can be implemented
...
 ```
 
-To define a pure interface that others are free to implement, that cannot be
-inherited but is allowed to have abstract members, define an
-`abstract interface class`.
+### `abstract interface`
+
+The most common use for the `interface` modifier is to [combine](#combining-modifiers)
+it with [`abstract`](#abstract) for an `abstract interface class`.
+This defines a pure interface.
```
atsansone, line 158: **suggestion:** Write in active voice per [GSG](https://developers.google.com/style/voice).
```suggestion
To create a pure interface, [combine](#combining-modifiers) the `interface` 
and [`abstract`](#abstract)  modifiers. 
The resulting `abstract interface class` defines a pure interface. 
```

src/language/class-modifiers.md
```diff
@@ -151,20 +151,31 @@ class MockVehicle          // Can be implemented
...
-inherited but is allowed to have abstract members, define an
-`abstract interface class`.
+### `abstract interface`
+
+The most common use for the `interface` modifier is to [combine](#combining-modifiers)
+it with [`abstract`](#abstract) for an `abstract interface class`.
+This defines a pure interface. 
+
+Like an `interface` class, other libraries can implement, but cannot inherit,
+a pure interface. Like an `abstract` class, a pure interface is allowed to have
```
atsansone, line 161: **nit:** Simplify the word choice.
```suggestion
a pure interface. Like an `abstract` class, a pure interface can have
```

src/language/class-modifiers.md
```diff
@@ -194,15 +205,16 @@ class MockVehicle          // ERROR: Cannot be implemented
 
 ## `sealed`
 
-Use the `sealed` modifier to create a known, enumerable set of subtypes, so that
-a switch over those types can be statically ensured to be [_exhaustive_][exhaustive].
+To create a known, enumerable set of subtypes, so that a switch over those types
+can be statically ensured to be [_exhaustive_][exhaustive], use the `sealed` modifier.
```
atsansone, line 209: **issue:** Write as shorter sentences per [GSG](https://developers.google.com/style/accessibility#ease-of-reading).
```suggestion
To create a known, enumerable set of subtypes, use the `sealed` modifier.
This ensures [_exhaustive_][exhaustive] coverage of all `cases` included in a `switch` over those types.
```

src/language/class-modifiers.md
```diff
@@ -194,15 +205,16 @@ class MockVehicle          // ERROR: Cannot be implemented
...
 
 The `sealed` modifier prevents a class from being extended or implemented outside
-of the library where the class is defined. The compiler is aware of any possible
-direct subtypes because they can only exist in the same library.
+of the library where the class is defined. They also prevent construction,
+and are therefore implicitly [abstract](#abstract).
```
atsansone, line 213: **suggestion:** Write in active voice per [GSG](https://developers.google.com/style/voice). Not certain of my construction here, so you can ignore.
```suggestion
of the library where you define the class. You can't construct these classes,
so they have implicit [abstraction](#abstract).
```

src/language/class-modifiers.md
```diff
@@ -194,15 +205,16 @@ class MockVehicle          // ERROR: Cannot be implemented
...
 
-This allows the compiler to alert you when a switch is not exhaustively handling
-all possible subtypes in its cases:
+The compiler is aware of any possible direct subtypes because they can only exist
+in the same library. This allows the compiler to alert you when a switch does not
+exhaustively handle all possible subtypes in its cases:
```
atsansone, line 217: **suggestion:** Not sure if this is better. I'm not sure if another construction could handle "exhaustively" better. Ignore if desired.
```suggestion
As possible direct subtypes can only exist in the same library,
their limited scope makes the compiler aware of them.
This limitation allows the compiler to alert you when a `switch` does not
exhaust all possible subtypes in its `case` statements:
```

src/language/class-modifiers.md
```diff
@@ -223,27 +235,29 @@ return switch (Vehicle vehicle) {       // ERROR: The switch is missing a subtyp
 };
 ```
 
-If [exhaustiveness checking][exhaustive] is not necessary for your use case, use
-[`final`](#final) instead.
+If you don't intend users to be able to [exhaustively switch][exhaustive] over
+the subtypes of the class, or you want the freedom to add new subtypes later
+without it being a breaking API change, use [`final`](#final) instead.
```
atsansone, line 240: **issue:** Write as shorter sentences per [GSG](https://developers.google.com/style/accessibility#ease-of-reading).
```suggestion
If you don’t want [exhaustive switching]][exhaustive] or want to add subtypes later,
use [`final`](#final).
```

atsansone (approved):
Great update! A few issues, suggestions, and nits. Handle the issues and no need for another pass.

src/language/class-modifiers.md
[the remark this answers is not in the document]
munificent: > I'm not sure exactly what you mean by the subject and object being backwards

As far as I know, to prepend means to put something before another thing. So "Modifiers prepend a class or mixin declaration" means that a modifier goes out and puts something (unspecified) before a class or mixin declaration.

> "Modifiers come before a class or mixin declaration"?

👍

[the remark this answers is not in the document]
MaryaBelanger: I'll fix this in the PR for #4826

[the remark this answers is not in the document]
lrhn: (As a third-language participant, I understand "A prepends B with C" to mean Subject A takes Object B and puts Object C before it. If we skip any of the parts, it becomes unclear which one is missing. "C prepends B" can be read as any of:
* Object C is put before Object B by unknown subject. (Matches non-transitive verbs like "predates".)
* Subject C puts Object B before unknown object.
* Subject C puts unknown object before Object B.

PR #4755, opened 2023-04-12
Rearrange collections page

MaryaBelanger:
The collections page was oddly organized.

For some reason, spread and collection operators were under the list section, even though they're applicable to all collections. So I moved them out to their own section.

Also, each collection section had the same few "learn more" lines at the bottom, all linking to the same resources, again applicable to all. So I consolidated all of that at the top.

### DIFF
--- a/src/language/collections.md
+++ b/src/language/collections.md
@@ -3,6 +3,9 @@ title: Collections
 description: Summary of the different types of collections in Dart.
 ---
 
+Lists, sets, and maps are all collections in Dart. You can learn more about each
+in the [Collections][] section of the Library tour, and on the [Generics][] page.
+
 ## Lists
 
 Perhaps the most common collection in nearly every programming language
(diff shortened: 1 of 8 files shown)


src/language/collections.md
```diff
@@ -3,6 +3,9 @@ title: Collections
 description: Summary of the different types of collections in Dart.
 ---
 
+Lists, sets, and maps are all collections in Dart. You can learn more about each
```
parlough, line 6: There are other collections in `dart:core` so perhaps we should be more specific. Something like:

```suggestion
Dart has built-in support for list, set, and map collections
and declaring them through literals.
You can learn more about each
```

src/language/collections.md
```diff
@@ -3,6 +3,9 @@ title: Collections
...
+in the [Collections][] section of the Library tour, and on the [Generics][] page.
```
parlough, line 7: This text seems to direct users away from this page to learn about collections. Could we provide some more specific context on why/when they would visit those pages? Or move the links to some sort of "Other resources" header at the end of the page.

For example, could state something like "To learn more about the underlying collection types and interacting with them, check out the [Collections][] section of the Library tour.".

parlough (commented):
Thanks so much for doing this. This new structure makes a lot more sense!

I have a few small comments, let me know what you think :)

src/language/collections.md
parlough: Oh I see you moved this from each section. I think I might prefer at least the specific library tour sections be linked to in the appropriate section. Since users may be reading the "map" section specifically and want to learn about the library usages. They may not think to find this at the top of page, and this also loses the specific section links.

At least as of now, each collection type isn't its own entry in the library tour TOC either, so they may have trouble finding it:

Collections in library tour tOC

MaryaBelanger: Ok great, I was thinking the same thing. I'll move a link to each types library section back into their language sections.

src/language/collections.md
MaryaBelanger: Could we say: "Dart has built in support for declaring list, set, and map collection literals", or does that change the meaning? 

The "...and declaring them..." clause reads a little awkward so I'm trying to consolidate them, but not sure I understand the distinction between supporting the collections vs. literals.

MaryaBelanger: I might be getting farther from the point here, but would it make sense to link to the library section on "collections", and the generics page on "literals"?

parlough: I think it makes sense to link to the library tour on collections, but not as much generics on literals.

I do think supporting collections vs supporting literals are related but slightly different. Perhaps we can just say built-in support for collections, I think "built-in" implies syntactical support sufficiently.

```
Dart has built-in support for list, set, and map [collections][].
```

For the generic mention, maybe we can be more specific, since readers may not know what generics are for. Something like?

```
To learn more about configuring the types collections contain,
check out [Generics][].
```

MaryaBelanger: Thanks for iterating on this with me!

parlough (approved):
Thanks for all of those adjustments, looks good to me!

Just as a note, I pushed a change updating all old links to point to the new header.

PR #4766, opened 2023-04-14
[Dart 3] Records

### LINKED ISSUE #4176: [Records] Language feature docs
### Summary

Records are a new type/collection type/data structure. They are great for working with patterns, but don't necessitate/are not necessitated by patterns, so the main records content will be separate from patterns.

### Evaluation

1. ~Records content will go with the rest of the collection types content (happening in #4612, "Language > Types > Collection types" )~
  - ~Could also separate the collection types into their own separate pages e.g. "Language > Types > Collection types > Records"~
  - see item 12 below for new plan

2. Topics to cover in general records content:
    - Record _expression_ = record _value_  
    - ~Multiple returns~
    - Emphasize difference from other collection types: 
      - type transparency
      - heterogenous
      - named and positional fields (though not really worth mentioning until argument list spreading becomes a thing in the future?)
    - Getters for each field (immutable, no setters)
    - Type system interpretation of equality
      - "If two unrelated libraries create records with the same set of fields, the type system understands that those records are the same type even though the libraries are not coupled to each other."

3. ~Maybe add mention of multiple returns being possible [here](https://dart.dev/guides/language/language-tour#return-values) that points back to records page~

3. Patterns docs will definitely discuss destructuring multiple returns with record patterns; link back to the general records doc

4. Maybe add content to "Core libraries > Tour > dart:core > [Collections](https://dart.dev/guides/libraries/library-tour#collections)" section (what kind of topics go here vs the language tour?)
    - maybe records' interaction with `toString`? [here](https://github.com/dart-lang/language/blob/master/accepted/future-releases/records/records-feature-specification.md#tostring)
    - compare other collection types library vs language tour sections to each other to understand what content goes on one page vs the other
    - [this content](https://github.com/dart-lang/language/blob/master/accepted/future-releases/records/records-feature-specification.md#identity) is kind of similar to what the other collection types have there
    - Other collections have examples of the methods they have, and then point to the API docs "for more"
      - here's records' API docs: https://api.dart.dev/dev/3.0.0-334.0.dev/dart-core/Record-class.html
      - it looks like only `toString` and `noSuchMethod` (in the API and in the spec), double check (find the commits)
      - The "Common collection methods" covers methods that apply to all collections, but again I can't find anything in the records spec referring to any of these.
 
6. Maybe something to add to Type System content, since records are the only data structure that the type system can "see through"
    - either goes with type system content or with record content

8. There's a "[testing functions for equality](https://dart.dev/guides/language/language-tour#testing-functions-for-equality)" section in the language tour; should there be a similar section for records to convey [this information](https://github.com/dart-lang/language/blob/master/accepted/future-releases/records/records-feature-specification.md#equality)?

10. ~[`on` clause](https://github.com/dart-lang/language/blob/master/accepted/future-releases/records/records-feature-specification.md#ambiguity-with-on-clauses) semantics change, add to language tour ["try" section](https://dart.dev/guides/language/language-tour#catch)~
- very contrived, no changes needed

11. ~VERY minor, probably won't do: maybe add [metadata disambiguation](https://github.com/dart-lang/language/blob/master/accepted/future-releases/records/records-feature-specification.md#ambiguity-with-metadata-annotations) (no space) to metadata language tour section~
- the metadata page already doesn't put whitespace between annotations and argument lists, so nothing to change
- couldn't find anywhere in the docs that used the incorrect syntax, nothing to change

12. ~I'm second guessing including records with the other collection types, because they're not part of [`dart:collection`](https://api.dart.dev/dev/3.0.0-436.0.dev/dart-collection/dart-collection-library.html) (i think....), and they're not [Iterables](https://api.dart.dev/dev/3.0.0-436.0.dev/dart-core/Iterable-class.html) (i think....).~
- going to make Records its own page

13. ~Add Records to list at the top of [Built in types](https://dart.dev/language/built-in-types) page~

#### TBD

10. ~Do records get into collection literals/generics like lists maps and sets do? Do we need records content in the [generics section](https://dart.dev/guides/language/language-tour#generics) of the language tour?~
    - Parker says no, related to the fact that records can maintain type transparency and the others can't...
11. ~Under “[Syntax](https://github.com/dart-lang/language/blob/master/accepted/future-releases/records/records-feature-specification.md#syntax)” section in spec (also just all over the spec), compile time errors associated with records. Where does that info go?~
    - ~asked Brian, assuming diagnostics have it covered~
14. Do upper/lower bounds have a place in user facing docs?

### Resources

- https://github.com/dart-lang/language/blob/master/accepted/future-releases/records/records-feature-specification.md

### LINKED ISSUE #4692: [Release] Record-related extensions on future
### Summary

Records allow multiple returns, so `Future` functionality extends to return multiple futures and maintain types of values (wip, still learning about it).

### Evaluation
WIP:
- Language content on `Future`: https://dart.dev/language/async#handling-futures
- Library content on `Future`: https://dart.dev/guides/libraries/library-tour#future
- Codelab, probably don't need to update: https://dart.dev/codelabs/async-await
- Maybe tie in to new [records](https://github.com/dart-lang/site-www/issues/4176) content as well

Technically dart 3 but probably doesn't need to be documented on the site immediately (unless you think otherwise @lrhn)

### Resources

- sdk tracking issue, there's multiple related core API additions : https://github.com/dart-lang/sdk/issues/49928
- https://dart-review.googlesource.com/c/sdk/+/288903

MaryaBelanger:
Reviewers: 
- New records page
- Added library tour section on new future extensions
- Added mention of records on Built-in types page
- Added mention of records to Return values section of Functions page
---
Major issues / questions:
- Composite map key example for records in value semantics section?
- Future extensions:
  - Unsure about attempt at examples
  - Unsure whether I captured the purpose / the difference between regular `wait` and the new `wait` properties
- Should there be a section on Records alone in the Library tour? I didn't add one because there's not a lot to say based on the [API docs](https://api.dart.dev/stable/2.19.6/dart-core/Record-class.html)

---
TODO:
- Change code to actual code excerpts
- In `/language/records#record-fields`, need to change link on "Patterns" to section on destructuring records, once that's written

--- 
Fixes #4176 
Fixes #4692

### DIFF
--- a/src/_guides/libraries/library-tour.md
+++ b/src/_guides/libraries/library-tour.md
@@ -1304,15 +1304,18 @@ or [record]({{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-async/Futu
 of futures.
 
 These extensions return a future with the resulting values of all the provided
-futures. Unlike `Future.wait()`, if any future in the collection completes with
+futures. Unlike `Future.wait`, if any future in the collection completes with
 an error, `wait` completes with a [`ParallelWaitError`][] that allows the caller
-to handle individual errors and dispose successful results if necessary:
+to handle individual errors and dispose successful results if necessary.
+
+A common use case is wanting to know when all the futures complete, and handle
+any errors, but not necessarily caring about the result of each individual future:
 
 ```dart
 void main() async {
-  Future<bool> delete() async =>  ...
-  Future<bool> copy() async =>  ...
-  Future<bool> errorResult() async =>  ...
+  Future<void> delete() async =>  ...
+  Future<void> copy() async =>  ...
+  Future<void> errorResult() async =>  ...
 
   try {      // Wait for each future in list.
     
@@ -1320,20 +1323,20 @@ void main() async {
 
     } on ParallelWaitError<List<bool?>, List<AsyncError?>> catch (e) {
 
-    print(e.values[0]);    // Prints result of future
-    print(e.values[1]);    // Prints result of future
-    print(e.values[2]);    // Prints null
+    print(e.values[0]);    // Prints successful future
+    print(e.values[1]);    // Prints successful future
+    print(e.values[2]);    // Prints null when the result is an error
 
-    print(e.errors[0]);    // Prints null
-    print(e.errors[1]);    // Prints null
+    print(e.errors[0]);    // Prints null when the result is succesful
+    print(e.errors[1]);    // Prints null when the result is succesful
     print(e.errors[2]);    // Prints error
   }
 
 }
 ```
 
-Waiting on a record of futures enables the same functionality, with the additional 
-benefit that the futures can be of different types:
+Waiting on a _record_ of futures enables the same functionality, with the additional 
+benefit that the futures can be of different types: 
 
 ```dart
 void main() async {
@@ -1349,8 +1352,18 @@ void main() async {
       (AsyncError?, AsyncError?, AsyncError?)> catch (e) {
     // ...
     }
+
+  // Do something with the results:
+  var deleteInt  = result.$1;
+  var copyString = result.$2;
+  var errorBool  = result.$3;
 }
 ```
+
+Waiting for a record of futures should be used when you need the individual
+result values from each future
+(otherwise, just use `wait` on an iterable as described in the previous example).
+
 
 ### Stream
 
--- a/src/_guides/libraries/library-tour.md
+++ b/src/_guides/libraries/library-tour.md
@@ -1293,9 +1293,9 @@ await Future.wait([
 print('Done with all the long steps!');
 ```
 
-`Future.wait()` returns a future which will complete once all the provided
-futures have completed, either with their results, or with an error if any of
-the provided futures fail. 
+`Future.wait()` returns a future which completes once all the provided
+futures have completed. It completes either with their results,
+or with an error if any of the provided futures fail. 
 
 #### Handling errors for multiple futures
 
@@ -1303,23 +1303,25 @@ You can also wait for parallel operations on an [iterable]({{site.dart-api}}/{{s
 or [record]({{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-async/FutureRecord2/wait.html)
 of futures.
 
-These extensions return a future with the resulting values of all the provided
-futures. Unlike `Future.wait`, if any future in the collection completes with
-an error, `wait` completes with a [`ParallelWaitError`][] that allows the caller
-to handle individual errors and dispose successful results if necessary.
+These extensions return a future with the resulting values of all provided
+futures. Unlike `Future.wait`, they also let you handle errors. 
 
-A common use case is wanting to know when all the futures complete, and handle
-any errors, but not necessarily caring about the result of each individual future:
+If any future in the collection completes with an error, `wait` completes with a
+[`ParallelWaitError`][]. This allows the caller to handle individual errors and
+dispose successful results if necessary.
+
+Use `wait` on an _iterable_ of futures when you _don't_ need the result values
+from each individual future:
 
 ```dart
 void main() async {
   Future<void> delete() async =>  ...
   Future<void> copy() async =>  ...
   Future<void> errorResult() async =>  ...
-
-  try {      // Wait for each future in list.
-    
-    var results = await [delete(), copy(), errorResult()].wait;    // List of futures
+  
+  try {
+    // Wait for each future in a list, returns a list of futures:
+    var results = await [delete(), copy(), errorResult()].wait;
 
     } on ParallelWaitError<List<bool?>, List<AsyncError?>> catch (e) {
 
@@ -1335,8 +1337,8 @@ void main() async {
 }
 ```
 
-Waiting on a _record_ of futures enables the same functionality, with the additional 
-benefit that the futures can be of different types: 
+Use `wait` on a _record_ of futures when you _do_ need the individual result values
+from each future. This also provides the additional benefit that the futures can be of different types:
 
 ```dart
 void main() async {
@@ -1345,7 +1347,7 @@ void main() async {
   Future<bool> errorResult() async =>  ...
 
   try {    
-    // Record of futures
+    // Wait for each future in a record, returns a record of futures:
     (int, String, bool) result = await (delete(), copy(), errorResult()).wait;
   
   } on ParallelWaitError<(int?, String?, bool?),
@@ -1359,11 +1361,6 @@ void main() async {
   var errorBool  = result.$3;
 }
 ```
-
-Waiting for a record of futures should be used when you need the individual
-result values from each future
-(otherwise, just use `wait` on an iterable as described in the previous example).
-
 
 ### Stream
 
--- a/src/_guides/libraries/library-tour.md
+++ b/src/_guides/libraries/library-tour.md
@@ -1293,6 +1293,60 @@ await Future.wait([
 print('Done with all the long steps!');
 ```
 
+`Future.wait()` returns a future which will complete once all the provided
+futures have completed, either with their results, or with an error if any of
+the provided futures fail. 
+
+#### Handling errors for multiple futures
+
+You can also wait for parallel operations on an [iterable]({{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-async/FutureIterable/wait.html)
+or [record]({{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-async/FutureRecord2/wait.html)
+of futures.
+
+These extensions return a future with the resulting values of the provided futures.
+Unlike `Future.wait()`, if any future in the collection completes with an error,
+`wait` completes with a [`ParallelWaitError`][] that allows the caller to
+handle individual errors and dispose successful results if necessary:
+
+```dart
+void main() async {
+  Future<bool> delete() async =>  ...
+  Future<bool> copy() async =>  ...
+  Future<bool> causesError() async =>  ...
+
+  try {
+    await [delete(), copy(), causesError()].wait;    // List of futures
+    Expect.fail("Didn't throw");
+  } on ParallelWaitError<List<bool?>, List<AsyncError?>> catch (e) {   
+    print(e.values[0]);    // Prints result of future
+    print(e.values[1]);    // Prints result of future
+    print(e.values[2]);    // Prints null
+
+    print(e.errors[0]);    // Prints null
+    print(e.errors[1]);    // Prints null
+    print(e.errors[2]);    // Preints error
+  }
+}
+```
+
+Waiting on a record of futures enables the same functionality, with the additional 
+feature that the futures can be of different types:
+
+```dart
+void main() async {
+  Future<int>    delete() async =>  ...
+  Future<String> copy() async =>  ...
+  Future<bool>   causesError() async =>  ...
+
+  try {
+    await (delete(), copy(), causesError()).wait;    // Record of futures
+    Expect.fail("Didn't throw");
+  } on ParallelWaitError<(int?, String?, bool?),
+      (AsyncError?, AsyncError?, AsyncError?)> catch (e) {
+    // ...
+    }
+}
+```
 
 ### Stream
 
@@ -1757,6 +1811,7 @@ To learn more about the Dart language, see the
 [NativeFinalizer]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-ffi/NativeFinalizer-class.html
 [NoSuchMethodError]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-core/NoSuchMethodError-class.html
 [Object]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-core/Object-class.html
+[`ParallelWaitError`]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-async/ParallelWaitError-class.html
 [Pattern]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-core/Pattern-class.html
 [Random]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-math/Random-class.html
 [RegExp]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-core/RegExp-class.html
--- a/src/_guides/libraries/library-tour.md
+++ b/src/_guides/libraries/library-tour.md
@@ -1303,44 +1303,48 @@ You can also wait for parallel operations on an [iterable]({{site.dart-api}}/{{s
 or [record]({{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-async/FutureRecord2/wait.html)
 of futures.
 
-These extensions return a future with the resulting values of the provided futures.
-Unlike `Future.wait()`, if any future in the collection completes with an error,
-`wait` completes with a [`ParallelWaitError`][] that allows the caller to
-handle individual errors and dispose successful results if necessary:
+These extensions return a future with the resulting values of all the provided
+futures. Unlike `Future.wait()`, if any future in the collection completes with
+an error, `wait` completes with a [`ParallelWaitError`][] that allows the caller
+to handle individual errors and dispose successful results if necessary:
 
 ```dart
 void main() async {
   Future<bool> delete() async =>  ...
   Future<bool> copy() async =>  ...
-  Future<bool> causesError() async =>  ...
+  Future<bool> errorResult() async =>  ...
 
-  try {
-    await [delete(), copy(), causesError()].wait;    // List of futures
-    Expect.fail("Didn't throw");
-  } on ParallelWaitError<List<bool?>, List<AsyncError?>> catch (e) {   
+  try {      // Wait for each future in list.
+    
+    var results = await [delete(), copy(), errorResult()].wait;    // List of futures
+
+    } on ParallelWaitError<List<bool?>, List<AsyncError?>> catch (e) {
+
     print(e.values[0]);    // Prints result of future
     print(e.values[1]);    // Prints result of future
     print(e.values[2]);    // Prints null
 
     print(e.errors[0]);    // Prints null
     print(e.errors[1]);    // Prints null
-    print(e.errors[2]);    // Preints error
+    print(e.errors[2]);    // Prints error
   }
+
 }
 ```
 
 Waiting on a record of futures enables the same functionality, with the additional 
-feature that the futures can be of different types:
+benefit that the futures can be of different types:
 
 ```dart
 void main() async {
-  Future<int>    delete() async =>  ...
+  Future<int> delete() async =>  ...
   Future<String> copy() async =>  ...
-  Future<bool>   causesError() async =>  ...
+  Future<bool> errorResult() async =>  ...
 
-  try {
-    await (delete(), copy(), causesError()).wait;    // Record of futures
-    Expect.fail("Didn't throw");
+  try {    
+    // Record of futures
+    (int, String, bool) result = await (delete(), copy(), errorResult()).wait;
+  
   } on ParallelWaitError<(int?, String?, bool?),
       (AsyncError?, AsyncError?, AsyncError?)> catch (e) {
     // ...
--- a/src/_guides/libraries/library-tour.md
+++ b/src/_guides/libraries/library-tour.md
@@ -1310,8 +1310,8 @@ If any future in the collection completes with an error, `wait` completes with a
 [`ParallelWaitError`][]. This allows the caller to handle individual errors and
 dispose successful results if necessary.
 
-Use `wait` on an _iterable_ of futures when you _don't_ need the result values
-from each individual future:
+When you _don't_ need the result values from each individual future,
+use `wait` on an _iterable_ of futures:
 
 ```dart
 void main() async {
@@ -1337,8 +1337,9 @@ void main() async {
 }
 ```
 
-Use `wait` on a _record_ of futures when you _do_ need the individual result values
-from each future. This also provides the additional benefit that the futures can be of different types:
+When you _do_ need the individual result values from each future,
+use `wait` on a _record_ of futures.
+This provides the additional benefit that the futures can be of different types:
 
 ```dart
 void main() async {
--- a/src/language/records.md
+++ b/src/language/records.md
@@ -13,21 +13,22 @@ functions, and store them in lists.
 
 ## Record syntax
 
-Records are comma-delimited lists of named or positional fields,
+_Records expressions_ are comma-delimited lists of named or positional fields,
 enclosed in parenthesis:
 
 ```dart
 var record = ('first', a: 2, b: true, 'last');    // Record expression
 ```
 
-Named fields inside record type annotations go inside a brace-delimited section
-of type and name pairs:
+Named fields inside _record type annotations_ go inside a brace-delimited section
+of type and name pairs, after all positional fields:
 
 ```dart
-(Sring, {int a, bool b}, String?) record;    // Record type annotation
+(String, String?, {int a, bool b}) record;    // Record type annotation
 ```
-Naming fields in record type annotations is optional for documentation purposes.
-The names don't need to match any record expression corresponding to the type.
+Naming fields in record type annotations is optional for documentation purposes,
+but the names of named fields are not optional. For more information,
+see the [Record equality](#record-equality) section.
 
 ## Record fields
 
@@ -38,6 +39,8 @@ Named fields expose getters of the same name. Positional fields expose getters
 of the name `$<position>`, skipping named fields:
 
 ```dart
+var record = ('first', a: 2, b: true, 'last');
+
 print(record.$1);    // Prints '123'
 print(record.a);     // Prints 2
 print(record.b);     // Prints true
@@ -52,17 +55,19 @@ To streamline record field access even more, see the page on [Patterns][].
 
 ## Record types
 
-There is no explicit record type; records are
-structurally typed based on the types of their fields.
+There is no type declaration for individual record types. Records are structurally
+typed based on the types of their fields. A record's _shape_ (the set of its fields,
+the fields' types, and their names, if any) uniquely determines the type of a record. 
 
 Each field in a record has its own type, which can be different from other
-fields' types in the same record. Records are transparent to the type system.
-The type system remembers and applies each fields' type wherever it is
-accessed from the record:
+fields' types in the same record. The type system is aware of each
+field's type wherever it is accessed from the record:
 
 ```dart
-var oneField = record.$1;       // oneField is a String
-var anotherField = record.a;    // anotherField is an int 
+(num, Object) pair = (42, "a");
+
+var first = pair.$1;     // static type `num`, runtime type `int`
+var second = pair.$2;    // static type `Object`, runtime type `String`
 ```
 
 If two unrelated libraries create records with the same set of fields,
@@ -73,48 +78,65 @@ libraries are not coupled to each other.
 
 Two records are equal if they have the same _shape_ (set of fields),
 and their corresponding fields have the same values.
-Named field order does not affect a record's shape:
+Since named field order is not part of a record's shape, the order of named
+fields does not affect equality:
 
 ```dart
-var a = (x: 1, 2);
-var b = (2, x: 1);
-assert(a == b);        // True
+(int x, int y, int z) point = (1, 2, 3);
+(int r, int g, int b) color = (1, 2, 3);
+
+// OK:
+point = color;
+```
+
+```dart
+({int x, int y, int z}) point = (x: 1, y: 2, z: 3);
+({int r, int g, int b}) color = (r: 1, g: 2, b: 3);
+
+// Error, different types:
+point = color;
 ```
 
 Records automatically define `hashCode` and `==` methods structurally based on
 their fields.
 
-{% comment %}
-    Example of record as a composite hash key in map? Supposedly a benefit of
-    record value semantice.
-{% endcomment %}
-
 ## Multiple returns
 
-Records allow functions to return multiple values bundled together:
+Records allow functions to return multiple values bundled together. Using [pattern matching][], you can retrieve record values from a return by destructuring them directly into local variables:
+
+{% comment %}
+    TODO: link to patterns page, specifically records destructuring section.
+{% endcomment %}
 
 ```dart
+// Returns multiple values in a record:
 (String, int) userInfo(Map<String, dynamic> json) {
   return (json['name'] as String, json['age'] as int);
 }
 
-main() {
+void main() {
+  // Destructures using a record pattern:
+  var (name, age) = userInfo(json);
+
+  /* Equivalent to:
   var info = userInfo(json);
   var name = info.$1;
   var age  = info.$2;
+  */
 }
 ```
 
-The only ways to accomplish multiple returns otherwise are
+Other ways to accomplish multiple returns are
 either to create a class (which is verbose), or use
-another collection type like List or Map (which lose type safety).
+another collection type like `List` or `Map` (which loses type safety).
 
 {{site.alert.note}}
   Records' multiple-return and heterogenous-type characteristics enable
-  parallelization of Futures of different types, which you can read about in the
+  parallelization of futures of different types, which you can read about in the
   [Library tour][].
 {{site.alert.end}}
 
 [collection types]: /language/collections
 [Patterns]: /language
+[pattern matching]: /language
 [Library tour]: /guides/libraries/library-tour#handling-errors-for-multiple-futures
--- a/src/language/records.md
+++ b/src/language/records.md
@@ -1,0 +1,120 @@
+---
+title: Records
+description: Summary of the record data structure in Dart.
+toc: false
+---
+
+Records are an anonymous, immutable, aggregate type. Like other [collection types][], 
+they let you bundle multiple objects into a single object. Unlike other collection 
+types, records are fixed-sized, heterogeneous, and typed.
+
+Records are real values; you can store them in variables, pass them to and from
+functions, and store them in lists.
+
+## Record syntax
+
+Records are comma-delimited lists of named or positional fields,
+enclosed in parenthesis:
+
+```dart
+var record = ('first', a: 2, b: true, 'last');    // Record expression
+```
+
+Named fields inside record type annotations go inside a brace-delimited section
+of type and name pairs:
+
+```dart
+(Sring, {int a, bool b}, String?) record;    // Record type annotation
+```
+Naming fields in record type annotations is optional for documentation purposes.
+The names don't need to match any record expression corresponding to the type.
+
+## Record fields
+
+Record fields are accessible through built-in getters. Records are immutable,
+so fields do not have setters. 
+
+Named fields expose getters of the same name. Positional fields expose getters
+of the name `$<position>`, skipping named fields:
+
+```dart
+print(record.$1);    // Prints '123'
+print(record.a);     // Prints 2
+print(record.b);     // Prints true
+print(record.$2);    // Prints 'last'
+```
+
+To streamline record field access even more, see the page on [Patterns][].
+
+{% comment %}
+    TODO: link to patterns page, specifically records destructuring section.
+{% endcomment %}
+
+## Record types
+
+There is no explicit record type; records are
+structurally typed based on the types of their fields.
+
+Each field in a record has its own type, which can be different from other
+fields' types in the same record. Records are transparent to the type system.
+The type system remembers and applies each fields' type wherever it is
+accessed from the record:
+
+```dart
+var oneField = record.$1;       // oneField is a String
+var anotherField = record.a;    // anotherField is an int 
+```
+
+If two unrelated libraries create records with the same set of fields,
+the type system understands that those records are the same type even though the
+libraries are not coupled to each other.
+
+## Record equality
+
+Two records are equal if they have the same _shape_ (set of fields),
+and their corresponding fields have the same values.
+Named field order does not affect a record's shape:
+
+```dart
+var a = (x: 1, 2);
+var b = (2, x: 1);
+assert(a == b);        // True
+```
+
+Records automatically define `hashCode` and `==` methods structurally based on
+their fields.
+
+{% comment %}
+    Example of record as a composite hash key in map? Supposedly a benefit of
+    record value semantice.
+{% endcomment %}
+
+## Multiple returns
+
+Records allow functions to return multiple values bundled together:
+
+```dart
+(String, int) userInfo(Map<String, dynamic> json) {
+  return (json['name'] as String, json['age'] as int);
+}
+
+main() {
+  var info = userInfo(json);
+  var name = info.$1;
+  var age  = info.$2;
+}
+```
+
+The only ways to accomplish multiple returns otherwise are
+either to create a class (which is verbose), or use
+another collection type like List or Map (which lose type safety).
+
+{{site.alert.note}}
+  Records' multiple-return and heterogenous-type characteristics enable
+  parallelization of Futures of different types, which you can read about in the
+  [Library tour][].
+{{site.alert.end}}
+
+[collection types]: /language/collections
+[Patterns]: /language
+[Library tour]: /guides/libraries/library-tour#handling-errors-for-multiple-futures
--- a/src/language/records.md
+++ b/src/language/records.md
@@ -1,7 +1,6 @@
 ---
 title: Records
 description: Summary of the record data structure in Dart.
-toc: false
 ---
 
 Records are an anonymous, immutable, aggregate type. Like other [collection types][], 
@@ -14,15 +13,16 @@ functions, and store them in lists.
 ## Record syntax
 
 _Records expressions_ are comma-delimited lists of named or positional fields,
-enclosed in parenthesis:
+enclosed in parentheses:
 
 ```dart
-var record = ('first', a: 2, b: true, 'last');    // Record expression
+// Record expression:
+var record = ('first', a: 2, b: true, 'last');
 ```
 
-_Record type annotations_ are comma-delimited lists of types enclosed in parenthesis.
-The `(int, int)` statements here are record type annotations used as a return
-type and as parameter types:
+_Record type annotations_ are comma-delimited lists of types enclosed in parentheses.
+The `(int, int)` statements in the following example are record type annotations
+used as a return type and as parameter types:
 
 ```dart
 (int, int) swap((int, int) record) {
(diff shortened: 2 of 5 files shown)


src/language/records.md
```diff
@@ -0,0 +1,120 @@
...
+description: Summary of the record data structure in Dart.
+toc: false
+---
+
+Records are an anonymous, immutable, aggregate type. Like other [collection types][], 
+they let you bundle multiple objects into a single object. Unlike other collection 
+types, records are fixed-sized, heterogeneous, and typed.
+
+Records are real values; you can store them in variables, pass them to and from
+functions, and store them in lists.
```
munificent, line 12: This is a great intro.

src/language/records.md
```diff
@@ -0,0 +1,120 @@
...
+```dart
+var record = ('first', a: 2, b: true, 'last');    // Record expression
+```
+
+Named fields inside record type annotations go inside a brace-delimited section
+of type and name pairs:
+
+```dart
+(Sring, {int a, bool b}, String?) record;    // Record type annotation
+```
```
munificent, line 28: The comments in the code samples help, but I worry this is still too subtle of a jump from a record *expression* to a record *type annotation*. Because it jumps from positional fields to named fields, it's easy to overlook that it's also moving from an expression to a type annotation too.

I'd probably insert an example of a record type annotation with only positional fields first and explain that it's a type annotation and where you might use it.

src/language/records.md
```diff
@@ -0,0 +1,120 @@
...
+Naming fields in record type annotations is optional for documentation purposes.
```
munificent, line 29: The names of positional fields are optional and only for documentation, but the names of named fields (I admit that's confusing) are not optional. So these two record types are equivalent as far as the type checker is concerned:

```dart
(int x, int y, int z) point = (1, 2, 3);
(int r, int g, int b) color = (1, 2, 3);

// OK:
point = color;
```

But these are not:

```dart
({int x, int y, int z}) point = (x: 1, y: 2, z: 3);
({int r, int g, int b}) color = (r: 1, g: 2, b: 3);

// Error, different types:
point = color;
```

(This is the same for function parameters, for what it's worth.)

src/language/records.md
```diff
@@ -0,0 +1,120 @@
...
+## Record fields
+
+Record fields are accessible through built-in getters. Records are immutable,
+so fields do not have setters. 
+
+Named fields expose getters of the same name. Positional fields expose getters
+of the name `$<position>`, skipping named fields:
+
+```dart
+print(record.$1);    // Prints '123'
```
munificent, line 41: Might be worth repeating the declaration of `record` here in case it's not clear that it's coming from the previous example.

src/language/records.md
```diff
@@ -0,0 +1,120 @@
...
+
+To streamline record field access even more, see the page on [Patterns][].
+
+{% comment %}
+    TODO: link to patterns page, specifically records destructuring section.
+{% endcomment %}
+
+## Record types
+
+There is no explicit record type; records are
```
munificent, line 55: It's probably more accurate to say that there are a plethora of record types, one for each unique set of field names, positions, and field types. So `(int, int)` is a record type, `({x: int, y: int})` is another record type, etc.

Also, all of those record types inherit from a built in type named `Record`. So there sort of is "an" explicit record type, though `Record` isn't useful for much. (It's directly analogous to `Function` and function types.)

I would maybe word this like:

> Records are structurally typed: the set of the fields, their types, and their names, uniquely determines the type of a record. Two records with the same set of field names and types have the same type.

src/language/records.md
```diff
@@ -0,0 +1,120 @@
...
+accessed from the record:
+
+```dart
+var oneField = record.$1;       // oneField is a String
+var anotherField = record.a;    // anotherField is an int 
+```
+
+If two unrelated libraries create records with the same set of fields,
+the type system understands that those records are the same type even though the
+libraries are not coupled to each other.
```
munificent, line 70: 👍

munificent (approved):
A few minor things here and there but this looks great.

src/language/records.md
lrhn: And *that* is where structural types differ from nominative types. :+1: indeed!

src/language/records.md
```diff
@@ -0,0 +1,120 @@
...
+## Multiple returns
+
+Records allow functions to return multiple values bundled together:
+
+```dart
+(String, int) userInfo(Map<String, dynamic> json) {
+  return (json['name'] as String, json['age'] as int);
+}
+
+main() {
```
lrhn, line 101: I'd add `void ` in front of `main`. Users get yelled at by lints if they forget the return type of top-level functions, so we should be setting a good example (when it doesn't cost more than that).

If anything, I think readers will visually pattern-match `void main() {` more easily than just `main() {`, because they are more used to the former.

src/language/records.md
```diff
@@ -0,0 +1,120 @@
...
+  var info = userInfo(json);
+  var name = info.$1;
+  var age  = info.$2;
+}
+```
```
lrhn, line 106: I'd risk showing an example of pattern matching here, as an appetite wetter for linking to patterns.
After all, we intend pattern matching to be primary way to destructure a record, so I think we should mention it in this document too. Maybe even prominently. Record destructuring using patterns *is* a record feature (too).

Maybe:
> or, using the recommended pattern destructuring approach:
> ```dart
> void main() {
>   var (name, age) = userInfo(json);
> }
> ```

src/language/records.md
```diff
@@ -0,0 +1,120 @@
...
+
+The only ways to accomplish multiple returns otherwise are
```
lrhn, line 108: I'd drop "only" here. Maybe "existing". And I'd drop "otherwise" too, it doesn't read well where it's placed, and the sentence works without it if using "existing".

src/language/records.md
```diff
@@ -0,0 +1,120 @@
...
+either to create a class (which is verbose), or use
+another collection type like List or Map (which lose type safety).
```
lrhn, line 110: I'd use code quotes on the source names: "like \`List\` or \`Map\`"

lrhn, line 110: lose -> loses  

Maybe?
(I read "which" as referring to the singular act of "using something". Consistent with how "which is verbose" uses the singular "is".)

src/_guides/libraries/library-tour.md
```diff
@@ -1293,6 +1293,60 @@ await Future.wait([
...
+handle individual errors and dispose successful results if necessary:
+
+```dart
+void main() async {
+  Future<bool> delete() async =>  ...
+  Future<bool> copy() async =>  ...
+  Future<bool> causesError() async =>  ...
+
+  try {
+    await [delete(), copy(), causesError()].wait;    // List of futures
```
lrhn, line 1318: This code is not using the result value. 
It's the first/only example of using `await iterable.wait`, so I think it should show a *full* use-case.

It might be the only example of using the `.wait` getter some people will see.

Consider:

  bool success;
  try {
    /// Wait for each future in list.
    var results = await [delete(), copy(), causesError()].wait;
    success = !results.contains(false);
  } on ParallelWaitError<List<bool?>, List<AsyncError?>> catch (e) {
    success = false;
    // print ... all the same prints.
  }
  if (success) print("Success");
```

File: src/_guides/libraries/library-tour.md [hunk 12]
```diff
@@ -1293,6 +1293,60 @@ await Future.wait([
...
+feature that the futures can be of different types:
+
+```dart
+void main() async {
+  Future<int>    delete() async =>  ...
+  Future<String> copy() async =>  ...
+  Future<bool>   causesError() async =>  ...
+
+  try {
+    await (delete(), copy(), causesError()).wait;    // Record of futures
```
lrhn, line 1342: Again, consider capturing the values, even if it's just:
```dart
  (int, String, bool) result = await (delete(), copy(), errorResult()).wait;
```

The fact that record-`.wait` allows different types of results 
is it's biggest feature, so it bears highlighting.

lrhn (approved):
This is looking very good!

src/_guides/libraries/library-tour.md
```diff
@@ -1293,6 +1293,60 @@ await Future.wait([
...
+```
+
+Waiting on a record of futures enables the same functionality, with the additional 
+feature that the futures can be of different types:
+
+```dart
+void main() async {
+  Future<int>    delete() async =>  ...
+  Future<String> copy() async =>  ...
+  Future<bool>   causesError() async =>  ...
```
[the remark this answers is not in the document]
MaryaBelanger: I like it too, but I'll skip it for now in favor of a wider discussion

src/language/records.md
```diff
@@ -0,0 +1,120 @@
...
+
+```dart
+var record = ('first', a: 2, b: true, 'last');    // Record expression
+```
+
+Named fields inside record type annotations go inside a brace-delimited section
+of type and name pairs:
+
+```dart
+(Sring, {int a, bool b}, String?) record;    // Record type annotation
```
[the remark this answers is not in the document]
MaryaBelanger: Nice, didn't know that!

src/language/records.md
MaryaBelanger: Oh ok, thanks for explaining. I might just use this example to illustrate that point.

src/language/records.md
MaryaBelanger: I think "existing" is dating the documentation too much; it'll sound weird once records have also existed for a while. But I get what you're saying, I think "other" will suffice.

src/language/records.md
MaryaBelanger: Very good to know!

src/language/records.md
```diff
@@ -0,0 +1,120 @@
...
+}
+```
+
+The only ways to accomplish multiple returns otherwise are
+either to create a class (which is verbose), or use
+another collection type like List or Map (which lose type safety).
+
+{{site.alert.note}}
+  Records' multiple-return and heterogenous-type characteristics enable
+  parallelization of Futures of different types, which you can read about in the
```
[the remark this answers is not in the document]
MaryaBelanger: I was wondering about this! I saw a lot of non-code-quoted upper-cased "Future"s in the existing section of the library tour, and then some non-code-quoted _lower_-cases, and code-quoted `Future`s in other places. So, any rhyme or reason is much appreciated 😅

src/language/records.md
munificent: For what it's worth, I always use "type annotation" to mean the syntax for how to write down a type. They aren't optional "annotations" in the Dart 1 sense, but as far as I know, "type annotation" is pretty standard across languages as a way to refer to type terms.

src/_guides/libraries/library-tour.md
MaryaBelanger: @lrhn Thanks for fixing this! Is the `success` part necessary? I realize it's adding extra clarification that the `try` block represents successfully waiting for all the futures, and that `ParallelWaitError` represents a failure, despite having results returned, but isn't that kind of implied by the nature of `try` anyway?

src/_guides/libraries/library-tour.md
MaryaBelanger: Similar to my last question, is the `Expect.fail()` line right below necessary? I know I added this part, but now I'd like to get anything that's not absolutely necessary out for brevity. 

I was hoping the `errorResult()` future would imply that the result is expected to fail, just for the sake of the example. And I'm hoping if they want to learn more they'll go to the API docs for more complete examples (assuming that [the test samples](https://dart-review.googlesource.com/c/sdk/+/288903/12/tests/lib/async/future_record_extension_test.dart) show up in the api docs somewhere, and if they don't, they really should!)

src/language/records.md
MaryaBelanger: I'm pretty sure you wrote that sentence 😅 definitely a great intro!

src/language/records.md
MaryaBelanger: > I'd probably insert an example of a record type annotation with only positional fields first and explain that it's a type annotation _and where you might use it._

@munificent Where _would_ you use one? 
Do they only come up for destructuring? 

I'm drawing a blank [from the spec](https://github.com/dart-lang/language/blob/master/accepted/future-releases/records/records-feature-specification.md#record-type-annotations). I was going to say, "to declare a record", but that doesn't sound right, since they're immutable it's not like you can declare a record then populate it later (can you?)

[Even for typing](https://github.com/dart-lang/language/blob/master/accepted/future-releases/records/records-feature-specification.md#record-type-annotations:~:text=(num%2C%20num)%20r%20%3D%20(3%2C%203.5)..%241.isEven%3B), it seems like the types are going to be inferred regardless of what you write in a type annotation, so that doesn't seem like a use case, just an example for the sake of the spec.

src/language/records.md
MaryaBelanger: Just realized, the exact opposite of what I wrote in the next line 😅 good thing you caught that!

MaryaBelanger: @munificent Actually, I'm still not sure I understand. In the first example, if I were to say `print(point.x)` or `print(color.b)`, it would work and print `1` and `3`, right?

And what if you were to write:
```
(int x, int y, int z) point = (a: 1, b: 2, c: 3);
```
That wouldn't work, right?

src/language/records.md
MaryaBelanger: Is it accurate to say a records shape is "the the set of its fields, the fields' types, and their names (if any)"? I use "shape" later on in the page so it seems like this would be a good place to define it first

src/language/records.md
MaryaBelanger: Nevermind, you can ignore all of this. I was somehow reading "the names _of named_ fields" as just "the names of fields" (which was very confusing!), but now I get it.

src/_guides/libraries/library-tour.md
```diff
@@ -1293,6 +1293,64 @@ await Future.wait([
...
 ```
 
+`Future.wait()` returns a future which will complete once all the provided
+futures have completed, either with their results, or with an error if any of
+the provided futures fail. 
+
+#### Handling errors for multiple futures
+
+You can also wait for parallel operations on an [iterable]({{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-async/FutureIterable/wait.html)
+or [record]({{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-async/FutureRecord2/wait.html)
```
kevmoo, line 1303:
```suggestion
of [record]({{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-async/FutureRecord2/wait.html)
```

Right?

MaryaBelanger: Records aren't actually part of the iterable family. So these future APIs for iterables and records are separate. 

Kind of unrelated, but the implementation distinction between records' implementation / API and the rest of the collections API has made it hard to generalize all them with a single word (e.g. "you can aggregate different values in a single object using Dart's collection types (lists, maps, sets) ... _and_ records"). I wish I could just call them a collection.

src/_guides/libraries/library-tour.md
lrhn: This is example code, possibly the first/only example code someone sees of `await (someRecord).wait`.

Unlike awaiting a list, the only reason to await a record of futures is if you *need* the result values. Otherwise you could just have made a `List<Future<void>>` instead of a record.

So I think the first example someone sees of `await (someRecord).wait` should show someone who *wants* to use the result values.

But that comes back to the purpose of the example:
* Is it to show an example of `await (someRecord).wait`, or
* Is it to show how `await (someRecord).wait` differs from ... something? `Future.wait`?

The latter seems too context dependent. If someone reads this example first, they won't have the context needed. They will read it as an example of using `await (someRecord).wait`, so it should be a *good* example of that first, and then something which can also be compared to other features second.

IMO. :)

src/_guides/libraries/library-tour.md
lrhn: The word I'd use is that they're *composite* objects (can contain other objects).
More precisely "composite value containers" (even if it's a bit redundant).

Most objects are composite in some sense, but some objects has that as their primary objective, being exactly the sum of their parts with no extra semantic interpretation or functionality. A list is just a list. A task queue might be a list, but it has a role. It's not _just_ a data container.

"Collections" are traditionally *mutable* homogeneous composite objects, whose primary task is to collect and access their values. Very traditionally they are lists, sets, queues and/or maps (maps are sometimes considered a collection, other times not, qv., `java.util.Collection` and `java.util.Map`). Good old "data structures".

"Structs", "records" or "tuples" are heterogeneous composite objects, but still mostly without extra functionality on top of just carrying the data values. Aka. "data objects"/"data classes".

Don't know if that helps at all :grin:

MaryaBelanger: Composite! That's exactly what I was looking for. I've weaseled my way around lumping records in with the rest of the collections so far, so I don't think I need to go back and correct anything, but if the need arises again (and I'm sure it will) I'll use "composite". Might even be a good entry for the new glossary. Thanks!

src/_guides/libraries/library-tour.md
MaryaBelanger: Ok, I think I understand now. I'll change them to `Future<void>` so the example is at least realistic (though still simpler, imo). 

If you think the use case shown in the example you wrote is much more valuable, though, let me know and I'll use it. I just worry that readers might conflate some of the other stuff that's going on in that example (like using `contains()` or somewhat "unpacking" the results into a boolean variable) with the question of "how to use `wait` on an iterable".

src/_guides/libraries/library-tour.md
MaryaBelanger: That's a _really_ good point. I was thinking that it'd be most useful to users, for some reason, to just see the bare minimum of how to set up using `wait` on a record. I didn't think of showing an actual useful use case. Thanks so much for explaining!

src/language/records.md
munificent: Yes, you rarely need them for variable declarations because they can be inferred. But they come up at API boundaries: Parameter types and return types. For example, here's a function that takes a record and swaps its fields:

```dart
(int, int) swap((int, int) record) {
  var (a, b) = record;
  return (b, a);
}
```

The two `(int, int)` here are record type annotations.

src/_guides/libraries/library-tour.md
lrhn: I think switching to `Future<void>` is fine. 
I expect "just tell me when they're done" to be the most common use-case for list-wait.

Having an _unknown_ number of same-typed futures, that you want to run in parallel, and still want all the values of, seems more like an edge case for a badly-written API than a common use-case.
(If you end up doing `var results = await ids.map((id) => server.fetch(id)).wait;`, I'd ask why the server API doesn't provide a `server.fetchMultiple(Iterable<Id>)` function to begin with.)

It'll still happen. A lot. But I have no doubt people doing that will already know `Future.wait`.

src/language/records.md
```diff
@@ -0,0 +1,142 @@
...
+// Error, different types:
+point = color;
+```
+
+Records automatically define `hashCode` and `==` methods structurally based on
+their fields.
+
+## Multiple returns
+
+Records allow functions to return multiple values bundled together. Using [pattern matching][], you can retrieve record values from a return by destructuring them directly into local variables:
```
johnpryan, line 105: Split into multiple lines?

src/language/records.md
```diff
@@ -0,0 +1,142 @@
...
+  /* Equivalent to:
+  var info = userInfo(json);
+  var name = info.$1;
+  var age  = info.$2;
+  */
+}
+```
+
+Other ways to accomplish multiple returns are
+either to create a class (which is verbose), or use
```
johnpryan, line 130:
```suggestion
either to create a class (which is more verbose), or use
```

src/_guides/libraries/library-tour.md
MaryaBelanger: > the only reason to await a record of futures is if you need the result values.

I added this description on top of the example, and showed some arbitrary "use" of the result values

src/language/records.md
```diff
@@ -0,0 +1,142 @@
...
+```
+
+Named fields inside _record type annotations_ go inside a brace-delimited section
+of type and name pairs, after all positional fields:
+
+```dart
+(String, String?, {int a, bool b}) record;    // Record type annotation
+```
+Naming fields in record type annotations is optional for documentation purposes,
+but the names of named fields are not optional. For more information,
```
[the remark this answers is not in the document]
MaryaBelanger: I think it means, once given a name in annotation, the name can be omitted but it can't be replaced. Changed this to

```
Naming fields in record type annotations is optional for documentation purposes.
The names of those named fields are not optional, though,  
once you create a record with named fields. The field names become part of the
[record's type definition](#record-types). They can be omitted, but not replaced.

For more information, see the [Record equality](#record-equality) section.
```

src/_guides/libraries/library-tour.md
```diff
@@ -1757,6 +1828,7 @@ To learn more about the Dart language, see the
 [NativeFinalizer]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-ffi/NativeFinalizer-class.html
 [NoSuchMethodError]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-core/NoSuchMethodError-class.html
 [Object]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-core/Object-class.html
+[`ParallelWaitError`]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-async/ParallelWaitError-class.html
```
Hixie, line 1831: why does this one use backticks when the others don't?

src/language/records.md
```diff
@@ -0,0 +1,160 @@
...
+If two unrelated libraries create records with the same set of fields,
+the type system understands that those records are the same type even though the
+libraries are not coupled to each other.
+
+## Record equality
+
+Two records are equal if they have the same _shape_ (set of fields),
+and their corresponding fields have the same values.
+Since named field _order_ is not part of a record's shape, the order of named
+fields does not affect equality:
```
Hixie, line 98: this sentence ends with a colon but the example doesn't demonstrate the sentence's claim

src/_guides/libraries/library-tour.md
```diff
@@ -1293,6 +1293,77 @@ await Future.wait([
...
+#### Handling errors for multiple futures
+
+You can also wait for parallel operations on an [iterable]({{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-async/FutureIterable/wait.html)
+or [record]({{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-async/FutureRecord2/wait.html)
+of futures.
+
+These extensions return a future with the resulting values of all the provided
+futures. Unlike `Future.wait`, if any future in the collection completes with
+an error, `wait` completes with a [`ParallelWaitError`][] that allows the caller
+to handle individual errors and dispose successful results if necessary.
```
atsansone, line 1309: **issue:** Write as shorter sentences per [GSG](https://developers.google.com/style/accessibility#ease-of-reading).
```suggestion
These extensions return a future with the values of all provided futures.
This differs from `Future.wait`.
If any future in the collection returns an error, `wait` returns a [`ParallelWaitError`][].
The caller can then handle each error and drop successful results if needed.
```

src/language/records.md
```diff
@@ -0,0 +1,160 @@
...
+
+```dart
+// Record type annotation in a variable declaration
+(String, String?, {int a, bool b}) record;
+```
+
+Naming fields in record type annotations is optional for documentation purposes.
+The names of those named fields are not optional, though,  
+once you create a record with named fields. The field names become part of the
+[record's type definition](#record-types). They can be omitted, but not replaced.
```
atsansone, line 45: **issue:** Sentence structure per [GSG](https://developers.google.com/style/sentence-structure)
```suggestion
Once you create a record with named fields,
those field names become part of the [record's type definition](#record-types).
They can be omitted, but not replaced.
```

**question:**  If they're not optional, how can you omit the field names?

src/language/records.md
```diff
@@ -0,0 +1,160 @@
...
+
+{% comment %}
+    TODO: link to patterns page, specifically records destructuring section.
+{% endcomment %}
+
+## Record types
+
+There is no type declaration for individual record types. Records are structurally
+typed based on the types of their fields. A record's _shape_ (the set of its fields,
+the fields' types, and their names, if any) uniquely determines the type of a record.
```
atsansone, line 76: **suggestion:** Use active voice per [GSG](https://developers.google.com/style/voice). Though the object doesn't need to own the action, adding agency helps here. 

```suggestion
Records don't declare the types for individual record types.
The record's _shape_ determines the specific type of a record. 
A shape includes the set of its fields, the fields' types, and any field names.
```

src/language/records.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Record types
+
+There is no type declaration for individual record types. Records are structurally
+typed based on the types of their fields. A record's _shape_ (the set of its fields,
+the fields' types, and their names, if any) uniquely determines the type of a record. 
+
+Each field in a record has its own type, which can be different from other
+fields' types in the same record. The type system is aware of each
+field's type wherever it is accessed from the record:
```
atsansone, line 80: **suggestion:** Active voice per [GSG](https://developers.google.com/style/voice). 
```suggestion
Each field in a record has its own type. Fields can differ from other fields' types in the same record.
The type system is aware of each field's type wherever the record accesses the type.
```

src/language/records.md
```diff
@@ -0,0 +1,160 @@
...
+// Record type annotation in a variable declaration
+(String, String?, {int a, bool b}) record;
+```
+
+Naming fields in record type annotations is optional for documentation purposes.
+The names of those named fields are not optional, though,  
+once you create a record with named fields. The field names become part of the
+[record's type definition](#record-types). They can be omitted, but not replaced.
+
+For more information, see the [Record equality](#record-equality) section.
```
atsansone, line 47: **nit:** Prefer different construction because there is no call to action for the reader.
```suggestion
To learn more, check out the [Record equality](#record-equality) section.
```

src/language/records.md
```diff
@@ -0,0 +1,160 @@
...
+```dart
+(num, Object) pair = (42, "a");
+
+var first = pair.$1;     // static type `num`, runtime type `int`
+var second = pair.$2;    // static type `Object`, runtime type `String`
+```
+
+If two unrelated libraries create records with the same set of fields,
+the type system understands that those records are the same type even though the
+libraries are not coupled to each other.
```
atsansone, line 91: **issue:** Write as shorter sentences per [GSG](https://developers.google.com/style/accessibility#ease-of-reading).
```suggestion
You can create records with the same field set in unrelated libraries.
Despite no link between the libraries, the type system infers that those records are the same type.
```

src/language/records.md
```diff
@@ -0,0 +1,160 @@
...
+```dart
+({int x, int y, int z}) point = (x: 1, y: 2, z: 3);
+({int r, int g, int b}) color = (r: 1, g: 2, b: 3);
+
+// Error, different types:
+point = color;
+```
+
+Records automatically define `hashCode` and `==` methods structurally based on
+their fields.
```
atsansone, line 117: **nits:** "Automatically" is redundant. The user can't manually define the hashCode, correct? "Structurally" reads as a hedge. 
```suggestion
Records define their `hashCode` and `==` methods based on the structure of their fields.
```

src/language/records.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Multiple returns
+
+Records allow functions to return multiple values bundled together.
+Using [pattern matching][], you can retrieve record values from a return by
+destructuring them directly into local variables:
```
atsansone, line 123: **issue:** Condition, then action per [GSG](https://developers.google.com/style/sentence-structure).
```suggestion
To retrieve record values from a return,
destructure the values into local variables using [pattern matching][].
```

src/language/records.md
```diff
@@ -0,0 +1,160 @@
...
+  var info = userInfo(json);
+  var name = info.$1;
+  var age  = info.$2;
+  */
+}
+```
+
+Other ways to accomplish multiple returns are
+either to create a class (which is more verbose), or use
+another collection type like `List` or `Map` (which loses type safety).
```
atsansone, line 149: **issue:** Avoid parentheses per [GSG](https://developers.google.com/style/parentheses). Try to ease reading per [GSG](https://developers.google.com/style/accessibility#ease-of-reading).
```suggestion
You can accomplish multiple returns in other ways that have some downsides.
* Create a class. This is more verbose.
* Use another collection type like `List` or `Map`. This loses type safety.
```

atsansone (commented):
@MaryaBelanger : Huge improvements! Most of the issues relate to sentence length. This includes some suggestions and nits. You can choose to just acknowledge but not change those.

src/language/records.md
```diff
@@ -0,0 +1,160 @@
...
+Record fields are accessible through built-in getters. Records are immutable,
+so fields do not have setters. 
+
+Named fields expose getters of the same name. Positional fields expose getters
+of the name `$<position>`, skipping named fields:
+
+```dart
+var record = ('first', a: 2, b: true, 'last');
+
+print(record.$1);    // Prints '123'
```
shilangyu, line 60: Shouldnt this print `'first'`?

src/language/records.md
shilangyu, line 45: Wording seems confusing, maybe more clearly differentiate what is the difference between a "field with a name" and "named field"? Maybe mention that the first one concerned "positional fields"? Bob's analogy to function parameters seems good.

src/language/records.md
lrhn: Yep, looks like a typo. There is no "123" to print.

src/language/records.md
MaryaBelanger: Changed 43-45 to:

> Once you create a record using a record expression, though,
> any field names become part of the [record's type definition](#record-types)
> and are immutable. They can be omitted, but not replaced.

And then the link to the record equality section shows the example:

> ```dart
> // point =/= color
> ({int x, int y, int z}) point = (x: 1, y: 2, z: 3);
> ({int r, int g, int b}) color = (r: 1, g: 2, b: 3);
> ```


 is that better? Admittedly, I've been having trouble capturing this point, so any feedback helps

src/_guides/libraries/library-tour.md
MaryaBelanger: I'd argue that the class names that _aren't_ back ticked need to change?

src/_guides/libraries/library-tour.md
MaryaBelanger: I went through your reviews bottom up on this section, but now that I'm up here, I realized that I could kind of rearrange the whole section to be more concise. I noticed I repeated what the new `wait` does 3 times (how it differs from `Future.wait`, then how it works for iterables, then how it works for records, which are all the same description). So I changed everything to now only mention how it works _once_ at the top, and then focus only on the differences:

> These extensions return a future with the resulting values of all provided
> futures. Unlike `Future.wait`, they also let you handle errors. 
> 
> If any future in the collection completes with an error, `wait` completes with a
> [`ParallelWaitError`][]. This allows the caller to handle individual errors and
> dispose successful results if necessary.
(NOTE: I think "completes" is different than "returns", so had to leave that in)
> 
> Use `wait` on an _iterable_ of futures when you _don't_ need the result values
> from each individual future:
> 
> ```dart
> // examples the same
> ```
> 
> Use `wait` on a _record_ of futures when you _do_ need the individual result values
> from each future. This also provides the additional benefit that the futures can be of different types:
>
> ```dart
> // examples the same
> ```
> (end section)

src/_guides/libraries/library-tour.md
```diff
@@ -1293,6 +1293,77 @@ await Future.wait([
 print('Done with all the long steps!');
 ```
 
+`Future.wait()` returns a future which will complete once all the provided
+futures have completed, either with their results, or with an error if any of
+the provided futures fail.
```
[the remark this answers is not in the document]
MaryaBelanger: Worked around to shorten this, because a future completing is different than returning (I don't understand it much more than that...):

> `Future.wait()` returns a future which completes once all the provided
futures have completed. It completes either with their results,
or with an error if any of the provided futures fail.

src/language/records.md
MaryaBelanger: It's confusing. How I understand it, which I haven't found a good way to put into words yet, is: once you name fields and actually create a record using those names (so, not just using a record type annotation to define its shape, but using a record expression to populate the fields), the names become part of the record's definition, i.e. it's shape. 

You can create a record from an annotation with named fields _without_ using the names, because they're implicit(maybe? not super sure about this part), but you _can't_ create a record from an annotation with named fields using different names in the expression...

```dart 
   ({int x, int y}) someRecord = (x: 1, y: 2);
// ^^^^^^^^^^^^^^^              ^^^^^^^^^^^^
// type annotation              expression

   ({int x, int y}) someRecord;
   someRecord = (1, 2);
// fine, field values fit the types in the annotation, and don't clash with the names
   print(someRecord.x);    // prints 1, i think?
   print(someRecord.y);    // prints 2, i think?

   ({int x, int y}) someRecord;
   someRecord = (a: 1, b: 2);
// doesn't work, can't give the fields different names because someRecord's shape is defined as named int x and named int y
```

So, something like that... but I'm not 100% sure that's accurate

src/language/records.md
MaryaBelanger: Skipping second sentence rewrite, the record doesn't access itself, it could be any construct doing the accessing hence generic "wherever"

src/language/records.md
MaryaBelanger: Users can manually define hasCode and ==, so "automatically" actually is the selling point here. 

"based on the structure of their fields" sounds good

src/language/records.md
MaryaBelanger: I don't want to use the bullet list because I feel like it draws too much attention to something that's just kind of a throwaway comment and not about records. Maybe see if the edit in the new commit is reasonable?

src/language/records.md
```diff
@@ -0,0 +1,178 @@
...
+// Type annotation:
+({int x, int y}) someRecord;
+// Record expression creating record from type annotation, omitting names:
+someRecord = (1, 2);
+// Prints 1, 2:
+print(someRecord.x, someRecord.y);
+```
+
+For more information and examples, check out [Record types](#record-types) and
+[Record equality](#record-equality).
```
MaryaBelanger, line 67: @munificent this was the only significant change from the last commit (lines 34-67), could you review it one more time? There were a couple comments on the confusing wording so I tried to elaborate

src/language/records.md
munificent: I'll see if I can help:

You can think of every record as having two separate sets of fields:

* A set of positional fields.
* A set of named fields.

These two sets are totally independent of each other: a named field isn't a positional field and a positional field isn't a named field. Either set may be empty, or they can both be. You can have records with just positional fields, just named, both, or no fields at all.

### Record expressions

When you write a record expression to create a record object, the positional fields are written as simple expressions:

```dart
var record = (1, 2, 3);
//          ^  ^  ^ Positional fields.
```

The named fields have a name and `:` before the field value:

```dart
var record = (x: 1, y: 2, z: 3);
//          ^^^^  ^^^^  ^^^^ Named fields.
```

You can have both in a single record expression and they can be arbitrarily
interleaved:

```dart
var record = (x: 1, 2, y: 3, 4);
//                ^        ^  Positional fields.
//          ^^^^     ^^^^  Named fields.
```

### Record types

Every record object has a static type, which is how the type checker "sees" the record. Users write down these types using "type annotations" (which are not related to metadata annotations, the `@foo` syntax).

In a record type annotation, the positional fields are written as simple type annotations for the field types:

```dart
  (int, String) record = (1, 'str');
// ^^^  ^^^^^^ Positional fields.
```

Named fields are enclosed in curly braces and must have a name after the type:

```dart
  ({int a, String b}) record = (a: 1, b: 'str');
//  ^^^^^  ^^^^^^^^ Named fields.
```

If a record type annotation has both positional fields and named ones, the positional ones must come first:

```dart
  (double, bool, {int a, String b}) record = (1.2, true, a: 1, b: 'str');
// ^^^^^^^ ^^^^ Positional fields.
//                ^^^^^  ^^^^^^^^ Named fields.
```

### Positional field names

The above is, I hope, fairly clear and straightforward. Where it gets a little confusing is that we do allow names on positional fields *in record type annotations*:

```dart
  (int x, String y) record = (1, 'str');
// ^^^^^  ^^^^^^^^ Positional fields.
```

Note that the fields aren't enclosed in curly braces. These are still positional fields and are not named fields. Note how in the record expression in the initializer, there are no names. It's just that the fields in the type annotation have, uh, "names".

But these "names" are purely for documentation and have no meaning to the language or the type checker. It's essentially the same as doing:

```dart
  (int /* x */, String /* y */) record = (1, 'str');
// ^^^          ^^^^^^ Positional fields.
```

And, in particular, these names are *not* part of the record's actual type:

```dart
(int x, String y) record1 = (1, 'str');
(int a, String b) record2 = record1; // OK!
```

These two variables, `record1` and `record2` have the exact same type even though the positional field "names" are different, which is why the assignment is valid.

This is different from actual named fields where the name *is* part of the type:

```dart
({int x, String y}) record1 = (x: 1, y: 'str');
({int a, String b}) record2 = record1; // Error!
```

This assignment is invalid because `record1` and `record2` have different types. Note also that the record expression in the initializer now says the field names.

The proposal sometimes talks about the "name" of a positional field, but really we should have a different term for it because it's not the same as an actual named field's name.

Sorry this is confusing. Much of this comes from deliberately designing records to mirror function parameter lists, so we inherit some weirdness from that. But I think the overall correspondence is useful (especially if we ever add the ability to spread a record into an argument list).

You can see the exact same confusion around positional parameter names in function typedefs:

```dart
typedef Positional1 = Function(int x, String y);
typedef Positional2 = Function(int a, String b);
```

These two typedefs define equivalent types even though the positional parameters have different names. But these two are different types from each other:

```dart
typedef Named1 = Function({int x, String y});
typedef Named2 = Function({int a, String b});
```

src/language/records.md
```diff
@@ -14,15 +13,16 @@ functions, and store them in lists.
...
+// Record expression:
+var record = ('first', a: 2, b: true, 'last');
 ```
 
-_Record type annotations_ are comma-delimited lists of types enclosed in parenthesis.
-The `(int, int)` statements here are record type annotations used as a return
-type and as parameter types:
+_Record type annotations_ are comma-delimited lists of types enclosed in parentheses.
+The `(int, int)` statements in the following example are record type annotations
+used as a return type and as parameter types:
```
atsansone, line 25: **suggestion:** Rephrase and split to ease readability.
```suggestion
The following example includes the `(int, int)` statements
as record type annotations.
The statements work as both a return type and as parameter types:
```

src/language/records.md
```diff
@@ -31,20 +31,40 @@ type and as parameter types:
...
 (String, String?, {int a, bool b}) record;
 ```
 
 Naming fields in record type annotations is optional for documentation purposes.
-The names of those named fields are not optional, though,  
-once you create a record with named fields. The field names become part of the
-[record's type definition](#record-types). They can be omitted, but not replaced.
+Once you create a record from a type annotation with named fields using a record
+expression, the names become part of the [record's type definition](#record-types),
+or its _shape_:
```
atsansone, line 46: **nit:** The comma isn't needed.
```suggestion
expression, the names become part of the [record's type definition](#record-types) or its _shape_:
```

src/language/records.md
```diff
@@ -31,20 +31,40 @@ type and as parameter types:
...
+work in functions. They go inside a curly brace-delimited
+section of type-and-name pairs, after all positional fields:
 
 ```dart
-// Record type annotation in a variable declaration
+// Record type annotation in a variable declaration:
 (String, String?, {int a, bool b}) record;
 ```
 
 Naming fields in record type annotations is optional for documentation purposes.
```
atsansone, line 43: **nit:** Write in active voice per [GSG](https://developers.google.com/style/voice).
```suggestion
Record type annotations don't require naming fields for documentation purposes.
```

src/_guides/libraries/library-tour.md
```diff
@@ -1293,33 +1293,35 @@ await Future.wait([
...
+futures. Unlike `Future.wait`, they also let you handle errors. 
 
-A common use case is wanting to know when all the futures complete, and handle
-any errors, but not necessarily caring about the result of each individual future:
+If any future in the collection completes with an error, `wait` completes with a
+[`ParallelWaitError`][]. This allows the caller to handle individual errors and
+dispose successful results if necessary.
+
+Use `wait` on an _iterable_ of futures when you _don't_ need the result values
+from each individual future:
```
atsansone, line 1314: **suggestion:** Condition, then instruction per [GSG](https://developers.google.com/style/sentence-structure).
```suggestion
When you _don't_ need the result values from each individual future,
use `wait` on an _iterable_ of futures.
```

src/_guides/libraries/library-tour.md
```diff
@@ -1335,8 +1337,8 @@ void main() async {
 }
 ```
 
-Waiting on a _record_ of futures enables the same functionality, with the additional 
-benefit that the futures can be of different types: 
+Use `wait` on a _record_ of futures when you _do_ need the individual result values
+from each future. This also provides the additional benefit that the futures can be of different types:
```
atsansone, line 1341: **suggestion:** Condition, then instruction per [GSG](https://developers.google.com/style/sentence-structure).
```suggestion
When you _need_ the individual result values from each future,
use `wait` on a _record_ of futures.
This also allows the futures can be of different types.
```

atsansone (approved):
LGTM % suggestions and nits. No further review needed.

src/language/records.md
shilangyu: I'll mark this as resolved since you adopted munificent's suggestion

src/language/records.md
munificent: It's getting there, but still not quite accurate. How about:

---

Fields in record expressions and type annotations mirror how [parameters and arguments][] work in functions. Positional fields go directly inside the parentheses:

```dart
// Record type annotation in a variable declaration:
(String, int) record;

// Initialize it with a record expression:
record = ('A string', 123);
```

In a record type annotation, named fields go inside a curly brace-delimited
section of type-and-name pairs, after all positional fields. In a record
expression, the names go before each field value with a colon after:

```dart
// Record type annotation in a variable declaration:
({int a, bool b}) record;

// Initialize it with a record expression:
record = (a: 123, b: true);
```

The names of named fields in a record type are part of thee [record's type
definition](#record-types), or its _shape_. Two records with named fields with
different names have different types:

```dart
({int a, int b}) recordAB = (a: 1, b: 2);
({int x, int y}) recordXY = (x: 3, y: 4);

recordAB = recordXY; // Compile error! These records don't have the same type.
```

In a record type annotation, you can also name the *positional* fields, but
these names are purely for documentation and don't affect the record's type:

```dart
(int a, int b) recordAB = (1, 2);
(int x, int y) recordXY = (3, 4);

recordAB = recordXY; // OK.
```

This is similar to how positional parameters in a function declaration or function typedef can have names but those names don't affect the signature of the function.

For more information and examples, check out [Record types](#record-types) and
[Record equality](#record-equality).

MaryaBelanger: Thanks, I'll open another PR for this!

PR #4770, opened 2023-04-17
Make prerelease section more current

timsneath:
With Dart 3 close to release and lots of package authors facing the scenario described for Dart 3, it makes sense to describe the prerelease versioning in that context, rather than based on an older version. This change will hopefully reduce confusion.


parlough (approved):
This makes sense to me. Thanks :)

mit-mit (approved):
This is great, thanks!

PR #4774, opened 2023-04-21
Remove no longer used properties from sidenav data file

parlough:
These were used in an old `sidenav.yml` include but we haven't used that since https://github.com/dart-lang/site-www/pull/1572


MaryaBelanger (approved):
👍

PR #4780, opened 2023-04-24
[Dart 3] Patterns-related control flow

### LINKED ISSUE #4635: [Patterns] Control flow docs

MaryaBelanger:
What I did:
- Split up existing control flow page to make room for longer switch content
- New pages, staged: 
  - [Assert](https://dart-3-marya.web.app/language/assert) (unchanged)
  - [Loops](https://dart-3-marya.web.app/language/loops) (unchanged)
  - [Branches](https://dart-3-marya.web.app/language/branches) (changed)
- Added all-new sections to branches.md, staged direct links to each section:
  - [if-case](https://dart-3-marya.web.app/language/branches#if-case) 
  - [switch expression](https://dart-3-marya.web.app/language/branches#switch-expressions)
  - [guard clause](https://dart-3-marya.web.app/language/branches#when)
  - [exhaustiveness checking](https://dart-3-marya.web.app/language/branches#exhaustiveness-checking)
- Modified existing switch statements content on branches.md (removed statements that are no longer accurate/recommended)
- Went through the site to fix links to old control flow page

---

Risks / Questions:

- I think the big "types of case patterns" section I added right under the "Switch case" heading should probably actually live on the Patterns page when I create that. Opinions welcome
- I tried removing `break` from existing switch statement code excerpts, wouldn't allow it

---

TODO:
- Actual code excerpts
- Add patterns page links and class modifiers links when all merged on same branch.

---

Fixes #4635


MaryaBelanger:
@parlough any idea what to do about these build test failures?

parlough:
> @parlough any idea what to do about these build test failures?

I've prepared a temporary fix in https://github.com/dart-lang/site-www/pull/4781.

src/language/branches.md
```diff
@@ -0,0 +1,295 @@
...
+}
+```
+
+The statement conditions must be expressions
+that evaluate to [boolean values][boolean], nothing else.
+To learn how to use `if` in an expression context, see [Conditional expressions][]
+
+## If case
+
+Dart `if` statements support `case` clauses followed by a [case pattern][pattern]:
```
eernstg, line 42: I don't think there's a need to introduce a new concept of 'case patterns'. The patterns occurring in this syntactic context are just patterns.

Of course, we do have the notions of 'declaration context', 'matching context', 'assignment context', and you could argue that we have different kinds of patterns in those contexts. (For instance, an identifier pattern has a very different meaning in each of those contexts.) But it's quite reasonable to say that we just have patterns, and they can occur in each of those different contexts. With that, we don't need to introduce different words to denote patterns when they occur in different contexts. I think that's going to be more comprehensible.

src/language/branches.md
```diff
@@ -0,0 +1,295 @@
...
+If-case is a simple way to match and destructure against a
+_single_ pattern. To test a value against _multiple_ patterns, use [switch](#switch-case).
+
+## Switch case
+
+Switch cases (both [statements](#switch-statements) and [expressions](#switch-expressions))
+in Dart compare an object against one or more [case patterns][pattern]. 
+
+Case patterns allow control flow to
+either match and destructure the object being switched, or continue execution
```
eernstg, line 71: Perhaps?:
```suggestion
either match and destructure the object being switched on, or continue execution
```
'Switching on' an object is a rather common phrase. Some people would say that object is the 'scrutinee', but perhaps that's too specialized (it's _the_ thing to say in the functional programming language community).

src/language/branches.md
```diff
@@ -0,0 +1,295 @@
...
+if the object doesn't match. They can be any kind of [refutable pattern][]:
```
eernstg, line 72: They can be refutable as well as irrefutable patterns. So it isn't going to help anyone to suggest that they can't be irrefutable. However, saying that it's 'any kind of pattern' and then 'including a [refutable pattern][]' might be helpful.

src/language/branches.md
```diff
@@ -0,0 +1,295 @@
...
+- **Object patterns**: the match exposes getters on the object type.
+  ```dart
+  switch (Shape shape) {
+    case Rect(width: var w, height: var h): ...
+    // Matches if shape is of type Rect, and then against the properties of Rect
+  }
+  ```  
+
+Values destructured by a case pattern become local variables, only available in the
+scope of that clause.
```
eernstg, line 117: People might not have an idea about how that scope is determined. Perhaps:
```suggestion
Values destructured by a pattern in a case become local variables, only available in the
body of that case (or group of cases, when one body is shared by several cases). 
```

src/language/branches.md
```diff
@@ -0,0 +1,295 @@
...
+
+_Switch expressions_ write a result value determined by a switch
+directly into a variable or return statement. 
+
+The syntax rules for a switch expression are:
+- The expression body is a series of case patterns.
+- Cases _do not_ use the `case` keyword.
+- Case patterns are followed by `=>`, and then a case body. 
+- Case bodies terminate with a comma `,`.
+- The switch expression body terminates with a semicolon `;`.
```
eernstg, line 186: That's not a property of the switch expression or its body, it is a requirement in some situations of the enclosing construct: A statement consisting of an expression ends in a semicolon (e.g., `cond ? 1 : switch (x) {...};` needs the semicolon because it's a statement of the form `e;` where `e` is an expression). In other situations the `}` could be followed by a `,`, say, if the switch is an actual argument in a function invocation.

So, maybe:
```suggestion
- The switch expression body may be followed by a semicolon `;`, e.g., if it is part of a declaration or a statement.
```

src/language/branches.md
```diff
@@ -0,0 +1,295 @@
...
+- The wildcard pattern `_` catches remaining values instead of `default`.
+
+This allows you to rewrite a 
+_statement_ like this:
```
eernstg, line 190:
```suggestion
_statement_ like this (where `slash`, `star`, ... are constant variables):
```
I think it's somewhat hard to understand what it is doing if we don't know that they must be constant variables.

src/language/branches.md
```diff
@@ -0,0 +1,295 @@
...
+```
+
+Into an _expression_, like this:
+
+```dart
+var token = switch (charCode) {
+  slash when nextCharCode == slash => skipComment(),
+  slash || star || plus || minus   => operator(charCode),
+  >= digit0 && <= digit9           => number(),
+  _                                => throw invalid()
```
eernstg, line 212: This looks tempting, but Dart standard formatting _never_ makes an attempt to use alignment such that specific tokens (`=>`) end up looking like a vertical line.

Perhaps we shouldn't give anybody the impression that they should align their code like that?

src/language/branches.md
```diff
@@ -0,0 +1,295 @@
...
+```
+
+### Exhaustiveness checking
+
+Exhaustiveness checking is a feature of switches that raises a compile-time error
+if a switch is not exhaustively covering all possible values that could flow through it.
+
+The compiler can only know if a set of cases is
+exhaustive if it is soundly aware of all possible enumerations of that type. So,
+exhaustiveness checking works only when switching on types that are fully enumerable:
```
eernstg, line 223:
```suggestion
exhaustiveness checking on switch _statements_ works only when switching on types that are fully enumerable:
```

src/language/branches.md
```diff
@@ -0,0 +1,295 @@
...
+    Square(length: var l) => l * l,
+    Circle(radius: var r) => math.pi * r * r
+  };
+```
+
+If a new subclass of `Shape` is ever added, exhaustiveness 
+checking will warn you about the missing subtype, because your switch is incomplete.
+This allows you to use Dart in a somewhat 
+[functional algrbraic datatype style](https://en.wikipedia.org/wiki/Algebraic_data_type). 
+
```
eernstg, line 262: Then we'd need to say that switch expressions are always required to be exhaustive, and that this is possible even with types that aren't enumerable if there is a case that covers all possible values (such as a default case or a case like `var x` which will match every value of the scrutinee and give `x` the same declared type as the static type of the scrutinee expression).

src/language/loops.md
```diff
@@ -0,0 +1,137 @@
+---
```
eernstg, line 1: So there are no examples where a `for` loop uses a pattern to process the values obtained from the iterable?

```dart
for (var (x, y) in listOfPairs) {...}
```
Maybe that kind of example is in some other file?

eernstg (approved):
Added a handful of comments.

src/language/branches.md
```diff
@@ -0,0 +1,295 @@
+---
+title: Branches 
+description: Learn how to use branches to control the flow of your Dart code.
+---
+
+You can control the flow of your Dart code using branches:
+
+-   `if`, `case`, and `else`
+-   `switch` and `case` statements
+-   `switch` and `case` expressions
```
munificent, line 10: `else` is always contained and subordinate to an `if` statement/element. Likewise, `case` is always part of some surrounding `switch` statement. I wouldn't refer to them as statements or expressions on their own since they really aren't. Maybe:

```
You can control the flow of your Dart code using branching:

-  `if` statements and elements
-  `if-case` statements and elements
-   `switch` statements and expressions
```

And then just describe `else` and `case` as part of their respective contained things.

src/language/branches.md
```diff
@@ -0,0 +1,295 @@
...
+
+Other methods of manipulating control flow in Dart are:
+
+-   [`for` loops][]
+-   [`while` and `do while` loops][]
+-   [`break` and `continue`][break]
+-   [`assert`][]
```
munificent, line 17: I'd remove `assert()` from here.

src/language/branches.md
```diff
@@ -0,0 +1,295 @@
...
+
+You can also affect control flow using [Exceptions][], like `try-catch` and `throw`.
+
+## If else
+
+Dart supports `if` statements with optional `else` statements:
```
munificent, line 23: "`else` statements" -> "`else` clauses".

src/language/branches.md
```diff
@@ -0,0 +1,295 @@
...
+if (isRaining()) {
+  you.bringRainCoat();
+} else if (isSnowing()) {
+  you.wearJacket();
+} else {
+  car.putTopDown();
+}
+```
+
+The statement conditions must be expressions
```
munificent, line 36: Maybe "The condition in parentheses after `if` must be an expression that evaluates to a [boolean][]."

src/language/branches.md
```diff
@@ -0,0 +1,295 @@
...
+Other methods of manipulating control flow in Dart are:
+
+-   [`for` loops][]
+-   [`while` and `do while` loops][]
+-   [`break` and `continue`][break]
+-   [`assert`][]
+
+You can also affect control flow using [Exceptions][], like `try-catch` and `throw`.
+
+## If else
```
munificent, line 21: I'd either leave off `else` here or add it to `## If case` below since both of them can either have or omit the `else` clause. Or maybe do:

```
## If and If-else

...

## If-case and If-case-else
```

src/language/branches.md
```diff
@@ -0,0 +1,295 @@
...
+  print('Was coordinate array $x,$y');
+} else {
+  throw FormatException('Invalid coordinates.');
+}
+```
+
+If-case is a simple way to match and destructure against a
+_single_ pattern. To test a value against _multiple_ patterns, use [switch](#switch-case).
+
+## Switch case
```
munificent, line 65: Just "## Switch".

Cases are the things inside the switch, but I think most people refer to the entire statement as a "switch" and then use "switch case" to refer to a single case within a switch. So, given:

```dart
switch (n) {
  case 1: ...
  case 2: ...
  default: ...
}
```

I'd say, "This switch statement matches the value of `n`. The first switch case checks if it's 1 and the second switch case check if it's 2. The default case handles any other numbers."

src/language/branches.md
```diff
@@ -0,0 +1,295 @@
...
+
+Switch cases (both [statements](#switch-statements) and [expressions](#switch-expressions))
+in Dart compare an object against one or more [case patterns][pattern].
```
munificent, line 68: Should this link work? It doesn't seem to.

src/language/branches.md
munificent, line 72: "Continue execution if the object doesn't match" is a little unclear about *where* it continues to. I'd probably merge the "Switch case" and "Switch statement" sections together and say something like:

> A switch statement allows you to test a single value against a series of patterns and execute code for the first pattern that matches. A switch statement has evaluates a *matched value expression*. Then it has a series of *cases*. Each case has a pattern, and a block of code to execute if the pattern matches. After evaluating the matched value expression, each pattern is matched against the value. The first one that matches has its block of code executed. After that, execution jumps to the end of the switch statement.

Note also that switch case patterns can be irrefutable patterns too. There are some places in the language that disallow refutable patterns, but irrefutable patterns are always allowed anywhere a pattern can appear.

src/language/branches.md
munificent, line 117: Probably "clause" -> "case body".

src/language/branches.md
```diff
@@ -0,0 +1,295 @@
...
+    executeClosed();
+    continue nowClosed;
+  // Continues executing at the nowClosed label.
+
+  nowClosed:
+  case 'NOW_CLOSED':
+    // Runs for both CLOSED and NOW_CLOSED.
+    executeNowClosed();
+}
+```
```
munificent, line 174: Continue to label is so rarely used, I would be tempted to omit it from the guide completely. (Honestly, I was hoping to remove it from the language entirely in Dart 3.0, but, alas, I wasn't able to pull it off.)

src/language/branches.md
```diff
@@ -0,0 +1,295 @@
...
+
+### Switch expressions
+
+_Switch expressions_ write a result value determined by a switch
+directly into a variable or return statement.
```
munificent, line 179: You can use a switch expression anywhere an expression can appear so "directly into a variable or return statement" is probably too specific. For example, this is fine:

```dart
print(switch (n) { 1 => 'one', 2 => 'two', _ => 'other'});
```

How about:

> A switch expression is similar to a switch statement, but can be used where an expression is expected and produces a value based on the expression body of the case that matched.

src/language/branches.md
munificent, line 186: You may see this in some examples, but that's not part of the switch expression, it's a result of the context where the expression appears. In:

```dart
var x = switch (n) { 1 => 'one', _ => 'other'};
```

The trailing `;` is part of the variable declaration statement, not the switch, the same as in:

```dart
var x = 3;
```

There's no semicolon at the end in, for example:

```dart
print(switch (n) { 1 => 'one', _ => 'other' })
```

src/language/branches.md
```diff
@@ -0,0 +1,295 @@
...
+_Switch expressions_ write a result value determined by a switch
+directly into a variable or return statement. 
+
+The syntax rules for a switch expression are:
+- The expression body is a series of case patterns.
+- Cases _do not_ use the `case` keyword.
+- Case patterns are followed by `=>`, and then a case body. 
+- Case bodies terminate with a comma `,`.
+- The switch expression body terminates with a semicolon `;`.
+- The wildcard pattern `_` catches remaining values instead of `default`.
```
munificent, line 187: You can actually use `_` as a default case in switch statements too. (Whether that will become idiomatic or not is anybody's guess.)

Maybe something like:

> Switch expression syntax is slightly different from switch statements:
>
> - Cases *don't* start with the `case` keyword.
> - The case pattern is separated from its body using `=>` instead of `:`.
> - The case body is a single expression instead of a series of statements. Every case must have a body -- there is no implicit fallthrough for empty cases.
> - Cases are separated by `,` (and an optional trailing `,` is allowed).

src/language/branches.md
```diff
@@ -0,0 +1,295 @@
...
+Into an _expression_, like this:
+
+```dart
+var token = switch (charCode) {
+  slash when nextCharCode == slash => skipComment(),
+  slash || star || plus || minus   => operator(charCode),
+  >= digit0 && <= digit9           => number(),
+  _                                => throw invalid()
+};
+```
```
munificent, line 214: These examples aren't quite 1-1. How about:

---

This allows you to rewrite a 
_statement_ like this:

```dart
switch (charCode) {
  case slash || star || plus || minus: // Logical-or pattern
    token = operator(charCode);
  case comma || semicolon: // Logical-or pattern
    token = punctuation(charCode);
  case >= digit0 && <= digit9: // Relational and logical-and patterns
    token = number();
  default:
    throw invalid();
}
```

Into an _expression_, like this:

```dart
token = switch (charCode) {
  slash || star || plus || minus => operator(charCode),
  comma || semicolon => punctuation(charCode),
  >= digit0 && <= digit9 => number(),
  _ => throw invalid()
};
```

src/language/branches.md
```diff
@@ -0,0 +1,295 @@
...
+
+### Exhaustiveness checking
+
+Exhaustiveness checking is a feature of switches that raises a compile-time error
```
munificent, line 218: I try to only use "raise" to refer to throwing runtime exceptions. (For comparison, the keyword to throw an exception in Python and Ruby is `raise`.) Maybe "that reports a compile-time error"?

src/language/branches.md
```diff
@@ -0,0 +1,295 @@
...
+if a switch is not exhaustively covering all possible values that could flow through it.
```
munificent, line 219: Defining exhaustiveness in terms of "exhaustively" may not be very clear. :)

Maybe "reports a compile-time error if it's possible for a value entering the switch to not match any of the case patterns".

src/language/branches.md
```diff
@@ -0,0 +1,295 @@
...
+
+The compiler can only know if a set of cases is
+exhaustive if it is soundly aware of all possible enumerations of that type. So,
+exhaustiveness checking works only when switching on types that are fully enumerable:
+
+- [Boolean][boolean] types, because there are only two possible values, `true` and `false`.
+- [Enum][enum] types, because they are always a fixed number of constant values.
+- [Sealed][sealed] types, because they are classes with a closed hierarcy of subtypes in the same library.
```
munificent, line 227: "hierarchy".

src/language/branches.md
munificent, line 223: Exhaustiveness does actually work for all kinds of types. (Otherwise switch expressions would be much less useful because they *have* to be exhaustive. Otherwise, you might not know what value the expression evaluates to.)

The reason that enums and sealed types are particularly useful for switches is that they allow you to "partition" the matched value in a way such that multiple cases each of which handle only a subset of the possible values can provably cover all possible values.

For other types, to be exhaustive, you basically need to have a single case that covers everything. So in:

```dart
int n = ...
var name = switch (n) {
  1 => 'one',
  2 => 'two',
  _ => 'other'
};
```

The switch is exhaustive because the final `_` pattern covers all possible ints.

I'd probably just remove this section. I don't think we go into a lot of detail about how type inference and flow analysis work and those are similar where there's a decent amount of algorithm sophistication going on under the hood and we just expect that it aligns with user intuition.

So if they have an intuition about the set of all possible values going into a switch and a good intuition about which values the set of patterns covers then exhaustiveness checking should mostly align with that. The actual mechanics of it get pretty complex.

src/language/branches.md
```diff
@@ -0,0 +1,295 @@
...
+
+double calculateArea(Shape shape) =>
+  switch (shape) {
+    Square(length: var l) => l * l,
+    Circle(radius: var r) => math.pi * r * r
+  };
+```
+
+If a new subclass of `Shape` is ever added, exhaustiveness 
+checking will warn you about the missing subtype, because your switch is incomplete.
```
munificent, line 259: "warn" -> "tell". (Since it's an error, not a warning.)

src/language/branches.md
```diff
@@ -0,0 +1,295 @@
...
+This allows you to use Dart in a somewhat 
+[functional algrbraic datatype style](https://en.wikipedia.org/wiki/Algebraic_data_type).
```
munificent, line 261: "algebraic".

src/language/branches.md
```diff
@@ -0,0 +1,295 @@
...
+    Circle(radius: var r) => math.pi * r * r
+  };
+```
+
+If a new subclass of `Shape` is ever added, exhaustiveness 
+checking will warn you about the missing subtype, because your switch is incomplete.
+This allows you to use Dart in a somewhat 
+[functional algrbraic datatype style](https://en.wikipedia.org/wiki/Algebraic_data_type). 
+
+## When
```
munificent, line 263: I'd name the heading "Guard clause" since that's what it is.

src/language/branches.md
MaryaBelanger: Great point, I'll incorporate this feedback throughout the page

src/language/branches.md
MaryaBelanger: Agreed it's a little too specialized here, but I'll definitely be using "scrutinee" in the functional-style-dart blog post I'm writing!

src/language/branches.md
MaryaBelanger: Aha, I took the point of _"you can't use refutable patterns in places that don't do control flow"_ in the spec to mean the inverse (_you can't use irrefutable patterns in places that do control flow_) as well. But, reading again, the spec doesn't say that anywhere. Thanks for catching!

src/_guides/language/language-tour.md
```diff
@@ -186,31 +186,31 @@ This content has moved to [Operators](/language/operators#other-operators).
 
 ## Control flow statements
 
-This content has moved to [Control flow](/language/control-flow).
+This content has moved to [Control flow](/language/loops).
```
Hixie, line 189: Most people reading this document aren't going to have read it before, so they don't care that the content was here before and moved.

I would phrase this as something like, "For information on control flow statements, see ..."

src/language/branches.md
Hixie: i would prefer to document the entire language than have hidden features. People will come across code that uses this, and then go looking for what it means, and the language tour is the only approachable language reference we provide, so if a feature is missing from the guide they'll be lost.

src/language/branches.md
Hixie: or maybe we should fix the formatter to do it :-)
this is way more readable than not doing it...

src/language/branches.md
Hixie: maybe "Guard clause (`when`)"

src/language/collections.md
```diff
@@ -293,6 +293,12 @@ to create a list with three or four items in it:
 var nav = ['Home', 'Furniture', 'Plants', if (promoActive) 'Outlet'];
 ```
 
+Dart also supports [if-case][] inside collection literals:
+
+```dart
+var nav = ['Home', 'Furniture', 'Plants', if (promoActive case String x) $x];
```
Hixie, line 299: what is `promoActive` in this example? I cannot guess what this code does by looking at it!

src/language/control-flow.md
```diff
@@ -1,306 +0,0 @@
----
```
Hixie, line 1: is there an HTTP redirect for people who get to this page from old links?

(not a JS-based redirector)

firebase.json
```diff
@@ -80,6 +80,7 @@
       { "source": "/codelabs/server{,/**}", "destination": "/tutorials/server/httpserver", "type": 301 },
       { "source": "/code-of-conduct", "destination": "/community/code-of-conduct", "type": 301 },
       { "source": "/concurrency", "destination": "/language/concurrency", "type": 301 },
+      { "source": "/control-flow", "destination": "/language/loops", "type": 301 },
```
parlough, line 83:
```suggestion
      { "source": "/language/control-flow", "destination": "/language/loops", "type": 301 },
```

src/language/collections.md
Hixie: oh is the idea that `promoActive` would be a `String?` and this is equivalent to `if (promoActive is String) promoActive`?

What is the `$` on `$x`?

I find this example very confusing.

MaryaBelanger: I just grabbed this example from the [existing collection literals content](https://dart.dev/language/collections#control-flow-operators), which also doesn't give any context.

> What is the `$` on `$x`?

Ahh, idk! Thanks for calling this out, I'll come up with something clearer 🤞 (or hopefully someone else caught it and suggested a better example :))

src/_guides/language/language-tour.md
MaryaBelanger: We [took the language tour down](https://github.com/dart-lang/site-www/pull/4612) and made it unsearchable, so the only people that will end up on that page actually are people who already have the link saved

src/language/branches.md
MaryaBelanger: There's no patterns page yet, so it just goes to the home page for now. All of these dart 3 prs will be merged into the `v3.0` branch before `main`, so I was thinking I'd fill in the actual links in another pr on that branch once they're all in.

src/language/branches.md
MaryaBelanger: I was thinking about merging those sections together myself but wasn't sure if it was important to distinguish "statements"; thanks for the confirmation!

src/language/branches.md
MaryaBelanger: I went with your version, except removed "patterns" from the end (just "cases") because of [Erik's idea](https://github.com/dart-lang/site-www/pull/4780#discussion_r1176761413) to not get into "case patterns" terminology too much here

src/language/branches.md
MaryaBelanger: Another case for needing fuller reference type docs per language topic, tracking here: #4783 

I'll leave the example for now, but will add this to the TODO list for reference docs.

src/language/assert.md
```diff
@@ -0,0 +1,75 @@
+---
+title: Assert
+description: Learn how to use the assert statement to manipulate control flow in Dart.
+toc: false
+---
+
+You can control the flow of your Dart code using `assert`, and other control flow
+statements:
```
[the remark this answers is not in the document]
MaryaBelanger: Not out of scope, I'll move assert to the exceptions/error handling page. Thanks!

src/language/branches.md
MaryaBelanger: I'll change it back to normal but I wish this formatting was ok to use in the docs! It's so much clearer

src/language/branches.md
MaryaBelanger: removed this section per [bob's comment](https://github.com/dart-lang/site-www/pull/4780#discussion_r1176833411)

src/language/branches.md
MaryaBelanger: I can add their declarations to the code block.

src/language/branches.md
MaryaBelanger: Removed the line entirely in favor of just saying it can be used anywhere an expression can be

src/language/branches.md
MaryaBelanger: I'm going to omit the parenthetical because I think "the body of a case", regardless of how that body ended up associated with that case, reasonably covers shared case bodies

src/language/loops.md
MaryaBelanger: I didn't even think of that! I'll add it right under the non-pattern `for-in` example

MaryaBelanger: Very arbitrary, but I added:
```dart
for (var (x, y) in listOfPairs) {
  assert(x != y);
}
```

src/language/collections.md
MaryaBelanger: Changed it to:
```dart
var nav = ['Home', 'Furniture', 'Plants', if (login case 'Manager') 'Inventory'];
```
Any better?

src/language/control-flow.md
MaryaBelanger: I'm not sure, maybe @parlough or @domesticmouse can answer that

parlough: Yes, you set it up in the firebase.json: https://github.com/dart-lang/site-www/pull/4780#discussion_r1176971293

src/language/branches.md
```diff
@@ -0,0 +1,279 @@
...
+  ```  
+- **Logical** or **relational patterns**: the object matches if the comparison to
+a constant returns `true`.
+  ```dart
+  switch (code) {
+    case >= first && <= last: ...    
+    // Matches if the value of code is between the constant values of first and last
+  }
+  ```  
+- **Null check** or **assert patterns**: the match is first on whether the object is not null, then on the value.
```
munificent, line 142: I would split null-check and null-assert patterns into two separate sections since they behave pretty differently on null. (Also, the latter are called "null assert" patterns not just "assert" patterns since they're specific to handling null.)

Also, the wording is pretty strange. Maybe: If the object is `null`, then doesn't match. Otherwise, matches the object against the inner pattern.

src/language/branches.md
```diff
@@ -0,0 +1,279 @@
...
+  ```dart
+  switch (maybe) {
+    case var s?:                 
+    // Matches if maybe is not null, and then against the inner pattern "var s"
+  }
+  ```  
+- **Object patterns**: the match exposes getters on the object type.
```
munificent, line 149: Maybe "Matches if the object has the given named type. If so, then calls getters on the object and matches them against corresponding inner patterns.

munificent (approved):
Looks great!

src/language/branches.md
MaryaBelanger: Moved this section to the types reference page where we addressed this, resolving!

src/language/branches.md
MaryaBelanger: Also moved and addressed this one, resolving!

src/language/branches.md
```diff
@@ -0,0 +1,229 @@
...
+
+```dart
+if (pair case [int x, int y]) return Point(x, y);
+```
+
+If the pattern (the list pattern `[int x, int y]` in the example above)
+matches the value (`pair`), then the branch (`return Point(x, y)`)
+executes with any variables the pattern defines in scope (`x` and `y`).
+
+Otherwise, the else branch is executed, if there is one:
```
atsansone, line 45: **issue:** Write in active voice per [GSG](https://developers.google.com/style/voice).

```suggestion
Otherwise, Dart executes the `else` branch when possible.
```

src/language/branches.md
```diff
@@ -0,0 +1,229 @@
...
+  switch (shape) {
+    Square(length: var l) => l * l,
+    Circle(radius: var r) => math.pi * r * r
+  };
+```
+
+If a new subclass of `Shape` is ever added, exhaustiveness 
+checking will tell you about the missing subtype, because your switch is incomplete.
+This allows you to use Dart in a somewhat 
+[functional algebraic datatype style](https://en.wikipedia.org/wiki/Algebraic_data_type).
```
atsansone, line 194: **issue:** Write in active voice per [GSG](https://developers.google.com/style/voice). 

```suggestion
If you add a new subclass of `Shape`,
exhaustiveness checking reports the missing subtype.
The check flags your `switch` as incomplete.
This allows you to use Dart in a somewhat 
[functional algebraic datatype style](https://en.wikipedia.org/wiki/Algebraic_data_type). 
```

src/language/branches.md
```diff
@@ -0,0 +1,229 @@
...
+
+## Guard clause
+
+<a id="when"></a>
+
+An optional guard clause can appear after any case clause
+(`if case`, and both `switch` statements and expressions)
+using the keyword `when`. Guards evaluate an arbitrary boolean expression _after_
+matching, allowing you to further constrain whether a case body should execute
+despite the case pattern matching the value.
```
atsansone, line 204: **issues:** Write as shorter sentences per [GSG](https://developers.google.com/style/accessibility#ease-of-reading). Avoid parentheticals. Condition, then instruction per [GSG](https://developers.google.com/style/sentence-structure).
```suggestion
To set an optional guard clause after a `case` clause, use the keyword `when`.
You can use this keyword with `if case`, and both `switch` statements and expressions.
Guards evaluate an arbitrary boolean expression _after_ matching.
This allows you to determine if a `case` body should execute
despite the `case` pattern matching the value.
```

atsansone (changes requested):
@MaryaBelanger : Back to you. Great content!

src/language/branches.md
```diff
@@ -0,0 +1,229 @@
...
+
+Dart `if` statements support `case` clauses followed by a [pattern][]: 
+
+```dart
+if (pair case [int x, int y]) return Point(x, y);
+```
+
+If the pattern (the list pattern `[int x, int y]` in the example above)
+matches the value (`pair`), then the branch (`return Point(x, y)`)
+executes with any variables the pattern defines in scope (`x` and `y`).
```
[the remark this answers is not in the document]
MaryaBelanger: I was hoping someone would fix this! 😅

src/language/branches.md
```diff
@@ -0,0 +1,229 @@
...
+
+If-case is a simple way to match and destructure against a
+_single_ pattern. To test a value against _multiple_ patterns, use [switch](#switch).
+
+## Switch
+
+Switch statements evaluate a value expression against a series of cases.
+Each case clause is a [pattern][] to match against the value being switched on,
+followed by a series of statements in the case body to execute if the value
+matches that pattern. You can use [any kind of pattern][] for a case.
```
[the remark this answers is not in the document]
MaryaBelanger: I use backticks to refer to the `case` keyword, so if I'm talking about the code itself. But mostly when I'm talking about cases, it's just the concept, so no backticks. Other than that, done

src/language/branches.md
```diff
@@ -0,0 +1,229 @@
...
+
+Non-empty `case` clauses implicitly jump to the end of the switch after completion.
+Other valid ways to end a non-empty `case` clause are a [`continue`][break],
+[`throw`][], or [`return`][] statement.
```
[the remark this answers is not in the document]
MaryaBelanger: Same about backticks for switch.

I'm a little torn on "implicitly"; I included it to kind of pose a counter to the previous behavior, which was not implicit (required `break` to end).  So it's providing an explanation to people who are used to break ("now it implicitly ends"), without being too on the nose for those who are new to Dart (like including a note about how it "used to be"). Lmk what you think.

src/language/branches.md
```diff
@@ -0,0 +1,229 @@
...
+    executeNowClosed();   // Runs for both OPEN and PENDING.
+}
+```
+For an empty case that does not fall through, use [`break`][break] for its body.
+
+### Switch expressions
+
+Switch expressions are similar to switch statements, but you can use them anywhere
+you can use an expression. They produce a value based on the expression body of
+whichever case matches.
```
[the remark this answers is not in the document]
MaryaBelanger: Not a fan of breaking the sentence like that, removing the counter "but" waters down the difference. I'd argue that the sentence is not egregiously long, and two clause sentences are reasonable.

src/language/branches.md
```diff
@@ -0,0 +1,229 @@
...
+  comma || semicolon => punctuation(charCode),
+  >= digit0 && <= digit9 => number(),
+  _ => throw invalid()
+};
+```
+
+### Exhaustiveness checking
+
+Exhaustiveness checking is a feature of switches that reports a compile-time
+error if it's possible for a value entering the switch to not match any of the cases.
```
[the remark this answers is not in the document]
MaryaBelanger: this is one of those things I'm hesitant to change now that we're past technical reviews. Exhaustiveness checking happens _to_ switches... I know that saying it's `switch` that's responsible for exhaustiveness checking would not go over well; it would probably be the compiler or something, but it's weird to bring up the compiler in the language tour... I changed it, when you see the next commit see what you think

src/language/branches.md
MaryaBelanger: The benefit here is that if a subclass is added _unbeknownst to you_, exhaustiveness checking would alert you to it, so using "you" as the actor doesn't work for the selling point. I tried this:

> If anyone were to add a new subclass of `Shape`, this `switch` expression would 
be incomplete. Exhaustiveness checking would inform you of the missing subtype.

src/language/branches.md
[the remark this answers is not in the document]
atsansone: **comment:** Be direct. You can say:
```markdown
on-empty `case` clauses jump to the end of the `switch` after completion without requiring a `break`.
```

src/language/branches.md
```diff
@@ -108,11 +116,11 @@ For an empty case that does not fall through, use [`break`][break] for its body.
 
 ### Switch expressions
 
-Switch expressions are similar to switch statements, but you can use them anywhere
-you can use an expression. They produce a value based on the expression body of
-whichever case matches.
+A `switch` expression is similar to a `switch` statement, but you can use them
```
atsansone, line 119: **suggestion:** Active voice.
```suggestion
A `switch` expression works like a `switch` statement, but you can use them
```

src/language/branches.md
```diff
@@ -151,8 +159,8 @@ token = switch (charCode) {
 
 ### Exhaustiveness checking
 
-Exhaustiveness checking is a feature of switches that reports a compile-time
-error if it's possible for a value entering the switch to not match any of the cases.
+Exhaustiveness checking is a feature that reports a compile-time
```
atsansone, line 162: **nit:** Consider shortening this.
```suggestion
Exhaustiveness checking reports a compile-time
```

src/language/loops.md
```diff
@@ -25,8 +27,8 @@ for (var i = 0; i < 5; i++) {
 }
 ```
 
-Closures inside of Dart’s `for` loops capture the _value_ of the index,
-avoiding a common pitfall found in JavaScript. For example, consider:
+Closures inside of Dart’s `for` loops capture the _value_ of the index.
+This avoidds a common pitfall found in JavaScript. For example, consider:
```
atsansone, line 31: **typo:** 
```suggestion
This avoids a common pitfall found in JavaScript. For example, consider:
```

src/language/branches.md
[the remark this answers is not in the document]
atsansone: **nit:** You mention the backticks but don't write what you mean. If you don't want to add them in this pass, don't. Let's land this.

atsansone (approved):
@MaryaBelanger : Looks good. There's a typo and a few nits/suggestions. No need for another pass through me.

src/language/branches.md
MaryaBelanger: Hmm.. I'm gonna keep it because "exhaustiveness checking" sounds like a verb, but it's kind of actually a "thing"... I guess. I think clarity helps here.

src/language/branches.md
MaryaBelanger: Gonna skip, the difference is more in how they work. Their similarities are also in how they work, but I just don't want to start with that.

src/language/branches.md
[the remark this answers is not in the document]
MaryaBelanger: Got it, using the more direct suggestion. 

I thought I wrote about the backticks in another comment, but I guess not. I'll keep them for now.

PR #4781, opened 2023-04-24
Add a dependency override for analyzer

parlough:
Pins analyzer to `5.10.0` due to an incompatibility between `5.11.0` and the latest published `dart_style`: https://github.com/dart-lang/sdk/issues/52159


parlough:
Looks like another analyzer version will be released with a fix, so might want to hold off on this in favor of the actual fix.

parlough:
Analyzer 5.11.1 was released which adds back the missing members. Build should be fixed with that :)

PR #4785, opened 2023-04-25
Add local note about old language tour use

parlough:


MaryaBelanger (approved):
Good idea. 

So recently when I moved the old control flow section a second time, I don't need to come in here and update it, just the map, right?

MaryaBelanger (approved):
Thanks for clarifying, approved again!

PR #4786, opened 2023-04-25
Update most SDK constraints to 3.0 and make adjustments

parlough:
Beyond updating the constraints, the primary change here is changing some classes to mixins and adjusting the [DO use = to separate a named parameter from its default value](https://dart.dev/guides/language/effective-dart/usage#do-use--to-separate-a-named-parameter-from-its-default-value) Effective Dart guideline. I've made an issue for future work there in https://github.com/dart-lang/site-www/issues/4787.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,8 +37,8 @@ jobs:
             experimental: false
           - sdk: beta
             experimental: false
-          - sdk: stable
-            experimental: true
+#          - sdk: stable
+#            experimental: true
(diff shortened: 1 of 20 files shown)


parlough:
@MaryaBelanger This will allow you to add Dart 3 code excerpts if you want to as part of your work, as well as make adjustments like removing `break` in existing snippets if you want.

MaryaBelanger (approved):
Thanks so much!

PR #4788, opened 2023-04-25
Update type_system example to use latest SDK constraint

parlough:
Have to also adjust for a new warning introduced with the new language version.


MaryaBelanger (approved):
Thanks!

PR #4790, opened 2023-04-26
Adjust Dart 3 alpha mentions to beta

parlough:


mit-mit (approved):
good catch!

but we should remember to update this again at 3.0 stable, so perhaps file an issue?

parlough:
Good idea, thanks. I filed an issue at https://github.com/dart-lang/site-www/issues/4791.

PR #4792, opened 2023-04-26
Add Dart 3 to spec page

parlough:
So I saw two ways to go about this change:

1. Separate Dart 3 from Dart 2, and keep the Dart 2 entry focused on versions 2.10 and earlier
2. Combine Dart 3 and 2, since the changes are minimal in terms of 2.12->2.19 vs 3.

I chose 1, since the separation felt less confusing, so users don't have to keep in mind a 2.12 vs before 2.12 world anymore. Open to other directions though :)

Staged: https://dart-dev--pr4792-dart-3-spec-ir63p46a.web.app/guides/language/spec

Closes https://github.com/dart-lang/site-www/issues/4793

### DIFF
--- a/src/_guides/language/spec.md
+++ b/src/_guides/language/spec.md
@@ -8,26 +8,43 @@ Use this page to find the formal Dart language specification.
 For a gentler introduction to Dart, see the
 [language tour](/language).
 
-## Dart 2
+## Dart 3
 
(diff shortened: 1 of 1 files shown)


MaryaBelanger (approved):
I think you made the right choice; the light explanation does a lot of work. Thanks!

atsansone (approved):
Makes sense.

PR #4796, opened 2023-04-27
Enable strict liquid checks to catch potential issues

parlough:
They prevents potential issues from landing on the site, as the CI will catch them and indicate the issue, so they can be addressed before PRs land.


atsansone (approved):
**question:** Do we know how strict checks could affect deploy or staging?

parlough:
They will prevent staging if someone breaks one of the checks, but the CI will indicate why and it can be addressed before merging.

Overall, it prevents potential issues showing up on the site, as they will instead be caught in PRs.

PR #4799, opened 2023-04-28
Mention `dart pub add`

mit-mit:


src/_guides/packages.md
```diff
@@ -54,6 +54,18 @@ dependencies:
   intl: ^0.17.0
 ```
 
+Rather than editing the pubspec manually, 
+you can also add a dependency by running a `dart` tool command. 
+This adds a dependency on `vector_math`:
```
atsansone, line 59: **issue:** Write in active voice per [GSG](https://developers.google.com/style/voice).
```suggestion
To update the `pubspec.yaml` file, you have a better option than hand editing. 
To add a dependency, you can run `dart pub add` command.
The following example adds a dependency on `vector_math`.
```

atsansone (changes requested):
@mit-mit : Need to rephrase to meet Google Style.

src/_guides/packages.md
mit-mit: Thanks! Committed a variant of that (wanted to avoid "hand" editing)

atsansone (approved):
LGTM!

Issue #4696, opened 2023-03-20
Document pubspec support for `topics:` field


parlough:
Reference: https://github.com/dart-lang/pub/pull/3796 and https://github.com/dart-lang/pub-dev/pull/6464

parlough:
@szakarias Have topics reached a point with https://github.com/dart-lang/pub-dev/pull/6601 that we should be documenting the field now, or not yet?

PR #4807, opened 2023-05-02
Add documentation for topics

szakarias:
Closes https://github.com/dart-lang/site-www/issues/4696


src/tools/pub/pubspec.md
```diff
@@ -410,6 +414,37 @@ Each download of the package includes all screenshot files.
...
+
+- At most 5 topics per package.
+- The length of the topic name should be 2-32 characters.
+- The topic name should consist of lowercase alphanumeric characters or dash
+  (but no double dash), starting with a-z and ending with a-z or 0-9.
+
+
+When choosing topics it is good practice to look at
+[existing topics]({{site.pub}}/topics) to see if there are established topics
+that match your package.
```
atsansone, line 442: **issue:** Write in active voice per [GSG](https://developers.google.com/style/voice).

```suggestion
When choosing topics, first look at
[existing topics]({{site.pub}}/topics) to check if established topics
exist that match your package.
```

src/tools/pub/pubspec.md
```diff
@@ -410,6 +414,37 @@ Each download of the package includes all screenshot files.
...
+```
+
+Pub.dev limits topics to the following specifications:
+
+- At most 5 topics per package.
+- The length of the topic name should be 2-32 characters.
+- The topic name should consist of lowercase alphanumeric characters or dash
+  (but no double dash), starting with a-z and ending with a-z or 0-9.
+
+
```
atsansone, line 439: **suggestion:** Add a heading for "Best Practices".
```markdown
#### Best Practices
```

src/tools/pub/pubspec.md
```diff
@@ -410,6 +414,37 @@ Each download of the package includes all screenshot files.
...
+When choosing topics it is good practice to look at
+[existing topics]({{site.pub}}/topics) to see if there are established topics
+that match your package.
+
+Another good idea, if you have a number of related packages, for instance
+belonging to the same framework, it is good practice to tag of all these
+packages with the same topic, e.g. the name of the framework.
```
atsansone, line 446: **issues:** Rephrase to make more concise and readable. 
Write in active voice per [GSG](https://developers.google.com/style/voice).

```suggestion
If you have a number of related packages,
tag of all these packages with the same topic.
Such packages could belong to the same framework
and be tagged with the name of the framework.
```

atsansone (changes requested):
@szakarias : A few small changes and I can approve.

src/tools/pub/pubspec.md
szakarias: done

src/tools/pub/pubspec.md
szakarias: done

src/tools/pub/pubspec.md
jonasfj: We should perhaps consider being less specific here.

Using a word like "may" is or say "may wish to", is perhaps a good idea.

There may be established topics that match your package, but which you don't want to use.
They may be too generic (then they still technically match your package), but it may be that you want to establish a new more specific topic -- that's fine, apply human judgement.

src/tools/pub/pubspec.md
jonasfj: We don't do this for other subjects on the page.

We could put it in an "Note" block, but I think that's overkill.

src/tools/pub/pubspec.md
jonasfj: > If you have a number of related packages,
> tag of all these packages with the same topic.

This changes the sentence from a recommendation to an instruction.

I think we should give advice, not instructions.
Package authors are writing code, this is complicated and we rely on them using their best judgement when it comes to presenting their package.

Hence, I would suggest we include words such as "may" and "should" or "consider", to explain to readers whether something is a suggestion or requirement.

> Such packages could belong to the same framework
> and be tagged with the name of the framework.

This just reads wrong to me.

src/tools/pub/pubspec.md
```diff
@@ -410,6 +414,38 @@ Each download of the package includes all screenshot files.
...
+
+#### Best Practices
+
+When choosing topics, first look at [existing topics]({{site.pub}}/topics) to
+check if established topics exist that match your package.
+
+If you have a number of related packages, tag of all these packages with the
+same topic. Such packages could belong to the same framework and be tagged with
+the name of the framework.
+
```
jonasfj, line 448:
```suggestion
When choosing topics it is good practice to look at
[existing topics]({{site.pub}}/topics) to see if there are established topics
that match your package.

Another good idea, if you have a number of related packages, for instance
belonging to the same framework, it is good practice to tag of all these
packages with the same topic, e.g. the name of the framework.
```
I suggest we go back to the original proposal.

--------------

```
#### Best Practices
```
We don't do this other places on this page.


> When choosing topics, first look at [existing topics]({{site.pub}}/topics) to
> check if established topics exist that match your package.

This is an instruction and not a recommendation or suggestion.
The original intent here is to suggest that users consider existing topics.


> If you have a number of related packages, tag of all these packages with the
> same topic. Such packages could belong to the same framework and be tagged with
> the name of the framework.

Again, we should suggest not instruct. Also the part about frameworks does not make sense to me.

jonasfj: If we wanted a different text I'd suggest:

```suggestion
When publishing a package consider if [existing topics]({{site.pub}}/topics)
are relevant. Tagging with existing topics helps users discover your package.

If you have a number of related packages, for example packages making up a framework,
consider establishing a new topic linking these packages together. This makes it easy
to discover related packages, and allows other package authors to tag their packages
as being relevant for your framework. For example, check out
[`topic:shelf`](https://pub.dev/packages?q=topic%3Ashelf).
```

We can also omit these lines entirely and just put the recommendations in `publishing.md`.

parlough: \cc @atsansone and @szakarias for thoughts on content and text

I think you are right though that these extensive details don't belong here and could be a section in [Publishing](https://dart.dev/tools/pub/publishing).

jonasfj (approved):
LGTM

src/tools/pub/pubspec.md
```diff
@@ -410,6 +414,31 @@ Each download of the package includes all screenshot files.
 Pub.dev generates the package's thumbnail image from the first screenshot. If 
 this screenshot uses animation, pub.dev uses its first frame.
  
+### Topics
+
+Package authors can use the `topics` field to categorize their package. Topics
+can be used to assist discoverability during search with filters on pub.dev.
+Pub.dev displays the topics on the package page as well as in the search
+results.
```
atsansone, line 422: **nit:** Simplify language a little. You can ignore this if you disagree.
```suggestion
Package authors can use the `topics` field to categorize their package. Topics
assist discoverability during search with filters on pub.dev.
Pub.dev displays the topics on the package page as well as in the search
results.
```

src/tools/pub/pubspec.md
```diff
@@ -410,6 +414,31 @@ Each download of the package includes all screenshot files.
...
+  - network
+  - http
+```
+
+Pub.dev limits topics to the following specifications:
+
+- At most 5 topics per package.
+- The length of the topic name should be 2-32 characters.
+- The topic name should consist of lowercase alphanumeric characters or dash
+  (but no double dash), starting with a-z and ending with a-z or 0-9.
```
atsansone, line 437: **issue:** This needs parallel structure, per [GSG](https://developers.google.com/style/lists#parallel-syntax)

```suggestion
Pub.dev requires topics to follow these specifications:

- Use no more than 5 topics per package.
- Set the length of the topic name to between 2 and 32 characters.
- Write the topic name following these requirements:
  - Use only lowercase alphanumeric characters or hyphens (`a-z`, `0-9`, `-`).
  - Don't use two consecutive hyphens.
  - Start the name with lowercase alphabet characters (`a-z`).
  - End with alphanumeric characters (`a-z` or `0-9`).
```

atsansone (changes requested):
A few changes to consider before merging.

src/tools/pub/pubspec.md
jonasfj: > - Set the length of the topic name to between 2 and 32 characters.

I don't think it makes sense to write that.

---------------------

We could perhaps do:
```suggestion
Pub.dev limits topics for a package as follows:
- At most 5 topics per package.
- Each topic name must:
  - Use only lowercase alphanumeric characters or hyphens (`a-z`, `0-9`, `-`).
  - Don't use two consecutive hyphens.
  - Start the name with lowercase alphabet characters (`a-z`).
  - End with alphanumeric characters (`a-z` or `0-9`).
  - Be between 2 and 32 characters long.
```

szakarias: Updated the bullet list.

src/tools/pub/pubspec.md
```diff
@@ -410,6 +414,34 @@ Each download of the package includes all screenshot files.
...
+
+Pub.dev requires topics to follow these specifications:
+
+- Tag each package with at most 5 topics.
+- Write the topic name following these requirements:
+  - Use between 2 and 32 characters.
+  - Use only lowercase alphanumeric characters or hyphens (`a-z`, `0-9`, `-`).
+  - Don't use two consecutive hyphens (`--`).
+  - Start the name with lowercase alphabet characters (`a-z`).
+  - End with alphanumeric characters (`a-z` or `0-9`).
```
jonasfj, line 440: This is not the specification I was promised on the line above.

parlough (approved):
Thanks for this and making those adjustments!

If anyone wants further changes, they can be handled in follow up work. Since topics were already announced I really want to get this in.

PR #4810, opened 2023-05-02
[Dart 3] Patterns

MaryaBelanger:
Please ignore missing links, that'll come once the contents are all merged into the v3.0 branch!

New:
- [patterns.md](https://dart-dev--pr4810-patterns-p1em9l86.web.app/language/patterns)
- [pattern-types.md](https://dart-dev--pr4810-patterns-p1em9l86.web.app/language/pattern-types)


src/language/patterns.md
```diff
@@ -0,0 +1,334 @@
...
+title: Patterns
+description: Summary of patterns in Dart.
+---
+
+{{site.alert.version-note}}
+  Patterns require a [language version][] of at least 3.0.
+{{site.alert.end}}
+
+Patterns are a syntactic category in the Dart language, like statements and expressions.
+They represent the form of a value, not an actual value.
```
leafpetersen, line 11: Maybe "shape of values"?

src/language/patterns.md
```diff
@@ -0,0 +1,334 @@
...
+
+To learn about the different types of patterns, visit the [pattern type reference][types].
+
+## Matching
+
+A pattern is always tested against a value to determine if the value has the form
+you expect, or in other words, if the value _matches_ the pattern. 
+
+What constitutes a match depends on the [pattern type][types].
+For example, matching a constant pattern means the value is equivalent to the pattern:
```
leafpetersen, line 26: I don't know what "the value is equivalent to the pattern" means here.  What is this sentence trying to say?  Something like "For example, the following switch usess a constant pattern to test that `number` is `1`?  I'm not sure.

src/language/patterns.md
```diff
@@ -0,0 +1,334 @@
...
+
+```dart
+switch (number) {
+  case 1: ...        // Constant pattern 1 matches if number == 1
```
leafpetersen, line 30: Technically, this tests `1 == number`, not sure if we want to be persnickety about it.

src/language/patterns.md
```diff
@@ -0,0 +1,334 @@
...
+For example, the individual fields of any [collection-type][] pattern could be 
+[variable patterns][variable] or [constant patterns][constant]:
+
+```dart
+switch (obj) {       // List pattern [a,b] matches obj first if obj
+  case [a, b]: ...   // is a list wth two fields, then if its  
+}                    // fields match the constant subpatterns a and b  
+```
+
+Patterns can also _partially_ match, by using a [wildcard pattern][] as a place holder,
```
leafpetersen, line 45: partially match is a somewhat odd term.  The pattern fully matches, it just "doesn't care" about certain parts of the value.  Maybe something like "Patterns can ignore pieces of the matched value that they don't care about by using a..."?

src/language/patterns.md
```diff
@@ -0,0 +1,334 @@
...
+or a [rest element][] in the case of list patterns.
+
+If you need to ensure `null` values aren't silently treated as match failures,
```
leafpetersen, line 48: This seems a little hard to situate without some context.  Maybe an example here?

src/language/patterns.md
```diff
@@ -0,0 +1,334 @@
...
+
+```dart
+var numList = [1, 2, 3];
+var [a, b, c] = numList;   // List pattern [a, b, c] destructures the list object
+print(a + b + c);          // numList into individual new parts. Prints 6.
+```
+
+You can use the [null-check pattern][] to check for `null` while destructuring. 
+
+All patterns attempt to match the objects they're compared against, but not all
```
leafpetersen, line 64: I'm not sure about this paragraph.  What precisely do you want to the reader to take away here?  It's true that whether a pattern destructures or not depends on the type of pattern, not sure about the context.  Maybe there's some confusion around destructure in the sense of "destructuring assignment" vs "extract into parts"?  That is, the following pattern destructures (extracts) a list into parts, whether it is used is in a destructuring assignment, or in a switch:

```
  var [a, b] = someList; // Destructuring assignment: unconditionally extract the parts from something that must be a list.
  switch (someList) {
     case [var a, var b]: // Switch case: extract the parts from something that might be a list, if it is a list (and otherwise fall through to the next case).
  }
```

src/language/patterns.md
```diff
@@ -0,0 +1,334 @@
...
+When an object and pattern match, the pattern can then access the object's data 
+and extract it in parts. In other words, the pattern _destructures_ the object:
+
+```dart
+var numList = [1, 2, 3];
+var [a, b, c] = numList;   // List pattern [a, b, c] destructures the list object
+print(a + b + c);          // numList into individual new parts. Prints 6.
+```
+
+You can use the [null-check pattern][] to check for `null` while destructuring.
```
leafpetersen, line 62: Well, this gets to my comment below.  You can use the null check pattern to destructure a value in a switch, but not in a destructuring assignment (which I think is more what this section is about).

leafpetersen: Or maybe you don't indent this to just be about destructuring assignment?  But then it seems somewhat redundant with the previous section?  I guess I'm not clear on how these two sections fit together.

src/language/patterns.md
```diff
@@ -0,0 +1,334 @@
...
+of the declaration, it destructures the value, binding it to new local variables:
+
+```dart
+var (a, [b, c]) = ("str", [1, 2]);   // Declares new variables a, b, and c.
+```
+
+A pattern variable declaration must start with either `var` or `final`, followed
+by a pattern. 
+
+Use the [null-assert pattern][] to eliminate `null` in variable declarations.
```
leafpetersen, line 94: This should probably be phrased more carefully.  It eliminates `null` by throwing an exception, which makes this dangerous as blanket advice.

src/language/patterns.md
```diff
@@ -0,0 +1,334 @@
...
+One convenient use case is swapping the values of two variables without needing
+to declare a third temporary one:
+
+```dart
+var (a, b) = ('left', 'right');
+(b, a) = (a, b);                   // Swap
+print('$a $b');                    // Prints "right left".
+```
+
+### Destrcuturing multiple returns
```
leafpetersen, line 111: Destructuring (sp)

src/language/patterns.md
```diff
@@ -0,0 +1,334 @@
...
+
+The `case` keyword is always followed by a pattern. _Case patterns_ are [refutable][];
+they allow control flow to either match and destructure the object being switched
+on, or continue execution if the object doesn't match. Case patterns can be any
+kind of pattern:
+
+- [**Constant patterns**][constant]: the object's value has to be equivalent to match.
+  ```dart
+  switch (number) {
+    case 1: ...          // Matches if number == 1
```
leafpetersen, line 147: Technically `1 == number`

src/language/patterns.md
```diff
@@ -0,0 +1,334 @@
...
+- [**List**][list], [**map**][map], or [**record**][record] **patterns**: the elements are destructured if matched.
+  ```dart
+  switch (obj) {
+    case [a, b]: ...     // Matches if obj is a list with two elements  
+  }
+  ```  
+- [**Variable patterns**][variable]: the matched value binds to a new variable.
+  ```dart
+  switch ((1, 2)) {
+    case (var a, var b): ...    // Matches if 1 and 2 can be assigned to a and b
```
leafpetersen, line 159: This one is a bit odd.  As written, it always matches: that is "if 1 and 2 can be assigned to a and b" is always true.  Maybe change `(1, 2)` to `obj` and replace the comment with `// Matches if obj is a record with two fields, and if so assign the fields to a and b`?

src/language/patterns.md
```diff
@@ -0,0 +1,334 @@
...
+  }
+  ```  
+- [**Logical**][logical] or [**relational**][relational] **patterns**: the object matches if the comparison to
```
leafpetersen, line 162: "The object matches if the comparison.." describes relational patterns only, so it reads a bit odd here when used to decribe "Logical or relational patterns".

src/language/patterns.md
```diff
@@ -0,0 +1,334 @@
...
+}
+
+class Circle implements Shape {
+  final double radius;
+  Circle(this.radius);
+
+  double calculateArea() => math.pi * radius * radius;
+}
+
+double calculateArea(Shape shape) =>
```
leafpetersen, line 251: This code is an error as written, since switch expressions must be exhaustive.  You could make `Shape` be `sealed`, or you could use a switch statement, or you could add a default case.  Given that you are talking about algebraic datatypes here, I think the first solution is the correct one, since that's the use case that `sealed` addresses.

src/language/pattern-types.md
```diff
@@ -0,0 +1,333 @@
...
+  `(a, b) = (3, 4);`
+- [Matching][] context: treated as a named constant pattern (unless its name is `_`):
+  ```dart
+  const c = 1;
+  switch (2) {
+    case c: print('match $c');
+    default: print('no match');    // Prints "no match".
+   }
+  ``` 
+- [Wildcard](#wildcard) identifier in any context: matches any value and discards it:
```
leafpetersen, line 61: I'd be inclined to pull this out to a separate ## Wildcard entry myself.  They're not really identifier patterns per se, and certainly users won't think of them that way.

src/language/pattern-types.md
```diff
@@ -0,0 +1,333 @@
...
+  `case [_, var y, _]: print('The middle element is $y');`
+
+## List
+
+`[subpattern1, subpattern2]`
+
+List patterns match values that implement [`List`][], and then recursively match
+its subpatterns against the list's elements to destructure them by position.
+
+List patterns require that the pattern match the entire list, but you can use
```
leafpetersen, line 71: Worth mentioning that in a declaration or assignment context, if the number of elements don't match an exception is thrown?

src/language/pattern-types.md
```diff
@@ -0,0 +1,333 @@
...
+var [a, b, ...rest, c, d] = [1, 2, 3, 4, 5, 6, 7];
+print('$a $b $rest $c $d'); // Prints "1 2 [3, 4, 5] 6 7".
+```
+
+## Logical-and	
+
+`subpattern1 && subpattern2`
+
+A pair of patterns separated by `&&` matches only if both subpatterns match. If the
+left branch does not match, the right branch is not evaluated.
```
leafpetersen, line 97: Maybe mention that the subpatterns may bind variables, but that they must be disjoint?

src/language/pattern-types.md
```diff
@@ -0,0 +1,333 @@
...
+Switch statements can have multiple cases share a body without using logical-or
+patterns, but they are still uniquely useful for allowing multiple cases to share
+a guard: 
+
+```dart
+switch (shape) {
+  case Square(size: var s) || Circle(size: var s) when s > 0:
+    print('Non-empty symmetric shape');
+}
+```
```
leafpetersen, line 138: Maybe mention that or patterns may bind variables, but both sub-patterns must bind the same variables (as in your example)?

src/language/pattern-types.md
```diff
@@ -0,0 +1,333 @@
...
+
+## Map
+
+`{"key": subpattern1, someConst: subpattern2}`
+
+Map patterns match values that implement [`Map`][], and then recursively 
+match its subpatterns against the map’s keys to destructure them.
+
+Map patterns don't require the pattern to match the entire map. If a map has
+extra keys that te pattern doesn't destructure, it can still match.
```
leafpetersen, line 148: te -> the

src/language/pattern-types.md
```diff
@@ -0,0 +1,333 @@
...
+
+Object patterns checks the matched value against a given named type to destructure
+data using the getters the object's class already defines. They are [refuted][]
+if the value doesn't have the same type.
+
+The getter name can be omitted and inferred from the [variable pattern](#variable)
+or [identifier pattern](#identifier) in the field subpattern:
+
+```dart
+var Point(:var x, :var y) = Point(1, 2);
```
leafpetersen, line 195: No `var` on x and y in a declaration or context (only in a matching context).

src/language/pattern-types.md
leafpetersen: Ah, I see you did so below.  :)

src/language/pattern-types.md
```diff
@@ -0,0 +1,333 @@
...
+The variables are in scope in a region of code that is only reachable when the
+pattern has matched.
+
+```dart
+switch ((1, 2)) {
+  case (var a, var b): // ...
+  // var a and var b are variable patterns that bind to 1 and 2, respectively.
+  // They are in scope in the case body.
+}
+```
```
leafpetersen, line 294: Maybe mention that a typed variable pattern only matches if the matched value has the declared type (and otherwise the match fails)?

src/language/patterns.md
```diff
@@ -0,0 +1,334 @@
...
+- [for and for-in loops][for]
+- [if-cases] and [switch-cases]
+- [Collection literals][]
+
+This section describes common use cases for matching and destructuring with patterns, 
+combining various contexts, pattern types, and surrounding code constructs.
+
+### Variable declaration
+
+Anywhere Dart allows regular, single-name local variable declaration, you can also
```
munificent, line 83: The "single-name" part throws me off a little. Maybe just: "In the body of a function where you can declare local variables, you can also use a _pattern variable declaration._"

Or do you intend to cover patterns in for and for-in statements/elements here too? (The variable declarations in those are also locals.)

src/language/patterns.md
munificent: +1. I just wouldn't talk about `!` here.

src/language/patterns.md
munificent, line 111: This section is an outlier because the other sections are describing distinct places in the language grammar where you can use a pattern. This section is more like a use case or idiom for patterns.

Maybe instead of having this as a separate section, you could just use this code as an example in the pattern variable declaration section sort of like how you use swapping as an example idiom under pattern assignment?

Taking a step back, maybe that's a smoother way to integrate the "places you can use a pattern" and "common idioms for patterns" together? For each place you use can use a pattern, the idiom then serves as a concrete example. Something like:

* Pattern variable declarations -> Destructuring multiple returns
* Pattern assignment -> Swapping variables
* Pattern variables in for and for-in -> The `MapEntry(...)` example of using an object pattern for destructuring
* Switch statements and expressions -> Algebraic datatype-style code with exhaustiveness checking over a sealed class's subtypes

src/language/patterns.md
```diff
@@ -0,0 +1,334 @@
...
+
+```dart
+main() {
+  var (name, age) = userInfo(json);
+}
+```
+
+### Switch cases
+
+The `case` keyword is always followed by a pattern. _Case patterns_ are [refutable][];
```
munificent, line 139: This is true but doesn't also cover switch *expressions* where the pattern isn't preceded by `case`. (It was in early versions of the proposal, but we changed the syntax.)

Maybe say "Every case in a switch statement or expression contains a pattern."

src/language/patterns.md
```diff
@@ -0,0 +1,334 @@
...
+  ```  
+- [**Object patterns**][object]: the match exposes getters on the object type.
+  ```dart
+  switch (Shape shape) {
+    case Rect(width: var w, height: var h): ...
+    // Matches if shape is of type Rect, and then against the properties of Rect
+  }
+  ```  
+
+Values destructured by a pattern in a case become local variables,
```
munificent, line 185: I'm not sure if it makes a lot of sense to list out all of the kinds of patterns here when they are also in the other reference page. Also, most of these patterns aren't specific to switch cases, so it arguably makes as much sense to have them here as it would to have them under any of the preceding sections: you can have object, list, and map patterns in pattern variable declarations and pattern assignments too.

src/language/patterns.md
```diff
@@ -0,0 +1,334 @@
...
+Binding the result of a getter call to a variable of the same name is a common
+use case, so object patterns can also infer the getter name from the
+[variable subpattern][variable]. This allows you to simplify the variable pattern
+from something redundant like `key: key` to just `:key`:
+
+```dart
+for (var MapEntry(:key, value: count) in hist.entries) { 
+  print('$key occurred $count times');
+}
+```
```
munificent, line 219: This section feels a little tacked on because it's more or less just a description of how object patterns work in general, so it overlaps the previous object pattern bullet point and the section on object patterns in the reference. Maybe expand/merge this into section that talks about using patterns in for/for-in loops (which isn't otherwise shown on this doc) and make this an example under that section?

src/language/patterns.md
```diff
@@ -0,0 +1,334 @@
...
+
+### Algebraic datatypes
```
munificent, line 221: If you don't want to move this as an example under the switch cases section, then I think it would help to separate out the "places you can use patterns" and "idioms patterns are helpful for" sets of sections under different subheadings like:

```
## Places patterns can appear

### Pattern variable declarations

...

### Pattern assignments

...

### Switch cases

...

## Use cases for patterns

### Multiple returns

...

### Algebraic datatype-style code

...

### Validating JSON

...

```

src/language/patterns.md
```diff
@@ -0,0 +1,334 @@
...
+      user[0] is String && user[1] is int) {
+    var name = user[2] as String;
+    var age = user[1] as int;
+    print('User $name is $age years old.');
+  }
+}
+```
+
+A single [case pattern](#switch-cases) in a switch achieves the same validation
+as above, but is much clearer, more declarative, and much shorter:
```
munificent, line 290: This might be a good place to use `if-case` instead of a switch since there's only a single case.

src/language/pattern-types.md
```diff
@@ -0,0 +1,333 @@
+---
+title: Pattern types
+description: Pattern type reference in Dart.
+---
+
```
munificent, line 5: Maybe a one-sentence intro here that links back to the other patterns page, like:

```
Here are all the kinds of [patterns][] that can be used.
```

src/language/pattern-types.md
```diff
@@ -0,0 +1,333 @@
...
+`123, null, 'string', math.pi, SomeClass.constant, const Thing(1, 2), const (1 + 2)`
+
+Constant patterns allow simple literals and references to named constants to be
+used directly as patterns. They match when the value is equal to the constant.
+
+Some valid constant expressions may not be valid constant patterns. In that case,
+to distinguish a constant expression from a constant pattern, prefix the expression
+with `const` and wrap it in parenthesis so it is unambiguosly an expression and
+not a constant pattern:
```
munificent, line 32: How about:

"Number (`123`, `45.56`), boolean (`true`), and string (`'hi'`) literals, and named constants (`someConstant`, `math.pi`, `double.infinity`) are directly usable as constant patterns. Constant constructors (`const Point(0, 0)`) and constant collection literals (`const []`, `const {1, 2}`) can also be used directly as patterns. More complex constant expressions must be parenthesized and prefixed with `const` (`const (1 + 2)`)."

(Or maybe pull those examples out into code blocks.)

src/language/pattern-types.md
```diff
@@ -0,0 +1,333 @@
...
+
+```dart
+// List or map pattern:
+case [a, b]: ...
+
+// List or map literal:
+case const [a, b]: ...
+```
+
+## Identifier
```
munificent, line 42: This section is great. It does a very good job of explaining one of the more confusing parts of the feature.

src/language/pattern-types.md
```diff
@@ -0,0 +1,333 @@
...
+  ``` 
+- [Wildcard](#wildcard) identifier in any context: matches any value and discards it:
+  `case [_, var y, _]: print('The middle element is $y');`
+
+## List
+
+`[subpattern1, subpattern2]`
+
+List patterns match values that implement [`List`][], and then recursively match
+its subpatterns against the list's elements to destructure them by position.
```
munificent, line 69: "its" -> "their", or (probably better) make the whole sentence singular. "A list pattern matches... matches its subpatterns... ."

src/language/pattern-types.md
```diff
@@ -0,0 +1,333 @@
...
+List patterns can contain _one_ rest element (`...`) which allows matching lists
+of arbitrary lengths.
+
+```dart
+var [a, b, ..., c, d] = [1, 2, 3, 4, 5, 6, 7];
+print('$a $b $c $d'); // Prints "1 2 6 7".
+```
+
+Rest elements can also have subpatterns that collects elements that don't match
+the other subpatterns in the list, into a new list:
```
munificent, line 85: Given that you can only have a single rest element in a pattern, I think this would read better singular.

src/language/pattern-types.md
```diff
@@ -0,0 +1,333 @@
...
+left branch does not match, the right branch is not evaluated. 
+
+## Logical-or
+
+`subpattern1 || subpattern2`
+
+A pair of patterns separated by `||` matches if either of the branches match.
+If the left branch matches, the right branch is not evaluated. 
+
+Logical-or patterns can be nested inside a destructuring pattern.
```
munificent, line 106: Actually any pattern can be nested inside a destructuring pattern. You can go wild: `case Foo([a]! && {1: b}? || Bar() as Baz)`. So I'd probably not call this out here since it suggests, at least, that `&&` *can't* be nested.

src/language/pattern-types.md
munificent, line 148: "te" -> "the".

"can still match" is kind of vague. Maybe: "A map pattern ignores any keys that the map contains that aren't matched by the pattern."

munificent (approved):

src/language/patterns.md
mit-mit: I may be wrong, but I think of a constant pattern as an "actual value"

src/language/patterns.md
MaryaBelanger: This is MUCH better, will do!

src/language/patterns.md
```diff
@@ -0,0 +1,334 @@
...
+combining various contexts, pattern types, and surrounding code constructs.
+
+### Variable declaration
+
+Anywhere Dart allows regular, single-name local variable declaration, you can also
+use a _pattern variable declaration_. If the pattern matches the value on the right
+of the declaration, it destructures the value, binding it to new local variables:
+
+```dart
+var (a, [b, c]) = ("str", [1, 2]);   // Declares new variables a, b, and c.
```
johnpryan, line 88: nit:
```suggestion
// Declares new variables a, b, and c.
var (a, [b, c]) = ("str", [1, 2]);  
```

src/language/patterns.md
```diff
@@ -0,0 +1,334 @@
...
+matched object and assigns the values to _existing_ variables,
+instead of binding new ones. 
+
+One convenient use case is swapping the values of two variables without needing
+to declare a third temporary one:
+
+```dart
+var (a, b) = ('left', 'right');
+(b, a) = (a, b);                   // Swap
+print('$a $b');                    // Prints "right left".
```
johnpryan, line 108: nit: for maintainability, don't align in-line comments manually:
```suggestion
var (a, b) = ('left', 'right');
(b, a) = (a, b); // Swap
print('$a $b'); // Prints "right left".
```

src/language/patterns.md
```diff
@@ -0,0 +1,334 @@
...
+```dart
+main() {
+  var info = userInfo(json);
+  var name = info.$1;
+  var age = info.$2;
+}
+```
+You can call the function directly into a [variable declaration](#variable-declaration)
+or [assigment pattern](#variable-assignment),
+with a [record subpattern][record] and fields for each value in the return:
```
johnpryan, line 129: The phrase "call a function into" doesn't parse for me. Maybe "You can destructure the fields of the record into local variables using a variable declaration pattern and a record pattern as its subpattern"

src/language/patterns.md
```diff
@@ -0,0 +1,334 @@
...
+  }
+  ```  
+
+Values destructured by a pattern in a case become local variables,
+only available in the body of that case. 
+
+### Object destructuring
+
+[Object patterns][object] match against named object types, allowing
+you to destructure their data using the getters the object's class already exposes.
```
johnpryan, line 191: nit: My understanding is an object's properties (fields or getters) can be used:
```suggestion
you to destructure their data using the object's properties.
```

src/language/pattern-types.md
```diff
@@ -0,0 +1,333 @@
...
+pattern against that same value.
+
+
+```dart
+String? maybeString = // ...
+switch (maybeString) {
+  case var s?:
+    // s has type non-nullable String here.
+}
+```
```
johnpryan, line 181: It might be worth pointing out that you use the constant pattern `null` to match when the value _is_ null. Someone asked me about this recently.

src/language/pattern-types.md
```diff
@@ -0,0 +1,333 @@
...
+
+// Null-check and null-assert:
+switch (obj) {
+  case (checked: var checked?, asserted: var asserted!): // ...
+  case (:var checked?, :var asserted!): // ...
+}
+
+// Cast:
+var (field: field as int) = // ...
+var (:field as int) = // ...
```
johnpryan, line 238: Aren't null-check, null-assert, and cast patterns already covered elsewhere?

src/language/pattern-types.md
MaryaBelanger: I can add the line "Any [kind of pattern][] can be nested inside a destructuring pattern" to the Destructuring section on the main page.

src/language/pattern-types.md
```diff
@@ -0,0 +1,333 @@
...
+
+```dart
+// Variable:
+var (untyped: untyped, typed: int typed) = // ...
+var (:untyped, :int typed) = // ...
+
+switch (obj) {
+  case (untyped: var untyped, typed: int typed): // ...
+  case (:var untyped, :int typed): // ...
+}
```
johnpryan, line 228: I would move the typed / untyped stuff to the variable patterns section, since it's not specific to how record patterns work.

src/language/pattern-types.md
MaryaBelanger: Added:

> Subpatterns in a logical-or pattern can bind variables, but the two branches must
> define the same set of variables, because only one branch will be evaluated when
> the pattern matches.

src/language/pattern-types.md
MaryaBelanger: Thanks, I was getting confused by that myself

src/language/pattern-types.md
MaryaBelanger: Nice, I actually had that as part of the definition originally but took it out because I wasn't sure if it was useful or not. Thanks for confirming!

src/language/pattern-types.md
```diff
@@ -0,0 +1,333 @@
...
+title: Pattern types
+description: Pattern type reference in Dart.
+---
+
+## Cast
+
+`foo as String`
+
+Cast patterns let you insert a [type cast][] as a pattern is
+[destructuring][destructure] values:
```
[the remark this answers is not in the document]
munificent: I think the original prose is deliberate, but maybe it would read better as:

> A cast pattern lets you insert a [type cast][] in the middle of destructuring before passing the value to another subpattern.

?

munificent (approved):

src/language/pattern-types.md
leafpetersen: Yep, that's great.

src/language/pattern-types.md
leafpetersen: > that any pattern in that context is irrefutable and that an exception is thrown when patterns don't match in any irrefutable context, for any type of pattern, if the value doesn't match match.

Sort of?  In an irrefutable context, mostly the value can't fail to match.  That is, mostly if you see `var <pattern> = expression`, you'll get a static error if the static type of expression doesn't guarantee that the value will match pattern.  There are two bug exceptions: if the type of `expression` is `dynamic`, in which case the match might fail, and if `<pattern>` is a list pattern, in which case it can fail if the length isn't correct.  So there's something fundamentally different between the pattern `(a, b)` and `[a, b]`.  In a refutable context, they either match or don't, but in an irrefutable context, the former will always match, but the latter may throw an exception if the destructured value doesn't have exactly two elements.

src/language/patterns.md
MaryaBelanger: Ok, good point. I placed this section here (originally from the switch case content) before writing the patterns reference page, but now that that exists this can definitely go

src/language/pattern-types.md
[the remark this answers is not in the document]
MaryaBelanger: Oh, yes the original meant "Cast patterns let you insert a type cast _while_ a pattern is destructuring some values", but I'll change it to Bob's suggestion

src/language/pattern-types.md
```diff
@@ -0,0 +1,333 @@
+---
+title: Pattern types
+description: Pattern type reference in Dart.
+---
+
+## Cast
+
+`foo as String`
```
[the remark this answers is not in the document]
MaryaBelanger: So, I did that originally but a full code block just has _so much_ extra margin/padding space, I hated the way it looked, and I felt like it distracted from the actual functioning example blocks within each section.

This:
image

vs. this:
image

I think the latter looks so much better! Idk, is it really that bad?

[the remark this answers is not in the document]
parlough: I do agree in the short term the inline code looks better.

While not a priority to land in this PR, we can always introduce a custom block or style for this type of heading later on :)

src/language/patterns.md
MaryaBelanger: Changed it to:

> A single [case pattern][] (conducive to an [if-case][] statement)
> achieves the same validation as above, but is much clearer, more declarative,
> and much shorter:
> 
> ```dart
> if (json case {'user': [String name, int age]}) 
>   print('User $name is $age years old.');
> ```

The parenthetical about if-case, keep or toss?

src/language/pattern-types.md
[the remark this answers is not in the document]
MaryaBelanger: Ok, resolving for now

src/language/pattern-types.md
MaryaBelanger: Yeah I just copied this block straight from the spec without putting much thought into it, but it looks like it's showing records + some other subpattern examples. I'll try to sort the extra ones out to where they belong, though I do think a few outer+inner pattern examples are alright

src/language/patterns.md
MaryaBelanger: I compared your two new structuring ideas ([the other one](https://github.com/dart-lang/site-www/pull/4810/files#r1183153697)) and I think this is the way to go.

src/language/patterns.md
leafpetersen: The `(int?, int?)` syntax there is a record type not a pattern.  The `p?` syntax meaning "nullable type" as a type and "check for null" is unfortunately very confusing - we just never found anything better.

src/language/patterns.md
johnpryan: For inline comments, the convention I usually see is to add one space before `//`.

src/language/patterns.md
leafpetersen: I think I somewhat misunderstood the point of this and the previous section, which perhaps partially reflects me being too close to the feature, and maybe reflects me reading to quickly, but might also suggest that other people might also go astray?  I think what you are trying to say with these sections is roughly "Hey, in general, there are two things going on in patterns.  First, given a value, you might want to check whether the value has a certain shape, or is a certain constant, or is equal to something else, or has a certain type.  And second, given a value, you might want a convenient declarative syntax to break it into a bunch of its constituent parts, possibly binding variables to some or all of those parts in the process.  The first we call matching, and the second we call destructuring.  Patterns let us to both at once. Depending on the context and the shape of the pattern, a pattern may match a value, destructure a value, or both.  Here are some examples to show you what we mean by matching, and here are some to show you what we mean by destructuring".  Is that a fair (rough summary)?  If so, I think these sections are actually pretty close, I just think that maybe a little bit of introductory text like the above to give the reader a roadmap of what's going on might avoid sending people like me down the wrong road (I was kind of reading this as talking more about refutable vs irrefutable contexts, whereas I think you mean something more like what I describe above).  Does that make sense?

src/language/patterns.md
MaryaBelanger: Oh, ok. I wonder if this warrants explaining somewhere, though I'm not sure where that explanation would go.

src/language/patterns.md
MaryaBelanger: Yes that's it exactly. I added a new section header to summarize this, and then made the matching and destructuring sections subsections under that introduction

src/language/patterns.md
```diff
@@ -0,0 +1,358 @@
...
+
+### Switch statements and expressions
+
+Every case clause contains a pattern. This applies to [switch statements][switch]
+and [expressions][], as well as [if-case statements][if].
+
+_Case patterns_ are [refutable][];
+they allow control flow to either match and destructure the object being switched
+on, or continue execution if the object doesn't match. Case patterns can be
+[any kind of pattern][types]:
```
atsansone, line 143: **issue:** If you can use a semi-colon, you can use a period.
```suggestion
_Case patterns_ are [refutable][].
They control flow toward one of two outcomes:
* Match and destructure the switched object.
* Continue execution if the object doesn't match.
Case patterns can be [any kind of pattern][types].
```

**nit:** I would change the first sentence to active voice, but that would involve changing the link.

src/language/patterns.md
```diff
@@ -0,0 +1,358 @@
...
+  case (var a, var b): ...
+    // Matches if obj is a record with two fields, then assigns the fields to 'a' and 'b'.
+  case Rect(width: var w, height: var h): ...
+    // Matches if obj is an instance of Rect, and then against the properties of Rect.
+  default: ...
+}
+```
+
+
+Values destructured by a pattern in a case become local variables,
```
atsansone, line 160: **issue:** Write in active voice per [GSG](https://developers.google.com/style/voice).
```suggestion
Values that a pattern destructures in a case become local variables.
```

src/language/patterns.md
```diff
@@ -0,0 +1,358 @@
...
+only available in the body of that case.
```
atsansone, line 161: **issue:** Rephrase.
```suggestion
This limits the scope of these local variables to the body of that `case` statement. 
```

src/language/patterns.md
```diff
@@ -0,0 +1,358 @@
...
+  print('one $one, two $two');
+}
+```
+
+### Algebraic datatypes 
+
+Object destructuring and switch cases are conducive to writing
+code in an [algebraic datatype][] style: 
+when you have a family of related types and an operation that needs
+specific behavior for each type.
```
atsansone, line 256: **issue:** Write as shorter sentences per [GSG](https://developers.google.com/style/accessibility#ease-of-reading). This is a 33 word sentence.
```suggestion
Object destructuring and switch cases aid in writing
code in an [algebraic datatype][] style.
Use these methods when you have:
* a family of related types
* an operation that needs specific behavior for each type
```

src/language/patterns.md
```diff
@@ -0,0 +1,358 @@
...
+
+Instead of implementing the operation as an instance method for every respective type,
+you can keep the operations in a single function, switching over subtypes for
+variations in behavior:
```
atsansone, line 260: **issue:** Write as shorter sentences per [GSG](https://developers.google.com/style/accessibility#ease-of-reading). One 29 word sentence to three ~10 word sentences.
```suggestion
This provides an alternative to each type having its own instance method.
Keep the operations, or instance methods, in one function.
Use a switch statement to to through subtypes for behavior variations.
```

src/language/patterns.md
```diff
@@ -0,0 +1,358 @@
...
+[Map][] and [list][] patterns work well for destructuring key-value pairs in
+JSON data:
+
+```dart 
+var json = {'user': ['Lily', 13]};
+var {'user': [name, age]} = json;
+```
+
+The example above makes sense when you already know that the JSON
+data has the structure you expect.
```
atsansone, line 293: **nits**: Don't use physical direction indicators. Also, if you know something, the already would be implied. 
```suggestion
When you know that the JSON data has the structure you expect,
the previous example makes sense.
```

src/language/patterns.md
```diff
@@ -0,0 +1,358 @@
...
+If the data comes from an external source, like over the network as is commonly
+the case, you can't be sure of it's structure and
+should validate it first.
```
atsansone, line 296: **issues:** Write as shorter sentences per [GSG](https://developers.google.com/style/accessibility#ease-of-reading).
**typo:** `its` not `it's`.
```suggestion
If the data comes from an external source, you can't validate its structure.
You should validate it first. 
```

src/language/patterns.md
```diff
@@ -0,0 +1,358 @@
...
+    var age = user[1] as int;
+    print('User $name is $age years old.');
+  }
+}
+```
+
+A single [case pattern](#switch-statements-and-expressions)
+(conducive to an [if-case][if] statement)
+achieves the same validation as above, but is much clearer, more declarative,
+and much shorter:
```
atsansone, line 316: **issue:** Write as shorter sentences per [GSG](https://developers.google.com/style/accessibility#ease-of-reading).
```suggestion
One [case pattern](#switch-statements-and-expressions) validates
in the same way as the previous example did.
It does so with more clarity, more declaration, and less size.
```

src/language/patterns.md
```diff
@@ -0,0 +1,358 @@
...
+
+```dart
+for (var MapEntry(:key, value: count) in hist.entries) { 
+  print('$key occurred $count times');
+}
+```
+
+## Uses cases for patterns
+
+The [section above](#places-patterns-can-appear)
```
atsansone, line 197: **issue:** Don't use physical directionality in documentation per [GSG](https://developers.google.com/style/word-list#above).

src/language/patterns.md
```diff
@@ -0,0 +1,358 @@
...
+describes _how_ you can use patterns in Dart code constructs, 
+sometimes with interesting uses cases as examples (like [swapping](#variable-assignment)
+the values of two variables, or
+[destructuring key-value pairs](#for-and-for-in-loops)
+in a map).
```
atsansone, line 202: **issue:** Write as shorter sentences per [GSG](https://developers.google.com/style/accessibility#ease-of-reading).
```suggestion
describes _how_ you can use patterns in Dart code constructs. 
Dart code constructs have interesting uses cases as examples.
These include [swapping](#variable-assignment)
the values of two variables
and [destructuring key-value pairs](#for-and-for-in-loops)
in a map. 
```

src/language/patterns.md
```diff
@@ -0,0 +1,358 @@
...
+The [section above](#places-patterns-can-appear)
+describes _how_ you can use patterns in Dart code constructs, 
+sometimes with interesting uses cases as examples (like [swapping](#variable-assignment)
+the values of two variables, or
+[destructuring key-value pairs](#for-and-for-in-loops)
+in a map). 
+
+This section describes even more use cases, answering _when and why_ you might
+want to use patterns, the kinds of problems they solve, and the idioms they're 
+best suited for.
```
atsansone, line 206: **issue:** 
Write as shorter sentences per [GSG](https://developers.google.com/style/accessibility#ease-of-reading).
```suggestion
This section describes even more use cases that answer:
* _when and why_ you might want to use patterns.
* the kinds of problems they solve.
* the idioms for which they're best suited.
```

src/language/patterns.md
```diff
@@ -0,0 +1,358 @@
...
+and much shorter:
+
+```dart
+if (json case {'user': [String name, int age]}) 
+  print('User $name is $age years old.');
+```
+
+This case pattern simultaneously validates that:
+
+- `json` is a map (because it must first match the outer [map pattern][map] to proceed).
```
atsansone, line 325: **issue:** Remove the parenthetical.
```suggestion
- `json` is a map because it must first match the outer [map pattern][map] to proceed.
```

atsansone (changes requested):
@MaryaBelanger : This looks worse than it is. Most of these comments are that the sentences are too long. I offered suggestions for each.

src/language/patterns.md
munificent: I think of it as a constant pattern *has* a value for the constant that it tests, just like a relational pattern has a value for the right-hand side operand. But I don't think of patterns as being value-like in general.

The "set of values" and "actual values" feels like a confusing level of repetition. The only "set of values" the pattern represents are the actual ones it would match.

If you look up "pattern" in the dictionary, the two definitions that are closest are:

> * an arrangement or sequence regularly found in comparable objects. "the house had been built on the usual pattern"

> * a regular and intelligible form or sequence discernible in certain actions or situations. "a complicating factor is the change in working patterns"

I think, "regular", "comparable", and "discernable" are the key bits.

A pattern describes some commonality that a set of objects share and you can determine whether a given object has that commonality or not. A pattern is like a mold, gauge, or yardstick that you can hold an object up to and see if it fits.

So maybe, "A pattern describes a set of properties an object might have. A value can be *matched* against the pattern to see if it has those properties."

src/language/patterns.md
```diff
@@ -0,0 +1,358 @@
...
+{{site.alert.end}}
+
+Patterns are a syntactic category in the Dart language, like statements and expressions.
+A pattern represents the shape of a set of values that it may match against actual
+values.
+
+This page describes:
+- What patterns do
+- Where patterns are allowed in Dart code
+- Common use cases for patterns
```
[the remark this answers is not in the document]
MaryaBelanger: Can I say "What the common use cases for patterns are"? Or is that grammatically worse? "What are..." sounds weird to me but I do want to think of something consistent for this point

src/language/patterns.md
```diff
@@ -0,0 +1,358 @@
...
+a certain type.
+
+Then, destructuring provides you with a convenient declarative syntax to break
+that value into its constituent parts, possibly binding variables to some or all
+of those parts in the process.
+
+### Matching
+
+A pattern is always tested against a value to determine if the value has the form
+you expect, or in other words, if the value _matches_ the pattern.
```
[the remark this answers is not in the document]
MaryaBelanger: The Dart team has been pretty clear against using "Dart" as an actor for anything in the docs, and I agree. I'll take the rest of the notes and say "Patterns compare against..." or something like that.

src/language/patterns.md
MaryaBelanger: I liked the way this sentence flowed, so I cut some words out to get it down to 26 per GSG instead of restructuring it 😬 

> Instead of implementing the operation as an instance method for every type,
keep the operations in a single function, switching over subtypes for
behavior variations:

src/language/pattern-types.md
```diff
@@ -0,0 +1,406 @@
...
+
+```dart
+String? maybeString = // ...
+switch (maybeString) {
+  case var s?: ...
+    // 's' has type non-nullable String here.
+}
+```
+
+Use a null-check pattern while matching if you need to ensure `null` values
```
whesse, line 227: I don't understand this sentence.  Doesn't a null-check pattern do exactly that: silently treat a null value as a match failure?

Should this be "Use a null-assert pattern ..."?  Maybe the sentence above, about null-assert patterns, should be the one about null-check patterns, and they are accidentally switched.

src/language/patterns.md
MaryaBelanger: Dropped "statement" at the end because want it to be generalized for expressions too

src/language/patterns.md
MaryaBelanger: I'm not sure how the first sentence can be made active. Ignoring the link, what would you do?

src/language/patterns.md
```diff
@@ -0,0 +1,358 @@
...
+
+<a id="pattern-uses"></a>
+
+- Local variable [declarations](#variable-declaration) and [assignments](#variable-assignment)
+- [for and for-in loops][for]
+- [if-case][if] and [switch-case][switch]
+- Control flow in [collection literals][]
+
+This section describes common use cases for matching and destructuring with patterns, 
+combining various contexts, pattern types, and surrounding code constructs.
```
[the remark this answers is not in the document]
MaryaBelanger: Just got rid of the second clause, it doesn't really add anything

src/language/patterns.md
```diff
@@ -0,0 +1,358 @@
...
+```dart
+switch (obj) {
+  // List pattern [a,b] matches obj first if obj is a list wth two fields,
+  // then if its fields match the constant subpatterns 'a' and 'b'.
+  case [a, b]: // ...   
+}                    
+```
+
+Patterns can ignore parts of a matched value by using a [wildcard pattern][]
+as a place holder, or a [rest element][] in the case of list patterns.
```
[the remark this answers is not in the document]
MaryaBelanger: I'm going to use "you" as the actor here, instead of "patterns", as it is a personal choice. But incorporating the gist of it.

src/language/patterns.md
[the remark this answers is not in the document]
atsansone: **comment:** Oh, yeah. It reads like a question how I wrote it. Your revision would be better. You can edit my comment suggestion to your wording and accept _that_.

src/language/patterns.md
atsansone: **comment:** What refutes case patterns? That would explain it.

That said, I'm trying to soften my approach to passive voice. The **When do I use passive voice?** section of [this page](https://advice.writing.utoronto.ca/revising/passive-voice/) from the University of Toronto seems to be a good guide. Look at that to see if what you have here just makes more sense to keep as is.

MaryaBelanger: Oh, ok. It didn't even cross my mind that that kind of sentence structure (Case patterns are refutable) is passive voice, good to know!

Who or what is refuting them is really hard to say so I'm going to leave this for now. I think it might be the pattern that does the refuting, but also is the one being refuted...idk. I vaguely remember Bob saying "refutable/irrefutable" are _types_ of patterns, so the passive voice is built into the descriptors.

src/language/pattern-types.md
whesse: I do think the other sentence is about null checks - it is "silently eliminating null (matches) from a pattern", without throwing. I don't see how assets could ever be silent.

MaryaBelanger: Got it. The null check pattern is about purposefully treating null values as match failures, and null assert is about _not_ treating them silently. I kept the line in both sections, emphasizing the difference

src/language/pattern-types.md
```diff
@@ -194,20 +194,31 @@ ignores any keys that the map contains that aren't matched by the pattern.
 `subpattern!`
 
 Null-assert patterns match first if the object is not null, then on the value.
-They permit non-null values to flow through, but [throw][] if the matched value is null. 
+They permit non-null values to flow through, but [throw][] if the matched value
+is null. 
+
+Use a null-assert pattern while matching if you need to ensure `null` values
+are not silently treated as match failures:
```
atsansone, line 201: **issue:** Condition, then instruction per [GSG](https://developers.google.com/style/sentence-structure).

```suggestion
To ensure `null` values don't get treated as match failures,
use a null-assert pattern while matching:
```

src/language/pattern-types.md
```diff
@@ -194,20 +194,31 @@ ignores any keys that the map contains that aren't matched by the pattern.
...
 List<String?> row = // ...
 
-// If the first column is 'user', we expect to have a name after it.
 switch (row) {
   case ['user', var name!]: // ...
     // 'name' is a non-nullable string here.
 }
 ```
 
 Use the null-assert pattern to eliminate `null` values from variable declaration
```
atsansone, line 212: **issue:** Condition, then instruction per [GSG](https://developers.google.com/style/sentence-structure).
```suggestion
To eliminate `null` values from variable declaration, use the null-assert pattern
```

src/language/pattern-types.md
```diff
@@ -216,16 +227,19 @@ patterns.
 Null-check patterns match first if the value is not null, and then match the inner
 pattern against that same value.
 
+Use the null-check pattern to silently treat `null` values as match failures
+without throwing.
```
atsansone, line 231: **issue:** Condition, then instruction per [GSG](https://developers.google.com/style/sentence-structure).
```suggestion
To treat `null` values as match failures
without throwing errors, use the null-check pattern.
```

**note:** The `silently treat` and `without throwing` are redundant. I removed one.

src/language/patterns.md
```diff
@@ -136,11 +137,12 @@ print('$a $b'); // Prints "right left".
...
+You can use [any kind of pattern][types] in a case.
 
-_Case patterns_ are [refutable][];
-they allow control flow to either match and destructure the object being switched
-on, or continue execution if the object doesn't match. Case patterns can be
-[any kind of pattern][types]:
+_Case patterns_ are [refutable][].
+They allow control flow to either:
+- Match and destructure the object being switched on,
+- Or continue execution if the object doesn't match.
```
atsansone, line 145: **nit:** For parallelism in the list, change this slightly:
```suggestion
- Match and destructure the object being switched on, or
- Continue execution if the object doesn't match.
```

src/language/patterns.md
```diff
@@ -194,16 +196,16 @@ for (var MapEntry(:key, value: count) in hist.entries) {
...
 [destructuring key-value pairs](#for-and-for-in-loops)
-in a map). 
+in a map. This section describes even more use cases, answering:
 
-This section describes even more use cases, answering _when and why_ you might
-want to use patterns, the kinds of problems they solve, and the idioms they're 
-best suited for.
+- _When and why_ you might want to use patterns.
+- The kinds of problems they solve.
+- The idioms for which they're best suited.
```
atsansone, line 208: **suggestion:** Parallelism.
```suggestion
- _What_ kinds of problems they solve.
- _Which_ idioms they best suit.
```

src/language/patterns.md
```diff
@@ -289,11 +293,10 @@ var json = {'user': ['Lily', 13]};
 var {'user': [name, age]} = json;
 ```
 
-The example above makes sense when you already know that the JSON
-data has the structure you expect.
-If the data comes from an external source, like over the network as is commonly
-the case, you can't be sure of it's structure and
-should validate it first. 
+IF you know that the JSON data has the structure you expect,
```
atsansone, line 296: **typo:**
```suggestion
If you know that the JSON data has the structure you expect,
```

src/language/patterns.md
```diff
@@ -289,11 +293,10 @@ var json = {'user': ['Lily', 13]};
...
+the previous example is realistic.
+But data typically comes from an external source, like over the network.
```
atsansone, line 298: **nit:** Rephrase.
```suggestion
But the typical data source comes from an external origin, like over the network.
```

src/language/patterns.md
```diff
@@ -311,9 +314,10 @@ if (json is Map<String, dynamic> &&
 ```
 
 A single [case pattern](#switch-statements-and-expressions)
-(conducive to an [if-case][if] statement)
-achieves the same validation as above, but is much clearer, more declarative,
-and much shorter:
+can achieve the same validation.
+Single cases are conducive to [if-case][if] statements.
```
atsansone, line 318: **nit:** This reads with some complexity. Simplify the word choice.
```suggestion
Single cases work best as [if-case][if] statements.
```

atsansone (approved):
@MaryaBelanger : Only a few issues and some suggestions / nits. Back to you. No need for another review.

src/language/patterns.md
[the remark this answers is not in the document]
MaryaBelanger: I guess the "or" isn't really even necessary

src/language/pattern-types.md
```diff
@@ -216,16 +227,19 @@ patterns.
...
 switch (maybeString) {
   case var s?: ...
     // 's' has type non-nullable String here.
 }
 ```
 
-Use a null-check pattern while matching if you need to ensure `null` values
-aren't silently treated as match failures.
+The null-check pattern provides a terse way to bind a variable whose type is the
+non-nullable base type of the nullable value being matched.
```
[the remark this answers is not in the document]
MaryaBelanger: Shortened and moved to the top of the section

PR #4808, opened 2023-05-02
[Dart 3] Migration guide

mit-mit:
fyi @MaryaBelanger I started on the Dart 3 migration guide in this PR

staging link:
https://dart-dev--pr4808-dart3migrationguide-lhaqtkuf.web.app/resources/dart-3-migration

PR #4819, opened 2023-05-08
Change mentions of optional null safety

atsansone:


src/overview.md
```diff
@@ -42,16 +42,15 @@ allowing the use of a `dynamic` type combined with runtime checks,
...
 null exceptions at runtime through static code analysis.
 Unlike many other null-safe languages,
-when Dart determines that a variable is non-nullable,
-that variable is _always_ non-nullable.
+when Dart determines that a variable can't be null,
+that variable can never be null.
 If you inspect your running code in the debugger,
-you’ll see that non-nullability is retained at runtime
-(hence _sound_ null safety).
+you see that runtime retains this non-nullability; hence _sound_ null safety.
```
parlough, line 53: It doesn't make sense to say that "runtime retains". You could technically say "the runtime retains" but developers shouldn't have to worry about "the runtime".

```suggestion
you see that non-nullability is retained at runtime; hence _sound_ null safety.
```

src/resources/books.md
```diff
@@ -3,21 +3,21 @@ title: Books about Dart
...
+  If you find a Dart book not listed on this page,
+  check for publication dates after June 2021.
+  Older books lack coverage of Dart 2 and 3 topics such as
   strong typing, null safety, FFI, the `dart` command-line utility,
-  and new developer tools. 
+  and new developer tools.
 {{site.alert.end}}
 
-## Dart books with null safety
+## Dart books
```
parlough, line 20: Honestly you can drop this title now. It was only added for the differentiation between null-safe books and non-null safe ones. The whole page is about Dart books.

```suggestion
```

src/tools/dartpad/troubleshoot.md
```diff
@@ -94,16 +94,15 @@ not **Solution** or **Tests**.
 
 If the code is using language or library features no longer supported
 in the latest stable release of Dart,
-the code may no longer analyze or run properly.
+the code may no longer analyze or run.
```
parlough, line 97: I think the word "properly" or something like it is important here. It will still analyze, just not without errors.

```suggestion
the code may no longer analyze or run without errors.
```

src/null-safety/index.md
```diff
@@ -5,37 +5,36 @@ description: Information about Dart's null safety feature
 
 The Dart language comes with sound null safety. 
 
-In Dart 3, you always get null safety. In Dart 2.x,
```
parlough, line 8: This page is a historical document for the introduction of null safety, not documentation of the null-safe type system, consider keeping this, especially since it's just after Dart 3 released.

src/null-safety/index.md
```diff
@@ -5,37 +5,36 @@ description: Information about Dart's null safety feature
...
-your app causes a runtime error. This type of error, a null dereference error,
-can be difficult to debug.
-
-With sound null safety variables are 'non-nullable' by default:
-They can be assigned only values of the declared type
-(e.g. `int i=42`), and never be assigned `null`.
-You can specify that a type of a variable is nullable
-(e.g. `int? i`),
-and only then can they contain either a `null` *or*
+your app returns a runtime error.
```
parlough, line 12: Errors are not returned.

```suggestion
your app causes a runtime error.
```

src/null-safety/index.md
```diff
@@ -5,37 +5,36 @@ description: Information about Dart's null safety feature
...
+This type of error, a null dereference error, can be difficult to debug.
+
+With sound null safety, variables default to 'non-nullable'.
+You can assigned values of the declared type only, like `int i=42`.
```
parlough, line 16:
```suggestion
You can assign values of the declared type only, like `int i=42`.
```

src/null-safety/index.md
```diff
@@ -5,37 +5,36 @@ description: Information about Dart's null safety feature
...
+You can never assign the value to `null` to default variable types.
```
parlough, line 17: I'm not 100% sure what this sentence is trying to say. Perhaps:

```suggestion
You can never assign the value `null` to non-nullable variables.
```

src/null-safety/index.md
```diff
@@ -5,37 +5,36 @@ description: Information about Dart's null safety feature
...
+To specify that a variable type can be nullable, you add a `?` after
+the type label: `int? i`.
+These specific types can contain either a `null` *or*
```
parlough, line 20:
```suggestion
To specify that a variable type can be nullable, 
add a `?` after the type label: `int? i`.
These nullable types can contain either `null` *or*
```

src/null-safety/index.md
```diff
@@ -5,37 +5,36 @@ description: Information about Dart's null safety feature
...
 
 ## Introduction through examples
 
-With null safety,
-all of the variables in the following code are non-nullable:
+With null safety, all variables in the following code can't be `null`:
 
 ```dart
-// In null-safe Dart, none of these can ever be null.
+// In Dart 3, none of these can ever be null.
```
parlough, line 37: This is a historical document and it also applies to null-safe code before Dart 3.

```suggestion
// With null safety, none of these can ever be null.
```

src/null-safety/index.md
```diff
@@ -5,37 +5,36 @@ description: Information about Dart's null safety feature
...
+These specific types can contain either a `null` *or*
 a value of the defined type.
 
 Sound null safety changes potential **runtime errors**
-into **edit-time** analysis errors, by flagging when
-any non-nullable variable hasn't been initialized with a 
-non-null value or is being assigned a `null`.
-This allows you to fix these errors before deploying your app.
+into **edit-time** analysis errors.
+Null safety does flagging that a non-null variable has either:
```
parlough, line 25: It reads weird to me to say that "null safety does flagging". Perhaps:

```suggestion
With null safety, the Dart analyzer and compilers
flag if a non-nullable variable has either:
```

src/null-safety/index.md
```diff
@@ -5,37 +5,36 @@ description: Information about Dart's null safety feature
...
+
+* Not been initialized with a non-null value
+* Been assigned a `null` value.
 
+This check allows you to fix these errors _before_ deploying your app.
```
parlough, line 30:
```suggestion
These checks allows you to fix these errors _before_ deploying your app.
```

src/null-safety/index.md
```diff
@@ -91,8 +90,7 @@ Because pkg1 doesn't support null safety, version solving failed.
 The lower bound of "sdk: '>=2.9.0 <3.0.0'" must be 2.12.0 or higher to enable null safety.
 ```
 
-Libraries that opt out of null safety will cause analysis or compilation
-errors:
+Libraries incompatible with Dart 3 return analysis or compilation errors.
```
parlough, line 93: This changes the meaning of the sentence. It's clear that libraries incompatible will cause errors, that's what incompatible means. The important part is the mention of the lack of null safety making them incompatible.

```suggestion
Pre-Dart 3 libraries that opt out of null safety will 
cause analysis or compilation errors.
```

src/null-safety/index.md
```diff
@@ -176,24 +176,9 @@ environment:
 
 [language version]: /guides/language/evolution#language-versioning
 
-## Migrating existing code {#migrate}
```
parlough, line 179: Since this is a historical document, I don't think you should remove this section. The page is about migrating, which still exists in this historical context.

src/null-safety/index.md
```diff
@@ -176,24 +176,9 @@ environment:
...
-```
-
-For more details on how to migrate your code to null safety,
-see the [migration guide][].
-
-
 ## Where to learn more
 
-For more information about null safety, see the following resources:
+To learn more about null safety, see the following resources:
```
parlough, line 181:
```suggestion
To learn more about null safety, check out the following resources:
```

src/null-safety/index.md
```diff
@@ -160,8 +158,10 @@ doesn't depend on Dart 3 breaking changes:
 
 ## Dart 2.x and null safety {#enable-null-safety}
 
-In Dart 2.12 to 2.19, null safety is a configuration option in the pubspec. Null
-safety is not available in SDK versions prior to Dart 2.12.
+From Dart 2.12 to 2.19, you need to enable null safety.
+To enable null safety, set the `sdk` option to `'>=2.12.0'`
+in the `pubspec.yaml` file.
```
parlough, line 163: The next sentences cover this in greater, more accurate detail. Can just remove this.

```suggestion
```

src/null-safety/index.md
```diff
@@ -115,9 +113,9 @@ To resolve these issues, check for
 of any packages you installed from pub.dev, and [migrate](#migrate) all
 of your own source code to use sound null safety.
 
-As of late April 2023, **Dart 3 beta** is available from the Dart beta channel
-and the Flutter beta channel; see [the download page][] for details.
-We recommend you test your code for Dart 3 compatibility using that release:
+**Dart 3** has been released to the Dart and Flutter stable channels.
```
parlough, line 116: nit: I'm not sure it makes sense to say "Dart 3" is released to the Flutter stable channel, since it's not released separately from Flutter. Perhaps "available" is still more accurate, at least in terms of Flutter.

src/null-safety/index.md
```diff
@@ -135,22 +133,22 @@ listed by the analyzer.
...
 ### Dart 3 backwards compatibility
 
-Packages and apps migrated to use null safety with Dart 2.12 or later will
-likely be backwards compatible with Dart 3. Specifically, for any package where
-the lower bound of the SDK constraint is 2.12.0 or higher, pub will allow
-resolution even when the upper bound is limited to versions below 3.0.0. For
-example, a package with the following constraint will be allowed to resolve with
-a Dart 3.x SDK:
+Packages and apps migrated to use null safety with Dart 2.12 or later
+should be backwards compatible with Dart 3.
```
parlough, line 137: question: Since the package was migrated with an older version, shouldn't this technically read "forwards compatible"?

```suggestion
should be forwards compatible with Dart 3.
```

src/null-safety/index.md
```diff
@@ -135,22 +133,22 @@ listed by the analyzer.
...
+If a any package has a lower bound SDK constraint of 2.12.0 or later,
+`dart pub` resolves that package.
+This remains true when the upper bound falls to a version earlier than 3.0.0.
```
parlough, line 140: I think "falls" is hard to understand here:

```suggestion
This remains true when the upper bound falls is
set to a version earlier than 3.0.0.
```

src/null-safety/index.md
```diff
@@ -135,22 +133,22 @@ listed by the analyzer.
...
-Packages and apps migrated to use null safety with Dart 2.12 or later will
-likely be backwards compatible with Dart 3. Specifically, for any package where
-the lower bound of the SDK constraint is 2.12.0 or higher, pub will allow
-resolution even when the upper bound is limited to versions below 3.0.0. For
-example, a package with the following constraint will be allowed to resolve with
-a Dart 3.x SDK:
+Packages and apps migrated to use null safety with Dart 2.12 or later
+should be backwards compatible with Dart 3.
+If a any package has a lower bound SDK constraint of 2.12.0 or later,
+`dart pub` resolves that package.
```
parlough, line 139: Resolution is a larger process than one individual package, and involves interactions with other packages, so I think the previous text is more accurate. Perhaps something like:

```suggestion
If a package has a lower bound SDK constraint of 2.12.0 or later,
`dart pub` will allow package resolution.
```

src/null-safety/index.md
```diff
@@ -135,22 +133,22 @@ listed by the analyzer.
...
 environment:
   sdk: '>=2.14.0 <3.0.0'
 ```
 
-This allows developers to use Dart 3 sound null safety with packages that have
-been migrated to 2.12 null safety without needing a second migration.
-Note that this only applies to code that
-doesn't depend on Dart 3 breaking changes:
+This allows developers to use Dart 3 sound null safety with packages
+that use 2.12 or later null safety without a second migration.
```
parlough, line 149: "2.12 or later null safety" doesn't make sense to me. How about the following?

```suggestion
This allows developers to use Dart 3 sound null safety with 
Dart 2 packages that have been migrated to null safety 
without a second migration.
```

src/null-safety/index.md
```diff
@@ -135,22 +133,22 @@ listed by the analyzer.
...
+
+Note: This doesn't apply to code that doesn't depend on Dart 3 breaking changes.
```
parlough, line 151: I think it's the other way around. It doesn't apply to code that _does_ depend on breaking changes in Dart 3.

```suggestion
Note: This doesn't apply to code that is affected by
other breaking changes in Dart 3.
```

parlough (changes requested):
Thanks for working on this and some of the miscellaneous cleanup. I noticed some issues and left a few other minor suggestions:

parlough (changes requested):
Thanks for working on this and some of the miscellaneous cleanup. I noticed some issues and left a few other minor suggestions:

src/_guides/language/coming-from/swift-to-dart.md
```diff
@@ -541,16 +539,15 @@ let str: String? = nil
...
-which ignores any operations when
-the target object is `null`.
+In addition, Dart offers a
+[null safe version of the cascade operator][] (`?..`).
+This operator ignores any operations when
+the target expression resolves to `null`.
 Dart also offers the null assignment operator (`??=`),
 which Swift doesn't.
-It allows you to only assign a value to a variable
+This operator permits you to assign a value to a variable
```
mit-mit, line 548: This changed meaning a bit. The point of this operator is that the assignment is conditional to to when the current value is null:

```
// Assign value to b if b is null; otherwise, b stays the same
b ??= value;
```

src/_guides/language/coming-from/swift-to-dart.md
```diff
@@ -790,14 +786,16 @@ func multiply(_ a: Int, _ b: Int, c: Int = 1, d: Int = 1) -> Int {
 }
 ```
 
-Named parameters must have a default value,
-be marked as nullable (which defaults to a `null` value,
-if not explicitly provided), or be marked as `required`
-for the function to compile.
-Nullable types and [null safety](#null-safety)
-are covered in a later section.
+For the function to compile in Dart, named parameters must include
```
mit-mit, line 789: The "to compile" part is confusing. Any violations of this are flagged during analysis, not compilation.

I'd just say "Named parameters must include...:"

src/codelabs/null-safety.md
```diff
@@ -128,38 +132,39 @@ void main() {
...
 // The following calls the 'action' method only if nullableObject is not null
 nullableObject?.action();
 ```
 
-In the code below, try using conditional property access
-in the `stringLength` method to fix the error and
-to return the length of the string or `null` if it is `null`:
+In the following code, try using conditional property access
+in the `stringLength` method. This fixes the error and
+to return the length of the string or `null` if it equals `null`:
```
mit-mit, line 167: Change "the error and to return the length" to "the error and returns the length" ?

src/codelabs/null-safety.md
```diff
@@ -348,23 +352,24 @@ as providing an alternative value if the left-hand side is `null`.
 
 ## Type promotion
 
-With sound null safety, Dart's
-[flow analysis](/null-safety/understanding-null-safety#flow-analysis) has
-been extended to take nullability into account. Nullable variables that
-can't possibly contain null values are treated like non-nullable variables.
-This behavior is called
+With sound null safety, we extended Dart's
```
mit-mit, line 355: I'd drop the history part of this (i.e. change "we extended Dart's flow analysis to account for..." to "Dart's flow analysis accounts for...")

src/codelabs/null-safety.md
```diff
@@ -427,19 +432,20 @@ For cases like that, use the
 [`late` keyword](/null-safety/understanding-null-safety#late-variables).
 
 When you put `late` in front of a variable declaration,
-that tells Dart the following:
+that tells Dart the following about the variable:
 
--   Don't assign that variable a value yet.
--   You *will* assign it a value later.
--   You'll make sure that the variable has a value *before* the variable is used.
+- It cannot be assigned a value yet.
```
mit-mit, line 437: This is too strong. I might be that it could be assigned a value, but that the code for that would just be too cumbersome.

`late` is the developer expressing *their* intent: "I don't want to assign assign value just yet"

src/language/variables.md
```diff
@@ -58,7 +58,7 @@ assert(lineCount == null);
   _condition_ is false. For details, check out [Assert][].
 {{site.alert.end}}
 
-If you enable null safety, then you must initialize the values
+With null safety, you must initialize the values
```
mit-mit, line 61: I think we need to tweak this page a bit more. This is, I think, the perfect early point to introduce null safety at is most basic level. I think we need to pull a bit of intro content into this section about variables.

src/null-safety/index.md
mit-mit: I think the intro part of this page should be brought up into the variables page above.

src/null-safety/index.md
```diff
@@ -78,11 +78,11 @@ Dart null safety support is based on the following three core design principles:
 
 ## Dart 3 and null safety
 
-Dart 3---planned for a mid-2023 release---always has sound null safety.
-Dart 3 will prevent code from running without it.
+Dart 3 always has sound null safety.
```
mit-mit, line 81: The term we've been using in the other PRs is ""Dart 3 has null safety built-in"

src/null-safety/index.md
mit-mit: +1 for "Dart 3 is available in the Dart and Flutter stable channels"

src/null-safety/index.md
```diff
@@ -135,22 +135,25 @@ listed by the analyzer.
 
 ### Dart 3 backwards compatibility
```
mit-mit, line 136: This whole section is being removed in https://github.com/dart-lang/site-www/pull/4808/files#diff-74cb202dbd963f8abf2a2a5cee0226b6396b1852141f6fa57422fbd8b89306a0, so to avoid conflicts please leave it as-is.

src/null-safety/index.md
mit-mit: +1

atsansone: **comment:** It's not a historical document, but this should point to the migration guide instead of explaining anything here.

mit-mit:
The build seems to be failing, but the logs confuse me as there is no detail. Perhaps @parlough knows what's going on?

src/codelabs/dart-cheatsheet.md
```diff
@@ -88,27 +88,25 @@ reference them inside single quotes, with a space in between.
...
-which is **invalid** because (with null safety)
-a variable of type `int` can't have the value `null`:
+For example, consider the following code.
+With null safety, this code returns an error.
+A variable of type `int` can't have the value `null`:
 
 <?code-excerpt "misc/bin/cheatsheet/nullable.dart (invalid-null)" replace="/null;/[!null!];/g"?>
 ```dart
-int a = [!null!]; // INVALID in null-safe Dart.
+int a = [!null!]; // INVALID
```
parlough, line 101: This is a code excerpt so you will need it update it in its original file: https://github.com/dart-lang/site-www/blob/main/examples/misc/bin/cheatsheet/nullable.dart#L7

src/codelabs/dart-cheatsheet.md
```diff
@@ -88,27 +88,25 @@ reference them inside single quotes, with a space in between.
...
 
-When creating a variable in Dart 2.12 or higher,
-you can add `?` to the type to indicate
+When creating a variable, add `?` to the type to indicate
 that the variable can be `null`:
 
 <?code-excerpt "misc/bin/cheatsheet/nullable.dart (valid-null)" replace="/int\?/[!int?!]/g"?>
 ```dart
-[!int?!] a = null; // Valid in null-safe Dart.
+[!int?!] a = null; // Valid
```
parlough, line 109: This is a code excerpt so you will need it update it in its original file: https://github.com/dart-lang/site-www/blob/main/examples/misc/bin/cheatsheet/nullable.dart#L13

src/codelabs/null-safety.md
```diff
@@ -348,23 +352,23 @@ as providing an alternative value if the left-hand side is `null`.
 
 ## Type promotion
 
-With sound null safety, Dart's
-[flow analysis](/null-safety/understanding-null-safety#flow-analysis) has
-been extended to take nullability into account. Nullable variables that
-can't possibly contain null values are treated like non-nullable variables.
-This behavior is called
+Dart's [flow analysis](/null-safety/understanding-null-safety#flow-analysis)
+to account for nullability.
```
parlough, line 356: The grammar got messed up here.

```suggestion
Dart's [flow analysis](/null-safety/understanding-null-safety#flow-analysis)
accounts for nullability.
```

src/codelabs/null-safety.md
```diff
@@ -384,8 +388,8 @@ void main() {
 
 ### Exercise: Null checking
 
-In the code below, add an `if` statement to the beginning of `getLength` that
-returns zero if `str` is null:
+In the following code, add an `if` statement to the beginning of
+`getLength` that returns zero if `str` has no value:
```
parlough, line 392: I feel "is null" is better than "no value" because "no value" could also be taken to mean it hasn't been assigned a value yet.

```suggestion
`getLength` that returns zero if `str` is `null`:
```

src/language/variables.md
```diff
@@ -38,7 +38,34 @@ String name = 'Bob';
...
 
+## Null safety
 
+The Dart language enforces sound null safety.
+
+Null safety prevents errors that result from unintentional access
+of variables set to `null`.
+
+For example, if a method expects an integer but receives `null`,
+your app returns a runtime error.
```
parlough, line 49:
```suggestion
your app causes a runtime error.
```

src/null-safety/index.md
```diff
@@ -3,39 +3,39 @@ title: Sound null safety
...
-your app causes a runtime error. This type of error, a null dereference error,
-can be difficult to debug.
-
-With sound null safety variables are 'non-nullable' by default:
-They can be assigned only values of the declared type
-(e.g. `int i=42`), and never be assigned `null`.
-You can specify that a type of a variable is nullable
-(e.g. `int? i`),
-and only then can they contain either a `null` *or*
+your app returns a runtime error.
```
parlough, line 12:
```suggestion
your app causes a runtime error.
```

parlough (commented):
I left a few more comments. 

The build is failing due to some code excerpts not being updated in the original /examples files. I pointed them out.

parlough (commented):
I left a few more comments. 

The build is failing due to some code excerpts not being updated in the original /examples files. I pointed them out.

atsansone:
@parlough : Thanks for the assist!

src/language/variables.md
```diff
@@ -38,7 +38,34 @@ String name = 'Bob';
...
+a value of the defined type.
+
+Sound null safety changes potential **runtime errors**
+into **edit-time** analysis errors.
+Null safety flags a non-null variable when it has been either:
+
+* Not initialized with a non-null value.
+* Assigned a `null` value.
+
+This check allows you to fix these errors _before_ deploying your app.
```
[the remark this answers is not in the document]
atsansone: **comment:** No difference was flagged. Should something have changed here?

src/null-safety/index.md
atsansone: Done.

src/language/variables.md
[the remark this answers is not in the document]
parlough: Just an empty line before the header.

examples/misc/bin/cheatsheet/nullable.dart
```diff
@@ -4,13 +4,13 @@ void main() {
   {
     // #docregion invalid-null
     // ignore: invalid_assignment
-    int a = null; // INVALID in null-safe Dart.
+    int a = null; // INVALID.
```
parlough, line 7: To match the snippet in the file:

```suggestion
    int a = null; // INVALID
```

examples/misc/bin/cheatsheet/nullable.dart
```diff
@@ -4,13 +4,13 @@ void main() {
...
     // #enddocregion invalid-null
   }
 
   {
     // #docregion valid-null
-    int? a = null; // Valid in null-safe Dart.
+    int? a = null; // Valid.
```
parlough, line 13: To match the snippet in the cheatsheet:

```suggestion
    int? a = null; // Valid
```

atsansone:
@parlough : Back to you. It passed those two prior build errors. I'd like to get this merged so @mit-mit has no conflicts with his work and so I can create a second PR for the larger changes.

src/null-safety/index.md
```diff
@@ -78,11 +73,11 @@ Dart null safety support is based on the following three core design principles:
 
 ## Dart 3 and null safety
 
-Dart 3---planned for a mid-2023 release---always has sound null safety.
-Dart 3 will prevent code from running without it.
+Dart 3 code has built-in sound null safety.
```
parlough, line 76:
```suggestion
Dart 3 has built-in sound null safety.
```

src/null-safety/index.md
```diff
@@ -160,8 +157,10 @@ doesn't depend on Dart 3 breaking changes:
 
 ## Dart 2.x and null safety {#enable-null-safety}
 
-In Dart 2.12 to 2.19, null safety is a configuration option in the pubspec. Null
-safety is not available in SDK versions prior to Dart 2.12.
+From Dart 2.12 to 2.19, you need to enable null safety.
+To enable null safety, set the `sdk` option to `'>=2.12.0'`
+in the `pubspec.yaml` file.
```
parlough, line 162: Remove these sentences, the next few sentences (L166-169) go into greater detail about it already.

```suggestion
```

src/null-safety/index.md
```diff
@@ -186,14 +194,12 @@ versions 2.12 to 2.19.
 $ cd my_app
 $ dart migrate
 ```
-
-For more details on how to migrate your code to null safety,
-see the [migration guide][].
-
+To learn how to migrate your code to null safety,
```
parlough, line 197: This needs a new-line before it:

```suggestion

To learn how to migrate your code to null safety,
```

src/overview.md
```diff
@@ -42,16 +42,15 @@ allowing the use of a `dynamic` type combined with runtime checks,
 which can be useful during experimentation or
 for code that needs to be especially dynamic.
 
-Dart offers [sound null safety](/null-safety),
-meaning that values can’t be null unless you say they can be.
+Dart offers [sound null safety](/null-safety).
```
parlough, line 45: "offers" make it seem optional.

```suggestion
Dart has built-in [sound null safety](/null-safety).
```

parlough (approved):
Left a few final comments. I don't need to see again after those are addressed as other cleanup can happen later. Thanks!

parlough:
Thanks for making those changes. If you have other changes planned, please land this and make them in a separate PR :)

src/language/variables.md
```diff
@@ -38,6 +38,34 @@ String name = 'Bob';
...
+
+Null safety prevents errors that result from unintentional access
+of variables set to `null`.
+
+For example, if a method expects an integer but receives `null`,
+your app causes a runtime error.
+This type of error, a null dereference error, can be difficult to debug.
+
+With sound null safety, all variables require a value.
+This means Dart considers all variables _non-nullable_.
```
munificent, line 53: This isn't true. This is fine:

```dart
var x = null;
```

Dart considers `null` to be a totally valid value and a perfectly fine value to store in variables. It's also perfectly happy to implicitly initialize variables to `null` on your behalf if you don't initialize it:

```dart
String? maybeString;
print(maybeString == null); // Prints "true".
```

The invariant Dart gives you is:

1. It's never possible to observe an uninitialized variable. It may get default initialized to `null`, but you'll never ever see garbage bits.
2. If a variable doesn't allow `null`, the language forces you to initialize it to some explicit value, since there is no default initialization for non-nullable types (unlike, say, Go or C#, where all types support default initialization).
3. If an expression or variable's type is non-nullable, then it will never evaluate to `null`.
4. If an expression or variable can be `null`, you can't access any properties or call any methods on it (unless those methods are valid for `null`, like `toString()`).

I'm not sure how best to summarize that here.

munificent (approved):

mit-mit:
@atsansone can you take a look at the comments Bob added? Then I think we're good to go.

src/language/variables.md
```diff
@@ -38,6 +38,34 @@ String name = 'Bob';
...
 
+## Null safety
+
+The Dart language enforces sound null safety.
+
+Null safety prevents errors that result from unintentional access
+of variables set to `null`.
+
+For example, if a method expects an integer but receives `null`,
+your app causes a runtime error.
```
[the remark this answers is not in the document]
atsansone: Thanks @munificent for this clarification. I went with a little of both:

```markdown
For example,
say you want to find the absolute value of a variable `i` of type `int`.
If `i` receives `null`, `i.abs()` causes a null dereference error.

In other languages, if you allow method calls and property accesses on
nullable expressions, this could lead to runtime errors.
As the Dart compiler prohibits these actions, its apps can't cause
these runtime errors.
```

src/language/variables.md
atsansone: I think I have a summary:
```markdown
In Dart, variables must be initialized before they can be used.
If a variable is not initialized, Dart assigns it the default value of null.
However, it is never possible to observe an uninitialized variable.
Dart also prevents you from accessing properties or calling methods
on a variable that equals `null`.
Unless you design the method to handle `null` values, you can't use that method
on a variable that equals `null`.
```

src/language/variables.md
```diff
@@ -38,6 +38,48 @@ String name = 'Bob';
   of using `var`, rather than type annotations, for local variables.
 {{site.alert.end}}
 
+## Null safety
```
atsansone, line 41: @munificent : I rewrote this section. Maybe I should move it to the overview. Let me know your thoughts.

src/language/variables.md
```diff
@@ -38,6 +38,48 @@ String name = 'Bob';
...
+
+The Dart language enforces sound null safety.
+
+Null safety prevents an error that results from unintentional access
+of variables set to `null`. The error is called a null dereference error.
+If you try to access a property or call a method on a variable set to `null`,
+that causes a null dereference error.
```
munificent, line 48: Null safety doesn't directly have to do with variables. For example:

```dart
int? someFunction() => null;

main() {
  someFunction().abs();
}
```

Here, there are no variables at all. But this program would throw a null reference exception if Dart didn't have sound null safety.

I would maybe say something like:

A null dereference error occurs when you access a property or call a method on an expression that evaluates to `null` (unless it's a property of method that `null` itself supports, like `toString()` or `hashCode`). With null safety, the Dart compiler detects these potential errors at compile time.

src/language/variables.md
```diff
@@ -38,6 +38,48 @@ String name = 'Bob';
...
+For example, say you want to find the absolute value of an `int` variable `i`.
+If `i` is `null`, calling `i.abs()` causes a null dereference error.
+In other languages, trying this could lead to a runtime error,
+but Dart's compiler prohibits these actions.
+Therefore, Dart apps can't cause runtime errors.
+
+Null safety introduces two key changes to type declaration.
```
munificent, line 55: "declaration" -> "declarations". (Unless you mean "declaration" as a verb, but that's a pretty uncommon use in this context.)

We tend to use "type annotations" to refer to what you're talking about here. "Type declaration" sounds like you're declaring a new type, like:

```dary
typedef Inches = int;
```

This is what I'd call a "type declaration".

src/language/variables.md
```diff
@@ -38,6 +38,48 @@ String name = 'Bob';
...
+First, you can specify that a variable _can_ be set to `null`.
+To call out nullability, you add a `?` to the end of the type declaration.
+
+```dart
+String? name  // Nullable type. Can be `null` or string.
+
+String name   // Default type. Cannot be `null` but can be string.
+```
+
+Second, you must initialize variables before using them.
```
munificent, line 66: This is a key change, but it's not a change to type declarations/annotations. I'd probably say:

---

Null safety introduces two key changes.

First, when specifying a type for a variable, parameter, return type, etc., you can control whether the type allows `null`. ...

Second, you must initialize variables before using them.

---

There is a third key change to call out here: You can't access properties or call methods on an expression whose type is nullable (unless it's a property or method that's defined on `Null` like `hashCode` or `toString()`).

src/null-safety/index.md
```diff
@@ -3,39 +3,39 @@ title: Sound null safety
...
-(e.g. `int? i`),
-and only then can they contain either a `null` *or*
+your app causes a runtime error.
+This type of error, a null dereference error, can be difficult to debug.
+
+With sound null safety, all variables require a value.
+This means Dart considers all variables _non-nullable_.
+You can assign values of the declared type only, like `int i=42`.
+You can never assign a value of `null` to default variable types.
+To specify that a variable type can have no value, add a `?` after
```
munificent, line 19: We definitely shouldn't describe a variable as having "no value". `null` is most definitely a value. Consider:

```dart
var thing = null;
var stuff = [1, thing, 'three']);
print(stuff.length); // 3.
```

If `null` isn't a value, then what's the second element in this list?

Even before null safety, Dart ensured that all variables had a value. To ensure that, it would default initialize everything to `null`.

What null safety does is prevent you from *doing stuff* (accessing members) on `null` that it doesn't support.

src/language/variables.md
atsansone: Added.

src/language/variables.md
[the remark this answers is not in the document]
munificent: I wouldn't say "If `i` receives `null`". In an object-oriented language "receive" is usually used for method calls ("message sends" in Smalltalk parlance). I'd just say "If `i` is `null`".

src/language/variables.md
munificent: > If a variable is not initialized, Dart assigns it the default value of null.

That will throw off some readers who will think it means that that variables aren't sound with respect to `null`. I'd summarize it as:

* In Dart, the compiler requires you to initialize a variable before it will let you use it.
* If the variable's type is nullable, the language will default initialize it to `null` for you. Otherwise, you have to explicitly initialize it yourself before you can use it.
* You can't call a method or getter on an expression whose type is nullable (unless it's one of a couple of methods that are defined on `null` too).

You can't really "design the method to handle `null`". There are a couple members defined in the Dart core libraries on the `Null` class that we add but end users can't add anything else to that class. (You *can* define extensions on nullable types, but that's probably out of scope here and not really related to null safety.)

src/language/variables.md
```diff
@@ -44,33 +44,41 @@ The Dart language enforces sound null safety.
 
 Null safety prevents an error that results from unintentional access
 of variables set to `null`. The error is called a null dereference error.
-If you try to access a property or call a method on a variable set to `null`,
-that causes a null dereference error.
+A null dereference error occurs when you access a property or call a method
+on an expression that evaluates to `null`.
+An exception to this rule is when `null` supports the property of method,
+like `toString()` or `hashCode`). With null safety, the Dart compiler
+detects these potential errors at compile time.
```
munificent, line 51: 👍

src/language/variables.md
```diff
@@ -44,33 +44,41 @@ The Dart language enforces sound null safety.
...
 Therefore, Dart apps can't cause runtime errors.
 
-Null safety introduces two key changes to type declaration.
+Null safety introduces three key changes:
 
-First, you can specify that a variable _can_ be set to `null`.
-To call out nullability, you add a `?` to the end of the type declaration.
+1.  When you specify a type for a variable, parameter, or another
+    relevant component, you can control whether the type allows `null`.
+    To call out nullability, you add a `?` to the end of the type declaration.
```
munificent, line 63: "call out nullability" -> "enable nullability".

("Call out" is sort of an auto-antonym.)

src/language/variables.md
```diff
@@ -44,33 +44,41 @@ The Dart language enforces sound null safety.
...
 
-```dart
-String? name  // Nullable type. Can be `null` or string.
+    ```dart
+    String? name  // Nullable type. Can be `null` or string.
 
-String name   // Default type. Cannot be `null` but can be string.
-```
+    String name   // Default type. Cannot be `null` but can be string.
```
munificent, line 68: "Default" -> "Non-nullable". I don't think we use "default" anywhere else to refer to types that aren't nullable, but we do use "non-nullable".

src/language/variables.md
```diff
@@ -44,33 +44,41 @@ The Dart language enforces sound null safety.
...
+    Nullable variables default to `null`, so they are initialized by default.
+    Dart doesn't set initial values to default types.
+    It forces you to set an initial value.
+    Dart doesn't allow you to observe an uninitialized variable.
+    This prevents you from accessing properties or calling methods
+    on a variable that equals `null`.
 
+3.  You can't access properties or call methods on an expression with a
+    nullable type. The same exception applies where it's a property or method
+    that `null` supports like hashCode or toString()).
```
munificent, line 81: Nit: Code font for `hashCode` and `toString()`?

src/null-safety/index.md
```diff
@@ -16,7 +16,7 @@ With sound null safety, all variables require a value.
 This means Dart considers all variables _non-nullable_.
 You can assign values of the declared type only, like `int i=42`.
 You can never assign a value of `null` to default variable types.
-To specify that a variable type can have no value, add a `?` after
+To specify that a variable type can have a `null` value, add a `?` after
 the type label: `int? i`.
```
munificent, line 20: "type label" -> "type" or "type annotation".

(Labels are a different, rarely used feature.)

src/language/variables.md
atsansone: We have used `default`, but `non-nullable` is better.

munificent (approved):

src/_guides/language/coming-from/swift-to-dart.md
```diff
@@ -541,21 +539,20 @@ let str: String? = nil
 let count = str?.count ?? 0
 ```
 
-Additionally, Dart offers a
-null safe version of the cascade operator (`?..`),
-which ignores any operations when
-the target object is `null`.
+In addition, Dart offers a
+[null safe version of the cascade operator][] (`?..`).
```
parlough, line 543:
```suggestion
null safe version of the cascade operator (`?..`).
```

PR #4821, opened 2023-05-08
No longer warn about caret syntax for SDK constraint

parlough:
Closes https://github.com/dart-lang/site-www/issues/4693


MaryaBelanger (approved):
Thanks!

PR #4823, opened 2023-05-08
[Dart 3] Adjustments for removed linter rules

parlough:
Includes the removals that I've manually reverted in the data file now that this branch documents Dart 3.

Also adds a specific text if a linter rule was removed.

Example: https://dart-dev--pr4823-misc-dart-3-linter-r-bbmsyqgj.web.app/tools/linter-rules#prefer_equal_for_default_values


MaryaBelanger (approved):
LGTM

PR #4824, opened 2023-05-08
[Dart 3] Cross links between new pages

MaryaBelanger:


src/language/branches.md
```diff
@@ -113,7 +115,8 @@ switch (command) {
     executeNowClosed();   // Runs for both OPEN and PENDING.
 }
 ```
-For an empty case that does not fall through, use [`break`][break] for its body.
+
+For more details, check out the patterns [section on switch][].
```
parlough, line 119: It might be helpful to add more information on what kind of details, otherwise I am not sure why I would click this link. Depending on what you are thinking, something like:

```suggestion
To learn more about patterns in case clauses, 
check out the patterns documentation on [Switch statements and expressions][].
```

src/language/records.md
```diff
@@ -84,11 +84,7 @@ print(record.b);     // Prints true
 print(record.$2);    // Prints 'last'
 ```
 
-To streamline record field access even more, see the page on [Patterns][].
-
-{% comment %}
-    TODO: link to patterns page, specifically records destructuring section.
-{% endcomment %}
+To streamline record field access even more, see the page on [Patterns][pattern].
```
parlough, line 87:
```suggestion
To streamline record field access even more, 
check out the page on [Patterns][pattern].
```

parlough (approved):
Two minor suggestions, but other than that, looks good to me. Thanks!

PR #4827, opened 2023-05-09
Misc redirect update, fixes, and cleanup

parlough:


firebase.json
```diff
@@ -89,16 +88,15 @@
       { "source": "/dart-tips{,/**}", "destination": "/guides/language/effective-dart", "type": 301 },
       { "source": "/dart-vm", "destination": "/server", "type": 301 },
       { "source": "/dart-vm/:page*", "destination": "/server/:page*", "type": 301 },
-      { "source": "/dart-vm/dart-by-example", "destination": "/samples", "type": 301 },
+      { "source": "/dart-vm/dart-by-example", "destination": "/language", "type": 301 },
```
MaryaBelanger, line 91: This is going to `/server/dart-by-example`

MaryaBelanger (approved):
Just the comment, not working as expected

firebase.json
parlough: I'm going to remove this one, as it was never working due to its position after the previous redirect.

PR #4829, opened 2023-05-09
Class modifier example fix

MaryaBelanger:


parlough (approved):
This is much easier to read, thanks for making these adjustments! The commas before the ... help too :D 

Handling cases like these can be a bit finicky with the code excerpter when we get to that, so I'm happy to help!

PR #4831, opened 2023-05-09
Update language repository links to use `main` branch

parlough:
Looks like the repository switched over sometime in the past few days :)


MaryaBelanger (approved):
Good catch, thank you!

PR #4834, opened 2023-05-09
Expand on `mixin class`

### LINKED ISSUE #4826: Feedback on Dart 3 mixins and class modifier documentation
### Page URL

https://dart.dev/language/mixins.html

### Page source

https://github.com/dart-lang/site-www/tree/main/src/language/mixins.md

### Describe the problem

The documentation should explain how the `mixin` modifier interacts with the `class` modifier. For example (not tested!):

```dart
abstract mixin class Musician {
  // Note that we don't an `on` clause here because the class defines the
  // abstract method that must be defined i.e. `playInstrument` 
  void playInstrument(String instrumentName);

  void playPiano() {
    playInstrument('Piano');
  }
  void playFlute() {
    playInstrument('Flute');
  }
}

class Virtuoso with Musician { // Use `Musician` as a mixin
  void playInstrument(String instrumentName) {
    print('Plays with $instrumentName beautifully');
  }  
} 

class Novice extends Musician { // Use `Musician` as a base class
  void playInstrument(String instrumentName) {
    print('Plays with $instrumentName poorly');
  }  
} 
```

### Expected fix

_No response_

### Additional context

_No response_

MaryaBelanger:
Fixes #4826 

I had trouble making the connection between what the `on` clause is for and how using `abstract` provides the same benefit. Here's what _I think_ the reason is (added to `mixins.md`), but I could be off:

> By declaring `Musician` abstract, you force any type that uses
it to define the abstract method its behavior depends on. 

> This is similar to how the `on` directive ensures a mixin has access to any
interfaces it depends on by specifying the superclass of that interface.

*Also not sure if I'mm using "interfaces" correctly there.

I assumed that this would be a common use case, so I put this explanation in its own subsection on the mixin page, "`abstract mixin class`".

### DIFF
--- a/src/language/mixins.md
+++ b/src/language/mixins.md
@@ -127,8 +127,8 @@ class Novice extends Musician { // Use Musician as a class
 } 
 ```
 
-By declaring `Musician` abstract, you force any type that uses
-it to define the abstract method its behavior depends on. 
+By declaring the `Musician` mixin as abstract, you force any type that uses
+it to define the abstract method upon which its behavior depends. 
 
 This is similar to how the `on` directive ensures a mixin has access to any
 interfaces it depends on by specifying the superclass of that interface.
--- a/src/language/mixins.md
+++ b/src/language/mixins.md
@@ -89,6 +89,50 @@ A `mixin` declaration defines a mixin. A `class` declaration defines a [class][]
 A `mixin class` declaration defines a class that is usable as both a regular class
 and a mixin, with the same name and the same type.
 
+Any restrictions that apply to classes and mixins also apply to mixin classes:
+
+- Mixins can't have `extends` or `with` clauses, so neither can a `mixin class`.
+- Classes can't have an `on` clause, so neither can a `mixin class`. 
+
+### `abstract mixin class`
+
+You can achieve similar behavior to the `on` directive for a mixin class. 
+Make the mixin class `abstract` and define the abstract methods its behavior 
+depends on:
+
+```dart
+abstract mixin class Musician {
+  // No 'on' clause, but an abstract method that other types must define if 
+  // they want to use (mix in or extend) Musician: 
+  void playInstrument(String instrumentName);
+
+  void playPiano() {
+    playInstrument('Piano');
+  }
+  void playFlute() {
+    playInstrument('Flute');
+  }
+}
+
+class Virtuoso with Musician { // Use Musician as a mixin
+  void playInstrument(String instrumentName) {
+    print('Plays with $instrumentName beautifully');
+  }  
+} 
+
+class Novice extends Musician { // Use Musician as a class
+  void playInstrument(String instrumentName) {
+    print('Plays with $instrumentName poorly');
+  }  
+} 
+```
+
+By declaring `Musician` abstract, you force any type that uses
+it to define the abstract method its behavior depends on. 
+
+This is similar to how the `on` directive ensures a mixin has access to any
+interfaces it depends on by specifying the superclass of that interface.
+
 [language version]: /guides/language/evolution#language-versioning
 [class]: /language/classes
 [class modifiers]: /language/class-modifiers
--- a/src/language/mixins.md
+++ b/src/language/mixins.md
@@ -89,7 +89,7 @@ A `mixin` declaration defines a mixin. A `class` declaration defines a [class][]
 A `mixin class` declaration defines a class that is usable as both a regular class
 and a mixin, with the same name and the same type.
 
-Any restrictions that apply to classes and mixins also apply to mixin classes:
+Any restrictions that apply to classes or mixins also apply to mixin classes:
 
 - Mixins can't have `extends` or `with` clauses, so neither can a `mixin class`.
 - Classes can't have an `on` clause, so neither can a `mixin class`. 
@@ -116,13 +116,13 @@ abstract mixin class Musician {
 
 class Virtuoso with Musician { // Use Musician as a mixin
   void playInstrument(String instrumentName) {
-    print('Plays with $instrumentName beautifully');
+    print('Plays the $instrumentName beautifully');
   }  
 } 
 
 class Novice extends Musician { // Use Musician as a class
   void playInstrument(String instrumentName) {
-    print('Plays with $instrumentName poorly');
+    print('Plays the $instrumentName poorly');
   }  
 } 
 ```


MaryaBelanger:
This is pointing at `v3.0` but if it can't get reviewed before merging that into `main` tomorrow morning, it'll be fine to merge this directly into `main` later on.

src/language/mixins.md
```diff
@@ -89,6 +89,50 @@ A `mixin` declaration defines a mixin. A `class` declaration defines a [class][]
 A `mixin class` declaration defines a class that is usable as both a regular class
 and a mixin, with the same name and the same type.
 
+Any restrictions that apply to classes and mixins also apply to mixin classes:
```
munificent, line 92: I'd probably change "classes and mixins" to "classes or mixins" because it's really if there is a restriction for either of those, then the restriction applies to mixin classes to. Otherwise, you could misread this as saying only the restrictions *shared* between mixins and classes are applied to mixin classes.

src/language/mixins.md
```diff
@@ -89,6 +89,50 @@ A `mixin` declaration defines a mixin. A `class` declaration defines a [class][]
...
+
+- Mixins can't have `extends` or `with` clauses, so neither can a `mixin class`.
+- Classes can't have an `on` clause, so neither can a `mixin class`. 
+
+### `abstract mixin class`
+
+You can achieve similar behavior to the `on` directive for a mixin class.
```
brianquinlan, line 99: This is much better IMO and clearly explains why you might want to use this approach.

I'm not sure how convincing the example is though...

There are some examples in the wild (e.g. `MapBase` in the SDK and `BaseClient` in `package:http`) but they are too complex and don't go with the Musician theme.

src/language/mixins.md
```diff
@@ -89,6 +89,50 @@ A `mixin` declaration defines a mixin. A `class` declaration defines a [class][]
...
+    playInstrument('Piano');
+  }
+  void playFlute() {
+    playInstrument('Flute');
+  }
+}
+
+class Virtuoso with Musician { // Use Musician as a mixin
+  void playInstrument(String instrumentName) {
+    print('Plays with $instrumentName beautifully');
```
brianquinlan, line 119: the $instrumentName

src/language/mixins.md
```diff
@@ -89,6 +89,50 @@ A `mixin` declaration defines a mixin. A `class` declaration defines a [class][]
...
+  }  
+} 
+
+class Novice extends Musician { // Use Musician as a class
+  void playInstrument(String instrumentName) {
+    print('Plays with $instrumentName poorly');
```
brianquinlan, line 125: the $instrumentName

brianquinlan (approved):

src/language/mixins.md
munificent, line 99: `on` clauses are really confusing. Fortunately, almost no one actually needs to use them.

People often think they are used to specify the methods than a mixin can call on itself and then get implemented by the thing that applies the mixin. That's what your example is about here.

But you don't need an `on` clause for that at all. As you do here, you can just make the type abstract and define abstract methods. Another way to accomplish something similar is to put an `implements` clause on the mixin or mixin class and then not actually implement the interface.

The only time you need an `on` clause is when the mixin wants to do `super` calls. It specifies the interface that *those* are resolved against. So in:

```dart
mixin M {}

class B {}
class C extends B with M {}
```

If there's some method in C that you want to be able to call from M, you can just declare it abstract in M and call it, like in your example:

```dart
mixin M {
  void c(); // Declare abstract.
  void test() {
    c();
  }
}

class B {}
class C extends B with M {
  void c() { ... }
}
```

You need an `on` clause if M wants to have a `super` call that needs to be resolved against *B* in the example here:

```dart
abstract class BStuff {
  void b();
}

mixin M on BInterface {
  void test() {
    super.b();
  }
}

class B {
  void b() { ... }
}

class C extends B with M {}
```

Honestly, `on` clauses and `super` calls in mixins are just not a great feature. Almost any time you think you need one, there's probably a better design you should use instead. But Flutter used `super` calls in a couple of key mixins early on (because the VM allowed it even though the language didn't specify it), so we were sort of forced to turn it into a real feature. But I definitely wouldn't emphasize it in the tour since it tends to confuse users and it's rarely what they want.

munificent: Summarizing: I think your example is OK here, but I wouldn't mention it in the context of `on` clauses. If that makes the entire example not worth keeping, I think it's fine to eliminate this section.

brianquinlan: The text before this section contains these sentences:

> Sometimes you might want to restrict the types that can use a mixin.
> For example, the mixin might depend on being able to invoke a method
> that the mixin doesn't define.

That is confusing given what you just said. There might not be time for this but it might make sense to reorder this page i.e. have an example of calling the mixing classes method from the mixin near the top (where the `on` example currently is), then have a section about `class mixin` and then finally a section explaining `on` with the "you will almost certainly never need" caveat and an explanation that it is only useful for `super` calls.

MaryaBelanger: Hmm.. ok, I think what I'll do _for now_ is keep the new section referencing similarity to using `on`, but more of an aside the purpose statement it currently is. Like Brian said, this is a direct response to the earlier text describing `on`. 

Later* I'll rewrite the section to take out any emphasis on `on` and instead describe the same example for both mixins and mixin classes simply as "the use case where you want to require that whatever uses the mixin specifies some methods".

*Only delaying it because, again like Brian said, I think it'd be best to rearrange the page, since ...
- this isn't just a use case for mixin classes, 
- and, like Bob said, the existing description of `on` for regular mixins seems misrepresented and could probably use a rewrite

And doing that will take a bit of attention that I won't have before publishing the new pages. I'd like to include this info on the first publish. @munificent if you think the content just absolutely shouldn't be published, please let me know!

parlough:
I pulled out the `sealed` changes in https://github.com/dart-lang/site-www/pull/4847 so I could land those and other fixes and so discussion can continue on mixins here.

parlough:
I've changed the base branch to `main` since the `v3.0` branch landed in `main`.

MaryaBelanger:
@atsansone could you approve this to get the information out for now, so I can come back and apply Bob's concerns later on?

src/language/mixins.md
```diff
@@ -89,6 +89,50 @@ A `mixin` declaration defines a mixin. A `class` declaration defines a [class][]
...
+} 
+
+class Novice extends Musician { // Use Musician as a class
+  void playInstrument(String instrumentName) {
+    print('Plays the $instrumentName poorly');
+  }  
+} 
+```
+
+By declaring `Musician` abstract, you force any type that uses
```
atsansone, line 130: **issue:** [GSG](https://developers.google.com/style/code-in-text#grammatical-treatment-of-code-elements) doesn't want us to refer to a class or any code object as a noun. This should be written as:
```suggestion
By declaring the `Musician` mixin as abstract, you force any type that uses
```

src/language/mixins.md
```diff
@@ -89,6 +89,50 @@ A `mixin` declaration defines a mixin. A `class` declaration defines a [class][]
...
+it to define the abstract method its behavior depends on.
```
atsansone, line 131: **nit:** Better not to end with a preposition.
```suggestion
it to define the abstract method upon which its behavior depends. 
```

src/language/mixins.md
```diff
@@ -89,6 +89,50 @@ A `mixin` declaration defines a mixin. A `class` declaration defines a [class][]
...
+  void playInstrument(String instrumentName) {
+    print('Plays the $instrumentName poorly');
+  }  
+} 
+```
+
+By declaring `Musician` abstract, you force any type that uses
+it to define the abstract method its behavior depends on. 
+
+This is similar to how the `on` directive ensures a mixin has access to any
```
atsansone, line 133: **nit:** Active voice. Not required, though.
```suggestion
This works like how the `on` directive ensures a mixin has access to any
```

src/language/mixins.md
atsansone: My two cents: It's worth mentioning in passing for coverage. It should be reviewed in the next week or so to see if it's truly relevant. I'm a fan of being opinionated in docs, even if the opinion is to skip something entirely.

atsansone (approved):
One issue, otherwise LGTM. No need for a second pass.

Issue #4828, opened 2023-05-09
Create code excerpts (and maybe tests) for Dart 3 documentation


parlough:
We should add code excerpts and maybe tests for the new code snippets coming in Dart 3 documentation. This will help us ensure that the snippets compile, are formatted properly, and have no analysis errors or other issues.

~~I'm going to create a new folder for the documentation rather than putting them in the super crowded misc.~~

**Edit:** I created a new folder under `/examples/language` which can replicate the language sidenav structure: https://github.com/dart-lang/site-www/tree/main/examples/language

parlough:
@MaryaBelanger Just wanted to note that I moved this to a higher priority, since I already caught a few switches that were broken in https://github.com/dart-lang/site-www/pull/4847. I'm worried there is other code that is slightly broken.

After I finish my finals, I think this is the first thing I will work on unless you want to work on it.

MaryaBelanger:
@parlough Ok, I'm going to write the language evolution entry and then I'll work on it right away. Thanks for pointing it out

MaryaBelanger:
- patterns.md done in #4857 
- records.md in #4856 
- class-modifiers.md in progress #4858 
- branches.md in progress #4859 
- ToDo: pattern-types.md

parlough:
There are still improvements to be made, but it seems all of the snippets are using excerpts now. Closing this in favor of more specific tracking as necessary. Thanks everyone!

PR #4838, opened 2023-05-09
Move and add code excerpts for new loops page

parlough:
Mostly I moved the loops code excerpts to a new `/language` examples directory to give us an example as we add new code excerpts for `/language`, so they aren't hidden deeply under `misc`. My thinking is they will emulate the `/language` sidebar structure to make it easy to navigate.

This does however changes the new example in loops to also follow the candidates theme of the page and adds a code excerpt for that. Let me know what you think :)

Staged updated example: https://dart-dev--pr4838-feature-control-flow-copvea4t.web.app/language/loops#:~:text=To%20process%20the%20values%20obtained

Contributes to https://github.com/dart-lang/site-www/issues/4828

### DIFF
--- a/examples/language/analysis_options.yaml
+++ b/examples/language/analysis_options.yaml
@@ -1,0 +1,1 @@
+include: ../analysis_options.yaml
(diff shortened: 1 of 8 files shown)


MaryaBelanger (approved):
I like this new organization, good start with these smaller examples!

PR #4842, opened 2023-05-10
[Dart 3] Priority fixes

parlough:
I've included the changes from https://github.com/dart-lang/site-www/pull/4766#discussion_r1189106464 so that there isn't a broken sample. We can land any adjustments you want to make afterwards.


parlough:
\cc @MaryaBelanger I'm going to land Bob's record changes so we don't have a broken sample there, but please feel free to take a closer look when you get a chance (https://github.com/dart-lang/site-www/issues/4843).

PR #4847, opened 2023-05-10
Expand on `sealed` modifier and fix some related examples

parlough:
Mostly pulled from https://github.com/dart-lang/site-www/pull/4834 since not all of it has been decided on.

I also added a few other fixes related to switches while I was at it.

\cc @MaryaBelanger For reference :)


parlough:
@domesticmouse I would like to get this one in, but would appreciate if you could have a really quick look over to confirm I didn't mess something dumb up :)

src/language/class-modifiers.md
```diff
@@ -202,9 +202,8 @@ To create a known, enumerable set of subtypes, use the `sealed` modifier.
 This allows you to create a switch over those subtypes that is statically ensured
 to be [_exhaustive_][exhaustive].
 
-The `sealed` modifier prevents a class from being extended or implemented outside
-of its own library. Sealed classes also prevent construction,
-and are therefore implicitly [abstract](#abstract).
+The `sealed` modifier prevents a class from being extended or
+implemented outside of its own library.
```
domesticmouse, line 206: This is different to the above. Arguably with this text a sealed class can be instantiated inside its library. Maybe split into two seperate sentences. Can't be extended outside of its library. Can't be instantiated directly.

domesticmouse (approved):

src/language/class-modifiers.md
parlough: Yeah this mention was moved below the code block to extend that discussion. But now that you say that, it does seem a bit far away for an important concept. 

I'll move that back up for now. Thanks!

parlough: Okay, I moved it back up, should cover that case fine now :)

PR #4845, opened 2023-05-10
Add Dart 3 Flutter codelab to index

parlough:
I know this is a Flutter codelab, but the material is relevant to Dart users too with its focus. Let me know what you think :)

PR #4855, opened 2023-05-10
Dart 3 in sidenav

MaryaBelanger:
I dislike how the migration guide is hidden at the moment. The banner announces Dart 3, which points to the blogpost, which I'm sure points back to the migration guide... but I feel like there should be a more obvious pointer to Dart 3 content.

I put a new Dart 3 bucket _right at the top_ of the sidenav:
image

Right now, all that's under it is the migration guide:
Pasted Graphic

Once @munificent 's class modifier guidance is ready, I think it should go under there too. Also, once the Dart 3 codelab (#4845) is ready, it should go there too:
Pasted Graphic 1

What does everyone think about this? No links have to change so I think it's a very minor effort / major impact change we can make to emphasize important new content

### DIFF
--- a/src/_data/side-nav.yml
+++ b/src/_data/side-nav.yml
@@ -1,3 +1,8 @@
+- title: Dart 3
+  expanded: false
+  children:
+    - title: Dart 3 migration guide
+      permalink: /resources/dart-3-migration
 - title: Tutorials & codelabs
   expanded: false
   children:
@@ -330,8 +335,6 @@
       permalink: /guides/language/evolution
     - title: Language specification
       permalink: /guides/language/spec    
-    - title: Dart 3 migration guide
-      permalink: /resources/dart-3-migration
     - title: Coming from ...
       expanded: false
       children:


mit-mit:
Hmm, not entirely sure.

My current assumption is that most existing code won't actually need Dart 3 migration, so perhaps this is all _too_ visible? There is a lot of content in that guide to consume if you are never going to his any of it.

mit-mit:
If the main goal is to get a clear Dart 3 landing page, then I think a page that has more of a focus on a high-level summary to what all the new features are (i.e. a short-form of the blog post) + then a link to migration guide, would be getter.

MaryaBelanger:
> If the main goal is to get a clear Dart 3 landing page, then I think a page that has more of a focus on a high-level summary to what all the new features are (i.e. a short-form of the blog post) + then a link to migration guide, would be getter.

@mit-mit that makes sense. Idk if I'll have time to do that this morning, and I'm sure it's mostly important for this morning, so I'll probably just not implement this for now. Thanks!

PR #4856, opened 2023-05-10
Added analysis to records snippets

craiglabenz:
Snippets moved to a `_test.dart` file, but I'm not actually sure that was critical vs an analysis-only migration.

Contributes to https://github.com/dart-lang/site-www/issues/4828

### DIFF
--- a/src/language/records.md
+++ b/src/language/records.md
@@ -81,11 +81,10 @@ these names are purely for documentation and don't affect the record's type:
 
 <?code-excerpt "language/test/records_test.dart (record-type-matched-names)"?>
 ```dart
-({int a, int b}) recordAB = (a: 1, b: 2);
-({int x, int y}) recordXY = (x: 3, y: 4);
+(int a, int b) recordAB = (1, 2);
+(int x, int y) recordXY = (3, 4);
 
-// Compile error! These records don't have the same type.
-// recordAB = recordXY;
+recordAB = recordXY; // OK.
 ```
 
 This is similar to how positional parameters
--- a/src/language/records.md
+++ b/src/language/records.md
@@ -19,8 +19,8 @@ functions, and store them in lists.
 _Records expressions_ are comma-delimited lists of named or positional fields,
 enclosed in parentheses:
 
+<?code-excerpt "language/test/records_test.dart (record-syntax)"?>
 ```dart
-// Record expression:
 var record = ('first', a: 2, b: true, 'last');
 ```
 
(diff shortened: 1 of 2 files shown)


MaryaBelanger (approved):
Preliminary approval, obviously tests have to pass first but we'll figure that out!
Thank you SO MUCH

src/language/records.md
```diff
@@ -64,21 +67,25 @@ the [record's type definition](#record-types), or its _shape_.
...
 
-recordAB = recordXY; // Compile error! These records don't have the same type.
+// Compile error! These records don't have the same type.
+// recordAB = recordXY;
 ```
 
 In a record type annotation, you can also name the *positional* fields, but
 these names are purely for documentation and don't affect the record's type:
 
+<?code-excerpt "language/test/records_test.dart (record-type-mismatched-names)"?>
```
parlough, line 82: This snippet doesn't match the text, is this supposed be:

```suggestion
<?code-excerpt "language/test/records_test.dart (record-type-matched-names)"?>
```

src/language/records.md
```diff
@@ -64,21 +67,25 @@ the [record's type definition](#record-types), or its _shape_.
...
 different names have different types:
 
+<?code-excerpt "language/test/records_test.dart (record-type-mismatched-names)"?>
 ```dart
 ({int a, int b}) recordAB = (a: 1, b: 2);
 ({int x, int y}) recordXY = (x: 3, y: 4);
 
-recordAB = recordXY; // Compile error! These records don't have the same type.
+// Compile error! These records don't have the same type.
+// recordAB = recordXY;
```
parlough, line 76: I could do this as separate follow-up, but I wanted to note:

Consider removing the comments before this, so it is actually a compile error.

You can use the code-excerpt replace functionality to remove them. 

Then you can add `{:.fails-sa}` before the code excerpt so it's clear it isn't a valid code snippet:

Example of fails-sa

parlough (commented):
Thanks so much for working on these @craiglabenz! 

Adding them in a tests file is definitely preferred :D

src/language/records.md
craiglabenz: How exactly to hold the `replace` command is eluding me, so I will take you up on the offer to do this in a second pass. Please tag me when you do, though - I'd love to see the right incantation.

parlough: Sounds good. I'll open that tonight or tomorrow and make sure to tag you :)

parlough (approved):
GitHub actions is having [some trouble](https://www.githubstatus.com/incidents/pr3498h3qkfy) right now. I'm going to merge this anyway as it successfully ran some of the jobs.

Thanks so much for doing this!

PR #4866, opened 2023-05-11
Fix TODOs for SCSS

atsansone:
Fixes https://github.com/dart-lang/site-www/issues/2859


atsansone:
@parlough : This change does affect the display of `h2` because it inherits the styles from `shared` for Flutter. I'm willing to go with that for now to get this cleared.

parlough:
I'll take a look at this early next week.

However, after taking a look at the staged site, the new h2 size is very large for the Dart site compared to other content around it and the other headers. I'd rather address that before landing this since there is no rush on these changes.

atsansone:
@parlough : You're right. I noted why in earlier comments. We may want to consider fixing this across both sites.

parlough (approved):
Thanks!

I pushed a few more fixes, including reverting h2 to the old size, adjusting some values, moving some variables, and removing some old comments.

PR #4867, opened 2023-05-11
Update banner post-I/O

MaryaBelanger:
I figure we should change this now that I/O is over?

image

### DIFF
--- a/src/_includes/banner.html
+++ b/src/_includes/banner.html
@@ -1,9 +1,10 @@
 <div class="banner">
   <p class="banner__text">
-    Dart 3 is here with records and patterns, alongside Google I/O!
+    Dart 3 is here with 
+    <a href="https://dart.dev/language/records" class="no-automatic-external">records</a>, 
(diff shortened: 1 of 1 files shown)


src/_includes/banner.html
```diff
@@ -1,9 +1,10 @@
 <div class="banner">
   <p class="banner__text">
-    Dart 3 is here with records and patterns, alongside Google I/O!
+    Dart 3 is here with
```
atsansone, line 3: **nit:** Active voice and avoid here in text.
```suggestion
    Dart 3 released, featuring
```

atsansone (approved):
LGTM% one suggestion. Not sure if mit needs to see this or not.

src/_includes/banner.html
MaryaBelanger: I'm so confused by active voice 😅 Wouldn't "released" beg the question, "_Who_ released Dart 3?", making "Dart 3 released" passive?

PR #4870, opened 2023-05-11
Add more detail to sealed definition

### LINKED ISSUE #4869: Revise explanation of sealed class in 'Class modifiers'
### Page URL

https://dart.dev/language/class-modifiers.html

### Page source

https://github.com/dart-lang/site-www/tree/main/src/language/class-modifiers.md

### Describe the problem

Some members of the Dart team have interpreted the docs to mean that `sealed` classes cannot have constructors. They can for the purposes of initializing their subclasses.

### Expected fix

Maybe:

Sealed classes also prevent construction **(but they can define constructors for their subclasses to use)**, and are therefore implicitly [abstract](https://dart.dev/language/class-modifiers#abstract). However, subclasses of sealed classes can be constructed, and sealed classes can have [factory constructors](https://dart.dev/language/constructors#factory-constructors).

### Additional context

_No response_

MaryaBelanger:
Fixes #4869


src/language/class-modifiers.md
```diff
@@ -210,9 +210,10 @@ to be [_exhaustive_][exhaustive].
...
 
-Sealed classes also prevent construction,
-and are therefore implicitly [abstract](#abstract).
-However, subclasses of sealed classes can be constructed, 
+Sealed classes are implicitly [abstract](#abstract).
+They cannot be constructed themselves,
+but they can define constructors for their subclasses to use.
+Subclasses of sealed classes can be constructed, 
 and sealed classes can have
 [factory constructors](/language/constructors#factory-constructors).
```
parlough, line 218: I wonder if we can combine these statements better? It seems a little convoluted to say they can have constructors for their subclasses. Then describe subclasses. Then describe another type of constructor they can have.

Also reading this again, it's also not guaranteed that their subclasses can be constructed (since they may be abstract as well), it's just that they are not automatically given this same restriction. Maybe we can just drop the mention of "subclasses can not be constructed" or modify it a bit? Not sure.

```suggestion
but they can define constructors for their subclasses to use and have 
[factory constructors](/language/constructors#factory-constructors).
Subclasses of sealed classes are, however, not implicitly abstract.
```

MaryaBelanger: I took this suggestion but broke it up into  bullets, lmk if that still makes sense or if it puts weird emphasis

parlough (approved):
Love it. Thanks!

PR #4872, opened 2023-05-11
Class modifiers for apis

munificent:
I'm sure this will need a decent amount of work to beat it into shape, but I think it's ready for review.

Is there something I need to change to get this to show up under "Resources" on the side navigation?


src/resources/class-modifiers-for-apis.md
```diff
@@ -0,0 +1,359 @@
...
+because it affects not just what users can do with your class or mixin,
+but also the affordances *their* subclasses can offer.
+Once you've put `base` on a type, the whole hierarchy under it
+is prohibited from being implemented.
+
+That sounds intense, but it's how most other programming languages
+have always worked.
+Most don't have implicit interfaces at all,
+so when you declare a class in Java, C#, or other languages,
+you effectively have the same limitation.
```
eernstg, line 208: Perhaps call it a 'constraint' rather than a 'limitation': It's actually fair to say that it is a limitation in those other languages, but in the context of Dart it is a deliberate choice: "I want the complete package consisting of some guarantees and some constraints", and if the constraint/limitation part is a problem then just don't write `base`.

eernstg:
@munificent, this is great, shape or non-shape! ;-)

MaryaBelanger:
> Is there something I need to change to get this to show up under "Resources" on the side navigation?

Yup, just add a `title` and `permalink` bullet for it here: https://github.com/dart-lang/site-www/blob/main/src/_data/side-nav.yml#L324

Taking a look at the rest of it now, thanks for doing this!

MaryaBelanger:
@munificent could you add a note to the main class modifiers page (`/language/class-modifiers.md`) as part of this PR? Wording variable, something like:

```
{{site.alert.note}}
  If you maintain a library, read the
  [Class modifiers for API maintainers][/resources/class-modifiers-for-apis]
  page for guidance on how to navigate these changes for your libraries. 
{{site.alert.end}}
```

I tried it at line 31 (the end of the intro section) just because there's already a version note at the top of the page, so it _kind of_ "looks better". You can do that or add it wherever you want, top of the page alongside the version note is fine too (and of course visibility is important so no point in hiding it just for looks)

MaryaBelanger (commented):
I didn't get all the way through yet, but nothing major anyway. 

Taking a break to add a couple thoughts:
- Would it make sense or be a nice plus to have each of the modifier sections in the regular class modifiers page link to their corresponding section on this page? I know the audiences are different, but I think just having "To learn more about (modifier) for library maintainers, check out..."
- I wonder if this would belong more under the Development or Packages sidenav section. Granted I couldn't really summarize the purposes of those sections, it just seems like this info might be useful to people who are concerned about package stuff or development stuff? Your call

src/resources/class-modifiers-for-apis.md
```diff
@@ -0,0 +1,359 @@
...
+will be reminded to implement the private methods.
+But note that the next section *does* apply:
+
+### Base transitivity
+
+The goal of marking a class `base` is to ensure that
+every instance of that type concretely inherits from it.
+To maintain this, the base restriction is "contagious".
+Every subtype of a type marked `base` -- *direct or indirect* --
+must also prevent being implemented too.
```
MaryaBelanger, line 194: "also" or "too", not both
```suggestion
must also prevent being implemented.
```

src/resources/class-modifiers-for-apis.md
```diff
@@ -0,0 +1,359 @@
...
+That means it must be marked `base`
+(or `final` or `sealed`, which we'll get to next).
+
+Applying `base` to a type requires some care then,
+because it affects not just what users can do with your class or mixin,
```
MaryaBelanger, line 199: Shorten sentence
```suggestion
Applying `base` to a type requires some care, then.
It affects not just what users can do with your class or mixin,
```

src/resources/class-modifiers-for-apis.md
```diff
@@ -0,0 +1,359 @@
...
+For this to be sound, the compiler enforces two restrictions:
+
+1.  The sealed class can't itself be directly constructible.
+    Otherwise, you could have an instance of `Amigo` that isn't
+    an instance of *any* of the subtypes.
+    So every `sealed` class is implicitly `abstract` too.
+
+2.  Every direct subtype of the sealed type must be in the same library
+    where the sealed type is declared.
+    This way the compiler can find them all and know that there aren't
```
MaryaBelanger, line 262:
```suggestion
    This way, the compiler can find them all. It knows that there aren't
```

src/resources/class-modifiers-for-apis.md
```diff
@@ -0,0 +1,359 @@
...
+
+// But these are both fine:
+class OtherLucky extends Lucky {}
+class OtherDusty implements Dusty {}
+```
+
+Of course, if you *want* the subtypes of your sealed type
+to be restricted as well, you can get that by marking them
+using `interface`, `base`, `final`, or `sealed`.
+
```
MaryaBelanger, line 290: I would make this it's own section (it'd be nice to point people to directly). Either top level or a subsection of sealed:

```suggestion

## `sealed` versus `final`

```

MaryaBelanger (approved):
Approved % the minor comments I made. 

I would really like this to be a blog post too, maybe it would reach more people that way? It kind of missed the Dart 3 announcement so I worry about it's visibility. It's great content and should be front and center but the site isn't really conducive to that. 

I don't want to add more work, but I would be ok with it:
- being both a page on the site AND a blog post, to balance visibility and longevity
  - that's probably bad practice but I feel like this is a special case...
- being just a blog post OR just a page on the site, if the work to achieve either is equivalent
  - we're obviously further along with adding it to the site, but if you want to choose to make it a blog post, I will personally steam roll it ahead and make sure it's published ASAP, just let me know. 

Edit: a review from @lrhn would be great before this gets published, or after if timing is a concern.

src/resources/class-modifiers-for-apis.md
```diff
@@ -0,0 +1,359 @@
...
+*   **Do you want users to be able to construct instances of it directly?**
+    In other words, is the class not abstract?
+
+*   **Do you *want* people to be able to use the declaration as a mixin?**
+    In other words, do you want them to be able to use it in `with` clauses?
+
+If the answer to both is "yes", then make it a mixin class. If the answer to
+the second is "no", then just leave it as a class. If the answer to the first
+is "no" and the second is "yes", then change it from a class to a mixin
+declaration.
```
MaryaBelanger, line 85:
```suggestion
- If the answer to both is "yes", then make it a mixin class. 
- If the answer to the second is "no", then just leave it as a class. 
- If the answer to the first is "no" and the second is "yes",
  then change it from a class to a mixin declaration.
```

munificent:
Thank you so much for picking this after I got sucked into the formatter and dropped it onto the floor.

I like the idea of it being both a blog post and a permanent guide. Maybe we could make the blog post be a condensed version that takes you to the full guide if you want to know more? @mit-mit might have an opinion.

MaryaBelanger:
I'm going to count my review as the copy review and merge this in. I'll follow up with @mit-mit about a possible blog post. Thanks!

PR #4878, opened 2023-05-12
Rename "Callable Classes" to "Callable objects"

### LINKED ISSUE #4876: Consider renaming 'Callable classes' as 'Callable instances'
### Page URL

https://dart.dev/language/callable-classes.html

### Page source

https://github.com/dart-lang/site-www/tree/main/src/language/callable-classes.md

### Describe the problem

This topic might more accurately be called "Callable Instances". The class itself isn't callable (call isn't static), but the instance is.

### Expected fix

Update language in this document (and site navigation) to refer to "Callable Instances" instead of "Callable Classes"

### Additional context

_No response_

gamebox:
Closes #4876


parlough (changes requested):
Thanks for catching and updating this!

A few last changes we'll need to make:

- `sidenav.yml` to update the entry in the side navigation menu - https://github.com/dart-lang/site-www/blob/main/src/_data/side-nav.yml#L89-L90
- `firebase.json` to add a redirect from the old page to the new one - https://github.com/dart-lang/site-www/blob/main/firebase.json#L184-L185
  ```
  { "source": "/language/callable-classes", "destination": "/language/callable-objects", "type": 301 },
  ```

gamebox:
I addressed those items, as well as a couple of other places where I saw references

gamebox:
My apologies, I didn't see that at all in my grep of the code

parlough:
> My apologies, I didn't see that at all in my grep of the code

No worries!

PR #4881, opened 2023-05-15
Prefer ^-constraint

jonasfj:
Just a suggestion, feel free to ignore the suggestion if you think the other thing is easier to understand.


parlough:
Yeah we considered this on the page but felt like the old syntax would make the difference more clear.

Perhaps we could show both or have a comment that's it's equivalent to `>= 3.0.0 < 4.0.0`?

atsansone:
@jonasfj , @parlough : For now, let's not muddy the waters by introducing two changes at the same time. Could we close this for now?

PR #4882, opened 2023-05-15
The PUB_CACHE have moved.

jonasfj:
Context: https://github.com/dart-lang/sdk/issues/52386


src/resources/dart-3-migration.md
```diff
@@ -596,6 +596,12 @@ analyzer:
   have been cleaned up [#50700](https://github.com/dart-lang/sdk/issues/50700).
 * The output of `dart format` changed a bit for
   [some code](https://github.com/dart-lang/sdk/blob/main/CHANGELOG.md#formatter).
+* On Windows the default location for the `PUB_CACHE` changed to
```
mit-mit, line 599: This sounds like the change happened in 3.0. I thought it happened in 2.8, and it's just that 3.0 no longer supports the old location?

jonasfj: Correct, I'll try to find better words :see_no_evil:

src/resources/dart-3-migration.md
```diff
@@ -596,6 +596,14 @@ analyzer:
...
 * The output of `dart format` changed a bit for
   [some code](https://github.com/dart-lang/sdk/blob/main/CHANGELOG.md#formatter).
+* Windows users with a default pub-cache in `%APPDATA%\Pub\Cache` will get a
+  new empty pub-cache in `%LOCALAPPDATA%\Pub\Cache`.
+  Starting Dart 2.8 a pub-caches would be created `%LOCALAPPDATA%\Pub\Cache`,
+  unless an existing pub-cache was found in `%APPDATA%\Pub\Cache`.
+  As of Dart 3 existing pub-caches in `%APPDATA%\Pub\Cache` will no longer be
+  used. Thus, users with such as pub-cached prior to Dart 2.8 will get a new
+  empty pub-cache. If you have `%APPDATA%\Pub\Cache\bin` in your `PATH` may wish
+  to update your `PATH` variable.
```
atsansone, line 606: **issue:** 
Write as shorter sentences per [GSG](https://developers.google.com/style/accessibility#ease-of-reading).

```suggestion
* The default location for pub-cache in Windows changed in Dart 2.8.
  It changed from `%APPDATA%\Pub\Cache` to `%LOCALAPPDATA%\Pub\Cache`.
  Between Dart 2.8 and 2.19, Dart used the pub-cache found in the old location.
  If a pub-cache wasn't found in the old location,
  Dart created an empty one in the new location.
  Dart 3 expects the pub-cache to exist in the new location.
  If one doesn't exist in the new location,
  Dart 3 creates a new, empty pub-cache in the new location.
  Dart 3 ignores any pub-cache in the old location.
  Check your `PATH` for `%APPDATA%\Pub\Cache\bin`.
  If that value exists, change it to `%LOCALAPPDATA%\Pub\Cache`.
```

jonasfj: I did a little rewrite, credits @whesse, who suggested we omitted a bunch of details.

Indeed there is no reason to explain when stuff happened.

src/resources/dart-3-migration.md
```diff
@@ -596,6 +596,11 @@ analyzer:
   have been cleaned up [#50700](https://github.com/dart-lang/sdk/issues/50700).
 * The output of `dart format` changed a bit for
   [some code](https://github.com/dart-lang/sdk/blob/main/CHANGELOG.md#formatter).
+* Ending backwards compatibility for the old location of pub-cache on Windows.
+  Prior to Dart 3 `%APPDATA%\Pub\Cache` was a fallback location for pub-cache.
+  As of Dart 3, the default pub-cache is always `%LOCALAPPDATA%\Pub\Cache`.
+  If you have added globally activated packages to your `PATH`, you may wish
+  to update `PATH` to contain `%LOCALAPPDATA%\Pub\Cache\bin`.
```
atsansone, line 603: **issue:** Clean up for length, clarity, and active voice.
```suggestion
* Ending backwards compatibility for the old location of pub-cache on Windows.
  Prior to Dart 3, `%APPDATA%\Pub\Cache` was a fallback location for pub-cache.
  Starting with Dart 3, the default pub-cache is located at `%LOCALAPPDATA%\Pub\Cache`.
  If you [added a package][]  to your `PATH` as a command line app,
  to update `PATH` to contain `%LOCALAPPDATA%\Pub\Cache\bin`.
  
  [added a package]:  /tools/pub/cmd/pub-global
```

atsansone (changes requested):
@jonasfj : A small update. I'll approve once fixed.

src/resources/dart-3-migration.md
jonasfj:
```md
   If you [added a package][]  to your `PATH` as a command line app,
   to update `PATH` to contain `%LOCALAPPDATA%\Pub\Cache\bin`.
```

You can do `dart pub global activate <package>`.

If the `<package>` being activated contains and `executable` section in `pubspec.yaml`, then executable files will be created in `$PUB_CACHE/bin`.

Thus, if you do `dart pub global activate mono_repo` you can execute: `$PUB_CACHE/bin/mono_repo`.
But the executable `mono_repo` is only available on `PATH` if you've added `$PUB_CACHE/bin` to `PATH`.

So I wouldn't really say that "you have added a package to your PATH".

What you did was:
 * (a) Globally activate one or more packages (which contains an `executable` section).
 * (b) Added `$PUB_CACHE/bin` to `PATH`.

We're trying to suggest that if you did (b), then action may be warranted.

> "you have added a package to your PATH"

Implies to me that there was some action you did that added a package to PATH.

I suggest using the term "added globally activated packages to PATH" because it might remind the user of that time they modified `PATH` in order to get globally activated packages working without having to type `dart pub global run <package>`.


--------------------------

We could also change the line to:
> If you previously added `path\to\pub-cache\bin` to `PATH` you may wish to update `PATH` to contain `%LOCALAPPDATA%\Pub\Cache\bin`.

Intentionally, avoiding to use the qualifier "If you previously added `%APPDATA\Pub\Cache\bin`", because while doing so might work, it's quite likely that users actually added the absolute path (e.g. `c:\path\to\pub-cache\bin`) to `PATH`.

I think it's viable to argue that `path\to\pub-cache\bin` can be a stand-in for both:
 * `%APPDATA\Pub\Cache\bin`, and,
 * `c:\path\to\pub-cache\bin`.

jonasfj: >  If you [added a package][]  to your `PATH` as a command line app,
>  to update `PATH` to contain `%LOCALAPPDATA%\Pub\Cache\bin`.

I don't think "added a package to your PATH" is a good description of what might have happened.

You can't add a package, only the executables declared within a package. This happens when you globally activate a package.
But the action that is concerning is "adding globally activated packages to PATH".

Which is really a shorter variant of "adding all executables from globally activated packages to PATH".

jonasfj:
Let's ship this, the goal was the help Dart 3 migration -- and we shipped Dart 3 a while ago :rofl:

PR #4887, opened 2023-05-15
Clarify switch expression usage

### LINKED ISSUE #4865: Clarify switch expression usage
https://github.com/dart-lang/language/issues/3061

https://dart.dev/language/branches#switch-expressions

~Change example to:~
```dart
enum Color { red, blue, green }

void main() {
  final color = Color.red;
  (switch (color) {
    Color.red => "red",
    Color.blue => "blue",
    Color.green => "green",
  });
}
```

The solution is to clarify that they can't be used as the beginning of an expression. Then, if they really need to do that, parenthesize it like the above example. But I think instead of parenthesizing it, these cases should really just be using a switch _statement_. I'm hesitant to mention the parenthesized syntax since it's a work around to achieve behavior that's already built into switch statements.

MaryaBelanger:
I don't think "return a value into some expression statement" in lines 136 and 137 _exactly_ makes sense:

> If your switch doesn't need to return a value into some expression statement,
use a [switch _statement_](#switch-statements) instead of an expression.

I could also just say "if you need to start an expression statement with a switch, use a switch statement" if that's clearer. This seems to be where the confusion is.

Fixes #4865 

Didn't change the bullet list, just moved it below the example since it's kind of weird to list out the syntax differences without having shown an example yet.

### DIFF
--- a/src/language/branches.md
+++ b/src/language/branches.md
@@ -131,10 +131,18 @@ check out the patterns documentation on [Switch statements and expressions][].
 
 A `switch` _expression_ produces a value based on the expression
 body of whichever case matches. 
-You can use a switch expression wherever Dart allows expressions,
-_except_ at the start of an expression statement.
-Not all switch statments need to return a value into some expression statement.
-If yours doesn't, use a [switch _statement_](#switch-statements) rather than an expression.
+You can use a switch expression wherever Dart allows expressions*, for example:
+
+```dart
+var x = switch (y) { ... };
+
+print(switch (x) { ... };
+
+return switch (x) { ... };
+```
+
+*You _can't_ use a switch expression at the start of an expression statement.
+Instead, use a [switch statement](#switch-statements).
 
 Switch expressions allow you to rewrite a switch _statement_ like this:
 
--- a/src/language/branches.md
+++ b/src/language/branches.md
@@ -133,8 +133,8 @@ A `switch` _expression_ produces a value based on the expression
 body of whichever case matches. 
 You can use a switch expression wherever Dart allows expressions,
 _except_ at the start of an expression statement.
-If your switch doesn't need to return a value into some expression statement,
-use a [switch _statement_](#switch-statements) instead of an expression.
+Not all switch statments need to return a value into some expression statement.
+If yours doesn't, use a [switch _statement_](#switch-statements) rather than an expression.
 
 Switch expressions allow you to rewrite a switch _statement_ like this:
 
--- a/src/language/branches.md
+++ b/src/language/branches.md
@@ -129,20 +129,14 @@ check out the patterns documentation on [Switch statements and expressions][].
 
 ### Switch expressions
 
-A `switch` expression is similar to a `switch` statement, but you can use them
-anywhere you can use an expression. They produce a value based on the expression
-body of whichever case matches.
+A `switch` _expression_ produces a value based on the expression
+body of whichever case matches. 
+You can use a switch expression wherever Dart allows expressions,
+_except_ at the start of an expression statement.
+If your switch doesn't need to return a value into some expression statement,
+use a [switch _statement_](#switch-statements) instead of an expression.
 
-The syntax of a `switch` expression differs from `switch` statement syntax:
-
-- Cases _do not_ start with the `case` keyword.
-- A case body is a single expression instead of a series of statements.
-- Each case must have a body; there is no implicit fallthrough for empty cases.
-- Case patterns are separated from their bodies using `=>` instead of `:`.
-- Cases are separated by `,` (and an optional trailing `,` is allowed).
-- Default cases can _only_ use `_`, instead of allowing both `default` and `_`.
-
-Switch expressions allow you to rewrite a _statement_ like this:
+Switch expressions allow you to rewrite a switch _statement_ like this:
 
 ```dart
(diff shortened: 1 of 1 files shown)


src/language/branches.md
```diff
@@ -129,20 +129,14 @@ check out the patterns documentation on [Switch statements and expressions][].
...
 
-A `switch` expression is similar to a `switch` statement, but you can use them
-anywhere you can use an expression. They produce a value based on the expression
-body of whichever case matches.
+A `switch` _expression_ produces a value based on the expression
+body of whichever case matches. 
+You can use a switch expression wherever Dart allows expressions,
+_except_ at the start of an expression statement.
+If your switch doesn't need to return a value into some expression statement,
+use a [switch _statement_](#switch-statements) instead of an expression.
```
atsansone, line 137: **nit:** This is a little long, but could go as is.
```suggestion
Not all switch statments need to return a value into some expression statement.
If yours doesn't, use a [switch _statement_](#switch-statements) rather than an expression.
```

atsansone (approved):
LGTM: One nit you can take or leave.

src/language/branches.md
MaryaBelanger: Thanks Tony! Sounds good

src/language/branches.md
```diff
@@ -129,20 +129,14 @@ check out the patterns documentation on [Switch statements and expressions][].
...
 ### Switch expressions
 
-A `switch` expression is similar to a `switch` statement, but you can use them
-anywhere you can use an expression. They produce a value based on the expression
-body of whichever case matches.
+A `switch` _expression_ produces a value based on the expression
+body of whichever case matches. 
+You can use a switch expression wherever Dart allows expressions,
+_except_ at the start of an expression statement.
+Not all switch statments need to return a value into some expression statement.
```
munificent, line 136: Yeah, this sentence doesn't quite work. I would maybe do:

---

You can use a switch expression wherever Dart allows expressions. For example:

```dart
// TODO: Real examples...
var x = switch (y) { example code here... };

print(switch (x) { example code ... };

return switch (x) { example code ... };
```

You can't use a switch expression at the start of an expression statement. Instead, use a switch statement.

munificent (approved):

src/language/branches.md
MaryaBelanger: I actually like the not-real examples here, going to leave them empty just quickly show the possible structures (until someone yells at me that that's horrible practice!)

src/language/branches.md
```diff
@@ -129,20 +129,22 @@ check out the patterns documentation on [Switch statements and expressions][].
 
 ### Switch expressions
 
-A `switch` expression is similar to a `switch` statement, but you can use them
-anywhere you can use an expression. They produce a value based on the expression
-body of whichever case matches.
+A `switch` _expression_ produces a value based on the expression
+body of whichever case matches. 
+You can use a switch expression wherever Dart allows expressions*, for example:
```
parlough, line 134: Is this asterisk intentional?

```suggestion
You can use a switch expression wherever Dart allows expressions, for example:
```

src/language/branches.md
```diff
@@ -129,20 +129,22 @@ check out the patterns documentation on [Switch statements and expressions][].
...
-- Case patterns are separated from their bodies using `=>` instead of `:`.
-- Cases are separated by `,` (and an optional trailing `,` is allowed).
-- Default cases can _only_ use `_`, instead of allowing both `default` and `_`.
+print(switch (x) { ... };
+
+return switch (x) { ... };
+```
 
-Switch expressions allow you to rewrite a _statement_ like this:
+*You _can't_ use a switch expression at the start of an expression statement.
```
parlough, line 144: How about this one?

```suggestion
You _can't_ use a switch expression at
 the start of an expression statement.
```

src/language/branches.md
```diff
@@ -129,20 +129,22 @@ check out the patterns documentation on [Switch statements and expressions][].
...
+```dart
+var x = switch (y) { ... };
 
-- Cases _do not_ start with the `case` keyword.
-- A case body is a single expression instead of a series of statements.
-- Each case must have a body; there is no implicit fallthrough for empty cases.
-- Case patterns are separated from their bodies using `=>` instead of `:`.
-- Cases are separated by `,` (and an optional trailing `,` is allowed).
-- Default cases can _only_ use `_`, instead of allowing both `default` and `_`.
+print(switch (x) { ... };
```
parlough, line 139:
```suggestion
print(switch (x) { ... });
```

parlough (approved):
Thanks so much for this. I love these adjustments. It's definitely a lot more clear!

A few minor fixes:

src/language/branches.md
MaryaBelanger: It is 😅 I thought because it says "wherever Dart allows expressions" but then there's a caveat, "not in the beginning of an expression statement". Unless "wherever Dart allows expressions" already doesn't include expression statements?

src/language/branches.md
MaryaBelanger: Yeah that was the caveat, like before it said "this is possible..." and this is like "..._unless_...". Again, if that's confusing to you I might be wrongly assuming that normally expressions _can_ be used at the start of an expression statement, just not _switch_ expressions.

PR #4888, opened 2023-05-15
Fix linter rules release information

parlough:
Generated after the fix in https://github.com/dart-lang/linter/commit/9fac1673f59458e101746707fb279b3a4bb8a41d


atsansone (approved):
LGTM presuming this is autogenerated.

PR #4890, opened 2023-05-16
Update record equality example

### LINKED ISSUE #4874: Update 'Record Equality' section in 'Records'
### Page URL

https://dart.dev/language/records.html

### Page source

https://github.com/dart-lang/site-www/tree/main/src/language/records.md

### Describe the problem

The example in section 'Record Equality', https://dart.dev/language/records#record-equality illustrates the typing properties which were already discussed at the end of the 'Record Syntax' section (in particular, it illustrates that types like `(int a, int b)` and `(int x, int y)` are the same type, but `({int a, int b})` and `({int x, int y})` are different types).

However, the example should be changed to illustrate equality (and possibly hashCode, but certainly equality), because that's the topic of the section.

### Expected fix

For example:

```dart
(int x, int y, int z) point = (1, 2, 3);
(int r, int g, int b) color = (1, 2, 3);

print(point == color); // Prints 'true'.
```

```dart
({int x, int y, int z}) point = (x: 1, y: 2, z: 3);
({int r, int g, int b}) color = (r: 1, g: 2, b: 3);

print(point == color); // Prints 'false'. Lint: Equals on unrelated types.
```

### Additional context

_No response_

MaryaBelanger:
Fixes #4874 

Added `// ignore_for_file: unrelated_type_equality_checks` to the excerpts file to deal with the lint; `point` and `color` aren't subtypes to be compared with `==`

### DIFF
--- a/examples/language/test/records_test.dart
+++ b/examples/language/test/records_test.dart
@@ -1,3 +1,4 @@
+// ignore_for_file: unrelated_type_equality_checks
 import 'package:test/test.dart';
 // import 'package:examples_util/print_matcher.dart' as m;
 
@@ -95,8 +96,7 @@ void main() {
     (int x, int y, int z) point = (1, 2, 3);
     (int r, int g, int b) color = (1, 2, 3);
 
-    // OK:
-    point = color;
+    print(point == color); // Prints 'true'.
     // #enddocregion record-shape
     expect(point, (1, 2, 3));
   });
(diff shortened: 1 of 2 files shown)


examples/language/test/records_test.dart
```diff
@@ -106,8 +106,7 @@ void main() {
     ({int x, int y, int z}) point = (x: 1, y: 2, z: 3);
     ({int r, int g, int b}) color = (r: 1, g: 2, b: 3);
 
-    // Compile error! These record don't have the same shape.
-    // point = color;
+    print(point == color); // Prints 'false'. Lint: Equals on unrelated types.
```
eernstg, line 109: Except that the lint is disabled?

Ah, perhaps the point is that the documentation web page should have the comment (because the documentation web page doesn't say anything about `// ignore...`, and we want to tell the reader that they'd get that lint. And `records_test.dart` should contain the comment because it must be identical (in some sense) to the code that occurs on the documentation web page. So it's all working as intended. ;-)

eernstg (approved):
LGTM

PR #4893, opened 2023-05-17
Add a screenshot to DevTools page

mit-mit:
Adding a screenshot to make it possible to visually get an impression of what this is.


src/tools/dart-devtools.md
```diff
@@ -8,6 +8,8 @@ for Dart and Flutter.
 These tools are distributed as part of the `dart` tool
 and interact with tools such as IDEs, `dart run`, and `webdev`.
 
+<img src="/tools/DevTools.png" width="400">
```
parlough, line 11: Seems a bit small for me. Also can we add some sort of alt text?

```suggestion
<img src="/tools/DevTools.png" width="640" alt="Screenshot of DevTools' Memory page">
```

MaryaBelanger: Agreed, very small

mit-mit:
I think this looks OK now?
https://dart-dev--pr4893-mit-mit-devtools-ojm2t1yu.web.app/tools/dart-devtools

parlough (approved):
That does look better, thanks!

I resized the source image and optimized it as it was quite large.

PR #4895, opened 2023-05-17
Reorder pattern types by precedence

### LINKED ISSUE #4837: [v3.0] pattern precedence is unstated
### Page URL

https://dart.dev/language/pattern-types (v3.0)

### Page source

https://github.com/dart-lang/site-www/blob/main/src/language/pattern-types.md

### Describe the problem

You've alphabetized the patterns, but this destroys any sense of hierarchy, and the precedence matters, which is why we have parens!  If you're going to describe them in alpha order, rather than the precedence order, you'll need a precedence chart at the end or beginning.

### Expected fix

Return to precedence order, or add a precedence chart.

### Additional context

_No response_

MaryaBelanger:
Fixes #4837 

- Reordered pattern-types.md in order of precedence
- Attempted to explain pattern precedence at the top of the page (h4 section to keep it minor)
- Added an example to the Parenthesized type section to elaborate on precedence more.


RandalSchwartz:
it makes a difference.  For example, in the absence of parens, pat1 || pat2 && pat3 will work like a math equation, with the && binding tighter than the ||, which defines order of execution and therefore overall outcome.  But if you don't have a chart to know that && binds tighter than ||, you can get lost about when you need parens and when you don't.

munificent (approved):

src/language/pattern-types.md
```diff
@@ -7,6 +7,97 @@ This page is a reference for the different kinds of patterns.
 For an overview of how patterns work, where you can use them in Dart, and common
 use cases, visit the main [Patterns][] page.
 
+#### Pattern precedence
+
+Similar to [operator precedence](/language/operators#operator-precedence-example),
+pattern evaluation adheres to precendence rules.
```
parlough, line 13:
```suggestion
pattern evaluation adheres to precedence rules.
```

src/language/pattern-types.md
```diff
@@ -7,6 +7,97 @@ This page is a reference for the different kinds of patterns.
...
+You can use [parenthesized patterns](#parenthesized) to evaluate lower-precedence
+patterns first.
```
parlough, line 15:
```suggestion
You can use [parenthesized patterns](#parenthesized) to 
evaluate lower-precedence patterns first.  
```

src/language/pattern-types.md
```diff
@@ -7,6 +7,97 @@ This page is a reference for the different kinds of patterns.
...
+* [Logical-or](#logical-or) patterns are lower-precedence than [logical-and](#logical-and),
+logical-and patterns are lower-precedence than [relational](#relational) patterns,
+and so on. 
+
+* Post-fix unary patterns ([cast](#cast), [null-check](#null-check),
+and [null-assert](#null-assert)) share the same level of precedence. 
+
+* The remaining primary patterns share the highest precedence.
+Collection-type ([record](#record), [list](#list), and [map](#map))
+and [Object](#object) patterns obviously encompass other
```
parlough, line 27: Can we drop the "obviously" here? It almost should never be used.

```suggestion
and [Object](#object) patterns encompass other
```

src/language/pattern-types.md
```diff
@@ -7,6 +7,97 @@ This page is a reference for the different kinds of patterns.
...
 use cases, visit the main [Patterns][] page.
 
+#### Pattern precedence
+
+Similar to [operator precedence](/language/operators#operator-precedence-example),
+pattern evaluation adheres to precendence rules.
+You can use [parenthesized patterns](#parenthesized) to evaluate lower-precedence
+patterns first.  
+
+We've listed the patterns ascending order of precedence:
```
parlough, line 17:
```suggestion
We've listed the patterns ascending order of precedence:

```

parlough (commented):
Thanks for doing this! A big win without too much extra writing.

src/language/pattern-types.md
parlough: Actually, I'm reading this again and I feel some word is missing?

```suggestion
This document lists the pattern types in ascending order of precedence.

```

src/language/pattern-types.md
```diff
@@ -59,6 +204,39 @@ case [a, b]: // ...
 case const [a, b]: // ...
 ```
 
+## Variable
+
+`var bar, String str, final int _`
+
+Variable patterns bind new variables to values that have been matched or destructured. 
+They usually occur as part of a [destructuring pattern][destructure] in order to
```
atsansone, line 212: **nit:** Remove `in order`. `In order to` and `To` mean the same thing.
```suggestion
They usually occur as part of a [destructuring pattern][destructure] to
```

src/language/pattern-types.md
```diff
@@ -59,6 +204,39 @@ case [a, b]: // ...
...
+capture a destructured value.
+
+The variables are in scope in a region of code that is only reachable when the
+pattern has matched.
```
atsansone, line 216: **issue:** Rephrase this. It's a little hard to follow. Maybe two sentences, one that establishes scope and one that covers pattern matching?

src/language/pattern-types.md
```diff
@@ -59,6 +204,39 @@ case [a, b]: // ...
...
+```dart
+switch ((1, 2)) {
+  // 'var a' and 'var b' are variable patterns that bind to 1 and 2, respectively.
+  case (var a, var b): // ...
+  // 'a' and 'b' are in scope in the case body.
+}
+```
+
+A _typed_ variable pattern only matches if the matched value has the declared type,
+and fails otherwise:
```
atsansone, line 228: **question:** So this pattern match matches on type annotation, not variable value?

atsansone (commented):
@MaryaBelanger : Nothing huge, but a few questions and suggestions. Let me know when you need another pass.

src/language/pattern-types.md
```diff
@@ -84,6 +262,31 @@ Identifier patterns may behave like a [constant pattern](#constant) or like a
...
+## Parenthesized
+
+`(subpattern)`
+
+Like parenthesized expressions, parentheses in a pattern let you control
+[pattern precedence](#pattern-precedence) and insert a lower-precedence
+pattern where a higher precedence one is expected.
+
+For example, imagine the boolean constants `x`, `y`, and `z` are 
+equivalent to `true`, `true`, and `false`, respectively:
```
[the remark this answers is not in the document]
MaryaBelanger: That's a great question! I don't actually know. I guess it depends on, in Dart, whether `x` and `true` are _exactly_ the same in every aspect (e.g., the same object, identity-wise), or just substitutable for each other. I'd go with the latter... but I'll find out.

[From our own docs](https://dart.dev/effective-dart/design#do-override-hashcode-if-you-override-), I thought this guidance was kind of telling about what constitutes "equal":

> ...two objects generally only have the same hash code if they are the _exact same object_...
> **Any two objects that are equal must have the same hash code.**

src/language/pattern-types.md
MaryaBelanger: Oh yeah, it should have been "..._in_ ascending...". I like "This document..." much better than "We've...", so thanks !

src/language/pattern-types.md
MaryaBelanger: This section wasn't changed as part of this PR, it was just moved. The reviews for the original PR were pretty stringent so I'm hesitant to make changes now. I also wouldn't know how to break it up since the scope is directly characterized by the match, and that's the point of the sentence.

parlough: I agree, it's better to leave a majority of the focus on this PR on the reordering.

src/language/pattern-types.md
MaryaBelanger: Yup, it's a pattern that declares new variables. So the value going into the declaration only matters in the sense of "can this value bind to this new variable". So in the case of a "regular" variable pattern, the value doesn't really matter because like with any variable declaration you can just write `var a = ` followed by pretty much anything (major generalization, but sufficient for this topic). But, as is the case for any variable declaration, if you annotate the declaration like `int a = `, then the value needs to adhere to the annotation.

src/language/pattern-types.md
```diff
@@ -84,6 +262,31 @@ Identifier patterns may behave like a [constant pattern](#constant) or like a
 - [Wildcard](#wildcard) identifier in any context: matches any value and discards it:
   `case [_, var y, _]: print('The middle element is $y');`
 
+## Parenthesized
+
+`(subpattern)`
+
+Like parenthesized expressions, parentheses in a pattern let you control
```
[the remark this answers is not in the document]
MaryaBelanger: https://github.com/dart-lang/language/blob/main/accepted/future-releases/0546-patterns/feature-specification.md#parenthesized-pattern

src/language/pattern-types.md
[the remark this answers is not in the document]
RandalSchwartz: > That's a great question! I don't actually know. I guess it depends on, in Dart, whether `x` and `true` are _exactly_ the same in every aspect (e.g., the same object, identity-wise), or just substitutable for each other. I'd go with the latter... but I'll find out.

I do believe that there is a singleton for each built-in enum, like true and false, as well as user-defined enums, so saying "equal" and "identical" mean the same thing here.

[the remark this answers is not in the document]
RandalSchwartz: In smalltalk terms, which is borrowed heavily for dart, true always has a well-known OOP, and a variable that is set to true also reads back as the same OOP... that is, an identical object, as well as equals of course.  OOPs are somewhat visible as hashcodes, although I think there's a hashing down from the full oop pointer, although I could be wrong.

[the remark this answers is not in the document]
MaryaBelanger: So, @atsansone it seems like I can use "equals" here, so I'll change that.

@RandalSchwartz if you don't mind me asking, I've never seen "OOP" used to describe single or individual...objects(?) like that. What does it mean in this context? Search results are pretty useless (lots of "smalltalk is the greatest OOP language ever!" 😅 ). Sorry if that's a massive question to answer, and please don't feel like you have to. You've already contributed plenty in the way of explanations, this ones mostly just for-my-information. Thanks!

parlough (approved):
Thanks for those adjustments! Looks good to me :)

PR #4897, opened 2023-05-17
Move logical-or switch content to patterns page

MaryaBelanger:
I left this in a comment on the patterns types page before the actual patterns page was ready, moving it now that everything's live. 

I thought about putting it in branches.md, but I think it's better in the patterns page because it's a little specific of a use case. Plus having it on the patterns page matches how the other uses sections are structured (explains where it can be used, then tells you why you might want to do that).

Added code excerpts too!

### DIFF
--- a/examples/language/lib/patterns/switch.dart
+++ b/examples/language/lib/patterns/switch.dart
@@ -4,6 +4,18 @@ class Rect {
   Rect({required this.width, required this.height});
 }
 
+enum Color { red, yellow, blue, green }
+
+class Square {
+  int size;
+  Square({required this.size});
+}
+
+class Circle {
+  int size;
+  Circle({required this.size});
+}
+
 void main() {
   var number = 1;
   // #docregion constant-pattern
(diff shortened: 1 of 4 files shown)


src/language/patterns.md
```diff
@@ -168,10 +171,33 @@ switch (obj) {
     print('a = $a, b = $b');
 
   default:
+}
 ```
 
-The values that a pattern destructures in a case become local variables.
-Their scope is only within the body of that case.
+<a id="or-pattern-swtich"></a>
```
parlough, line 177:
```suggestion
<a id="or-pattern-switch"></a>
```

src/language/patterns.md
```diff
@@ -168,10 +171,33 @@ switch (obj) {
...
+```dart
+var isPrimary = switch (color) {
+  Color.red || Color.yellow || Color.blue => true,
+  _ => false
+};
+```
+
+Switch statements can have multiple cases share a body without using logical-or
+patterns, but they are still uniquely useful for allowing multiple cases to share
+a guard:
```
parlough, line 192:
```suggestion
Switch statements can have multiple cases share a body
without using logical-or patterns, but they are
still uniquely useful for allowing multiple cases to share a guard:
```

examples/language/lib/patterns/switch.dart
```diff
@@ -45,7 +57,28 @@ void main() {
...
+
+  {
+    final color = Color.red;
+    // #docregion or-share-body
+    var isPrimary = switch (color) {
+      Color.red || Color.yellow || Color.blue => true,
+      _ => false
+    };
+    // #enddocregion or-share-body
+    isPrimary;
```
parlough, line 72: This isn't important, it just feels weird to have a lone variable there to me.

```suggestion
    print(isPrimary);
```

src/language/branches.md
```diff
@@ -265,3 +266,4 @@ rather than exiting the entire switch.
 [any kind of pattern]: /language/pattern-types
 [destructure]: /language/patterns#destructuring
 [section on switch]: /language/patterns#switch-statements-and-expressions
+[logical-or patterns]: /language/patterns#or-pattern-swtich
```
parlough, line 269:
```suggestion
[logical-or patterns]: /language/patterns#or-pattern-switch
```

parlough (approved):
Reads well and makes sense. Thanks!

A fix for the anchor and minor nits:

Issue #4886, opened 2023-05-15
Too much margin above `{{site.alert...}}`


MaryaBelanger:
### Describe the problem

Can we adjust the upper margin around the `{{site.alert.whatever}}` blocks? I think they're too big

<img width="658" alt="image" src="https://github.com/dart-lang/site-www/assets/111139605/b4b29b52-8146-4592-938e-b56a6092a781">

See how the margin between every other section is more narrow and consistent across different areas (paragraph to paragraph, after a heading, before the code block, etc.)

### Expected fix

I haven't checked if we have a custom CSS file yet, I imagine we'd just adjust that a bit.

PR #4898, opened 2023-05-17
Reduce margin above site alerts

parlough:
Thanks for catching this! The increased consistency with this change looks fine in most spots :D

Closes https://github.com/dart-lang/site-www/issues/4886


MaryaBelanger (approved):
PERFECT! Thank you!

PR #4909, opened 2023-05-22
Cleanup and standardize Effective Dart location and headers

parlough:
- Moves to `/effective-dart` to match the relatively recent change to the top-level. I've been liking that location so thought its path should match, like we've done with `/language`. **!! Let me know** if you think `/language/effective-dart` would be better though.
- Removes periods from headings of Effective Dart rules.
  - It is inconsistent with headings elsewhere on the site and our newest rules didn't follow that scheme.
  - Update the tool to check and fail if one ends in a period
  - Also update the tool to add a period when generating the Effective Dart TOC
- Updates the Effective Dart TOC tool to the latest dependencies and Dart 3
  - Also adds `prefer_final_locals` while I'm there :)


atsansone (approved):
@parlough : I love this. I'd like @MaryaBelanger to sign off, though, too, as she's leading the IA changes.

parlough:
Thanks for the reviews!

Issue #4912, opened 2023-05-23
[PAGE ISSUE]: 'Concurrency in Dart'


sereksim:
### Page URL

https://dart.dev/language/concurrency/

### Page source

https://github.com/dart-lang/site-www/tree/main/src/language/concurrency/index.md

### Describe the problem

At the beginning of the section "Code Examples", the link to [Flutter’s compute() function - https://api.flutter.dev/flutter/foundation/compute-constant.html](https://api.flutter.dev/flutter/foundation/compute-constant.html) does not work:
![grafik](https://github.com/dart-lang/site-www/assets/74626816/11680dff-9666-4034-80f9-864c8a245d3d)

### Expected fix

The link should propably lead to https://api.flutter.dev/flutter/foundation/compute.html

Issue #4789, opened 2023-04-26
Clarify `compute` function limits on web in 'Concurrency in Dart'


parlough:
### Page URL

https://dart.dev/language/concurrency/

### Page source

https://github.com/dart-lang/site-www/tree/main/src/language/concurrency/index.md

### Describe the problem

The note about `compute` doesn't indicate the limitations of the `compute` function in web, in that it doesn't run in a separate isolate, but rather in the event loop of the current isolate.

### Expected fix

Either add a mention of this and add link to the concurrency in web section, or just remove this note completely.

PR #4913, opened 2023-05-23
Fix Flutter compute link and mention

parlough:
The goal is to make it clear that `compute` doesn't run on a different isolate or thread when on the web. Also removes the wording "consider" since I don't think it's worth pushing users towards.

Fixes https://github.com/dart-lang/site-www/issues/4789
Fixes https://github.com/dart-lang/site-www/issues/4912

Staged: https://dart-dev--pr4913-fix-flutter-compute-pt3d3d91.web.app/language/concurrency#implementing-a-simple-worker-isolate

### DIFF
--- a/src/language/concurrency/index.md
+++ b/src/language/concurrency/index.md
@@ -318,7 +318,7 @@ setting up and managing worker isolates:
   instead of `Isolate.run()`.
   On the [web](#web), the `compute` function falls back
   to running the specified function on the current event loop.
-  Use `Isolate.run()` when targeting native platforms only
+  Use `Isolate.run()` when targeting native platforms only,
   for a more ergonomic API.
 {{site.alert.end}}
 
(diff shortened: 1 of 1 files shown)


src/language/concurrency/index.md
```diff
@@ -314,17 +314,16 @@ setting up and managing worker isolates:
...
   If you're using Flutter,
-  consider using [Flutter's `compute()` function][]
+  you can use [Flutter's `compute` function][]
   instead of `Isolate.run()`.
-  The `compute` function
-  allows your code to work  on both
-  [native and non-native platforms][].
+  On the [web](#web), the `compute` function falls back
+  to running the specified function on the current event loop.
   Use `Isolate.run()` when targeting native platforms only
```
MaryaBelanger, line 321: Unrelated but could you toss a comma in there for me:
```suggestion
  Use `Isolate.run()` when targeting native platforms only,
```

MaryaBelanger (approved):
Better, thank you!

src/language/concurrency/index.md
parlough: Just for you!

PR #4915, opened 2023-05-23
Add redirect for pinned SDK packages

devoncarew:
- add a go link to reference documentation for Flutter pinned packages (links to https://github.com/dart-lang/sdk/wiki/Flutter-Pinned-Packages)

This adds a redirect for `https://dart.dev/go/sdk-version-pinning` (some of our tools are now emitting this link).


parlough:
I'm fine with this destination for now, but it seems like we should be documenting something on dart.dev or docs.flutter.dev for this?

parlough (approved):

devoncarew:
> I'm fine with this destination for now, but it seems like we should be documenting something on dart.dev or docs.flutter.dev for this?

That sounds fine to me - something on pub.dev or dart.dev perhaps? Getting a redirect in place - even to a wiki initially - gives us a flexibility wrt changing where we document it going forward.

cc @sigurdm

PR #4917, opened 2023-05-23
Update linter rules for new rules and accurate Effective Dart links

parlough:
This is generated in `dart-lang/linter` and the contents don't need to be reviewed. The main changes here:

- Updates to use the new `dart.dev/effective-dart` links
- Updates some links to Effective Dart to link to the actual related rule rather than just the doc
- Includes docs for the new `no_self_assignments` rule


MaryaBelanger (approved):
Thank you!

PR #4922, opened 2023-05-24
Update patterns.md to fix example

praxder:
Update to the example so that it's accurate and compiles. Previously, it would not compile since `a` and `b` were being treated as variables, but didn't have a type, `var`, or `final` keyword prefixing it. I updated the example to match a constant 'a' and 'b', like the pre-existing comment describes. Tested on dartpad.dev.

---

#### Contribution guidelines:

- See our [contributor guide](https://github.com/dart-lang/.github/blob/main/CONTRIBUTING.md) for general expectations for PRs.
- Contributions to our repos should follow the [Dart style guide](https://dart.dev/guides/language/effective-dart) and use `dart format`.
- Most changes should add an entry to the changelog and may need to [rev the pubspec package version](https://github.com/dart-lang/sdk/wiki/External-Package-Maintenance#making-a-change).
- Changes to packages require [corresponding tests](https://github.com/dart-lang/.github/blob/main/CONTRIBUTING.md#Testing).

(note that many Dart repos have a weekly cadence for reviewing PRs - please allow for some latency before initial review feedback)

### DIFF
--- a/src/language/patterns.md
+++ b/src/language/patterns.md
@@ -60,10 +60,10 @@ For example, the individual fields of any [collection-type][] pattern could be
 <?code-excerpt "language/lib/patterns/switch.dart (list-pattern)"?>
 ```dart
 switch (obj) {
-  // List pattern [a, b] matches obj first if obj is a list with two fields,
+  // List pattern ['a', 'b'] matches obj first if obj is a list with two fields,
   // then if its fields match the constant subpatterns 'a' and 'b'.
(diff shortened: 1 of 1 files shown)


praxder:
Closing because I realize it's a separate docs issue, after reading through the Dart docs more for how Patterns work. According to [this](https://dart.dev/language/pattern-types#identifier) page, it should be working, but isn't, so it must be a different problem.

Issue #4921, opened 2023-05-24
Fix Matching example on 'Patterns' page


praxder:
### Page URL

https://dart.dev/language/patterns.html

### Page source

https://github.com/dart-lang/site-www/tree/main/src/language/patterns.md

### Describe the problem

On the "Matching" section, this example is given:
```dart
switch (obj) {
  // List pattern [a, b] matches obj first if obj is a list with two fields,
  // then if its fields match the constant subpatterns 'a' and 'b'.
  case [a, b]:
    print('$a, $b');
}
```
But that doesn't compile. I'm not sure why it doesn't compile, since [the docs](https://dart.dev/language/pattern-types#identifier) for the `identifier` type of pattern matching does show this as valid syntax. The example simply doesn't work on dartpad.dev.

### Expected fix

Not sure.

### Additional context

Tested on dartpad.dev with Dart 3.0.1.

MaryaBelanger:
Thanks for checking on this @praxder 

Can you show us what you're getting from dartpad? I've tried and it's compiling. 

![image](https://github.com/dart-lang/site-www/assets/111139605/3fc28f89-f7fd-4cdf-8a94-e27f3b005a9b)

For reference, that code snippet is built in a dart file and added to the doc page. You can find the source [here](https://github.com/dart-lang/site-www/blob/main/examples/language/lib/patterns/switch.dart#L29) (lines 29-38).

On the doc page, it leaves out the declaration of `a` and `b` as constants, but the comment is meant to make up for showing that declaration:

>   // List pattern [a, b] matches obj first if obj is a list with two fields,
>   // then if its fields match **_the constant subpatterns 'a' and 'b'_.**

praxder:
Ah, that makes sense, I was mis-interpreting that another way. In that case, this is working fine. Thanks for clarifying!

Closing this issue.

MaryaBelanger:
No worries at all! It seems that having the actual constant declarations visible in the example code might've helped you avoid this confusion (correct me if I'm wrong). That's valuable feedback; we can add those to help other users who might run into the same question. If it was something else, or just something about the wording leading up to the example, let me know that too. Every detail matters!

praxder:
Sounds good. I went ahead and opened another PR with those changed. 
https://github.com/dart-lang/site-www/pull/4924

parlough:
Looks like https://github.com/dart-lang/site-www/pull/4924 handled the concerns here. Thank you!

PR #4924, opened 2023-05-24
Patch 3

praxder:
This PR expands the docs for the `identifier` type of pattern matching to include extra, helpful info.

### DIFF
--- a/examples/language/lib/patterns/switch.dart
+++ b/examples/language/lib/patterns/switch.dart
@@ -26,10 +26,10 @@ void main() {
   }
   // #enddocregion constant-pattern
 
+  var obj = Object();
   // #docregion list-pattern
   const a = 'a';
   const b = 'b';
-  var obj = Object();
   switch (obj) {
     // List pattern [a, b] matches obj first if obj is a list with two fields,
     // then if its fields match the constant subpatterns 'a' and 'b'.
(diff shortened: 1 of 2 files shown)


praxder:
Note: this is in reference to issue: https://github.com/dart-lang/site-www/issues/4921

examples/language/lib/patterns/switch.dart
```diff
@@ -26,10 +26,10 @@ void main() {
   }
   // #enddocregion constant-pattern
 
+  // #docregion list-pattern
   const a = 'a';
   const b = 'b';
   var obj = Object();
```
MaryaBelanger, line 32: Thanks for doing this! Could you move the `obj` declaration line outside of the `docregion`? That part is pretty arbitrary.

```suggestion
  var obj = Object();
  // #docregion list-pattern
  const a = 'a';
  const b = 'b';
```

praxder: Sounds good. Would you also like me to remove it from the 'src/language/patterns.md' file too?

parlough: Thanks! Yes please :)

The markdown snippet is matched against the `.dart` file to make sure they match.

praxder: done. :)

MaryaBelanger (approved):
Thank you! I'll take care of the build failures, but otherwise this is as good as landed

Issue #4657, opened 2023-03-13
Implement automatically generated prev -> next system for new language tour


parlough:
### Describe the problem

We want to maintain some sort of flow for the new language content and should be able to automatically generate the prev->next breadcrumbs for each page.

PR #4925, opened 2023-05-25
Add previous and next links to language pages

parlough:
This will provide a reference of correctness in the future once I eventually get to implementing automatic generation of these, while benefiting users in the mean time.

Contributes to https://github.com/dart-lang/site-www/issues/4657

Example of previous and next links


MaryaBelanger:
This is AWESOME

atsansone (approved):
LGTM, but it omits null safety.

parlough:
That's on purpose, since those null safety docs are less about learning the language and more about migrating. Will add it in to the flow once we improve our standalone null safety documentation.

PR #4926, opened 2023-05-25
Rename `/language/enum` page to `/language/enums`

parlough:
This makes it consistent with our other entries like `classes.md`, `mixins.md`, etc.


MaryaBelanger (approved):
Makes sense. Just want to check if you think continuing to call the page "Enum types" in the side nav is the best choice, or if you think changing that to just "Enums" would make sense too? The page title itself is "Enumerated types" so I don't feel strongly either way.

parlough:
That's fair. I think people will be most familiar with "Enums" so how about we try that for the short-title?

PR #4927, opened 2023-05-25
Make stage workflow not require test workflow

parlough:
`stage` doesn't need `test` to finish as the site build doesn't need the Dart code to be tested. Removing this dependency allows for automatic staging to occur faster after opening a pull request and link checking to occur whether the initial tests pass or not.


MaryaBelanger (approved):
Nice!

Issue #4908, opened 2023-05-22
Introduce expandable subset code snippets


parlough:
### Describe the problem

In many code snippets, we include code that isn't necessarily runnable on its own, such as code that belongs in a function. That allows us to focus on the important piece of the snippet. However, sometimes users expect to be able to run the snippet or try it out, but without a surrounding function or sometimes other code, it isn't possible. The resulting diagnostics or different behavior can be confusion.

### Expected fix

We should introduce a mechanism which allows us to focus on a certain portion of a snippet, but allow it to be expanded to be runnable. This would also make it possible to integrate with the dart analyzer highlighter and make every snippet runnable.

MaryaBelanger:
Was just about to create a new issue, but found this! My issue was going to say, we could link in the snippets to the source (based on[ this comment ](https://github.com/dart-lang/site-www/issues/4938#issuecomment-1569873778)). 

I'm going to paste here what I was writing in the new issue, but lmk if you want to move it to it's own issue since it's a _slightly_ different idea:

---

Users aren't seeing that, e.g. variables are declared a certain way that makes the example possible, because those variables are declared outside of the `docregion` that we're trying to focus on. 

Also, we've had a couple instances of contributors editing code snippets when they come across these issues, which is _amazing_ and we want to encourage and make as easy as possible. But they understandably miss that there's also a code excerpt to update as well. 

It's not _that_ big a deal, but it'd be a smoother experience for consumers _and_ contributors of our documentation if that were obvious.

I'm imagining something like how GitHub Gists look in Medium articles:

![image](https://github.com/dart-lang/site-www/assets/111139605/da1a6ad0-cb6b-4520-8f1e-7e9cb25bc80e)

Could we add a link to the source file in the already existing margin of the snippets? 

![image](https://github.com/dart-lang/site-www/assets/111139605/561deecc-5558-4b30-ab6f-e63ee9b94c9c)

We already have a similar concept of "something in the margin" of some code snippets in practice:

![image](https://github.com/dart-lang/site-www/assets/111139605/48d7e5c0-1f76-48d4-a45c-2f0d007166da)

It'd be great if the link could:
- automatically populate just from the `<code-excerpt ...>` tag in the md
- link not only to the source code file, but the exact doc region

MaryaBelanger:
Expanding is an interesting solution too. It would solve the problem for consumers of the docs, but I don't think it would solve the problem for contributors who are unaware of the source file they need to also update when editing snippets. 

[...]

PR #4940, opened 2023-05-31
Clarify list pattern snippet

### LINKED ISSUE #4938: Update `switch` example on 'Pattern types' page
### Page URL

https://dart.dev/language/pattern-types.html

### Page source

https://github.com/dart-lang/site-www/tree/main/src/language/pattern-types.md

### Describe the problem

List section has code snippet that doesn't compile:
```dart
switch (obj) {
  // Matches if obj is a list with two elements.
  case [a, b]: // ...
}
```

### Expected fix

This should probably be replaced with 
```dart
switch (obj) {
  // Matches if obj is a list with two elements.
  case [final a, final b]: // ...
}
```

### Additional context

_No response_

MaryaBelanger:
Fixes #4938 

Instead of updating the snippet in `pattern_types.dart`, I just re-used the `docregion` from `switch.dart` that clarifies `a` and `b` are constants. So then I just got rid of the `list` `docregion` from `pattern_types.dart` because it's useless. 

(Granted, these types of fixes are just temporary solutions before we find a better way to indicate the surrounding context of a code snippet so users don't run into these issues anymore - see #4908)

### DIFF
--- a/examples/language/lib/patterns/pattern_types.dart
+++ b/examples/language/lib/patterns/pattern_types.dart
@@ -81,15 +81,6 @@ void miscDeclAnalyzedButNotTested() {
   }
 
   {
-    // #docregion list
-    switch (obj) {
-      // Matches if obj is a list with two elements.
-      case [a, b]: // ...
-    }
-    // #enddocregion list
-  }
-
-  {
     // #docregion rest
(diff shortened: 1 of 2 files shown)


atsansone (approved):
LGTM

Issue #4585, opened 2023-02-08
Upcoming deprecations


MaryaBelanger:
[Changelog 3.0](https://github.com/dart-lang/sdk/blob/main/CHANGELOG.md#300---2023-05-10)

atsansone:
@MaryaBelanger : Can you add any detail to this issue?

MaryaBelanger:
I opened this as a place holder, but did not get around to checking on them. I believe others have been updating minor deprecation mentions gradually, so there might be nothing left to fix. 

What I would still like to do is just what I did for the last release: comb through the changelog for any mention of any deprecations, then search the site for any mention of the deprecated item. 

Since these are usually API deprecations, and dart.dev doesn't get into specific APIs too much, there's probably not much to fix, but I like the peach of mind from checking anyway!

PR #4950, opened 2023-06-02
Account for Dart 3 deprecations and removals

parlough:
I went through and looked for instances of all of the deprecations and removals in Dart 2.18, 2.19, and 3.

The only I could find was using the removed `onError` named argument in a symbols example that isn't even currently used.

Closes https://github.com/dart-lang/site-www/issues/4585

### DIFF
--- a/examples/misc/lib/language_tour/built_in_types.dart
+++ b/examples/misc/lib/language_tour/built_in_types.dart
@@ -203,20 +203,12 @@ var s2 = """This is also a
 multi-line string.""";
 // #enddocregion triple-quotes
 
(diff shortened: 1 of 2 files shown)


MaryaBelanger (approved):
Huh, that's weird (the part that's commented out, not your part). Thanks for checking and closing that dart 3 work!

parlough:
Yeah I didn't look into why it's commented out yet, but `Function.apply` does fight against the static nature of modern Dart, and symbols are a bit more of a niche feature, so maybe the example didn't make sense in this context.

PR #4952, opened 2023-06-04
Update typedefs.md, changed function typedef return type

PickledPenguins:
The last function typedef example declares an int as the return type. This seemed silly to me since the inputs of the function are parameter types, implying it is intended to be a generic function typedef.  For my edit, I've replaced the "int" with "T" since that appears to be have been the intent.

- [ x] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

<details>

- See our [contributor guide](https://github.com/dart-lang/.github/blob/main/CONTRIBUTING.md) for general expectations for PRs.
- Contributions to our repos should follow the [Dart style guide](https://dart.dev/guides/language/effective-dart) and use `dart format`.
- Most changes should add an entry to the changelog and may need to [rev the pubspec package version](https://github.com/dart-lang/sdk/wiki/External-Package-Maintenance#making-a-change).
- Changes to packages require [corresponding tests](https://github.com/dart-lang/.github/blob/main/CONTRIBUTING.md#Testing).

Note that many Dart repos have a weekly cadence for reviewing PRs - please allow for some latency before initial review feedback.
</details>


parlough (commented):
Hi @PickledPenguins, thanks for the pull request!

I can see where the confusion can rise, but I believe this is intentional as compare/sort functions usually return an integer to allow for comparisons. As the `Comparable` [`compareTo`](https://api.dart.dev/stable/3.0.3/dart-core/Comparable/compareTo.html) method documentation outlines:

> That is, it returns a negative integer if this is ordered before other, a positive integer if this is ordered after other, and zero if this and other are ordered together.

If they returned something with the same type as the parameters (`T`), it would no longer be a comparison.

It's not really core to the example, so I'm not sure if it's necessary but do you think a small inline comment would be helpful? Or perhaps a different example?

Thanks again! Let us know what you think :)

parlough:
I'm going to close this following my explanation in https://github.com/dart-lang/site-www/pull/4952#pullrequestreview-1465848814. If you think we should adjust the example or add some further comments, please comment here, open a new PR, or open an issue. Thanks so much.

And thanks again for the pull request!

PR #4954, opened 2023-06-05
"Library package" to just "package" or "regular package"

### LINKED ISSUE #4944: [PAGE ISSUE]: 'Glossary of package terms' - "library package" term is misleading
### Page URL

https://dart.dev/tools/pub/glossary.html

### Page source

https://github.com/dart-lang/site-www/tree/main/src/tools/pub/glossary.md

### Describe the problem

The term "Library package" is misleading imo, see https://github.com/dart-lang/language/issues/3123. I think we should come up with a different term (I would just call it a package).

In Dart a Library and a Package are two different things, and we are conflating the meanings with terms like this. A library is a single _compilation unit_, which is made up of a single primary file and then some number of parts, and has its own private scope.

A package is a collection of libraries under a directory, with a pubspec.yaml in the root of that directory (technically a pubspec.yaml is not actually required but for essentially all external users this would be how they make a package, so it is probably fine to omit this detail).

Packages also have a directory under which they can put their "public libraries", which is the `lib` directory (again technically doesn't have to be `lib/`, but does be for any normal developer, so probably not worth mentioning).

### Expected fix

My suggestion would be to change the term "Library package" to "Pub package", or possibly just "package", or some other term. I chose "Pub package" because typically a shared package would be published to pub. But you could also have your own non-published packages that are not apps, so it isn't a perfect term.

If we just called these a "package", then we would have the special "app package" term still, which would just be a special type of package that isn't intended to be depended on by other packages.

I think it would also be good to have a dedicated doc explaining more concretely exactly the "library" term and the "package" term.

### Additional context

_No response_

MaryaBelanger:
Fixes #4944 

It seems like "library package" was a phrase meant to distinguish between what I would consider a standard or "regular" Dart package and an application package. Since application packages seem like the outlier, I think it's fair to just refer to them as "packages" when talking about them, and "regular packages" when comparing them to application packages.

- This PR focuses on [rewriting the pub glossary entry for library package](https://dart-dev--pr4954-lib-pkg-qximd446.web.app/tools/pub/glossary#package).
- I also added a ["library" definition](https://dart-dev--pr4954-lib-pkg-qximd446.web.app/tools/pub/glossary#library) to accompany it. 
- I also corrected any other instances and redirect any links across the site to the new section; I think I caught everything.

Maybe fixes https://github.com/dart-lang/language/issues/3123 too?


src/_guides/libraries/create-library-packages.md
```diff
@@ -1,13 +1,11 @@
 ---
```
jakemac53, line 1: Note the file name here ideally would change also. But if that will break links and things we would want a redirect. If we can't do that it might not be worth changing.

src/_guides/libraries/create-library-packages.md
```diff
@@ -1,13 +1,11 @@
...
-description: Learn how to create library packages in Dart.
+description: Learn how to create packages in Dart.
 ---
 
 The Dart ecosystem uses [packages](/guides/packages)
 to share software such as libraries and tools.
-This page tells you how to create a package,
-with a focus on the most common kind of package,
-[_library_ packages](/tools/pub/glossary#library-package).
+This page tells you how to create the most common kind of [package](/tools/pub/glossary#package).
```
jakemac53, line 8: nit: I would actually guess this is _not_ the most common type of package. There are probably many more apps out there than there are shared packages, and every app has to be a package (technically not true but true enough).

Apps might create multiple packages just for organization but I get the sense that they don't typically do that.

src/language/libraries.md
```diff
@@ -11,7 +11,7 @@ nextpage:
 ---
 
 The `import` and `library` directives can help you create a
-modular and shareable code base. Libraries not only provide APIs, but
+modular and shareable code base. [Libraries][] not only provide APIs, but
 are a unit of privacy: identifiers that start with an underscore (`_`)
 are visible only inside the library. *Every Dart app is a library*, even
```
jakemac53, line 16:
```suggestion
are visible only inside the library. *Every Dart file is a library*, even
```

Technically its a bit more complicated that this description also, because of part files, but I don't know how to simplify this and the distinction is probably not all that important.

The previous "app" description here suggests that the _entire app_ is one library though imo.

src/tools/pub/cmd/pub-upgrade.md
```diff
@@ -34,7 +34,7 @@ When `dart pub upgrade` upgrades dependency versions, it writes a lockfile to en
 dependencies. For application packages, check in the lockfile to
 source control; this ensures the application has the exact same
 versions of all dependencies for all developers and when deployed to
-production. For library packages, don't check in the lockfile,
+production. For regular packages, don't check in the lockfile,
 because libraries are expected to work with a range of dependency versions.
```
jakemac53, line 38:
```suggestion
because packages are expected to work with a range of dependency versions.
```

src/tools/pub/glossary.md
```diff
@@ -145,34 +144,12 @@ A [dependency](#dependency) that your package directly uses itself. The
...
-Since [semantic versioning](https://semver.org/spec/v2.0.0-rc.1.html) requires
-that libraries increment their major version numbers for any backwards
-incompatible changes, library packages will usually require their dependencies'
-versions to be greater than or equal to the versions that were tested and less
-than the next major version. So if your library depended on the (fictional)
-`transmogrify` package and you tested it at version 1.2.1, your version
-constraint would be [`^1.2.1`][].
-
-[`^1.2.1`]: /tools/pub/dependencies#caret-syntax
+[parts]: /language/libraries#importing-only-part-of-a-library
```
jakemac53, line 152: This doc is not describing part files, so I don't think its a good resource for this link. Do we have a better one?

jakemac53 (commented):
Thanks for working on this!

src/language/libraries.md
MaryaBelanger: I could write "*Every Dart file (plus its parts) is a library*".

But I think since "library" will now link to the glossary entry, that now says 

> A library is a single compilation unit, made up of a single primary file _and any optional number of [parts][]_

That kind of covers the caveat. Especially since it's not worth mentioning front and center like you said, I think this is a reasonable compromise to include the info just for people who already want to learn more, so they're clicking the link.

jakemac53: Either way is fine with me, although I probably would lean towards adding the `(plus its parts)` since that is fairly short and more accurate.

src/tools/pub/glossary.md
MaryaBelanger: Thanks for catching that, changed it to https://dart.dev/effective-dart/usage#do-use-strings-in-part-of-directives. Another option is https://dart.dev/tools/diagnostic-messages#part-file but that's kind of sparse. Probably some consolidating that needs to happen there but I'll save that for another PR

jakemac53: Those are both better - probably the latter one is best if it was filled out a bit.

src/tools/pub/glossary.md
```diff
@@ -12,11 +12,10 @@ the [pub tool][].
 
 ## Application package
 
-A package that is not intended to be used as a library. Application packages may
-have [dependencies][] on other packages, but are never depended on
-themselves. They are usually meant to be run directly, either on the command
-line or in a browser. The opposite of an application package is a
-[library package][].
+A package that is not intended to be used as a library, unlike regular [packages][].
```
jakemac53, line 15:
```suggestion
A package that is not intended to be shared, unlike regular [packages][].
```

Individual libraries within an app package _are_ used like libraries - they just aren't intended for use outside the package, so I think just using the term "not intended to be shared" is better here?

jakemac53 (approved):

src/tools/pub/glossary.md
MaryaBelanger: Thanks for explaining! I thought "library" and "application" were mutually exclusive for some reason (probably because I've been reading the same docs that confused our users 😅)

src/tools/pub/glossary.md
```diff
@@ -12,11 +12,10 @@ the [pub tool][].
 
 ## Application package
 
-A package that is not intended to be used as a library. Application packages may
-have [dependencies][] on other packages, but are never depended on
-themselves. They are usually meant to be run directly, either on the command
-line or in a browser. The opposite of an application package is a
-[library package][].
+A package that is not intended to be shared, unlike regular [packages][].
```
mit-mit, line 15: I think I'd start with what this is, rather than what this isn't. Maybe something like:

---

A package that contains a program or app, with a main entrypoint in `bin`. They are meant to be run directly, either on the command line or in a browser. 
 
Application packages may have [dependencies][] on other packages, but are never
depended on themselves.

src/tools/pub/glossary.md
```diff
@@ -12,10 +12,12 @@ the [pub tool][].
 
 ## Application package
 
-A package that is not intended to be shared, unlike regular [packages][].
-Application packages may have [dependencies][] on other packages, but are never
-depended on themselves. They are usually meant to be run directly, either on the
-command line or in a browser. 
+A package that contains a program or app, with a main entrypoint in `/bin`.
```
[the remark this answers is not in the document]
MaryaBelanger: Isn't it under `/lib` for Dart too? I haven't seen `/bin` mentioned at all while working on this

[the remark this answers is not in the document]
jakemac53: If we have a document describing the general term "entrypoint" (which we should, :D), then I would just link that and  end the sentence at `with a main [entrypoint](link)`.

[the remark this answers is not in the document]
jakemac53: LGTM

atsansone (approved):

PR #4956, opened 2023-06-05
[New page] Class modifiers reference

### LINKED ISSUE #4809: [Dart 3] Add class modifiers reference page
Many asks for this, here's one: https://github.com/dart-lang/site-www/pull/4754#issuecomment-1522287240

Possibly might get in before release, otherwise right after

MaryaBelanger:
Fixes #4809 

Pretty much just copied straight from the spec. Did some restructuring as well:

- took the Class modifiers page out from under "Classes & objects"
- made a section "Class modifiers" directly under "Language" for both the main page and the reference to go under
- this is so we can eventually add #4872 under there as well, and keep all the class mod. pages together

I deliberately left some of the very specific invalid mixin combinations out, because I felt like the nature of those "invalid" combinations doesn't go with the rest of the class modifier stuff... 🤷🏻‍♀️


src/_data/side-nav.yml
```diff
@@ -80,14 +80,18 @@
...
-          permalink: /language/class-modifiers
         - title: Enums
           permalink: /language/enums
         - title: Extension methods
           permalink: /language/extension-methods
         - title: Callable objects
           permalink: /language/callable-objects
+    - title: Class modifiers
+      children: 
+        - title: Overview & usage
```
munificent, line 91: Out of curiosity, what style rules do you have about when to use "&" versus "and"?

munificent (approved):
Looks great!

Issue #4955, opened 2023-06-05
New check-links tool issues


MaryaBelanger:
### Describe the problem

Copied from #4900 post-merge conversation:

I'm wondering how to read this output from the new tool (I'm guessing?) The errors are pointing to the `check_link_references.dart`, I'm not sure if that means that it found anything broken.  Ran `make run` and then `tool/check-links.sh`:

```
Resolving dependencies... 
  _fe_analyzer_shared 60.0.0 (61.0.0 available)
  analyzer 5.12.0 (5.13.0 available)
  build_runner_core 7.2.8 (7.2.9 available)
  built_value 8.5.0 (8.6.1 available)
  code_builder 4.4.0 (4.5.0 available)
  collection 1.17.1 (1.17.2 available)
  csslib 0.17.2 (0.17.3 available)
  file 6.1.4 (7.0.0 available)
  glob 2.1.1 (2.1.2 available)
  graphs 2.3.0 (2.3.1 available)
  lints 2.1.0 (2.1.1 available)
  logging 1.1.1 (1.2.0 available)
  matcher 0.12.15 (0.12.16 available)
  test_api 0.5.2 (0.6.0 available)
  watcher 1.0.2 (1.1.0 available)
Got dependencies!
Checking for valid link references...
tool/check_link_references.dart:3:8: Error: Error when reading '/Users/mbelanger/.pub-cache/hosted/pub.dev/path-1.8.3/lib/path.dart': No such file or directory
import 'package:path/path.dart' as path;
       ^
tool/check_link_references.dart:87:26: Error: Method not found: 'extension'.
      .where((p) => path.extension(p) == '.html')) {
                         ^^^^^^^^^
tool/check_link_references.dart:91:30: Error: Method not found: 'relative'.
      invalidReferences[path.relative(filePath, from: directory)] = results;
                             ^^^^^^^^
```

I also tried `tool/test.sh` locally which didn't give me any explicit issues to fix. But then after pushing the commit (https://github.com/dart-lang/site-www/pull/4954), the build/stage CI failed, showing me exactly what links are broken:

```
Starting Firebase emulator async...
Emulator is running
Crawling...

http://localhost:5500/guides/language/coming-from/js-to-dart
- (3660:15) 'avoid us..' => http://localhost:5500/guides/libraries/create-library-packages#organizing-a-library-package (HTTP 200 but missing anchor)

http://localhost:5500/guides/language/coming-from/swift-to-dart
[...]

PR #4959, opened 2023-06-06
Simplify tool pubspec setup

parlough:
We had a separate pubspec in the `/tool` directory, potentially causing issues depending on where you ran `dart pub get` previously. This simplifies to just using the pubspec in the root directory where you'll run related commands.

Once this lands in your branch, **you will** need to rerun `make setup` to get the tool commands working.

Closes https://github.com/dart-lang/site-www/issues/4955


MaryaBelanger (approved):
Thank you!

PR #4960, opened 2023-06-06
Fix some misc broken external links

parlough:
Caught by temporarily enabling external link checks with the linkchecker.

Also ignores Twitter, since it is failing to connect.


MaryaBelanger (approved):
Nice

> Caught by temporarily enabling external link checks with the linkchecker.

Is that something we should be regularly auditing? "regularly" like once or twice a year maybe, but still 😅

Issue #4164, opened 2022-08-08
Update bootstrap to 5.x or remove it


parlough:
### Describe the problem

We currently use Bootstrap 4 across the site, which has now been superseded by Bootstrap 5, and it currently requires we pull in jquery as well. Beyond that, our current use of bootstrap is quite messy, with lots of overrides, and custom styles anyway.

### Expected fix

We should either migrate to Bootstrap 5 if the migration isn't too hard, otherwise consider removing our usage of it completely, as we have a lot of custom styles and overrides anyway.

PR #4963, opened 2023-06-07
Remove unused styles and other cleanup

parlough:
- Removes unused styles
- Improves formatting and spacing
- Moves some code to proper locations
- Removes now unnecessary files

Preparation to make future work towards https://github.com/dart-lang/site-www/issues/4164 easier. Has the added benefit of reducing final CSS size a bit.


parlough:
Thanks for the review!

PR #4974, opened 2023-06-09
Migrate from deprecated Sass `@import` to `@use` and `@forward`

parlough:
`@import` is being deprecated in favor of `@use` in Dart Sass (which we are now using). Read more within Sass's [`@import` documentation](https://sass-lang.com/documentation/at-rules/import).

Contributes to https://github.com/dart-lang/site-www/issues/4164 since it will make that migration easier, with all bootstrap usages namespaced. Also enables better code analysis and auto-completion in IDEs.


parlough:
I did my best to verify this didn't break or change any styles across the site, but would appreciate someone else taking a look through as well. Thanks so much!

domesticmouse (approved):

PR #4975, opened 2023-06-09
No longer include prettify-css

parlough:
We don't use prettify for any important CSS highlighting, and are aiming to move away from prettify completely. This moves us a bit closer while reducing amount of assets needed to download during page load.


atsansone:
@parlough : Makes sense, but it's not the biggest impact to load times. Run [Lighthouse](https://developer.chrome.com/docs/lighthouse/overview/) against the site and you'll see that FontAwesome and jQuery are our longest drags. Our internal CSS could use some updating, too.

atsansone:
There's also [this option in StackOverflow](https://stackoverflow.com/questions/52340070/how-to-reduce-the-size-of-fontawesome-5-3-1).

atsansone (approved):

PR #4979, opened 2023-06-10
Fix incorrect class modifier highlighting in sidenav

parlough:
Our currently activenav system ended up highlighting the overview page since `/language/class-modifiers-for-apis` contains  `/language/class-modifiers`. This is a temporary work around until I finalize an updated system for docs.flutter.dev and here.

**Previously:**
Incorrectly highlighting overview
<br>
**After(fixed):**
Correctly highlighting API doc entry

### DIFF
--- a/src/_data/side-nav.yml
+++ b/src/_data/side-nav.yml
@@ -87,8 +87,9 @@
         - title: Callable objects
           permalink: /language/callable-objects
     - title: Class modifiers
-      children: 
+      children:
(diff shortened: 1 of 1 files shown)


MaryaBelanger (approved):
Interesting, thank you for catching this

parlough:
Yeah, it's more of a bug with the active nav system, but this is quicker than fixing that for now :)

PR #4980, opened 2023-06-10
Fix prev/next information for class modifier pages

parlough:
"Reference" didn't feel like enough information, because I wasn't sure what it is a reference of. This switches to use the full name.

It also fixes the async page's previous to be the reference, which points to it, to allow a consistent back-forward flow.


parlough:
Also this makes me wonder, should the reference be titled "Class modifier reference" instead of "Class modifiers reference"? @MaryaBelanger

MaryaBelanger (approved):
Thank you Parker!

PR #4990, opened 2023-06-13
Add a file listing all released, non-removed linter rules

parlough:
Despite personally not fully understanding why, https://dart-lang.github.io/linter/lints/options/options.html appears to be a relatively(?) popular resource. Some developer's linter workflows include copying all rules, then disabling until they're happy. I believe at least some developers on the Flutter team have done this before.

To still support that workflow and provide a location to redirect to from https://dart-lang.github.io/linter/lints/options/options.html, this PR introduces a similar file on dart.dev.

There is a difference though in that it only lists (non-removed) linter rules that have been released in the latest stable SDK. **Question:** Is that fine? Should we list them even if they're unreleased?

**Staged:** https://dart-dev--pr4990-feature-all-linter-r-96iuxgfx.web.app/tools/linter-rules/all

\cc @pq Do you think there is value in keeping a page like this?
\cc @MaryaBelanger For wording help 🙇  and copy review


domesticmouse (approved):

pq:
> Do you think there is value in keeping a page like this?

It's come up a fair amount and enough folks depend on it so yeah, I think it's worth maintaining.

Thanks!

Issue #4499, opened 2023-01-09
[Tracking] Restructure diagnostic message docs


MaryaBelanger:
There are a lot of small inconsistencies across the diagnostic messages documentation that altogether make the topic's content appear disjointed. For example,

- The way lint messages are written varies slightly from message to message. 
- They differ completely from diagnostic messages, despite really just being two different types of the same thing. 
- Lints docs are built on two separate sites and tools link to each. 
- And other related maintenance issues.

Goal: Restructure the diagnostic message documentation to be more consistent, across messages, pages, and sites. 

This issue is for planning/tracking. Most of the points can be addressed asynchronously. Some of the relevant content and the work described lives outside of this repo.

### 1. Consolidate linter rules with other diagnostic messages
_Decision: TBD / Needs input_

Issue: https://github.com/dart-lang/site-www/issues/4498

### 2. Consistent format and information across diagnostic messages
_Decision: Rewrite individual linter rules to match diagnostic message format_

Issue: https://github.com/dart-lang/linter/issues/3924

The diagnostic messages have a more exact/succinct format to them than the lint messages, so we'll rewrite the lint messages to match that syntax. 

This means using the structure of:
- Single line summary
- Description 
- Example
- Common fixes (instead of GOOD/BAD)

Also need to standardize other features of lint messages not included in diagnostic messages:
- How we link to Effective Dart
- "This rule is available as of...."
- "Rule sets:..."
- "Quick fix available..."

@atsansone I know you had ideas about restructuring or standardizing the layout of individual messages, if you want to share here

### 3. Deprecate github.io/linter
_Decision: repoint all the tools to dart.dev/tools/linter-rules and BE CAREFUL WITH REDIRECTS_

Hosting the lint docs on github.io was decided before the linter rules were added to dart.dev. Now that they're on dart.dev (where they should be, to keep docs in one place), the io page should be deprecated.

A lot of tools point to the io pages though, so the redirects need to be dealt with very carefully.

### 4. Update how to enable lint instructions 
_Decision: page hasn't been reviewed in a while, make sure it's up to date_

https://dart.dev/guides/language/analysis-options

### 5. Create guidance page for which lints to use
_Decision: Add a page to the static analysis set of pages that provides guidance on which lints to enable for your needs_

This will go into detail on the existing common lint sets, and maybe other use cases.

### 6. Standardize process for contributing lint messages 
_Decision: Document the exact format we decide on for linter rules, and a plan for how they are reviewed_

[...]

Issue #4498, opened 2023-01-09
Combine Linter rules and Diagnostic messages pages?


MaryaBelanger:
This is part of the greater effort to [restructure and improve the diagnostic messages documentation](https://github.com/dart-lang/site-www/issues/4499).

Incorporating the [linter rules](https://dart.dev/tools/linter-rules) page into the [diagnostic messages](https://dart.dev/tools/diagnostic-messages) page seems like a good idea because lints are a type of diagnostic message (along with errors and warnings), so it makes sense conceptually join them instead of having a separate, slightly different, standalone page. 

- Keeps all the diagnostics messages in one place, easy to ctrl+f search
- Reduces page clutter. Especially important as we plan to restructure dart.dev 
- Creates a cognitive connection between lints and other diagnostics 

BUT, the potential length of the page is a concern, so we haven't finalized whether to do this.  
- Load time could become a problem. 
  - The number of messages on the diagnostic messages page is already very high. 
  - There are many already-existing diagnostics that still need to be added. 
  - There will always be more created as features develop.
  - Adding the linter messages to that page, that are built from a different source, that will also always continually be added to over time, exacerbates that.
- Scroll length is not the issue. 
  - Both pages are only used as references, where users will likely only ever be linked to the specific message they're concerned with.

@mit-mit, @bwilkerson @pq and I talked about combining the pages, but are concerned with the page length becoming an issue. We wanted to check with you before deciding one way or another, benefit vs. cost, etc.

mit-mit:
We could use a detail page for each lint -- like on https://dart-lang.github.io/linter/lints/ -- and then just have index pages that list them all?

bwilkerson:
> We could use a detail page for each lint ...

In terms of content, the detail pages for lints are effectively equivalent to the individual diagnostic message docs on https://dart.dev/tools/diagnostic-messages. If each lint's detail page is on a separate page, then we should consider putting each non-lint diagnostic's detail page on a separate page, though as was pointed out above, that could cause page clutter.

> ... and then just have index pages that list them all?

I think we need a page that does more than just list them all, I think we need a page (or multiple pages?) that guides users through the process of configuring lints for their package. One possibility that I could imagine is to have the lints (or maybe only those that aren't in either 'core' or 'recommended' grouped in some meaningful way with a brief description of why a user might want to enable it (a brief sales pitch, if you will).

parlough:
[...]

PR #4999, opened 2023-06-17
Implement individual linter rule pages

parlough:
This PR is meant to finalize work to make dart.dev a satisfactory replacement for [`dart-lang.github.io/linter/`](https://dart-lang.github.io/linter/) while `dart-lang/linter` moves to the SDK and we work on consolidating and improving documentation (https://github.com/dart-lang/site-www/issues/4498). 

It does this by (staged links after colon (`:`)):

- Updating the information on dart.dev/lints for the latest state of the linter: [/lints](https://dart-dev--pr4999-feature-linter-indiv-zetj76b0.web.app/tools/linter-rules)
- Making dart.dev/lints an index page rather than including all information: [/lints](https://dart-dev--pr4999-feature-linter-indiv-zetj76b0.web.app/tools/linter-rules)
- Moving the long-form lint rule documentation to individual pages found at `dart.dev/lints/<rule>`: [/lints/avoid_dynamic_calls](https://dart-dev--pr4999-feature-linter-indiv-zetj76b0.web.app/lints/avoid_dynamic_calls)
  - It does this by using a Jekyll page generator based on `linter_rules.json`.
- Adding a textual reference pointing to the recently released all linter rules page (https://github.com/dart-lang/site-www/commit/38973deedb07ef0dfda7e13d5a29306dac1e435e) found at: [/lints/all](https://dart-dev--pr4999-feature-linter-indiv-zetj76b0.web.app/lints/all)
- Updating and links and references for these previous changes (mostly Effective Dart).

Contributes to https://github.com/dart-lang/linter/issues/4460, https://github.com/dart-lang/site-www/issues/4498, https://github.com/dart-lang/site-www/issues/4499

**Note:** This is an intermediate step and doesn't aim to improve the formatting or style of the information on the index or individual pages. That will be part of follow-up work as we work on improvements to the diagnostic linter rule documentation.


bwilkerson (approved):
I don't know the infrastructure being used well enough to catch problems, so if there are areas of concern you'll probably want a review from someone else.

MaryaBelanger:
This is really cool. The new structure and navigation between pages feels fluid and looks great. Clutter is significantly down. 

For my own understanding, are the individual lint pages actually created (like, under `tools/linter-rules` there's an actual `.md` file for each), or is each one just spun up dynamically?

I'm wondering about the URLs. I was expecting the *actual* URL to be `dart.dev/lints/<etc>` and the old `tools/linter-rules/` URLs would redirect to that, but it seems like it's the other way around. I'm assuming it might be too much/too complicated to actually change the file structure around? Or maybe I'm off, could you explain it?

parlough:
@MaryaBelanger 

> For my own understanding, are the individual lint pages actually created (like, under tools/linter-rules there's an actual .md file for each), or is each one just spun up dynamically?

This was a bit tricky to get working properly and I plan to document it as I improve the functionality, especially since I hope to make it generic enough to support diagnostic messages too. The summary is that they are generated automatically during the Jekyll build (currently from the `linter_rules.json`), like they are Markdown files under `tools/linter-rules` (which you'll see in the built site), but you won't see them in your `src` directory.

For a longer explanation: The Jekyll plugin in [`src/_plugins/linter_page_generator.rb`](https://github.com/dart-lang/site-www/pull/4999/files#diff-a982c3bb220ab534b0e1e7987b83a7d3444a6514d9c66a576fe36f6e162b7404) generates each Markdown page with the [layout](https://jekyllrb.com/docs/step-by-step/04-layouts/) [`src/_layouts/linter-rule-standalone.html`](https://github.com/dart-lang/site-www/pull/4999/files#diff-a8f74b0a50a4e6e3372bae279b1db92518965a48dc0fa36cda8dff41f93bfa25), which uses the [`src/_includes/linter-page-content.md`](https://github.com/dart-lang/site-www/pull/4999/files#diff-97c0dc4be66e30473888887eec6488d739be6dfdf2d38efe8516ec1dfe9cd3ca) include to generate the page's content. I had to have a separate layout and include file because I guess layout files have to HTML. I could have done it in the Ruby plugin, but the Markdown file seemed more friendly to edit and review, since not many of us (including me) know Ruby well.

> I'm wondering about the URLs. I was expecting the actual URL to be dart.dev/lints/<etc> and the old tools/linter-rules/ URLs would redirect to that, but it seems like it's the other way around.

I tried to keep it the same as before intentionally, as I feel it still fits within the other `tools` content (especially once/if we enable breadcrumbs), and the redirect is just a link destination we can guarantee will continue to work, no matter where the content actually lives. Let me know if you think this content should be directly under `/lints` though, it's an option, just perhaps with less flexibility for moving in the future.

domesticmouse (approved):

MaryaBelanger:
> Let me know if you think this content should be directly under `/lints`

Nope, your explanation makes sense so let's keep it the way you've set it up so far. 

> For a longer explanation:

Couple more questions regarding this paragraph (but feel free to merge in the meantime). I know this design/plan might not be fully figured out yet, but wondering if it is/what you're thinking:

- For the "justification" content per lint, will that live on these generated pages? And if yes, how so? Will it still be written into the yaml files in the sdk? 
- It sounds like we won't be able to manually adjust these generated pages in this repo -- I'm not sure how important this might be to the plan so I just want to hear your thoughts on it

I'm still catching up on notes from our last meeting / what you worked on while I was out, sorry I keep asking you to re-explain everything! I'll be caught up soon, and will also sync with Brian if these questions are beyond your scope of the project 🤞

MaryaBelanger (approved):

parlough:
> For the "justification" content per lint, will that live on these generated pages? And if yes, how so? Will it still be written into the yaml files in the sdk?

Yes that is what I'm thinking. I don't think we've fully decided where the source justification content will live or in what format, but it sounds like it may be better in the SDK so it can be used there as well. I'm not exactly sure when, but perhaps when you hover a linter rule or whatever. In the meantime, it is still coming from the machine-generated file on `dart-lang/linter`.

> It sounds like we won't be able to manually adjust these generated pages in this repo -- I'm not sure how important this might be to the plan so I just want to hear your thoughts on it

Similar to the previous question, it's not clear where we will store the content eventually. But most likely, the format and layout will continued to be adjustable and controllable in this repo at the very least.

PR #5001, opened 2023-06-20
Folders named packages are no longer ignored by publishing.

sigurdm:
Fixes https://github.com/dart-lang/pub/issues/3948


parlough (approved):
Thanks!

PR #5005, opened 2023-06-21
Fixed double negative

### LINKED ISSUE #5004: Rephrase double negative on 'Effective Dart: Documentation' page
### Page URL

https://dart.dev/effective-dart/documentation.html

### Page source

https://github.com/dart-lang/site-www/tree/main/src/effective-dart/documentation.md

### Describe the problem

[DO format comments like sentences
](https://dart.dev/effective-dart/documentation#do-format-comments-like-sentences)

DO format comments like sentences

`"// Not if there is nothing before it." 
` 
is not a good use of English as it is double negative and I would not use it as an example of a sentence (https://dictionary.cambridge.org/grammar/british-grammar/double-negatives-and-usage)

### Expected fix

`"// Not if there is anything before it." 
` 

### Additional context

_No response_

atsansone:
Fixes #5004

### DIFF
--- a/src/effective-dart/documentation.md
+++ b/src/effective-dart/documentation.md
@@ -32,7 +32,7 @@ generated documentation.
 {:.good}
 <?code-excerpt "docs_good.dart (comments-like-sentences)"?>
 {% prettify dart tag=pre+code %}
-// Not if there is nothing before it.
+// Not if anything comes before it.
 if (_chunks.isEmpty) return false;
 {% endprettify %}
 
--- a/src/effective-dart/documentation.md
+++ b/src/effective-dart/documentation.md
@@ -33,7 +33,7 @@ generated documentation.
 <?code-excerpt "docs_good.dart (comments-like-sentences)"?>
 {% prettify dart tag=pre+code %}
 // Not if anything comes before it.
-if (_chunks.isEmpty) return false;
+if (_chunks.isNotEmpty) return false;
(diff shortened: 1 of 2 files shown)


src/effective-dart/documentation.md
```diff
@@ -32,7 +32,7 @@ generated documentation.
 {:.good}
 <?code-excerpt "docs_good.dart (comments-like-sentences)"?>
 {% prettify dart tag=pre+code %}
-// Not if there is nothing before it.
+// Not if anything comes before it.
```
parlough, line 35: This modified comment seems to no longer match the code sample. It's returning false if nothing comes before it (`_chunks.isEmpty`). The "not" comes from the false, the "nothing" came from the `isEmpty`. 

If you don't like the original comment, I suppose you can change the snippet, perhaps simply:

```
if (_chunks.isNotEmpty) return false;
```

Then the "Not if anything comes before it." will make a bit more sense.

parlough (changes requested):
Thanks for looking into this. Makes sense to try to include a clear sentence.

No matter what the change ends up looking like, you'll need to modify the source file too since this is a code excerpt: https://github.com/dart-lang/site-www/blob/main/examples/misc/lib/effective_dart/docs_good.dart#L19-L20

parlough (approved):
Works for me, thanks!

PR #5006, opened 2023-06-21
Red underlines for diagnostics

parlough:
This PR replaces the previous yellow background used in highlighting on the page with a red-dashed underline.

I tried out wavy text underlining, but the behavior is different between browsers and it resets in each span, resulting in a discontinuous line. 

**Staged:** https://dart-dev--pr5006-misc-try-red-underli-lhe7tgba.web.app/tools/diagnostic-messages

**Example:**

Example with red underline


parlough:
@bwilkerson How is this for a replacement of the yellow highlighting or do you think it is too hard to visually identify?

parlough:
Generator change landed in https://github.com/dart-lang/sdk/commit/5200d086d15132b3e02a09aab560ae66dd329117

MaryaBelanger (approved):
Does this get written into the yaml the same way? `[! ... !]`

I like it, but I think @bwilkerson should approve it too.

parlough:
Yep, it is still written with the `[! ... !]`. The added class just replaces the yellow highlighting with the red underlining.

bwilkerson (approved):

PR #5008, opened 2023-06-23
Rename `funcThatThrows` to `asyncErrorFunction`

### LINKED ISSUE #5007: Issue with whenComplete example on dart.dev documentation website
In the section about `whenComplete` in the article on [Futures and Error handling](https://dart.dev/guides/libraries/futures-error-handling#completing-the-future-returned-by-whencomplete) there is a code example which doesn't work as expected when implemented

```dart
void main() {
  funcThatThrows()
      // Future completes with an error:
      .then((_) => print("Won't reach here"))
      // Future completes with the same error:
      .whenComplete(() => print('Reaches here'))
      // Future completes with the same error:
      .then((_) => print("Won't reach here"))
      // Error is handled here:
      .catchError(handleError);
}
```
The name suggests that `funcThatThrows()` will throw an exception, and could have any return type. But in that case `then()` can't be attached to it. For `then()` to be attached, `funcThatThrows()` has to return a `Future` along with a `throw()`. But in that case `throw()` is not caught and we get a runtime error (not as suggested by comments). To get the output we need `funcThatThrows` to not throw and return `Future.error()` object, which makes it a misnomer.

I think this can be fixed by changing code as follows:
```dart
void main() {
  Future(funcThatThrows)
      ...
```

Code examples with outputs at the end of file:
1. [`funcThatThrows` that doesn't `throw`, but returns `Future.error`](https://github.com/niteshb/dart-bug-reports/blob/master/2022_10_15_doc_future_then/l21_whencomplete_dart_doc_return_future_error.dart) In this case output is as documented, but misnomer issue.
2. [`funcThatThrows` that `throw`s and returns a `Future`](https://github.com/niteshb/dart-bug-reports/blob/master/2022_10_15_doc_future_then/l21_whencomplete_dart_doc_doesnt_work.dart) Return subtype of Future doesn't matter because we end up getting runtime error
3. [Code that works as it is supposed to](https://github.com/niteshb/dart-bug-reports/blob/master/2022_10_15_doc_future_then/l21_whencomplete_dart_doc.dart)

I think doc should be changed to reflect case 3.

Various other examples around this example suffer from similar bugs.

natebosch:
The old name can be confused for a function with a synchronous `throw` instead of a function returning an future error or an `async` function with a `throw`.

Fixes #5007

### DIFF
--- a/examples/futures/lib/early_error_handlers.dart
+++ b/examples/futures/lib/early_error_handlers.dart
@@ -5,9 +5,9 @@ import 'package:futures_examples/util.dart';
 
 // #docregion bad
 void mainBad() {
-  Future<Object> future = funcThatThrows();
+  Future<Object> future = asyncErrorFunc();
 
-  // BAD: Too late to handle funcThatThrows() exception.
+  // BAD: Too late to handle asyncErrorFunc() exception.
   Future.delayed(const Duration(milliseconds: 500), () {
     future.then(ellipsis()).catchError(ellipsis());
   });
@@ -17,7 +17,7 @@ void mainBad() {
 // #docregion good
 void mainGood() {
   Future.delayed(const Duration(milliseconds: 500), () {
-    funcThatThrows().then(ellipsis()).catchError(ellipsis()); // We get here.
+    asyncErrorFunc().then(ellipsis()).catchError(ellipsis()); // We get here.
   });
 }
 // #enddocregion good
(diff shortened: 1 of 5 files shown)


natebosch:
cc @lrhn - can you think of any clear name that doesn't include the abbreviation `Func`?

natebosch:
Maybe `returnsAsyncError` or `returnsFutureError`?

lrhn:
Just `asyncErrorFunction` seems fine.
Just avoid abbreviations, the test is precise and describes the function semantically.

parlough (approved):
Thanks for this! Definitely makes it clearer :)

As a note for myself and others, this is one of our older docs, might be time for us go through it and make sure it is up to our current standards.

PR #5010, opened 2023-06-23
Specify specific Dart versions in Docker file

parlough:
Previously the build broke with every new Dart release, reducing velocity and breaking unrelated PRs. This PR ties the Dockerfile to specific versions which can be updated as desired.

This does limit control from the `make` commands of specifying the dart version outside of the channel (`stable`, `beta`, or `dev`), but I've never used that functionality as you rarely need a specific Dart version. If someone needs to test one, they can adjust the version in the Docker file if they need to for their work.

Closes https://github.com/dart-lang/site-www/issues/4297

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,16 +17,10 @@ RUN echo "alias lla='ls -lAhG --color=auto'" >> ~/.bashrc
 WORKDIR /root
 
 
-# google-chrome-stable
-
-
 # ============== DART ==============
 # See https://github.com/dart-lang/dart-docker
-# See https://github.com/dart-lang/setup-dart/blob/main/setup.sh
 FROM base as dart
-ARG DART_VERSION=latest
 ARG DART_CHANNEL=stable
-ENV DART_VERSION=$DART_VERSION
 ENV DART_CHANNEL=$DART_CHANNEL
 ENV DART_SDK=/usr/lib/dart
 ENV PATH=$DART_SDK/bin:$PATH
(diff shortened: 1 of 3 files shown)


MaryaBelanger:
What's changing in the workflow compared to using checksums? Will we have to manually update `DART_VERSION` the same as the checksums? Or does Docker some how keep track of the correct version and pull that into the Dockerfile on its own?

If `DART_VERSION` is supposed to be manually updated, where are we getting the version numbers from? Is it just absolute latest?

I'm not sure if my questions make sense; I guess I'd just like to hear how this works all around.

parlough:
@MaryaBelanger Sorry that I wasn't super clear.

This route does lose some flexibility, as now the `make fetch-sums` pulls the latest versions of each branch as well as their checksums. The workflow to update the Dockerfile will be the same from our end(`make fetch-sums` then update Dockerfile), the generated text just now contains the versions too, and can be updated whenever you want rather than being required when a release is made. It does this while maintaining the security benefits of verifying the checksums.

We could technically allow specifying versions when getting the checksums or avoid the checksum check when specifying a version manually, but in the two years since this infrastructure was added, we've only wanted to update to the latest releases. So I didn't think it was worth the time adding those pieces of functionality, especially since I want to move away from the Dockerfile as soon as possible. Let me know if you think that functionality would be worthwhile to add though.

Dockerfile
```diff
@@ -17,39 +17,39 @@ RUN echo "alias lla='ls -lAhG --color=auto'" >> ~/.bashrc
...
 ENV DART_CHANNEL=$DART_CHANNEL
 ENV DART_SDK=/usr/lib/dart
 ENV PATH=$DART_SDK/bin:$PATH
 RUN set -eu; \
     case "$(dpkg --print-architecture)_${DART_CHANNEL}" in \
       amd64_stable) \
         DART_SHA256="05fba372d64932dffce90bbd45116b76806bf9adc6203967b56faf5c64b2b66c"; \
-        SDK_ARCH="x64";; \
+        SDK_ARCH="x64"; \
+        DART_VERSION="3.0.5";; \
```
domesticmouse, line 32: I'd suggest putting these as versions above to separate configuration from code.

domesticmouse (approved):

tool/fetch-dart-sdk-sums.sh
```diff
@@ -43,13 +43,17 @@ for CHANNEL in $CHANNELS; do
...
+    _version=$(curl -fsSL "$_urlNoFile/VERSION" | awk -F: '/version/{print $2}' | sed 's/[", ]//g')
+    _url="$_urlNoFile/sdk/$_filename"
     curl -fsSLO $_url
     _checksum=$(shasum -a 256 $_filename)
     read -a _fname_arr <<< "${_checksum}" # Read in string output as array
     _checkonly="${_fname_arr%:*}" # Remove filename portion of checksum output
     printf "$(yellow "  DART_SHA256=\"$_fname_arr\";") $ENDING"
-    printf "$(yellow "  SDK_ARCH=\"$_arch\";;") $ENDING"
+    printf "$(yellow "  SDK_ARCH=\"$_arch\";") $ENDING"
+    printf "$(yellow "  DART_VERSION=\"$_version\";;") $ENDING"
```
atsansone, line 56: **issue:** This dumps numeric formatting characters instead of color coding, which has made updating even more difficult. Can this be changed?

That way, we could run the script to output to a file that we could then just copy/paste into this file. As is, I need to highlight and copy.

atsansone (changes requested):
**question:** I'm unsure as to how this resolves the linked issue. The concern there was that the manual updating of checksums breaks docs builds. This PR doesn't appear to change that issue. 

We would want something closer to [this StackOverflow response](https://stackoverflow.com/questions/2699666/replace-delimited-block-of-text-in-file-with-the-contents-of-another-file). The script should write the changes to a file without any formatting code, then the script should replace the block as a whole. That would simplify build issues greatly.

MaryaBelanger:
> Let me know if you think that functionality would be worthwhile to add though.

Nope, I was just asking to understand it better. It sounds good to me overall (but definitely consider Brett and Tony's more technical reviews). 

My reason for asking is, even though it _is_ annoying when everything breaks because checksums are out of date, I was concerned this new method might allow us to get too out of date since we're not _forced_ to update all the time anymore (just thinking of worst case, I know you're on top of things regardless). I'm still unsure if this issue might be possible with this new design, since you said:

> the generated text just now contains the versions too, _and can be updated whenever you want rather than being required when a release is made_

So I guess, is it possible that the absence of disruptive breaks could lead to us getting out of sync with Dart? And if so, would the ramifications of that be more dire than dealing with the current checksum problem? 

Again, _very_ not sure if these questions are on topic or make sense at all. If this is not related to your work here, just responding with "not an issue!" is all I need :)

PR #5017, opened 2023-06-27
Update Effective Dart on mixins for Dart 3 changes

### LINKED ISSUE #4941: Update `mixin` guidance on Effective Dart
### Page URL

https://dart.dev/effective-dart/design#avoid-mixing-in-a-class-that-isnt-intended-to-be-a-mixin

### Page source

https://github.com/dart-lang/site-www/blob/main/src/effective-dart/design.md?plain=1#L728

### Describe the problem

> For compatibility, Dart still allows you to mix in classes that aren’t defined using mixin.

I don't think that's the case now. The purpose of adding mixin classes was to no longer have this ambiguity. 

### Expected fix

Rewrite the section, or more likely remove it, or maybe replace it with different guidance e.g. "DO use a mixin class to..." 

### Additional context

_No response_

MaryaBelanger:
Fixes #4941 

The two existing sections on mixins in Effective Dart: Design don't make sense anymore, now that the `mixin` keyword is _required_ to be able to mixin a type:
- [DO use `mixin` to define a mixin type](https://dart.dev/effective-dart/design#do-use-mixin-to-define-a-mixin-type)
- [AVOID mixing in a type that isn’t intended to be a mixin](https://dart.dev/effective-dart/design#avoid-mixing-in-a-class-that-isnt-intended-to-be-a-mixin)

I honestly think both could be completely removed, and nothing would be lost. There's nothing to update here, since everything to do with the `mixin` keyword is now an explicit rule.

But for the sake of keeping some mixin guidance here (and having somewhere to anchor the old section header links to), I removed only the more egregiously incorrect section, and rewrote the other. _I vaguely remembered @munificent telling me that the idea is `mixin class` shouldn't really be recommended for anyone to use unless they're trying to maintain behavior for pre-dart-3 classes that might be being used as mixins._ 

So, that's what the new section is about. If that's no good for Effective Dart, we can just completely erase it. But there won't be anywhere to link the old mixin sections to (maybe @parlough you can advise what we can do in that case).


MaryaBelanger:
> However, there is still a lot of code written with Dart 2.12 and 2.19 language versions (which can run with Dart 3), so I'm assuming the original guidelines still apply to some extent, at least to help in future migrations.

Yeah, I just read you and Brian's response to my [`prefer_mixins` question](https://github.com/dart-lang/linter/issues/4502), and I think I need to reevaluate this PR in that context. I'll switch to a draft while I do that and see if my idea still makes sense (and like you said, you can do the same! :))

munificent (approved):
This is great!

Issue #5016, opened 2023-06-27
View page source button for individual lint pages doesn't work


MaryaBelanger:
### Describe the problem

For example, clicking the "View page source" button on this page https://dart.dev/tools/linter-rules/prefer_mixin leads to this 404 in the site repo https://github.com/dart-lang/site-www/tree/main/src/tools/linter-rules/prefer_mixin.md

### Expected fix

I'm sure this has something to do with how they live in the linter repo.

### Additional context

Not a big deal, but something to look into eventually.

parlough:
I did not think of this, thank you so  much for catching it. What do you think about hiding it for now on those pages until we figure out where the content will live in the future?

MaryaBelanger:
> What do you think about hiding it for now on those pages until we figure out where the content will live in the future?

That sounds good!

PR #5018, opened 2023-06-27
Hide page source button on generated lint pages

parlough:
It does this by adding a class (`generated-page`) to the body of the generated linter rule pages, then removing the first of the two buttons when that class is present.

Closes https://github.com/dart-lang/site-www/issues/5016


parlough:
\cc @MaryaBelanger Thanks for catching this so quickly!

domesticmouse (approved):

MaryaBelanger:
Woo! Thanks Parker :)

Issue #5011, opened 2023-06-25
Add styles for breadcrumbs


parlough:
### Describe the problem

We have support for generating breadcrumbs, but there is no proper style and the position causes the header to change positions.

### Expected fix

Adjust location and add styles to be similar to Flutter's, perhaps with some coloring differences.

PR #5023, opened 2023-07-02
Improve breadcrumbs style and enable for linter rules

parlough:
We had a breadcrumbs style, but they were too small, in the wrong position, and placed incorrectly due to the h1 bottom margin. This PR adjusts the location and style to be more consistent with Flutter, below the h1 title, and a readable size to match other text.

Using this new functionality, this PR then enables breadcrumbs for the linter rule pages, so users can easily navigate back to the index page after learning more about a specific lint.

**Example:** https://dart-dev--pr5023-feature-linter-rule-jrf93owm.web.app/lints/avoid_dynamic_calls
Example of avoid_dynamic_calls with breadcrumbs

Closes https://github.com/dart-lang/site-www/issues/5011

### DIFF
--- a/src/_layouts/default.html
+++ b/src/_layouts/default.html
@@ -11,15 +11,15 @@
       <article>
         <div class="content">
           {% include navigation-sub.html %}
-          {% if page.show_breadcrumbs -%}
-            {% include shared/breadcrumbs.html %}
-          {% endif -%}
-          <div>
+          <div id="site-content-title">
             {% include shared/page-github-links.html %}
             {% if page.underscore_breaker_titles -%}
             <h1>{{page.title | underscore_breaker}}</h1>
             {% else %}
             <h1>{{page.title }}</h1>
+            {% endif -%}
+            {% if page.show_breadcrumbs -%}
+              {% include shared/breadcrumbs.html %}
             {% endif -%}
           </div>
           {% include navigation-toc.html id='site-toc--inline' collapsible=true %}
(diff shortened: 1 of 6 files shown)


domesticmouse (approved):

MaryaBelanger (approved):
I was just looking at some Flutter docs and thought "those breadcrumbs are cool!" 😅 

One question about closing #5011 though, was that issue originally meant to only cover linter pages, or was it about adding breadcrumbs to all pages? Not a blocking question, obviously, but just wondering if you want to leave it open to eventually add the feature to all pages like Flutter's docs have.

parlough:
Breadcrumbs were already possible to enable for a page or directory, but before this PR, the styles and position were wrong if enabled, so that issue focused on improving that.

The reason I discovered that is for linter rules and being able to navigate back from a specific rule to the index page, otherwise it's not visually clear how to get there.

We can enable it for more locations, but it mostly only works well when the file path and organization in a subdirectory make sense and are consistent. That currently isn't the case in a lot of locations yet, but is slowly improving. I opened https://github.com/dart-lang/site-www/issues/5049 to track this work and if there's any specific location you'd like to enable breadcrumbs, we can now after this PR lands! Just like docs.flutter.dev, I expect it will continue to be opt-in, rather than opt-out.

MaryaBelanger:
> it mostly only works well when the file path and organization in a subdirectory make sense and are consistent. That currently isn't the case in a lot of locations yet

It most certainly is not 😅  (yet!), so that makes a lot of sense. Thank you for explaining!

PR #5026, opened 2023-07-02
Extend local caching to more file types

parlough:
This expands local caching to also cache `webp` images, `webm` videos (used in home page), `svg` images, as well as `css` and `js` files. We append cache busts to our styles on build as well as some scripts as necessary.

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -7,7 +7,7 @@
       {
         "source": "**/*.@(jpg|jpeg|gif|png|md|txt|json|webp|webm|svg|css|js)",
         "headers": [
-          { "key": "Cache-Control", "value": "max-age=86400" },
+          { "key": "Cache-Control", "value": "max-age=3600" },
           { "key": "Access-Control-Allow-Origin", "value": "*" }
         ]
       },
--- a/firebase.json
+++ b/firebase.json
@@ -5,9 +5,9 @@
     "trailingSlash": false,
     "headers": [
       {
-        "source": "**/*.@(jpg|jpeg|gif|png|md|txt|json)",
+        "source": "**/*.@(jpg|jpeg|gif|png|md|txt|json|webp|webm|svg|css|js)",
         "headers": [
-          { "key": "Cache-Control", "value": "max-age=3600" },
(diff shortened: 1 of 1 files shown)


firebase.json
```diff
@@ -5,9 +5,9 @@
     "trailingSlash": false,
     "headers": [
       {
-        "source": "**/*.@(jpg|jpeg|gif|png|md|txt|json)",
+        "source": "**/*.@(jpg|jpeg|gif|png|md|txt|json|webp|webm|svg|css|js)",
         "headers": [
-          { "key": "Cache-Control", "value": "max-age=3600" },
+          { "key": "Cache-Control", "value": "max-age=86400" },
```
domesticmouse, line 10: This may hurt if you don't adopt hashed urls

domesticmouse (approved):

firebase.json
parlough: We add a cache bust to the critical files and don't apply this cache configuration to textual content, so I think the 1 day should be fine.

domesticmouse: It'll mean you have a day of confusion if you change CSS or images in place. 

Most of my knowledge around the upsides and downsides of caching are at least a decade out of date, which is why I said "may cause pain".

What I'd ask is what problem are you solving by extending cache times to a whole day? We are serving from well connected edge servers, and I'm guessing session lengths for people browsing our site is an hour or two at most.

Maybe you can get most of the upside you are looking for by going from five minutes to an hour. That way the confusion time for an in place edit will be an hour, instead of 24 hours.

domesticmouse: I misread, you are already caching for an hour. Now I am curious, what problem are you solving?

parlough: My primary goal here was to expand the file types cached as I noticed the lack of it for a few file types. When adding those, I imagined users using the site multiple times throughout a day or week while learning Dart or otherwise programming. I was thinking a week since we do append time stamp generated cache busts on build to styles and most scripts, but not for other assets currently so I went with a more conservative choice.

Your explanation makes sense though and one hour is likely enough while avoiding potential issues. I’ll adjust the PR to just expand the types cached :)

Issue #3199, opened 2021-05-05
Show indicator of additional length to on-page TOC on narrow pages


kwalrath:
Per feedback from @sfshaza2, it's not obvious that, in a medium-width page with a longish TOC, the TOC is abbreviated. E.g. check out https://dart.dev/guides/libraries/writing-package-pages:

![image](https://user-images.githubusercontent.com/2164483/117173661-d7edbb80-ad81-11eb-9b94-38994a164802.png)

If you aren't familiar with dart.dev TOC conventions, it looks like the TOC stops at 1. The arrow at the upper right is too far away to give this hint.

We should instead have some indication that the TOC has been shortened, and have the way to lengthen it be more obvious. Something like what d.g.c uses would be good.

![image](https://user-images.githubusercontent.com/2164483/117173982-2602bf00-ad82-11eb-880d-eb69d03e2bfb.png)
_Screenshot from https://developers.google.com/style/word-list_

sfshaza2:
Thanks for filing this, @kwalrath! Perfection!  @jayoung-lee, what confused me wasn't your page, per se, it's how the Dart site implements it's table of contents. Kathy and I met yesterday to work it out.  :)

jayoung-lee:
I see it now! The new design looks great :D Thanks for the update!

PR #5027, opened 2023-07-02
Add indicator that narrow TOC can be expanded

parlough:
Adds a `...` indicator at the bottom of the TOC when the page is narrow to make it more clear that it can be extended and there are more contents. The dots also work as a way to expand the TOC. This PR also causes the cursor to switch to a pointer when hovering over the dots and the arrows to better indicate they are interactable.

Closes https://github.com/dart-lang/site-www/issues/3199

**Staged _(make browser narrow)_:** https://dart-dev--pr5027-feature-3199-a4vm172d.web.app/language

Example of new interactable expansion icon


domesticmouse (approved):

atsansone:
Staged changes are no longer working.

image

parlough:
@atsansone I re-staged the site :)

For future reference, the staged links only stay live for 7 days after the most recent commit. You can always re-run the `build / stage` workflow from the UI manually for it to re-stage.

atsansone (approved):

PR #5030, opened 2023-07-02
Improve analyzer page location

parlough:
We're trying to remove items from within `/guides/language` slowly. These contents belong under `/tools` with the other diagnostic and linter files, better matching its location in the sidenav, and with an easier to understand URL.

I went with `/tools/analysis`, but I'm heavily open to suggestions. Perhaps `/tools/analyzer`, `/tools/analysis-options`, `/tools/customize-analysis`, etc. Let me know what makes the most sense to you  :) 

\cc @bwilkerson If you have any opinion or thoughts.

**Staged:** https://dart-dev--pr5030-misc-improved-analys-hyo50m5g.web.app/tools/analysis


bwilkerson:
I don't have any strong opinions, but IMO `/tools/analysis` has a slight advantage over `/tools/analyzer` in that `analyzer` is a command-line tool and `analysis` could be either the command-line tool or the analysis server, and most of these docs appear (after a small sample) to apply to both.

domesticmouse (approved):

src/_guides/language/sound-problems.md
```diff
@@ -640,6 +640,6 @@ also supported on setters and fields.
 [bottom type]: https://en.wikipedia.org/wiki/Bottom_type
 [cast()]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-core/Iterable/cast.html
 [Iterable]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-core/Iterable-class.html
-[implicit casts]: /guides/language/analysis-options#enabling-additional-type-checks
+[implicit casts]: /tools/analysis#enabling-additional-type-checks
```
MaryaBelanger, line 643: This link isn't used on the page anywhere, could probably just delete it

src/_guides/language/index.md
MaryaBelanger, line 1: Not related to this PR, but what page is this? Are we just maintaining this in case anyone external has the link? Because I don't see any way to get here from anywhere else within the site -- searched the code base and nothing's linking to it. Just curious, no need to deal with it in the PR!

src/effective-dart/style.md
```diff
@@ -78,7 +78,7 @@ const foo = Foo();
 class C { ... }
 {% endprettify %}
 
-[Linter rule]: /guides/language/analysis-options#the-analysis-options-file
+[Linter rule]: /tools/analysis#the-analysis-options-file
```
MaryaBelanger, line 81: I don't think this is actually linking to any text, probably can delete

src/tools/linter-rules/index.md
```diff
@@ -146,5 +146,5 @@ These rules identify possible issues around
 
 {% include linter-rules-section.md type="pub" %}
 
-[Enabling and disabling linter rules]: /guides/language/analysis-options#enabling-linter-rules
+[Enabling and disabling linter rules]: /tools/analysis#enabling-linter-rules
```
MaryaBelanger, line 149: I don't think this link is used anywhere on the page (it's already linked on 66 and only used in one place), can probably delete it

MaryaBelanger (approved):
Thanks Parker, good idea! All my comments are just unused links (not added by you) and one unrelated question, so feel free to merge!

src/_guides/language/index.md
parlough: Nice catch. I forgot about this page. It was probably linked to automatically at some point when the language tour was located in that path, such as in breadcrumbs. I'll follow up and probably remove it in a separate PR.

parlough:
Thanks for catching all of those! I push a commit removing the unused definitions :)

src/_guides/language/index.md
MaryaBelanger: Thank you for taking the initiative on that!

parlough: @MaryaBelanger After more thought, while we have other docs under `/guides/language/<etc>`, it makes sense to keep a page there. Since it's common to find an index when navigating to the base path. We can always revisit its removal if/when we remove the other docs under there. Does that seem fine for you?

MaryaBelanger: Yeah of course! I figured it was still in place for this reason. So no need to do anything now, thanks for the input.

Issue #5031, opened 2023-07-02
Remove out-of-date Dart 2 page


parlough:
The guidance on [the page](https://dart.dev/articles/archive/dart-2) has long been out of date, and with Dart 3's release, even more so.

We can remove the pages and redirect to the language evolution pages. 

The article's content will be available in Git history if needed and I've taken an up-to-date snapshot of the page on the Internet Archive: https://web.archive.org/web/20230702211824/https://dart.dev/articles/archive/dart-2.

PR #5032, opened 2023-07-02
Remove out of date Dart 2 pages and guidance

parlough:
The page's contents were out of date for a while and became even more so with Dart 3, and will continue to be increasingly out of date and a source of potential confusion. Rather than keeping the outdated content around, this PR removes it.

I've [archived the page](https://web.archive.org/web/20230702211824/https://dart.dev/articles/archive/dart-2) as of 07/02/2023 as a backup and it will also remain in Git history.

I've adjusted some Dart 2 redirects to the 2.0 entry in the language evolution page, others to `package:build`, and others to the `webdev` docs depending on what fits best.

Closes https://github.com/dart-lang/site-www/issues/5031


mit-mit (approved):

mit-mit:
Nice cleanup!

Issue #5020, opened 2023-06-30
Right shift operator assert example fails on web


iehrais:
### Page URL

https://dart.dev/language/operators.html

### Page source

https://github.com/dart-lang/site-www/tree/main/src/language/operators.md

### Describe the problem

The assertion [`assert((-value >> 4) == -0x03);`](https://github.com/dart-lang/site-www/blob/e4dd7e19c18a5e4e031c80c392e86008f44a1721/src/language/operators.md?plain=1#L334) in the [**Bitwise and shift operators**](https://github.com/dart-lang/site-www/blob/e4dd7e19c18a5e4e031c80c392e86008f44a1721/src/language/operators.md?plain=1#L303) section of the page fails.

danagbemava-nc:
Hi @iehrais, thanks for filing the issue. Does the assertion fail with a message that you can share? What platform did you run this on?

iehrais:
Hello @danagbemava-nc . There is just a message `Uncaught Error: Assertion failed` due to an expression `(-value >> 4) == -0x03` evaluates to `false` with `value` equals to `0x22;`.
I ran this in **DartPad** (Flutter 3.10.5 _Dart SDK 3.0.5_). I got the same error in  **DartPad** (Flutter 3.12.0-3.0.pre _Dart SDK 3.1.0-171.0.dev_) as well.

parlough:
Thanks for opening an issue! Trying on native compilers, it seems to work as expected, so it might be due to a difference on how Dart [represents numbers on the web](https://dart.dev/guides/language/numbers).

However, writing equivalent code in JavaScript seems to match Dart native:

JS:
```js
const value = 0x22;

console.log(-value >> 4); // Prints -3 (matching Dart native)
```

Dart (web compilers):
```dart
void main() {
  final value = 0x22;
  print((-value >> 4)); // Prints 4294967293
}
```

@sigmundch Is this difference in behavior expected?

eernstg:
This is indeed caused by the JavaScript-like number semantics on the web, cf. https://dart.dev/guides/language/numbers#bitwise-operations.

I think it would make sense to add a comment to that example about the special treatment of numbers on the web.

Here's some background info. The following variant of the example succeeds in DartPad:

```dart
void main() {
  final value = 0x22;
  final bitmask = 0x0f;

  assert((value & bitmask) == 0x02); // AND
  assert((value & ~bitmask) == 0x20); // AND NOT
[...]

PR #5033, opened 2023-07-03
Direct users to bitwise operation platform differences

parlough:
I didn't want to repeat too much of what was covered on the number semantics page or focus too much on the platform differences, so I did my best to find a minimal balance. Let me know what you think.

If we want to cover more of what @eernstg mentioned in https://github.com/dart-lang/site-www/issues/5020#issuecomment-1617650660, such as potential workarounds, that would be best as a follow-up expansion on the number semantics page rather than here.

**Staged:** https://dart-dev--pr5033-fix-5020-spu2i1id.web.app/language/operators#bitwise-and-shift-operators

Closes https://github.com/dart-lang/site-www/issues/5020

### DIFF
--- a/examples/misc/test/language_tour/operators_test.dart
+++ b/examples/misc/test/language_tour/operators_test.dart
@@ -159,13 +159,18 @@ void main() {
     assert((value & ~bitmask) == 0x20); // AND NOT
     assert((value | bitmask) == 0x2f); // OR
     assert((value ^ bitmask) == 0x2d); // XOR
+
     assert((value << 4) == 0x220); // Shift left
     assert((value >> 4) == 0x02); // Shift right
+
+    // Shift right example that results in different behavior on web
+    // due to negative operand:
+    assert((-value >> 4) == -0x03);
+
     assert((value >>> 4) == 0x02); // Unsigned shift right
-    assert((-value >> 4) == -0x03); // Shift right
     assert((-value >>> 4) > 0); // Unsigned shift right
     // #enddocregion op-bitwise
-  });
+  }, testOn: 'vm');
 
   test('if-null', () {
     // #docregion if-null
--- a/src/language/operators.md
+++ b/src/language/operators.md
@@ -317,6 +317,13 @@ you’d use these bitwise and shift operators with integers.
 | `>>>`                       | Unsigned shift right
 {:.table .table-striped}
 
+{{site.alert.end}}
+  The behavior of bitwise operations with large or negative operands
+  might differ between platforms.
+  To learn more, check out
+  [Bitwise operations platform differences][].
+{{site.alert.end}}
+
 Here’s an example of using bitwise and shift operators:
 
 <?code-excerpt "misc/test/language_tour/operators_test.dart (op-bitwise)"?>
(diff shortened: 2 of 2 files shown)


eernstg:
Sorry, I must have overlooked this one. Looking at it now.

examples/misc/test/language_tour/operators_test.dart
```diff
@@ -159,13 +159,18 @@ void main() {
     assert((value & ~bitmask) == 0x20); // AND NOT
     assert((value | bitmask) == 0x2f); // OR
     assert((value ^ bitmask) == 0x2d); // XOR
+
     assert((value << 4) == 0x220); // Shift left
     assert((value >> 4) == 0x02); // Shift right
+
+    // Shift right example that results in different behavior on web
+    // due to negative operand:
```
eernstg, line 167: This is delicate, as always, but I think we need to get it right (such that we don't send any readers who actually need to use bitwise operations on the web in the wrong direction).

If the operand is negative and we're on the web then the 32 bit unsigned value which is used as the receiver (left hand operand) or right hand operand in a bitwise operation is obtained by masking off all higher bits of the 2's complement representation of the IEEE 754 encoding, which can yield every single one of the 2^32 bit vectors that an unsigned 32 bit int can be. So we don't know anything about the outcome of `.toSigned(32)`; for instance, it could be a negative number and it could be a positive number.

You could say that if the initial value is negative and "not too negative (-2^32 or bigger)" then masking the value to 32 bits and doing `.toSigned(32)` _will_ yield the same value. But then we'd need to mention "32 bits".

In particular, it really does matter that non-zero bits at position 32 or bigger are outright discarded before we even begin to compute the bitwise operator result, and this matters for every bitwise operator:

```dart
void main() {
  var big = 1024 * 1024 * 1024 * 1024; // 2^40, >32 bits, and OK on the web.
  print(big); // '1099511627776'.
  print(big >> 0); // '0'.
  print(big >>> 0); // '0'.
  print(-big >> 0); // '0'.
  print(-big >>> 0); // '0'.
}
```

I think this implies that it isn't very helpful to establish the rule of thumb in the minds of readers that "I just need to look out for negative operands to `>>`", we really need to mention the fact that these operations take place in a 32 bit unsigned representation on the web.

So how about this:
```suggestion
    // because the operand value changes when masked to 32 bits:
```

and then we should probably have a link to [Numbers in Dart](https://dart.dev/guides/language/numbers).

src/language/operators.md
```diff
@@ -328,10 +335,15 @@ assert((value & bitmask) == 0x02); // AND
 assert((value & ~bitmask) == 0x20); // AND NOT
 assert((value | bitmask) == 0x2f); // OR
 assert((value ^ bitmask) == 0x2d); // XOR
+
 assert((value << 4) == 0x220); // Shift left
 assert((value >> 4) == 0x02); // Shift right
+
+// Shift right example that results in different behavior on web
+// due to negative operand:
```
eernstg, line 343: I guess this text is updated automatically because it's extracted from examples/misc/test/language_tour/operators_test.dart, or vice versa?

examples/misc/test/language_tour/operators_test.dart
eernstg: Ah, we have that already.

eernstg (approved):
LGTM. Lots of weeping and gnashing of teeth to get there, and then suggesting a very small change. ;-)

atsansone (approved):

parlough:
@atsansone I didn't get to look into and apply Erik's changes yet. I'll follow up with a PR for that now.

examples/misc/test/language_tour/operators_test.dart
parlough: I really appreciate the thoughtful insights and the focus on what really matters for users who need to use bitwise operations on the web.

I like your adjusted text and it makes sense to me. I opened https://github.com/dart-lang/site-www/pull/5065 which applies these changes as this PR was accidentally merged without applying them.

Thanks again!

src/language/operators.md
parlough: Yep, we use an excerpt tool which pulls snippets from the `.dart` files where they are tested :)

examples/misc/test/language_tour/operators_test.dart
eernstg: Thanks! And thank you for the kind words!

Issue #4736, opened 2023-04-04
[Dart 3] Cleanup `dart pub add` page after changes have settled


parlough:
### Page URL

https://dart.dev/tools/pub/cmd/pub-add.html

### Page source

https://github.com/dart-lang/site-www/tree/main/src/tools/pub/cmd/pub-add.md

### Describe the problem

The page got into a little awkward state since a bunch of new options and syntax was added recently and we wanted to document both. It could do with some standardization and cleanup.

### Expected fix

Alongside Dart 3 or a little after, revisit the page and make sure it all makes sense and perhaps remove some mentions of old syntax where they no longer make sense. Standardize the various headers as well.

PR #5040, opened 2023-07-04
Clean up and fix `dart pub add` page

parlough:
- Accounts for support for `override` and `dev` in CLI syntax explanations
- Clarifies that it's not the latest stable version added by default, but the latest compatible stable version
- Puts dev dependency before dependency overrides, since (I believe?) they are more common
- Removes more content about the old syntax into the related note
- Add link to hosted dependency information to match other sources
- Fix link syntax to path dependency

**Staged:** https://dart-dev--pr5040-fix-4736-ecxjqodj.web.app/tools/pub/cmd/pub-add

Closes https://github.com/dart-lang/site-www/issues/4736


src/tools/pub/cmd/pub-add.md
```diff
@@ -37,12 +38,32 @@ $ dart pub add foo:'>=2.0.0 <3.0.1'
 ```
 
 If `dart pub add <package>:<constraint>` is an existing dependency,
-it will update the constraint.
+it will instead update the existing constraint.
```
MaryaBelanger, line 41: This is a huge nitpick/overthink,, so you don't have to take it, but I thought this wording could be a little more explicit (the original too, which I wrote, so not your fault). 

I think your edit adding "existing" moves in the right direction, but I'm still concerned saying it "updates the constraint" could be mistakenly conflated with "progresses to the next latest version", when really it's updating the pubspec entry, and the "new" constraint could be anything, even an older version. Maybe for the sake of overly-clarifying we could say:

> If `<package>` (or `<package>:<constraint>`) is an existing dependency in `pubspec.yaml`,
> `dart pub add <package>:<constraint>` updates the dependency's
> constraint to the one specified in the command.

Or maybe it'd be more useful to make this more action-driven, like:
> You can change an existing dependency's constraint in your `pubspec.yaml`
> by running `dart pub add <package>:<constraint>` again. 

And if you decide to leave it as your first edit, I'd remove "instead": "...it will update the existing constraint."

MaryaBelanger (approved):
Couple comments but mostly just for my education, so thanks in advance!

src/tools/pub/cmd/pub-add.md
```diff
@@ -6,7 +6,7 @@ description: Use dart pub add to add a dependency.
 _Add_ is one of the commands of the [pub tool](/tools/pub/cmd).
 
 ```nocode
-$ dart pub add [dev:]<package>[:<constraint>] [[dev:]<package>[:<constraint>]... ] [options]
+$ dart pub add [{dev|override}:]<package>[:descriptor] [[{dev|override}:]<package>[:descriptor] ...] [options]
```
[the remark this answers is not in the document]
parlough: My understanding is just one type of descriptor. It just often looks simpler since there is no nested information. You can also include `version` as part of the JSON descriptor:

`'test:{"hosted":"https://pub.dev","version":"1.24.1"}'`

src/tools/pub/cmd/pub-add.md
parlough: What you suggested makes a lot more sense, thanks for the thoughtful suggestion. I went with a derivative of your first suggestion, let me know how that sounds :)

src/tools/pub/cmd/pub-add.md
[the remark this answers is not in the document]
MaryaBelanger: Ok, so descriptor includes constraint and other things. Makes sense then, good change

src/tools/pub/cmd/pub-add.md
MaryaBelanger: Everything looks good! Double approved :)

PR #5050, opened 2023-07-07
Update verified-publishers.md

jonasfj:
Explain that what is needed for a verified publisher is a "domain property".

And link to documentation for how to create properties in Google Search Console: https://support.google.com/webmasters/answer/34592


parlough (commented):
Thanks for the extra clarification and references added here. Looks good to me!

parlough (approved):

Issue #5048, opened 2023-07-06
Add Breaking changes page 


MaryaBelanger:
_Part of https://github.com/dart-lang/sdk/issues/52828_:

> Add a new web page and populate it with past and currently planned breaking changes and deprecations.

> Flutter documents past and upcoming breaking changes in a nicely discoverable web page [here](https://docs.flutter.dev/release/breaking-changes)...While the same information is available for Dart via the [CHANGELOG](https://github.com/dart-lang/sdk/blob/main/CHANGELOG.md), we'd like to provide a similar page for Dart. 

### Plan:

- The goal is just to have a breaking changes page that's easily accessible/discoverable for users (currently, the change log is the only place that info is collected). We don't need to recreate Flutter's whole implementation.

- The CHANGELOG will stay in the sdk repo, and won't be duplicated on the site in any way. 

- The new page will be modeled after Flutter's:
  - "Not yet released to stable" section for planned breaking changes
  - "Released in Dart x.x.x" sections, descending order, most recent at top

- Past breaking changes per release will need to be pulled in to the new page to populate past release sections. 

- For now, we've decided to link to the change's corresponding issue (like the change log does now). 
  - I think we should revisit this; individual migration guides per breaking change seem like they could be very useful. 
    - But, there are so many very very minor breaking changes, _requiring_ a page be written for each one is too much.
  - I've personally found in the past that change issues aren't always a straightforward means of understanding the work, outcome, steps to take, etc. 
    - But, breaking change issues have always followed a more stringent outline than others, so it's probably ok.
  - At the very least, evaluating whether a page might need to be written should be part of the new breaking change process.

parlough:
> Link to issues of the changes

So you think the breaking change index should only link to the issues? Or just for the breaking changes from before this plan?

I was originally thinking each breaking change would have its own page with background information a migration guide, even if small for consistency, similar to the Flutter version. Then that page could link to the related issues and pull requests. However, perhaps it makes sense for simple ones to just link to the issue, not sure.

MaryaBelanger:
> So you think the breaking change index should only link to the issues? Or just for the breaking changes from before this plan?

So, I just asked some of the Dart team leads and they said this is what they were imagining. I was previously imagining what you described, the way Flutter does it, because I think that's a really useful kind of documentation. But my question was more along the lines of "do you see your engineers writing that kind of documentation and will that be part of the official breaking change process" and the answer was basically no, along the lines of:
[...]

PR #5061, opened 2023-07-13
Add breaking changes page

MaryaBelanger:
_Fixes: #5048_ 

This page will become part of the [breaking change process](https://github.com/dart-lang/sdk/issues/52828) for Dart.

## This PR:

- Copied all the breaking changes from the change log to fill out the page. 

- Made some decisions on writing/organizing standards to keep the page consistent. 

- Edited the past content somewhat, so that past entries can serve as an example for how future entries should look.

## Decision on standards:

I strongly suggest it be a REQUIREMENT that each breaking change listed here _link to more info_, rather than summarizing all the details inline. 

Links should go to either:
- a breaking change issue
- a breaking change page (like Flutter uses)
- a combination of both (i.e. engineer chooses which to use).

If we decide on both (allowing the engineer to decide), then the template must be _exactly_ identical for both the issues and pages. 

There are also compelling arguments for ONLY utilizing a dedicated page on the site, and not linking to repo issues: https://github.com/dart-lang/site-www/issues/5048#issuecomment-1631160427. We should discuss this as part of the policy updates ([#52884](https://github.com/dart-lang/sdk/issues/52884), [#52885](https://github.com/dart-lang/sdk/issues/52885)).

The rest of the writing standards I came up with are predicated on the fact that all supporting information will NOT be explained on this page:

## Writing rules _for this page_:

These rules, once agreed on, need to be synthesized somewhere (dart.dev? alongside the policy in the sdk repo?) for engineers to follow when adding their breaking changes.   

1. Section layout on the page is:
  ```
  Not yet released in stable
  Released in x.x.x
  | -- Language                  # Doesn't have subections
  | -- Libraries                 # No more "core libraries / other libraries" like in the changelog
      | -- `library:name`
  | -- Tools
      | -- Tool name / area      # This includes pub, vm, linter, etc 
  Released in x.x.y              # etc
  ```

   - Don't include the header if no changes exist in that area or release (e.g. if there's no Language breaking changes, don't include the Language header)

2. The bullet point should ONLY describe the break:
    - Don't include the fix.
    - Don't explain why.
    - Don't explain the previous state.
    - Don't detail changes across versions ("previously announced in..., deprecated in..., removed in ..., etc)
    - Don't include caveats
    
    See "Decision on standards"; these supporting details are necessary, but do not belong on the index page.

3. Start each sentence with a past-tense verb describing the break.
    E.g.: "Deprecated ...., Added...., Disallowed...."
    - To the best of your ability; sometimes this doesn't make sense.
    - But whatever you do, DON'T describe a change in any way that includes "...has been deprecated", "....was removed", etc.
    - E.g. ["`dart_api_dl.cc` is renamed to..."](https://github.com/dart-lang/sdk/blob/main/CHANGELOG.md#281---2020-05-06:~:text=%3A%20dart_api_dl.cc-,is%20renamed%20to,-dart_api_dl.c%20and), change to "Renamed `dart_api_dl.cc`..."

4. Link to the issue/migration page over the _shortest part of the sentence_ that makes sense to describe the break.
    - E.g. "[Updated the SecurityContext class](https://github.com/dart-lang/sdk/issues/46875) to set the minimum TLS protocol version to TLS1_2_VERSION (1.2) instead of TLS1_VERSION."

5. ONLY use "For more information, see..." if the break has multiple issues to link to. 
    - E.g.: "See [48730](https://github.com/dart-lang/sdk/issues/48730) and [49941](https://github.com/dart-lang/sdk/issues/49941) for more details."
    - Otherwise, follow 4.

6. Break multifaceted breaking changes into separate bullets (linking to the same issue). 
    - E.g. Compare how [this](https://github.com/dart-lang/sdk/blob/main/CHANGELOG.md#281---2020-05-06:~:text=Breaking%20change%20%2340483%3A) was written originally, with how I broke it up in this PR.

7. Try not to add -s, -'s, -ed, etc., to the end of code formatted words. 
    - E.g. [Here](https://github.com/dart-lang/sdk/blob/main/CHANGELOG.md#281---2020-05-06:~:text=Dart_NewWeakPersistentHandle%27s%20and%20Dart_NewFinalizableHandle%27s)

8. Don't say "...in x.x.x" or "...in `library`", etc., that's what the section headings are for.
    - E.g. [Here](https://github.com/dart-lang/sdk/blob/main/CHANGELOG.md#281---2020-05-06:~:text=Constants%20in%20dart%3Aio%27s)

9. If a break is language versioned, mark it as such by starting the bullet with "**Language versioned:**"

10. Edited to add: address instructions for moving "Not yet released to stable" entries to new release sections (see [this comment](https://github.com/dart-lang/site-www/pull/5061#discussion_r1265948401) below)


MaryaBelanger:
I made the release sections' headings h2, but the next level (Language, Libraries, Tools) are h4, and their subsections are h5. I did this because all the repetitive h3s were crowding the sidebar and not useful in a TOC.
    
@atsansone I know that skipping heading levels is an accessibility issue, so @parlough I'm wondering if there's another way to hide these from showing up in the contents?

MaryaBelanger:
@itsjustkevin @leafpetersen I drafted up my idea for the page, [staged here](https://dart-dev--pr5061-breaking-changes-4t42rms7.web.app/resources/breaking-changes). Open to suggestions on layout (like, do we still want to distinguish between "Language", "Libraries", etc. like in the change log?)

I'm also convinced that the new policy should enforce a standard for expanding on details of the change ("Decision on standards" section above). That's beyond the scope of adding this page, though, but we should talk about it sometime because it does affect this.

src/resources/breaking-changes.md
```diff
@@ -0,0 +1,788 @@
...
+
+This page lists all the breaking changes in Dart, organized by release and area.
+
+According to our [breaking change policy], each listed item contains a link to more
+information on the change, such as migration guidance, background, reasoning,
+caveats, and more.
+
+{{site.alert.note}}
+Changes prior to Dart 3.1.0 were not subject to our updated breaking change policy,
+so might not have links to additional information. More details per change might
```
itsjustkevin, line 14: This seems a bit off, I would expect one of the following:

- So we might not...
- Additional information may be missing...

I don't know but it feels a bit weird.

src/resources/breaking-changes.md
```diff
@@ -0,0 +1,788 @@
+---
+title: Breaking changes
+description: A list of breaking changes by release in Dart.
+---
+
+This page lists all the breaking changes in Dart, organized by release and area.
```
itsjustkevin, line 6: This page lists the breaking changes in Dart?

itsjustkevin (commented):
This looks great @MaryaBelanger, feel free to ignore my comments, I may be completely wrong in these.

src/resources/breaking-changes.md
MaryaBelanger: Yeah, it's kind of an awkward note overall. I rewrote it to get rid of that clause entireIy, but I was also looking for feedback on whether this is even worth mentioning. I included it because I thought if anyone's digging through old breaking changes they might be confused, but I could also just leave it out.

src/resources/breaking-changes.md
```diff
@@ -0,0 +1,828 @@
...
+Complete release notes are available in the [Dart SDK change log][change log].
+
+According to our [breaking change policy], each listed change contains a link to more
+information on the change, such as migration guidance, background, reasoning,
+caveats, and more.
+
+{{site.alert.note}}
+Changes prior to Dart 3.1.0 and our updated breaking change policy 
+might not have links to additional information. More details on some of those
+changes might be available in their corresponding [change log] entry.
```
parlough, line 16: Maybe we can just keep this for a while. I think most users will be looking for new releases rather those old ones, so this note will slow down the average use case.

src/resources/breaking-changes.md
```diff
@@ -0,0 +1,828 @@
...
+  generate slightly different JS code for methods that have optional parameters.
+
+[52801]: https://github.com/dart-lang/sdk/issues/52801
+[52334]: https://github.com/dart-lang/sdk/issues/52334
+[51486]: https://github.com/dart-lang/sdk/issues/51486
+[52027]: https://github.com/dart-lang/sdk/issues/52027
+
+{% comment %}
+Create new section from these headers for each release.
+Copy entries from "Not yet released to stable".
```
parlough, line 73: I'm a bit confused on this instruction. Is it talking about moving changes from "Not yet released to stable" to a specific release when it becomes stable? If not, you can ignore the following.

If it is though, we probably should be more specific here as after the beta cut there is a period where a stable version isn't released but breaking changes that are landed aren't part of that coming stable release. Also some changes may have been reverted in that timeline. Essentially there may need to be a bit of a manual verification process.

src/resources/breaking-changes.md
```diff
@@ -0,0 +1,828 @@
...
+{{site.alert.note}}
+Changes prior to Dart 3.1.0 and our updated breaking change policy 
+might not have links to additional information. More details on some of those
+changes might be available in their corresponding [change log] entry.
+{{site.alert.end}}
+
+**Dart has two types of breaking changes**:
+* **Standard**: Causes breaks as soon as you [upgrade your sdk version][sdk]. 
+  *These are the majority of changes and are not specially marked in this list.*
+* **Language versioned**: Causes breaks only when you [upgrade the
```
parlough, line 22: If we keep this labeling mechanism, I think it would be worth going through past versions (perhaps to Dart 2.7 when language versioning was released) and labeling language versioned changes, so we can be consistent throughout. At least to 2.12 since those language versions are still supported.

src/resources/breaking-changes.md
parlough, line 16:
```suggestion
  Changes prior to Dart 3.1.0 and Dart's updated breaking change policy 
  might not have links to additional information. 
  More details on some of those changes might
  be available in their corresponding [change log][] entry.
```

src/resources/breaking-changes.md
```diff
@@ -0,0 +1,828 @@
+---
+title: Breaking changes
+description: A list of breaking changes by release in Dart.
+---
+
+This page lists the breaking changes in Dart, organized by release and area.
+Complete release notes are available in the [Dart SDK change log][change log].
```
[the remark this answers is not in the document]
MaryaBelanger: I prefer that too! External places I looked all say "change log", so I figured I should do that, but I didn't even think to look around our own site or pub.dev, so I'm happy to use "changelog"!

src/resources/breaking-changes.md
MaryaBelanger: I agree. How long do you think makes sense? I could also just not include it at all, open to either

src/resources/breaking-changes.md
```diff
@@ -0,0 +1,828 @@
...
+caveats, and more.
+
+{{site.alert.note}}
+Changes prior to Dart 3.1.0 and our updated breaking change policy 
+might not have links to additional information. More details on some of those
+changes might be available in their corresponding [change log] entry.
+{{site.alert.end}}
+
+**Dart has two types of breaking changes**:
+* **Standard**: Causes breaks as soon as you [upgrade your sdk version][sdk].
```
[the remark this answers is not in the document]
MaryaBelanger: I'm not sure I understand, do you mean that the first bullet (`Standard`) should say something like "Breaks code related to the change as soon as you update your sdk version"?

src/resources/breaking-changes.md
```diff
@@ -0,0 +1,828 @@
...
+#### (`dart:core`, `package:js`, etc)
+
+### Tools
+{: .no_toc}
+
+#### (Dart VM, Pub, Linter, `dart2js`, etc)
+{% endcomment %}
+
+## Released in 3.0.0
+
```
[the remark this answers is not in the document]
MaryaBelanger: Good point, that page also allows me to remove the changes listed out under `dart:collection` in this section and just link to the migration guide, shortening the section.

src/resources/breaking-changes.md
MaryaBelanger: I think I'll remove the instruction to copy from "Not yet released" to the new section (line 73) from this doc. Since there needs to be instructions in the breaking change policy about how to handle all of this anyway, we can elaborate there. 

I'll leave the heading layout for ease of copying in the future.

Lmk if that doesn't address your concern!

src/resources/breaking-changes.md
MaryaBelanger: I'm _pretty sure_ I labeled all the ones that were explicitly called out already in their original changelog entries. I could've missed some though, but I'd need someone else to confirm because I have no way of knowing unless they're labeled

src/resources/breaking-changes.md
```diff
@@ -0,0 +1,818 @@
+---
+title: Breaking changes
```
leafpetersen, line 2: I think we should make this cover deprecations as well, so perhaps "Breaking changes and deprecations"?

src/resources/breaking-changes.md
```diff
@@ -0,0 +1,818 @@
...
+description: A list of breaking changes by release in Dart.
+---
+
+This page lists the breaking changes in Dart, organized by release and area.
```
leafpetersen, line 6: "lists the breaking changes and deprecations..."

src/resources/breaking-changes.md
```diff
@@ -0,0 +1,818 @@
...
+Complete release notes are available in the [Dart SDK changelog][changelog].
+
+According to Dart's [breaking change policy][], each listed change contains a link to more
```
leafpetersen, line 9: This sentence reads a little odd to me.  It's not really according the breaking change policy that there is a link.  Perhaps restructure this something like:

The Dart language policy and process around breaking changes and deprecations is described in the [breaking change policy][] document.  In order to help Dart programmers understand and manage the impact of language and library breaking changes and deprecations, all such changes (both upcoming and previously released) are listed on this page.  Each listing contains a link ....

src/resources/breaking-changes.md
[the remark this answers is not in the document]
leafpetersen: I might say something like:

**Unversioned**:  The compiler does not maintain backward compatibility, and code may break as soon as you upgrade your sdk version if it relies on the previous behavior. *These are the majority....*
**Language versioned**: The compiler maintains backward compatibility for existing code, and the behavior change only takes effect (potentially breaking code which relied on the previous behavior) when a you upgrade ....

src/resources/breaking-changes.md
```diff
@@ -0,0 +1,818 @@
...
+  More details on some of those changes might
+  be available in their corresponding [changelog][] entry.
+{{site.alert.end}}
+
+**Dart has two types of breaking changes**:
+* **Standard**: Causes breaks as soon as you [upgrade your sdk version][sdk]. 
+  *These are the majority of changes and are not specially marked in this list.*
+* **Language versioned**: Causes breaks only when you upgrade the
+  [language version] of your code. *These are marked "**Language versioned**"*.
+
```
leafpetersen, line 25: If we cover deprecations here, we should probably have a comment and marker about deprecations.

src/resources/breaking-changes.md
```diff
@@ -0,0 +1,818 @@
...
+If you have questions or concerns about any of these breaking changes, please 
+[file an issue][]. To be notified about future breaking changes,
```
leafpetersen, line 27: Perhaps say "comment on the breaking change issue linked from the relevant entry" to avoid people filing new issues?

src/resources/breaking-changes.md
MaryaBelanger: I added this bullet for deprecations, but am not sure it's correct:

```md
* **Deprecations**: The compiler maintains compatibility for deprecated code,
  with a warning. Deprecations are then completely removed in the release
  following their deprecation, breaking any code that relies on the previous behavior.
  
  *These are marked "**Deprecated**" and "**Removed**", respectively*.
```

MaryaBelanger: I was also wondering, going back through older entries, is _anything_ that even mentions "removed/removal/etc" considered a removal, based on my definition above?

src/resources/breaking-changes.md
```diff
@@ -0,0 +1,821 @@
...
+---
+
+This page lists all language and library breaking changes and deprecations in Dart,
+organized by release and area, to help Dart users understand and manage their
+impact. Complete release notes are available in the [Dart SDK changelog][changelog].
+The [breaking change policy][] document describes the policy and process
+around breaking changes and deprecations in Dart. 
+
+**This page includes two types of breaking changes, as well as deprecations**:
+* **Unversioned**: The compiler does not maintain backward compatibility, and
```
parlough, line 13: Was "compiler" explicitly chosen here? As far as I understand, breaking changes can result from tooling other than the compiler itself. Such as the analyzer introducing a new error, a tool being removed or changed (`dartfmt` in 2.14), etc.

Perhaps "Dart does not guarantee..."? Or the "(Dart) SDK does not guarantee..."

src/resources/breaking-changes.md
```diff
@@ -0,0 +1,821 @@
...
+  code may break as soon as you [upgrade your sdk version][sdk] if it relies on
+  the previous behavior.
+  
+  *These are the majority of changes and are not specially marked in this list.*
+* **Language versioned**: The compiler maintains backward compatibility for
+  existing code, and the behavior change only takes effect (potentially breaking
+  code that relies on the previous behavior) when you upgrade the
+  [language version] of your code.
```
parlough, line 21:
```suggestion
  [language version][] of your code.
```

src/resources/breaking-changes.md
```diff
@@ -0,0 +1,821 @@
...
+  
+  *These are marked "**Language versioned**"*.
+* **Deprecations**: The compiler maintains compatibility for deprecated code,
+  with a warning. Deprecations are then completely removed in the release
+  following their deprecation, breaking any code that relies on the previous behavior.
```
parlough, line 26: In the past, deprecations have been removed a bit inconsistently. The next release, a few releases later, a major release (Dart 3). Is there a documented process we can refer to here? Or just mention a minimum instead?

src/resources/breaking-changes.md
```diff
@@ -0,0 +1,821 @@
...
+
+If you have questions or concerns about any of these breaking changes, please 
+comment on the breaking change issue linked from the relevant entry.
+To be notified about future breaking changes, join the [Dart announce][] group.
+
+[breaking change policy]: https://github.com/dart-lang/sdk/blob/main/docs/process/breaking-changes.md
+[changelog]: https://github.com/dart-lang/sdk/blob/main/CHANGELOG.md
+[sdk]: /get-dart
+[language version]: /guides/language/evolution#language-versioning
+[Dart announce]: https://groups.google.com/a/dartlang.org/g/announce
```
parlough, line 38: I feel like we have a few places we link to this now. Consider adding a site variable in the [`config.yml`](https://github.com/dart-lang/site-www/blob/main/_config.yml#L32-L40).

src/resources/breaking-changes.md
```diff
@@ -0,0 +1,821 @@
...
+
+## Not yet released to stable
```
parlough, line 40: Just a note to update this for the 3.1 release. Note that the changelog may have updated since you wrote this as well.

parlough, line 40: Now that 3.1 has released, could you add a new "Not yet released to stable" section for 3.2. Looks like there are already a few breaking changes there as well. Thank you!

src/resources/breaking-changes.md
```diff
@@ -0,0 +1,821 @@
...
+[50231]: https://github.com/dart-lang/sdk/issues/50231
+[`Metrics`]: https://api.dart.dev/stable/2.18.2/dart-developer/Metrics-class.html
+[`Metric`]: https://api.dart.dev/stable/2.18.2/dart-developer/Metric-class.html
+[`Counter`]: https://api.dart.dev/stable/2.18.2/dart-developer/Counter-class.html
+[`Gauge`]: https://api.dart.dev/stable/2.18.2/dart-developer/Gauge-class.html
+[49536]: https://github.com/dart-lang/sdk/issues/49536
+[51035]: https://github.com/dart-lang/sdk/issues/51035
+
+
+## Released in 2.19.0
```
parlough, line 176: This isn't necessary, but a question/consideration:

I wonder if it's worth it to explicitly set the IDs of the version headers for easier to remember/write link targets. For example, this one could be:

```
## Released in 2.19.0 {#2-19-0}
```

Then the link would just be https://dart.dev/resources/breaking-changes#2-19-0

parlough (commented):
Thanks for continuing to work on this! A few more comments:

src/resources/breaking-changes.md
MaryaBelanger: Yeah I got it from [this suggestion](https://github.com/dart-lang/site-www/pull/5061#discussion_r1268755106) but I'm not tied to that phrasing, I'll change it to "Dart/SDK"

src/resources/breaking-changes.md
MaryaBelanger: I can just remove the "in the release following their deprecation" and make it more general, thanks for letting me know!

src/resources/breaking-changes.md
MaryaBelanger: Great idea! Will do

src/resources/breaking-changes.md
MaryaBelanger: About 3.1: Do you think [the removal of `ObjectLiteral`](https://github.com/dart-lang/sdk/blob/main/CHANGELOG.md#dartjs_interop-1) is a breaking change or should be marked "Removed" under the deprecation format I just added to this PR?

src/resources/breaking-changes.md
MaryaBelanger: Are we sure that's how that works? 😅 
image

src/resources/breaking-changes.md
MaryaBelanger: So how does this work? When I add the link to config.yml does that mean I no longer need to include this line? I can just write `[Dart announce]` anywhere on the site and it will link?

src/resources/breaking-changes.md
parlough: Good question. That's tough since `dart:js_interop` is WIP and comes with a warning that it may change, so its not a traditional breaking change since that status allows for it.

I guess it's fine to just call it a breaking change even though any users should be prepared for the break.

src/resources/breaking-changes.md
parlough: You'll instead just link to `{{site.announce}}` (or whatever you name it) so you don't have to find/time that long URL everytime. Plus if it ever changes, it'll be easy to update.

```
[Dart announce]: {{site.announce}}
```

src/resources/breaking-changes.md
parlough: Maybe that's what I meant 👀 ...

Emm after some testing, it seems overriding the ID doesn't work if the override begins with a number. I think it's fine to drop this idea for now.

MaryaBelanger: I figured it was the numbers. Oh well, good idea while it lasted!

MaryaBelanger:
@parlough So regarding some of the `js_interop` stuff, I added an **Experimental** tag based on a quick convo with Siggi... what do you think? Does the description make sense?

I'm also not sure how to write the **Removed** ones like lines 74, 76, and 132. It's kind of awkward ("Removed: Removed `thing`" seems repetitive, but "Removed: `thing`" seems weird) so any suggestions welcome

src/resources/breaking-changes.md
leafpetersen: > is anything that even mentions "removed/removal/etc" considered a removal

I think this is case to case.  Someone might mention removing something in the CHANGELOG that still wasn't part of the public API.

parlough:
> So regarding some of the `js_interop` stuff, I added an **Experimental** tag based on a quick convo with Siggi... what do you think? Does the description make sense?
> 
> I'm also not sure how to write the **Removed** ones like lines 74, 76, and 132. It's kind of awkward ("Removed: Removed `thing`" seems repetitive, but "Removed: `thing`" seems weird) so any suggestions welcome

You're right that they are a little odd currently, but I think they are fine for now. I'm thinking we can later replace these with little tags rather than text that won't look as odd next to each other. We can always come back to that as follow-up after we're happy with the format.

I'm envisioning something like this:

Example tags

src/resources/breaking-changes.md
```diff
@@ -0,0 +1,873 @@
...
+  *These are marked "**Language versioned**"*.
+* **Deprecations**: The Dart SDK maintains compatibility for deprecated code,
+  with a warning. Deprecations are then completely removed in a subsequent release,
+  breaking any code that relies on the previous behavior.
+  
+  *These are marked "**Deprecated**" and "**Removed**", respectively*.
+* **Experimental**: Part of the release but not yet treated as stable in the SDK,
+  and can break from one version to another. Experimental changes do not
+  always have a corresponding breaking change issue, but may have more detail in
+  the [changelog].
```
parlough, line 33: Consider specifying "SDK changelog" instead.

src/resources/breaking-changes.md
```diff
@@ -0,0 +1,873 @@
...
+
+  *These are marked "**Experimental**", possibly in conjunction with another change type*
```
parlough, line 35:
```suggestion
  *These are marked "**Experimental**", possibly in conjunction with another change type*.
```

src/resources/breaking-changes.md
```diff
@@ -0,0 +1,873 @@
...
+The [breaking change policy][] document describes the policy and process
+around breaking changes and deprecations in Dart. 
+
+**This page includes the following types of breaking changes**:
+
+* **Unversioned**: The Dart SDK does not maintain backward compatibility, and
+  code may break as soon as you [upgrade your sdk version][sdk] if it relies on
+  the previous behavior.
+  
+  *These are the majority of changes and are not specially marked in this list.*
```
parlough, line 18: Consider using underscores (`_`) to indicate italic to visually differentiate the raw markdown from bold.

parlough (approved):
Thanks for working on this @MaryaBelanger!

How about we drop the page from the sidenav for now and land this PR. Then iterate on the page and process and add it back to the sidenav alongside Dart 3.2?

MaryaBelanger:
@parlough Thanks for the for the final review! 

> How about we drop the page from the sidenav for now and land this PR. Then iterate on the page and process and add it back to the sidenav alongside Dart 3.2?

I don't _need_ it to be in the sidenav just yet, but I also think it's fine to add now (or once this lands, without waiting for 3.2). Why do you think we should wait? Do you want 3.2 to be up to date in the "Not yet released" section, or for some other reason? (I feel like we might've talked about this already, so sorry if I forgot!)

parlough:
> Why do you think we should wait?

I just didn't know where you were at personally with the page/process. I think it's completely fine in the sidenav, just wanted to convey some flexibility!

parlough:
If you like the colored tags, please create an issue so we don't forget :)

leafpetersen (approved):
LGTM, thanks!

PR #5064, opened 2023-07-14
Diagnostic message changes

MaryaBelanger:
I just added some [very minor edits](https://dart-review.googlesource.com/c/sdk/+/309780) to the diagnostic messages file in the sdk. I followed [these steps](https://github.com/dart-lang/site-www/issues/1949#issuecomment-535257768) to copy the generated file from there to here.

However, there seems to be some other changes that I didn't make that are also new to the site-www version of the diagnostics. I checked side by side and they seem to have been recently added in the last few weeks, so maybe we just haven't updated the site since then. 

_Fixes #4740_


parlough (approved):
Thank you!

Issue #5054, opened 2023-07-10
Move diagnostic glossary entries to site-wide glossary


parlough:
Really rough steps. Feel free to rewrite :) 

- [x] 1. Add entries to dart.dev/resources/glossary with any desired copy/technical updates (https://github.com/dart-lang/site-www/pull/5066)
- [x] 2. Update links in the SDK generator to link to `/glossary/<word>` (https://dart-review.googlesource.com/c/sdk/+/313960)
- [x] 3. Remove old glossary entries from `generate.dart` (https://dart-review.googlesource.com/c/sdk/+/313960)
- [ ] 4. Update version of diagnostic messages file on site-www #5070

MaryaBelanger:
> 2. Update links in the SDK generator to link to /glossary/<word>

Does this mean I need to update links in [`generate.dart`](https://github.com/dart-lang/sdk/blob/master/pkg/analyzer/tool/diagnostics/generate.dart)? Or just as normal in `messages.yaml`, then run the generation tool?

And I'm guessing after generating `diagnostics.md` in the sdk, I'll have to copy it over here as with any diagnostic message updates? Thanks @parlough!

parlough:
I apologize for not being super clear, I just made this issue pretty quickly. 

> 2. Update links in the SDK generator to link to /glossary/<word>
>
> Edit: Flip flopping again, did you mean I need to update the links on lines [241](https://github.com/dart-lang/sdk/blob/master/pkg/analyzer/tool/diagnostics/generate.dart#L241)-247 in generate.dart?

There are two potential solutions, I was originally thinking the first one but both are fine. Let me know what you think :) 

1. You can keep the link definitions in https://github.com/dart-lang/sdk/blob/acd2ad41b66187564b65d93cd70538c01727c720/pkg/analyzer/tool/diagnostics/generate.dart#L241-L247 but just update them to point the glossary entries. This way you don't need to update a bunch of links in `messages.yml` which use the link definitions.
2. You can specifying the links in each diagnostic message they are used in and remove those link definitions.

> And I'm guessing after generating `diagnostics.md` in the sdk, I'll have to copy it over here as with any diagnostic message updates?

Yep, exactly! You probably want to make sure to land the new glossary entries first in a separate to avoid any potential issues.

MaryaBelanger:
Great! 1) Seems interesting, I might try that first

MaryaBelanger:
[...]

PR #5066, opened 2023-07-14
Add diagnostic glossary entries to main glossary

MaryaBelanger:
Part of #5054 (step 1)

- This PR is just the move.
- Once this lands, I will update the links and then remove the glossary entries from the diagnostics page, both in the sdk repo (see #5054)
- Now that the glossary is a little more filled out, I added it to the sidenav under Resources and removed `sitemap:false` so it's contents can show up in searches.

---

- I didn’t update any of the text. Much of this could be considered to move to other places in the docs more conducive to long form content (like, most of the mixin application content might make sense on the mixins page?)
- We don’t have a standard or rules for glossary entries or "how to write glossary entries" yet, though, so for now I’m not going to say “this isn’t a glossary entry”. They'll have to be reconsidered as part of site IA work.

### DIFF
--- a/src/_data/side-nav.yml
+++ b/src/_data/side-nav.yml
@@ -347,6 +347,8 @@
           permalink: /guides/language/coming-from/js-to-dart
         - title: Swift to Dart
           permalink: /guides/language/coming-from/swift-to-dart
+    - title: Glossary
+      permalink: /resources/glossary
     - title: Books
       permalink: /resources/books
     - title: Videos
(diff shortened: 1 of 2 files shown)


parlough:
I agree some of these things ought to be better documented elsewhere on the site, especially "constant context", but keeping them in the glossary seems fine for now like you said :)

parlough (approved):
Looks good to me!

Thanks for adding it to the sidenav as well.

Consider filing individual issues for entries you think you should be visited for editing, expansion, simplification, covering elsewhere, etc. That way we don't lose track of potential improvements.

PR #5071, opened 2023-07-18
Update part file link

MaryaBelanger:
Idk if this is useful or not, but this updates a link in the pub glossary to point from the Effective Dart entry on part files, to the glossary entry. 

I had this as a To-Do on #4622. I think it makes sense in the effort to utilize the glossary and make it more ubiquitous, but since this particular entry isn't filled out a ton, it just points to the Effective Dart entry anyway.

I think that's ok though, as we want the glossary to function as a set of definitions for other content to point to, and not extensive explanations.


parlough (approved):
Makes sense to me!

PR #5088, opened 2023-07-31
Fix document about understanding-null-safety

RyochanUedasan:
The example code seems to be about "without null safety".

### DIFF
--- a/src/null-safety/understanding-null-safety/index.md
+++ b/src/null-safety/understanding-null-safety/index.md
@@ -813,7 +813,7 @@ print(notAString?.length?.isEven);
 This is annoying, but, worse, it obscures important information. Consider:
 
 ```dart
-// Using null safety:
+// Without null safety:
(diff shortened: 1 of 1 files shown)


RyochanUedasan:
I have just signed the Google CLA!

parlough (commented):
Thanks for opening a pull request and signing the CLA!

However, this snippet does appear to be "using null safety" due to the `Thing? thing` parameter which is marked as nullable which is only possible with null safety enabled.

What about it makes you think it's meant to be an example without null safety?

RyochanUedasan:
@parlough Thanks for your reply. 
I initially thought this code was another example of using the null-aware operator before implementing sound null safety, suggesting that `?` should be used every time in a method chain, regardless of whether `doohickey` is non-nullable.
However, as you pointed out, `Thing? thing` is used here, so this is about `using null safety`.

I apologize for the mistake!

PR #5092, opened 2023-08-04
Fix server tutorial prev->next links

parlough:
The Fetch Data tutorial wasn't set up as in this flow. This PR adds it.


MaryaBelanger (approved):
Looks good, links work, thanks!

PR #5102, opened 2023-08-11
Incorporate latest diagnostic message and linter rule updates

parlough:
These were written and already reviewed in the SDK and linter repositories.


MaryaBelanger (approved):
Thank you!

PR #5108, opened 2023-08-13
Cleanup and replace more dartdevc mentions

parlough:
We completed some related cleanup work in the past, but there were still some dartdevc links/mentions left. It's best to avoid linking there as it's a historical page that just links elsewhere.


src/_tutorials/web/get-started.md
```diff
@@ -219,9 +221,8 @@ Check out these resources:
   * [IDEs](/tools#ides-and-editors)
   * [Web libraries and packages](/web/libraries)
 
-If you get stuck, find help at [Community and Support.](/community)
+If you get stuck, find help at [Community and support.](/community)
```
MaryaBelanger, line 224: This has nothing to do with this pr, but I always wonder what the best way to do this is... We don't Title Case our page titles, so if you go to that page it says "Community and support". But when linking to a page or writing out the title inline in some text, title case looks way better, e.g. [Community and Support].

I usually stick to sentence case no matter what because that's how the titles are actually written. What made you want to change this instance? Have you seen more examples across the site that do it this way?

MaryaBelanger (approved):
Thanks, checked all the links and pages, all good!

src/_tutorials/web/get-started.md
parlough: Generally the style guide suggests linking with the exact title (in its sentence case) when referencing the title and it's mostly what we do across the site, so I stick with that. Consistency is great for setting expectations even if it doesn't look the best, so when I see otherwise, I try to update those references.

If that style doesn't feel right or the full title doesn't make sense in the context, it is fine to describe the topic/link as if its part of the sentence and follow the sentence's capitalization. I guess it is a better choice when something isn't important enough to explicitly call out, but you want to link to something from a word/phrase for extra/optional context.

https://developers.google.com/style/link-text discusses both choices and probably goes into more detail.

In this instance, I felt the title of the page provides enough context, so I updated the link text to match the page's title.

parlough:
Thanks for taking a look!

PR #5115, opened 2023-08-15
[3.2] Update `utf8encoder` return type in library tour

MaryaBelanger:
Changing the return type [here](https://dart.dev/guides/libraries/library-tour#:~:text=List%3Cint%3E%20encoded%20%3D%20utf8.encode) to align with [breaking change #52801](https://github.com/dart-lang/sdk/issues/52801)

@mkustermann is changing the return type sufficient in this example, or does more of it need to change?

```dart
    Uint8List encoded = utf8.encode('Îñţérñåţîöñåļîžåţîờñ');

    assert(encoded.length == utf8Bytes.length);
    for (int i = 0; i < encoded.length; i++) {
      assert(encoded[i] == utf8Bytes[i]);
    }
``` 

Also, I had to import `dart:typed_data` for the code excerpt, is that necessary for users to use the new return type? I didn't add any mention to the doc about it but I will if it's part of using the type. Thanks!


examples/misc/test/library_tour/convert_test.dart
```diff
@@ -56,7 +57,7 @@ void main() {
     assert(funnyWord == 'Îñţérñåţîöñåļîžåţîờñ');
     // #enddocregion utf8-decode
     // #docregion utf8-encode
-    List<int> encoded = utf8.encode('Îñţérñåţîöñåļîžåţîờñ');
+    Uint8List encoded = utf8.encode('Îñţérñåţîöñåļîžåţîờñ');
```
mkustermann, line 60: If you want to avoid the `import "dart:typed_data";` import, you can also rely on type inference, just making this `var` or `final`.

Looking at some other code in the docs at [0] it seems to use `var` a lot.

[0] https://dart.dev/guides/libraries/library-tour#dartconvert---decoding-and-encoding-json-utf-8-and-more

mkustermann:
> @mkustermann is changing the return type sufficient in this example, or does more of it need to change?

The example before was also correct (as `List<int>` is a supertype of `Uint8List`) but it's better written using `Uint8List`.

> Also, I had to import dart:typed_data for the code excerpt, is that necessary for users to use the new return type?
> I didn't add any mention to the doc about it but I will if it's part of using the type. Thanks!

If you want to use `Uint8List`, then yes the code has to `import "dart:typed_data";`.

You could also use `var` / `final` and rely on type inference instead of typing the `Uint8List` type explicitly. That would avoid the import.

mkustermann (approved):
lgtm with comments

examples/misc/test/library_tour/convert_test.dart
parlough: I think either way is fine. Relying on type inference is consistent with the nearby examples and what our docs default to, but showing the type here helps a bit to support the leading text that it returns a "list of UTF8-encoded bytes:".

parlough (approved):
Thanks! Looks good to me.

parlough (changes requested):
Oops. I just updated the PR to run on the 3.1.0 stable release and it failed. After checking the [commit](https://github.com/dart-lang/sdk/commit/81df36216feda133754e88458fd640d8edc038ef), looks like the change landed after the cut-off and isn't in 3.1.0, but rather 3.2.0.

MaryaBelanger:
> it looks like the change landed after the cut-off and isn't in 3.1.0, but rather 3.2.0

Oops! Well, it's ready for next time 😅 

Though, I might just drop it since relying on type inference is a good point too, as you said. I'll think about it

parlough:
I'll defer to you on if we should revisit this in 3.2 or adjust the example to rely on type inference now :)

MaryaBelanger:
Ok closing and will reassess for 3.2

MaryaBelanger:
@parlough sorry I was trying to resolve merge conflicts since this branch is pretty old, and I pulled in all the commits from main somehow. Should I just close this and start a new PR for this change? Not sure how to fix the missing CLA issue that one of those commits pulled in.

parlough:
@MaryaBelanger You could do an interactive rebase but it might just be easier to open a new PR. Did you decide to rely on type inference or specify the type?

MaryaBelanger:
@parlough I decided to use the specific type, so I added the import to the start of the excerpt. Idk if that makes sense though (since the rest of the code isn't in `main` or anything like that, it looks kinda weird...)

parlough:
> @parlough I decided to use the specific type, so I added the import to the start of the excerpt. Idk if that makes sense though (since the rest of the code isn't in `main` or anything like that, it looks kinda weird...)

I think you can wrap it in a method or just not show the import, whichever you prefer.

PR #5119, opened 2023-08-16
Update build to use stable

parlough:
I forgot to revert setting the `Makefile` to use `beta` and a new beta release came out since the last update, breaking the build.


MaryaBelanger (approved):
Ah, that explains it. thank you!

PR #5120, opened 2023-08-17
Update the name of the dart_community discord

parlough:
The Dart community Discord has renamed to dart_community to match Dart's package/file naming scheme :)


MaryaBelanger (approved):
Nice, I like the reasoning!

PR #5130, opened 2023-08-20
Finish language evolution entry for Dart 3.0

parlough:


src/_guides/language/evolution.md
```diff
@@ -49,15 +49,41 @@ _Released 10 May 2023_
...
+* [Patterns][], a new category of grammar that lets you
+  match and destructure values.
+* [Records][], a new type that lets you aggregate
+  multiple values of different types in a single function return.
+* [Class modifiers][], a new set of keywords that let you
+  control how a class or mixin can be used.
+* [Switch expressions][], a new form of multi-way branching
+  allowed where expressions are expected.
+* [If-case causes][], a new conditional construct that matches a value
+  against a pattern and executes the then or else branch depending
```
MaryaBelanger, line 61:
```suggestion
* [If-case clauses][], a new conditional construct that matches a value
  against a pattern and executes the `then` or `else` branch, depending
```

MaryaBelanger (approved):
Thanks!

src/_guides/language/evolution.md
parlough: I removed the suggested code syntax around "then" and "else", since there is no code constructor for `then` in Dart. I applied the other changes though. Thanks so much!

parlough:
Thanks for taking a look!

src/_guides/language/evolution.md
MaryaBelanger: No problem, I wasn't sure about that one and yeah your take sounds like the best way to deal with it. Thanks!

PR #5132, opened 2023-08-21
Update js-to-dart.md - quickfix "increment a before b" example

evgeniikozhanov:
The PR fixes the small issue in the "js-to-dart" docs: it seems that an assertion in the "Increment a before b" example is incorrect.


parlough (approved):
Nice catch. Thanks for fixing this!

PR #5134, opened 2023-08-21
Fix linter rule dupe

### LINKED ISSUE #5127: Remove duplicate code in 'implicit_reopen' page example
### Page URL

https://dart.dev/tools/linter-rules/implicit_reopen.html

### Page source

https://github.com/dart-lang/site-www/tree/main/src/tools/linter-rules/implicit_reopen.md

### Describe the problem

The following code is duplicated:

```dart
interface class I {}

final class C extends I {}
```

### Expected fix

_No response_

### Additional context

_No response_

atsansone:
Partial fix to #5127.


parlough (approved):
Note that these docs are written, reviewed, and then generated in the linter and copied from there. Generally we avoid changes directly within this repo otherwise they will got lost.

I haven't updated the source file yet since the linter is currently in the process of being moved to the SDK, so we can't make new updates until that is complete.

Since we can't make updates to the source file temporarily, I suppose it's fine to make this change for now. However, I removed the "Fix #issue number" text to keep the issue open. Otherwise it'll be easy to lose track of the change and the fix will be reverted when updating this file the next time. Thanks!

Issue #5041, opened 2023-07-04
Wide tables cause layout issues on narrow devices


parlough:
### Describe the problem

Large tables narrow widths, such as on the [Numbers in dart](https://dart.dev/guides/language/numbers) page, go off the screen, requiring scrolling the whole page horizontally.

### Expected fix

Make tables horizontally scrollable when necessary.

im-victor-mendez:
One solution (impractical in my opinion) is to put the tables inside a div tag and add the overflow: auto styling property to it.
I propose the use of the following syntax:
`<div class="table-container"><table ...>...</table></div>` and (in src\_sass\site.scss) `.table-container { overflow: auto; }`

Before:
<img width="689" alt="Before" src="https://github.com/dart-lang/site-www/assets/106124660/bbb0b502-8684-475b-8aa4-68701c7ce5ff">

After:
<img width="690" alt="After" src="https://github.com/dart-lang/site-www/assets/106124660/00231eb9-8169-4384-8153-35774868ff37">

I have been trying to make the changes in my environment but I have a problem with a command at the moment to make a build that I have not been able to solve.

danagbemava-nc:
Another instance of this is in https://github.com/dart-lang/site-www/issues/5062

parlough:
@im-victor-mendez Thanks so much for looking into this!

I forgot about it, but we do have a similar class already (`table-wrapper`) [declared in `site.scss`](https://github.com/dart-lang/site-www/blob/8232f9469e6592065bf761af29c57e9abd5c8577/src/_sass/site.scss#L571-L573). An example of us using it is on the the table on the [Dart keywords page](https://dart.dev/language/keywords) ([code](https://github.com/dart-lang/site-www/blob/main/src/language/keywords.md?plain=1#L18)). 

We just haven't added the div container and class consistently to all tables such as on the numbers page. If you wanted to do that, I'd be happy to stage and review your changes!

If you'd like to get things working locally, let us know what issue you are facing and perhaps we can help get your build working :)

PR #5141, opened 2023-08-23
Fixes instance of #5041 (wide tables)

im-victor-mendez:
_Fixes: Instance of #5041_

Wrapped all tables of [Codelabs pages](https://dart.dev/codelabs) that were causing horizontal overflow

### DIFF
--- a/src/codelabs/dart-cheatsheet.md
+++ b/src/codelabs/dart-cheatsheet.md
@@ -33,11 +33,13 @@ If the expression is an identifier, you can omit the `{}`.
 
 Here are some examples of using string interpolation:
 
+<div class="table-wrapper" markdown="1">
 | String                      | | Result |
 |-----------------------------+-+ -------|
 | `'${3 + 2}'`                | | `'5'` |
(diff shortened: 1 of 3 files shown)


src/codelabs/dart-cheatsheet.md
```diff
@@ -33,11 +33,13 @@ If the expression is an identifier, you can omit the `{}`.
 
 Here are some examples of using string interpolation:
 
+<div class="table-wrapper" markdown="1">
 | String                      | | Result |
 |-----------------------------+-+ -------|
 | `'${3 + 2}'`                | | `'5'` |
 | `'${"word".toUpperCase()}'` | | `'WORD'` |
 | `'$myObject'`               | | The value of `myObject.toString()` |
+</div>
```
parlough, line 42: Without the extra new line in this one, it fails to render the table, because the `.table` class isn't added on the inside.

```suggestion

</div>
```

parlough (approved):
Wow, thanks so much for fixing all of these! It's really appreciated!

Looks good to me after one minor fix:

PR #5153, opened 2023-09-05
Update extension-methods.md

hrishikesh-kadam:
Add a line on how to invoke a static method of extension.

---

<details>

</details>

### DIFF
--- a/src/language/extension-methods.md
+++ b/src/language/extension-methods.md
@@ -220,6 +220,7 @@ extension NumberParsing on String {
 
 The members of an extension can be methods, getters, setters, or operators.
 Extensions can also have static fields and static helper methods.
+Make sure to use the Extension name for invoking them.
 
 ### Unnamed extensions
 
--- a/src/language/extension-methods.md
+++ b/src/language/extension-methods.md
@@ -220,7 +220,10 @@ extension NumberParsing on String {
 
 The members of an extension can be methods, getters, setters, or operators.
 Extensions can also have static fields and static helper methods.
-Make sure to use the Extension name for invoking them.
+To access static members outside the extension declaration, 
+invoke them through the declaration name like [class variables and methods][]. 
+
+[class variables and methods]: /language/classes#class-variables-and-methods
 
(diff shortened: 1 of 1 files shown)


src/language/extension-methods.md
```diff
@@ -220,6 +220,7 @@ extension NumberParsing on String {
 
 The members of an extension can be methods, getters, setters, or operators.
 Extensions can also have static fields and static helper methods.
+Make sure to use the Extension name for invoking them.
```
parlough, line 223: Could you explain further what you are trying to suggest here? Are you suggesting using the Extension name explicitly to call the methods? Such as `NumberParsing('42').parseInt()`?

We do cover when you need to do in [Using extension methods](https://dart.dev/language/extension-methods#using-extension-methods) and its [API conflicts](https://dart.dev/language/extension-methods#api-conflicts) subcategory.

parlough (commented):
Thanks for opening a pull request!

I have some questions about the intent of the change:

src/language/extension-methods.md
parlough, line 223: When reading this, it wasn't clear that it was just about the static members. Perhaps we could call that out more directly? Also I believe they can be used without the extension name within the declaration, so we should account for that somehow.

How about something like:

```
To access static members outside of the extension declaration,
invoke them through the declaration name like [class variables and methods][].

[class variables and methods]: /language/classes#class-variables-and-methods
```

This text also references class static members to provide extra context and reading material.

parlough (commented):
Sorry about previous review, I misread the previous text, so I dismissed it.

Thanks for the PR! Great idea to clarify this.

A few potential suggestions to consider:

src/language/extension-methods.md
hrishikesh-kadam: Hi, @parlough! 😄 
Yes, it would be better to clarify that an Extension name is only required for static fields and static methods.
Please feel free to co-edit this PR.

parlough (approved):
Thank you!

PR #5174, opened 2023-09-15
Missing quotes in Update patterns.md

marcmcd:
Quotes must be added to change "first element is a or b:" where a and b seem to be variables which they are not, to "first element is 'a' or 'b':" where 'a' and 'b' are the correct string constants.

Recommended fix:

Change

You can nest [any kind of pattern][types] inside a destructuring pattern. 
For example, this case pattern matches and destructures a two-element
list whose first element is `a` or `b`:

<?code-excerpt "language/lib/patterns/destructuring.dart (nested-pattern)"?>
```dart
switch (list) {
  case ['a' || 'b', var c]:
    print(c);
}
```

to

You can nest [any kind of pattern][types] inside a destructuring pattern. 
For example, this case pattern matches and destructures a two-element
list whose first element is `'a'` or `'b'`:

<?code-excerpt "language/lib/patterns/destructuring.dart (nested-pattern)"?>
```dart
switch (list) {
  case ['a' || 'b', var c]:
    print(c);
}
```


parlough (approved):
Makes sense to me. Thanks for making this clearer!

Issue #5177, opened 2023-09-18
☂️ Migrate to 11ty


parlough:
An alternative to https://github.com/dart-lang/site-www/issues/4163 with a much lower up-front cost that still brings many of the benefits, greatly simplifying the site architecture and making it easier to add new features.

**Details to be filled out...**

parlough:
As an update for those following this: Now that I'm back from Prague, I've made good progress:

- A bunch of initial cleanup to make the migration easier has been completed (linked to this issue above).
- The site is running with 11ty locally and almost everything is functional.
  - Syntax highlighting is working but missing some of our custom functionality.
  - I'm implementing syntax highlighting in a way to solve https://github.com/dart-lang/site-www/issues/3846 while also reducing maintenance burden.
- Incremental rebuilding is supported but not working in all cases yet. I'll consider this as not a requirement, just a bonus for the initial migration.
- The recommended way to build the site will be without Docker, but it will still be there if desired or if running in environments like codespaces.
- Some cleanup, performance fixes, and general improvements will land after the primary change to make reviewing easier.

PR #5181, opened 2023-09-18
Remove unused settings in Jekyll config

parlough:
These settings aren't used for anything anymore, so they can safely be removed.


atsansone (approved):
@parlough : LGTM, but could you include something in the description that explains why these changes were needed or desirable? Otherwise, looking back on this PR will be confusing.

PR #5186, opened 2023-09-18
Simplify setup of `/guides` docs

parlough:
Moves all docs under `/_guides` to `/guides` then removes now unnecessary collection setup.


atsansone (approved):
Will all of the links work as before?

If so, LGTM.

parlough:
> Will all of the links work as before?

Yep. I manually verified the pages and checked pre-existing links to them as well.

PR #5189, opened 2023-09-19
Situate get-dart JS alongside other JS files

parlough:
This simplifies our site setup, as we work towards `/assets` being the standard location for all asset like files.

Contributes to https://github.com/dart-lang/site-www/issues/5177 by making asset pass-through easier to setup.


atsansone (approved):
Great!

PR #5196, opened 2023-09-19
Don't use the shared includes for now

parlough:
The usage of these includes on dart.dev and docs.flutter.dev have diverged and will continue to. This PR copies the used ones to the local includes folder. This will allow easier atomic modifications of the files. We can eventually move them back to site-shared if/when their behavior stabilizes and is sufficient for both sites.

Should slightly reduce build time as well.

Contributes to https://github.com/dart-lang/site-www/issues/5177 by allowing for changes which only one site will support temporarily.

### DIFF
--- a/src/_includes/breadcrumbs.html
+++ b/src/_includes/breadcrumbs.html
@@ -1,0 +1,33 @@
+{% comment %}
+Embeds breadcrumb RDFa, follows ARIA guidelines. References:
+- https://developers.google.com/search/docs/data-types/breadcrumb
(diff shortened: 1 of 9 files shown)


atsansone (approved):
Not sure if this will remain true, but it makes sense for now.

parlough:
> Not sure if this will remain true, but it makes sense for now.

Yeah, this is likely a temporary move, but will make some future work of mine much much easier to complete and to review. Thanks for taking a look!

PR #5205, opened 2023-09-19
Simplify `/articles` setup to be normally generated directory

parlough:
All pages within `/_articles` are generated, so this PR converts it to be setup just like all the other content directories.

As this removes our last collection, also uses a simpler (and faster) sitemap that just showcases all pages.


parlough:
The build will fail on this until https://github.com/dart-lang/site-www/pull/5180 lands.

atsansone (approved):
LGTM

PR #5213, opened 2023-09-21
Cleanup generated human readable index

parlough:
- Prevents the 404 page being included on the generated page
- Prevents non-HTML files (like CSS) being included
- Stops looping through `site.documents` since we don't use it (therefore it's empty)

Contributes to https://github.com/dart-lang/site-www/issues/5177 which won't support `site.documents` anyway

**Staged:** https://dart-dev--pr5213-misc-cleanup-human-r-kij4d07f.web.app/map

### DIFF
--- a/src/map.html
+++ b/src/map.html
@@ -6,15 +6,7 @@ toc: false
 
 <ul>
 {% for document in site.pages %}
-  {% if document.url and document.url != '/404' %}
-  <li>
-    <a href="{{document.url}}">{{ document.title | xml_escape }}</a><br>
-    <small>{{document.url}}</small>
-  </li>
-  {% endif %}
-{% endfor %}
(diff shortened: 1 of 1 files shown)


atsansone (approved):
Two questions:
1. Why do we need this?
2. Why not a table (other than Markdown issues)?

parlough:
> Why do we need this?

I'm not fully sure. I didn't even notice this page existed for the longest time. It could potentially help search engines find/promote new pages. It doesn't hurt to have though, plus it's helpful as a debugging tool.

> Why not a table (other than Markdown issues)?

I have no opinion since this page isn't super useful for users, and perhaps the simple unordered list is easier for tools to parse. We can always change it to a table if you prefer :)

Issue #4254, opened 2022-10-03
[3.2] Private field promotion docs


MaryaBelanger:
## This is an _evaluation_ ticket

Evaluating the possible need for doc updates due to dart-lang/language#2020

### Summary

Type promotion is currently only available for _local_ variables. [#2020](https://github.com/dart-lang/language/issues/2020) will make it possible for _class_ fields to be safely promoted if they meet certain criteria: when the field is `private` and `final` (and there are no concrete getters of the same name, in the same library). 

Limiting promotable fields to `final` tells the compiler the instance value can't change (and therefore, type promotion remains sound). But, `final` doesn't protect soundness if another class implements the original class and overrides its field declarations. So, the compiler scans the library for implementing classes and can prevent type promotion of the original because of it.

But what if a class outside of the library implements the defining class and overrides its fields? The compiler can't check for implementing classes outside of the original's library, which is why the field must also be `private`. A `private` field tells the compiler that an implementing class from another library isn't able to change the value. 

So `private` and `final` allow the compiler to soundly promote a type because they inherently mean the value cannot be overridden outside of the compiler's purview.

One loophole (`nosuchmethod` overriding interfaces in an implementing class that doesn't fulfill the interface contract, i.e. doesn't explicitly declare the fields of the original class) is covered in [#49687](https://github.com/dart-lang/sdk/issues/49687) and another evaluation ticket (https://github.com/dart-lang/site-www/issues/4263).

### Evaluation

- [Fixing type promotion failures](https://dart.dev/tools/non-promotion-reasons)
  - #5246 
  - The sections on this page are actually linked to in type promotion error messages. 
  - [Only local variables can be promoted](https://dart.dev/tools/non-promotion-reasons#property-or-this) is no longer an error
  - "Write captured by a local function", "Written outside of the current closure or function expression", "Write captured outside of the current closure or function expression" sections all have "create a local variable" as a solution / workaround, so I'm wondering if they'll still be errors once non-local variables can be promoted, or maybe the workaround will change to utilize private final field promotion
    -  "Other causes and workarounds" section also mentions creating a local variable to fix most type promotion issues

  - New sections will need to be added to support new error messages from 2020 (@stereotype441 evaluating actual error code):
    - When non-local field promotion is not allowed because they're using an older version of Dart
    - When promotion doesn't work because the field is not final and private
    - When there's a getter with the same name in the same library that isn't final
[...]

PR #5237, opened 2023-10-09
Point new promo go links to /non-promotion-reasons

MaryaBelanger:
Part of #4254 

I added the new private final field promotion go links to `firebase.json` so they at least point to the correct page. Since these are accessible to some (limited) users now it's good to have them go somewhere. CL here: https://dart-review.googlesource.com/c/sdk/+/325803/3/pkg/_fe_analyzer_shared/lib/src/flow_analysis/flow_analysis.dart#2799

I'd like to merge this ASAP, then change their `destination` fields to the correct new sections on the `/non-promotion-reasons` page once those are written (to do).


parlough (approved):
Sounds like a plan!

PR #5246, opened 2023-10-12
[3.2] Update "Fixing type promotion failures" for private final field promo

MaryaBelanger:
This is the work to add all the new field promotion failure anchors to the site. These are linked from context messages.


firebase.json
MaryaBelanger, line 1: @stereotype441 can you confirm the go links make sense / are all accounted for?

src/tools/non-promotion-reasons.md
```diff
@@ -111,16 +163,546 @@ one or more of the following:
...
+
+**Message:**
+
+```nocode
+'overriden' couldn’t be promoted because there is a conflicting getter in class 'Override'
+```
+
+**Solution**:
+
+*// what is the solution here? would you be able to rename them if they ARE related, too?*
```
MaryaBelanger, line 505: @stereotype441

src/tools/non-promotion-reasons.md
```diff
@@ -111,16 +163,546 @@ one or more of the following:
...
+
+**Message:**
+
+```nocode
+'overridden' couldn’t be promoted because there is a conflicting non-promotable field in class 'Override'.
+```
+
+**Solution:**
+
+*// Try renaming one of the conflicting fields? What if they ARE related?*
```
MaryaBelanger, line 619: @stereotype441

src/tools/non-promotion-reasons.md
```diff
@@ -111,16 +163,546 @@ one or more of the following:
...
+
+**Message:**
+
+```nocode
+'_i' couldn’t be promoted because there is a conflicting noSuchMethod forwarder in class 'MockExample'.
+```
+
+**Solution:**
+
+*// Is there a solution here? Maybe explicitly supporting/defining the getter in `MockExample` would fix it?*
```
MaryaBelanger, line 678: @stereotype441

MaryaBelanger (commented):
Solutions for the conflict failures: I'm not sure if just renaming them is applicable when the two conflicting entities _are_ related. Does the regular perscriptive advice apply to these three too? (i.e. create a local variable, use a null check, etc)

src/tools/non-promotion-reasons.md
```diff
@@ -111,16 +163,546 @@ one or more of the following:
...
+If the [`on` type][] of the extension method is a nullable type,
+you'd want to do a null check to see whether `this` is null:
+
+**Example:**
+
+{:.bad}
+```dart
+extension E on int? {
+  int? get i {
+    return this != null;
```
stereotype441, line 182: This example doesn't actually trigger a type promotion failure. It's just a straight up ordinary compilation error because the result of a `!=` check is a `bool`, not an `int?`.

I would recommend a slightly different example, perhaps something like:
```dart
extension E on int? {
  int get valueOrZero {
    return this == null ? 0 : this;
  }
}
```

The corresponding "good" code would be:
```dart
extension E on int? {
  int get valueOrZero {
    final self = this;
    return self == null ? 0 : self;
  }
}
```

src/tools/non-promotion-reasons.md
```diff
@@ -111,16 +163,546 @@ one or more of the following:
...
+**Solution:**
+
+Making the field private lets the compiler be sure that no outside libraries
+could possibly override its value, so it's safe to promote.
+
+{:.good}
+```dart
+class C {
+  final int? _n;
+  C(this.n);
```
stereotype441, line 253: Since you've renamed the field, you need to update references to it. This line needs to be `C(this._n);`.

The `test` method below needs to be:
```dart
test(C c) {
  if (c._n != null) {
    print(c._n + 1); // OK
  }
}
```

src/tools/non-promotion-reasons.md
```diff
@@ -111,16 +163,546 @@ one or more of the following:
...
+```
+
+**Solution:**
+
+Make the field `final`
+
+{:.good}
+```dart
+class Example {
+  int? final _immutablePrivateField;
```
stereotype441, line 303: `final` needs to go before the type, so this line should be `final int? _immutablePrivateField;`.

src/tools/non-promotion-reasons.md
```diff
@@ -3,60 +3,116 @@ title: Fixing type promotion failures
...
+  where `version` is lower than 3.2.
+
+**Example:**
+
+{:.bad}
+```dart
+// @dart=3.1
+
+class C {
+  final int? _i; // CONTEXT
```
stereotype441, line 37: I don't see any explanation of what this `// CONTEXT` comment means. Consider either:
- Adding an explanation somewhere that these `// CONTEXT` comments are to let the reader know where the context message will point to
- Or maybe just drop all the `// CONTEXT` comments in this file. After all, the user got to here by clicking on a link that was in a context message; they really don't need us to tell them where the context message pointed to.

I would personally favor dropping all the `// CONTEXT` comments.

src/tools/non-promotion-reasons.md
```diff
@@ -111,16 +163,546 @@ one or more of the following:
...
+but only instance fields can be promoted, not getters. 
+
+The compiler has no way to guarantee that a getter returns the same result every time.
+Because their stability can't be confirmed, getters are not safe to promote.
+
+**Example:**
+
+{:.bad}
+```dart
+abstract class C {
```
stereotype441, line 326: We're planning to make a breaking change to the language in 3.3 that's going to render this example invalid. (Details, if you're curious, are here: https://github.com/dart-lang/language/issues/3328#issuecomment-1792511446). We should probably tweak this example so that the breaking change won't affect it.

Here's an example you can use:
```dart
import 'dart:math';
abstract class C {
  int? get _i => Random().nextBool() ? 123 : null; // CONTEXT
}

f(C c) {
  if (c._i != null) {
    pint(c._i.isEven); // ERROR
  }
}
```

And the corresponding "good" code would be:
```dart
import 'dart:math';
abstract class C {
  int? get _i => Random().nextBool() ? 123 : null;
}

f(C c) {
  final i = c._i;
  if (i != null) {
    print(i.isEven);
  }
}
```

(Note: I switched to printing whether the integer is even rather than trying to assign it to an `int` variable, just so that I wouldn't have to name the new variable `j`).

src/tools/non-promotion-reasons.md
```diff
@@ -111,16 +163,546 @@ one or more of the following:
...
+
+class Derived extends Base {
+  @override
+  final int? _i;
+  Derived(this._i);
+}
+
+f(Base b) {
+  if (b._i != null) {
+    int i = b._i; // ERROR
```
stereotype441, line 385: When we make the breaking change described in https://github.com/dart-lang/language/issues/3328#issuecomment-1792511446, this error will go away.

You might want to consider just dropping this subsection, since it's a subtlety that most people are not going to run into, and it will stop being a problem in a few months.

If you do decide to keep it, maybe mention that this is a known bug and that we're planning to fix it in a future release.

src/tools/non-promotion-reasons.md
stereotype441: Good point. The only way to fix the error through renaming is to rename either the getter or the field but not both. If you've given them the same name on purpose because they're related (e.g. because one of them overrides the other, as in the above example), then renaming one of them but not the other will break the relationship between them, and your code will no longer do what you intend (it might not even compile).

Maybe we need to rewrite this section a bit. As it stands we've got the section "Conflict with getter elsewhere in library" that the user will see first, with an example where the classes are related, and then we have the subsection "Note about unrelated classes", with an example where the classes aren't related.

Maybe it would be better to lead with the second example, where the classes aren't related, and the solution is to do a rename. Then follow up with a subsection "But what if the classes are related?", and that subsection breaks the news to the user that if the two conflicting things are related, they're probably going to have to do the trick with a local final variable (as you've done in most of the other examples in this file).

Does that seem reasonable? I'm spitballing here.

src/tools/non-promotion-reasons.md
stereotype441: See my comment above 😃

src/tools/non-promotion-reasons.md
stereotype441: Yeah, I like that idea. We could give something like this as a solution:

```dart
import 'package:mockito/mockito.dart';

class Example {
  final int? _i;
  Example(this._i);
}

class MockExample extends Mock implements Example {
  @override
  late final int? _i;
}

f(Example x) {
  if (x._i != null) {
    int i = x._i;
  }
}
```

(I made the override `late` because that's consistent with the way mocks generally get used--first you construct them and then you tell them how they should behave.)

stereotype441 (commented):
This is looking great Marya! I think a couple things need some additional work, but nothing major. Let me know if any of my comments aren't clear.

src/tools/non-promotion-reasons.md
MaryaBelanger: Thank you! I tried simplifying the last example you gave me but messed it up in the process

src/tools/non-promotion-reasons.md
MaryaBelanger: Ok, I'll drop the section and add a temporary note.

src/tools/non-promotion-reasons.md
MaryaBelanger: Lmk if the note I added on [line 361](https://github.com/dart-lang/site-www/pull/5246/files#diff-68e8ac8e2954d6b7811e6c6a95e2fe171417e5fbd368a6f03f30f6ac9645fc1fR361-R367) sounds good!

stereotype441: Yeah, that looks totally reasonable.

stereotype441 (approved):
lgtm!

src/tools/non-promotion-reasons.md
```diff
@@ -111,16 +163,566 @@ one or more of the following:
...
+  final i = x._overridden;
+  if (i != null) {
+    print(i.isEven); // ERROR
+  }
+}
+```
+
+If the fields are unrelated, then simply rename one of the fields so they don't
+conflict.
+Read the [Note about unrelated clasess](#note-about-unrelated-classes).
```
parlough, line 616:
```suggestion
Read the [Note about unrelated clasess](#note-about-unrelated-classes). 
```

firebase.json
```diff
@@ -158,16 +158,17 @@
...
-      { "source": "/go/non-promo-non-final-field", "destination": "/tools/non-promotion-reasons", "type": 301 },
-      { "source": "/go/non-promo-property", "destination": "/tools/non-promotion-reasons#property-or-this", "type": 301 },
-      { "source": "/go/non-promo-public-field", "destination": "/tools/non-promotion-reasons", "type": 301 },
-      { "source": "/go/non-promo-this", "destination": "/tools/non-promotion-reasons#property-or-this", "type": 301 },
+      
+      { "source": "/go/non-promo-conflicting-getter", "destination": "/tools/non-promotion-reasons#getter-name", "type": 301 },
+      { "source": "/go/non-promo-conflicting-non-promotable-field", "destination": "/tools/non-promotion-reasons#field-name", "type": 301 },
+      { "source": "/go/non-promo-conflicting-noSuchMethod-forwarder", "destination": "/tools/non-promotion-reasons#nosuchmethod", "type": 301 },
+      { "source": "/go/non-promo-external-field", "destination": "/tools/non-promotion-reasons#external", "type": 301 },
+      { "source": "/go/non-promo-field-promotion-unavailable", "destination": "/tools/non-promotion-reasons##language-version", "type": 301 },
```
parlough, line 166:
```suggestion
      { "source": "/go/non-promo-field-promotion-unavailable", "destination": "/tools/non-promotion-reasons#language-version", "type": 301 },
```

src/tools/non-promotion-reasons.md
```diff
@@ -111,16 +163,566 @@ one or more of the following:
...
+
+The compiler creates this implicit implementation of `_i` because
+`MockExample` promises to support a getter for `_i` when it implements
+`Example` in its declaration, but doesn't fulfill that promise. 
+So, the undefined getter implementation is handled by [`Mock`'s `noSuchMethod`
+definition](https://pub.dev/documentation/mockito/latest/mockito/Mock/noSuchMethod.html),
+which creates an implicit `noSuchMethod` forwarder of the same name.
+
+The failure can also occur between fields in
+[unrelated clasess](#note-about-unrelated-classes).
```
parlough, line 662:
```suggestion
[unrelated classes](#note-about-unrelated-classes).
```

parlough (approved):
I haven't done a full proper review of this yet _(sorry)_, but quickly reading through it, it looks amazing and will be a great resource for those trying to better understand type promotion failures.

I did make sure to test the links though and they all work as expected.

A few minor adjustments before landing though:

MaryaBelanger:
Thank you Parker!

PR #5258, opened 2023-10-18
Update guard clause content

### LINKED ISSUE #5255: Document pattern guards

The docs for patterns are excellent and very comprehensive, but I was talking to a couple of folks on the team and realized that they don't cover guard clauses. Not too much of a surprise given how much other stuff there was to cover, but we should probably explain them too. :)

[Here is the relevant part of the spec](https://github.com/dart-lang/language/blob/main/accepted/3.0/patterns/feature-specification.md#guard-clause). Guards are allowed in switch statement cases, switch expression cases, and if-case statements:

```dart
// Switch statement:
switch (something) {
  case somePattern when some || boolean || expression:
    //             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Guard clause.
    body;
}

// Switch expression:
var value = switch (something) {
  somePattern when some || boolean || expression => body,
  //               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Guard clause.
}

// If-case statement:
if (something case somePattern when some || boolean || expression) {
  //                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Guard clause.
  body;
}
```

MaryaBelanger:
Fixes #5255

### DIFF
--- a/examples/language/lib/control_flow/branches.dart
+++ b/examples/language/lib/control_flow/branches.dart
@@ -1,4 +1,4 @@
-// ignore_for_file: unused_local_variable
+// ignore_for_file: unused_local_variable, dead_code, unreachable_switch_case
 
 class Point {
   final int x;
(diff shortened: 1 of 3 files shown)


atsansone (approved):
LGTM. I believe you can do inline highlighting, but I may be wrong there. Ship it!

PR #5263, opened 2023-10-19
Exclude internal linter rules from generated docs

parlough:


parlough:
@pq Do you not want pages generated at all for internal linter rules? Or do you want them generated, just not listed/enumerated in lists of rules?

pq:
Thanks, Parker!

I'd say we not generate at all.

pq (approved):
🎉

Issue #4987, opened 2023-06-12
Rewrite 'Write command-line apps' tutorial


parlough:
### Page URL

https://dart.dev/tutorials/server/cmdline

### Page source

https://github.com/dart-lang/site-www/tree/main/src/_tutorials/server/cmdline.md

### Describe the problem

This is an old tutorial that could benefit from rewriting and a letter extra focus.

It is also the last use of tooltips, which we'd like to remove (https://github.com/dart-lang/site-www/issues/2316) to simplify the site infrastructure.

### Expected fix

Rewrite the tutorial, considering some stylistic choices from the more recent [Fetch data tutorial](https://dart.dev/tutorials/server/fetch-data). Keep it valuable and teach some basic concepts without using tooltips if possible.

atsansone:
@parlough : #2316 asks to delete _or_ use more consistently. I'd prefer the latter. They have some great applications that I don't think we've taken full advantage.

parlough:
I think we can and should revisit tooltips in code blocks in the future, especially since browsers are implementing native popover mechanisms, but currently this functionality is hard to maintain, write, properly test, and ensure accessibility.

I think for practical purposes in the meantime, it makes sense to drop its current use.

atsansone:
Agreed if we revisit later.

Issue #2316, opened 2020-03-25
Delete/improve+use create_code_with_tooltips.dart


kwalrath:
We apparently have some tooltips still in code, e.g. see #2315, which discusses some code at https://dart.dev/tutorials/server/cmdline#overview-of-the-dcat-app-code.

These code snippets are auto-generated and not included or tested.

We should fix this, one way or another.

PR #5271, opened 2023-10-22
Remove popover tooling and update CLI tutorial

parlough:
Continuing the goal of reducing site maintenance costs, removes the unmaintained popover support and updates the one remaining use to no longer be reliant on it.

This will also slightly reduce site loading time with less JS and CSS downloaded on every page :) 

We can revisit popover support in code snippets in the future if desired. Browsers are currently implementing [standardized popovers](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API) that are more accessible and easier to implement. They have shipped in WebKit and Chromium based browsers and are set to be enabled in Firefox soon.

Closes https://github.com/dart-lang/site-www/issues/2316
Fixes https://github.com/dart-lang/site-www/issues/4987

**Staged:** https://dart-dev--pr5271-feature-remove-popup-farcxcqp.web.app/tutorials/server/cmdline


atsansone (approved):
Agreed that it's not needed for now and was inconsistently implemented.

PR #5272, opened 2023-10-22
Reduce home page load time by lazy loading DartPad iframe

parlough:
[Lazy loads](https://web.dev/articles/iframe-lazy-loading) the DartPad iframe on the (bottom of the homepage) since it is almost always offscreen on first load and does not need to delay other parts of site loading. The browsers that respect this can choose to load it based on certain heuristics, so it may not make a difference, but across a few runs I saw the largest contentful paint time in lighthouse reduce by 1-5%.


atsansone:
@parlough : Could you add a summary to this PR? It looks good, but I have no way to know if it effective. Where can I see the change and what should the change affect? Does it load the page faster? Smoother?

atsansone (approved):
LGTM!

Are we looking at adding this to all video embeds, too?

PR #5281, opened 2023-10-24
[IGNORE] Proof of concept of migrating to 11ty

parlough:
This isn't done and there are many TODOs, documented both here and elsewhere, but it showcases the site at least building with 11ty.


parlough:
Superseded by https://github.com/dart-lang/site-www/pull/5284

PR #5282, opened 2023-10-24
Don't run tests on the dev channel for now to fix build

parlough:
This is temporary until either the next release or the whitespace change causing the failure is reverted from the underlying diagnostic in the SDK.

We don't rely on the `dev` testing much anyway, with most of our focus on `stable` and `beta`.


atsansone (approved):
LGTM

PR #5283, opened 2023-10-24
Update constructors page

### LINKED ISSUE #5265: Update 'Constructors' docs so first constructor shown is an idiomatic one
### Page URL

https://dart.dev/language/constructors.html

### Page source

https://github.com/dart-lang/site-www/tree/main/src/language/constructors.md

### Describe the problem

This is a sort of meta comment - the very first constructor shown on this page is not an idiomatic one, it looks like this:

```dart
class Point {
  double x = 0; 
  double y = 0;

  Point(double x, double y) {
    // See initializing formal parameters for a better way
    // to initialize instance variables.
    this.x = x;
    this.y = y;
  }
}
```

Note that it even has a comment showing how to better improve the constructor :).

My main concern is somebody comes to this page just trying to figure out how to write a basic constructor, copies the first example without really reading anything, and ends up using this pattern throughout their codebase, without ever finding out about initializing formal parameters, etc :).

### Expected fix

So, I think it would be better if the very first constructor we showed was one that we would be happy to see in any codebase, if that makes sense?

MaryaBelanger:
Fixes #5265


examples/misc/lib/language_tour/classes/point_alt.dart
```diff
@@ -4,18 +4,24 @@
...
+  Point(this.x, this.y);
+  // #enddocregion idiomatic-constructor
 
+  /* 
+  // #docregion constructor-long-way
+  
   Point(double x, double y) {
-    // See initializing formal parameters for a better way
-    // to initialize instance variables.
     this.x = x;
```
jakemac53, line 20: While we are here, in Dart you really shouldn't be assigning fields inside of constructor bodies ever, so it does feel weird for me to highlight this as an example. Especially with null safety, this often would force you into a `late` or nullable field.

This might be what some people are used to seeing in other languages, but you really don't ever want to do this in Dart. If you need to do some logic that you can't fit into the initializer list, you should create a factory constructor with that logic and then pass the computed values from there into a normal constructor.

What if we changed this example to use an explicit initializer list? That is the more canonical "long way" constructor.

MaryaBelanger: @jakemac53 I removed the longer example all together because I didn't think it was necessary to show a wrong way at all. 

But, if you think a lot of users default to writing out explicit initializer lists without realizing they don't have to, then I can add it back in as the longer "Instead of..." part of the example. What would the syntax for that look like though? Same as the [Initializer lists](https://dart.dev/language/constructors#initializer-list) section?

> In order to perform some logic that cannot be expressed in the initializer list, it is better to create a factory constructor (or static method) with that logic and then pass the computed values to a normal constructor.

I added that note to the end of the initializing formal parameters section, to point to factory constructors and static methods. Seems like it'd be helpful to know if you're reading the page from top to bottom and the first thing you see is the simplistic initializing formal parameters example, but let me know if that doesn't make sense, thanks!

jakemac53 (approved):

src/language/constructors.md
```diff
@@ -96,11 +97,6 @@ class Point {
   final double y;
 
   Point(this.x, this.y);
-
```
jakemac53, line 99: This section is about named constructors but now there isn't a named constructor :). Fwiw there wasn't anything wrong with this constructor (this is the "initializer list" style I was talking about above)

examples/misc/lib/language_tour/classes/point_alt.dart
jakemac53: I definitely think it makes a lot of sense for the simplistic example to be at the top. The transition between that and the next section is a bit awkward now though - maybe just add a comment on the first constructor, saying the next section will describe what it is doing?.

src/language/constructors.md
MaryaBelanger: Oops, that was just an accident! Lots of layered excerpts for different examples in that code 😅

src/language/constructors.md
```diff
@@ -43,29 +41,35 @@ The `this` keyword refers to the current instance.
...
 
+  Point(this.x, this.y);
   // Sets the x and y instance variables
   // before the constructor body runs.
-  Point(this.x, this.y);
 }
 ```
 
 The variables introduced by the initializing formals
 are implicitly final and only in scope of the initializer list.
```
parlough, line 66: Since we haven't introduced initializer lists yet, but this paragraph and (now) the next refer to them, can you update this usage to link to [its section](https://dart.dev/language/constructors#initializer-list).

This makes me think we should cover initializer lists earlier, since the named constructors section also references them. I'll leave that to a separate issue issue though and a link will be good enough for now :)

parlough (approved):
Thanks for working on this @MaryaBelanger! Looks great.

Just one suggestion related to an existing paragraph and its new sibling.

src/language/constructors.md
MaryaBelanger: Definitely will do. And yes please on the issue! Even just this small amount of content was tricky to rearrange, so that'll definitely need to be another PR 😅

PR #5310, opened 2023-11-02
Clarify use of `this` in initializing instance variables

pdhankhar61:
I found this conceptual bug or mistake. Checke here https://dart.dev/language/classes#instance-variables

This is original-
"If you initialize a non-late instance variable where it’s declared, the value is set when the instance is created, which is before the constructor and its initializer list execute. As a result, non-late instance variable initializers can’t access this."

Suggested Changes-
"If you initialize a non-late final instance variable where it’s declared, the value is set when the instance is created, which is before the constructor and its initializer list execute. As a result, non-late final instance variable initializers can’t access this."


src/language/classes.md
```diff
@@ -175,10 +175,10 @@ Non-final instance variables and
 an implicit *setter* method. For details,
 check out [Getters and setters][].
 
-If you initialize a non-`late` instance variable where it's declared,
+If you initialize a non-`late` `final` instance variable where it's declared,
```
eernstg, line 178: I'm not sure why you'd think that this doesn't initialize `i`:

```dart
class A {
  int i = 1;
}

void main() {
  print(A().i); // '1'.
}
```

eernstg (commented):
This change seems to reduce the coverage of the given text by removing some cases that should be included (namely: initialization of non-final instance variables).

pdhankhar61:
Respected,

I am talking about non-late but final instance variable which is
initialized at declaration time. In that case "this" keyword will not be
accessible.

class Point{
final double a;
final double b;

 //initialized at declaration
final double c=2.7;

Point(this.a,this.b,this.c);

}

In this case "this.c" will show error. Because it is already initialized.

But if "c" is not final then you can reassign it a new value.


Prince Kumar Dhankhar

src/language/classes.md
pdhankhar61: Respected,

I am talking about non-late but final instance variable which is initialized at declaration time. In that case "this" keyword will not be accessible.

```
class Point{
final double a;
final double b;

 //initialized at declaration
final double c=2.7;

Point(this.a,this.b,this.c);

}

```
In this case "this.c" will show error. Because it is already initialized.

But if "c" is not final then you can reassign it a new value.

MaryaBelanger: If I understand correctly, @pdhankhar61 you're issue is with the existing statement:

> non-late instance variable initializers can’t access this.

And you're saying that statement is incorrect because you tried accessing `this` for the non-late instance variables `a`, `b` and `c`, and it works:

```dart
class Point{
double a;
double b;

 //initialized at declaration
double c=2.7;

Point(this.a,this.b,this.c); // no errors, accessing `this` DOES work even though c is already initialized
}
```

But you found that statement *is* true if the variable is `final`, so you're updating the statement to specify `final`?

```dart
class Point{
final double a;
final double b;

 //initialized at declaration
final double c=2.7;

Point(this.a,this.b,this.c); // error - 'c' is final and was given a value when it was declared, so it can't be set to a new value.
}
```

I could be totally missing the point of the section, but I'm finding it confusing too. It also doesn't seem like the example code starting on line 183 is related to the paragraph immediately preceding it (the same paragraph in question), which makes it even more confusing. 

@eernstg (or anyone) So it's not just about initialization of non-late instance variables in general, but specifically that the paragraph makes it sound like you can't use `this` to initialize non-late instance variables -- why does it say that doesn't work?

pdhankhar61:
Respected,

Yes, you understood it correctly.


Prince Kumar Dhankhar

src/language/classes.md
pdhankhar61: Respected,

Yes, you understood it correctly.

eernstg (commented):
I think the comments I wrote a few days ago might not have been sent yet. (The behavior of Github reviewing isn't always obvious. Maybe this helps. ;-)

src/language/classes.md
pdhankhar61: Okay Now, I have understood you point @eernstg sir. Like that statement was confusing without example. 

Delay in response is because of github, because it sends notifications with a delay.

eernstg: Thanks, @pdhankhar61!

So what have we learned here? Taking a look at [the section](https://github.com/dart-lang/site-www/blob/4ff61a8486a80b14b8d5153b1b88c8693a942df7/src/language/classes.md#instance-variables) as a whole, the following comes to mind:

I think it's confusing that this section says 'All uninitialized instance variables have the value `null`', and it does not even mention that for many instance variables (namely the ones whose declared type is potentially non-nullable) it is simply a compile-time error if they are left uninitialized.

If we mention this then it also makes sense to mention the error, perhaps right after "can't access `this`":

```dart
double initialX = 1.5;

class Point {
  double? x = initialX; // OK, can access declarations that do not depend on `this`.
  double? y = this.x; // Compile-time error, cannot access `this`.
}
```

@MaryaBelanger, I understand that this kind of documentation should be extremely succinct and to the point, but what do you think about adding something like this?

We should probably also mention somewhere that an initializing formal isn't an expression and hence it isn't an "access to `this`" even in the case where it's written literally as `this.x`, but that might be better placed in a section where initializing formals are introduced (the same section does use one initializing formal, but only refers to it as 'a constructor parameter', so perhaps initializing formals aren't described here at all).

eernstg:
@MaryaBelanger, I unresolved the thread about line 178 and tried to summarize what we can learn from this discussion. What do you think would be the best way to proceed?

MaryaBelanger:
@eernstg I think the section would benefit from more explicit explanation, and possibly code showing the thing it's saying you _can't_ do, since that seems to be where we got confused.

@pdhankhar61 I'm going to add some commits to this PR to incorporate @eernstg's explanation. I'll request both your reviews again when I'm done and hopefully we can land this with you as the original author. :) Thank you for bringing this to our attention!

pdhankhar61:
Sounds great, thank you!

Prince Kumar Dhankhar

eernstg:
Hi @MaryaBelanger and @pdhankhar61, I'm not sure whether anyone is waiting for input from me, so let me just refer to a comment that I gave a while ago:

> It's a bit like seeing "You can't go faster than 25MPH here!" and suggesting that we should change it to "You can't go faster than 25MPH here on rainy days!". If the first sentence is true then the second one is true, too, but it isn't helpful to introduce a reduction of scope when the statement is true for the original, broader scope. It indirectly implies that the original statement wasn't true, but that's incorrect: It was true.
>
> Same structure here: The fact that there is no access to `this` in an initializing expression is true for all non-late instance variables, not just for final ones.

In short, I still think it's misleading to add the word 'final' in those two locations.

MaryaBelanger:
Hi @eernstg, sorry for the delay! I did see that quote was your conclusion and I was just working on the rewrite myself. Thanks for the reminder, will be done shortly!

MaryaBelanger:
Here's what I changed:

- Removed mentions of `final` added in the PR's first draft
- Changed the statement "All uninitialized instance variables have the value `null`" to account for non-nullable ([lines 170-172](https://github.com/dart-lang/site-www/pull/5310/files#diff-a8961bcc36d2bca8862ce398e1aa3ed05c42a0fee28361000835001dc8a8ebfdR170-R172))
- Moved the paragraph about `this`/initializers _below_ the code sample on [line 180](https://github.com/dart-lang/site-www/pull/5310/files#diff-a8961bcc36d2bca8862ce398e1aa3ed05c42a0fee28361000835001dc8a8ebfdR180), so that code sample could immediately follow [the paragraph about getters and setters](https://github.com/dart-lang/site-www/pull/5310/files#diff-a8961bcc36d2bca8862ce398e1aa3ed05c42a0fee28361000835001dc8a8ebfdR174-R178) that it's related to.
- Rewrote the [`this`/initializers paragraph](https://github.com/dart-lang/site-www/pull/5310/files#diff-a8961bcc36d2bca8862ce398e1aa3ed05c42a0fee28361000835001dc8a8ebfdR195-R199) slightly, to emphasize that it's referring to _initializing expressions_ (which I think is what @pdhankhar61 and I got confused about), and [added a code sample](https://github.com/dart-lang/site-www/pull/5310/files#diff-533d21108e7b2eb198096689e3e24a7d4f70d58c9453d0fcbf379497535226f3R2-R16) to visualize the paragraph.  


@eernstg please review the wording and code sample (mostly used your comments, but just to double check!) and @pdhankhar61 please let us know if you think the new code sample and wording is clearer than it was before, thank you!

examples/misc/lib/language_tour/classes/point_this.dart
```diff
@@ -0,0 +1,16 @@
...
+class Point {
+  double? x = initialX;
+  // OK, can access declarations that do not depend on `this`.
+  double? y = this.x;
+  // ERROR, can't access `this` in non-`late` initializer.
+  late double? z = this.x;
+  // OK, can access `this` in `late` initializer.
+
+  Point(this.x, this.y);
+  // OK, not an initializing expression, can access `this`.
```
eernstg, line 14: This one is difficult. ;-)

I think it's very useful to state explicitly that `this.x` and `this.y` are not expressions at all, not even the otherwise plausible construct that we call `<assignableExpression>` which is a specification of the target of an assignment (occurring on the left hand side of `=`).

If we had considered each of them as the left hand side of an assignment then we could be initializing an instance of a subclass of `Point` that declares `set x(int value) {...}`, and in that case we'd have to call that setter. That's not the actual semantics: `this.x` specifies that the actual argument is stored in the instance variable `x` in the current class body, it will never call an overriding setter. So `this.x` is not an expression and it's not an assignable expression.

In order to make that point even more firmly I think it's safer to avoid talking about "can access `this`" at all: that sounds a bit like "the semantics of `this.x` includes expression evaluation, and during that expression evaluation we can evaluate `this`, just like any other expression". And we don't want the reader to think like that.

Perhaps we could say something like this:

```suggestion
  // OK, `this.someName` is a parameter declaration, not an expression.
```

src/language/classes.md
```diff
@@ -195,6 +192,29 @@ void main() {
...
+  // OK, can access declarations that do not depend on `this`.
+  double? y = this.x;
+  // ERROR, can't access `this` in non-`late` initializer.
+  late double? z = this.x;
+  // OK, can access `this` in `late` initializer.
+
+  Point(this.x, this.y);
+  // OK, not an initializing expression, can access `this`.
+}
+```
```
eernstg, line 216: Similar comments as in examples/misc/lib/language_tour/classes/point_this.dart. (I guess that library and this snippet are kept in sync automatically).

src/language/classes.md
```diff
@@ -167,19 +167,16 @@ class Point {
 }
 ```
 
-All uninitialized instance variables have the value `null`.
+Uninitialized instance variables declared with
+[nullable types][] have the value `null`.
```
eernstg, line 171: I usually try to minimize the use of plural in situations like this, because of the tendency to introduce ambiguities. This could also be a matter of style/convention (it would be a bad idea to switch to singular if this location would then stand out as an exception). Otherwise:

```suggestion
An uninitialized instance variable declared with a
[nullable type][] has the value `null`.
```

src/language/classes.md
```diff
@@ -350,3 +370,5 @@ can pass a static method as a parameter to a constant constructor.
 [initializer list]: /language/constructors#initializer-list
 [factory constructor]: /language/constructors#factory-constructors
 [late-final-ivar]: /effective-dart/design#avoid-public-late-final-fields-without-initializers
+[nullable types]: /null-safety/understanding-null-safety#using-nullable-types
```
eernstg, line 373: If the corresponding change was made above:
```suggestion
[nullable type]: /null-safety/understanding-null-safety#using-nullable-types
```

eernstg (approved):
LGTM!

pdhankhar61:
Agreed!

Prince Kumar Dhankhar

examples/misc/lib/language_tour/classes/point_this.dart
```diff
@@ -0,0 +1,16 @@
+// ignore_for_file: invalid_reference_to_this, unnecessary_this
+// #docregion this-late
+double initialX = 1.5;
+
+class Point {
+  double? x = initialX;
+  // OK, can access declarations that do not depend on `this`.
```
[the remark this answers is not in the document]
MaryaBelanger: I originally had the comments on the same line, but when I save the file in VSCode it automatically breaks some lines into new lines in odd places...
image

It happens when I save the file, any idea why? I'm checking my VSCode settings now

src/language/classes.md
MaryaBelanger: Yeah they're linked, and since it's happening on Save I can't manually overwrite.

src/language/classes.md
MaryaBelanger: That's a great point, thank you!

examples/misc/lib/language_tour/classes/point_this.dart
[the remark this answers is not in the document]
MaryaBelanger: I just turned the "format on save" setting off, all set now!

parlough:
Thank you @pdhankhar61 and @MaryaBelanger for working on this!! I'm excited for the improvements to have landed :)

pdhankhar61:
Thanks to all of you too.

Prince Kumar Dhankhar

Issue #5296, opened 2023-10-27
Fix duplication of in-page ToC


atsansone:
a11y script found duplicate ID for the in-page ToC. The ToC is rendered twice: once for desktop layout and once for mobile layout. This raises a [WCAG F77 error](https://squizlabs.github.io/HTML_CodeSniffer/Standards/WCAG2/#sc4_1_1) on pages with in-page ToCs.  This causes issues with accessibility technology. This should be fixed.

PR #5317, opened 2023-11-04
Fix and remove duplicated `toc` ID

parlough:
We use a TOC generated from a plugin, so we need to do the replacement manually.

Fixes https://github.com/dart-lang/site-www/issues/5296
Fixes [#5302](https://github.com/dart-lang/site-www/issues/5302)


atsansone (approved):
No errors found on IDs with Wave. Approved.

Issue #5293, opened 2023-10-26
Change base colors to improve contrast


atsansone:
Change any colors that don't match the Dart branding guidelines. This results in multiple violations of [WCAG 2.0 G18](https://www.w3.org/TR/WCAG20-TECHS/G18.html) and GAR criteria 1.4.

PR #5318, opened 2023-11-05
Update text colors and highlighting for increased contrast

parlough:
Also updates various colors to more closely align with Dart/Flutter brand guidelines.

Fixes https://github.com/dart-lang/site-www/issues/5293

### DIFF
--- a/src/_sass/core/_variables.scss
+++ b/src/_sass/core/_variables.scss
@@ -2,9 +2,12 @@
 // Colors
 $flutter-color-teal: #60caf6;
 $flutter-color-blue-500: #1389FD;
-$flutter-color-dark-blue: #075b9a;
+$flutter-color-blue-600: #0553B1;
+$flutter-color-blue-700: #043875;
+$flutter-color-blue-800: #042B59;
 $flutter-color-fuchsia: #FF2B5B;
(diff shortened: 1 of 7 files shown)


src/_sass/core/_variables.scss
```diff
@@ -2,9 +2,13 @@
 // Colors
 $flutter-color-teal: #60caf6;
```
atsansone, line 3: These colors should match [those provided for Flutter](https://github.com/flutter/website/blob/main/src/_sass/base/_variables.scss#L2-L19). This would be consistent with the Dart design guidelines and Design's latest advice for the Flutter website.

atsansone (changes requested):
Need to match the Dart design guidelines for color.

src/_sass/core/_variables.scss
parlough: I can update this teal, but the other ones added come from the brand/design guidelines. Any in particular you're worried about?

parlough:
@atsansone These colors all come from the branding guidelines. I've gone ahead and updated a few others to match.

Can we land this for the contrast fixes and any further branding alignment can happen as follow up?

atsansone (approved):
LGTM.

PR #5319, opened 2023-11-05
Avoid duplicated search box IDs

parlough:
The `name` field on an input is used for submitting the search, so it's okay for the IDs to be different, but the names to remain the same.

Fixes https://github.com/dart-lang/site-www/issues/5295


atsansone (approved):
No error found using Wave or Lighthouse. Approved.

Issue #5247, opened 2023-10-12
☂️ Consistent terms for language features


MaryaBelanger:
### What information is missing?

We need consistent and user-friendly terminology across our docs for various language features. We should start standardizing the terminology we use for topics/ideas that have more than one word/phrase to describe them, by choosing one, defining, and sticking to it. A couple of examples:

- The `!` operator following an expression that will throw an exception if the value of the expression is `null`. It gets called several things like `null assert operator` and `non-null operator`.
  - [ ] #5252 
- Instance fields are sometimes called 'fields', sometimes 'variables', and sometimes 'instance' is dropped because it's assumed to be clear from context (which might sometimes be valid).
  - [ ] (conversation below)
- Very subtle but meaningful differences like subtype vs. subclass
  - [ ] dart-lang/sdk#59464 
- Function vs. method
  - [x] dart-lang/sdk#58820 
- (future) `implements` for classes vs extension types (probably)
- etc.

The concern is that having different terms might be confusing and definitely makes searching harder.

### How would you like us to fix this problem?

- Add to [glossary](https://dart.dev/resources/glossary) page, possibly include synonyms for variable terms
- Once standards are defined, go through the site and make sure the same term is used consistently (as much as possible)

parlough:
I thought a lot about the non-null assertion operator yesterday: https://github.com/dart-lang/linter/issues/3008#issuecomment-1758424665. I'd really like the team to come to a conclusion for that one, and it's better to solidify earlier rather than later.

parlough:
In terms of the `non-null assertion operator`, maybe we just need to make a `dart-lang/language` issue, ping the entire language team, and get that discussion over with. In the end, we just need a term. I am biased towards `non-null assertion operator` since it's familiar to developers coming from other languages, but I will accept anything.

MaryaBelanger:
Wow, thanks for the write up! 

> I'm not sure if it's a coincidence?

Not a coincidence, @bwilkerson just told me that this and similar conversations have been going on.

> maybe we just need to make a dart-lang/language issue, ping the entire language team, and get that discussion over with

I think you should definitely create that issue with basically all the text from your comment. 

Another option is I could create a PR _here_, and ping the language team / others on the team to review it? I'll take your choice, define it in the glossary, and go through the site and replace the existing terminology. PR for a couple reasons:

- people get way more involved in PRs
- would let people be able to see the change in action / evaluate it differently.

What do you think?

parlough:
[...]

PR #5324, opened 2023-11-07
[3.2] Update effective dart mention of "only local variables can be promoted"

MaryaBelanger:
This is the same as #4272 but that was stale (not requesting Bob's review again since he approved that one), plus added more links to more info.

### DIFF
--- a/src/effective-dart/usage.md
+++ b/src/effective-dart/usage.md
@@ -353,10 +353,8 @@ to a non-nullable type. That lets you access members on the variable and pass it
 to functions expecting a non-nullable type.
 
 Type promotion is only supported, however, for local variables, parameters, and
-private final fields. Values that are open to manipulation, whether
-directly, like top-level variables, or indirectly, like fields with a concrete
-getter of the same name in the same library, can't be type promoted.
-Read [Fixing type promotion failures][] for a full list of non-promotion reasons.
+private final fields. Values that are open to manipulation
+[can't be type promoted][].
 
 Declaring members [private][] and [final][], as we generally recommend, is often
 enough to bypass these limitations. But, that's not always an option.
@@ -414,7 +412,7 @@ local variable [`final`][] can prevent such mistakes.) Also, if the field might
 change while the local is still in scope, then the local might have a stale
 value. Sometimes it's best to simply [use `!`][] on the field.
 
-[Fixing type promotion failures]: /tools/non-promotion-reasons
+[can't be type promoted]: /tools/non-promotion-reasons
 [private]: /effective-dart/design#prefer-making-declarations-private
 [final]: /effective-dart/design#prefer-making-fields-and-top-level-variables-final
 [`final`]: /effective-dart/usage#do-follow-a-consistent-rule-for-var-and-final-on-local-variables
--- a/src/effective-dart/usage.md
+++ b/src/effective-dart/usage.md
@@ -352,16 +352,17 @@ Checking that a nullable variable is not equal to `null` promotes the variable
 to a non-nullable type. That lets you access members on the variable and pass it
 to functions expecting a non-nullable type.
 
-Type promotion is only sound, however, for local variables, parameters, and
-private final fields. Instances that are open to manipulation, whether
-explicitly, like top-level variables, or implicitly, like fields with a concrete
+Type promotion is only supported, however, for local variables, parameters, and
+private final fields. Values that are open to manipulation, whether
+directly, like top-level variables, or indirectly, like fields with a concrete
 getter of the same name in the same library, cannot be type promoted.
 Read [Fixing type promotion failures][] for a full list of non-promotion reasons.
 
-Declaring members [private] and [final], as we generally recommend, bypasses
-these limitations, but that's not always an option. One pattern to work around
-this is to assign the field's value to a local variable. Null checks on that
-variable will promote, so you can safely treat it as non-nullable.
+Declaring members [private] and [final], as we generally recommend, is often
+enough to bypass these limitations. But, that's not always an option.
+One pattern to work around this is to assign the field's value
+to a local variable. Null checks on that variable will promote,
+so you can safely treat it as non-nullable.
 
 {:.good}
 <?code-excerpt "usage_good.dart (shadow-nullable-field)"?>
@@ -385,7 +386,7 @@ class UploadException {
 ```
 
 Assigning to a local variable can be cleaner and safer than using `!` every
-place the instance is used:
+place the value is used:
 
 {:.bad}
 <?code-excerpt "usage_bad.dart (shadow-nullable-field)" replace="/!\./[!!!]./g"?>
--- a/src/effective-dart/usage.md
+++ b/src/effective-dart/usage.md
@@ -350,13 +350,18 @@ then it probably does make sense to have a separate boolean field.
 
 Checking that a nullable variable is not equal to `null` promotes the variable
 to a non-nullable type. That lets you access members on the variable and pass it
-to functions expecting a non-nullable type. Unfortunately, promotion is only
-sound for local variables and parameters, so fields and top-level variables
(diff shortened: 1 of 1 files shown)


src/effective-dart/usage.md
```diff
@@ -350,13 +350,18 @@ then it probably does make sense to have a separate boolean field.
...
-to functions expecting a non-nullable type. Unfortunately, promotion is only
-sound for local variables and parameters, so fields and top-level variables
-aren't promoted.
+to functions expecting a non-nullable type.
 
-One pattern to work around this is to assign the field's value to a local
-variable. Null checks on that variable do promote, so you can safely treat
-it as non-nullable.
+Type promotion is only sound, however, for local variables, parameters, and
+private final fields. Instances that are open to manipulation, whether
```
stereotype441, line 356: I found this a bit confusing because "instances" has a technical meaning that's different from what you're trying to communicate here. E.g., an object on the heap is often said to be an "instance" of a certain class. But we're not talking about heap objects being manipulated here, we're talking about things like non-final fields, which might _refer_ to different heap objects from time to time, and getters, which might return a different heap reference from time to time.

Maybe change "instances" to "values"? It's not really any more precise than "instances", but at least it avoids the confusing connotation.

src/effective-dart/usage.md
```diff
@@ -350,13 +350,18 @@ then it probably does make sense to have a separate boolean field.
...
 to a non-nullable type. That lets you access members on the variable and pass it
-to functions expecting a non-nullable type. Unfortunately, promotion is only
-sound for local variables and parameters, so fields and top-level variables
-aren't promoted.
+to functions expecting a non-nullable type.
 
-One pattern to work around this is to assign the field's value to a local
-variable. Null checks on that variable do promote, so you can safely treat
-it as non-nullable.
+Type promotion is only sound, however, for local variables, parameters, and
```
stereotype441, line 355: This is a little bit of a fib because type promotion _would_ be sound for top-level variables, provided that they're final. We just didn't implement type promotion for top-level variables because we didn't think it would be useful enough to justify the engineering effort.

Maybe we should just change "sound" to "supported"?

src/effective-dart/usage.md
```diff
@@ -350,13 +350,18 @@ then it probably does make sense to have a separate boolean field.
...
+private final fields. Instances that are open to manipulation, whether
+explicitly, like top-level variables, or implicitly, like fields with a concrete
```
stereotype441, line 357: Consider changing "explicitly" and "implicitly" to "directly" and "indirectly". Usually when we talk about something being implicit, we're talking about it being left out of the source code entirely (like how `var x = 0;` has an implicit type of `int`), whereas in this case there's nothing left out, it's just that the thing that can cause the value to change is somewhere else; hence, "indirect".

src/effective-dart/usage.md
```diff
@@ -350,13 +350,18 @@ then it probably does make sense to have a separate boolean field.
...
+getter of the same name in the same library, cannot be type promoted.
+Read [Fixing type promotion failures][] for a full list of non-promotion reasons.
+
+Declaring members [private] and [final], as we generally recommend, bypasses
```
stereotype441, line 361: Technically, declaring members private and final isn't sufficient to bypass a conflict with a getter. Maybe change "bypasses these limitations" to "can often bypass these limitations"?

src/effective-dart/usage.md
```diff
@@ -380,7 +385,7 @@ class UploadException {
 ```
 
 Assigning to a local variable can be cleaner and safer than using `!` every
-place the field or top-level variable is used:
+place the instance is used:
```
stereotype441, line 388: It took me a while to understand what "the instance" was referring to, due to the technical meaning of "instance" I mentioned earlier. If you follow my suggestion above and change "instance" to "value", then this sentence says "...safer than using `!` every place the value is used", which I think will be easier to follow.

stereotype441 (approved):
Lgtm! I made a bunch of nit-picky suggestions; it's totally your call whether you want to follow any of them.

MaryaBelanger:
Thanks for the review @stereotype441, and the lessons about terminology ("implicit", "explicit", and "instance" might be great candidates for the [glossary](https://github.com/dart-lang/site-www/issues/5247), linking the umbrella issue to keep in mind)

src/effective-dart/usage.md
```diff
@@ -380,7 +386,7 @@ class UploadException {
 ```
 
 Assigning to a local variable can be cleaner and safer than using `!` every
-place the field or top-level variable is used:
+place the value is used:
```
parlough, line 389: I understand why you made this change, but now the sentence reads slightly awkward with how close "using" and "used" are. 

Maybe we could be more specific? Since technically you can use the still nullable version without adding `!` as well. Plus it would avoid the extra "use" :)

```suggestion
time you need to treat the value as non-null:
```

src/effective-dart/usage.md
```diff
@@ -350,13 +350,19 @@ then it probably does make sense to have a separate boolean field.
...
-aren't promoted.
+to functions expecting a non-nullable type.
 
-One pattern to work around this is to assign the field's value to a local
-variable. Null checks on that variable do promote, so you can safely treat
-it as non-nullable.
+Type promotion is only supported, however, for local variables, parameters, and
+private final fields. Values that are open to manipulation, whether
+directly, like top-level variables, or indirectly, like fields with a concrete
+getter of the same name in the same library, cannot be type promoted.
```
parlough, line 358: > "indirectly, like fields with a concrete getter of the same name in the same library"

While Effective Dart is best read after finishing the language tour, it is still targeted at those newer to the language and/or programming. I think this sentence/concept might be too abstract or confusing for some of its targeted readers.

With extra context on why and an inline code example, such as a link to the corresponding entry in your updated non-promotion doc, it might be fine though.

However, since readers often won't make that page switch and it doesn't seem super essential for the point of this Effective Dart entry, what do you think about dropping the expansion of "Values that are open to manipulation"?

If you have related docs in "Understanding null safety" or type system docs (now or in future), perhaps a cross link to there could be added as an alternative.

src/effective-dart/usage.md
```diff
@@ -404,10 +410,15 @@ class UploadException {
...
 make sure that you don't write back to the local variable instead. (Making the
-local variable `final` can prevent such mistakes.) Also, if the field might
+local variable [`final`] can prevent such mistakes.) Also, if the field might
 change while the local is still in scope, then the local might have a stale
-value. Sometimes it's best to simply use `!` on the field.
+value. Sometimes it's best to simply [use `!`] on the field.
 
+[Fixing type promotion failures]: /tools/non-promotion-reasons
+[private]: design#prefer-making-declarations-private
+[final]: design#prefer-making-fields-and-top-level-variables-final
```
parlough, line 419: I guess we haven't really documented this anywhere yet (I will), but try to avoid relative links:

```suggestion
[private]: /effective-dart/design#prefer-making-declarations-private
[final]: /effective-dart/design#prefer-making-fields-and-top-level-variables-final
```

parlough (approved):
Thanks for expanding this @MaryaBelanger! Reads great.

I do have one concern about a complicated sentence and a few minor formatting suggestions, but nothing blocking :)

src/effective-dart/usage.md
MaryaBelanger: > what do you think about dropping the expansion of "Values that are open to manipulation"?

Done 👍 

> If you have related docs in "Understanding null safety" or type system docs (now or in future), perhaps a cross link to there could be added as an alternative.

I think the "fixing type promotion page" is the most complete list of reasons you can't type promote, so I treated your other comment about linking to that page in combination with this one: I removed the explicit "check out <that page>..." link and instead linked to it over the text `Values that are open to manipulation [can't be type promoted][].`

PR #5325, opened 2023-11-07
[3.2] Understanding null safety page updates for private final field promotion

MaryaBelanger:
---

Part of #4254 

---

Questions / notes: I'm not sure if the wording in these parts need to be slightly adjusted, or are unrelated.

- Line 405: 
    > **Top level variable and static field declarations must have an initializer.** Since these can be accessed and assigned from anywhere in the program, it’s impossible for the compiler to guarantee that the variable has been given a value before it gets used. The only safe option is to require the declaration itself to have an initializing expression that produces a value of the right type:

  - not sure if this statement is affected, because later is says that local variables are the most flexible case, I'm assuming because they can be type promoted? So are the rules about fields now also more flexible, at least for private final fields, since they can be promoted now too?

- [Late variables](https://dart.dev/null-safety/understanding-null-safety#late-variables)
  - "The most common place where the type checker cannot prove the safety of code is around top-level variables and fields." ... technically still true, maybe adjust?

### DIFF
--- a/src/null-safety/understanding-null-safety.md
+++ b/src/null-safety/understanding-null-safety.md
@@ -648,7 +648,7 @@ crashing.
 But if the value isn't `null`, it would be good to be able to move it over to
 the non-nullable side so you can call methods on it. Flow analysis is one of the
 primary ways to do this for local variables and parameters
-(and private, final fields, as of Dart 3.2). We've extended type
+(and private final fields, as of Dart 3.2). We've extended type
 promotion to also look at `== null` and `!= null` expressions.
 
 If you check a local variable with nullable type to see if it is not `null`, 
@@ -698,7 +698,7 @@ reasonable to figure that out statically, the analysis should be clever enough
 to do so.
 
 Note that type promotion originally only worked on local variables,
-and now also works on private, final fields as of Dart 3.2.
+and now also works on private final fields as of Dart 3.2.
 For more information about working with non-local variables,
 see [Working with nullable fields](#working-with-nullable-fields).
 
@@ -757,7 +757,7 @@ We've now corralled `null` into the set of nullable types. With flow analysis,
 we can safely let some non-`null` values hop over the fence to the non-nullable
 side where we can use them. That's a big step, but if we stop here, the
 resulting system is still painfully restrictive. Flow analysis only helps with
-locals, parameters, and private, final fields.
+locals, parameters, and private final fields.
 
 To try to regain as much of the flexibility that Dart had before null
 safety—and to go beyond it in some places—we have a handful of other
@@ -1534,7 +1534,7 @@ The core points to take away are:
     nullable operand to the underlying non-nullable type.
 
 *   Flow analysis lets you safely turn nullable local variables and parameters
-    (and private, final fields, as of Dart 3.2)
+    (and private final fields, as of Dart 3.2)
     into usable non-nullable ones. The new flow analysis also has smarter rules
     for type promotion, missing returns, unreachable code, and variable
     initialization.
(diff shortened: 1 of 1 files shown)


src/null-safety/understanding-null-safety.md
```diff
@@ -647,7 +647,8 @@ crashing.
 
 But if the value isn't `null`, it would be good to be able to move it over to
 the non-nullable side so you can call methods on it. Flow analysis is one of the
-primary ways to do this for local variables and parameters. We've extended type
+primary ways to do this for local variables and parameters
+(and private, final fields, as of Dart 3.2). We've extended type
```
stereotype441, line 651: Nit: I find "private, final fields" harder to scan than "private final fields". IMHO the comma isn't necessary, for the same reason that "delicious fried plantains" doesn't need a comma.

src/null-safety/understanding-null-safety.md
```diff
@@ -755,10 +757,10 @@ We've now corralled `null` into the set of nullable types. With flow analysis,
 we can safely let some non-`null` values hop over the fence to the non-nullable
 side where we can use them. That's a big step, but if we stop here, the
 resulting system is still painfully restrictive. Flow analysis only helps with
-locals and parameters.
+locals, parameters, and private, final fields.
```
stereotype441, line 760: This is an example of a sentence that I think is harder to follow due to the comma between "private" and "final". We're trying to say that flow analysis helps with 3 things, but because of the extra comma, my brain initially interprets it as 4 things, one of which is "private", which isn't a thing. Then I notice that the "and" is in the wrong place. And then finally I figure out it.

stereotype441:
> Part of #4254
> 
> Questions / notes: I'm not sure if the wording in these parts need to be slightly adjusted, or are unrelated.
> 
> * Line 405:
>   > **Top level variable and static field declarations must have an initializer.** Since these can be accessed and assigned from anywhere in the program, it’s impossible for the compiler to guarantee that the variable has been given a value before it gets used. The only safe option is to require the declaration itself to have an initializing expression that produces a value of the right type:
>   
>   
>   
>   * not sure if this statement is affected, because later is says that local variables are the most flexible case, I'm assuming because they can be type promoted? So are the rules about fields now also more flexible, at least for private final fields, since they can be promoted now too?

No change needed here. The additional flexability we're talking about here is the fact that since local variables can only be accessed inside the function that declares them, flow analysis can see all uses of a local variable and reason about when they occur. So in many cases you don't have to initialize a non-nullable local variable where you declare it; you just have to make sure to write a value to it before you try to read from it, and if flow analysis can prove that you've done that, it won't complain. Flow analysis isn't sophisticated enough to prove things about the ordering of reads and writes of top level variables, static fields, or instance fields, so for those things, if they have non-null types, you have to initialize them. Nothing about private field promotion changes that.

> * [Late variables](https://dart.dev/null-safety/understanding-null-safety#late-variables)
>   
>   * "The most common place where the type checker cannot prove the safety of code is around top-level variables and fields." ... technically still true, maybe adjust?

Still true. I don't think this sentence needs adjustment, but maybe I'm not understanding what you have in mind.

stereotype441 (approved):

src/null-safety/understanding-null-safety.md
MaryaBelanger: I agree now that you've pointed it out. I'll try to fix this in all the new PRs so don't worry about commenting every time you see it (it's probably a lot)

parlough (approved):
Great updates! Thanks @MaryaBelanger 🚀

PR #5329, opened 2023-11-08
[3.2] Very minor changes to null safety codelab

MaryaBelanger:
Just one addition of "fields can be promoted too!" (and fixed a messed up bullet list)


parlough (approved):
Nice catch!

PR #5330, opened 2023-11-09
[3.2] Language evolution

MaryaBelanger:
Added 3.2 section with language updates from the [changelog](https://github.com/dart-lang/sdk/blob/main/CHANGELOG.md#language)


src/guides/language/evolution.md
```diff
@@ -10,6 +10,8 @@ Dart programming language.
 * To learn specific details about the most recent supported language version,
   check out the [language documentation][] or the [language specification][].
 * For a full history of changes to the Dart SDK, see the [SDK changelog][].
+* For a full history of breaking changes, including [language versioned] changes,
+  check out the [Breaking changes][] page.
```
parlough, line 14:
```suggestion
* For a full history of breaking changes,
  including [language versioned][] changes,
  check out the [Breaking changes][] page.
```

src/guides/language/evolution.md
```diff
@@ -37,6 +40,50 @@ environment:
 
 ## Changes in each release
 
+### Dart 3.2
+_Released 15 November 2023_
+| [Dart 3.2 announcement](https://medium.com/dartlang/)
```
parlough, line 45: Leaving comment so we don't forget to update :)

src/guides/language/evolution.md
```diff
@@ -37,6 +40,50 @@ environment:
...
+      }
+    }
+  }
+
+  // Private field promotions also work from outside of the class:
+  f(Example1 x) {
+    if (x._privateField != null) {
+      int i = x._privateField; // OK
+    }
+  }
```
parlough, line 74: I'm assuming it wasn't intended for there to be a mix of `Example` and `Example1`.

Fixes to that and formatting/indentation:

```suggestion
  class Example {
    final int? _privateField;
  
    Example(this._privateField);
  
    void f() {
      if (_privateField != null) {
        // _privateField has now been promoted; you can use it without
        // null checking it.
        int i = _privateField; // OK
      }
    }
  }
  
  // Private field promotions also work from outside of the class:
  void f(Example x) {
    if (x._privateField != null) {
      int i = x._privateField; // OK
    }
  }
```

src/guides/language/evolution.md
```diff
@@ -37,6 +40,50 @@ environment:
...
+  ```
+  
+  For more information on when private final fields can and can't promote, check
+  out [Fixing type promotion failures](/tools/non-promotion-reasons).
+
+* Corrected inconsistencies in type promotion behvaior of
```
parlough, line 80:
```suggestion
* Corrected inconsistencies in type promotion behavior of
```

src/guides/language/evolution.md
```diff
@@ -37,6 +40,50 @@ environment:
...
+  [if-case](/language/branches#if-case) statements
+  where the scrutinee throws an exception.
```
parlough, line 82: I don't think we used the term "scrutinee" in our pattern matching docs, so likely want to avoid it here as Dart developers might not be familiar with it.

Taking a look at your branching and pattern docs, it seems you mostly referred to the scrutinee as a "value" and added context to explain. So perhaps "_where the value being matched_ throws an exception"?

parlough (approved):
Woo! Thanks for working on this @MaryaBelanger 🎸 

I have a few fixes and suggestions you should take a look at, but I don't need to take another look. Thanks again :)

PR #5331, opened 2023-11-09
[3.2] Update breaking changes page

MaryaBelanger:
### Do not merge until 3.2

---

- Moved "not yet released to stable" contents to "released in 3.2" section, and added newer entries from the changelog. 
- Added 3.3.0 changelog contents to "not yet released in stable"

fyi @leafpetersen

### DIFF
--- a/src/resources/breaking-changes.md
+++ b/src/resources/breaking-changes.md
@@ -47,49 +47,21 @@ To be notified about future breaking changes, join the [Dart announce][] group.
 
 ## Not yet released to stable
 
-### Language
-{: .no_toc}
-
-* **Language versioned**: [Changed the split point for refutable patterns][53167]
-  to the top level pattern so type promotion in if-case statements is consistent
-  regardless of whether the scrutinee might throw an exception.
-
 ### Libraries
 {: .no_toc}
 
-#### `dart:convert`
+#### `dart:nativewrappers`
 
-* [Changed return types of `utf8.encode()` and `Utf8Codec.encode()`][52801]
- from `List<int>` to `Uint8List`.
+* [Marked classes belonging to `NativeWrapperClass` as `base`][51896] so that
+  none of their subtypes can be implemented.
 
-#### `dart:developer`
+#### `dart:typed_data`
 
-* **Deprecated**: The `Service.getIsolateID` method.
-
-#### `dart:io`
-
-* [Eliminated trailing whitespace from HTTP headers][53005].
-
-#### `dart:js_interop`
-
-* **Experimental** **Removed**: `JSNumber.toDart` in favor of `toDartDouble` and
-  `toDartInt`.
-* **Experimental** **Removed**: `Object.toJS` in favor of `Object.toJSBox.`
-
-### Tools
-{: .no_toc}
-
-#### Dart Dev Compiler (DDC) and Dart2js
-
-* [Added interceptors for JavaScript `Symbol` and `BigInt` types][53106];
-  they should no longer be used with `package:js` classes.
+* **Deprecated**: [Unmodifiable view classes for typed data][53218].
 
 
-[53167]: https://github.com/dart-lang/sdk/issues/53167
-[52801]: https://github.com/dart-lang/sdk/issues/52801
-[53005]: https://github.com/dart-lang/sdk/issues/53005
-[53106]: https://github.com/dart-lang/sdk/issues/53106
-
+[51896]: https://github.com/dart-lang/sdk/issues/51896
+[53218]: https://github.com/dart-lang/sdk/issues/53218
 
 {% comment %}
 Create new section from these headers for each release.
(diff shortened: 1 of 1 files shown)


parlough (commented):
Thanks for working on this @MaryaBelanger, looks great!

I think a few additional adjustments might be nice:

- It depends a bit on what you want to call a breaking change, but private final field promotion can result in the [`unnecessary_non_null_assertion`](https://dart.dev/tools/diagnostic-messages#unnecessary_non_null_assertion) diagnostic triggering on old code that passed analysis, but with the new language version now has a warning due to the unnecessary use of `!` or similar. This can technically result in CI failures. @bwilkerson might have more well-defined thoughts on if scenarios like these are worth calling out as breaking changes.
- I think DDC and dart2js entries should referenced as `Development JavaScript compiler (DDC)` and `Production JavaScript compiler (dart2js)` respectively since the underlying compiler names are less and less visible to users.
- While [`waitFor`](https://api.dart.dev/stable/3.1.5/dart-cli/waitFor.html) has been deprecated for many releases, before 3.2 it just said "This functionality is incomplete and may be removed in a later version". Now in [3.2](https://github.com/dart-lang/sdk/commit/61b03b49b1b34b4757e47ce9778a80d68dd18694) it explicitly states it "is deprecated and will be removed in Dart 3.4". I think it would be good to call this deprecation out again so people are aware (https://github.com/dart-lang/sdk/issues/52121).

MaryaBelanger:
> It depends a bit on what you want to call a breaking change...

Interesting, also curious to hear if @bwilkerson or maybe @stereotype441 thinks it'd be useful and/or accurate to include that info here.

> I think DDC and dart2js entries...

Do you think I should update the whole page or just this release? If their underlying names were better known at some point maybe it makes sense to leave old entries since this is somewhat of a chronological reference.

> While [waitFor](https://api.dart.dev/stable/3.1.5/dart-cli/waitFor.html) has been deprecated for many releases...

Thank you! Nice catch

parlough:
>DDC/dart2js: Do you think I should update the whole page or just this release?

I think it's fine to keep the change to this release and going forward, but the repositioning as more of an implementation detail has been going on for a few releases.

MaryaBelanger:
> I would personally say it's worth a mention,

I agree with that sentiment; this page is for users who'd expect to be alerted of something like that. If someone thinks it shouldn't be mentioned, I can take it out but until then it's in :)

stereotype441:
> > It depends a bit on what you want to call a breaking change...
> 
> Interesting, also curious to hear if @bwilkerson or maybe @stereotype441 thinks it'd be useful and/or accurate to include that info here.

Looking at the top of the file, we do say "This page includes the following types of breaking changes ... Language versioned". So I definitely agree with @parlough that it's worth calling out the change. As to what we should say about it, I'm not sure. Technically, any change to type analysis can potentially cause any number of errors, warnings, or lints to occur, because a change in a type in one place can cause the meaning of the code that follows it to change. For example, when we turned the feature on in our internal code base, there was _one_ compile error (out of millions of lines of Dart code), because the type of a local variable changed. Here's a contrived example of how that might occur:
```dart
class C {
  final int? _x;
  C(this._x);

  f() {
    if (_x != null) {
      var y = _x; // Used to have type `int?`, now has type `int`
      ...
      y = null; // Used to be ok, now is an error
    }
  }
}
```
But it's so rare for something like this to crop up that I'm not sure whether it's worth mentioning; I worry that it would be a lot of wasted mental effort for people to understand this example, and why it's broken, when there's such a tiny chance of their own code having similar problems.

In practice, we did see a fair number of these three kinds of warnings:

- unnecessary_non_null_assertion
- invalid_null_aware_operator
- unnecessary_cast

Here's an example that illustrates all three:
```dart
class C {
  final num? _x = null;

  void test() {
    if (_x != null) {
      print(_x! * 2); // unnecessary_non_null_assertion
      print(_x?.abs()); // invalid_null_aware_operator
    }
    if (_x is int) {
      print((_x as int).bitLength); // unnecessary_cast
    }
  }
}
```

parlough:
I love the example that Paul used to illustrate all three potential warnings that may start being triggered. Can we include that @MaryaBelanger?

MaryaBelanger:
> I love the example that Paul used to illustrate all three potential warnings that may start being triggered. Can we include that @MaryaBelanger or does it not fit this page?

Added it! This page is supposed to eventually be like Flutter's breaking changes page, where they link out to longer form explanations and migrations, so that kind of example is actually exactly with the purpose of this content (and since we don't utilize those kinds of extra-page-explanations yet, sticking it right on the page is the next best thing :))

parlough (approved):
Thanks for this @MaryaBelanger and those adjustments :D 

Feel free to land this one whenever, it doesn't need to wait for the 3.2 release.

Issue #3720, opened 2021-12-09
Update content in 'Concurrency in Dart' page


kwalrath:
### Page URL

https://dart.dev/language/concurrency

### Page source

https://github.com/dart-lang/site-www/tree/master/src/_guides/language/concurrency/index.md

### Describe the problem

Definitely:
- [x] Convert code snippets to use code excerpts. (@parlough)
- [x] Add this page to the sidenav. Link to it appropriately. (https://github.com/dart-lang/site-www/pull/3850)
- [ ] Figure out how to save an editable version of the source that has the right fonts.

Maybe:
- [ ] Change diagrams to mention port/ReceivePort/SendPort.
- [ ] Add a figure (or code?) high up in the doc to make it more inviting, and to make clear that it isn't just words.
- [x] Give more information about performance, perhaps in its own page. https://github.com/dart-lang/site-www/pull/3696#discussion_r762007443 has some details. So does the [2.15 blog post](https://medium.com/dartlang/dart-2-15-7e7a598e508a).
- [x] Add a new macro & style for "Flutter note:" asides?
- [ ] Add a something to the page (and/or a FAQ entry) with the term _single-threaded language_. It might also say _Dart code executes in a predictable sequence that can’t be interrupted by other Dart code._

### Additional context

Followup to #3696.

PR #5332, opened 2023-11-09
Isolate examples documentation

ericwindmill:
This PR is being split into two. The first one can be found [here](https://github.com/dart-lang/site-www/pull/5402).

This PR updates documentation at language/concurrency with more information, and adds a page language/isolates with examples.

Fixes https://github.com/dart-lang/site-www/issues/3720

**Staged:**

- [Concurrency](https://dart-dev--pr5332-isolate-documentatio-gmqcme0e.web.app/language/concurrency) _(Updated)_
- [Isolates](https://dart-dev--pr5332-isolate-documentatio-gmqcme0e.web.app/language/isolates) _(New)_


ericwindmill:
@parlough Can you help me understand what's going on with the test failures? In the refreshing code excerpt part, it ends with this message, `Processed 203 Dart/Jade/Markdown files: 9 out of 937 fragments needed updating.`

But when I scroll through the output, there are only 8 messages that aren't INFO. So I'm guessing one of my newly added code excerpts is causing the problem, but I don't know how to find out which one.

examples/concurrency/lib/isolate_run_syntax.dart
```diff
@@ -0,0 +1,10 @@
+import 'dart:isolate';
+
+// #docregion
+int slowFib(int n) => n <= 1 ? 1 : slowFib(n - 1) + slowFib(n - 2);
+
+// Compute without blocking current isolate.
+void fib40() async {
+  await Isolate.run(() => slowFib(40));
```
lrhn, line 8: I'd do something with the value, even if just:
```dart
  var result = await Isolate.run(() => slowFib(40));
  print("Fib(40) = $result");
```

parlough:
> @parlough Can you help me understand what's going on with the test failures? It looks like it's code excerpt issues, but is there a way to get a verbose output?

It's unfortunately a bit obscure right now. I guess the original thinking was that people would just run the `./tool/refresh-code-excerpts.sh` script.

I'm working on finalizing an updated version of the tool that is simpler to run and will provide some more details on what needs to be updated. Until then, let me know if you need me to run the tool.

src/language/concurrency.md
```diff
@@ -16,90 +19,99 @@ prevpage:
...
+
+## Event Loop
+
+Dart’s runtime model is based on an event loop. This event loop is responsible
+for executing the code of your program, collecting and processing events,
+running garbage collection processes, and more. As your application runs, all
+events are added to a queue, called the event queue. Events can be anything from
+requests to repaint the UI, to user taps and keystrokes, to i/o from the disk.
+Because your app can’t predict what order these events will happen, it runs this
+event loop that processes events in the order they're queued, one at a time.
```
MaryaBelanger, line 43:
```suggestion
Dart’s runtime model is based on an event loop. 
The event loop is responsible for executing your program's code,
collecting and processing events, running garbage collection processes, and more.

As your application runs, all events are added to a queue,
called the *event queue*. Events can be anything from requests to repaint the UI,
to user taps and keystrokes, to I/O from the disk.
Because your app can’t predict what order events will happen, 
the event loop processes events in the order they're queued, one at a time.
```
(also just a minor restructure)

src/language/concurrency.md
```diff
@@ -16,90 +19,99 @@ prevpage:
...
+Implementing an event loop might resemble this:
 
-[isolates section]: #how-isolates-work
+```dart
+while (eventQueue.waitForEvent()) {
+  eventQueue.processNextEvent();
+}
+```
 
+Even though this event loop is synchronous and runs on a single thread, most
```
MaryaBelanger, line 55: "synchronous and runs on a single thread" -- is this the same as "this event loop is synchronous, meaning it runs on a single thread, ..." or are those two different things? (Haven't defined "synchronous" at this point, though it might be reasonably understood)

src/language/concurrency.md
```diff
@@ -16,90 +19,99 @@ prevpage:
...
+requests to repaint the UI, to user taps and keystrokes, to i/o from the disk.
+Because your app can’t predict what order these events will happen, it runs this
+event loop that processes events in the order they're queued, one at a time.
 
-## Asynchrony types and syntax
+![A figure showing events being fed, one by one, into the event loop](/assets/img/language/concurrency/event-loop.png)
 
-If you're already familiar with `Future`, `Stream`, and async-await,
-then you can skip ahead to the [isolates section][].
+Implementing an event loop might resemble this:
```
MaryaBelanger, line 47: This statement might be confusing for readers looking for practical information: implementing an event loop *might* resemble this? Is implementing an event loop something *I* need to do? Is it something that people actually do when writing Dart programs or is this just a visualization?

src/language/concurrency.md
```diff
@@ -16,90 +19,99 @@ prevpage:
...
-Many Dart apps use only one isolate, the _main isolate_.
-You can create additional isolates to enable
-parallel code execution on multiple processor cores.
-
-Although Dart's isolate model is built with underlying primitives
-such as processes and threads
-that the operating system provides,
-the Dart VM's use of these primitives
-is an implementation detail that this page doesn't discuss.
+Concurrent programming in Dart refers to both asynchronous APIs, like `Future`
```
MaryaBelanger, line 22: Somewhere at the start, I think there should be a clear purpose statement for the page, e.g. "This page contains a conceptual overview of Dart's options for concurrent programming" or something like that. "Conceptual overview" is also not the clearest statement, but something that tells the reader "this is reading/understanding/foundational info, not doing/practical/instructional info" before they dive in

src/language/concurrency.md
```diff
@@ -16,90 +19,99 @@ prevpage:
...
+while (eventQueue.waitForEvent()) {
+  eventQueue.processNextEvent();
+}
+```
 
+Even though this event loop is synchronous and runs on a single thread, most
+Dart applications need to do more than one thing at a time. For example, a
+client application might need to execute an HTTP request, while also listening
+for a user to tap a button. To handle this, Dart offers many async APIs, like
+Futures, Streams, and async-await. These APIs are built around this event loop.
```
MaryaBelanger, line 59:
```suggestion
[Futures, Streams, and async-await](/language/async). These APIs are built around this event loop.
```

src/language/concurrency.md
```diff
@@ -16,90 +19,99 @@ prevpage:
...
-represent values to be provided in the future.
-For example, a promise to eventually provide an `int` value
-is typed as `Future<int>`.
-A promise to provide a series of `int` values
-has the type `Stream<int>`.
+When this code is processed by the event loop, it will immediately call the
+first clause - `http.get`, and return a `Future`. It will also tell the event
+loop to take note that sometime in the future, a specific event will happen (
+which is in this case an HTTP response), and when that event happens, it should
+execute the callback passed to `then`.
```
MaryaBelanger, line 74:
```suggestion
When this code reaches the event loop, it immediately calls the
first clause -- `http.get`, and returns a `Future`. It also tells the event
loop to take note that sometime in the future, a specific event will happen
(in this case, an HTTP response), and when that event happens, it should
execute the callback passed to `then`.
```

Questions:
First sentence: Just from looking at the code, it's not obvious why it returns a Future. Is it `then` that causes that to happen? Or is Future the return type of `http.get`? 

Second sentence: it says it "*also* tells the event loop..." to expect a future, but isn't the code telling the event loop to expect a Future *because* the first clause returns a Future, not in addition-to? If that's the case, could change it to something like: "The returned `Future` tells the event loop to look out for a specific event to occur (in this case, ..."

Also, last clause: what's a callback? Why is it passed to `then`? Is the callback the result of the `Future`?

src/language/concurrency.md
```diff
@@ -16,90 +19,99 @@ prevpage:
...
-reads a file synchronously,
-blocking until the file is either fully read or an error occurs.
-The method then either returns an object of type `String`
-or throws an exception.
-The asynchronous equivalent, [`readAsString()`][],
-immediately returns an object of type `Future<String>`.
-At some point in the future,
-the `Future<String>` completes with either a string value or an error.
+![Figure showing async events being added to an event loop and holding onto 
+a callback to execute later.](/assets/img/language/concurrency/async-event-loop.png)
```
MaryaBelanger, line 77: The idea that the event loop "holds onto" the callback for later is an illuminating description that you should consider putting in the preceding text.

src/language/concurrency.md
```diff
@@ -16,90 +19,99 @@ prevpage:
...
 
-[`readAsStringSync()`]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-io/File/readAsStringSync.html
-[`readAsString()`]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-io/File/readAsString.html
+This same model is basically how the event loop handles all other 
+asynchronous events in Dart, such as [`Streams`][].
```
MaryaBelanger, line 80:
```suggestion
This same model is generally how the event loop handles all other 
asynchronous events in Dart, such as [`Streams`][].
```

src/language/concurrency.md
```diff
@@ -16,90 +19,99 @@ prevpage:
...
-Asynchronous computations are often the result of performing computations
-outside of the current Dart code; 
-this includes computations that don't complete immediately, 
-and where you aren't willing to block your Dart code waiting for the result.
-For example, an app might start an HTTP request,
-but need to update its display or respond to user input
-before the HTTP request completes.
-Asynchronous code helps apps stay responsive.
+A promise to eventually provide a string value is typed as `Future<String>`.
+This is an example of a function that returns a Future:
```
MaryaBelanger, line 94:
```suggestion
This is an example of a function that returns a `Future`:
```

src/language/concurrency.md
```diff
@@ -16,90 +19,99 @@ prevpage:
...
-or maybe just waiting for a timer to trigger. 
-All of these processes either run in a different thread, 
-or are handled by the operating system or the Dart runtime, 
-which allows Dart code to run concurrently with the computation.
+<?code-excerpt "lib/future_syntax.dart"?>
+```dart
+Future<String> _readFileAsync(String filename) {
+  final file = File(filename);
 
+  // .then() returns a Future
```
MaryaBelanger, line 101: There's the answer to my earlier question. So I guess you were just introducing a very high-level idea before and this section goes more in depth. I think you could either put this detail in that earlier paragraph, or at least say in that paragraph that the exact mechanics of what you're describing will be covered later (I prefer the former)

MaryaBelanger (commented):
Not done reviewing yet, but here's what I've got so far!

src/language/isolates.md
```diff
@@ -0,0 +1,784 @@
...
+this analogy, the StreamController is called a SendPort, and you can “add” 
+messages with the [`send()` method][]. ReceivePorts are the listeners, and 
+when these listeners receive a new message, they call a provided callback with the message as an argument.
+
+[`send()` method]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}
+/dart-isolate/SendPort/send.html
+
+#### Setting up ports
+
+Setting up a port requires a few steps, which must be done in a specific order. The following figure shows the conceptual steps required to spawn a new isolate and establish 2-way communication between it and the main isolate.
```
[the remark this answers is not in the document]
lrhn: Here is a more documented version, as a gist:
https://gist.github.com/lrhn/4426e19fc3511be5b15c03c3e8960676

src/language/concurrency.md
```diff
@@ -235,305 +268,268 @@ After handling the events, the isolate exits.
...
+using isolates.
+
+### Message types
+
+`message`s sent via SendPort can be almost any type of Dart object, but there
+are a few exceptions:
+
+- Objects with native resources (subclasses of e.g. NativeFieldWrapperClass1). A
+  Socket object for example refers internally to objects that have native
+  resources attached and can therefore not be sent.
```
MaryaBelanger, line 469:
```suggestion
- Objects with native resources
  - For example, subclasses of [`NativeFieldWrapperClass1`](), or a
    [`Socket`]() object that refers internally to objects that have native
    resources attached and can therefore not be sent.
```

I don't understand the given example(s ?), but basically just keep any elaboration in a sub-bullet and code font + link on any API mentions

src/language/concurrency.md
```diff
@@ -235,305 +268,268 @@ After handling the events, the isolate exits.
...
+- ReceivePort
+- DynamicLibrary
+- Finalizable
+- Finalizer
+- NativeFinalizer
+- Pointer
+- UserTag
+- MirrorReference
```
MaryaBelanger, line 477: Code font and link all of these

src/language/concurrency.md
```diff
@@ -235,305 +268,268 @@ After handling the events, the isolate exits.
...
+- Instances of classes that either themselves are marked with @pragma
+  ('vm:isolate-unsendable'), extend or implement such classes cannot be sent
+  through the ports.
+
+Apart from those exceptions any object can be sent. See documentation for
+  [`SendPort`][] for more information.
```
MaryaBelanger, line 483:
```suggestion
Apart from those exceptions, any object can be sent. 
Check out the [`SendPort`][] documentation for more information.
```

src/language/concurrency.md
```diff
@@ -235,305 +268,268 @@ After handling the events, the isolate exits.
...
-is slower when isolates are in different groups.
-
-[`Isolate.spawnUri()`]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-isolate/Isolate/spawnUri.html
-
-{{site.alert.flutter-note}}
-  Flutter doesn't support `Isolate.spawnUri()`.
-{{site.alert.end}}
-
 <a id="web"></a>
 ## Concurrency on the web
```
MaryaBelanger, line 488: I know this is another topic the Dart VM team heard user feedback about, wanting more practical examples. If that's covered in the main isolates page, link to it here.

MaryaBelanger (commented):
Finished the main concurrency page, will get to the new isolates page next. Thanks for working on this.

src/language/concurrency.md
```diff
@@ -235,305 +268,268 @@ After handling the events, the isolate exits.
...
-1. Spawns (starts and creates) an isolate
-2. Runs a function on the spawned isolate
-3. Captures the result
-4. Returns the result to the main isolate
-5. Terminates the isolate once work is complete
-6. Checks, captures, and throws exceptions and errors back to the main isolate
+For doing a single computation on a separate thread, use the static
+method `Isolate.run`. When you need to create an isolate that will handle
+multiple messages over time, or a background worker, use the `Isolate.spawn`
+method.
```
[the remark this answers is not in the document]
ericwindmill: I don't know the answer to your question, but "execute a single callback" is 100% accurate, so I'll change it to that.

src/language/concurrency.md
ericwindmill: It's not necessarily the same -- Dart's async APIs like Futures are asynchronous, but still run in a single thread

src/language/concurrency.md
ericwindmill: I'm not sure what you're referring to.

ericwindmill: I've also realized that this code comment is misleading. Both `readAsString` and `then` return Futures. `then` is a function on the Future class (that also returns a future.)  I updated that code comment, and I'll update the text if you can clarify it for me.

src/language/concurrency.md
```diff
@@ -235,305 +268,268 @@ After handling the events, the isolate exits.
...
+}
+```
 
-However, now the isolate sends a [closure][].
-Closures are less limited than typical named functions,
-both in how they function and how they're written into the code.
-In this example, `Isolate.run()` executes what looks like local code, concurrently.
-In that sense, you can imagine `run()` to work like a control flow operator
-for "run in parallel".
+## Performance and isolate groups
```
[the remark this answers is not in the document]
ericwindmill: If you're talking specifically about isolate groups, I think it should be a follow up issue.
I believe it's extremely niche for a Dart developer to need to know about isolate groups. When I surveyed the Insiders and the flutterdev reddit, and in my follow up conversations with a couple insiders who had advanced use-cases, isolate groups came up exactly one time (IIRC).

And less importantly, I don't know how to write those docs :) I wouldn't know how to make an example app that cared about Isolate groups, so I certainly wouldn't know how to talk about performance related to isolate groups.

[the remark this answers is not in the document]
ericwindmill: Also, FWIW, this section is copied exactly from the existing Dart documentation.  

Also, I realize I have little-to-no practical knowledge of isolate groups, but it does seem like the current paragraphs say everything there is to say. So the updates would largely just be code examples. 

I'm happy to do the follow up work, by the way. I just don't want to add more to this already large chunk of updates.

src/language/concurrency.md
ericwindmill: We do need to follow up on this, and John has even offered to help on the subject. 

In the Flutter PR about isolates, everyone agreed (including SWEs), that it's outside the scope of this PR, because on the web you don't use the Isolate API.

src/language/concurrency.md
```diff
@@ -16,87 +19,108 @@ prevpage:
...
+The event loop is what makes asynchronous and 
+concurrent programming possible in Dart.
 
-[isolates section]: #how-isolates-work
+## Event Loop
 
+Dart’s runtime model is based on an event loop. 
+The event loop is responsible for executing your program's code,
+collecting and processing events, 
+running garbage collection processes, and more.
```
MaryaBelanger, line 44: @ericwindmill Lasse said garbage collection probably isnt part of the event loop [here](https://github.com/dart-lang/site-www/pull/5332#discussion_r1402421621)

src/language/concurrency.md
```diff
@@ -16,87 +19,108 @@ prevpage:
...
 
-### Future and Stream types
+As your application runs, all events are added to a queue,
+called the *event queue*. 
+Events can be anything from requests to repaint the UI,
+to user taps and keystrokes, to I/O from the disk.
+Because your app can’t predict what order events will happen,
```
MaryaBelanger, line 50:
```suggestion
Because your app can’t predict the order events will happen, 
```

src/language/concurrency.md
```diff
@@ -16,87 +19,108 @@ prevpage:
...
-The Dart language and libraries use `Future` and `Stream` objects to
-represent values to be provided in the future.
-For example, a promise to eventually provide an `int` value
-is typed as `Future<int>`.
-A promise to provide a series of `int` values
-has the type `Stream<int>`.
+![A figure showing events being fed, one by one, into the
+event loop](/assets/img/language/concurrency/event-loop.png)
+
+The functioning of the event loop resembles this:
```
MaryaBelanger, line 56:
```suggestion
The way the event loop functions resembles this code:
```

src/language/concurrency.md
```diff
@@ -16,87 +19,108 @@ prevpage:
...
-reads a file synchronously,
-blocking until the file is either fully read or an error occurs.
-The method then either returns an object of type `String`
-or throws an exception.
-The asynchronous equivalent, [`readAsString()`][],
-immediately returns an object of type `Future<String>`.
-At some point in the future,
-the `Future<String>` completes with either a string value or an error.
+Even though this event loop is synchronous and runs on a single thread, most
+Dart applications need to do more than one thing at a time. For example, a
```
MaryaBelanger, line 65:
```suggestion
This example event loop is synchronous and runs on a single thread.
However, most Dart applications need to do more than one thing at a time. For example, a
```

src/language/concurrency.md
MaryaBelanger: Ok added another explanation, thanks for explaining

src/language/concurrency.md
```diff
@@ -16,87 +19,108 @@ prevpage:
...
+client application might need to execute an HTTP request, while also listening
+for a user to tap a button. To handle this, Dart offers many async APIs, like
+[Futures, Streams, and async-await](/language/async). 
+These APIs are built around this event loop.
 
-[`readAsStringSync()`]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-io/File/readAsStringSync.html
-[`readAsString()`]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-io/File/readAsString.html
+For example, consider making a network request:
```
MaryaBelanger, line 71:
```suggestion
For example, consider making a network request:

```
space between text and code blocks

examples/concurrency/lib/isolate_spawn_syntax.dart
```diff
@@ -0,0 +1,58 @@
...
+
+// #docregion
+class BackgroundWorker {
+  late Isolate _isolate;
+  late SendPort workerIsolateSendPort;
+
+  Future<void> initIsolate() async {
+    // Create a recieve port to recieve messages from spawned isolate
+    final mainIsolateReceivePort = ReceivePort();
+    mainIsolateReceivePort.listen((dynamic message) {
```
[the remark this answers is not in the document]
ericwindmill: This comment has really confused me, so I'll try to address different parts in order...

> Consider using forEach instead of listen.
Makes sense

> It's a little worrisome that there is no way to close the port. That means it will keep the current isolate alive forever.
There is a way to close the port, unless I'm misunderstanding something. Toward the bottom of the file, there's a `dispose` method which calls _isolate.kill

> Consider making this an async factory:
> rather than creating an uninitialized object with public uninitialized `late` fields. (That's generally discouraged.)
Makes sense

> Generally, I'm not sure what this is an example of. It's doing intricate setup, then only to do nothing but `print` on the results that come back. That dissonance irks me :)
This example is only showing syntax of how one would set up a long-lived isolate. It purposefully leaves off everything that isn't necessary.

> Personally, I'd probably use `Isolate.run` to start the other isolate, instad of `Isolate.spawn`, 
This example is under the heading "Long-lived isolates" and the file is called "isolate_spawn_syntax". Unless I'm missing something, it would not be appropriate to use Isolate.run here.


If I am missing something, which would most likely be that I don't understand `Isolate.run`, then I'm happy to change this example, but at the moment it feels like from a high-level, the example is "correct" (barring some implementation details that you pointed out).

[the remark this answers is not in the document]
ericwindmill: @lrhn I've updated the code to include the async factory, but I'm still wondering if other portions are incorrect

examples/concurrency/lib/isolate_spawn_syntax.dart
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+// This listener callback will be called each time a subsequent message
+// is sent from the main isolate to the worker isolate
+    receivePortInSpawnedIsolate.listen((dynamic message) async {
+      if (message is String) {
+        final dartObjects = jsonDecode(message);
+        sendPortToMainApp.send(dartObjects);
+      } else {
+        throw const SocketException(
```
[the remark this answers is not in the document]
ericwindmill: This makes sense to me and I agree. I am curious why I used Socket exception here... I wouldn't have come up with that on my own.

src/language/isolates.md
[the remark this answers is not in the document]
ericwindmill: This comment may address my questions/concerns from my previous reply. I wanted to say that incase you see these comments before my workday tomorrow when I have more time to go through this.  Thanks for being thorough!

src/language/isolates.md
```diff
@@ -0,0 +1,802 @@
...
+
+{{site.alert.note}}
+This diagram, and the following diagram, are high-level and intended to convey the concepts necessary to use isolates, but actual implementation requires a bit more code. A full code example is shown later on this page.  
+{{site.alert.end}}
+
+![A figure showing events being fed, one by one, into the event loop](/assets/img/language/concurrency/ports-setup.png)
```
MaryaBelanger, line 194: Hate to do this because the diagrams look great, but could you take the numbered list out of the diagram and just have that as text below? I think it's an accessibility thing (it's not screen readable) and also not searchable (this one and line 200, sorry!)

src/language/isolates.md
```diff
@@ -0,0 +1,802 @@
...
+
+[`send()` method]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}
+/dart-isolate/SendPort/send.html
+
+#### Setting up ports
+
+Setting up a port requires a few steps, which must be done in a specific order. The following figure shows the conceptual steps required to spawn a new isolate and establish 2-way communication between it and the main isolate.
+
+{{site.alert.note}}
+This diagram, and the following diagram, are high-level and intended to convey the concepts necessary to use isolates, but actual implementation requires a bit more code. A full code example is shown later on this page.
```
MaryaBelanger, line 191:
```suggestion
The diagrams in this section are high-level and intended to convey the concept of
using ports for isolates. Actual implementation requires a bit more code,
which you will find [later on this page](link-to-ports-example-section).  
```

MaryaBelanger (commented):
Another round, got to just before the ports example. Thanks for doing all this!

src/language/isolates.md
```diff
@@ -0,0 +1,802 @@
...
+Isolates are most commonly used in Flutter applications, when you 
+need to perform large computations that might otherwise cause the 
+UI to become unresponsive.
+They’re also useful in many server-side applications. In general, 
+you should use isolates whenever your application is handling computations that 
+are large enough to temporarily block other computations.
+
+While there aren't any rules about when you _must_ use isolates, here are
+common situations in which you can consider using isolates:
+
```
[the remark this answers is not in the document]
ericwindmill: Nah, makes sense to me.

src/language/concurrency.md
```diff
@@ -16,87 +19,108 @@ prevpage:
...
+[`Streams`]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-async/Stream-class.html
 
+## Asynchrony types and syntax
+
+If you're already familiar with `Future`, `Stream`, and async-await,
+then you can skip ahead to the [isolates section][].
+
+[isolates section]: #isolates
+
+### Future type syntax
```
parlough, line 102: I'm not sure I would call this section "syntax", as it's not about a syntax specific to futures, it's about what a `Future` is. Perhaps just "Future type" is fine? Or even "Futures"?

src/language/concurrency.md
```diff
@@ -144,80 +168,106 @@ void main() [!async!] {
...
+executes non-Dart code, in either the Dart virtual machine (VM) or the operating
+system (OS). Once `readAsString()` returns a value, Dart code execution resumes.
 
-![Flowchart-like figure showing app code executing from start to exit, waiting for native I/O in between](/assets/img/language/concurrency/basics-await.png)
+![Flowchart-like figure showing app code executing from start to exit, waiting 
+for native I/O in between](/assets/img/language/concurrency/basics-await.png)
 
-If you'd like to learn more about using `async`, `await`, and futures,
-visit the [asynchronous programming codelab][].
+### Stream syntax
```
parlough, line 188: Same as the futures section, this section is not about syntax, it's about what streams are.

src/language/concurrency.md
```diff
@@ -16,87 +19,108 @@ prevpage:
...
+## Asynchrony types and syntax
+
+If you're already familiar with `Future`, `Stream`, and async-await,
+then you can skip ahead to the [isolates section][].
+
+[isolates section]: #isolates
+
+### Future type syntax
+ 
+A promise to eventually provide a string value is typed as `Future<String>`.
```
parlough, line 104: As a reader, I was a bit surprised this jumped straight to an example. I think it would be helpful to offer a more generic introduction before jumping to a specific instance of it.

parlough, line 104: Also, it's not necessarily a promise to provide a string value because the future could complete with an error.

Technically it may never complete either, but that's probably not worth worrying about for the purposes of this doc.

I think the error case is definitely worth mentioning though. There's a good sentence from the [codelab](https://dart.dev/codelabs/async-await#what-is-a-future) that could be modified to mention errors.

```
A `Future` represents the result of an asynchronous operation that will eventually complete with a value or an error.
```

Could then go on with your specific example.

src/language/concurrency.md
```diff
@@ -144,80 +168,106 @@ void main() [!async!] {
...
+The `await` keyword works only in functions that have `async` before the 
+function body. 
 {{site.alert.end}}
 
-As the following figure shows,
-the Dart code pauses while `readAsString()` executes non-Dart code,
-in either the Dart virtual machine (VM) or the operating system (OS).
-Once `readAsString()` returns a value, Dart code execution resumes.
+As the following figure shows, the Dart code pauses while `readAsString()`
+executes non-Dart code, in either the Dart virtual machine (VM) or the operating
```
parlough, line 182: I know you didn't write this, but I think we should drop "the Dart virtual machine" mention, as Dart could be running on other platforms too. I think just "the Dart runtime" is sufficient.

```suggestion
executes non-Dart code, in either the Dart runtime or the operating
```

src/language/concurrency.md
```diff
@@ -144,80 +168,106 @@ void main() [!async!] {
...
-If you'd like to learn more about using `async`, `await`, and futures,
-visit the [asynchronous programming codelab][].
+### Stream syntax
 
-[asynchronous programming codelab]: /codelabs/async-await
+Dart also supports asynchronous code in the form of streams. Streams
+provide values in the future and repeatedly over time. A promise to provide a
+series of `int` values over time has the type `Stream<int>`.
+
+This example will return a new `int` value every second.
```
parlough, line 194: Not a big deal, but it's not the example that will return a new `int`, but rather the stream created in the example.

Also, it might be better to stick to the phrasing "emits" so the reader doesn't think `Stream.periodic` is being returned every second or something like that.

What do you think about an introduction like:

```suggestion
In the following example, the stream created with `Stream.periodic`
repeatedly emits a new `int` value every second.
```

src/language/concurrency.md
```diff
@@ -144,80 +168,106 @@ void main() [!async!] {
...
+```dart
+Stream<int> stream = Stream.periodic(const Duration(seconds: 1), (i) => i * i);
+```
+
+#### await-for and yield
 
+Await-for is a type of for loop that executes each subsequent iteration of the
+loop as new values are provided. In other words, it’s used to “loop over”
+streams. In this example, a new value will be returned from the function
+`sumStream` as new values are emitted from the stream that’s provided as an
```
parlough, line 206: Same idea here with the "return" vs "emit" terminology, since `sumStream` doesn't keeping returning a new value, but instead the stream it returns keeps emitting new values.

src/language/concurrency.md
```diff
@@ -144,80 +168,106 @@ void main() [!async!] {
...
+## Isolates
+
+Dart also supports concurrency via isolates. Most modern devices have multi-core
+CPUs. To take advantage of multiple cores, developers sometimes use
+shared-memory threads running concurrently. However, shared-state concurrency is
+[error prone](https://en.wikipedia.org/wiki/Race_condition#In_software) and can
+lead to complicated code.
+
+Instead of threads, all Dart code runs inside of isolates. Each isolate has its
+own memory heap, ensuring that none of the state in an isolate is accessible
```
parlough, line 234: I know this is pre-existing, but since we're here, let's avoid the term "heap" since they can share the same heap. It sounds like "has its own global fields" is the most accurate, but "has its own memory" is probably fine too.

src/language/concurrency.md
```diff
@@ -144,80 +168,106 @@ void main() [!async!] {
...
-
+won't occur in Dart. That said, isolates don't prevent race conditions all
+together. For more information on this concurrency model, read about
+the [Actor model](https://en.wikipedia.org/wiki/Actor_model).
 
 Using isolates, your Dart code can perform multiple independent tasks at once,
-using additional processor cores if they're available.
-Isolates are like threads or processes,
-but each isolate has its own memory and a single thread running an event loop.
+using additional processor cores if they’re available. Isolates are like threads
```
parlough, line 244:
```suggestion
using additional processor cores if they're available. Isolates are like threads
```

src/language/concurrency.md
```diff
@@ -144,80 +168,106 @@ void main() [!async!] {
...
+or processes, but each isolate has its own memory and a single thread running an
+event loop.
```
parlough, line 246:
```suggestion
event loop.
```

src/language/concurrency.md
```diff
@@ -237,303 +287,272 @@ After handling the events, the isolate exits.
...
-{{site.alert.end}}
-
-[native and non-native platforms]: /overview#platform
-[Flutter's `compute` function]: {{site.flutter-api}}/flutter/foundation/compute.html
+(reading and writing files, for example), set timers, and more. It has its own
+memory and doesn't share any state with the main isolate. The worker isolate can
+block without affecting other isolates.
 
-#### Running an existing method in a new isolate
+### Isolate syntax
```
parlough, line 333:
```suggestion
### Using isolates
```

src/language/concurrency.md
```diff
@@ -237,303 +287,272 @@ After handling the events, the isolate exits.
...
-  // Read some data.
-  final jsonData = await Isolate.run(_readAndParseJson);
+In most cases, `Isolate.run` is the recommended 
+API to run processes in the background.
 
-  // Use that data.
-  print('Number of JSON keys: ${jsonData.length}');
-}
-```
+### `Isolate.run()`
```
parlough, line 344: Is this intended to be a subheading of the previous section similar to the `Isolate.spawn` section below?

```suggestion
#### `Isolate.run()`
```

src/language/concurrency.md
```diff
@@ -237,303 +287,272 @@ After handling the events, the isolate exits.
...
-   The worker isolate performs a verification pass to ensure
-   the objects are allowed to be transferred.
-
-`_readAndParseJson()` is an existing,
-asynchronous function that could just as easily
-run directly in the main isolate.
-Using `Isolate.run()` to run it instead enables concurrency.
-The worker isolate completely abstracts the computations
-of `_readAndParseJson()`. It can complete without blocking the main isolate.
+#### Isolate.spawn
```
parlough, line 360: Do you want this in code syntax to be consistent with the `Isolate.run` section?

```suggestion
#### `Isolate.spawn`
```

src/language/concurrency.md
```diff
@@ -235,305 +268,268 @@ After handling the events, the isolate exits.
...
+communication between the main isolate and the worker isolate. This code 
+snippet shows the bare-minimum needed to create a long-lived isolate. You 
+can see a more detailed code example and explanation on the [Isolates][] page.
 
-You can also create a simple worker isolate with `run()` using a
-function literal, or closure, directly in the main isolate.
+[Isolates]: /language/isolates
 
-<?code-excerpt "lib/simple_isolate_closure.dart (main)"?>
+<?code-excerpt "lib/isolate_spawn_syntax.dart"?>
```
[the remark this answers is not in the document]
parlough: I do think it's likely better to just reference the other page rather than reproduce a pared down version here.

src/language/concurrency.md
```diff
@@ -237,303 +287,272 @@ After handling the events, the isolate exits.
...
 
-{% assign samples = "https://github.com/dart-lang/samples/tree/main/isolates" %}
+#### Isolates aren't threads
 
-[isolate samples]: {{ samples }}
-[send_and_receive.dart]: {{ samples }}/bin/send_and_receive.dart
-[long_running_isolate.dart]: {{ samples }}/bin/long_running_isolate.dart
+If you’re coming to Dart from a language with multithreading, it’d be reasonable
+to expect isolates to behave like threads, but that isn’t the case. Each isolate
+has its own memory heap, ensuring that none of the state in an isolate is
```
parlough, line 469:
```suggestion
has its own memory, ensuring that none of the state in an isolate is
```

src/language/concurrency.md
```diff
@@ -237,303 +287,272 @@ After handling the events, the isolate exits.
...
+- [`ReceivePort`][]
+- [`DynamicLibrary`][]
+- [`Finalizable`][]
+- [`Finalizer`][]
+- [`NativeFinalizer`][]
+- [`Pointer`][]
+- [`UserTag`][]
+- Instances of classes that either themselves are marked with @pragma
+  ('vm:isolate-unsendable'), extend or implement such classes cannot be sent
+  through the ports.
```
parlough, line 495: The pragma annotation should be in code syntax.

Also, it seems like this sentence is from the API docs, but it is hard to read. I'm struggling with an improvement though. @MaryaBelanger any ideas?

src/language/concurrency.md
```diff
@@ -237,303 +287,272 @@ After handling the events, the isolate exits.
...
 
-[`Isolate.spawnUri()`]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-isolate/Isolate/spawnUri.html
+Apart from those exceptions, any object can be sent. 
+Check out the [`SendPort`][] documentation for more information.
```
parlough, line 498: Consider linking directly to `SendPort.send` docs here since almost all of the details are there.

src/language/concurrency.md
```diff
@@ -237,303 +287,272 @@ After handling the events, the isolate exits.
...
+#### Message types
 
-When an isolate calls [`Isolate.spawn()`][],
-the two isolates have the same executable code
-and are in the same _isolate group_.
-Isolate groups enable performance optimizations such as sharing code;
-a new isolate immediately runs the code owned by the isolate group.
-Also, `Isolate.exit()` works only when the isolates
-are in the same isolate group.
+Messages sent via [`SendPort`][]
```
parlough, line 482: Do consider mentioning `Isolate.run` and `Isolate.exit` here as it may not be clear to readers that they are subject to the same limitations.

src/language/concurrency.md
```diff
@@ -237,303 +287,272 @@ After handling the events, the isolate exits.
...
+as the spawning isolate. Web workers don't have an equivalent API.
 
 [Dart web platform]: /overview#platform
 [web workers]: https://developer.mozilla.org/docs/Web/API/Web_Workers_API/Using_web_workers
+
+
+## Additional resources
+
+- If you’re using many isolates, consider
+  the [IsolateNameServer][]
```
parlough, line 539:
```suggestion
  the [`IsolateNameServer`][]
```

(and update the reference)

src/language/concurrency.md
```diff
@@ -237,303 +287,272 @@ After handling the events, the isolate exits.
...
+  in Flutter, or
+  the [pub package][] that clones
```
parlough, line 541: Consider naming the package here:

```
[`package:isolate_name_server`][] that provides
similar functionality for non-Flutter Dart applications.
```

src/language/concurrency.md
```diff
@@ -237,303 +287,272 @@ After handling the events, the isolate exits.
...
+  the functionality for Dart applications not using Flutter.
+- Dart’s isolates are an implementation of
```
parlough, line 543: This reads less like an "Additional resource" and more like an instructional sentence. Try to keep the entries in a list parallel in their construction.

src/language/isolates.md
```diff
@@ -0,0 +1,847 @@
+---
+title: Isolates
+description: Information on writing isolates in Dart.
+short-title: Isolates
+prevpage:
+  url: /language/async
+  title: Asynchronous support
+nextpage:
+  url: /null-safety
+  title: Sound Null Safety
```
parlough, line 10:
```suggestion
  title: Sound null safety
```

parlough (commented):
Thanks for working on these articles @ericwindmill! I'm super excited for these to get out to Dart developers who I know are anxious to better understand Dart's event loop and isolates.

I left some comments on the concurrency article, but am going to wait on the isolates article. Do consider breaking the two page updates in to two new separate PRs if that's possible. I think it'll be easier for you to manage and easier for others to review. Plus it's nice if some of the work can land earlier.

Also it seems like some sections got reformatted (automatically?) to break at 80 chars rather than their original _(more semantic)_ breaks, which made it hard to identify what changed. Consider disabling that at least for the Dart and Flutter sites.

Let me know if any of my comments weren't clear, and please feel free to push back :)

src/language/concurrency.md
```diff
@@ -16,87 +19,108 @@ prevpage:
...
-non-blocking I/O, performing an HTTP request, or communicating with a browser. 
-Other scenarios include waiting for computations
-performed in another Dart isolate as described below, 
-or maybe just waiting for a timer to trigger. 
-All of these processes either run in a different thread, 
-or are handled by the operating system or the Dart runtime, 
-which allows Dart code to run concurrently with the computation.
+[`Streams`]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-async/Stream-class.html
 
+## Asynchrony types and syntax
```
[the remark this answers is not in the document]
ericwindmill: This is fine with me. I'm pretty sure all the titles in this sections are relics of the original docs, and I just didn't question them. The fact that you have any opinion at all is evidence that you've thought about it more than I have.

src/language/concurrency.md
ericwindmill: It is copied from the API docs. I think the better thing to say is just "classes that are marked @pragma...", because the API docs are then immediately referenced and linked, so folks can easily see the explanation. Also, if someone understands and is working with these niche, advanced objects, I think we can assume they understand that the same rules would apply whether the class is annotated itself or it's extending an annotated class

ericwindmill: That's what I did in the separate PR that I put in for just the concurrency page. Feel free to disagree there, but I'm resolving here so we can keep the conversation sane.

src/language/concurrency.md
MaryaBelanger: I was referring to my questions here, which you addressed. So I'll resolve this

src/language/isolates.md
```diff
@@ -0,0 +1,1102 @@
...
+A `ReceivePort` is an object that handles messages that are sent from other
+isolates. Those messages are sent via a `SendPort`.
+
+{{site.alert.note}}
+`SendPort` objects are associated with exactly one `ReceivePort`,
+but a single `ReceivePort` can have many `SendPorts`.
+The `ReceivePort` creates a `SendPort` when it is created itself.
+You can create new `SendPorts` that
+can send messages to an existing `ReceivePort`.
+{{site.alert.note}}
```
MaryaBelanger, line 205:
```suggestion
{{site.alert.end}}
```

This will fix the nested notes, I think

MaryaBelanger:
Not finished reviewing yet, but as part of this PR you should link _to_ this new isolates page somewhere on the Concurrency page. I suggest [the end of the first paragraph](https://dart-dev--pr5332-isolate-documentatio-wj7iadc7.web.app/language/concurrency#concurrency-on-the-web:~:text=For%20more%20practical%20code%20examples%20of%20using%20async%20features%2C%20read%20the%20Asynchrony%20support.) (and you can fix the typo there too):

> For more practical code examples of using ~async features~ concurrency in Dart, 
> read the [Asynchrony support](https://dart-dev--pr5332-isolate-documentatio-wj7iadc7.web.app/language/async) **page and the [Isolates](https://dart-dev--pr5332-isolate-documentatio-wj7iadc7.web.app/language/isolates) page.**

src/language/isolates.md
```diff
@@ -0,0 +1,1106 @@
...
+  the `startupPort`.
+
+<?code-excerpt "lib/robust_ports_example/spawn_1.dart"?>
+```dart
+class Worker {
+  final SendPort _commands;
+  final ReceivePort _responses;
+
+  static Future<Worker> spawn() async {
+    // Create a receive port and add it's initial message handler
```
MaryaBelanger, line 621:
```suggestion
    // Create a receive port and add it's initial message handler.
```
Fix in code file too

src/language/isolates.md
```diff
@@ -0,0 +1,1106 @@
...
+to create a `ReceivePort` straight away, you’d only be able to add
+one `listener`, because `ReceivePort` implements [`Stream`][], rather
+than [`BroadcastStream`][].
+
+Effectively, this allows you to separate your isolate start-up logic from the
+logic that handles receiving messages after setting up communication is
+complete. This benefit will become more obvious as the logic in the other
+methods grows.
+
+#### Step 3: Spawn a worker isolate with `Isolate.spawn`.
```
MaryaBelanger, line 646:
```suggestion
#### Step 3: Spawn a worker isolate with `Isolate.spawn`
```

src/language/isolates.md
```diff
@@ -0,0 +1,1106 @@
...
+  and passing in the ports from that completer.
+
+<?code-excerpt "lib/robust_ports_example/spawn_2.dart"?>
+```dart
+class Worker {
+  final SendPort _commands;
+  final ReceivePort _responses;
+
+  static Future<Worker> spawn() async {
+    // Create a receive port and add it's initial message handler
```
MaryaBelanger, line 670:
```suggestion
    // Create a receive port and add its initial message handler.
```
Fix in code file too

src/language/isolates.md
```diff
@@ -0,0 +1,1106 @@
...
+    final initPort = RawReceivePort();
+    final connection = Completer<(ReceivePort, SendPort)>.sync();
+    initPort.handler = (initialMessage) {
+      final commandPort = initialMessage as SendPort;
+      connection.complete((
+        ReceivePort.fromRawReceivePort(initPort),
+        commandPort,
+      ));
+    };
+    // Spawn the isolate
```
MaryaBelanger, line 680:
```suggestion
    // Spawn the isolate.
```
Fix in code file too

src/language/isolates.md
```diff
@@ -0,0 +1,1106 @@
...
+        await connection.future;
+
+    return Worker._(sendPort, receivePort);
+  }
+```
+
+Note that in this example (compared to the [previous example][]), `Worker.spawn`
+acts as an asynchronous static constructor for this class and is the only way to
+create an instance of `Worker`. This simplifies the API, making the code that
+will create an instance of Worker cleaner.
```
MaryaBelanger, line 698:
```suggestion
creates an instance of `Worker` cleaner.
```

src/language/isolates.md
```diff
@@ -0,0 +1,1106 @@
...
+
+#### Step 4: Complete the isolate setup process
+
+In this step, you will complete the basic isolate setup process. This correlates
+almost entirely to the [previous example][], and there are no new concepts.
+There is a slight change in that the code is broken into more methods, which
```
MaryaBelanger, line 704:
```suggestion
There is a slight change in that the code is broken into more methods, which
is a design practice that
```

src/language/isolates.md
```diff
@@ -0,0 +1,1106 @@
...
+outside code to send JSON to the worker isolate to be decoded.
+
+<?code-excerpt "lib/robust_ports_example/step_4.dart (parse-json)"?>
+```dart
+Future<Object?> parseJson(String message) async {
+  _commands.send(message);
+}
+```
+
+This method will be updated in the next step.
```
MaryaBelanger, line 798:
```suggestion
You will update this method in the next step.
```

src/language/isolates.md
```diff
@@ -0,0 +1,1106 @@
...
+<?code-excerpt "lib/robust_ports_example/step_5_add_completers.dart (vars)"?>
+```dart
+class Worker {
+  final SendPort _commands;
+  final ReceivePort _responses;
+  final Map<int, Completer<Object?>> _activeRequests = {};
+  int _idCounter = 0;
+```
+
+The `_activeRequests` map will associate a message sent to the worker isolate
```
MaryaBelanger, line 825:
```suggestion
The `_activeRequests` map associates a message sent to the worker isolate
```

src/language/isolates.md
```diff
@@ -0,0 +1,1106 @@
...
+with a `Completer`. The keys used in `_activeRequests` are taken
+from `_idCounter`, which will be increased as more messages are sent.
```
MaryaBelanger, line 827:
```suggestion
from `_idCounter`, which increases as more messages are sent.
```

src/language/isolates.md
```diff
@@ -0,0 +1,1106 @@
...
+messages to the worker isolate.
+
+- First create a `Completer`.
+- Next, increment `_idCounter`, so that each `Completer` is associated with a
+  unique number.
+- Add an entry to the `_activeRequests` map in which the key is the current
+  number of `_idCounter`, and the completer is the value.
+- Send the message to the worker isolate, along with the id. Because you can
+  only send one value through the `SendPort`, wrap the id and message in a
+  record.
```
MaryaBelanger, line 839:
```suggestion
  [record][].
```

src/language/isolates.md
```diff
@@ -0,0 +1,1106 @@
...
+ports on the main isolate and the worker isolate.
+
+First, add a class-level boolean that tracks if the ports are closed.
+
+Then, add the `Worker.close` method. Within this method:
+Update `_closed` to be true. Send a final message to the worker isolate. This
+message is a `String` that reads “shutdown”. This isn’t a specific string, and
+it could be any object you’d like. It will be used in the next code snippet.
+Finally, check if `_activeRequests` is empty. If it is, close down the main
+isolate’s `ReceivePort` called `_responses`
```
MaryaBelanger, line 915:
```suggestion
- First, add a class-level boolean that tracks if the ports are closed.
- Then, add the `Worker.close` method. Within this method:
  - Update `_closed` to be true.
  - Send a final message to the worker isolate.
    This message is a `String` that reads “shutdown”,
    but it could be any object you’d like.
    You will use it in the next code snippet.
- Finally, check if `_activeRequests` is empty. If it is, close down the main
  isolate’s `ReceivePort` named `_responses`.
```
The long paragraph is formatted kind of weird, so I made the whole section match the rest of the sections (double check that the sub-bullet layers render properly)

src/language/isolates.md
```diff
@@ -0,0 +1,1106 @@
...
+      if (_activeRequests.isEmpty) _responses.close();
+      print('--- port closed --- ');
+    }
+  }
+```
+
+Next, you need to handle the “shutdown” message in the worker isolate. Add the
+following code to the `_handleCommandsToIsolate` method. This code will check if
+the message is a `String` that reads “shutdown”. If it is, it will close the
+worker isolate’s `ReceivePort`, and return.
```
MaryaBelanger, line 935:
```suggestion
- Next, you need to handle the “shutdown” message in the worker isolate. Add the
  following code to the `_handleCommandsToIsolate` method. This code will check if
  the message is a `String` that reads “shutdown”. If it is, it will close the
  worker isolate’s `ReceivePort`, and return.
```

src/language/isolates.md
```diff
@@ -0,0 +1,1106 @@
...
+
+<?code-excerpt "lib/robust_ports_example/step_6_close_ports.dart (handle-commands)"?>
+```dart
+static void _handleCommandsToIsolate(
+  ReceivePort receivePort,
+  SendPort sendPort,
+) {
+  receivePort.listen((message) {
+    // New if-block
```
MaryaBelanger, line 944:
```suggestion
    // New if-block.
```
fix in code file too

src/language/isolates.md
```diff
@@ -0,0 +1,1106 @@
...
+  _activeRequests[id] = completer;
+  _commands.send((id, message));
+  return await completer.future;
+}
+```
+
+#### Complete example
+
+<details>
+  <summary>Expand here to see full example</summary>
```
MaryaBelanger, line 978:
```suggestion
  <summary>Expand here to see the full example</summary>
```

src/language/isolates.md
```diff
@@ -0,0 +1,1106 @@
...
+      sendPort.send((id, jsonData));
+    } catch (e) {
+      sendPort.send((id, RemoteError(e.toString(), '')));
+    }
+  });
+}
+```
+
+Finally, you should add code to check if the ports are closed before trying to
+send messages. Add one line in the `Worker.parseJson` method.
```
MaryaBelanger, line 961:
```suggestion
- Finally, you should add code to check if the ports are closed before trying to
  send messages. Add one line in the `Worker.parseJson` method.
```

MaryaBelanger (approved):
PHEW! What a mammoth piece of work, and it turned out so great! Thank you!

Don't forget to get another technical review for your new examples (if you haven't already done so offline :)).

ericwindmill:
@johnpryan, @lrhn, @domesticmouse, can one of you please review the ports code examples? Most importantly examples/concurrency/lib/robust_ports_example/complete.dart

examples/concurrency/lib/robust_ports_example/step_4.dart
```diff
@@ -0,0 +1,80 @@
...
+import 'dart:convert';
+import 'dart:isolate';
+
+// #docregion constructor
+class Worker {
+  final SendPort _commands;
+  final ReceivePort _responses;
+// #enddocregion constructor
+  static Future<Worker> spawn() async {
+    // Create a receive port and add it's initial message handler
```
lrhn, line 13: it's -> its

lrhn (approved):
The complete example, and the steps towards it, LGTM!

PR #5346, opened 2023-11-14
Use "enable" rather than "allow" in objective-c-interop.md

dcharkes:
> The usual distinction between allow and enable is that allow means "to not prohibit something, to let it happen, to remove any constraint that would prevent something from happening" and enable means "to make something possible"

https://ell.stackexchange.com/questions/182862/the-difference-between-allow-and-enable

---

### DIFF
--- a/src/guides/libraries/objective-c-interop.md
+++ b/src/guides/libraries/objective-c-interop.md
@@ -17,7 +17,7 @@ to call Objective-C and Swift APIs.
   and [in active development](https://github.com/dart-lang/sdk/issues/49673).
 {{site.alert.end}}
 
-`dart:ffi` allows Dart code to interact with native C APIs.
+`dart:ffi` enables Dart code to interact with native C APIs.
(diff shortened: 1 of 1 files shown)


src/guides/libraries/objective-c-interop.md
```diff
@@ -94,7 +94,7 @@ which are very large.
 If bindings are generated without any filters,
 the resulting file can be millions of lines long.
 To solve this problem,
-the ffigen config has fields that allow you to filter out
+the ffigen config has fields that enables you to filter out
```
MaryaBelanger, line 97:
```suggestion
the ffigen config has fields that enable you to filter out
```

MaryaBelanger (approved):
Just one typo

MaryaBelanger:
I like this distinction. Not exactly "language feature" terminology, but adjacent. So maybe a candidate for #5247 (enable vs. allow)

PR #5366, opened 2023-11-21
Move interop section out from development section

### LINKED ISSUE #5364: Take "Interop" contents out from under "Development"
Right now "Interoperability" is nested under "Development" in the side nav. 
image

I know interop _is_ inherently about development, and it makes a lot of sense for it to be nested there. Moving it out to it's own section would objectively be _less_ organized than it's current position. 

But, with a lot of work to [JS interop](#4578) upcoming, and [C interop](#5269) always expanding, it'd be really nice to have a sidenav section where you can _see_ the expanded contents under each interop type. But the side nav only expands to the 3rd level. That means it won't be obvious that JS interop has a bunch of tutorials and reference pages underneath it, it's just going to look like there's nothing new there, and that's pretty underwhelming for such a big feature release.

So, like the language tour expansion to make way for all the new Dart 3 features we did earlier this year, can we move "Interoperability" to it's own top level section to be able to highlight new interop docs?

Any opinions @mit-mit @parlough @atsansone @sfshaza2, or anyone?

MaryaBelanger:
Fixes #5364


src/_data/side-nav.yml
```diff
@@ -242,6 +232,17 @@
     - title: Environment declarations
       permalink: /guides/environment-declarations
 
+- title: Interoperability
```
MaryaBelanger, line 235: "Interoperability" or just "Interop"?

parlough: To be honest, I like "Interop" better, but "Interoperability" looks nicer in the sidenav it feels, so I actually think that's the better choice for now. Up to you though!

parlough (approved):
Thanks! This will provide you and everyone else a much cleaner slate to work with.

I'm assuming `/interop/js-interop` will become a new subcategory with the new work?

src/_data/side-nav.yml
```diff
@@ -242,6 +232,17 @@
...
+  expanded: false
+  children:
+  - title: JavaScript interop
+    permalink: /interop/js-interop
```
parlough, line 239: Actually, consider moving this to the bottom of the list for now, especially while it's in limbo.

Issue #5382, opened 2023-11-30
☂️ Migrate embedded DartPads away from multifile


parlough:
To prepare for DartPad no longer supporting multi-file and related features (tests, solutions, hints), we need to update, migrate, or remove the following content.

#### Codelabs

The should either be updated to be useful without the testing capability, consolidated with other material elsewhere, or just removed for now.

- [x] [Asynchronous programming: futures, async, await](https://dart.dev/codelabs/async-await) codelab
  - Consider moving valuable content to [Asynchrony support](https://dart.dev/language/async) and [`dart:async` library tour](https://dart.dev/guides/libraries/library-tour#dartasync---asynchronous-programming), then deleting.
  - Another option is to include the solution as a separate details drop-down as an intermediate step
  - https://github.com/dart-lang/site-www/pull/5544
- [x] [Iterable collections](https://dart.dev/codelabs/iterables) codelab
  - I think this one can live on without the multi-file functionality in the short term with the solutions of the exercises as a details dropdown.
  - https://github.com/dart-lang/site-www/pull/5542
- [x] [Null safety](https://dart.dev/codelabs/null-safety) codelab
  - Never was completed, and less useful standalone now that null safety is just Dart.
  - Can move useful portions into the cheat sheet for now.
  - Can revisit more when we rewrite our type system docs (https://github.com/dart-lang/site-www/issues/4745).
  - https://github.com/dart-lang/site-www/pull/5539
- [x] [Dart cheatsheet](https://dart.dev/codelabs/dart-cheatsheet) codelab
  - Popular way to learn about and try out some of Dart's features.
  - Not super well maintained, but can likely live on for now with tests and solutions removed from embeds.
  - Should be replaced by language tour / interactive step-by-step tutorial eventually.
  - https://github.com/dart-lang/site-www/pull/5392

#### Low-level HTML tutorials

These will be replaced by new `package:web` and static interop docs, so we likely don't need to worry about them too much. If we move to the new DartPad before they are replaced, dropping the embedded DartPads from the page is fine.

- [x] [Get started: Web apps](https://dart.dev/tutorials/web/get-started)
  - https://github.com/dart-lang/site-www/pull/5415
- [x] [Add elements to the DOM](https://dart.dev/tutorials/web/low-level-html/add-elements)
  - https://github.com/dart-lang/site-www/pull/5415
- [x] [Remove DOM elements](https://dart.dev/tutorials/web/low-level-html/remove-elements)
[...]

PR #5383, opened 2023-11-30
Remove DartPad tutorial best practices

parlough:
Support for tests/solutions/hints/etc (at least in their current form) is going away soon in embedded DartPads, so I don't think we should be surfacing guidelines on how to best use them.

This PR removes that page and redirects to an underlying paper which publishes some of the same findings.


atsansone (approved):
LGTM.

PR #5386, opened 2023-12-01
Fix accidental replacement on type system page

parlough:
In https://github.com/dart-lang/site-www/pull/5217, when adding code excerpts to these samples, I accidentally didn't make the replace specific enough and replaced too much.


wngtk:
Why use Root as a return type? Where is the Root?

Issue #4787, opened 2023-04-25
[Future] Remove "DO use = to separate a named parameter from its default value" guideline


parlough:
### Page URL

https://dart.dev/guides/language/effective-dart/usage

### Page source

https://github.com/dart-lang/site-www/tree/main/src/_guides/language/effective-dart/usage.md

### Describe the problem

The guideline [DO use = to separate a named parameter from its default value](https://dart.dev/guides/language/effective-dart/usage#do-use--to-separate-a-named-parameter-from-its-default-value) is no longer relevant in a Dart 3 world.

### Expected fix

The guideline should be removed in the future, perhaps alongside Dart 3.1 or 3.2.

PR #5389, opened 2023-12-01
Remove guideline on named parameter default value syntax

parlough:
It's time. The old syntax was deprecated and linted for many years, then had a warning in the analyzer for a while, and now has been removed from the language for a few releases. 

🫡

Closes https://github.com/dart-lang/site-www/issues/4787


MaryaBelanger (approved):
Thanks for catching those other unused headings in the toc!

PR #5391, opened 2023-12-05
Make the recommendation to not use `dart:io`/`dart:html` for HTTP requests stronger

brianquinlan:
Part of https://github.com/dart-lang/sdk/issues/52023

---

- [X] I’ve reviewed the contributor guide and applied the relevant portions to this PR.
- [X] This PR doesn’t contain automatically generated corrections or text (Grammarly, LLMs, and similar).
- [X] This PR follows the [Google Developer Documentation Style Guidelines](https://developers.google.com/style) — for example, it doesn’t use _i.e._ or _e.g._, and it avoids _I_ and _we_ (first person).
- [X] This PR uses [semantic line breaks](https://github.com/dart-lang/site-shared/blob/main/doc/writing-for-dart-and-flutter-websites.md#semantic-line-breaks) of 80 characters or fewer.

<details>

</details>


src/guides/libraries/_dart-html-tour.md
```diff
@@ -322,119 +322,13 @@ subclasses. Some common events include:
...
-
-  await request.onLoadEnd.first;
-
-  if (request.status == 200) {
-    // Successful URL access...
-  }
-  // ···
-}
-```
+You should avoid directly using `dart:html` to make HTTP requests.
```
brianquinlan, line 325: Let me know if you think that I should just remove this section completely.

src/guides/libraries/_dart-io-tour.md
```diff
@@ -234,27 +234,13 @@ void processRequest(HttpRequest request) {
...
-  var url = Uri.parse('http://localhost:8888/dart');
-  var httpClient = HttpClient();
-  var request = await httpClient.getUrl(url);
-  var response = await request.close();
-  var data = await utf8.decoder.bind(response).toList();
-  print('Response ${response.statusCode}: $data');
-  httpClient.close();
-}
-```
+You should avoid directly using `dart:io` to make HTTP requests.
```
brianquinlan, line 237: Let me know if you think that I should just remove this section completely.

parlough: I think it's good to keep these sections for two releases or so to point users in the right direction. Curious what Marya thinks though.

In the sections you remove, do consider linking to the `/tutorials/server/fetch-data` page as an alternative to learn more though since these changes remove learning material.

I do think making the corresponding note/warning on the API's dartdocs more prominent will have a larger effect than any changes here.

brianquinlan: Great suggestion! I've added a link to the tutorial.

There is a code review out for making the corresponding change to the API docs:
https://dart-review.googlesource.com/c/sdk/+/340283

src/tutorials/server/fetch-data.md
```diff
@@ -129,12 +129,17 @@ see [What is a URL?][] on the mdn web docs.
 
 ## Retrieve the necessary dependencies
 
-You can directly use `dart:io` or `dart:html` to make HTTP requests,
-however those libraries are platform dependent.
 `package:http` provides a cross-platform library
 for making composable HTTP requests,
 with optional fine-grained control.
 
+{{site.alert.note}}
```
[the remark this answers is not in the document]
brianquinlan: Thanks!

src/guides/libraries/_dart-io-tour.md
brianquinlan: We want to be pretty forceful about our recommendation - we are using the same initial wording in the SDK documentation. **But** we are preserving the example so users can find it there.

Wouldn't it be better to avoid having examples in our library tour that should not be used except in rare circumstances?

brianquinlan:
> I left some initial comments, but have some questions too. There are a few other mentions of using HttpRequest elsewhere on the library tour page (like in code blocks in the [Future](https://dart.dev/guides/libraries/library-tour#basic-usage) section, and the [HTTP server](https://dart.dev/guides/libraries/library-tour#http-server) section under dart:io). Do those all need to be changed to use `package:http`, or is it not interchangeable like that?

I updated the `Future` example. The `HTTPServer` example is fine.

src/guides/libraries/_dart-io-tour.md
MaryaBelanger: Thanks for explaining, definitely trust your judgement here! 

You can make the final call here, but just to be extra explicit for readers, you could explicitly state that there's still an example in the sdk docs (that's the API docs right?) _if you really need it_ (like Parker said, having an example in one place then making it disappear _too well_ all of a sudden isn't ideal). So, after lines 245 in this file, and 333 at the end of the html-tour file, something like:

```
Though no longer recommended / not recommended for the majority of use cases, 
you can find more information on / an example of `HttpClient`/`HttpRequest` 
in the [API documentation](link-to-api-docs).
```

Approving in the mean time so you can make the final call, but I'll come back and check again if you do change something else and want another look. Thanks for doing this!

MaryaBelanger (approved):

src/guides/libraries/_dart-io-tour.md
brianquinlan: Isn't the purpose the library tour to indoctrinate developers new to Dart? Those developers should start with `package:http` (or some equivalent package).

I preserved the link to the class and it still has an example there.

brianquinlan: Oh, and don't trust my judgement; if often sucks ;-)

brianquinlan:
1. Are the check failures related to my code? I don't know how to interpret the results.
2. I don't have permission to merge this.

parlough:
> Are the check failures related to my code? I don't know how to interpret the results.

We pull code snippets from source files so they can be analyzed and tested. So the two you edited needed to be updated at the source. I did this for you in https://github.com/dart-lang/site-www/pull/5391/commits/22437f25e212a18d47070f783450af306b7e67f2 :)

Thanks for working on this :D

brianquinlan:
> > Are the check failures related to my code? I don't know how to interpret the results.
> 
> We pull code snippets from source files so they can be analyzed and tested. So the two you edited needed to be updated at the source. I did this for you in [22437f2](https://github.com/dart-lang/site-www/commit/22437f25e212a18d47070f783450af306b7e67f2) :)
> 
> Thanks for working on this :D

Thanks for fixing it for me!

src/guides/libraries/_dart-io-tour.md
MaryaBelanger: > Isn't the purpose the library tour to indoctrinate developers new to Dart? 

Definitely so. What Parker and I are referring to is the kind of nebulous, hard to measure, off-chance that "what if someone, somewhere has a link to this specific section that they hold onto for reference that says "example here for whatever", and the off-chance that they'll be greatly perturbed if all of a sudden that "whatever"s not there any more and it's not immediately obvious where to find it. 

This definitely isn't a big deal so again, your call :) And your judgement about how important it is that no one remotely even goes poking around the old stuff unless they're absolutely digging for it is based on a lot more context than I have, I'm just citing a general best practice for tech writing. And yes, you do have the link on the API names which I'm ok with!

brianquinlan: Thanks for the detailed answer. I think that I'm OK with it as-is.

brianquinlan:
Could someone press the merge button for me?

parlough (approved):
Thank you!!

PR #5392, opened 2023-12-05
Remove multi-file DartPad usages from cheatsheet

MaryaBelanger:
Part of #5382 

- The tests feature contained `main` for each example, but that features being removed. The examples need `main` to run, and also are not helpful without the feedback that the tests provided, so we're adding the tests to code visible in each dartpad.
  - Comment at the top of each that this part is tests, don't edit
  - Will have to adjust the tests for more useful messages to users. 
  - Since `main` is a big focus now, add a new section on the `main` function at the top of the cheatsheet.

- Also remove hint and solution feature and place in expandable "solution" section under each example.

- Rewrite the intro section to explain the new way the page works.


parlough:
Thanks for raising your concerns. I think we can still figure out something beneficial here.

Do you have time to meet for 30 minutes today (12/5)? Might be easier to discuss our options synchronously.

MaryaBelanger:
@parlough I was already planning to leave the new section on `main` for another PR, just to get this one finished. But then I saw in the intro that it says:

> The Dart language is designed to be easy to learn for coders coming from other languages, **but it has a few unique features. This codelab walks you through the most important of these language features.**

In that light, going over `main` doesn't really fit to the theme of the page. What do you think?

parlough:
I think it's fine to move to a follow-up CL, but I'd say it does fit the theme of being a unique feature to Dart. Python and JS don't have a `main` method, Java (until a few weeks ago) required a `static public void main(String[] args)`, and C# requires something similar.

MaryaBelanger:
Oh perfect, didn't know that and that info will help me write the section too. Created #5414 , thanks!

src/codelabs/dart-cheatsheet.md
```diff
@@ -473,19 +458,45 @@ void main() {
...
 ```
 
+<details markdown="1">
+ <summary>
+   <b>Solution for collection literals example</b>
+  </summary>
+
+  This exercise is fairly straightforward.
+  Just add a list, set, or map literal after each equals sign.
+  See the codelab text for the correct syntax to use.
```
parlough, line 478: I know you didn't write this, but while we're here, can we remove this text that makes it sound easy? It might not be easy for someone.

Also the "see the codelab text..." portion doesn't make sense anymore with the answer right below.

I can't really think of any super valuable hint here, if you can't either, I'd be fine just dropping it. Perhaps we could mention something like "Just remember to specify the types when they can't be inferred." or something like that for the empty ones?

I'll come back to this comment if I think of anything else. Thanks!!

src/codelabs/dart-cheatsheet.md
```diff
@@ -53,41 +57,36 @@ For example, `stringify(2, 3)` should return `'2 3'`.
...
 }
-{$ end test.dart $}
-{$ begin hint.txt $}
-Both x and y are simple values,
-and Dart's string interpolation will handle
-converting them to string representations.
-All you need to do is use the $ operator to
-reference them inside single quotes, with a space in between.
-{$ end hint.txt $}
+{$ end main.dart $}
```
parlough, line 68: Can you get rid of these `begin` and `end main.dart` tags across the file? Support for them will also be going away and they're not necessary with just one Dart file anyway.

src/codelabs/dart-cheatsheet.md
```diff
@@ -559,58 +554,72 @@ void main() {
...
 ```
 
+<details markdown="1">
+ <summary>
+   <b>Solution for arrow syntax example</b>
+  </summary>
+
+  For the product, you can just multiply the three values together.
```
parlough, line 600: Maybe we can be a bit more specific? As long as we drop "just" I'll be a happy camper.

```suggestion
  For the product, you can use `*` to multiply the three values together.
```

src/codelabs/dart-cheatsheet.md
```diff
@@ -559,58 +554,72 @@ void main() {
...
+  For `incrementValue1`, you can use the increment operator (`++`).
+  For `joinWithCommas`, try using the `join` method found in the `List` class.
```
parlough, line 602: Since this is in the solution rather than a hint now, "try" doesn't make sense anymore.

```suggestion
  For `joinWithCommas`, use the `join` method found in the `List` class.
```

src/codelabs/dart-cheatsheet.md
```diff
@@ -986,77 +988,83 @@ Here are some examples of function calls and returned values:
...
-{$ end hint.txt $}
+{$ end main.dart $}
 ```
 
+<details markdown="1">
+ <summary>
+   <b>Solution for positional parameters example</b>
+  </summary>
+  The `b`, `c`, `d`, and `e` parameters are null if they aren't provided by the
+  caller. The important thing, then, is to check whether those arguments are null
```
parlough, line 1053:
```suggestion
  caller. The important thing, then, is to check whether those arguments are `null`
```

src/codelabs/dart-cheatsheet.md
```diff
@@ -1417,35 +1395,53 @@ void main() {
...
-{$ end hint.txt $}
+{$ end main.dart $}
 ```
 
+<details markdown="1">
+ <summary>
+   <b>Solution for exceptions example</b>
+  </summary>
+  This exercise looks tricky, but it's really one big `try` statement.
+  Just call `untrustworthy` inside the `try`, and
```
parlough, line 1426:
```suggestion
  Call `untrustworthy` inside the `try`, and
```

src/codelabs/dart-cheatsheet.md
```diff
@@ -1507,55 +1503,55 @@ class MyClass {
...
-{$ end hint.txt $}
+{$ end main.dart $}
 ```
 
+<details markdown="1">
+ <summary>
+   <b>Solution for `this` example</b>
+  </summary>
+  This exercise has a one-line solution.
+  Just declare the constructor with
```
parlough, line 1546:
```suggestion
  Declare the constructor with
```

src/codelabs/dart-cheatsheet.md
```diff
@@ -1507,55 +1503,55 @@ class MyClass {
...
-}
-{$ end main.dart $}
-{$ begin solution.dart $}
-class MyClass {
-  final int anInt;
-  final String aString;
-  final double aDouble;
-  
-  MyClass(this.anInt, this.aString, this.aDouble);
+  // TODO: Create a constructor here.
```
parlough, line 1506: Should this be "the constructor" since we're instructed them to create a specific one?

```suggestion
  // TODO: Create the constructor here.
```

src/codelabs/dart-cheatsheet.md
```diff
@@ -1507,55 +1503,55 @@ class MyClass {
...
+ <summary>
+   <b>Solution for `this` example</b>
+  </summary>
+  This exercise has a one-line solution.
+  Just declare the constructor with
+  `this.anInt`, `this.aString`, and `this.aDouble`
+  as its parameters in that order.
+
+  ```dart    
+    MyClass(this.anInt, this.aString, this.aDouble);
```
parlough, line 1551: No need to indent here since it's alone in the snippet.

```suggestion
MyClass(this.anInt, this.aString, this.aDouble);
```

src/codelabs/dart-cheatsheet.md
```diff
@@ -1667,22 +1653,32 @@ void main() {
...
-{$ end hint.txt $}
+{$ end main.dart $}
 ```
 
+<details markdown="1">
+ <summary>
+   <b>Solution for initializer lists example</b>
+  </summary>
+  Two assignments need to happen:
+  `letterOne` should be `word[0]`, and `letterTwo` should be `word[1]`.
```
parlough, line 1673: Do you think it's worth being more specific here?

```suggestion
  `letterOne` should be assigned to `word[0]`, and
  `letterTwo` should be assigned to `word[1]`.
```

src/codelabs/dart-cheatsheet.md
```diff
@@ -1732,59 +1728,57 @@ class Color {
...
+      errs.add('Called Color.black() and got a Color with green equal to \n ${result.green} instead of the expected value (0).');
     }
 
     if (result.blue != 0) {
-  errs.add('Called Color.black() and got a Color with blue equal to ${result.blue} instead of the expected value (0).');
+  errs.add('Called Color.black() and got a Color with blue equal to \n ${result.blue} instead of the expected value (0).');
     }
   } catch (e) {
-    _result(false, ['Called Color.black() and got an exception of type ${e.runtimeType}.']);
+    _result(false, ['Called Color.black() and got an exception of type \n ${e.runtimeType}.']);
```
parlough, line 1755: Looks like this step still need to be migrated away from `_result`.

src/codelabs/dart-cheatsheet.md
```diff
@@ -1732,59 +1728,57 @@ class Color {
...
-{$ end hint.txt $}
+{$ end main.dart $}
 ```
 
+<details markdown="1">
+ <summary>
+   <b>Solution for named constructors example</b>
+  </summary>
+  The declaration for your constructor should be `Color.black() {}`.
+  Inside the braces, set `red`, `green`, and `blue` to zero.
```
parlough, line 1773: We must have not updated this hint when it was changed to use an initializer list for null safety.

How does this sound?

```suggestion
  The declaration for your constructor should begin with `Color.black(): `.
  In the initializer list (after the colon), set `red`, `green`, and `blue` to `0`.
```

src/codelabs/dart-cheatsheet.md
```diff
@@ -1901,7 +1861,7 @@ void main() {
   } on Error {
     _throwed = true;
   } catch (e) {
-    _result(false, ['Called IntegerSingle.fromList([]) and got an exception of type ${e.runtimeType}.']);
+    _result(false, ['Called IntegerSingle.fromList([]) and got an exception of \n type ${e.runtimeType}.']);
```
parlough, line 1864: This sample also needs to be updated away from `_result`.

src/codelabs/dart-cheatsheet.md
```diff
@@ -2114,51 +2088,60 @@ class Recipe {
...
+      errs.add('Called Recipe([\'1 egg\', \'Pat of butter\', \'Pinch salt\'], 120, 200) \n and got an object with a calorie value of ${obj.calories} rather than the expected value (120).');
     }
     
     if (obj.milligramsOfSodium != 200) {
-      errs.add('Called Recipe([\'1 egg\', \'Pat of butter\', \'Pinch salt\'], 120, 200) and got an object with a milligramsOfSodium value of ${obj.milligramsOfSodium} rather than the expected value (200).');
+      errs.add('Called Recipe([\'1 egg\', \'Pat of butter\', \'Pinch salt\'], 120, 200) \n and got an object with a milligramsOfSodium value of ${obj.milligramsOfSodium} rather than the expected value (200).');
     }
   } catch (e) {
-    _result(false, ['Tried calling Recipe([\'1 egg\', \'Pat of butter\', \'Pinch salt\'], 120, 200) and received a null.']);
+    _result(false, ['Tried calling Recipe([\'1 egg\', \'Pat of butter\', \'Pinch salt\'], 120, 200) \n and received a null.']);
```
parlough, line 2116: This sample also needs to be updated away from `_result`.

src/codelabs/dart-cheatsheet.md
```diff
@@ -1913,73 +1873,90 @@ void main() {
...
-{$ end hint.txt $}
+{$ end main.dart $}
 ```
 
+<details markdown="1">
+ <summary>
+   <b>Solution for factory constructors example</b>
+  </summary>
+  Inside the factory constructor,
+  check the length of the list and create an
```
parlough, line 1943:
```suggestion
  check the length of the list, then create and return an
```

parlough (approved):
This is great @MaryaBelanger!! Thanks so much for taking this on, especially unplanned and on a short notice.

I was worried how it would turn out without all of the original DartPad functionality, but it's not too different and I think the core benefits of the page still remain.

I left a few suggestions and questions, but nothing blocking.

A few sample tests do need still need to be updated to move away from `_result`, but I'd be happy to that as follow-up if you're busy with other work.

src/codelabs/dart-cheatsheet.md
MaryaBelanger: Oh, sure! Didn't know

src/codelabs/dart-cheatsheet.md
MaryaBelanger: I put:

    Add a list, set, or map literal after each equals sign.
    Remember to specify the types for the empty declarations,
    since they can't be inferred.

I think it's ok just as a description of the solution, not necessarily a hint

src/codelabs/dart-cheatsheet.md
MaryaBelanger: Agree with specifying the action, but wouldn't it be more correct to say it the other way around? Like, I'd read that code as "`word[0]` is assigned to `letterOne`", so I guess the instruction would be:

`letterOne` should be assigned `word[0]`

I put this in my edit but lmk if that's less accurate

src/codelabs/dart-cheatsheet.md
MaryaBelanger: Nice catch!

MaryaBelanger:
@parlough Thanks for the thorough review. I went through and there were a lot of `_results` still in use so I got rid of all of them. I also changed any "try" language in the solution/hint text to "use" like you suggested in a few places (there were a few more).

parlough (approved):
Thank you so much, looks great!

PR #5402, opened 2023-12-11
Concurrency overview page update

ericwindmill:
This is the first PR of two that will update isolate documentation. It was decided that we should split [the original PR](https://github.com/dart-lang/site-www/pull/5332) into two.

This PR updates documentation at language/concurrency with more information.

Fixes https://github.com/dart-lang/site-www/issues/3720

---


src/language/concurrency.md
```diff
@@ -16,87 +19,111 @@ prevpage:
...
 </style>
 
-Dart supports concurrent programming with async-await, isolates, and
-classes such as `Future` and `Stream`.
-This page gives an overview of async-await, `Future`, and `Stream`,
-but it's mostly about isolates.
+This page contains a conceptual overview of how concurrent programming works in
+Dart. It explains the event-loop, async language features, and isolates from
+a high-level. For more practical code examples of using async features,
+read the [Asynchrony support](/language/async).
```
ericwindmill, line 25: This should be updated in the next PR to also reference the new isolates page.

src/language/concurrency.md
```diff
@@ -226,314 +280,195 @@ using asynchronous operations as necessary.
...
-However, now the isolate sends a [closure][].
-Closures are less limited than typical named functions,
-both in how they function and how they're written into the code.
-In this example, `Isolate.run()` executes what looks like local code, concurrently.
-In that sense, you can imagine `run()` to work like a control flow operator
-for "run in parallel".
-
-[closure]: /language/functions#anonymous-functions
-
-### Sending multiple messages between isolates
```
ericwindmill, line 446: This section is removed until the next PR. Per suggestion by @parlough  (which I agree with), I'll want to reference the isolate example from the new isolates page rather than include it on both pages. So this will need to be added once there's something to reference.

src/language/concurrency.md
```diff
@@ -144,80 +171,107 @@ void main() [!async!] {
...
+In the following example, the stream created with `Stream.periodic`
+repeatedly emits a new `int` value every second.
 
-## How isolates work
+<?code-excerpt "lib/stream_syntax.dart"?>
+```dart
+Stream<int> stream = Stream.periodic(const Duration(seconds: 1), (i) => i * i);
+```
+
+#### await-for and yield
```
MaryaBelanger, line 205: This is a great section! I didn't know anything about streams and found them intimidating before and now they seem simple and approachable

src/language/concurrency.md
```diff
@@ -144,80 +171,107 @@ void main() [!async!] {
...
-[error prone](https://en.wikipedia.org/wiki/Race_condition#In_software) and
-can lead to complicated code.
+## Isolates
 
-Instead of threads, all Dart code runs inside of isolates.
-Each isolate has its own memory heap,
-ensuring that none of the state in an isolate is accessible from
-any other isolate.
-No shared state between isolates means concurrency complexities like 
+Dart also supports concurrency via isolates. Most modern devices have multi-core
```
MaryaBelanger, line 231: Since this section probably will be linked to from around the site (and outside the site) as the defacto isolate definition, we can't assume the reader read anything before this so "also" doesn't really work. We can say:

 ```suggestion
Dart supports concurrency via isolates, in addition to [asynchronous APIs](#types-and-syntax-for-asynchronous-programming).
Most modern devices have multi-core
```

The link to the earlier section is, again, because this might be the first part of the page someone reads. 

(Idk if saying "asynchronous APIs" makes sense though so feel free to change that)

File: src/language/concurrency.md [hunk 5]
```diff
@@ -16,87 +19,111 @@ prevpage:
...
-The synchronous `File` method [`readAsStringSync()`][]
-reads a file synchronously,
-blocking until the file is either fully read or an error occurs.
-The method then either returns an object of type `String`
-or throws an exception.
-The asynchronous equivalent, [`readAsString()`][],
-immediately returns an object of type `Future<String>`.
-At some point in the future,
-the `Future<String>` completes with either a string value or an error.
+## Types and syntax for asynchronous programming
```
MaryaBelanger, line 95: Once I got to the `## Isolates` section, I started thinking about how these sections will probably be used as the link to "define" async and isolates around the site, so maybe a more definitive section title would be better here. Also, one that mirrors the `## Isolates` section since they're kind of sister sections.

```suggestion
## Asynchronous programming
```

And then add the "types and syntax" thing to the intro sentence (commented below)

src/language/concurrency.md
```diff
@@ -16,87 +19,111 @@ prevpage:
...
 
-[`readAsStringSync()`]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-io/File/readAsStringSync.html
-[`readAsString()`]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-io/File/readAsString.html
+If you're already familiar with `Future`, `Stream`, and async-await,
```
MaryaBelanger, line 97:
```suggestion
This section summarizes the different types and syntaxes of asynchronous programming in Dart.
If you're already familiar with `Future`, `Stream`, and async-await,
```

src/language/concurrency.md
```diff
@@ -144,80 +171,107 @@ void main() [!async!] {
...
-any other isolate.
-No shared state between isolates means concurrency complexities like 
+Dart also supports concurrency via isolates. Most modern devices have multi-core
+CPUs. To take advantage of multiple cores, developers sometimes use
+shared-memory threads running concurrently. However, shared-state concurrency is
+[error prone](https://en.wikipedia.org/wiki/Race_condition#In_software) and can
+lead to complicated code.
+
+Instead of threads, all Dart code runs inside isolates. Each isolate has its
+own global fields, ensuring that none of the state in an isolate is accessible
```
MaryaBelanger, line 238: The last paragraph seems like it fits better up here (deleted in subsequent comment):

```suggestion
Instead of threads, all Dart code runs inside isolates.
Using isolates, your Dart code can perform multiple independent tasks at once.
Isolates are like threads or processes, but each isolate has its own memory
and a single thread running an event loop.

Each isolate has its own global fields,
ensuring that none of the state in an isolate is accessible
```

Notice I removed the clause "using additional processor cores if they're available." While that might be the case, I think it introduces unnecessary uncertainty, like "do I have additional processor cores available? what if I don't, can I use isolates? what happens if I try?" and stuff like that. If there's nothing the user can do about it and we're not going to elaborate on what to do otherwise, probably best not to mention it.

If you think it's the kind of thing that's important for developers to know or that they'd want to know, add it back (I don't have that context)

src/language/concurrency.md
```diff
@@ -226,314 +280,195 @@ using asynchronous operations as necessary.
...
-A common case, shown in the following figure,
-is spawning a simple worker isolate that
-performs a computation and then exits.
-The worker isolate returns its result in a message when the worker exits.
+If your app's UI becomes unresponsive due to a time-consuming
+computation—[parsing a large JSON file][json], for example—consider offloading
+that computation to a worker isolate, often called a _background worker._
+A common case, shown in the following figure, is spawning a simple worker
+isolate that performs a computation and then exits. The worker isolate returns
+its result in a message when the worker exits.
```
MaryaBelanger, line 325:
```suggestion
its result in a message when it exits.
```

MaryaBelanger (approved):
LGTM % some comments, I think there's a misprint or two that _needs_ to be addressed, but other than that the rest are just suggestions

atsansone:
@parlough : I think this is affected by the build issue in https://github.com/dart-lang/site-www/pull/5410. Can you re-run that action once it merges?

PR #5408, opened 2023-12-14
Add note about `--git-ref` to `pub global`

nilsreichardt:
I wanted to know how to specify a commit for a package I want to install, and I noticed that the option isn't in the documentation.


parlough (approved):
Thanks for thinking to add this!

PR #5409, opened 2023-12-14
Remove PATH instructions for Linux install

### LINKED ISSUE #4609: Remove the "Modify PATH for access to all Dart binaries" section to simplify the installation in 'Get the Dart SDK' Linux page
### Page URL

https://dart.dev/get-dart/

### Page source

https://github.com/dart-lang/site-www/tree/main/src/get-dart/index.md

### Describe the problem

I think the Dart Linux installation instructions may be more complex than necessary. On my machine, installing Dart through `apt` automatically added a symbolic link to the `/usr/bin` directory:
 
```
tims@penguin:~$ ls -al /usr/bin/dart
lrwxrwxrwx 1 root root 20 Feb  7 10:57 /usr/bin/dart -> ../lib/dart/bin/dart
```

As a result, the exhortation to add `/usr/lib/dart` to the search path (the "Modify PATH for access to all Dart binaries" section) seems redundant.

It was probably useful in the past when Dart had a number of different CLI utilities, but now they have been unified into a single `dart` tool, the symlink should be sufficient.

### Expected fix

Remove the "Modify PATH for access to all Dart binaries" section to simplify the installation instructions.

### Additional context

/cc @mit-mit for validation that I'm right here :)

atsansone:
Fixes #4609

### DIFF
--- a/src/get-dart/_linux.md
+++ b/src/get-dart/_linux.md
@@ -27,18 +27,3 @@ $ sudo apt-get install dart
 
 Alternatively, download Dart SDK [as a Debian package](#){:.debian-link-stable}
 in the `.deb` package format.
-
-#### Modify PATH for access to all Dart binaries
-
-After installing the SDK, **add its `bin` directory to your `PATH`**. For example,
-use the following command to change `PATH` in your active terminal session:
-
-```terminal
-$ export PATH="$PATH:/usr/lib/dart/bin"
-```
-
-To change the PATH for future terminal sessions, use a command like this:
-
-```terminal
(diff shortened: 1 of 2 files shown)


parlough (approved):
Thanks for going back to this longstanding request!

I think removing this is fine, but there's one case where this is still needed (`dartaotruntime`).

Could you add a note to https://dart.dev/tools/dartaotruntime that to access `dartaotruntime`, the `bin` folder in your Dart SDK installation must be added to the path.

As for the build breaking, https://github.com/dart-lang/site-www/pull/5410 will fix it :)

PR #5419, opened 2023-12-20
Update diagnostic messages and linter rules

parlough:
These changes were previously reviewed on Gerrit before landing in the SDK.


MaryaBelanger:
I think these are the diagnostics that are causing issues: https://github.com/dart-lang/sdk/issues/54416

If that's correct, should we wait until that's sorted out to copy them here?

bwilkerson:
One of the problems was that the error range wasn't highlighted. That's been fixed.

The remaining problems are compilation errors in the code (things like declaring a non-nullable top-level variable without initializing it). I don't know how likely that is to create problems for users.

I also don't know how long it will take to get them updated given the holiday season.

MaryaBelanger:
Thanks for the context @bwilkerson 

I'm ok with landing this then, but I wonder @parlough would it be ok to just manually add the fixes from https://dart-review.googlesource.com/c/sdk/+/342941 ? 

It's basically just adding `external` on `NATIVE_FIELD_INVALID_TYPE` and `NATIVE_FIELD_MISSING_TYPE`, lines [14091](6a32d77fc362261ec94fa06e3152e56256eccdf4609d7eb4252c8e9f1d623de2R14091) and [14102](https://github.com/dart-lang/site-www/pull/5419/files#diff-6a32d77fc362261ec94fa06e3152e56256eccdf4609d7eb4252c8e9f1d623de2R14102),  and [14130](https://github.com/dart-lang/site-www/pull/5419/files#diff-6a32d77fc362261ec94fa06e3152e56256eccdf4609d7eb4252c8e9f1d623de2R14130) and [14144](https://github.com/dart-lang/site-www/pull/5419/files#diff-6a32d77fc362261ec94fa06e3152e56256eccdf4609d7eb4252c8e9f1d623de2R14144), respectively. And then adding `final` to `NATIVE_FIELD_NOT_STATIC`, line [14209](https://github.com/dart-lang/site-www/pull/5419/files#diff-6a32d77fc362261ec94fa06e3152e56256eccdf4609d7eb4252c8e9f1d623de2R14209).

Idk if that messes up the process you use for diagnostics but I figured from an information standpoint it'd probably safe? If not, again just land it, no problem imo.

MaryaBelanger (approved):

bwilkerson:
I was pleasantly surprised by how quickly the issue was fixed. The try bots that test the changes haven't completed, so I can't confirm that all of the issues are fixed by those changes, but I suspect they will be.

Applying changes to the website that haven't yet landed in `analyzer/messages.yaml` is fine as long as those changes will also be applied in that file (which I believe they will).

parlough:
Simon has always been super helpful and on top of things =D

PR #5429, opened 2023-12-26
Update to slimmer firebase-tools v13

parlough:
This reduces some deprecation notices and has improved support for Node v20 which we are using.


parlough:
Superseded by https://github.com/dart-lang/site-www/pull/5445 :)

PR #5431, opened 2023-12-26
Added info on dependency version constraints

### LINKED ISSUE #4811: Add best practice about setting version constraints to 'Package dependencies' page
### Page URL

https://dart.dev/tools/pub/dependencies.html

### Page source

https://github.com/dart-lang/site-www/tree/main/src/tools/pub/dependencies.md

### Describe the problem

From a conversation with @munificent: 

@timsneath:
> I've often worked on the assumption that looser dependencies are good (e.g. foo: ^2.0.0), because they increase the chances of the version solver being successful, but tighter dev_dependencies are good (e.g. test: ^1.21.4), because there are no transitive concerns. Am I right? I've been looking for somewhere where this is documented, and can't find it -- which makes me wonder if my priors are false...

@munificent:
> I work under the same assumption, Tim. I don't know if we've ever written it down anywhere.

### Expected fix

The Best Practices section of this doc (https://dart.dev/tools/pub/dependencies#best-practices) would be a great place to add a comment or two here.

### Additional context

_No response_

atsansone:
Fixes #4811

### DIFF
--- a/src/content/tools/pub/dependencies.md
+++ b/src/content/tools/pub/dependencies.md
@@ -443,6 +443,29 @@ To identify dependencies in your app or package that
 aren't on the latest stable versions,
 use [`dart pub outdated`][].
 
+### Tighten version constraints for dev dependencies
+
+A dev dependency defines a package that you need only when developing.
+A finished app won't need these packages.
+Examples of these packages include tests or code generation tooling.
+Set the version constraints of packages in [`dev_dependencies`][dev-dep]
+to have a lower bound of the latest version on which your package depends.
+
+Tightening the version constraints of your dev dependencies might
+resemble the following:
+
+```yaml
+dev_dependencies:
+  build_runner: ^2.4.8
+  lints: ^2.1.1
+  test: ^1.25.1
+```
+
+This YAML sets the `dependencies` to the major version and the
+`dev_dependencies` to a specific patch version.
+
+[dev-dep]: /tools/pub/dependencies#dev-dependencies
+
 ### Test whenever you update package dependencies
 
 If you run [`dart pub upgrade`][] without updating your pubspec,
--- a/src/content/tools/pub/dependencies.md
+++ b/src/content/tools/pub/dependencies.md
@@ -419,10 +419,12 @@ are ignored by all users of your package.
 ## Best practices
 
 Be proactive in managing your dependencies.
-Ensure that your packages use the freshest versions possible.
-Any stale dependencies could themselves depend on other stale packages.
-These stale versions can have a negative impact on
-the stability, performance, and quality of apps.
+Ensure that your packages depend on the freshest versions of packages
+when possible.
+If your package depends on a stale package,
+that stale package may depend on other stale packages in its dependency tree.
+Stale versions of packages can have a negative impact on
+the stability, performance, and quality of your app.
 
 We recommend the following best practices for package dependencies.
 
--- a/src/content/tools/pub/dependencies.md
+++ b/src/content/tools/pub/dependencies.md
@@ -418,12 +418,9 @@ are ignored by all users of your package.
 
 ## Best practices
 
-It's important to actively manage your dependencies and
-ensure that your packages use the freshest versions possible.
-If any dependency is stale,
-then you might have not only a stale version of that package,
-but also stale versions of other packages in your dependency graph that
-depend on that package.
+Be proactive in managing your dependencies.
+Ensure that your packages use the freshest versions possible.
+Any stale dependencies could themselves depend on other stale packages.
 These stale versions can have a negative impact on
 the stability, performance, and quality of apps.
 
--- a/src/content/tools/pub/dependencies.md
+++ b/src/content/tools/pub/dependencies.md
@@ -461,8 +461,7 @@ dev_dependencies:
   test: ^1.25.1
 ```
 
-This YAML sets the `dependencies` to the major version and the
-`dev_dependencies` to a specific patch version.
+This YAML sets the `dev_dependencies` to the latest patch versions.
 
 [dev-dep]: /tools/pub/dependencies#dev-dependencies
 


src/tools/pub/dependencies.md
```diff
@@ -442,6 +439,32 @@ To identify dependencies in your app or package that
 aren't on the latest stable versions,
 use [`dart pub outdated`][].
 
+### Use tight dev_dependencies
```
parlough, line 442: We tend to use "dependency" to refer to the package depended on, so it's not really the dependencies themselves that are "loose" or "tight", but rather the version constraint we specify for each of them.

Perhaps: 

```suggestion
### Use tight version constraints for dev dependencies
```

src/tools/pub/dependencies.md
```diff
@@ -442,6 +439,32 @@ To identify dependencies in your app or package that
...
+
+Dependencies can be loose, specifying dependency at the highest version digit
+like `^2.0.0`, or tight, specifying dependency at the lowest version digit like
+`^1.2.4`.
+
+In your `pubspec.yaml` file, follow these recommendations.
```
parlough, line 448: I'm not sure we need to mention this as this whole section is about dependencies in the `pubspec.yaml` file and recommendations for those dependencies. I think we can just focus on the best practices.

src/tools/pub/dependencies.md
```diff
@@ -442,6 +439,32 @@ To identify dependencies in your app or package that
...
+
+* Set regular `dependencies` loose to increase the chance of a version solver success.
```
parlough, line 450: This section's title only mentions dev dependencies. Rather than expanding the focus of this section, I'd consider dropping this as the best practices for normal dependencies are much more nuanced and very situation dependent.

That's not to say we shouldn't have guidelines for them, but if we do have guidelines, we have to make them very clear.

src/tools/pub/dependencies.md
```diff
@@ -442,6 +439,32 @@ To identify dependencies in your app or package that
...
+* Set [`dev_dependencies`][dev-dep] tighter as they lack transitive concerns.
```
parlough, line 451: If we can drop the note on normal dependencies, I think we should just combine this with the above text to better colocate what a tight dependency constraint is and why you should use them for dev dependencies.

src/tools/pub/dependencies.md
```diff
@@ -442,6 +439,32 @@ To identify dependencies in your app or package that
...
+
+```yaml
```
parlough, line 453: This code block confused me when reading as I didn't know why it was there. I think it can be dropped, but if you want to keep it, do add some more introduction/content, and perhaps some inline comments.

I also recommend using a non-Flutter example that doesn't use `sdk` dependencies, as they are more obscure and perhaps less familiar or completely unknown to non-Flutter developers.

src/tools/pub/dependencies.md
parlough: Also if we keep using the wording "transitive concerns", could we add an example or two of those concerns? I'm afraid the terminology will be less familiar and potentially unclear about what concerns are being referred to.

I'd be happy to help expand here if you need any ideas. Let me know!

src/tools/pub/dependencies.md
```diff
@@ -442,6 +439,32 @@ To identify dependencies in your app or package that
 aren't on the latest stable versions,
 use [`dart pub outdated`][].
 
+### Use tight dev_dependencies
+
+Dependencies can be loose, specifying dependency at the highest version digit
+like `^2.0.0`, or tight, specifying dependency at the lowest version digit like
+`^1.2.4`.
```
parlough, line 446: I don't think we should necessarily define this in terms of which "version digit" as that's not a term we've used on the site. If someone is reading the best practices I think we can assume prerequisite knowledge of what each portion of the version specifier means, so let's specify it in terms of that (major, minor, patch).

Also, as mentioned above, it's best to refer to the constraints being tight or loose.

So in the case of a tight version constraint, that means the lower bound being the newest compatible version.

src/tools/pub/dependencies.md
```diff
@@ -416,16 +416,13 @@ are ignored by all users of your package.
...
 
-It's important to actively manage your dependencies and
-ensure that your packages use the freshest versions possible.
-If any dependency is stale,
-then you might have not only a stale version of that package,
-but also stale versions of other packages in your dependency graph that
-depend on that package.
+Be proactive in managing your dependencies.
+Ensure that your packages use the freshest versions possible.
+Any stale dependencies could themselves depend on other stale packages.
```
parlough, line 421: This is a nice simplification of this text, but I'm worried it loses its meaning. It seems the original is actually talking about the other direction in which various packages in your dependency graph depend on that (stale) package. But if you are depending on a stale version of that package, the version solver might end up having to pull stale versions of your other dependencies, since their newer versions aren't compatible with the stale version you're depending on.

It's a bit intricate and this new direction is also a slight concern, so maybe it's fine to lose that meaning?

parlough (changes requested):
Thanks for taking this on @atsansone!

I have a few comments and suggestions, including reducing the scope a bit to focus on what we can be more confident in suggesting. Let me know if you have any questions, would like to chat further, or if you'd like me to further propose some text.

Thanks!

src/tools/pub/dependencies.md
MaryaBelanger:
```suggestion
### Tighten version constraints for dev dependencies
```

src/tools/pub/dependencies.md
MaryaBelanger:
```suggestion
A stale dependency could mean stale versions of other packages in your dependency graph
that depend on that package. 
```
To add back the correct relationship (I think?)

src/tools/pub/dependencies.md
MaryaBelanger: I'd skip the "loose" definition here if the best practice is just "tighten", define a "tight" constraint ("_that means the lower bound being the newest compatible version._"), and like Parker said [below](https://github.com/dart-lang/site-www/pull/5431/files#r1449864165), add the reasoning and definition of the reasoning to the opening description of the best practice, e.g.

    Specify dev dependencies at their lower bound, or newest compatible version,
    to ... (what does this achieve / avoid?)  
    OR
    because ... (something something transitive concerns?)

src/tools/pub/dependencies.md
atsansone: It's at the top of the page.

atsansone:
@parlough : PTAL.

src/tools/pub/dependencies.md
```diff
@@ -442,6 +441,34 @@ To identify dependencies in your app or package that
...
 
+### Tighten version constraints for dev_dependencies
+
+A dev dependency retrieves packages that you need only when developing.
+These packages aren't needed in a finished app, like tests or examples.
+Set the versions of package in [`dev_dependencies`][dev-dep]
+to the exact version on which your package depends.
+
+An example of how to set versions for dependencies would resemble the
+following `pubspec.yaml` example.
```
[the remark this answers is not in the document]
atsansone: Fixed.

src/content/tools/pub/dependencies.md
```diff
@@ -444,6 +443,29 @@ To identify dependencies in your app or package that
...
+
+```yaml
+dev_dependencies:
+  build_runner: ^2.4.8
+  lints: ^2.1.1
+  test: ^1.25.1
+```
+
+This YAML sets the `dependencies` to the major version and the
+`dev_dependencies` to a specific patch version.
```
parlough, line 465: Since you dropped the `dependencies` from the example, I guess need to modify this text:

```suggestion
This YAML sets the `dev_dependencies` to the latest patch versions.
```

parlough (approved):
Thanks for those adjustments @atsansone! One more area that needs to be updated, but otherwise looks great.

PR #5433, opened 2023-12-27
Updated image usage

### LINKED ISSUE #1789: Make sure image files are used

atsansone:
Fixes #1789


src/index.html
parlough, line 1: Looks like [L204](https://github.com/dart-lang/site-www/blob/20dc96baba986b378f1f73c6ac3153bd2bf21afb/src/index.html#L204) is still using the old file name.

While you're here, consider standardizing the "optimise" spelling in the file names and alt text to "optimize" to be consistent with our site and the style guide.

parlough (approved):
Thanks for this cleanup @atsansone!

I think a few of the diagram files may still be used, but if they need to be updated, they likely need to be remade for current site standards anyway.

Just one small issue and suggestion:

Issue #5285, opened 2023-10-25
Update next-generation JS interop discussion 


domesticmouse:
### Page URL

https://dart.dev/web/js-interop.html

### Page source

https://github.com/dart-lang/site-www/tree/main/src/web/js-interop.md

### Describe the problem

This page mentions "when Dart 3 lands."

Which is *checks watch* a while ago now.

### Expected fix

Update the content of this page to reflect Dart 3.1 as current stable.

domesticmouse:
Adding https://github.com/dart-lang/sdk/issues/35084 is useful context for this issue.

parlough:
@kevmoo @srujzs @sigmundch 

Would you be able to take a look at [this page](https://dart.dev/web/js-interop#next-generation-js-interop-preview) and see if everything still is correct and matches the latest suggestions and expectations? Also, should we call out anything else? Suggest trying out `package:web`, etc?

srujzs:
There are a few changes here that I think we need:

- Dart 3 -> Dart 3.3
- We want people to `dart:js_interop` and `dart:js_interop_unsafe` wherever they can instead of `package:js` and `dart:js_util`. We want to remove support for the latter two in dart2wasm. `dart:js_interop` should give you all that you need for `@staticInterop` interfaces that you could previously define with `package:js`. Both `dart:js_interop` and `dart:js_interop_unsafe` should be enough to cover most of the functionality in `dart:js_util`. There are a few small pieces missing that I need to add to reach parity.
- I think we're ready for users to try out extension types pretty soon here (maybe once we hit the 3.3 cutoff EOY).
- It's worth suggesting `package:web` but we're still not quite at stable yet. I'd like to migrate that package to extension types and address some of the missing gaps before we're ready to commit to few breaking changes.

I'm chatting with @MaryaBelanger next week to discuss docs for interop in general now that we're basically feature-complete.

MaryaBelanger:
Adding to note we'll probably continue to work in this draft/outline PR https://github.com/dart-lang/site-www/pull/4578

PR #5436, opened 2023-12-28
Update next-gen JS interop docs for current situation

parlough:
This is more of a stop gap to correct the page of current inaccuracies until we have a chance to create the actual static interop docs. Primarily pushing new uses to `dart:js_interop` rather than `package:js`.

Fixes https://github.com/dart-lang/site-www/issues/5285

**Staged:** https://dart-dev--pr5436-misc-js-interop-3-2-akb3v6pt.web.app/interop/js-interop


MaryaBelanger (approved):
Thank you, makes a big difference!

PR #5437, opened 2023-12-28
Accuracy adjustment of non-null assertion operator

parlough:
We haven't come to a conclusion yet, and I'd like to leave that conversation on hold, but we should at least improve the from status quo which is currently more inaccurate and inconsistent with the [null assert pattern type](https://dart.dev/language/pattern-types#null-assert).


MaryaBelanger (approved):
Good idea, thank you!

PR #5444, opened 2024-01-04
Add next-gen JS interop short link

parlough:
I'd like to add this to link to from DartPad and the Flutter site. Perhaps elsewhere as well (such as tools or docs in the SDK).

That way we can adjust the destination of the redirect in the future as we add more documentation.


MaryaBelanger (approved):

parlough:
Thanks for the quick review @MaryaBelanger!

PR #5447, opened 2024-01-04
Add a Cloud Build config to deploy website to Firebase.

khanhnwin:
As the title says. This is the Cloud Build configuration that we'll use to build dart.dev and publish to Firebase hosting. 

Note: This current config currently publishes to a staging Flutter project to confirm everything works on main. 

Once this PR lands and we've confirmed deploy works as expected, 
- [ ] I'll open another PR to switch to prod project `dart-dev`
- [ ] Decommission the [deploy.yaml GitHub action](https://github.com/dart-lang/site-www/blob/main/.github/workflows/deploy.yml). 
- [ ] Update the trigger branch to `main`

cc @drewroengoogle @parlough


cloud_build/deploy.yaml
```diff
@@ -0,0 +1,19 @@
...
+      - |-
+        set -e
+        echo "Building the website using a makefile..."
+        make build
+  - name: gcr.io/flutter-dev-230821/firebase-ghcli
+    entrypoint: '/bin/bash'
+    args:
+      - '-c'
+      - |-
+        firebase deploy --project=flutter-website-staging-349021 --only=hosting
```
drewroengoogle, line 17: Is this supposed to be to a dart project instead?

khanhnwin: Yes! I'm going to open another PR to change that once this lands and we've confirmed it deploys properly. I don't want to remove the current deploy GitHub action until we confirm this deploy works & I don't want them to deploy on top of each other.

cloud_build/deploy.yaml
```diff
@@ -0,0 +1,19 @@
...
+options:
```
parlough, line 18: The other ones you two added had to specify what secrets were available: 

https://github.com/dart-lang/site-www/blob/2b4e25631c9d9cd194febe98fc04f575b403cb0d/cloud_build/stage.yaml#L26-L29

Is that not necessary in this case?

khanhnwin: Nope, the GH_PAT_TOKEN secret is only required for the staging configuration because it needs to add a comment back to the GitHub PR with a staging link. This is just a one way deploy to Firebase Hosting so no GitHub access needed.

parlough: That makes a lot of sense...

Well this senseless developer say thanks for the explanation :)

parlough (approved):
# 🫡

Thanks!

cloud_build/deploy.yaml
khanhnwin: It made more sense to me _after_ I wrote out that explanation so thanks for asking! hehe

PR #5449, opened 2024-01-04
Change deploy project to dart-dev

khanhnwin:
Addresses remaining AI's from #5447.

- [x] Swapped out deploy project to `dart-dev`
- [x] I've manually disabled the [deploy.yaml GitHub action](https://github.com/dart-lang/site-www/blob/main/.github/workflows/deploy.yml) workflow from the "Actions" menu. 
- [x] Updated the Cloud Build trigger to go off on push to `main` branch

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,3 +1,6 @@
+# This GitHub workflow has been disabled and is no longer used. 
+# Website deploys are now done through Cloud Build. 
+# See https://github.com/dart-lang/site-www/pull/5447
 name: deploy
 
 on:
--- a/cloud_build/deploy.yaml
+++ b/cloud_build/deploy.yaml
@@ -14,6 +14,6 @@ steps:
     args:
       - '-c'
       - |-
-        firebase deploy --project=flutter-website-staging-349021 --only=hosting
+        firebase deploy --project=dart-dev --only=hosting
 options:
   logging: CLOUD_LOGGING_ONLY
--- a/cloud_build/deploy.yaml
+++ b/cloud_build/deploy.yaml
@@ -9,6 +9,7 @@ steps:
         set -e
         echo "Building the website using a makefile..."
         make build
+        make write-prod-robots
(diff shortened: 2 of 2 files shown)


cloud_build/deploy.yaml
parlough, line 1: I forgot about this. Can you add a call to `make write-prod-robots` after `make build`. It won't be needed for much longer, but it is for now to configure the `robots.txt` file.

.github/workflows/deploy.yml
```diff
@@ -1,3 +1,6 @@
+# This GitHub workflow has been disabled and is no longer used. 
+# Website deploys are now done through Cloud Build. 
+# See https://github.com/dart-lang/site-www/pull/5447
```
parlough, line 3: Thanks for adding a comment to explain this. However, can we just delete the file or do you foresee a reason it will be needed again soon?

parlough (approved):
Thanks @khanhnwin!

One request and one question:

.github/workflows/deploy.yml
khanhnwin: That's a fair point! I was leaving it mostly so that the specification deployment folks can have a reference when moving their stuff over.

cloud_build/deploy.yaml
khanhnwin: Yes! Just added. WDYT?

.github/workflows/deploy.yml
parlough: Ah I see. That's fine then since you disabled it :)

khanhnwin: It'll be good to remove once all the cloud build migrations are completed, though. I'll file an issue so that we remember.

parlough (approved):
Looks good to me. Thanks @khanhnwin!

Consistent staging and deploys here we come. 🚅

khanhnwin:
wahoooo!

khanhnwin:
We're good! https://github.com/dart-lang/site-www/runs/20180528321

Khammunh:
Hi sister can you help me write code scan barcode flutter project?

PR #5459, opened 2024-01-12
Add go/security-advisories redirection

szakarias:
For now this points at the issue tracking the documentation. The go link is to be used in the dart sdk changelog.

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -181,7 +181,7 @@
       { "source": "/go/publishing-with-service-account", "destination": "/tools/pub/automated-publishing#publishing-from-google-cloud-build", "type": 301 },
       { "source": "/go/sdk-constraint", "destination": "/tools/pub/pubspec#sdk-constraints", "type": 301 },
       { "source": "/go/sdk-version-pinning", "destination": "https://github.com/dart-lang/sdk/wiki/Flutter-Pinned-Packages", "type": 301 },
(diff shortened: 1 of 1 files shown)


firebase.json
```diff
@@ -181,6 +181,7 @@
       { "source": "/go/publishing-with-service-account", "destination": "/tools/pub/automated-publishing#publishing-from-google-cloud-build", "type": 301 },
       { "source": "/go/sdk-constraint", "destination": "/tools/pub/pubspec#sdk-constraints", "type": 301 },
       { "source": "/go/sdk-version-pinning", "destination": "https://github.com/dart-lang/sdk/wiki/Flutter-Pinned-Packages", "type": 301 },
+      { "source": "/go/security-advisories", "destination": "https://github.com/dart-lang/site-www/issues/5458", "type": 301 },
```
parlough, line 184: Since the SDK could potentially have security advisories as well, could we make this more specific?

```suggestion
      { "source": "/go/pub-security-advisories", "destination": "https://github.com/dart-lang/site-www/issues/5458", "type": 301 },
```

parlough (commented):
Thanks for thinking to add this! One question:

parlough (approved):
I'll approve this in case you are already using this somewhere, but do consider my suggestion to make the redirect link a bit more specific. Thanks! :D

firebase.json
szakarias: I didn't yet land the place where I am using it. I like your suggestion, so I'll make a new PR.

PR #5467, opened 2024-01-17
Verified publishers: add recommendation for several members

szakarias:
We get many support emails asking for regaining access to a verified publisher after the original creator is not at the company anymore.

### DIFF
--- a/src/tools/pub/publishing.md
+++ b/src/tools/pub/publishing.md
@@ -128,10 +128,12 @@ To create a verified publisher, follow these steps:
    * When the verification flow is complete, return to step 4.
 
 
-{{site.alert.note}} We strongly recommend you invite other members of your
-  organization as members to the verified publisher. This ensures that your
-  organization retains access to the publisher when you are not available.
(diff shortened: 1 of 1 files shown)


src/tools/pub/publishing.md
```diff
@@ -128,6 +128,11 @@ To create a verified publisher, follow these steps:
    * When the verification flow is complete, return to step 4.
 
 
+{{site.alert.note}} We strongly recommend you invite other members of your
+  organization as members to the verified publisher. This ensures that your
+  organization retains access to the publisher when you are not available.
+  {{site.alert.end}}
```
parlough, line 134:
```suggestion
{{site.alert.tip}}
  We strongly recommend you invite other members of your
  organization to be members of the verified publisher.
  This helps ensure that your organization retains access to
  the publisher when you are not available.
{{site.alert.end}}
```

parlough (approved):
Good idea to add this note. Thanks!

Some minor formatting fixes:

PR #5471, opened 2024-01-17
Add UTF-8 note

### LINKED ISSUE #5453: Document that SDK tooling expects Dart files to be UTF-8 encoded
### Page URL

https://dart.dev/tools/sdk.html

### Page source

https://github.com/dart-lang/site-www/tree/main/src/tools/sdk.md

### Describe the problem

We don't clearly document anywhere what encoding the tools support, resulting in potential confusion by developers (https://github.com/dart-lang/sdk/issues/54523).

While it seems to have been decided to not document this at the specification level (https://github.com/dart-lang/language/issues/2186), we likely should document it at the SDK or tooling level.

### Expected fix

Document somewhere, maybe on the SDK overview or the `dart` tool page that Dart files are expected to be UTF-8 encoded.

atsansone:
Fixes #5453

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -38,6 +38,11 @@ pub-pkg: https://pub.dev/packages
 dartpad: https://dartpad.dev
 news: https://news.dartlang.org
 announce: https://groups.google.com/a/dartlang.org/g/announce
+gh: https://github.com
+gh-dart: 
+  org: https://github.com/dart-lang
+  sdk: https://github.com/dart-lang/sdk
+  site: https://github.com/dart-lang/website
 
 show_banner: true
 
--- a/src/tools/sdk.md
+++ b/src/tools/sdk.md
@@ -4,12 +4,13 @@ description: Dart libraries and command-line tools.
 ---
 
 The Dart SDK has the libraries and command-line tools that you need to develop
-Dart web, command-line, and server apps. To get the Dart SDK, see [Get Dart](/get-dart).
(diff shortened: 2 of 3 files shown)


_config.yml
```diff
@@ -38,6 +38,10 @@ pub-pkg: https://pub.dev/packages
 dartpad: https://dartpad.dev
 news: https://news.dartlang.org
 announce: https://groups.google.com/a/dartlang.org/g/announce
+gh: https://github.com
+  dart: https://github.com/dart-lang
```
parlough, line 42: `dart` can't have an inline value and be a map. Consider something like this instead:

```suggestion
  dart:
    org: https://github.com/dart-lang
```

src/tools/sdk.md
```diff
@@ -21,30 +22,34 @@ the [Dart tools](/tools) page.
 {% comment %}
   IMPORTANT: After each release, EDIT src/_data/pkg-vers.json
   to update the SDK version number.
-  More info: https://github.com/dart-lang/site-www/wiki/Updating-to-new-SDK-releases
+  More info: https://github.com/site-www/wiki/Updating-to-new-SDK-releases
```
parlough, line 25:
```suggestion
  More info: https://github.com/dart-lang/site-www/wiki/Updating-to-new-SDK-releases
```

parlough (approved):
Thanks! Looks good to me :)

PR #5479, opened 2024-01-22
Revise Dart FAQ

### LINKED ISSUE #4673: Evaluate and/or rewrite the FAQ
### Page URL

https://dart.dev/resources/faq

### Page source

https://github.com/dart-lang/site-www/tree/main/src/faq.md

### Describe the problem

Most of the FAQ entries were written a long time ago. While many still be correct, a lot of them may be outdated or irrelevant, or import questions may be missing. Most of them are focused on the web, which has overall modified focus and web interop and web features are evolving heavily in the near feature. Our writing style has evolved since then as well.

### Expected fix

Evaluate each FAQ entry for its:

1. Usefulness
2. Correctness
3. Writing

Then either remove or update it accordingly.

Also consider other frequent questions that may be useful to add.

### Additional context

_No response_

atsansone:
Revised for currency.

Fixes #4673

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -42,7 +42,8 @@ gh: https://github.com
 gh-dart: 
   org: https://github.com/dart-lang
   sdk: https://github.com/dart-lang/sdk
-  site: https://github.com/dart-lang/website
+  lang: https://github.com/dart-lang/language
+  site: https://github.com/dart-lang/site-www
 
 show_banner: true
 
--- a/firebase.json
+++ b/firebase.json
@@ -232,6 +232,7 @@
       { "source": "/pub/tools/editor", "destination": "/tools", "type": 301 },
       { "source": "/redirects/sdk-download-*", "destination": "/get-dart", "type": 301 },
       { "source": "/resources/dart-tips*{,/**}", "destination": "/effective-dart", "type": 301 },
+      { "source": "/resources/faq", "destination": "/language", "type": 301 },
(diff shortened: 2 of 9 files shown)


MaryaBelanger:
Can you update the PR description to explain whether you checked the page for any info that might be beneficial to preserve since it doesn't exist anywhere else on the site. I can't go through all of them but at a glance some of these warrant at least a single sentence added to their relevant corresponding pages, e.g. [Why are generics covariant?](https://dart.dev/resources/faq#q-why-are-generics-covariant) probably deserves a mention on the generics page, something about [AOT vs JIT tradeoffs](https://dart.dev/resources/faq#q-which-is-faster-aot--or-jit-compiled-code) on the `dart compile` page, etc.

atsansone:
@MaryaBelanger : I did in part. Maybe removal isn't the right move here. I'll take another look.

parlough:
I think until someone has a chance to collect common questions and determine their answers, I think it's fine to delete this. It's one of the least visited pages.

It's just someone needs to go through each answer to determine if it is still relevant and if so, should it be documented elsewhere? Some cases, such as in the browser version support change you already made, already should have been. Other answers, even if they're still relevant, might already be documented elsewhere or not important enough to document.

Happy to discuss different ones if needed :)

atsansone:
@parlough @MaryaBelanger : I did discuss with @munificent and his views matched @parlough . I moved some information, but didn't do all of it yet. @MaryaBelanger : Would you prefer to have a full accounting first or can I remove and we can revisit after I/O?

MaryaBelanger (approved):
@atsansone please create the issue to revisit

PR #5481, opened 2024-01-23
Add section on `external`

### LINKED ISSUE #5478: Document `external` on the Functions page
The `external` keyword isn't explained anywhere on the site. Right now, references of it point to the spec. 

With [JS interop](https://github.com/dart-lang/site-www/issues/5438) and [extension type](https://github.com/dart-lang/site-www/issues/4177) docs developing, I think it's important to have a high-level summary of `external` to point to when we're talking about it. 

I think the section should go on the [Functions](https://dart.dev/language/functions) page, and be relatively high-level (description in the [spec](https://spec.dart.dev/DartLangSpecDraft.pdf#External%20Functions) is good). Since it's highly platform dependent, we can just summarize it and point to interop docs (maybe use [C](https://dart.dev/interop/c-interop) and [JS](https://dart.dev/interop/js-interop) specifically as an example, until we have a [interop overview](https://github.com/dart-lang/site-www/issues/5439) page to point to instead).

MaryaBelanger:
Fixes #5478

### DIFF
--- a/src/language/functions.md
+++ b/src/language/functions.md
@@ -502,8 +502,7 @@ Iterable<int> naturalsDownFrom(int n) sync* {
 
 ## External functions {#external}
 
-Declare a function `external` if its body is defined separately from its
-declaration in your Dart program. 
+An external function is a function whose body is provided separately from its declaration.
 
 ```dart
 external void someFunc(int i);
--- a/src/language/functions.md
+++ b/src/language/functions.md
@@ -520,8 +520,6 @@ and to make them usable in Dart. Implementation and usage is
 heavily platform specific, so check out the interop docs on, for example,
 [C][] or [JavaScript][] to learn more.
 
-
-
 [`Iterable`]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-core/Iterable-class.html
 [`Stream`]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-async/Stream-class.html
 [record]: /language/records#multiple-returns
--- a/src/language/functions.md
+++ b/src/language/functions.md
@@ -502,23 +502,25 @@ Iterable<int> naturalsDownFrom(int n) sync* {
 
 ## External functions {#external}
 
-An external function is a function whose body is provided separately from its declaration.
+An external function is a function whose body is implemented separately from its
+declaration. Include the `external` keyword before a function declaration, like so:
 
 ```dart
 external void someFunc(int i);
 ```
+
+An external function's implementation can come from another Dart library,
+or, more commonly, from another language. In interop contexts, `external`
+introduces type information for foreign functions or values,
+making them usable in Dart. Implementation and usage is
+heavily platform specific, so check out the interop docs on, for example,
+[C][] or [JavaScript][] to learn more.
 
 External functions can be top-level functions, [instance methods][],
 [getters or setters][], or [non-redirecting constructors][].
 An [instance variable][] can be `external` too,
 which is equivalent to an external getter and (if the variable
 is not `final`) an external setter.
-
-The `external` keyword usually appears in interop contexts,
-to introduce type information for foreign functions or values,
-and to make them usable in Dart. Implementation and usage is
-heavily platform specific, so check out the interop docs on, for example,
-[C][] or [JavaScript][] to learn more.
 
 [`Iterable`]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-core/Iterable-class.html
 [`Stream`]: {{site.dart-api}}/{{site.data.pkg-vers.SDK.channel}}/dart-async/Stream-class.html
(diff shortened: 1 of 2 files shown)


src/language/functions.md
```diff
@@ -500,6 +500,26 @@ Iterable<int> naturalsDownFrom(int n) sync* {
...
+External functions can be top-level functions, [instance methods][],
+[getters or setters][], or [non-redirecting constructors][].
+[Instance variables][] can be `external` too,
+and essentially equate to a getter or setter on the external value. 
+
+The `external` keyword usually appears in interop contexts,
+to introduce type information for foreign functions or values,
+and to make them usable in Dart. Implementation and usage is
+heavily platform specific, so check out the interop docs on, for example,
+[C][] or [JavaScript][] to learn more.
```
parlough, line 521: This "C" in the raw markdown looks funny for some reason 😆. It's so short. Good thing Dart has 4 letters 👀

src/language/functions.md
```diff
@@ -500,6 +500,26 @@ Iterable<int> naturalsDownFrom(int n) sync* {
 }
 ```
 
+## External functions {#external}
+
+Declare a function `external` if its body is defined separately from its
+declaration in your Dart program.
```
parlough, line 506: This sentence being imperative makes me slightly concerned developers will misinterpret it as "if you make a function `external` in one file, **you** can implement it in another **Dart** file". Is that worry justified?

If so, would it make sense to instead just mention _how_ they are declared and rely on the various interop docs for actual instructions? Or move up the discussion from the last paragraph about when it appears, since that's the main reason we're documenting it? Not sure 😕 

Maybe I'm overthinking it. I'll think some more but will leave this comment for now to get your thoughts :)

parlough (commented):
Another one down _(from the external keyword table links)_ 🥳!

Thanks for working on this Marya! Reads well to me.

I do have one question that is maybe me overthinking things. Let me know what you think :)

src/language/functions.md
MaryaBelanger: > Good thing Dart has 4 letters 👀

Clearly the superior language 😆

src/language/functions.md
MaryaBelanger: No that's a really good point (in fact, it sounds like exactly the kind of question I'd leave in a review of someone else's work haha).

I do want to leave the last paragraph where it is (though your reasoning makes sense, I think since this is purely a reference section the first sentence should be a generic definition), but I think I'll restate the opening more or less as it's stated in the [spec](https://spec.dart.dev/DartLangSpecDraft.pdf#External%20Functions):

> An external function is a function whose body is provided separately from its declaration.

src/language/functions.md
```diff
@@ -500,6 +500,25 @@ Iterable<int> naturalsDownFrom(int n) sync* {
 }
 ```
 
+## External functions {#external}
+
+An external function is a function whose body is provided separately from its declaration.
```
parlough, line 505: Starting with this does seem safer, thanks! However, do consider briefly introducing the following code snippet/syntax, otherwise it's not clear why it's there/what it means.

parlough (approved):
Looks great to me after one more consideration. Thanks for documenting this Marya!

Do consider waiting for Erik's review before landing, unless you need this landed for something else. We can always make changes as the various uses of the keyword start to be documented too :D

src/language/functions.md
```diff
@@ -500,6 +500,25 @@ Iterable<int> naturalsDownFrom(int n) sync* {
...
+
+```dart
+external void someFunc(int i);
+```
+
+External functions can be top-level functions, [instance methods][],
+[getters or setters][], or [non-redirecting constructors][].
+[Instance variables][] can be `external` too,
+and essentially equate to a getter or setter on the external value.
```
eernstg, line 514: We don't even have to say 'essentially' --- a final external variable is just a different syntax specifying an external getter, and a non-final external variable is just a different syntax specifying an external getter and an external setter. Also, we can't know whether there is anything out there that matches the description 'external value', it's perfectly possible that such getters/setters are bound to C functions, with no direct connection to anything that resembles a variable. Perhaps the text could be adjusted a bit like this:

```suggestion
An [instance variable][] can be `external` too,
which is equivalent to an external getter and (if the variable
is not `final`) an external setter.
```

It is also possible for a static variable to be external, and a library variable (aka a top-level variable), but not a local variable (and not a function parameter). I don't know if it's important to mention the static/library variants, that would depend on the level of detail/completeness which is customary and desired here.

src/language/functions.md
```diff
@@ -511,3 +530,10 @@ Iterable<int> naturalsDownFrom(int n) sync* {
 [conditional expression]: /language/operators#conditional-expressions
 [Flutter]: {{site.flutter}}
 [trailing commas]: /language/collections#lists
+
+[instance methods]: /language/methods#instance-methods
+[getters or setters]: /language/methods#getters-and-setters
+[non-redirecting constructors]: /language/constructors#redirecting-constructors
+[Instance variables]: /language/classes#instance-variables
```
eernstg, line 537:
```suggestion
[instance variable]: /language/classes#instance-variables
```

eernstg (approved):
LGTM!

Issue #5458, opened 2024-01-12
Security advisory documentation


szakarias:
### What information needs to be added?

Add documentation about security advisories in pub

### I would like to fix this problem.

- [x] I will try and fix this problem on dart.dev.

PR #5480, opened 2024-01-23
Add documentation about security advisories

szakarias:

PR #5485, opened 2024-01-24
Minor fixes and cleanup for security advisories page

parlough:
Follow-up to https://github.com/dart-lang/site-www/pull/5480

Fixes https://github.com/dart-lang/site-www/issues/5458

A few things:

- Adjusts the placeholder redirect to point to the new page
- Add a cross-link from the pubspec reference page
- Fixes a few typos
- Use descriptive link text instead of "Here"
- Remove `/en` from the GitHub link, so it's more likely redirect to the desired language of the user
- In the "Ignoring security advisories" section, if putting the pubspec field `ignored_advisories` in code font, it's not really "a list of `ignored_advisories`, but rather a list itself.
- Uses proper capitalization for "GitHub"
- Use "the Dart team" instead of "we" for the recommendation
- Some other minor wording adjustments _(please take a look :])_


parlough:
@MaryaBelanger Thanks for reviewing and landing the original PR!!

I didn't take a look until after it landed, but found a few minor issues and areas of improvements.

Consider using as a reference for some things I look for when reviewing :)

MaryaBelanger (approved):
Thanks for catching all that!

PR #5487, opened 2024-01-25
Add reference documentation for 'ignored_advisories' in the pubspec

szakarias:


src/tools/pub/pubspec.md
```diff
@@ -449,6 +449,24 @@ Pub.dev requires topics to follow these specifications:
 When choosing topics, consider if [existing topics]({{site.pub}}/topics)
 are relevant. Tagging with existing topics helps users discover your package.
 
+### Ignored_advisories
+
+If a package has a dependency which is affected by a security advisory, the pub
+client will warn about the advisory at dependency resolution. Package authors
+can use the 'ignored_advisories' property to allowlist an advisory that is not
+relevant for the package. To suppress the warning about an advisory, add the
+advisory identifier to the 'ignored_advisories' list. For example:
```
parlough, line 458: A few minor suggestions here:

- The writers alway tells me to avoid "will" if possible.
- I recommend putting `ignored_advisories` in code font.
- The Google style guide says we shouldn't use "allowlist" as a verb (https://developers.google.com/style/word-list#allowlist), but as a noun is ok.
- To be consistent with elsewhere on the page, consider calling it a "field" rather than "property".
- Since the paragraph gets a little long, maybe splitting off the instruction portion would help readability?

```suggestion
If a package has a dependency that is affected by a security advisory,
pub warns about the advisory during dependency resolution.
Package authors can use the `ignored_advisories` field as an allowlist
of triggered advisories that are not relevant for the package.

To suppress the warning about an advisory,
add the advisory identifier to the `ignored_advisories` list.
For example:
```

src/tools/pub/pubspec.md
```diff
@@ -449,6 +449,24 @@ Pub.dev requires topics to follow these specifications:
...
+
+```yaml
+name: myapp
+dependencies:
+  foo: ^1.0.0
+ignored_advisories:
+ - GHSA-4rgh-jx4f-qfcq
+```
+
+For more information, see [Security advisories](/tools/pub/security-advisories).
```
parlough, line 468: We've been mostly standardizing to "check out" rather than "see":

```suggestion
For more information, check out
[Security advisories](/tools/pub/security-advisories).
```

parlough (approved):
Thanks for expanding on this @szakarias! I appreciate it =D

Looks good to me, but some small suggestions to consider:

szakarias:
Thanks @parlough for your suggestions!

PR #5489, opened 2024-01-26
Updated deferred language

### LINKED ISSUE #5100: Clarify library documentation around support for deferred loading in AOT mode
https://dart.dev/language/libraries states that deferred loading is not supported in the Flutter and Dart VM.
This is not entirely true, we do support deferred loading in AOT mode, this text needs to be updated to reflect
the current state.

Please see https://github.com/dart-lang/sdk/issues/33118 which was closed as implemented for AOT

atsansone:
Fixes #5100

### DIFF
--- a/src/content/language/libraries.md
+++ b/src/content/language/libraries.md
@@ -89,29 +89,24 @@ import 'package:lib1/lib1.dart' show foo;
 import 'package:lib2/lib2.dart' hide foo;
 ```
 
-<a id="deferred-loading"></a>
-#### Lazily loading a library
+#### Lazily loading a library {:#deferred-loading}
 
-_Deferred loading_ (also called _lazy loading_)
+*Deferred loading* (also called *lazy loading*)
 allows a web app to load a library on demand,
 if and when the library is needed.
-Here are some cases when you might use deferred loading:
+Use deferred loading when you want to meet one or more of the following needs.
 
-* To reduce a web app's initial startup time.
-* To perform A/B testing—trying out
+* Reduce an Android or web app's initial startup time.
+* Perform A/B testing—trying out
   alternative implementations of an algorithm, for example.
-* To load rarely used functionality, such as optional screens and dialogs.
+* Load rarely used functionality, such as optional screens and dialogs.
 
-:::warning
-**Only `dart compile js` supports deferred loading.**
-Flutter and the Dart VM don't support deferred loading.
-To learn more, see
-[issue #33118](https://github.com/dart-lang/sdk/issues/33118) and
-[issue #27776.](https://github.com/dart-lang/sdk/issues/27776)
-:::
+The Android and JavaScript deferred loading differ.
+The Android app bundles all deferred components into the APK at compile time.
+That doesn't mean it loads them all at start time.
+The web app can download deferred components via the web when needed.
 
-To lazily load a library, you must first
-import it using `deferred as`.
+To lazily load a library, first import it using `deferred as`.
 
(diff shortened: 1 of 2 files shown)


src/language/libraries.md
```diff
@@ -92,34 +92,25 @@ import 'package:lib2/lib2.dart' hide foo;
...
-  **Only `dart compile js` supports deferred loading.**
-  Flutter and the Dart VM don't support deferred loading.
-  To learn more, see
-  [issue #33118](https://github.com/dart-lang/sdk/issues/33118) and
-  [issue #27776.](https://github.com/dart-lang/sdk/issues/27776)
-{{site.alert.end}}
-
-To lazily load a library, you must first
-import it using `deferred as`.
+To lazily load a library, import it using `deferred as`.
```
mkustermann, line 105: AFAIK the flutter-AOT deferred loading solution will still bundle the deferred components in e.g. the Android APK. So they will not lazily be downloaded from network.

This is quite different from JS, where the actual JS files will be downloaded when needed.

It may be worth pointing out this aspect.

mkustermann (approved):

src/content/language/libraries.md
```diff
@@ -89,29 +89,24 @@ import 'package:lib1/lib1.dart' show foo;
 import 'package:lib2/lib2.dart' hide foo;
 ```
 
-<a id="deferred-loading"></a>
```
parlough, line 92: Can you keep this extra anchor fragment target? It's been used for a while and there are a few links from other sites to it.

src/content/language/libraries.md
```diff
@@ -89,29 +89,24 @@ import 'package:lib1/lib1.dart' show foo;
...
-#### Lazily loading a library
+#### Lazily loading a library {:#lazily-loading-a-library}
```
parlough, line 92: I wouldn't expect the header ID customization `{:#lazily-loading-a-library}` is needed in this case since it matches the original header. Any reason you needed to add it?

src/content/language/libraries.md
```diff
@@ -89,29 +89,24 @@ import 'package:lib1/lib1.dart' show foo;
...
-**Only `dart compile js` supports deferred loading.**
-Flutter and the Dart VM don't support deferred loading.
-To learn more, see
-[issue #33118](https://github.com/dart-lang/sdk/issues/33118) and
-[issue #27776.](https://github.com/dart-lang/sdk/issues/27776)
-:::
+* Load rarely used functionality, such as optional screens and dialogs.
+
+The Android and JavaScript deferred loading differ.
+The Android app bundles all deferred components into the APK at compile time.
```
parlough, line 105: My understanding is that this support for deferred loading on Android functionality is specific to Flutter, and other Dart users can't take advantage it without further customizations to the VM or their embedder.

Due to this and the fact we don't generally document Flutter special handling on the site, I'd instead recommend not mentioning Android in this discussion, and linking to the Flutter site.

Maybe something roughly like:

```
The `dart` tool doesn't support deferred loading for its other compilation targets.
If you're building a Flutter app, to learn about the platform it supports for deferred loading,
check out the flutter.dev docs on
[deferred components](https://docs.flutter.dev/perf/deferred-components).
```

That also prevents us from having to make further updates if Flutter changes that functionality or implements support for other target platforms.

parlough (commented):
Thanks for looking into this update Tony! However, since this is currently Flutter specific, it's best to just point to the documentation there.

To account for the VM technically having some of the underlying support, we can shift the discussion of lack of support from the "Dart VM" to the `dart` tool.

PR #5501, opened 2024-02-02
Add recommendation for several package uploaders.

szakarias:
In the same spirit as https://github.com/dart-lang/site-www/pull/5467


parlough (approved):
Thanks! Good idea to add it here too :)

PR #5503, opened 2024-02-03
Update breaking changes page for Dart 3.3

parlough:
- Added various deprecations, breaking changes, and potentially disruptive changes to 3.3 changes
- Begin filling out the 3.4 section as well
- Since neither version is released yet, added disclaimers in their intros.
- Modified the headers to just the version number since the "Released in" felt unnecessary **(What do you think about this?)**

_A few of the significant entries are missing from the upstream changelog. I'll work on getting those added._

Resolves https://github.com/dart-lang/site-www/issues/5505

**Staged:** https://dart-dev--pr5503-feat-3-3-breaking-changes-jqc29fvc.web.app/resources/breaking-changes


src/resources/breaking-changes.md
```diff
@@ -45,46 +45,241 @@ To be notified about future breaking changes, join the [Dart announce][] group.
...
+#### Production JavaScript compiler (dart2js)
+
+* You should now specify a format to the `--dump-info` flag
+  of either `binary` or `json`.
+  The `json` format is deprecated and might be removed in a future Dart release.
+
+#### Wasm compiler (dart2wasm)
+
+* **Experimental**: Various `dart compile wasm` CLI arguments have
+  been updated, removed, or replace.
```
MaryaBelanger, line 139:
```suggestion
  been updated, removed, or replaced.
```

src/resources/breaking-changes.md
```diff
@@ -45,46 +45,241 @@ To be notified about future breaking changes, join the [Dart announce][] group.
...
+- This library is now marked as legacy and will see less support in the future.
+  Usages should be migrated to `dart:js_interop` and `dart:js_interop_unsafe`.
+  To learn more, check out [`/go/next-gen-js-interop`][].
+
+[`/go/next-gen-js-interop`]: https://dart.dev/go/next-gen-js-interop
 
 #### `dart:typed_data`
 
-* **Deprecated**: [Unmodifiable view classes for typed data][53218].
+* [The unmodifiable view classes for typed data are deprecated][53128].
```
MaryaBelanger, line 121:
```suggestion
* **Deprecated**: [The unmodifiable view classes for typed data are deprecated][53128].
```

src/resources/breaking-changes.md
```diff
@@ -45,46 +45,241 @@ To be notified about future breaking changes, join the [Dart announce][] group.
...
+### Tools
+{: .no_toc}
 
-{% comment %}
-Create new section from these headers for each release.
-If no changes exist in a section (e.g. Language, `dart:async`, etc.),
-don't include the section header.
+#### Production JavaScript compiler (dart2js)
+
+* You should now specify a format to the `--dump-info` flag
```
MaryaBelanger, line 132:
```suggestion
* **Deprecated**: You should now specify a format to the `--dump-info` flag
```

src/resources/breaking-changes.md
```diff
@@ -45,46 +45,241 @@ To be notified about future breaking changes, join the [Dart announce][] group.
...
+The following changes are expected to be included in the 3.3.0 stable release,
+but the final list might change before then.
+To reduce any potential impact of these changes, consider
+accounting for them before the 3.3.0 release.
+
+### SDK
+{: .no_toc}
 
-## Released in x.x.x
+* The following experiments are now retired as they were released in Dart 3.
```
MaryaBelanger, line 153: Would this count as an Experimental / Removed?
```suggestion
* **Experimental** **Removed**: The following experiments are now retired as they were released in Dart 3.
```

src/resources/breaking-changes.md
```diff
@@ -45,46 +45,241 @@ To be notified about future breaking changes, join the [Dart announce][] group.
...
+* Due to [improvements in type promotion][54056], the following analyzer
+  diagnostics might trigger on existing code that previously passed analysis:
+
+  * `unnecessary_non_null_assertion`
+  * `unnecessary_cast`
+  * `invalid_null_aware_operator`
+
+[54056]: https://github.com/dart-lang/sdk/issues/54056
+
+#### Linter
```
MaryaBelanger, line 266: Do these two count as "removed" also?

MaryaBelanger (approved):
Thanks for doing this! Getting rid of "released in..." makes sense to me. I'm wondering about a few possible "removed"s in the new version entries (left comments). Are we only labeling things "Removed" if its already had a corresponding past "Deprecated" entry? Or do some of things I flagged not technically count as our definition of removals (not in the SDK or something)? 

Edit: all my comments besides the one typo are more like questions ("Should we add this tag?")

src/resources/breaking-changes.md
parlough: I was on the fence about this one as it has been deprecated for many years now and it's not clear we want to deprecate it still. Makes sense to add this though, thanks!

I'll likely also add experimental.

src/resources/breaking-changes.md
parlough: Ooops!

src/resources/breaking-changes.md
parlough: Hmm. I guess so? For now, works for me. Definitely makes sense to add Experimental though!

Once I get around to working on the labels we can think more precisely about the categories and how they help users.

src/resources/breaking-changes.md
parlough: I'm not sure to be honest. I think a bold "Removed" is probably not necessary for these as they don't affect code.

It seemed like **Removed** was more orientated towards language features or APIs that are removed, so users have to change their Dart code.

For now, I don't think these need a lot of attention, so I'll lean towards not adding it.

Let's discuss this page and its categories after things have calmed down a bit :)

parlough:
Thanks for the review and suggestions @MaryaBelanger!

I'm going to land this as it can already help Dart developers know what changes they might need to make.

Please feel free to make any follow-up adjustments you see fit :D

PR #5483, opened 2024-01-24
Initial migration of site to 11ty

parlough:
I'm sorry that this ended up as such a large pull request, but that's mostly because it essentially touched every file with a small set of changes. See the different sections below for considerations for review or future reference. Please note that we'd like this to land despite it not being perfect, as then pain points can be identified and improved. Then I can also follow up with proper documentation as the implementation and syntax is stabilized. I'll be staying on top of any reported issue. If anything goes majorly wrong, we can always revert the change as well. Thanks for your time 💙

Resolves https://github.com/dart-lang/site-www/issues/5177
Resolves https://github.com/dart-lang/site-www/issues/3846
Resolves https://github.com/dart-lang/site-www/issues/5323
Fixes https://github.com/dart-lang/site-www/issues/4297
Fixes https://github.com/dart-lang/site-www/issues/4919
Closes https://github.com/dart-lang/site-www/issues/4631
Contributes to https://github.com/dart-lang/site-www/issues/4163

> [!TIP]
> Reviewing this in GitHub's UI might be slow. Consider cloning the branch locally.

**Changes potentially worth reviewing from a technical perspective are:**

- Implemented 11ty in `eleventy.config.js` with customizations in `/src/_11ty`.
  - `eleventy.config.js` configures the site build, enables used plugins, and connects the custom logic added in `/src/_11ty`. This will be greatly simplified and better documented in follow-up work.
- Replaced prettify, rouge, and custom code excerpter logic with custom markdown plugin.
  - You can find this at `/src/_11ty/plugins/highlight.js` _(super messy for now)_.
  - New custom theme is at `/src/_11ty/syntax/dash-light.js`.
  - The implementation of this needs a lot of cleanup and will be simplified greatly in follow-up work to rely on recent improvements in underlying highlighting package.
- Added a `build` and `serve` command to `dart run dart_site` (found in `/tool/dart_site/lib`).
- The deploy configs in `/cloud_build` have been migrated from using Docker to use Node to build the site. I reused the Docker image we built for firebase_tools, as that already has Node. Let me know if I shouldn't though :)
- Removed remnants of previous infra (Jekyll files, gem files, docker files, makefile, etc).
- All written docs and assets have been moved into `/content` to enable a separation from site tooling and configuration.

**Changes potentially worth reviewing or knowing about from a writer/contributor perspective:**

- **Major:** All doc content has been moved into `/content`. This allows site implementation to be separate from content, potentially enabling easier downstream localization as well.
[...]

PR #5510, opened 2024-02-05
Fix ARIA errors

### LINKED ISSUE #5509: Fix ARIA errors
Fix errors involving ARIA attributes.

- **Error/Serious:** [ARIA hidden element must not be focusable or contain focusable elements](https://dequeuniversity.com/rules/axe/4.8/aria-hidden-focus?application=axeAPI)
  Ensures aria-hidden elements are not focusable nor contain focusable elements

- **Error/Critical:** [Elements must only use supported ARIA attributes](https://dequeuniversity.com/rules/axe/4.8/aria-allowed-attr?application=axeAPI)
  Ensures an element's role supports its ARIA attributes

atsansone:
Fixes #5509

Updated the script for 11ty.
image

PTAL @parlough .


src/assets/js/main.js
```diff
@@ -124,6 +124,19 @@ function adjustToc() {
   $(document.body).scrollspy({ offset: 100, target: '#site-toc--side' });
 }
 
+function adjustHeadingAnchors() {
+  document.addEventListener('DOMContentLoaded', function() {
+    const headingAnchors = document.querySelectorAll('a.anchor[aria-hidden]');
```
parlough, line 129: Thanks for thinking about this Tony! However, I'd prefer this specific change not land if possible as these anchors won't exist once https://github.com/dart-lang/site-www/pull/5483 lands, which should be sometime in the next few days.

If you can share more details about what the adjustment is intended to do, I can make sure it works with the new setup!

atsansone: @parlough : Can do. The issue is that the a11y check fails when it encounters the anchor links that appear next to each heading. This change removes those heading links from the tabindex, which allows the test to pass.

atsansone:
Superceded by #5537

Issue #4278, opened 2022-10-12
Standardize all icon usage to Material Symbols


parlough:
### Describe the problem

Material Symbols is a variable font with expanded and updated icons. dartdoc has switched to using it and we should as well.

### Expected fix

Migrate all Material Icons usages to Material Symbols.

Font Awesome usage should also be migrated: https://github.com/dart-lang/site-www/issues/3790

### Additional context

Material symbols: https://fonts.google.com/icons
Material symbols guide: https://developers.google.com/fonts/docs/material_symbols

PR #5518, opened 2024-02-08
Standardize icons on Material Symbols

parlough:
Corresponds to https://github.com/flutter/website/pull/10105

Fixes https://github.com/dart-lang/site-www/issues/4278


atsansone (approved):
LGTM, but check any inline symbols later. They may need to offset by 1 px or so, IME.

PR #4578, opened 2023-02-06
New JS interop docs

MaryaBelanger:
Fixes #5438 

WIP (lots of notes interspersed on the pages)

Adds new pages and content to cover static, inline-class based, JS interop.

**Staged:** https://dart-dev--pr4578-js-interop-jct9fv7o.web.app/interop/js-interop

New pages:
- JavaScript interoperability (Overview) (web/js-interop)
- JS Interop Reference (web/js-interop/reference)
- How to interop with DOM APIs (web/js-interop/dom)
- How to interop with JavaScript libraries and apps (web/js-interop/js-app)
- How to test and mock JavaScript interop in Dart (web/js-interop/test-and-mock)
- Non-static JS interop (web/js-interop/past-js-interop)
- Migration (web/js-interop/migration)
- FAQ / Troubleshooting (web/js-interop/faq-troubleshooting)

PR #5524, opened 2024-02-09
`package:web` migration page

MaryaBelanger:
This is a very high level migration page that mostly points to JS interop pages and the `web` repo for more detail. 

Part of #4578


src/interop/js-interop/package-web.md
```diff
@@ -0,0 +1,195 @@
...
+}
+```
+
+`package:web` is generated automatically from [Web IDL][@webref/idl] declarations,
+and implemented exclusively using [static interop][].
+Packages will only be compatible with [Wasm][] if they use static interop,
+which is why `package:web` is replacing Dart's previous web interop solution,
+[`dart:html`][html] (and similar libraries like `dart:svg` and `dart:web_gl`).
+
+:::important
```
MaryaBelanger, line 26: This doesn't render yet because the base branch isn't updated for 11ty

src/interop/js-interop/package-web.md
```diff
@@ -0,0 +1,195 @@
...
+
+To migrate unsupported method calls from `dart:html` to `package:web`,
+you'll mostly want to use conversion methods like `toJS()` and `toDart()`.
+Learn how to use static interop conversion methods from the [Conversions][]
+section of the JS types page.
+
+```dart
+_clickSubscription = window.onClick.listen(callback); // Remove
+window.onclick = callback.toJS; // Add
+```
```
MaryaBelanger, line 114: This example is from a doc written before the helper methods became a thing. I'm guessing technically there's a helper to address this kind of thing now, so maybe a bad example?

srujzs: There is, although the helpers are on `Element`, but yeah, I think the example is probably not what we want. We can put a TODO here to look at some of the migrations and look for a better example.

src/interop/js-interop/package-web.md
```diff
@@ -0,0 +1,195 @@
...
+
+Generally, you can spot which methods need to migrate because they'll be flagged 
+with some variation of the exception:
+
+```nocode
+A value of type '...' can't be assigned to a variable of type 'JSFunction?'
+```
+
+Also important to note that methods are [statically dispatched][static interop]
```
srujzs, line 123: I wonder if the extension types docs has a section on this that we can link to here.

src/interop/js-interop/package-web.md
```diff
@@ -0,0 +1,195 @@
...
+[repo][helpers].
+
+### Zones
+
+In `dart:html`, calls are automatically zoned.
+This is not the case in `package:web`. There is no
+automatic binding of async callbacks in the current zone.
+
+To migrate, you can still to use zones,
+but you will have to [write them yourself][zones].
```
srujzs, line 138: TODO on me to come up with an example.

src/interop/js-interop/package-web.md
```diff
@@ -0,0 +1,195 @@
...
+There is no [helper](#helper-layer) available yet.
+
+## Helper layer
+
+The core of `package:web` provides direct interop with external
```
srujzs, line 143: Alternate wording: "The core of `package:web` contains `external` members to directly interop with the web, but does not ..."

src/interop/js-interop/package-web.md
```diff
@@ -0,0 +1,195 @@
...
+members from the IDL, but does not provide solutions for some other
+functionality that `dart:html` handled by default.
+To mitigate these differences,
+`package:web` provides a [`helpers`][helpers] library for additional support
+in handling a number of use cases that aren't directly
+available through the core interop.
+The helper library contains classes, typedefs, and extension methods
```
srujzs, line 150: I think we're actually planning on removing this now that we have the dart fix. :)

I'd maybe reword this sentence as "The helper library contains various types and members to expose some legacy features that exist within the Dart web libraries."

srujzs (commented):
Thanks! I have an initial pass with some comments.

src/interop/js-interop/package-web.md
```diff
@@ -0,0 +1,195 @@
...
+```dart
+import 'package:web/web.dart';
+
+void main() {
+  final div = document.querySelector('div')!;
+  div.text = 'Text set at ${DateTime.now()}';
+}
+```
+
+`package:web` is generated automatically from [Web IDL][@webref/idl] declarations,
```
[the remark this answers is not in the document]
MaryaBelanger: > I think we're shying away from "static interop" going forward to reduce confusion.

Interesting, like, in all contexts or just this one for some reason? Why is that? 
I think I've been saying it a lot so I'll have to check all my recent work!

MaryaBelanger:
@srujzs I added a rough draft of the comparison section. I'll definitely go over it again to improve the wording, but just want to get it out so you could take a look. Mostly wondering if that's the kind of content you had in mind. Thanks!

src/interop/js-interop/package-web.md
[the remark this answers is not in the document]
srujzs: It's mostly messaging really. Part of it is that we're saying users should really have one form of interop going forwards, so prefacing it with static is just confusing because for them, it's just interop. The other part is `@staticInterop` could be considered "static interop" but we don't want people using that now that extension types are out.

It's fine for us to say colloquially/internally to differentiate it from older interop.

srujzs:
Thanks for the edits Marya! I went through and added a few edits (the commit message should contain a summary) and additional sections. From my perspective, this is good to go.

cc @sigmundch for a review. I'll tackle any comments in another PR instead so this can be unblocked.

src/content/interop/js-interop/package-web.md
```diff
@@ -0,0 +1,307 @@
...
+
+It's common for code that uses `dart:html` to utilize runtime checks like `is`.
+When used with a `dart:html` object, `is` and `as` verify that the object is
+the JS type within the `@Native` annotation.
+In contrast, all `package:web` types are reified to [`JSObject`][]. This means a
+runtime type check will result in different behavior between `dart:html` and
+`package:web` types.
+
+To be able to perform type checks, migrate any `dart:html` code
+using `is` type tests to use [interop methods][] like `instanceOfString`.
```
sigmundch, line 152: To encourage the more convenient and typed API, should we illustrate `isA<T>()` instead?

src/content/interop/js-interop/package-web.md
```diff
@@ -0,0 +1,307 @@
...
+export 'src/hw_none.dart'
+    if (dart.library.io) 'src/hw_io.dart'
+    if (dart.library.js_interop) 'src/hw_web.dart';
+```
+
+### Virtual dispatch and mocking
+
+`dart:html` classes supported virtual dispatch, whereas because JS interop uses
+extension types, virtual dispatch is [not possible]. Similarly, `dynamic` calls
+with `package:web` types won't work as expected as their members are only
```
sigmundch, line 212: well. there is a caveat: if `dart:html` is included in the SDK, dart2js will continue to find the interceptor and it may continue to work by accident. Once `dart:html` is removed it will stop working.

Not sure if it's worth flagging this, since it's a very confusing concept to explain.

sigmundch (approved):
The article looks really great and reads very nicely! Minor suggestions for latter, below.

src/content/interop/js-interop/package-web.md
MaryaBelanger: I added a caveat without getting into detail:

> ...won't work as expected (or, they might continue to work
just by chance, but will stop when `dart:html` is removed), ...

src/content/interop/js-interop/package-web.md
MaryaBelanger: Added

srujzs: I think it's fine to add it here. In the JS interop docs, I mention it's not available until 3.4 and the `@Since` annotation should make that clear anyways.

src/content/interop/js-interop/package-web.md
srujzs: > Not sure if it's worth flagging this, since it's a very confusing concept to explain.

I had a comment for this and then removed it figuring it would just confuse users, but either way is fine. :)

PR #5531, opened 2024-02-10
Update set up instructions and pnpm dependencies

parlough:
Updated the README to make it clear that `corepack` is not required, it's just an easy way to install `pnpm` and match the version required by the current project.

Also update default pnpm version and various transitive dependencies, including some fixes for 11ty dev server.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,8 @@ The latest stable LTS release of Node.js is required to build the site.
 If you don't have Node.js or need to update, download your
 computer's corresponding version and follow the instructions
 from the [Node.js download archive][].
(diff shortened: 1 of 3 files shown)


atsansone (approved):
LGTM. Would it be worth having more that covers if you already have other versions of node or npm installed? We saw what happened when that's the case. It doesn't have to be in this PR, though.

parlough:
Thanks for the review!

Having other versions of node or npm isn't a problem. It's just the person needs to update the version they use. It's always a problem to have differing installs of a program. I added a section in a previous PR to explain this, including checking your current Node version first.

PR #5537, opened 2024-02-13
Fix accessibility concerns around heading anchors

parlough:
An alternate solution that introduces a wrapper and configures the generation of the anchors directly.

Fixes https://github.com/dart-lang/site-www/issues/5509


atsansone (approved):
LGTM. No error for ARIA issues.

image

Issue #4603, opened 2023-02-17
Type system > add null safety


parlough:
## Notes

Currently we have a lot of documentation covering null safety as both an integrated and an isolated concept, but it really is just a part of the type system. In the longer term (far post Dart 3) "null safety" as a term likely shouldn't be super common on the site, but instead be incorporated naturally across the entire documentation, particularly updated type and type systems sections.

Prerequisites:
- https://github.com/dart-lang/site-www/issues/4588
- https://github.com/dart-lang/site-www/issues/4584

Some steps:

- [ ] Consider a new nullable type section or subsection to new language docs
- [ ] Verify nullable and non-nullable types are covered and used as necessary across all docs on the site
- [ ] Introduce specific or better examples in cheatsheets and/or codelabs around nullable types
- [ ] Update or **rewrite** type system article to be more thorough in nullable and non-nullable types ([Understanding null safety](https://dart.dev/null-safety/understanding-null-safety) has great related content)

**Way** later steps:

- [ ] Remove all mentions of unsound null safety (`--no-sound-null-safety`)
- [ ] Remove or archive articles about enabling and configuring null safety

## Notes from Marya

Probably part of https://github.com/dart-lang/site-www/issues/4603, the [Abstract fields](https://dart.dev/null-safety/understanding-null-safety#abstract-fields) section of the Understanding null safety page seems like it is only part of the page because the feature was added with null safety. But now, that connection probably isn't too relevant any more and that content might be better suited somewhere else, like maybe in the Classes section of the Language tour?

MaryaBelanger:
Small subset of this, but there's been conversation about `?.` and getting the useful info from https://dart.dev/null-safety/understanding-null-safety#smarter-null-aware-methods into the "core docs", specifically the operators page [where `?.` is "discussed"](https://dart.dev/language/operators#:~:text=from%20expression%20foo-,%3F.,-Conditional%20member%20access) (there's like, a sentence about it at the bottom of the page).

From @fishythefish :

> Do we have a good reference page for what ?. does? I've found https://dart.dev/language/operators#other-operators and https://dart.dev/language/classes#using-class-members, but I'm wondering if there are any others. I've noticed two common misunderstandings of this operator in google3:
> 1. Some code uses ?. for every property access. I can sort of see the reasoning prior to null safety, but many of these uses flow into code that's going to blow up on null anyway. It feels like users think ?. magically avoids NPEs altogether.
[...]

PR #5539, opened 2024-02-13
Remove null safety codelab for now

parlough:
Removes the null safety codelab and references to it. In a few days, I will follow up soon with some new samples on the cheatsheet to replace this.

This codelab is the least viewed on the site and no longer makes sense to separate from other language learning. As mentioned, I will follow up with a few new samples on the cheatsheet as a stop-gap until we implement a consolidated new-learner experience.

Contributes to https://github.com/dart-lang/site-www/issues/4603
Closes https://github.com/dart-lang/site-www/issues/3093


atsansone (approved):
LGTM. Thanks!

PR #5542, opened 2024-02-14
Migrate iterables codelab away from multi-file

parlough:
- Removes usage of multi-file DartPad support (tests, hints, solutions)
- Introduces new separate **Hint** and **Solution** details-based dropdowns

**Staged:** https://dart-dev--pr5542-fix-migrate-iterables-codelab-4glpkbdh.web.app/codelabs/iterables

### DIFF
--- a/src/content/codelabs/iterables.md
+++ b/src/content/codelabs/iterables.md
@@ -353,9 +353,6 @@ void main() {
     print('Tried calling singleWhere, but received an exception: $e');
   }
 }
-
-// ignore: non_constant_identifier_names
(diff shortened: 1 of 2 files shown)


MaryaBelanger (approved):
LGTM! Aside question, what does the "Show tests" option (and the empty tab it opens) mean? Will that just not show up anymore after the new dartpad update?
image

parlough:
Yeah, that's just shown on all embeds with the old UI, so there's unfortunately no way to disable it. It won't be an option anymore with the new one :)

PR #5545, opened 2024-02-14
Add custom styles for details-based dropdowns

parlough:
Adds interaction styles and spacing to the details summary as well as spacing to the contents.

An example of how this effects on a details usage on hover at https://dart-dev--pr5545-feat-details-styling-imoer7pa.web.app/language/isolates#complete-example: 

Hovering over the summary of a details dropdown

Contributes to https://github.com/dart-lang/site-www/issues/5382 which will use `<details>` as a replacement in some cases.

### DIFF
--- a/src/_sass/core/_base.scss
+++ b/src/_sass/core/_base.scss
@@ -86,13 +86,22 @@ blockquote {
 }
 
 details {
-  summary {
+  margin-bottom: 0.75rem;
+
+  > summary {
     font-weight: 500;
     user-select: none;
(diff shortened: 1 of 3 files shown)


parlough:
@MaryaBelanger  This is to make the details-based dropdowns look a bit less cramped and have a more accessible interaction style. Especially since in the iterables codelab I added one for hint and solution.

MaryaBelanger (approved):
Nice, very subtle but makes a huge difference. Thanks!

PR #5546, opened 2024-02-14
Misc cleanup to class modifier page

parlough:
- Remove unnecessary library comments now that code blocks can have filenames
- Consistently follow site contraction standards
- Add punctuation to the end of comments to match Effective Dart
- Use semantic breaks in a few more locations
- Adjust or expand wording slightly in a few spots


MaryaBelanger (approved):
Thank you!

PR #5551, opened 2024-02-14
Banner for 3.3

MaryaBelanger:
TODO:
- Add in Medium link when ready

### DIFF
--- a/src/_includes/banner.html
+++ b/src/_includes/banner.html
@@ -1,9 +1,9 @@
 <div class="banner">
   <p class="banner__text">
-    Check out the
-    <a href="https://medium.com/dartlang/dart-3-2-c8de8fe1b91f">Dart 3.2 blog post</a>!
+    Dart 3.3 is out!
+    Get extension types, better JavaScript integration, and more.
     <br>
-    This release brings enhancements to type promotion, interop capabilities,
-    DevTools, and more.
+    Read the
+    <a href="https://medium.com/dartlang/"></a>blog post</a>.
   </p>
 </div>


src/_includes/banner.html
```diff
@@ -1,9 +1,9 @@
...
   <p class="banner__text">
-    Check out the
-    <a href="https://medium.com/dartlang/dart-3-2-c8de8fe1b91f">Dart 3.2 blog post</a>!
+    Dart 3.3 is out!
+    Get extension types, better JavaScript integration, and more.
     <br>
-    This release brings enhancements to type promotion, interop capabilities,
-    DevTools, and more.
+    Read the
+    <a href="https://medium.com/dartlang/dart-3-3-325bf2bf6c13"></a>blog post</a>.
```
parlough, line 7:
```suggestion
    <a href="https://medium.com/dartlang/dart-3-3-325bf2bf6c13">blog post</a>
```

src/_includes/banner.html
```diff
@@ -1,9 +1,9 @@
 <div class="banner">
   <p class="banner__text">
-    Check out the
-    <a href="https://medium.com/dartlang/dart-3-2-c8de8fe1b91f">Dart 3.2 blog post</a>!
+    Dart 3.3 is out!
```
parlough, line 3: Personal preference, but consider an adjustment here:

```suggestion
    Dart 3.3 is here!
```

src/_includes/banner.html
```diff
@@ -1,9 +1,9 @@
...
+    Get extension types, better JavaScript integration, and more.
```
parlough, line 4: Not necessary for this PR if you prefer this wording, but I think it's best to avoid "get" if possible. It's not clear what "getting" extension types means.

Also would something like "improved JavaScript interop" provide more details to readers than "better JavaScript integration"?

parlough (approved):
Thanks Marya! 🚀 

One fix and a few suggestions to consider :D

src/_includes/banner.html
MaryaBelanger: I agree, this was the "marketing approved" text but idk what that means and it's not technically great 😅 

Changing!

PR #5556, opened 2024-02-15
Correct return type of representation object getter

abitofevrything:
Corrects the return type of the implicit representation object getter in the extension types page.


parlough (approved):
Thanks @abitofevrything 😄 

This makes sense to me but I'll defer to @MaryaBelanger for landing it as she's still working on some updates to this page.

parlough:
Actually I'll merge this as those changes can always be adapted. Thanks again!

MaryaBelanger:
Thanks so much @abitofevrything , lgtm too!

PR #5557, opened 2024-02-15
Extension types follow up

MaryaBelanger:
@lrhn and @eernstg, I had to merge the previous pull request (#5508) for the release, but I am still concerned with some of the changes I made based on feedback from your reviews. So I opened another PR with no changes so you can provide line by line feedback if this is a good way to do it. Would you mind reviewing the major changes I made to the following sections? I'll list out the sections that need a closer look below:

- major changes to the [Implements section](https://github.com/dart-lang/site-www/pull/5557/files#diff-aaa2c0eabea037b696a026dcdb15170ffa02c5f8c988b7146348b16c9fd8a2ecR200-R248) (or view the live page [here](https://dart.dev/language/extension-types#implements))
- added part about [private constructors](https://github.com/dart-lang/site-www/pull/5557/files#diff-aaa2c0eabea037b696a026dcdb15170ffa02c5f8c988b7146348b16c9fd8a2ecR200-R248) (live page [here](https://dart.dev/language/extension-types#constructors))
- changed the [Usage](https://github.com/dart-lang/site-www/pull/5557/files#diff-aaa2c0eabea037b696a026dcdb15170ffa02c5f8c988b7146348b16c9fd8a2ecR276) section to have two subsections. Instead of just talking about "transparency" as a singular topic, reframed it as the two opposite use cases Erik described [here](https://github.com/dart-lang/site-www/pull/5508#discussion_r1487421070):
  - [transparent case](https://github.com/dart-lang/site-www/pull/5557/files#diff-aaa2c0eabea037b696a026dcdb15170ffa02c5f8c988b7146348b16c9fd8a2ecR310-R355) (live page [here](https://dart.dev/language/extension-types#1-provide-an-extended-interface-to-an-existing-type))
  - [non transparent case](https://github.com/dart-lang/site-www/pull/5557/files#diff-aaa2c0eabea037b696a026dcdb15170ffa02c5f8c988b7146348b16c9fd8a2ecR357-R397) (live page [here](https://dart.dev/language/extension-types#2-provide-a-different-interface-to-an-existing-type))

Sorry, I'm not sure how to make the file commentable.... let me know if you also can't figure it out and we can do this some other way


src/content/language/extension-types.md
```diff
@@ -443,9 +443,10 @@ void main() {
 ```
 
 It's important to be aware of this quality when using extension types,
-and never rely on them in scenarios where the representation type must be concealed. 
-The trade off for using an extension type over a more-secure real object (wrapper class)
-is their lightweight implementation, which can greatly improve performance in some scenarios.
+and never rely on them in scenarios where the representation type must be
+concealed. The trade off for using an extension type over a more-secure real
+object (wrapper class) is their lightweight implementation, which can greatly
+improve performance in some scenarios.
```
eernstg, line 449: This text tends to associate extension types with 'reliability' and 'security' (that is: lack of both) in a very vague and broad sense. I think it's more helpful for developers to be reminded about the basic facts, such that they know _how_ they need to be careful. Perhaps we could say something like this?:

```suggestion
and always keep in mind that the extension type exists and matters at
compile time, but it gets erased during compilation. If an expression has
an extension type `E` with representation type `R` as its static type then
the value at run time will be an object with type `R`, and even the type
itself is erased (`List<E>` is exactly the same thing as `List<R>` at run time).
In other words, a real wrapper class can encapsulate a wrapped object
whereas an extension type is just a compile-time view on the wrapped object.
In that sense, the real wrapper is safer; the trade-off is that the
extension type allows us to avoid the wrapper object, which can greatly
improve performance in some scenarios.
```

eernstg (approved):
LGTM.

I suggested a change which would make the text more specific. This means that we don't just leave the reader with the impression that "extension types are unreliable and unsafe", but we refresh their memory about the specific semantic facts that they need to be aware of and cautious about.

eernstg:
About github reviews of text that hasn't been changed: That's terrible, github has all kinds of behaviors to prevent that. I think it would be better to just look at the updated text at some point in time (like: "please comment on _this URL_ before _this date_"), and then we can make changes in a new PR if needed.

src/content/language/extension-types.md
MaryaBelanger: Thanks Erik, this is a much better way to wrap up the page!

PR #5558, opened 2024-02-15
3.3 language evolution

MaryaBelanger:


src/content/guides/language/evolution.md
```diff
@@ -41,6 +41,33 @@ on the Dart language GitHub repo.
...
 ## Changes in each release
 
+### Dart 3.3
+_Released 15 February 2024_
+| [Dart 3.3 announcement](https://medium.com/dartlang/dart-3-3-325bf2bf6c13)
+
+Dart 3.3 added some enchancements to the language:
+
+* [Extension types][] are a new feature in Dart that allow zero-cost wrapping
+  of existing type. They are similar to wrapper classes and to extension methods,
```
parlough, line 51: I'm not sure if this is best/most accurate the solution, but I think a word or two is missing here:

```suggestion
  of values with an existing type. They are similar to wrapper classes and to extension methods,
```

src/content/guides/language/evolution.md
```diff
@@ -41,6 +41,33 @@ on the Dart language GitHub repo.
...
+  but with implementation difference and benefits over each.
```
parlough, line 52: Would tradeoffs be more accurate?

```suggestion
  but with implementation differences and different tradeoffs.
```

If benefits is more accurate, it seems like "differences" should still be plural here.

```suggestion
  but with implementation differences and benefits over each.
```

parlough (approved):
Thanks for adding this so swiftly @MaryaBelanger!

Some considerations for you:

src/content/guides/language/evolution.md
MaryaBelanger: I know there was a lot of discussion around what exactly is being wrapped, what the value is in these situations, etc. so I'm going to stick with just "wrapping _an_ (missing word) existing type", not because you're wording is wrong, I just can't remember if saying "values" in this context was cause for concern in other reviews...

PR #5564, opened 2024-02-16
Fix a getter name in a comment from an extension types usage example

rxlabz:
This PR just fixes a typo in the NumberE usage example :  a comment refers to a 'i' getter when it should be the 'next' getter


parlough (approved):
Nice catch, thank you!

Issue #1105, opened 2018-08-27
Make it easier to copy-paste commands


kwalrath:
Feedback from a user of https://webdev.dartlang.org/guides/get-started:

> It was hard to copy-paste from the console instructions. I think some of our other pages have some fix so that the selection doesn’t include the `>` character.

/cc @mit-mit

chalin:
Yes, the plan is to make use of all of tri-site tricks once the shared infrastructure lands. Thanks for opening this issue to track that particular improvement made to the Flutter site but not yet used here (or on webdev).

parlough:
I'm going to close this in favor of https://github.com/dart-lang/site-www/issues/3846 where I plan to tackle this alongside some of the other work.

PR #5568, opened 2024-02-17
Implement code block copy button

parlough:
Fixes https://github.com/dart-lang/site-www/issues/1105

### DIFF
--- a/src/_sass/components/_code.scss
+++ b/src/_sass/components/_code.scss
@@ -58,13 +58,42 @@ pre {
   }
 }
 
+.code-copy-button {
+  position: absolute;
+  top: 6px;
+  right: 6px;
+  z-index: 10;
+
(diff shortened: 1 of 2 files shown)


atsansone (approved):
LGTM. The fact that it's hidden might not be ideal, but let's see how the functionality is received.

MaryaBelanger (approved):
Nice! Agree with @atsansone that maybe it should just show the copy icon by default, but much needed and works as it should for now!

parlough:
Thanks for the review! I'm happy to make further adjustments here if they come up.

As for my thinking for this design:

I tried to match the experience of other doc sites by making it hidden until hover, that way there's a way to view text it might cover, and not be distracted by the button when it's not necessary (most times).

I did make the button also show up if the code block or the copy button is focused.

PR #5574, opened 2024-02-20
Unmask Windows ARM64 and Linux RISCV64 stable channel downloads.

rmacnak-google:


tool/get-dart/dart_sdk_archive/lib/src/version_selector.dart
```diff
@@ -181,8 +181,9 @@ class VersionSelector {
                 versionInfo.version < Version(3, 0, 0, pre: '290.2.beta')) {
               continue;
             }
-            // No stable builds yet.
-            if (versionInfo.channel == 'stable') {
+            // Stable builds start at 3.2.3.
+            if (versionInfo.channel == 'stable' &&
+                versionInfo.version < Version(3, 2, 3)) {
```
parlough, line 186: Even if the builds existed at this point, would it instead make sense to only surface them in 3.3.0 and later? That way there's a more clear reference point on when they were declared stable. Generally anyone wanting to try on these platforms should use Dart 3.3 or later anyway.

I don't feel strongly either way. But curious to see what you both think?

parlough (commented):
Thanks Ryan! It's exciting to see these two targets able to be marked as stable.

I have one question for consideration, but otherwise I'll leave approval to mit.

mit-mit:
SGTM

tool/get-dart/dart_sdk_archive/lib/src/version_selector.dart
rmacnak-google: Updated to show from 3.3.

parlough (approved):
Thanks @rmacnak-google for these updates, including to the table. Looks good to me!

Of course thank you for the great work implementing support for these targets as well :)

PR #5578, opened 2024-02-20
Link to `ignored_advisories` from security advisories page

MaryaBelanger:
Point to the definition on the pubspec page


parlough (approved):
Ooo, nice catch. Thanks for thinking of this!

PR #5584, opened 2024-02-21
Specify hook directory

dcharkes:
Specify the `hook/` directory in the package layout.

* https://github.com/dart-lang/sdk/issues/54334

### DIFF
--- a/src/content/tools/pub/package-layout.md
+++ b/src/content/tools/pub/package-layout.md
@@ -501,10 +501,12 @@ enchilada/
 Packages can define hooks to be invoked by the Dart and Flutter SDK.
 These hooks have a predefined CLI, and will be invoked by the SDK tools if present.
 
-:::note
 Because these hooks are invoked by the SDKs on runs and builds, the dependencies
 of these hooks must be normal dependencies and not dev dependencies.
-:::
+
+Hooks are currently an experimental feature.
+For more information see the
+[build.dart hook documentation](https://github.com/dart-lang/native/blob/main/pkgs/native_assets_cli/README.md).
 
 ## Project-specific caching for tools
 
--- a/src/content/tools/pub/package-layout.md
+++ b/src/content/tools/pub/package-layout.md
@@ -505,9 +505,15 @@ Because these hooks are invoked by the
 `dart` and `flutter` tools on runs and builds, the dependencies
 of these hooks must be normal dependencies and not `dev_dependencies`.
 
-Hooks are currently an experimental feature.
-For more information see the
-[build.dart hook documentation](https://github.com/dart-lang/native/blob/main/pkgs/native_assets_cli/README.md).
+:::note
+Hook support is **experimental** and in active development.
+
+To learn more about how to define hooks and their current status,
+refer to the [`build.dart` hook documentation][].
+:::
+
+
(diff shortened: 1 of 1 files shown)


src/content/tools/pub/package-layout.md
```diff
@@ -488,6 +490,22 @@ documentation generators, or other bits of automation.
...
 
+## Hooks
+
+```plaintext
+enchilada/
+  hook/
+    build.dart
+```
+
+Packages can define hooks to be invoked by the Dart and Flutter SDK.
```
mit-mit, line 501: I think we should link to something that has more details here. Even if we don't have docs we can link to, can't we link to an issue or similar?

parlough: Yes please. Otherwise developers will need to find details on their own or potentially just try adding file.

Also, is the usage of `/hook` still experimental in a sense that this could change or the functionality dropped, or perhaps only available when a certain experiment is enabled or on a certain branch? Whatever state, consider making that clear to avoid any incorrect assumptions of support.

dcharkes: Maybe it's better to just point to https://github.com/dart-lang/native/blob/main/pkgs/native_assets_cli/README.md, this will be updated on a more continuous basis. WDYT?

mit-mit: Linking to README SGTM

src/content/tools/pub/package-layout.md
```diff
@@ -488,6 +490,24 @@ documentation generators, or other bits of automation.
...
+These hooks have a predefined CLI, and will be invoked by the SDK tools if present.
+
+Because these hooks are invoked by the SDKs on runs and builds, the dependencies
+of these hooks must be normal dependencies and not dev dependencies.
```
jonasfj, line 505:
```suggestion
of these hooks must be normal dependencies and not `dev_dependencies`.
```

jonasfj: We don't really need to document this in general.

Since each entry in `hook/` should have its own documentation.

jonasfj (approved):

src/content/tools/pub/package-layout.md
dcharkes: Yeah, I would expect this to hold for all hooks, given they are Dart code and we will not do multiple resolutions, which means we will run them with the package resolution for the dependencies of the root package.

mit-mit (approved):

dcharkes:
I have no merge powers. Over to you @parlough.

dcharkes:
Suggestions look good! Thanks @parlough!

Ready to merge! 👌

parlough:
Thanks! Sorry about the delay, I've been traveling :) 

One question though. Is this directory reserved for the Dart and Flutter SDKs and other packages should avoid relying on files in the directory?

parlough:
> I believe packages wanting to run code from other packages should just invoke the APIs in a package's lib/ dir. No need for a protocol or hooks. Or did you have another use case in mind?

That makes sense to me, thanks! No specific use case, I just envision it to be a consideration some developers might have, as it's the first directory we reserve in that way. So I want to make sure I keep it in mind for any future documentation :)

Issue #5593, opened 2024-02-23
Implement and migrate to simpler excerpter tool


parlough:
The current excerpter is based on build_runner and is completing in two steps: determining fragments, then inserting them. We've never taken advantage of this functionality and a simpler, one-step, solution would run fast enough where it's not a problem. The current tool is quite complicated as well, supporting lots of functionality and syntax we no longer use or need on the sites, making it much harder to understand and maintain. It's current implementation also has a few issues on Windows.

To solve all of this, reimplement tool focused only on what the sites use and without worrying about saving fragments or caching results between runs.

PR #5594, opened 2024-02-23
Standardize and clean up code excerpt regions

parlough:
Updates code excerpt region openings, closings, and names to follow a consistent standard (`lower-case-with-dashes`) to make the site contribution experience more consistent and enable the new excerpter tool to warn about more potential mistakes. It also helps the implementation of the excerpter tool simpler as well :)

Contributes to https://github.com/dart-lang/site-www/issues/5593


MaryaBelanger (approved):
Those are in fact `lower-case-with-dashes` name. Approved!

Issue #5573, opened 2024-02-20
Contents bar doesn't show up on Libraries sub pages


MaryaBelanger:
Contents heading with the page's subheadings are not showing up on the right of the page. E.g. Library page:

<img width="1360" alt="image" src="https://github.com/dart-lang/site-www/assets/111139605/2466b750-f0f8-4097-a0e2-dfa8fa4c4393">

Any other page:
<img width="1396" alt="image" src="https://github.com/dart-lang/site-www/assets/111139605/62e18559-801a-4009-aec2-570bf4b26a52">

PR #5595, opened 2024-02-26
Fix heading levels on libraries pages

parlough:
I also added a check to the build to ensure `h3` headers are always located below an `h2`, so that this doesn't happen again :)

Fixes https://github.com/dart-lang/site-www/issues/5573


MaryaBelanger (approved):
> ensure `h3` headers are always located below an `h2`

Aha, that's why 😅 Definitely the product of my sloppy copy-pasting. The error message is a great touch, thanks!

Issue #5579, opened 2024-02-20
Unclear line in README


MaryaBelanger:
*_Possibly just unclear to me_ 😅 

[Lines 178-179](https://github.com/dart-lang/site-www/blob/main/README.md?plain=1#L178-L179) say:

> To instead view the source of generated site files,
   check the `_site` directory.

1. What does that mean / why might someone want to do that?
2. To actually do it, what is it actually telling us to do?
    Does it mean append `_site` to `http://localhost:4000`?
    Or going into your IDE to view the markdown? 

I am confused by this suggestion.

Issue #5576, opened 2024-02-20
Possible error in README about Node version constraint


MaryaBelanger:
[Line 72](https://github.com/dart-lang/site-www/blob/main/README.md?plain=1#L72) says Node has to be version 20 _or later_. I had `20.0.0` installed. But, `corepack enable` didn't work for me until I upgraded node. 

Unfortunately I closed the terminal without saving the complaint from corepack about updating (and it's possible that closing and reopening the terminal is what solved the issue), but maybe it's worth someone else testing to see if we need to change the guidance to a newer version (I think it told me `>=20.5.0` or `>=20.20.0`)

huycozy:
I'm using Node `v20.11.1` but don't see any issue with `corepack` commands (enable or install) on my machine. It would be helpful for investigation if you could find the history of command's output.

huycozy:
I checked [nodejs v20.0.0 changelog](https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V20.md#2023-04-18-version-2000-current-rafaelgss) but I still see `corepack` is mentioned:

```
deps: update corepack to 0.17.2 (Node.js GitHub Bot)
```

I'm not sure if we should update the version constraint in the docs or not as there is not much information (history command output or error stack trace). Will label this for other's input.

huycozy:
You're welcome. Yes, I think we can close the issue. Once someone else also experiences this, we can re-open and investigate this further.

PR #5597, opened 2024-02-26
Clarify a few mentions in the README

parlough:
Generally I recommend (and earlier in the README suggest) using the latest stable version (so the biggest x and y in (20.x.y)), which should mostly avoid these mismatches. However, the site does have a hard minimum of `20.10` to set a relatively recent baseline from when we did the migration. For clarity, this PR specifies that in the README. That won't need to increase again for quite a while though.

I also plan to introduce some tooling in front of dart_site that will catch tooling issues and explain the steps to resolve them :) - https://github.com/dart-lang/site-www/issues/5596

Fixes https://github.com/dart-lang/site-www/issues/5579
Fixes https://github.com/dart-lang/site-www/issues/5576

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ dart --version
 
 #### Node.js
 
-The latest stable LTS release of Node.js is required to build the site.
+The **latest** stable LTS release of Node.js is required to build the site.
 If you don't have Node.js or need to update, download your
 computer's corresponding version and follow the instructions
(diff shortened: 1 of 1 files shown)


MaryaBelanger (approved):
Tiny changes make a big difference! Thank you for this PR and ALL this work, Parker, it's really such a huge improvement ❤️

PR #5602, opened 2024-02-28
Separate excerpts check from other tests

parlough:
Excerpt checking is independent of Dart channels and the formatting, analysis, or testing of the Dart code, so it can be pulled out to a separate job. This reduces duplicated checks and results in CI being completed slightly faster, enabling quicker visibility into failure reasons :D


MaryaBelanger (approved):
🎉

PR #5606, opened 2024-02-28
Add a shell script to validate setup and defer to tooling

parlough:
This will simplify `dart run dart_site <command>` to just `./dash_site <command>`. The command also validates that the setup has been completed correctly and with the correct versions, with tips if not, and then defers to the underlying `dart_site` tool.

This only works on macOS and Linux for now as I don't currently have a Windows device with me to build and test a Powershell equivalent. Windows users can keep using `dart run dart_site` for now, but I'll work on get an equivalent ps alternative added as soon as possible.

Fixes https://github.com/dart-lang/site-www/issues/5596

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,37 +162,44 @@ following the instructions in [Get the prerequisites](#get-the-prerequisites).
    pnpm install
    ```
 
-5. From the root directory, serve the site locally.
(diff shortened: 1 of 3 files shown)


atsansone (approved):
LGTM

PR #5607, opened 2024-02-28
Mark IA32 support as deprecated in system requirements table

parlough:
This change doesn't commit to anything but indicates to developers that support for IA32 might be less and that there is a chance for removal in the future.

Adds a new deprecated icon to facilitate this and adds aria-labels to each entry for improved accessibility of the table cells.

Contributes to https://github.com/dart-lang/sdk/issues/49969

**Staged:** https://dart-dev--pr5607-misc-mark-ia32-as-deprecated-8mye30gd.web.app/get-dart#system-requirements

### DIFF
--- a/src/_sass/_site.scss
+++ b/src/_sass/_site.scss
@@ -588,7 +588,13 @@ thead:has(th:empty) {
 .content > p {
   > i.material-symbols, > span.material-symbols {
     vertical-align: bottom;
+    user-select: none;
   }
+}
+
+// Make icons used in system requirements table unselectable.
+.system-support {
+  user-select: none;
 }
 
 // Overwrite bootstrap's breadcrumbs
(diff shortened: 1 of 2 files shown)


atsansone (approved):
@parlough : LGTM. 

@mit-mit : Any issues if I merge this?

parlough:
> There was a concern regarding using [Ubuntu LTS] as it implies supporting older LTS versions too (versions older than 20.04 

@a-siva I'd prefer we don't list specific versions if possible as that's harder to maintain, but I've opened https://github.com/dart-lang/site-www/pull/5647 to address the concern of versions before 20.04. We used to clarify it's only for LTS versions currently "under standard support" which would only include 20.04 and 22.04 currently, with 24.04 replacing 20.04 soon. https://github.com/dart-lang/site-www/pull/5647 adds that back. How does that sound?

@mit-mit Any concerns with landing this?

PR #5611, opened 2024-02-29
Update breaking changes to account for waitFor removal

parlough:
It was removed in https://github.com/dart-lang/sdk/commit/9e11d796570bc1b83ea15709c3b40266b5cc0c54.


MaryaBelanger (approved):
Good catch!

PR #5620, opened 2024-03-04
Correct type explanation on Extension types page

### LINKED ISSUE #5610: Correct the explanation on `Extension types` page
### Page URL

https://dart.dev/language/extension-types/

### Page source

https://github.com/dart-lang/site-www/tree/main/./src/content/language/extension-types.md

### Describe the problem

The sentence 

> An expression with the extension type `E` and the representation type `R` as its static type will be an object with type `R` at run time.

is slightly misleading. The problem is that we can read 'the representation type `R` as its static type' (and forget about 'the extension type `E`'), and conclude that the static type of the expression is `R`.

The static type of the expression is actually `E`, and the type `E` has a representation type which is denoted by `R` here, but `R` itself doesn't have any guaranteed relationship with the static type of that expression.

### Expected fix

We could say the following:

> If `e` is an expression whose static type is an extension type `E`, and the representation type of `E` is `R`, then the run-time type of the value of `e` is a subtype of `R`.

It is not guaranteed that the run-time type of this object is precisely `R`, and `e` might have many different values because we could reach that location in code and evaluate it many times, but it might still be OK to say that 'the run-time type of `e` is `R`', which would make it a bit less heavy.

### Additional context

_No response_

MaryaBelanger:
Fixes #5610

### DIFF
--- a/src/content/language/extension-types.md
+++ b/src/content/language/extension-types.md
@@ -446,8 +446,9 @@ It's important to be aware of this quality when using extension types.
 Always keep in mind that an extension type exists and matters at compile time,
 but gets erased _during_ compilation.
 
-An expression with the extension type `E` and the representation type `R`
-as its static type will be an object with type `R` at run time.
+For example, consider an expression `e` whose static type is the
+extension type `E`, and the representation type of `E` is `R`.
+Then, the run-time type of the value of `e` is a subtype of `R`.
 Even the type itself is erased;
 `List<E>` is exactly the same thing as `List<R>` at run time.
 


src/content/language/extension-types.md
```diff
@@ -446,8 +446,9 @@ It's important to be aware of this quality when using extension types.
 Always keep in mind that an extension type exists and matters at compile time,
 but gets erased _during_ compilation.
 
-An expression with the extension type `E` and the representation type `R`
-as its static type will be an object with type `R` at run time.
+For example, consider an expression `e` whose static type is the
+extension type `E`, and the representation type of `E` is `R`.
+Then, the run-time type of the value of `e` is a subtype of `R`.
```
MaryaBelanger, line 451: @eernstg I think you said it would be "OK" to say `e`'s type is R, so could we maybe combine the two points just to be extra explicit?:
```suggestion
Then, the run-time type of the value of `e` is `R` or a subtype of `R`.
```

src/content/language/extension-types.md
```diff
@@ -446,8 +446,9 @@ It's important to be aware of this quality when using extension types.
...
 Even the type itself is erased;
 `List<E>` is exactly the same thing as `List<R>` at run time.
```
MaryaBelanger, line 453: @eernstg Does this statement still make sense after changing the previous? I think so but just checking

src/content/language/extension-types.md
eernstg: It's always tricky to try to predict the interpretation that "most readers" will use for any given wording. For example, "the object `o` has type `R`" could mean "the run-time type of `o` is `R`" or it could mean "the run-time type of `o` is a subtype of `R`", and the latter _always_ means "the run-time type of `o` is `R` or a subtype of `R`".

So should we go over all of site-www and change "is a subtype of `T`" to "is `T` or a subtype of `T`", or should we try to say, once and for all, that the subtype relation is reflexive and "a subtype of `T`" _always_ includes `T` itself? (The specification documents, and perhaps some other documents, use 'a proper subtype of `T`' to denote that rare exception where `T` is not included.)

I think that's a sitewide policy matter, so I'll just rely on you knowing much more about which approach is preferred globally.

src/content/language/extension-types.md
eernstg: Yes, that's equally useful after the change.

The fact that an expression has a certain static type `T` and the value(s) of that expression can have various types as their run-time type is one thing. Line 451-452 is about the meaning of the type itself and that's an orthogonal topic. The type itself will determine the semantics of expressions like `e is List<E>`, and in this expression the static type of `e` does not matter at all.

eernstg (approved):
LGTM!

src/content/language/extension-types.md
MaryaBelanger: Aha, I did not realize that subtypes are a reflexive relation and was just speaking colloquially (like, for the sake of overt clarity). If that's the definition, then I think it's safe to use it correctly! I'll scrap the suggestion

PR #5621, opened 2024-03-04
Revise "Web libraries and packages" page

### LINKED ISSUE #5608: Update 'Web libraries and packages' to reference `package:web` instead
### Page URL

https://dart.dev/web/libraries/

### Page source

https://github.com/dart-lang/site-www/tree/main/./src/content/web/libraries.md

### Describe the problem

This page potentially has limited benefit, but might be useful in the short term as the JS interop and `package:web` docs are located in the newer **Interop** category now.

Let's update it and revisit its necessity later down the road.

### Expected fix

Replace mentions of `dart:html` with `package:web`.

Can leave a small note of the historical use of `dart:html`, linking to its docs, but with a suggestion to migrate (and link to migration doc).

### Additional context

_No response_

MaryaBelanger:
Fixes #5608 

@parlough It ended up being a little more of a change than just replacing dart:html with package:web, since the page was all structured by "web libraries" vs "web packages". I got rid of that distinction and just listed everything together. 

The "note" about dart:html is in the description about the migration page, the first solutions link. Lmk if you wanted something more overt than that

### DIFF
--- a/src/content/web/libraries.md
+++ b/src/content/web/libraries.md
@@ -5,24 +5,26 @@ description: Libraries and packages that can help you write Dart web apps.
 ---
 
 Dart provides several packages and libraries to support
-web app development, the recommended option being [package:web][].
+web app development, the recommended option being [`package:web`][web].
 The [Dart SDK][] also contains other libraries that provide low-level web APIs.
 
 ## Web solutions
 
-[Migrate to package:web][migrate]
+[Migrate to `package:web`][migrate]
 : Learn how to migrate to `package:web`
-  from Dart's previous web solution, [dart:html][].
+  from Dart's previous web library solutions, like [`dart:html`][html].
 
-[package:web API reference][package:web]
-: Complete reference documentation for `package:web` on pub.dev.
+[`package:web` API reference][web]
+: Dart's recommended web interop solution `package:web` exposes browser
+  APIs with lightweight bindings built around static JS interop. 
 
-[JavaScript interoperability documentation](/interop/js-interop)
-: Interact with existing JavaScript or TypeScript libraries
+[JavaScript interoperability documentation][js]
+: Learn how to interact with existing JavaScript or TypeScript libraries
   using Dart's JS interop support.
 
-[The dart:js_interop documentation][js_interop]
-: Complete reference documentation for the `dart:js_interop` library.
+[`dart:js_interop` API reference][js_interop]
+: Dart's web library `dart:js_interop` provides all the necessary members to
+  facilitate sound interop between JavaScript and Dart types. 
 
 [Flutter web support][flutter-web]
 : The [Flutter framework][flutter] supports web development with Dart,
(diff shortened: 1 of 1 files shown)


src/content/web/libraries.md
```diff
@@ -4,46 +4,41 @@ short-title: Web libraries
...
-[dart:html]: {{site.dart-api}}/{{site.sdkInfo.channel}}/dart-html/dart-html-library.html
+## Web solutions
 
+[Migrate to package:web][migrate]
+: Learn how to migrate to `package:web`
+  from Dart's previous web solution, [dart:html][].
 
-## SDK libraries
+[package:web API reference][package:web]
+: Complete reference documentation for `package:web` on pub.dev.
```
craiglabenz, line 18: I wonder if it might be better to give a one-line summary of what this package is, instead of indicating that this link goes to it's documentation. Doing that for this and `js_interop` below might help a reader more quickly identify the right link for them without having to click on them all.

craiglabenz (approved):
LGTM with one thought that I defer to @MaryaBelanger's judgement

src/content/web/libraries.md
```diff
@@ -4,46 +4,41 @@ short-title: Web libraries
 description: Libraries and packages that can help you write Dart web apps.
 ---
 
-The [Dart SDK][] contains [dart:html][] and other libraries
-that provide low-level web APIs.
-You can supplement or replace these APIs using web packages.
+Dart provides several packages and libraries to support
+web app development, the recommended option being [package:web][].
```
parlough, line 8: Perhaps it's only me, but if using the `package:<name>` or `dart:<name>` syntax, I personally prefer to put it in code syntax. What do you think (here and elsewhere on the page)?

src/content/web/libraries.md
```diff
@@ -4,46 +4,41 @@ short-title: Web libraries
...
-## Web packages
-
-Many [packages](/guides/packages) support web development with Dart.
-In particular, the [Flutter framework][flutter] has [web support][flutter-web],
-in addition to mobile, desktop, and embedded device support.
+[Build a web app with Dart](/web/get-started)
+: A quick overview of how to build, run, and debug a web app with Dart.
 
-To find more libraries that support writing web apps, 
+To find more support for writing web apps,
```
parlough, line 34: "To find more support" reads to me as the link will lead to me docs not packages, so I think the previous text or some other modification is likely a bit more clear.

Perhaps:

```suggestion
To find other libraries that support the web platform,
```

parlough (approved):
Thanks for tackling this Marya!! Two additional considerations and a +1 to Craig's suggestion.

src/content/web/libraries.md
MaryaBelanger: Definitely not just you! The original page didn't code syntax any of them, so that tripped me up. Thanks for catching that!

src/content/web/libraries.md
MaryaBelanger: That's a great idea, added!

PR #5622, opened 2024-03-04
Add destructuring named fields to "Records" page

### LINKED ISSUE #5262: Add example on how to destructure named records

i was reading the page about records, hoping to find out how to destructure records with named fields, but found nothing.  
[this section](https://dart.dev/language/records#multiple-returns) only showcases how to destructure a record with positional fields.  
i tried the following:
```dart
({ String name, int age }) userInfo() { ... }

void test(){
  // this is the pattern for positional fields. maybe it will work here.
  final ( name, age ) = userInfo();
  // i get the error:
  // The matched value of type '({int age, String name})' isn't assignable to the required type '(Object?, Object?)'
	
  // maybe use curly braces?
  final ({ name, age }) = userInfo();
  // nope:
  // The matched value of type '({int age, String name})' isn't assignable to the required type 'Map<Object?, Object?>'
}
```

would be nice to have an example for how to destructure named fields as well.

MaryaBelanger:
Fixes #5262 

Added an example (w/ code excerpt) of destructuring named fields to the "Multiple returns" section, and improved links connecting more info between the patterns, pattern types, and records pages

### DIFF
--- a/examples/language/test/records_test.dart
+++ b/examples/language/test/records_test.dart
@@ -115,7 +115,7 @@ void main() {
   test('record-multiple-returns', () {
     // #docregion record-multiple-returns
     // Returns multiple values in a record:
-    (String, int) userInfo(Map<String, dynamic> json) {
+    (String name, int age) userInfo(Map<String, dynamic> json) {
       return (json['name'] as String, json['age'] as int);
     }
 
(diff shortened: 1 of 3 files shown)


craiglabenz (approved):
LGTM

Issue #5506, opened 2024-02-04
Update outdated mention that factory constructors can return `null`


robsenshemsedin:
### Page URL

https://dart.dev/codelabs/dart-cheatsheet.html

### Page source

https://github.com/dart-lang/site-www/tree/main/src/codelabs/dart-cheatsheet.md

### Describe the problem

The Factory constructors section of this page, describes factory constructors as follows:

>  Dart supports factory constructors, which can return subtypes or even null.

  But factory constructors no longer return `null ` with null-safe Dart. 

### Expected fix

We need to update the existing description to something like, 

> Dart supports factory constructors, which can return instances of a class or even subtypes.

### I would like to fix this problem.

- [X] I will try and fix this problem on dart.dev.

parlough:
Great catch @robsenshemsedin! I'm surprised no one noticed this early. Thank you :D

You mentioned "I will try and fix this problem on dart.dev". Feel free to open a PR with the change and I'd be happy to review it. If you have any issues, let me know :)

robsenshemsedin:
@parlough I have opened PR with a change to the definition of Factory constructors.

PR #5623, opened 2024-03-05
Update outdated mention that factory constructors can return null.

robsenshemsedin:
Factory constructors no longer return null with null-safe Dart. Instead, they can be used to return instance of a class or subtypes.

Fixes <https://github.com/dart-lang/site-www/issues/5506>

### DIFF
--- a/src/content/codelabs/dart-cheatsheet.md
+++ b/src/content/codelabs/dart-cheatsheet.md
@@ -1727,8 +1727,9 @@ void main() {
 
 ## Factory constructors
 
-Dart supports factory constructors,
-which can return instances of a class or even subtypes.
+Dart supports factory constructors.
+Factory constructors can return an instance of the surrounding class
+or a subtype of that class.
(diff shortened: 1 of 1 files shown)


src/content/codelabs/dart-cheatsheet.md
```diff
@@ -1728,7 +1728,7 @@ void main() {
 ## Factory constructors
 
 Dart supports factory constructors,
-which can return subtypes or even null.
+which can return instances of a class or even subtypes.
```
munificent, line 1731: Everything is an instance of a class in Dart (even `null`), so that doesn't say much. But you're exactly right that factory constructors can't return `null` anymore. That's a good catch. How about:

"which can return an instance of the surrounding class or even a subtype of it."

atsansone (approved):

atsansone:
Fixed in https://github.com/dart-lang/site-www/pull/5757.

PR #5627, opened 2024-03-06
Switch to injecting new DartPad UI

parlough:
- Updates to use the new, smaller injection script (introduced in https://github.com/dart-lang/site-shared/pull/205)
- Updates the markdown renderer to set the new expected data variables
- Removes some old code relating to multi-file DartPad

Staged example: https://dart-dev--pr5627-feat-new-dartpad-injection-n1bx8le4.web.app/codelabs/async-await

### DIFF
--- a/eleventy.config.js
+++ b/eleventy.config.js
@@ -91,6 +91,7 @@ export default function (eleventyConfig) {
 
   eleventyConfig.addPassthroughCopy('src/content/assets/dash');
   eleventyConfig.addPassthroughCopy('src/content/assets/js');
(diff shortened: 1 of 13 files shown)


parlough:
I'm getting the following warning from CanvasKit on pages with lots of embeds: `Too many active WebGL contexts. Oldest context will be lost.`

I'm not yet sure how it effects them though.

johnpryan:
There might be a timing issue, where we are injecting the dart code before the Editor widget is instantiated or something. The first few embeds are showing the correct code, but if I scroll down I see some that print the default `hello ${i + 1}` snippet.

johnpryan (approved):
LGTM, I filed an issue to investigate the timing: https://github.com/dart-lang/dart-pad/issues/2893

PR #5628, opened 2024-03-06
Replace custom mini-toc container with secondary

parlough:
The `mini-toc` styling and setup are so similar to the secondary asides now that we can set custom titles for them. So this PR removes support for the `mini-toc` container type in favor of reusing `secondary`, reducing mostly duplicated JS and CSS.


atsansone (approved):
LGTM % the link text feels a little big compared to the right-hand side ToC, but not a blocker.

Issue #5632, opened 2024-03-08
Update JS annotation requirement now that #54801 is Fixed


Cat-sushi:
### Page URL

https://dart.dev/interop/js-interop/usage/

### Page source

https://github.com/dart-lang/site-www/tree/main/./src/content/interop/js-interop/usage.md

### Describe the problem

Following warning should be deleted, because the bug is fixed and Flutter 3.19.3 is released.

> There's a bug that currently requires object literal constructors to have an [@JS](https://dart.dev/interop/js-interop/usage#js) annotation on the library. See [#54801](https://github.com/dart-lang/sdk/issues/54801) for more details.

### Expected fix

remove the note.

### I would like to fix this problem.

- [ ] I will try and fix this problem on dart.dev.

parlough:
Good catch! Thanks for calling this out.

I think it's best to keep the note there for a bit, for those using interop on earlier releases, as it doesn't hurt, but we can perhaps adjust the note for now.

parlough:
I'm going to close this as resolved by https://github.com/dart-lang/site-www/pull/5638, our normal version note triage will catch this in the future alongside future releases. Thanks again!

PR #5638, opened 2024-03-10
Update 54801 issue note

parlough:
Update the note now that a fix has been released.

Contributes to https://github.com/dart-lang/site-www/issues/5632


MaryaBelanger (approved):
Great idea adjusting it for the version instead of completely removing it. Thanks!

PR #5639, opened 2024-03-11
Change html formatting to md

MaryaBelanger:
Working on the Operators page and the html formatting in some of the tables made them hard to read and edit. I'm not sure if there's a reason we do this? But if not I changed all of the `<em><code>` tags in the tables to just use backticks and markdown italices

### DIFF
--- a/src/content/language/operators.md
+++ b/src/content/language/operators.md
@@ -19,24 +19,24 @@ and [operator precedence](#operator-precedence-example) from highest to lowest,
 which are an **approximation** of Dart's operator relationships.
 You can implement many of these [operators as class members][].
 
-| Description                             | Operator                                                                                                                                                                                          | Associativity |
-|-----------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------------|
-| unary postfix                           | <code><em>expr</em>++</code>    <code><em>expr</em>--</code>    `()`    `[]`    `?[]`    `.`    `?.`    `!`                                                                                       | None          |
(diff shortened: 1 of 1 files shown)


src/content/language/operators.md
```diff
@@ -262,10 +262,10 @@ an operation with an assignment.
 
 Here's how compound assignment operators work:
 
-|                                  | Compound assignment           | Equivalent expression            |
-|----------------------------------|-------------------------------|----------------------------------|
-| **For an operator <em>op</em>:** | <code>a <em>op</em>= b</code> | <code>a = a <em>op</em> b</code> |
-| **Example:**                     | `a += b`                      | `a = a + b`                      |
+|                           | Compound assignment | Equivalent expression |
+|---------------------------|---------------------|-----------------------|
+| **For an operator *op*:** | `a `  *`op`*  `= b` | `a = a ` *`op `* `b`  |
```
parlough, line 267: I think this one still makes sense with no space before the `=` if possible to match the spacing of the actual operators:

```suggestion
| **For an operator *op*:** | `a `  *`op`*`= b`   | `a = a ` *`op `* `b`  |
```

parlough (approved):
I forgot why. Perhaps was because the Markdown processor we used with Jekyll didn't handle the mixing too well? Not sure.

Thanks for doing this though! Looks good to me. I looked through the stage site and didn't notice any big issues, just a few spacing changes. Only this one seemed worth addressing:

src/content/language/operators.md
MaryaBelanger: thank you for catching that!

Issue #5635, opened 2024-03-08
[PAGE ISSUE]: 'Operators': The spread operators are missing


dtonhofer:
### Page URL

https://dart.dev/language/operators/

### Page source

https://github.com/dart-lang/site-www/tree/main/./src/content/language/operators.md

### Describe the problem

On page [Operators](https://dart.dev/language/operators) of the Tutorial, we are presented with a quite informative list of Dart operators.

However, it is still missing the new `...collection` and `...?collection` "spread operators" which inject collections into collections.

Additionally, the "if null" operator `a ?? b` is apparently generally called a ["null-coalescing"](https://en.wikipedia.org/wiki/Null_coalescing_operator) operator. One might want to note that.

### Expected fix

As described above

### I would like to fix this problem.

- [ ] I will try and fix this problem on dart.dev.

dtonhofer:
Thank you Marya 👌!

PR #5649, opened 2024-03-14
Add note about if-null operator

MaryaBelanger:
Wanted to address a side point @dtonhofer made in #5635 

> Additionally, the "if null" operator `a ?? b` is apparently generally called a ["null-coalescing"](https://en.wikipedia.org/wiki/Null_coalescing_operator) operator. One might want to note that.

Not trying to rename it, it seems like we use if-null across the board. But I confirmed we do use "null coalescing" to describe our if-null operator in the spec at least once: 

image

So I think it makes sense to say "also known as the null coalescing operator" in at least one place, for people more familiar with that term.


munificent (approved):
Yeah, this is one of those annoying places where a number of languages have added the same or similar feature but some of them picked different names, so it's hard to decide what to call it. I like mentioning the other name.

parlough (approved):
Thanks for this Marya! Will help with searching and those familiar with the other terminology.

PR #5657, opened 2024-03-19
Clarified parenthetical patterns

### LINKED ISSUE #5653: Parenthesized patterns text is misleading
### Page URL

https://dart.dev/language/pattern-types#parenthesized

### Page source

https://github.com/dart-lang/site-www/blob/main/src/content/language/pattern-types.md?plain=1#L291

### Describe the problem

The text says that the pattern `x || y && z` evaluates `y && z` first because logical-and patterns have higher precedence than logical-or.

But in this case, the pattern `x` is evaluated first, and evaluates to true, and the patterns `y` and `z` and the pattern `y && z` are not evaluated at all (and would be evaluated second if x were false).

The example can't really be changed, because && and || are the only binary operators where parentheses can meaningfully be introduced.

Perhaps just say "the subpattern y && z is evaluated" versus "the subpattern x || y is evaluated". Or just say that `x || y && z` is equivalent to `x || ( y && z )`, which is different from `( x || y ) && z` . And don't talk about "evaluated first".

I also think that any example where the different parentheses give different results will also shortcut in at least one case, and not evaluate all the subpatterns, which makes it hard to make an example.

### Expected fix

_No response_

### Additional context

_No response_

atsansone:
Fixes #5653

Based on your comment in the issue, I thought you would be the best to review @eernstg .

### DIFF
--- a/src/content/language/pattern-types.md
+++ b/src/content/language/pattern-types.md
@@ -287,7 +287,7 @@ the example matches patterns.
 // ...
 x || y => 'matches true',
 x || y && z => 'matches true',
-x || (y && z) => 'matches true',
+x || (y && z) => 'matches true', // `x || y && z` is the same thing as `x || (y && z)`.
 (x || y) && z => 'matches nothing',
 // ...
 ```
--- a/src/content/language/pattern-types.md
+++ b/src/content/language/pattern-types.md
@@ -287,7 +287,8 @@ the example matches patterns.
 // ...
 x || y => 'matches true',
 x || y && z => 'matches true',
-x || (y && z) => 'matches true', // `x || y && z` is the same thing as `x || (y && z)`.
+x || (y && z) => 'matches true',
+// `x || y && z` is the same thing as `x || (y && z)`.
 (x || y) && z => 'matches nothing',
 // ...
 ```
--- a/src/content/language/pattern-types.md
+++ b/src/content/language/pattern-types.md
@@ -278,20 +278,34 @@ Like parenthesized expressions, parentheses in a pattern let you control
 pattern where a higher precedence one is expected.
 
 For example, imagine the boolean constants `x`, `y`, and `z`
-equal `true`, `true`, and `false`, respectively:
+equal `true`, `true`, and `false`, respectively.
+Though the following example resembles boolean expression evaulation,
+the example matches patterns.
 
 <?code-excerpt "language/lib/patterns/pattern_types.dart (parens)"?>
 ```dart
 // ...
+x || y => 'matches true',
 x || y && z => 'matches true',
-(x || y) && z => 'matches false',
+x || (y && z) => 'matches true',
+(x || y) && z => 'matches nothing',
 // ...
 ```
 
-Dart evaluates `x || y && z` as `x || (y && z)`,
-therefore the first expression matches `true`.
-As the parenthetical subpattern matches first,
-the second expression matches `false`.
+Dart starts matching the pattern from left to right.
+
+1. The first pattern matches `true` as `x` matches `true`.
+1. The second pattern matches `true` as `x` matches `true`.
+1. The third pattern matches `true` as `x` matches `true`.
+1. The fourth pattern `(x || y) && z` has no match.
+
+   * The `x` matches `true`, so Dart doesn't try to match `y`.
+   * Though `(x || y)` matches `true`, `z` doesn't match `true`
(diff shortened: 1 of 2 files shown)


src/content/language/pattern-types.md
```diff
@@ -288,11 +288,10 @@ x || y && z => 'matches true',
 // ...
 ```
```
eernstg, line 289: This example is a little bit convoluted because `(x || y) && z` doesn't match any boolean values at all. E.g.,

```dart
const x = true;
const y = true;
const z = false;

int f(bool b) => switch (b) {
      x || y && z => 1, // Matches `true`.
      (x || y) && z => 2, // Does not match any `bool` value.
      _ => 3, // Matches every `bool` value.
    };

void main() {
  print(f(true)); // Prints '1'.
  print(f(false)); // Prints '3'.
}
```

We can't just consider the pattern `(x || y) && z` to be similar to the boolean expression `(x || y) && z`.

The expression evaluates to false because `(x || y)` evaluates to true and `z` evaluates to false, which is then the result of the `&&`.

However, the pattern `(x || y) && z` doesn't match true: `x` matches true, so we don't even try to match `y`, we know that `(x || y)` matches true. But `z` does not match true, so `(x || y) && z` doesn't match true.

Similarly, `(x || y) && z` doesn't match false, because `(x || y)` doesn't match false (and we don't even try matching with `z`).

This means that the strings in lines 286 and 287 are misleading.

src/content/language/pattern-types.md
```diff
@@ -288,11 +288,10 @@ x || y && z => 'matches true',
...
 
-In the first case, the logical-and pattern `y && z` evaluates first because
-logical-and patterns have higher precedence than logical-or.
-In the next case, the logical-or pattern is parenthesized. It evaluates first,
-which results in a different match.
-
+Dart evaluates `x || y && z` as `x || (y && z)`,
```
eernstg, line 291: It is again dangerous to say this, because Dart doesn't evaluate a pattern: It performs matching operations that may succeed or fail (or throw), and the resulting behavior depends on whether or not there was a match for the entire pattern.

What we _can_ say is that `p1 || p2` proceeds by matching the matched value with `p1` and succeeds if the match succeeded, otherwise it matches the matched value with `p2` and succeeds if the match succeeded; and otherwise the match fails.

I think it's safe to say that the pattern `x || y && z` is equivalent to the pattern `x || (y && z)`, and then we can use the description of how to match an object against `x` respectively how to match that object against `y && z` to understand how the pattern `x || (y && z)` behaves.

eernstg: I think it might actually be a really good point to make that `(x || y) && z` doesn't match anything, because this puts a strong emphasis on the fact that pattern matching isn't boolean expression evaluation.

eernstg (commented):
Lots of comments, it probably makes sense to take another iteration. ;-)

src/content/language/pattern-types.md
```diff
@@ -277,22 +277,35 @@ Like parenthesized expressions, parentheses in a pattern let you control
...
+Though the following example resembles boolean expression evaulation,
+the example matches patterns.
 
 <?code-excerpt "language/lib/patterns/pattern_types.dart (parens)"?>
 ```dart
 // ...
+x || y => 'matches true',
 x || y && z => 'matches true',
-(x || y) && z => 'matches false',
+x || (y && z) => 'matches true',
```
eernstg, line 290: Maybe something like this would be helpful:
```suggestion
x || (y && z) => 'matches true', // `x || y && z` is the same thing as `x || (y && z)`.
```

src/content/language/pattern-types.md
eernstg: That's great, thanks!

src/content/language/pattern-types.md
```diff
@@ -277,22 +277,35 @@ Like parenthesized expressions, parentheses in a pattern let you control
...
+(x || y) && z => 'matches nothing',
 // ...
 ```
 
-In the first case, the logical-and pattern `y && z` evaluates first because
-logical-and patterns have higher precedence than logical-or.
-In the next case, the logical-or pattern is parenthesized. It evaluates first,
-which results in a different match.
+Dart starts matching the pattern from left to right.
```
eernstg, line 295: That's very good, we don't even get anywhere near the confusion about the ordering.

eernstg (approved):
LGTM!

PR #5667, opened 2024-03-25
Fix and standardize various links in JS interop docs

parlough:
- Don't link to `dev` version of API docs
- Use site variables where possible
- Don't link to en-US MDN docs, so the site can choose the appropriate locale for the user


MaryaBelanger (approved):
Thanks so much Parker!

kevmoo:
Bless you, kind sir. 🙏

PR #5668, opened 2024-03-26
Add docs for ExternalDartReference

srujzs:
This type was added as part of https://github.com/dart-lang/sdk/issues/55187 to dart:js_interop for a faster alternative to JSBoxedDartObject, so we should add some documentation on it.

### DIFF
--- a/src/content/interop/js-interop/js-types.md
+++ b/src/content/interop/js-interop/js-types.md
@@ -35,17 +35,12 @@ JS types form a natural type hierarchy:
       - JS typed arrays like `JSUint8Array`
     - `JSBoxedDartObject`, which allows users to box and pass Dart values
       opaquely within the same Dart runtime
+      - From Dart 3.4 onwards, the type `ExternalDartReference` in
+      `dart:js_interop` also allows users to pass Dart values opaquely, but is
+      *not* a JS type. Learn more about the tradeoffs between each option
+      [here](#jsboxeddartobject-vs-externaldartreference).
 
 You can find the definition of each type in the [`dart:js_interop` API docs].
-
-From Dart 3.4 onwards, there also exists one type in `dart:js_interop` that can
-be used but is *not* a JS type called `ExternalDartReference`.
-
-Both `JSBoxedDartObject` and `ExternalDartReference` pass opaque references to
-Dart `Object`s through JavaScript. However, `JSBoxedDartObject` wraps the opaque
-reference in a JavaScript object, while `ExternalDartReference` is the reference
-itself and therefore is not a JS type. Use `ExternalDartReference` where
-performance is needed and `JSBoxedDartObject` where a JS type is needed.
 
 {% comment %}
 TODO (srujzs): Should we add a tree diagram instead for JS types?
@@ -81,7 +76,7 @@ Generally, the conversion table looks like the following:
 
 <div class="table-wrapper" markdown="1">
 
-| JS type                             | Dart type                                |
+| Interop type                        | Dart type                                |
 | ----------------------------------- | ---------------------------------------- |
 | `JSNumber`, `JSBoolean`, `JSString` | `num`, `int`, `double`, `bool`, `String` |
 | `JSExportedDartFunction`            | `Function`                               |
@@ -89,6 +84,7 @@ Generally, the conversion table looks like the following:
 | `JSPromise<T extends JSAny?>`       | `Future<T extends JSAny?>`               |
 | Typed arrays like `JSUint8Array`    | Typed lists from `dart:typed_data`       |
 | `JSBoxedDartObject`                 | Opaque Dart value                        |
+| `ExternalDartReference`             | Opaque Dart value                        |
 
 {:.table .table-striped}
 </div>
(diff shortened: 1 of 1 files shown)


sigmundch (approved):
Only some suggestions, but LGTM even as is

src/content/interop/js-interop/js-types.md
```diff
@@ -73,7 +82,7 @@ Generally, the conversion table looks like the following:
 <div class="table-wrapper" markdown="1">
 
 | JS type                             | Dart type                                |
```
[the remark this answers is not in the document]
MaryaBelanger: > I believe we use JS-type and interop-type interchangingly.

We actually don't use them interchangeably. I think JS types are interop types, but not all interop types are JS types. From the [Usage](https://dart.dev/interop/js-interop/usage#interop-types) page:

> Interop types are either a ["JS type"](https://dart-dev--pr5668-addexternaldartreference-45iczn14.web.app/interop/js-interop/js-types) provided by Dart or an [extension type](https://dart-dev--pr5668-addexternaldartreference-45iczn14.web.app/language/extension-types) wrapping an interop type.

src/content/interop/js-interop/js-types.md
```diff
@@ -38,6 +38,15 @@ JS types form a natural type hierarchy:
 
 You can find the definition of each type in the [`dart:js_interop` API docs].
```
MaryaBelanger, line 40: I think the new content should either be 1) primarily in the api docs, or 2) in a breakout section on this page.

It's a sizable amount of digression for the Type hierarchy section. I think either of these options would also allow more room for Siggi's [comment](https://github.com/dart-lang/site-www/pull/5668/files#r1540041570) about elaborating on when to use each.

1) API docs:

Add a sub-bullet under the `JSBoxedDartObject` item in the list (line 38) and put your first two sentences there (slight rewrite for wording):

```markdown
- `JSBoxedDartObject`, which allows users to box and pass Dart values
    opaquely within the same Dart runtime.
  - From Dart 3.4 onwards, the type `ExternalDartReference` in `dart:js_interop` can be used to
    accomplish the same thing, but is *not* a JS type. To learn more about the tradeoffs between
    each option, read [API docs link].
```

And then your second paragraph (lines 44-48) could be a note in the API docs for either or both types (I think both makes sense).

2) Another section on this page:

Add the same sub-bullet in 1), but link to the section on the page instead (`To learn more about the tradeoffs between each option, read [link to section]`).

At the bottom of the page, add a section and include your second paragraph + Siggi's "when a JS type is needed" explanation:

```markdown
## `JSBoxedDartObject` vs `ExternalDartReference`

Both `JSBoxedDartObject` and `ExternalDartReference` pass opaque references to
Dart `Object`s through JavaScript. However, `JSBoxedDartObject` wraps the opaque
reference in a JavaScript object, while `ExternalDartReference` is the reference
itself and therefore is not a JS type. Use `ExternalDartReference` where
performance is needed and `JSBoxedDartObject` where a JS type is needed.

(edit as needed: )
For example, if you just want to pass references opaquely back and forth using the
reference as an identity, then `ExternalDartReference` is a good fit. However, if you want
a reference that can also store data and you want to extend it to add data to that reference,
then we recommend using `JSBoxedDartObject`. 
```

And I'd even add more of the tradeoff details from your issue https://github.com/dart-lang/sdk/issues/55187

---

I don't have a preference, though I'm not too familiar with API doc best practices so I'm not sure if 1) makes sense for the kind of information usually kept there, or if repeating the explanation on both type's pages is bad practice 🤷‍♀️

srujzs: Thanks for the suggestion! I like the idea of splitting it into a separate section (2) and avoiding polluting the type hierarchy section so I went ahead and did that.

src/content/interop/js-interop/js-types.md
```diff
@@ -252,4 +275,8 @@ TODO: add links (with stable) when ready:
 [`instanceOfString`]: https://api.dart.dev/dev/dart-js_interop/JSAnyUtilityExtension/instanceOfString.html
 [`isA`]: https://api.dart.dev/dev/dart-js_interop/JSAnyUtilityExtension/isA.html
 [#4841]: https://github.com/dart-lang/linter/issues/4841
-[#54025]: https://github.com/dart-lang/sdk/issues/54025
\ No newline at end of file
+[#54025]: https://github.com/dart-lang/sdk/issues/54025
+[`JSBoxedDartObject`]: https://api.dart.dev/main/dart-js_interop/JSBoxedDartObject-extension-type.html
+[`ExternalDartReference`]: https://api.dart.dev/main/dart-js_interop/ExternalDartReference-extension-type.html
+[`toExternalReference`]: https://api.dart.dev/main/dart-js_interop/ObjectToExternalDartReference/toExternalReference.html
+[`toDartObject`]: https://api.dart.dev/main/dart-js_interop/ExternalDartReferenceToObject/toDartObject.html
```
MaryaBelanger, line 282:
```suggestion
[`JSBoxedDartObject`]:  {{site.dart-api}}/{{site.sdkInfo.channel}}/dart-js_interop/JSBoxedDartObject-extension-type.html
[`ExternalDartReference`]:  {{site.dart-api}}/{{site.sdkInfo.channel}}/dart-js_interop/ExternalDartReference-extension-type.html
[`toExternalReference`]:  {{site.dart-api}}/{{site.sdkInfo.channel}}/dart-js_interop/ObjectToExternalDartReference/toExternalReference.html
[`toDartObject`]:  {{site.dart-api}}/{{site.sdkInfo.channel}}/dart-js_interop/ExternalDartReferenceToObject/toDartObject.html
```

The rest of the links were updated in [this PR](https://github.com/dart-lang/site-www/pull/5667/files), so I think you'll get those changes if you just update the branch. This suggestion is to correct the new ones.

MaryaBelanger (approved):
Thanks for incorporating my suggestions! One more thing about the links but otherwise this looks great

src/content/interop/js-interop/js-types.md
srujzs: Done, thanks!

PR #5676, opened 2024-04-01
Clarify renames section on package:web page

### LINKED ISSUE #5601: Clarify `package:web` migration text
Hey @srujzs! There's a [part](https://dart.dev/interop/js-interop/package-web#migrating-from-dart-html:~:text=that%20it%20annotates.-,Similarly,-%2C%20if%20you%20find) under the [Renames](https://dart.dev/interop/js-interop/package-web#renames) section that I don't quite understand (sorry, missed it during the PR review).

> Similarly, if you find an API with the keyword `native` in `dart:html` that doesn't have an equivalent in `package:web`, check to see if there was a rename with the `@JSName` annotation. `native` is an internal keyword that means the same as `external` in this context.

"...if you find an API with the keyword `native` in `dart:html` that doesn't have an equivalent in `package:web`..." 
- I'm assuming by "if you find" you mean "if the two methods we just mentioned (the `fix` and looking up the definition manually) don't turn up an answer", is that correct?
  - Or, is it more specific, like "if neither of those other methods turned up anything _AND_ the API has the `native` keyword?
- What does it mean for an API to be "_with_ the keyword `native`"? Like, does that mean `native` is literally in the API's name, or in its declaration somewhere? I can't picture it.

"...check to see if there was a rename with the `@JSName` annotation."
- Where do they go to check that? 
- Once they get to the place where they can check, how will they know "...if there was a rename with the `@JSName` annotation"?
- If they find the answer to "...if there was a rename..." is "yes, there was", what do they do now?

I guess I'm having trouble picturing all of this, could you show an example (or make one up) of an API with the `native` keyword and what looking that up would look like, plus what they would see that would confirm an `@JSName` rename? Thank you!

MaryaBelanger:
Fixes #5601

### DIFF
--- a/src/content/interop/js-interop/package-web.md
+++ b/src/content/interop/js-interop/package-web.md
@@ -111,28 +111,45 @@ You can address these either:
 - In your IDE, by selecting the `dart fix`: **Rename to '`package:web name`'**.
 
 {% comment %}
-Maybe a pic here of menu selection in IDE?
 TODO: Update this documentation to refer to symbols instead of just types once
 we have a dart fix for that.
 {% endcomment -%}
 
 The `dart fix` covers many of the common type renames.
-If you come across a `dart:html` type without a built-in fix, let us know by
-filing an [issue][].
-You can manually discover the `package:web` type name
-by looking up the `dart:html` class' `@Native` annotation.
-You can do this by either:
+If you come across a `dart:html` type without a `dart fix` to rename it,
+first let us know by filing an [issue][].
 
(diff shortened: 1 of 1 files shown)


srujzs (approved):
LGTM, thanks!

There are maybe a handful of members we rename in `package:web` because they use the same name as unshadowable language keywords, but it's not worth mentioning here as that would lead to unnecessary confusion.

parlough (approved):

Issue #11, opened 2016-07-14
Confusing results from the link checker


Sfshaza:
I've just removed the trailing slash from all permalinks on www and have fixed all places where these links were referenced (also by removing the trailing links).

When I now run deploy/html_proof.rb, and I remove all "duplicate ID" errors, I get 13 errors that I don't understand. For several articles, and some other files, the link checker claims that these files reference themselves with a trailing backslash. They do NOT. I've checked the generated files in publish. I've checked the JavaScript console in the browser. I've checked in "view source".

This is a head scratcher. Here's the list of offending files:

```
- /var/folders/b6/r_1vr90j2mj1bzcg8gbxsx_m00713m/T/___articles___dart-vm___numeric-computation_____20160714-4473-1q11n7v.html
  *  External link http://localhost:5000/articles/dart-vm/numeric-computation/ failed: 404 No error

- /var/folders/b6/r_1vr90j2mj1bzcg8gbxsx_m00713m/T/___articles___dart-vm___simd_____20160714-4473-du0i8z.html
  *  External link http://localhost:5000/articles/dart-vm/simd/ failed: 404 No error

- /var/folders/b6/r_1vr90j2mj1bzcg8gbxsx_m00713m/T/___articles___libraries___converters-and-codecs_____20160714-4473-j9blb2.html
  *  External link http://localhost:5000/articles/libraries/converters-and-codecs/ failed: 404 No error

- /var/folders/b6/r_1vr90j2mj1bzcg8gbxsx_m00713m/T/___articles___libraries___creating-streams_____20160714-4473-1pu1z18.html
  *  External link http://localhost:5000/articles/libraries/creating-streams/ failed: 404 No error

- /var/folders/b6/r_1vr90j2mj1bzcg8gbxsx_m00713m/T/___articles___libraries___futures-and-error-handling_____20160714-4473-1yc62q5.html
  *  External link http://localhost:5000/articles/libraries/futures-and-error-handling/ failed: 404 No error

- /var/folders/b6/r_1vr90j2mj1bzcg8gbxsx_m00713m/T/___articles___libraries___zones_____20160714-4473-13x22m5.html
  *  External link http://localhost:5000/articles/libraries/zones/ failed: 404 No error

- /var/folders/b6/r_1vr90j2mj1bzcg8gbxsx_m00713m/T/___dart-vm___tools_____20160714-4473-1ai2vox.html
  *  External link http://localhost:5000/dart-vm/tools/ failed: 404 No error

[...]

Issue #10, opened 2016-07-14
Spec version is 4, not 3


kwalrath:
We somehow missed this change when creating the new site.

See https://github.com/dart-lang/www.dartlang.org/issues/1554 for more details.

Sfshaza:
Fixed.

PR #5679, opened 2024-04-02
Suggest null-check patterns for type promo

### LINKED ISSUE #5666: Suggest using null-check patterns for working with nullable fields or expressions that can't be promoted
One of the main ways we make nullable types easy to work with is by promoting a nullable typed variable to a non-nullable one when you check it for `null`. Unfortunately, that doesn't work for most fields. We have an [existing guideline that suggests you work around the limitation by copying the field to a local variable first](https://dart.dev/effective-dart/usage#consider-assigning-a-nullable-field-to-a-local-variable-to-enable-type-promotion). Now that we have pattern matching and null-check patterns, another—possibly superior—idiom is to use a null-check pattern. So instead of:

```dart
  String toString() {
    final response = this.response;
    if (response != null) {
      return 'Could not complete upload to ${response.url} '
          '(error code ${response.errorCode}): ${response.reason}.';
    }

    return 'Could not upload (no response).';
  }
```

You can write:

```dart
  String toString() {
    if (this.response case var response?) {
      return 'Could not complete upload to ${response.url} '
          '(error code ${response.errorCode}): ${response.reason}.';
    }

    return 'Could not upload (no response).';
  }
```

I think we should suggest this idiom with a CONSIDER rule. I'm not sure if we should retire the existing rule, merge it with this one, or just have two. I lean towards having two, but I would put using a null-check pattern first.

MaryaBelanger:
Fixes #5666 

**Staged:** https://dart-dev--pr5679-null-check-wujb9sng.web.app/effective-dart/usage#consider-work-arounds-to-enable-type-promotion-for-nullable-fields

### DIFF
--- a/src/content/effective-dart/_toc.md
+++ b/src/content/effective-dart/_toc.md
@@ -105,7 +105,7 @@ the project:
 * <a href='/effective-dart/usage#dont-use-an-explicit-default-value-of-null'>DON'T use an explicit default value of <code>null</code>.</a>
 * <a href='/effective-dart/usage#dont-use-true-or-false-in-equality-operations'>DON'T use <code>true</code> or <code>false</code> in equality operations.</a>
 * <a href='/effective-dart/usage#avoid-late-variables-if-you-need-to-check-whether-they-are-initialized'>AVOID <code>late</code> variables if you need to check whether they are initialized.</a>
-* <a href='/effective-dart/usage#consider-work-arounds-to-enable-type-promotion-for-nullable-fields'>CONSIDER work arounds to enable type promotion for nullable fields.</a>
+* <a href='/effective-dart/usage#consider-type-promotion-or-null-check-patterns-for-using-nullable-types'>CONSIDER type promotion or null-check patterns for using nullable types.</a>
 
 **Strings**
 
--- a/src/content/effective-dart/_toc.md
+++ b/src/content/effective-dart/_toc.md
@@ -105,7 +105,7 @@ $ dart run dart_site effective-dart
 * <a href='/effective-dart/usage#dont-use-an-explicit-default-value-of-null'>DON'T use an explicit default value of <code>null</code>.</a>
 * <a href='/effective-dart/usage#dont-use-true-or-false-in-equality-operations'>DON'T use <code>true</code> or <code>false</code> in equality operations.</a>
 * <a href='/effective-dart/usage#avoid-late-variables-if-you-need-to-check-whether-they-are-initialized'>AVOID <code>late</code> variables if you need to check whether they are initialized.</a>
-* <a href='/effective-dart/usage#consider-assigning-a-nullable-field-to-a-local-variable-to-enable-type-promotion'>CONSIDER assigning a nullable field to a local variable to enable type promotion.</a>
+* <a href='/effective-dart/usage#consider-work-arounds-to-enable-type-promotion-for-nullable-fields'>CONSIDER assigning a nullable field to a local variable to enable type promotion.</a>
 
 **Strings**
 
--- a/src/content/effective-dart/_toc.md
+++ b/src/content/effective-dart/_toc.md
@@ -105,7 +105,7 @@ the project:
 * <a href='/effective-dart/usage#dont-use-an-explicit-default-value-of-null'>DON'T use an explicit default value of <code>null</code>.</a>
 * <a href='/effective-dart/usage#dont-use-true-or-false-in-equality-operations'>DON'T use <code>true</code> or <code>false</code> in equality operations.</a>
 * <a href='/effective-dart/usage#avoid-late-variables-if-you-need-to-check-whether-they-are-initialized'>AVOID <code>late</code> variables if you need to check whether they are initialized.</a>
-* <a href='/effective-dart/usage#consider-work-arounds-to-enable-type-promotion-for-nullable-fields'>CONSIDER assigning a nullable field to a local variable to enable type promotion.</a>
+* <a href='/effective-dart/usage#consider-work-arounds-to-enable-type-promotion-for-nullable-fields'>CONSIDER work arounds to enable type promotion for nullable fields.</a>
 
 **Strings**
 
--- a/src/content/effective-dart/usage.md
+++ b/src/content/effective-dart/usage.md
@@ -340,8 +340,9 @@ Declaring members [private][] and [final][], as we generally recommend, is often
 enough to bypass these limitations. But, that's not always an option.
 
 One pattern to work around type promotion limitations is to use a
-[null-check pattern][] to simultaneously confirm the member's value is not null,
-and bind that value to a new non-nullable variable of the same base type.
+[null-check pattern][]. This simultaneously confirms the member's value
+is not null, and binds that value to a new non-nullable variable of
+the same base type.
 
 <?code-excerpt "usage_good.dart (null-check-promo)"?>
 ```dart tag=good
--- a/src/content/effective-dart/usage.md
+++ b/src/content/effective-dart/usage.md
@@ -325,7 +325,8 @@ Of course, if `null` is a valid initialized value for the variable,
 then it probably does make sense to have a separate boolean field.
 
 
-### CONSIDER assigning a nullable field to a local variable to enable type promotion
+### CONSIDER work arounds to bypass type promotion limitations
+(or, "...to enable type promotion")
 
 Checking that a nullable variable is not equal to `null` promotes the variable
 to a non-nullable type. That lets you access members on the variable and pass it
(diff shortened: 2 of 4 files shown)


src/content/effective-dart/usage.md
```diff
@@ -350,19 +374,27 @@ class UploadException {
 
   @override
   String toString() {
+    // ···
```
MaryaBelanger, line 377: @parlough I know I've asked about this before, so sorry for forgetting! But is there a way around these `// ...` that happen when you end and restart docregions? Also on line 358

parlough: You should be able to add `plaster="none"` to the `code-excerpt` instruction.

MaryaBelanger:
@parlough the effective dart toc isn't updating, something to do with style.md?

```
PathNotFoundException: Cannot open file, path = 'src/effective-dart/style.md' (OS Error: No such file or directory, errno = 2)
#0      _File.throwIfError (dart:io/file_impl.dart:675:7)
#1      _File.openSync (dart:io/file_impl.dart:490:5)
#2      _File.readAsBytesSync (dart:io/file_impl.dart:574:18)
#3      _File.readAsStringSync (dart:io/file_impl.dart:624:18)
#4      _File.readAsLinesSync (dart:io/file_impl.dart:630:36)
#5      _generateToc (package:dart_site/src/commands/generate_effective_dart_toc.dart:52:10)
#6      GenerateEffectiveDartToc.run (package:dart_site/src/commands/generate_effective_dart_toc.dart:36:36)
#7      CommandRunner.runCommand (package:args/command_runner.dart:212:27)
#8      CommandRunner.run.<anonymous closure> (package:args/command_runner.dart:122:25)
#9      new Future.sync (dart:async/future.dart:305:31)
#10     CommandRunner.run (package:args/command_runner.dart:122:14)
#11     main (file:///Users/mbelanger/Projects/dart.dev/dart-3/tool/dart_site/bin/dart_site.dart:23:22)
#12     _delayEntrypointInvocation.<anonymous closure> (dart:isolate-patch/isolate_patch.dart:295:33)
#13     _RawReceivePort._handleMessage (dart:isolate-patch/isolate_patch.dart:184:12)
```

I tried removing the `path-base` command on line 12 of style.md but still got the error (in case that's related)

src/content/effective-dart/usage.md
```diff
@@ -325,7 +326,7 @@ Of course, if `null` is a valid initialized value for the variable,
 then it probably does make sense to have a separate boolean field.
 
 
-### CONSIDER assigning a nullable field to a local variable to enable type promotion
+### CONSIDER work arounds to enable type promotion for nullable fields
```
munificent, line 329: Grammar nit: I think "workaround" is a single word when used as a noun.

More generally, I think the naming of this guideline makes less sense when it includes the null-check pattern because that technically isn't using type promotion. Instead of promoting the type of the variable after it's declared, it binds an entirely new variable whose static type is already the non-nullable base type.

Maybe it's better to split this into two separate guidelines? Something like:

---

**CONSIDER using a null-check pattern to work with a nullable expression**

*null-check pattern example...*

**CONSIDER assigning a nullable expression to a local variable to enable type promotion**

*other example...*

---

I'm saying "expression" instead of "field" here because these patterns are equally applicable to more complex expressions that return nullable values, like:

```dart
if (foo.bar.baz.someString != null) {
  print(foo.bar.baz.someString.length); // Error, can't promote to non-nullable.
}

if (foo.bar.baz.someString case var someString?) {
  print(someString.length); // Works.
}

var someString = foo.bar.baz.someString;
if (someString != null) {
  print(someString.length); // Also works.
}
```

src/content/effective-dart/_toc.md
```diff
@@ -105,7 +105,7 @@ the project:
 * <a href='/effective-dart/usage#dont-use-an-explicit-default-value-of-null'>DON'T use an explicit default value of <code>null</code>.</a>
 * <a href='/effective-dart/usage#dont-use-true-or-false-in-equality-operations'>DON'T use <code>true</code> or <code>false</code> in equality operations.</a>
 * <a href='/effective-dart/usage#avoid-late-variables-if-you-need-to-check-whether-they-are-initialized'>AVOID <code>late</code> variables if you need to check whether they are initialized.</a>
-* <a href='/effective-dart/usage#consider-assigning-a-nullable-field-to-a-local-variable-to-enable-type-promotion'>CONSIDER assigning a nullable field to a local variable to enable type promotion.</a>
+* <a href='/effective-dart/usage#consider-work-arounds-to-enable-type-promotion-for-nullable-fields'>CONSIDER assigning a nullable field to a local variable to enable type promotion.</a>
```
munificent, line 108: This text needs to be updated too.

src/content/effective-dart/usage.md
MaryaBelanger: I did see your suggestion in the issue to split them into two. The reason I kept them together is because both the situation and the outcome are identical. Like, both the local variable and the pattern methods are just two different ways to achieve the same thing when you have the same problem, if that makes sense, and it felt natural to group them based on that common purpose. 

Another title I thought of was "**Consider workarounds to bypass type promotion limitations**" -- is that general enough to keep both strategies together under one heading? 

Or, is the issue _because_ of the generic title, and Effective Dart headings need to be more specific than that? If that's the case then I will split them up.

munificent: > I did see your suggestion in the issue to split them into two.

Oh, heh, I completely forgot I already suggested that. At least I'm consistent with my past self. :)

> The reason I kept them together is because both the situation and the outcome are identical. Like, both the local variable and the pattern methods are just two different ways to achieve the same thing when you have the same problem

SGTM.

> Another title I thought of was "**Consider workarounds to bypass type promotion limitations**" -- is that general enough to keep both strategies together under one heading?

It's not really about type promotion itself, because type promotion is another solution. The problem is really that you can't do much with an expression whose type is nullable. It's about escaping the nullable type shackles.

Maybe we can be more concrete and just mention both solutions in the guideline? Something like "CONSIDER type promotion or null-check patterns for using nullable types"?

MaryaBelanger: > CONSIDER type promotion or null-check patterns for using nullable types

That sounds great! Thanks for working something out with me. Your consistency is appreciated :)

atsansone:
> @parlough the effective dart toc isn't updating, something to do with style.md?
> 
> ```
> PathNotFoundException: Cannot open file, path = 'src/effective-dart/style.md' (OS Error: No such file or directory, errno = 2)
> #0      _File.throwIfError (dart:io/file_impl.dart:675:7)
> #1      _File.openSync (dart:io/file_impl.dart:490:5)
> #2      _File.readAsBytesSync (dart:io/file_impl.dart:574:18)
> #3      _File.readAsStringSync (dart:io/file_impl.dart:624:18)
> #4      _File.readAsLinesSync (dart:io/file_impl.dart:630:36)
> #5      _generateToc (package:dart_site/src/commands/generate_effective_dart_toc.dart:52:10)
> #6      GenerateEffectiveDartToc.run (package:dart_site/src/commands/generate_effective_dart_toc.dart:36:36)
> #7      CommandRunner.runCommand (package:args/command_runner.dart:212:27)
> #8      CommandRunner.run.<anonymous closure> (package:args/command_runner.dart:122:25)
> #9      new Future.sync (dart:async/future.dart:305:31)
> #10     CommandRunner.run (package:args/command_runner.dart:122:14)
> #11     main (file:///Users/mbelanger/Projects/dart.dev/dart-3/tool/dart_site/bin/dart_site.dart:23:22)
> #12     _delayEntrypointInvocation.<anonymous closure> (dart:isolate-patch/isolate_patch.dart:295:33)
> #13     _RawReceivePort._handleMessage (dart:isolate-patch/isolate_patch.dart:184:12)
> ```
> 
> I tried removing the `path-base` command on line 12 of style.md but still got the error (in case that's related)

IIRC, all of the code is under `src/content/<path>`. The [code for this command](https://github.com/dart-lang/site-www/blob/main/tool/dart_site/lib/src/commands/generate_effective_dart_toc.dart#L42) points there, so maybe your local copy doesn't have the latest version of the command? Just a thought.

src/content/effective-dart/usage.md
```diff
@@ -337,7 +338,30 @@ private final fields. Values that are open to manipulation
 
 Declaring members [private][] and [final][], as we generally recommend, is often
 enough to bypass these limitations. But, that's not always an option.
-One pattern to work around this is to assign the field's value
+
+One pattern to work around type promotion limitations is to use a
+[null-check pattern][] to simultaneously confirm the member's value is not null,
+and bind that value to a new non-nullable variable of the same base type.
```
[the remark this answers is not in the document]
MaryaBelanger: - **Condition / instruction**: Isn't it already written in this format except that it starts with "One pattern to \<condition>, \<instruction>"? Which I think usefully indicates that another option will be covered next. 
- **Active voice**: What part of the original isn't in active voice? I wrote "..._use_ a null-check pattern..._bind_ that value...", both active, right?
- **Shorten**: Shortened, but just fyi the second sentence suggestion inaccurately attributes binding the new variable to the user, when it's something that is simultaneously done by the null-check pattern, as I originally wrote.

src/content/effective-dart/usage.md
```diff
@@ -355,14 +379,21 @@ class UploadException {
...
+Be careful when using a local variable. If you need to write back to the field,
+make sure that you don't write back to the local variable instead. (Making the
+local variable [`final`][] can prevent such mistakes.) Also, if the field might
+change while the local is still in scope, then the local might have a stale
+value. 
+
+Sometimes it's best to simply [use `!`][] on the field. 
+In some cases, though, using either a local variable or a null-check pattern 
+can be cleaner and safer than using `!` every time you need to treat the value
+as non-null:
```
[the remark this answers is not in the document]
MaryaBelanger: Going to sip this change as I don't think elongating the text and recapping the points that were just covered in the prior two paragraphs is necessary for such a short section.

munificent (approved):

src/content/effective-dart/usage.md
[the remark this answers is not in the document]
atsansone: - The condition / instruction would come from the fact that two "to X..." writing patterns are there. If you change to "To X, do Y", you remove both the passive voice ("...limitations is to use...") and set the condition on which to act.
- There are two "is" verbs in the one sentence. That makes it all passive as it's unclear who is performing the action.
- Agreed on the last part. Just a suggested way of doing it. Feel free to rework if you want to correct it.

I'm not blocking this, but would encourage you to look at how you can move closer to Google style. Look at the embedded links and see if that helps. 

You can merge if you want, of course.

PR #5680, opened 2024-04-02
Fix Effective Dart TOC script

parlough:
The script didn't account for the updated file location :)


MaryaBelanger (approved):
Thank you 🙏

Issue #5079, opened 2023-07-25
Consider adding a redirect system for language feature specifications similar to Flutter design docs


parlough:
Currently it's a bit awkward to link to feature specifications on the language repository. When they are being worked on, then accepted, then released, their link changes every time. Beyond that, the links are long and definitely not memorable.

Developers often love to see what's coming, learn from the specifications, track them, share them, etc. We also use these links occasionally on dart.dev. 

We should consider tracking redirects that we handle updating, so we can have a persistent link to link to from anywhere, sort of similar to Flutter design docs that live at `flutter.dev/go/`.

This could be done on this repository, but it would probably be better on the language repository's `spec.dart.dev` Firebase instance, which would allow atomic updates of the redirects as the documents move. Then we could add a single redirect from `dart.dev` to `spec.dart.dev` for those design docs.

If we managed this in `dart-lang/language`, the workflow would look like:

1. When adding a design doc:
   1. Add the redirect to `dart-lang/language`'s `firebase.json` under a name `/feature-spec/class-modifiers` or something similar.
   2. Add the dart.dev redirect link to the document at the top of the feature specification itself for easy sharing.
2. When moving it to a different directory, replace the destination URL. 

\cc @MaryaBelanger

mit-mit:
Should we just go with `dart.dev/go/` ? We still have time to change https://github.com/dart-lang/site-www/pull/5684 to `/go/macros`...

PR #5684, opened 2024-04-03
Add /macros redirect to current language spec

kevmoo:
In the (near) future this can be updated to point to a getting started
guide or similar

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -225,6 +225,7 @@
       { "source": "/lints", "destination": "/tools/linter-rules", "type": 301 },
       { "source": "/lints/:lint*", "destination": "/tools/linter-rules/:lint", "type": 301 },
       { "source": "/logos", "destination": "/brand", "type": 301 },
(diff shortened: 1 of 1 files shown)


johnpryan (approved):

parlough (approved):
This is fine as a one-off, but long-term I'd prefer to implement a proper solution for sharing WIP features/specs: https://github.com/dart-lang/site-www/issues/5079

kevmoo:
> This is fine as a one-off, but long-term I'd prefer to implement a proper solution for sharing WIP features/specs rather than filling up the top-level scope: #5079

If this was JUST for the spec, yes, but the theory is that this link will evolve to point to a wiki page for getting started with the preview to the eventual `https://dart.dev/language/macoros` or whatever page we want to have in the end.

parlough:
> If this was JUST for the spec, yes, but the theory is that this link will evolve to point to a wiki page for getting started with the preview to the eventual https://dart.dev/language/macoros or whatever page we want to have in the end.

Sounds good to me :)

PR #5703, opened 2024-04-09
Continue docregion/excerpt cleanup

parlough:
- Migrates away from last unnamed regions
- Close some unclosed regions
- Some other misc cleanup

Contributes to https://github.com/dart-lang/site-www/issues/5593


atsansone (approved):
Thanks @parlough !

PR #5706, opened 2024-04-09
Add replacement suggestion

### LINKED ISSUE #3750: Document suggesting a replacement after a package is discontinued

Page: https://dart.dev/tools/pub/publishing#discontinue

Currently we document how to discontinue a package, but not the fact you can suggest a replacement or how to go about doing so.

For example, [pedantic](https://pub.dev/packages/pedantic) has the `lints` package suggested as a replacement:

image

Within the "Marking packages as discontinued" section we should bring attention to the fact you can suggest a replacement and highlight why it is beneficial/important to do so if possible. 

Extra details in https://github.com/dart-lang/site-www/issues/4782

### LINKED ISSUE #4783: Need for more extensive reference type docs
Related: #4565 

We need a type of documentation, and dedicated place for that type, that allows extensive reference information per topic. 

Currently, the language tour is really the only place to elaborate on a topic, which is counter-intuitive because, as a tour, it's purpose is brevity/"quick-start" info. The library tour is the same, but it can point to the API docs. 

Some topics with critical details that don't belong in the language tour:
- [Class modifiers combinations](https://github.com/dart-lang/site-www/pull/4754#discussion_r1165196287)
- [Patterns types ](https://github.com/dart-lang/language/blob/master/accepted/future-releases/0546-patterns/feature-specification.md#patterns)
- Definitions of patterns terminology ([refutable vs irrefutable](https://dart.dev/resources/glossary#refutable-pattern), outer and inner/sub patterns, [pattern contexts](https://github.com/dart-lang/language/blob/master/accepted/future-releases/0546-patterns/feature-specification.md#pattern-context), etc.)
- (Add more as they come up)

As with any IA rule, this needs to be a really, _really_ well defined doc type, and that definition needs to be strictly enforced. Introducing different documentation types to cover individual topics can become a nightmare of spread and watered-down content; it needs to be very clear for users, without explicit direction, what kind of info they will find in a reference coverage of a topic vs. language tour coverage or tutorial coverage of the same topic.

Keep in mind the [Glossary page](https://github.com/dart-lang/site-www/issues/4622) and its purpose. Overlap needs to be _strongly_ scrutinized as we define what the purpose of a reference page vs a glossary entry is.

atsansone:
Fixes #3750

### DIFF
--- a/src/content/tools/pub/publishing.md
+++ b/src/content/tools/pub/publishing.md
@@ -23,7 +23,7 @@ Watch the following video for an overview of building and publishing packages.
   {{yt.set}}>
 </iframe>
 
-## Publishing is forever
+## Publish is forever
 
 Keep in mind that publishing is forever. As soon as you publish your package,
 users can depend on it. Once they start doing that, removing
@@ -37,16 +37,15 @@ upgrade yet.
 For already published packages that are no longer relevant or being maintained,
 you can [mark them as discontinued](#discontinue).
 
-## Preparing to publish
+## Prepare to publish
 
-When publishing a package, it's important to follow the [pubspec
-format][pubspec] and
+When publishing a package, follow the [pubspec format][pubspec] and
 [package layout conventions][].
-Some of these are required in order for others to be able to use your package.
-Others are suggestions to help make it easier for users to understand and work
-with your package. In both cases, pub tries to help you by pointing out what
-changes will help make your package play nicer with the Dart ecosystem. There
-are a few additional requirements for uploading a package:
+Some of these are required for others to use your package.
+Others are suggestions to simplify concepts other users to understand and work
+with your package. In both cases, pub points out what changes help make your package play nicer with the Dart ecosystem.
+Beyond these conditions, follow these additional requirements
+to upload your package:
 
 * You must include a `LICENSE` file.
   We recommend the [BSD 3-clause license][],
@@ -89,7 +88,6 @@ affect how your package's page looks:
   details about your package on the right side of your package's page, like its
   description, homepage, etc.
 
-
 ### Advantages of using a verified publisher {:#verified-publisher}
 
 You can publish packages using either a verified publisher (recommended)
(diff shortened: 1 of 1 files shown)


src/content/tools/pub/publishing.md
```diff
@@ -37,16 +37,15 @@ upgrade yet.
...
-format][pubspec] and
+When publishing a package, follow the [pubspec format][pubspec] and
 [package layout conventions][].
-Some of these are required in order for others to be able to use your package.
-Others are suggestions to help make it easier for users to understand and work
-with your package. In both cases, pub tries to help you by pointing out what
-changes will help make your package play nicer with the Dart ecosystem. There
-are a few additional requirements for uploading a package:
+Some of these are required for others to use your package.
+Others are suggestions to simplify concepts other users to understand and work
```
bkonyi, line 45: "simplify concepts other users to understand" sounds strange to me. Is this right?

src/content/tools/pub/publishing.md
```diff
@@ -226,57 +233,55 @@ Here's how to transfer a package to a verified publisher:
...
+## Detect supported platforms
 
 The [pub.dev site]({{site.pub}}) detects which platforms a package supports,
 displaying these platforms on the package page.
 Users of pub.dev can filter searches by platform.
 
-To change the automatically generated list of supported platforms,
-[specify supported platforms][] in the pubspec.
+To change the generated list of supported platforms,
+[specify supported platforms][] in the `pubspec/yaml` file.
```
bkonyi, line 276: Nit: `pubspec.yaml` file

src/content/tools/pub/publishing.md
```diff
@@ -23,7 +23,7 @@ Watch the following video for an overview of building and publishing packages.
   {{yt.set}}>
 </iframe>
 
-## Publishing is forever
+## Publish is forever
```
bkonyi, line 26: This sounds strange to me compared to the original. Any particular reason for this change?

bkonyi (commented):
LGTM overall.

src/content/tools/pub/publishing.md
atsansone: Missed how this change would look. We're trying to remove gerunds from headings.

bkonyi (approved):

PR #5707, opened 2024-04-09
Fix 404 page missing from site

parlough:
After https://github.com/dart-lang/site-www/commit/faa3ca3ad31e374f3a6573403c092fea107ac535, the 404 page was generating but being placed in a folder instead of in the root directory, causing it to not be found.

Before fix: https://dart.dev/parker-is-cool
After fix: https://dart-dev--pr5707-fix-404-missing-9rfsj1qx.web.app/parker-is-cool


atsansone (approved):
LGTM

PR #5719, opened 2024-04-12
Updated methods/operators section

### LINKED ISSUE #5634: Add proposed text changes for the paragraph on "operators" on 'Methods' page
### Page URL

https://dart.dev/language/methods/

### Page source

https://github.com/dart-lang/site-www/tree/main/./src/content/language/methods.md

### Describe the problem

At

[Method/Operators](https://dart.dev/language/methods#operators)

We read:

> Operators are instance methods with special names. Dart allows you to define operators with the following names:
> < | + | | | >>> > | / | ^ | [] <= | ~/ | & | []= >= | * | << | ~ - | % | >> | ==

It would be better to list the operators in a list. The above list seems to be missing the BNF `|` at arbitrary places (?).
It is also unclear why the operators come in the order chosen. It is not an ordering by priority or anything else that makes sense to me. 

Compare with page [Operators](https://dart.dev/language/operators)

See also Chapter 10.2.1 (page 37) of the [Dart language specification, 6th edition draft](https://spec.dart.dev/DartLangSpecDraft.pdf) (PDF), which I attach.

-----
Operators from the draft spec
-----

Additionally, it would be cool to show an example of overriding the `[]` operator. That's unusual!

```
class Vector {
  final int x, y;

  Vector(this.x, this.y);

  Vector operator +(Vector v) => Vector(x + v.x, y + v.y);
  Vector operator -(Vector v) => Vector(x - v.x, y - v.y);
  Vector operator -() => Vector(-x,-y);

  @override
  bool operator ==(Object other) =>
      other is Vector && x == other.x && y == other.y;

  @override
  int get hashCode => Object.hash(x, y);

  int operator [](int index) {
    switch (index) {
      case 0: return x;
      case 1: return y;
      default: throw ArgumentError("Index $index is out of range [0,1] for 2-D vector of int");
    }
  }

}


void main() {
  final v = Vector(2, 3);
  final w = Vector(2, 2);

  assert(v + w == Vector(4, 5));
  assert(v - w == Vector(0, 1));

  assert(v[0] == 2 && v[1] == 3);
  assert(-v[0] == -2 && -v[1] == -3);
}
```

### Expected fix

As described above

### Additional context

_No response_

atsansone:
Fixes #5634

### DIFF
--- a/src/content/language/methods.md
+++ b/src/content/language/methods.md
@@ -41,7 +41,7 @@ class Point {
 
 ## Operators
 
-Operators are instance methods with special names.
+Most operators are instance methods with special names.
 Dart allows you to define operators with the following names:
 
 |       |      |      |      |       |      |
@@ -55,11 +55,9 @@ Dart allows you to define operators with the following names:
 
 :::note
 You may have noticed that some [operators][], like `!=`, aren't in
-the list of names. That's because they're just [syntactic sugar][sugar].
-For example, the expression `e1 != e2` is syntactic sugar for `!(e1 == e2)`.
+the list of names. These operators aren't instance methods.
+Their behavior is built in to Dart.
 :::
-
-[sugar]: https://en.wikipedia.org/wiki/Syntactic_sugar
 
 {%- comment %}
   Internal note from https://github.com/dart-lang/site-www/pull/2691#discussion_r506184100:
@@ -67,9 +65,10 @@ For example, the expression `e1 != e2` is syntactic sugar for `!(e1 == e2)`.
   - `!` is probably excluded for historical reasons
 {% endcomment %}
 
-To identify an operator declaration, use the built-in identifier `operator`.
-The following example defines vector 
-addition (`+`), subtraction (`-`), and equality (`==`):
+To declare an operator declaration, use the built-in identifier
+`operator` then the operator you are defining.
+The following example defines vector addition (`+`), subtraction (`-`),
+and equality (`==`):
 
 <?code-excerpt "misc/lib/language_tour/classes/vector.dart"?>
 ```dart
--- a/src/content/language/methods.md
+++ b/src/content/language/methods.md
@@ -65,7 +65,7 @@ Their behavior is built in to Dart.
   - `!` is probably excluded for historical reasons
 {% endcomment %}
 
-To declare an operator declaration, use the built-in identifier
+To declare an operator, use the built-in identifier
 `operator` then the operator you are defining.
 The following example defines vector addition (`+`), subtraction (`-`),
 and equality (`==`):
--- a/src/content/language/methods.md
+++ b/src/content/language/methods.md
@@ -44,21 +44,22 @@ class Point {
 Operators are instance methods with special names.
 Dart allows you to define operators with the following names:
 
-|      |      |      |       |
-|------|------|------|-------|
-| `<`  | `+`  | `\|` | `>>>` |
-| `>`  | `/`  | `^`  | `[]`  |
-| `<=` | `~/` | `&`  | `[]=` |
-| `>=` | `*`  | `<<` | `~`   |
-| `-`  | `%`  | `>>` | `==`  |
+|       |      |      |      |       |      |
+|-------|------|------|------|-------|------|
+| `<`   | `>`  | `<=` | `>=` | `==`  | `~`  |
+| `-`   | `+`  | `/`  | `~/` | `*`   | `%`  |
+| `\|`  | `ˆ`  | `&`  | `<<` | `>>>` | `>>` |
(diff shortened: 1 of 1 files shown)


src/content/language/methods.md
```diff
@@ -44,29 +44,30 @@ class Point {
 Operators are instance methods with special names.
```
munificent, line 44: This isn't true for all operators in Dart. (`.`, `?.`, `..`, `||`, and `&&` aren't instance methods, for example.)

Maybe add "Most" to this sentence while you're inhere?

src/content/language/methods.md
```diff
@@ -44,29 +44,30 @@ class Point {
...
+| `\|`  | `ˆ`  | `&`  | `<<` | `>>>` | `>>` |
+| `[]=` | `[]` |      |      |       |      |
 
 {:.table}
 
 :::note
 You may have noticed that some [operators][], like `!=`, aren't in
-the list of names. That's because they're just syntactic sugar. For example,
-the expression `e1 != e2` is syntactic sugar for `!(e1 == e2)`.
+the list of names. That's because they're just [syntactic sugar][sugar].
```
munificent, line 58: I don't know if it's really accurate (or useful) to characterize all non-instance method operators as syntactic sugar. Maybe just say some operators are not instance methods and instead have behavior that is built directly into the language?

src/content/language/methods.md
```diff
@@ -44,29 +44,30 @@ class Point {
...
+[sugar]: https://en.wikipedia.org/wiki/Syntactic_sugar
+
 {%- comment %}
   Internal note from https://github.com/dart-lang/site-www/pull/2691#discussion_r506184100:
   -  `??`, `&&` and `||` are excluded because they are lazy / short-circuiting operators
   - `!` is probably excluded for historical reasons
 {% endcomment %}
 
-An operator declaration is identified using the built-in identifier `operator`.
+To identify an operator declaration, use the built-in identifier `operator`.
```
munificent, line 70: I'm not sure what "identify" is doing in this sentence. Maybe: "To declare an operator, use the built-in identifier `operator` followed by the operator you are defining."

MaryaBelanger: +1

src/content/language/methods.md
MaryaBelanger: +1

src/content/language/methods.md
```diff
@@ -44,29 +44,30 @@ class Point {
...
-| `>`  | `/`  | `^`  | `[]`  |
-| `<=` | `~/` | `&`  | `[]=` |
-| `>=` | `*`  | `<<` | `~`   |
-| `-`  | `%`  | `>>` | `==`  |
+|       |      |      |      |       |      |
+|-------|------|------|------|-------|------|
+| `<`   | `>`  | `<=` | `>=` | `==`  | `~`  |
+| `-`   | `+`  | `/`  | `~/` | `*`   | `%`  |
+| `\|`  | `ˆ`  | `&`  | `<<` | `>>>` | `>>` |
+| `[]=` | `[]` |      |      |       |      |
```
MaryaBelanger, line 52: Why not the original dimensions? There's 20, 4x5 seems like the obvious choice

src/content/language/methods.md
MaryaBelanger: Agree. This is the operators section _on the methods page_. It should be clear that this info doesn't apply to _all_ operators, and this note feels like a good place to call that out (more so than the syntactic sugar comment)

MaryaBelanger (commented):
Consider opening another issue to track the second part of #5634 , showing an example of overriding the `[]` operator.

src/content/language/methods.md
atsansone: Less space, honestly. Looked pretty strange for 4 columns across a broad page. I updated the order to the spec. It's not a big deal to me either way.

MaryaBelanger: Ok I'm fine with it, just checking

src/content/language/methods.md
```diff
@@ -66,9 +65,10 @@ the expression `e1 != e2` is syntactic sugar for `!(e1 == e2)`.
   - `!` is probably excluded for historical reasons
 {% endcomment %}
 
-An operator declaration is identified using the built-in identifier `operator`.
-The following example defines vector 
-addition (`+`), subtraction (`-`), and equality (`==`):
+To declare an operator declaration, use the built-in identifier
```
MaryaBelanger, line 68: Just to not say "declare" twice in a row like that:

```suggestion
To write an operator declaration, use the built-in identifier
```

MaryaBelanger (approved):
One more minor change, but that's all. Thanks for working in the feedback!

PR #5720, opened 2024-04-13
Update and prune pnpm dependencies

parlough:
Many dependencies were able to be removed due to some downstream dependencies no longer relying on polyfills. Results in a slightly quicker dependency retrieval and fresh build :)


atsansone (approved):
LGTM. Thanks!

PR #5725, opened 2024-04-16
Update isolates.md: Robust Example Fix

maryamshal:
Fix robust example inconsistency. The constructor signature changes order of parameters mid example which causes issues for a person following along step by step.


MaryaBelanger:
Hi @maryamshal, thanks for fixing this! Could you also update the code excerpt source for each of these changes (that's why the build is failing, both the markdown page and the source for example code need to be updated).

You can find the source of a code block right above it usually, indicated by the `<?code excerpt ...` directive. So on lines [546](https://github.com/dart-lang/site-www/pull/5725/files#diff-2f75b394563841310a6aca2698a5f102497102269dba193a21565e2a34451893R546) and [664](https://github.com/dart-lang/site-www/pull/5725/files#diff-2f75b394563841310a6aca2698a5f102497102269dba193a21565e2a34451893R664) you can see that the source corresponding to the changes you made are in:

- `examples/concurrency/lib/robust_ports_example/start.dart` [line 21](https://github.com/dart-lang/site-www/blob/main/examples/concurrency/lib/robust_ports_example/start.dart#L21), and
- `examples/concurrency/lib/robust_ports_example/spawn_2.dart` [line 33](https://github.com/dart-lang/site-www/blob/main/examples/concurrency/lib/robust_ports_example/spawn_2.dart#L33)

Make the same changes there and we should be all set! (might help to follow the [instructions](https://github.com/dart-lang/site-www?tab=readme-ov-file#refresh-code-excerpts) in the readme too, but it might not be necessary). 

If you're having trouble just let us know and we can take over that part for you, Thanks!

parlough (approved):
Thanks for catching this and opening the PR! Having the order be consistent makes sense and will definitely help readers :)

I went ahead and updated the source `.dart` files for these snippets as well.

Issue #5732, opened 2024-04-18
`./dash_site serve` error


MaryaBelanger:
### Describe the problem

I'm not sure how to interpret this. I'm pretty sure it's not my local edits, as I've just been writing plain text in a single markdown file...

Ran `dart pub get` and `pnpm install` and then tried again but it didn't change anything.

```
Building package executable... 
Built dart_site:dart_site.
[11ty] Problem writing Eleventy templates: (more in DEBUG output)
[11ty] 1. Having trouble rendering liquid template ./src/content/language/macros.md (via TemplateContentRenderError)
[11ty] 2. Cannot read properties of null (reading 'length') (via TypeError)
[11ty] 
[11ty] Original error stack trace: TypeError: Cannot read properties of null (reading 'length')
[11ty]     at markdown.renderer.rules.fence (file:///Users/mbelanger/Projects/dart.dev/dart-3/src/_11ty/plugins/highlight.js:49:37)
[11ty]     at Renderer.render (file:///Users/mbelanger/Projects/dart.dev/dart-3/node_modules/.pnpm/markdown-it@14.1.0/node_modules/markdown-it/lib/renderer.mjs:313:28)
[11ty]     at MarkdownIt.render (file:///Users/mbelanger/Projects/dart.dev/dart-3/node_modules/.pnpm/markdown-it@14.1.0/node_modules/markdown-it/lib/index.mjs:530:24)
[11ty]     at file:///Users/mbelanger/Projects/dart.dev/dart-3/node_modules/.pnpm/@11ty+eleventy@3.0.0-alpha.5/node_modules/@11ty/eleventy/src/Engines/Markdown.js:79:33
[11ty]     at async Template._render (file:///Users/mbelanger/Projects/dart.dev/dart-3/node_modules/.pnpm/@11ty+eleventy@3.0.0-alpha.5/node_modules/@11ty/eleventy/src/TemplateContent.js:532:19)
[11ty]     at async Template.renderPageEntryWithoutLayout (file:///Users/mbelanger/Projects/dart.dev/dart-3/node_modules/.pnpm/@11ty+eleventy@3.0.0-alpha.5/node_modules/@11ty/eleventy/src/Template.js:432:25)
[...]

PR #5743, opened 2024-04-22
Provide more details on error when Markdown code block is misformatted

parlough:
@MaryaBelanger I believe https://github.com/dart-lang/site-www/issues/5732 is caused by a code block without a language specified in your local branch in the `/src/content/language/macros.md` file. This PR makes that more obvious by providing an error explaining that.

I'll follow-up with a better error message, but wanted to make sure this actually was the issue you're facing. I'm also open to changing the requirement of a language if that's helpful or important to you. Thanks!

### DIFF
--- a/src/_11ty/plugins/highlight.js
+++ b/src/_11ty/plugins/highlight.js
@@ -46,6 +46,12 @@ export async function configureHighlighting(markdown) {
     const token = tokens[index];
 
     const splitTokenInfo = token.info.match(/(\S+)\s?(.*?)$/m);
+
+    if (!splitTokenInfo) {
+      throw new Error('Each Markdown code block should specify a language ' +
+          'after the opening backticks like: ```dart.');
+    }
+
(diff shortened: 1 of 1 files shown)


MaryaBelanger (approved):
Yes that was it! I think the error message reads perfectly, thanks for figuring it out.

How did you figure that out by the way? Reading the error log didn't seem to point to anything concrete

parlough:
@MaryaBelanger I determined it roughly by:

I looked at the lines at the beginning of the stack trace:

```
[11ty] Original error stack trace: TypeError: Cannot read properties of null (reading 'length')
[11ty]     at markdown.renderer.rules.fence (file:///Users/mbelanger/Projects/dart.dev/dart-3/src/_11ty/plugins/highlight.js:49:37)
```

This line tried to use which was the result of `token.info.match(/(\S+)\s?(.*?)$/m);` (stored in `splitTokenInfo`) when parsing a Markdown code block(fence). So `splitTokenInfo` must have been `null`, meaning that regex didn't find any matches when parsing the opening fence line.

PR #5747, opened 2024-04-23
Adds footnote to Keywords page

### LINKED ISSUE #5715: Add footnote to 'Keywords' page
### Page URL

https://dart.dev/language/keywords/

### Page source

https://github.com/dart-lang/site-www/tree/main/src/content/language/keywords.md

### Describe the problem

I'm not quite sure how this page is funneled through the deployment machinery, so maybe an update is already in the pipeline.

However, I noticed on the current version of this page that some keywords should have a superscript number (let's use "3", which is the first unused number). It would refer to a footnote along the lines of "³ There are no restrictions on the usage of this keyword as an identifier". However, none of those things are shown.

The point is that words like `async` and `while` are shown in a way that makes them look equivalent, but the former can be the name of a type or an import prefix or a variable (or indeed anything that has a name) whereas the latter can never be used as the declared name of anything. So, quite a big difference.

### Expected fix

I don't understand the mechanics of translating `type: bit` in `src/_data/keywords.yml` into a footnote marker like `{{bii}}` which is used in `src/content/language/keywords.md`, but I'd assume that a new type should be recognized, `type: nothing`, and it should be translated to some new marker, say `{{not}}`, and the new footnote would then use that marker.

The relevant words in `src/_data/keywords.yml` already have the right type, so there should not be a need to change that file.

### Additional context

_No response_

atsansone:
Fixes #5715

### DIFF
--- a/src/_data/keywords.yml
+++ b/src/_data/keywords.yml
@@ -9,13 +9,13 @@
   type: reserved
 - term: 'async'
   link: /language/async
-  type: nothing
+  type: unrestricted
 - term: 'await'
   link: /language/async
   type: context
 - term: 'base'
   link: /language/class-modifiers#base
-  type: nothing
+  type: unrestricted
 - term: 'break'
   link: /language/loops#break-and-continue
   type: reserved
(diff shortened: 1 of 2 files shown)


MaryaBelanger (approved):
LGTM

PR #5754, opened 2024-04-25
fix text display inconsistency

ZBrushIsNotAvailable:
Fixes markdown
image


parlough (approved):
Nice catch and fix, thanks so much!

PR #5757, opened 2024-04-26
Refactor Constructors page

### LINKED ISSUE #2009: Add Documentation on redirecting factory constructor
The Language spec defines implementation for the following syntax:
`factory Listenable.merge(List<Listenable> listenables) = _MergingListenable`

This can be found in the Dart Language Specification, but not in the language tour. This syntax could be made more visible.

Related: https://github.com/flutter/flutter/issues/35727

### LINKED ISSUE #2221: Give example of using constructor variable initialization with named parameters
Language tour have many example of classes and their initialization, but none of them includes named arguments that also initialize. Would be nice have something like:

```dart
class Point {
  num x, y;
  Point({@required this.x, @required this.y});
}
```

As when you have multiple arguments with the same type it's usually better to make them named to avoid mistakes / make the callsite easier to read. There is also an edge case here when you want to assign to a private variable, which requires to do that:

```dart
class Point {
  num _x, _y;
  Point({@required num x, @required num y}) : _x = x, _y = y;
}
```

I know from code reviews that newcomers struggle to figure it out initially.

### LINKED ISSUE #5227: Fix description for constructor's super parameters
I was reading through the documentation for constructors, when I arrived here: https://dart.dev/language/constructors#super-parameters

The description says:

> Super-initializer parameters cannot be positional if the super-constructor invocation already has positional arguments, but they can always be named:

And then the code is as follow:

```
class Vector2d {
  // ...

  Vector2d.named({required this.x, required this.y});
}

class Vector3d extends Vector2d {
  // ...

  // Forward the y parameter to the named super constructor like:
  // Vector3d.yzPlane({required double y, required this.z})
  //       : super.named(x: 0, y: y);
  Vector3d.yzPlane({required super.y, required this.z}) : super.named(x: 0);
}

```

The description talks about positional arguments, and the shown example below has named arguments instead. Which one is it?

### LINKED ISSUE #5506: Update outdated mention that factory constructors can return `null`

### LINKED ISSUE #5728: Clarify 'Constructors' warning & example
### Page URL

https://dart.dev/language/constructors/

### Page source

https://github.com/dart-lang/site-www/tree/main/src/content/language/constructors.md

### Describe the problem

# Confusing warning

At 

https://dart.dev/language/constructors

we find the following highlighted remark:

> Use `this` only when there is a name conflict. Otherwise, Dart style omits the this.

For clarity, one should add that:

> For the generative constructor however, you *must* prefix the _initializing formal parameter_ name with `this.`.

# Confusing initialization in example code

Example code is given as:

~~~
class Point {
  double x = 0;
  double y = 0;

  // Generative constructor with initializing formal parameters:
  Point(this.x, this.y);
}
~~~

However, the explicit initialization of `x` and `y` to 0 is confusing, as "it does nothing". The values are immediately overwritten by the generative constructor.

Maybe one should show the two (or three) possibilities:

Example with initialization determined directly at the line of field definition:

~~~
class PointA {
  double x = 1.0;
  double y = 2.0;

  // The parameterless constructor is not even be needed to set to (1.0,2.0)
  PointA();

  @override
  String toString() {
    return "PointA($x,$y)";
  }
}
~~~

Example with mandatory initialization via constructor:

~~~
class PointB {
  double x; // must be set in constructor
  double y; // must be set in constructor

  // Generative constructor with initializing formal parameters
  PointB(this.x, this.y);

  @override
  String toString() {
    return "PointB($x,$y)";
  }
}
~~~

Constructor which may or may not set to non-null value:

~~~
class PointC {
  double? x; // null if not set in constructor
  double? y; // null if not set in constructor

  // Generative constructor with initializing formal parameters
  PointC(this.x, this.y);

  @override
  String toString() {
    return "PointC($x,$y)";
  }
}
~~~

And thus:

~~~
void main() {
  print(PointA());
  print(PointB(2.4,4.7));
  print(PointC(null,6.67));
}
~~~

Result:

~~~
PointA(1.0,2.0)
PointB(2.4,4.7)
PointC(null,6.67)
~~~

### Expected fix

_No response_

### Additional context

_No response_

### LINKED ISSUE #5738: Fix potential contradiction on 'Constructors' page
### Page URL

https://dart.dev/language/constructors/

### Page source

https://github.com/dart-lang/site-www/tree/main/src/content/language/constructors.md

### Describe the problem

These two sentences appear to be contradictory:
"Subclasses don't inherit constructors from their superclass."
"By default, a constructor in a subclass calls the superclass's unnamed, no-argument constructor."

Surely the subclass must inherit at least the no-name, no-arg constructor in order to call it?

### Expected fix

_No response_

### Additional context

_No response_

### LINKED ISSUE #5739: Clarify copy around factory on 'Constructors' page
### Page URL

https://dart.dev/language/constructors/

### Page source

https://github.com/dart-lang/site-www/tree/main/src/content/language/constructors.md

### Describe the problem

At 

*Factory constructors*, https://dart.dev/language/constructors#factory-constructors

we read:

> Use the `factory` keyword when implementing a constructor that doesn't always create a new instance of its class. For example, a factory constructor might return an instance from a cache, or it might return an instance of a subtype. Another use case for factory constructors is initializing a final variable using logic that can't be handled in the initializer list.

That was confusing to me. Suggesting:

> Use the `factory` keyword when implementing a constructor that doesn't always create a new instance of its class. Although a factory constructor cannot return `null`, it might return an existing instance from a cache instead of creating a new one, or it might return a new instance of a subtype. Another use case for a factory constructor is performing non-trivial work prior to instance construction, like checking arguments or doing any other processing that cannot be handled in the initializer list.

### Expected fix

_No response_

### Additional context

_No response_

atsansone:
Fixes #2009 
Fixes #2221
Fixes #5227 
Fixes #5506
Fixes #5728
Fixes #5738
Fixes #5739

### DIFF
--- a/src/content/language/constructors.md
+++ b/src/content/language/constructors.md
@@ -10,15 +10,182 @@ nextpage:
   title: Methods
 ---
 
+Constructors create instances of classes.
+
+* To declare a constructor, create a function named the same as its class.
+  You can add an optional additional identifier as described in
+  [Named constructors](#named-constructors).
+
+* To instantiate a class, use the
+  [generative constructor](#generative-constructors).
+
+* To instantiate any instance variables,
+  [initialize formal parameters](#parameter-initialization).
+
 <?code-excerpt replace="/ *\/\/\s+ignore_for_file:[^\n]+\n//g; /(^|\n) *\/\/\s+ignore:[^\n]+\n/$1/g; /(\n[^\n]+) *\/\/\s+ignore:[^\n]+\n/$1\n/g; / *\/\/\s+ignore:[^\n]+//g; /([A-Z]\w*)\d\b/$1/g"?>
 
-Declare a constructor by creating a function with the same name as its
-class (plus, optionally, an additional identifier as described in
-[Named constructors](#named-constructors)). 
+## Constructor inheritance
 
-Use the most common constructor, the generative constructor, to create a new
-instance of a class, and [initializing formal parameters](#initializing-formal-parameters)
-to instantiate any instance variables, if necessary:
+Subclasses don't inherit constructors from their superclass.
+Parameters of a superclass can be inherited.
+These are called [Super parameters](#super-parameters)
+A subclass without constructor declarations can only use
+a [default constructor](#default-constructors).
+
+## Parameter initialization
+
+Dart can initialize parameters in a constructor in three ways.
+
+### Initialize parameters when declaring variables
+
+Initialize the constructor parameters when you declare variables.
+
+```dart
+class PointA {
+  double x = 1.0;
+  double y = 2.0;
+
+  // The parameterless constructor is not even be needed to set to (1.0,2.0)
+  PointA();
+
+  @override
+  String toString() {
+    return "PointA($x,$y)";
+  }
+}
+```
+
+### Use initializing formal parameters
+
+To simplify the common pattern of assigning a constructor argument
+to an instance variable,
+Dart has *initializing formal parameters*.
+
+In the constructor declaration, include `this.<propertyName>`
+and omit the body. The `this` keyword refers to the current instance.
+
+When the name conflict exists, use `this`.
+Otherwise, Dart style omits the `this`.
+An exception exists for the generative constructor:
+you must prefix the initializing formal parameter name with `this`.
+
+Certain constructors and parts of constructors can't access `this`:
+
+* Factory constructors
+* The right-hand side of an initializer list
+* Arguments to a superclass constructor
+
+Initializing formal parameters also allow you to initialize
+non-nullable or `final` instance variables.
+Both of these types of variables require initialization or a default value.
+
+```dart
+class PointB {
+  double x; // must be set in constructor
+  double y; // must be set in constructor
+
+  // Generative constructor with initializing formal parameters
+  PointB(this.x, this.y);
+
+  @override
+  String toString() {
+    return "PointB($x,$y)";
+  }
+}
+```
+
+All variables introduced from initializing formal parameters are both
+final and only in scope of the initialized variables.
+
+To perform logic that you can't express in the initializer list,
+create a [factory constructor](#factory-constructors)
+or [static method][] with that logic.
+You can then pass the computed values to a normal constructor.
+
+The constructor parameters could be set as nullable and not be initialized.
+
+```dart
+class PointC {
+  double? x; // null if not set in constructor
+  double? y; // null if not set in constructor
+
+  // Generative constructor with initializing formal parameters
+  PointC(this.x, this.y);
+
+  @override
+  String toString() {
+    return "PointC($x,$y)";
+  }
+}
+```
+
+### Use an initializer list
+
+Before the constructor body runs, you can initialize instance variables.
+Separate initializers with commas.
+
+<?code-excerpt "misc/lib/language_tour/classes/point_alt.dart (initializer-list)"?>
+```dart
+// Initializer list sets instance variables before
+// the constructor body runs.
+Point.fromJson(Map<String, double> json)
+    : x = json['x']!,
+      y = json['y']! {
+  print('In Point.fromJson(): ($x, $y)');
+}
+```
+
+:::warning
+The right-hand side of an initializer list can't access `this`.
+:::
+
+To validate inputs during development,
+use `assert` in the initializer list.
+
+<?code-excerpt "misc/lib/language_tour/classes/point_alt.dart (initializer-list-with-assert)" replace="/assert\(.*?\)/[!$&!]/g"?>
+```dart
+Point.withAssert(this.x, this.y) : [!assert(x >= 0)!] {
+  print('In Point.withAssert(): ($x, $y)');
+}
+```
+
+Initializer lists help set up `final` fields.
+
+The following example initializes three `final` fields in an initializer list.
+To execute the code, click **Run**.
+
+<?code-excerpt "misc/lib/language_tour/classes/point_with_distance_field.dart"?>
+```dart:run-dartpad:height-340px:ga_id-initializer_list
+import 'dart:math';
+
+class Point {
+  final double x;
+  final double y;
+  final double distanceFromOrigin;
+
+  Point(double x, double y)
+      : x = x,
+        y = y,
+        distanceFromOrigin = sqrt(x * x + y * y);
+}
+
+void main() {
+  var p = Point(2, 3);
+  print(p.distanceFromOrigin);
+}
+```
+
+## Types of constructors
+
+### Default constructors
+
+If you don't declare a constructor, a default constructor is provided.
+The default constructor has no arguments and invokes the
+no-argument constructor in the superclass.
+
+### Generative constructors
+
+To instantiate a class, use the generative constructor.
 
 <?code-excerpt "misc/lib/language_tour/classes/point_alt.dart (idiomatic-constructor)" plaster="none"?>
 ```dart
@@ -31,62 +198,7 @@ class Point {
 }
 ```
 
-The `this` keyword refers to the current instance.
-
-:::note
-Use `this` only when there is a name conflict. 
-Otherwise, Dart style omits the `this`.
-:::
-
-
-## Initializing formal parameters
-
-Dart has *initializing formal parameters* to simplify the common pattern of
-assigning a constructor argument to an instance variable. 
-Use `this.propertyName` directly in the constructor declaration,
-and omit the body. 
-
-Initializing parameters also allow you to initialize
-non-nullable or `final` instance variables,
-which both must be initialized or provided a default value:
-
-<?code-excerpt "misc/lib/language_tour/classes/point.dart (constructor-initializer)" plaster="none"?>
-```dart
-class Point {
-  final double x;
-  final double y;
-
-  // Sets the x and y instance variables
-  // before the constructor body runs.
-  Point(this.x, this.y);
-}
-```
-
-The variables introduced by the initializing formals
-are implicitly final and only in scope of the
-[initializer list](/language/constructors#initializer-list).
-
-If you need to perform some logic that cannot be expressed in the initializer list,
-create a [factory constructor](#factory-constructors) 
-(or [static method][]) with that logic
-and then pass the computed values to a normal constructor.
-
-
-## Default constructors
-
-If you don't declare a constructor, a default constructor is provided
-for you. The default constructor has no arguments and invokes the
-no-argument constructor in the superclass.
-
-
-## Constructors aren't inherited
-
-Subclasses don't inherit constructors from their superclass. A subclass
-that declares no constructors has only the default (no argument, no
-name) constructor.
-
-
-## Named constructors
+### Named constructors
 
 Use a named constructor to implement multiple constructors for a class
 or to provide extra clarity:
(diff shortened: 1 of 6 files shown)


src/content/language/constructors.md
```diff
@@ -10,88 +10,68 @@ nextpage:
...
-assigning a constructor argument to an instance variable. 
-Use `this.propertyName` directly in the constructor declaration,
-and omit the body. 
+### Default constructors
 
-Initializing parameters also allow you to initialize
-non-nullable or `final` instance variables,
-which both must be initialized or provided a default value:
+If you don't declare a constructor, Dart uses the default constructor.
+The default constructor has no arguments and no name.
```
natebosch, line 52: [optional] I don't know for sure if the detail of "generative constructor" is specifically useful to understand the concept of the default constructor - but knowing that the default constructor is a generative constructor helps to understand the concept of generative constructors in my opinion. I think since this flows into the next section, it could be worth adding this detail. What do you think?

```suggestion
The default constructor is a generative constructor with no arguments and no name.
```

src/content/language/constructors.md
```diff
@@ -10,88 +10,68 @@ nextpage:
...
 
-<?code-excerpt "misc/lib/language_tour/classes/point.dart (constructor-initializer)" plaster="none"?>
+### Generative constructors
+
+To instantiate a class, use the generative constructor.
```
natebosch, line 56: [nit] It's possible for a class to have more than one generative constructor.

```suggestion
To instantiate a class, use a generative constructor.
```

src/content/language/constructors.md
```diff
@@ -10,88 +10,68 @@ nextpage:
...
+
+<?code-excerpt "point_alt.dart (idiomatic-constructor)" plaster="none"?>
 ```dart
 class Point {
-  final double x;
-  final double y;
+  double x = 0;
+  double y = 0;
```
natebosch, line 62: [nit] I don't think it is particularly useful to show these initializing expressions (nor making the fields mutable). I'd keep these two lines as they were.

src/content/language/constructors.md
```diff
@@ -10,88 +10,68 @@ nextpage:
...
-Otherwise, Dart style omits the `this`.
-:::
+## Types of constructors
+
+### Constructor inheritance
+
+_Subclasses_, or child classes, don't inherit *constructors*
+from their _superclass_, or immediate parent class.
+If a class doesn't declare a constructor, it can only use the
+[default constructor](#default-constructors).
```
natebosch, line 36: [nit] I don't think this sentence is necessary here.

```suggestion
```

src/content/language/constructors.md
```diff
@@ -10,88 +10,68 @@ nextpage:
...
+These are called [super parameters](#super-parameters)
+A subclass without constructor declarations can only use
+a [default constructor](#default-constructors).
 
-## Initializing formal parameters
+Constructors work in a somewhat similar way to
+how you call a chain of static methods.
+Each subclass can call its superclass's constructor to initialize an instance,
+like a subclass can call a superclass's static method.
+This process doesn't involve any "inheritance".
```
natebosch, line 47: [optional] I think it might be more clear to add some specificity.

```suggestion
This process doesn't involve any "inheritance" of constructor bodies or signatures.
```

src/content/language/constructors.md
```diff
@@ -111,34 +91,28 @@ class Point {
...
-The superclass's constructor is called at the beginning of the
-constructor body. If an [initializer list](#initializer-list)
-is also being used, it executes before the superclass is called.
-In summary, the order of execution is as follows:
-
-1. initializer list
-1. superclass's no-arg constructor
+1. [initializer list](#use-an-initializer-list)
+1. superclass's unnamed, no-arg constructor
 1. main class's no-arg constructor
```
natebosch, line 104: This bullet point feels out of place - that's not a new problem in this PR though.

I don't know why we need to talk specifically about the arguments to the main class constructor - this description applies to any constructor definition that doesn't have a super constructor invocation.

I was trying to find a suggestion for a rewrite of this section and it struck me that waiting until the end of the doc to show what an "initializer list" is may be hurting us.

I wonder if we should try to show syntax and terminology near the top of this doc. If a reader knows that the "initializer list" is "the part with `fieldName = parameterName` because they saw that syntax labeled, I think they would have an easier time understanding what it means for the "initializer list" to run in a given position in the instantiation process, even if they don't know the details that will be introduced later.

src/content/language/constructors.md
```diff
@@ -211,20 +196,34 @@ class Vector3d extends Vector2d {
...
-Super-initializer parameters cannot be positional 
-if the super-constructor invocation already has positional arguments,
-but they can always be named:
+To further illustrate, consider the following example.
 
-<?code-excerpt "misc/lib/language_tour/classes/super_initializer_parameters.dart (named)" plaster="none"?>
+```dart
+  // If you invoke the super constructor (`super(0)`) with any
+  // positional arguments, using a super parameter (`super.x`)
+  // returns an error.
```
natebosch, line 204: [nit] I don't know what "returns" would mean in this context. How about "is an error" or "results in an error"?

```suggestion
  // is an error.
```

src/content/language/constructors.md
```diff
@@ -233,79 +232,34 @@ class Vector3d extends Vector2d {
...
+  final double x, y;
 
-void main() {
-  var p = Point(2, 3);
-  print(p.distanceFromOrigin);
+  const ImmutablePoint(this.x, this.y);
 }
 ```
 
+Constant constructors don't always create constants.
```
natebosch, line 253: [optional]

```suggestion
Constant constructors may be invoked in a non-const context and don't always create constants.
```

src/content/language/constructors.md
```diff
@@ -318,50 +272,34 @@ class Point {
...
-  const ImmutablePoint(this.x, this.y);
-}
-```
-
-Constant constructors don't always create constants.
-For details, see the section on
-[using constructors][].
-
+When encountering one of two cases of implementing a constructor,
+use the `factory` keyword:
```
natebosch, line 278: [nit] Is this missing a word at the end? `use the factory keyword when:` ? `use the factory keyword if:` ?

src/content/language/constructors.md
```diff
@@ -402,6 +340,211 @@ var logMap = {'name': 'UI'};
...
+  final double y;
+
+  // Sets the x and y instance variables
+  // before the constructor body runs.
+  PointB(this.x, this.y);
+
+  // Initializing formal parameters can also be optional.
+  PointB.optional([this.x = 0.0, this.y = 0.0]);
+  PointB.named({required this.x, required this.y});
+}
```
natebosch, line 423: With the current layout of the doc the last line is covered well in `PointC`- I think it flows well now to omit the mention here. The optional positional parameters are still useful to show here.

```suggestion
  // Initializing formal parameters can also be optional.
  PointB.optional([this.x = 0.0, this.y = 0.0]);
}
```

src/content/language/constructors.md
```diff
@@ -402,6 +340,211 @@ var logMap = {'name': 'UI'};
...
+
+  PointB.namedPrivate({required double x, required double y})
+      : _x = x,
+        _y = y;
+
+// ...
+}
+```
+
+This also works with named variables.
```
natebosch, line 444: I think we should mention named variables before private fields. WDYT about flipping them - move lines 426-442 to after the example currently ending on line 464

natebosch (commented):
I have a few re-orderings suggested. I _think_ these should make the doc flow fairly well, but I'm having a hard time landing on a clearly optimal way to lay it out since so many of the details are intermingled.

natebosch:
Do we have the capability to create a link to another header _inside_ a code block? (I expect the answer is no)

I think having some kind of syntax "glossary" early in the doc could help?

```dart
class Super {
  final String field1; // Instance variable declaration
  final String field2 =
      someExpression(); // Instance variable declaration with initializing expression

  Super() // Unnamed generative constructor
      : field1 = 'default'; // Initializer list

  Super.named(String argument) // Named generative constructor
      : field1 = argument;

  Super.initializingFormal(this.field1); // Initializing formal parameter

  factory Super.factory() => Super.named('from factory'); // Factory constructor

  Super.redirecting() // Redirecting constructor
      : this.named('from redirecting');
}

class Foo extends Super {
  Foo()
      : super.named('foo unnamed constructor'); // Super constructor invocation
  Foo.named(super.argument)
      // Super parameter
      : super.named();
}
```

src/content/language/constructors.md
```diff
@@ -10,88 +10,68 @@ nextpage:
...
+* To instantiate a class, use the
+  [generative constructor](#generative-constructors).
 
-<?code-excerpt "misc/lib/language_tour/classes/point_alt.dart (idiomatic-constructor)" plaster="none"?>
-```dart
-class Point {
-  double x = 0;
-  double y = 0;
+* To instantiate any instance variables,
+  [initialize formal parameters](#parameter-initialization).
```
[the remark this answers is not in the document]
atsansone: Refactored with explanatory list.

src/content/language/constructors.md
```diff
@@ -10,88 +10,68 @@ nextpage:
...
-The `this` keyword refers to the current instance.
+<?code-excerpt replace="/ *\/\/\s+ignore_for_file:[^\n]+\n//g; /(^|\n) *\/\/\s+ignore:[^\n]+\n/$1/g; /(\n[^\n]+) *\/\/\s+ignore:[^\n]+\n/$1\n/g; / *\/\/\s+ignore:[^\n]+//g; /([A-Z]\w*)\d\b/$1/g"?>
 
-:::note
-Use `this` only when there is a name conflict. 
-Otherwise, Dart style omits the `this`.
-:::
+## Types of constructors
+
+### Constructor inheritance
```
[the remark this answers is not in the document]
atsansone: Can do, though I think this has info relevant to the succeeding descriptions of types of constructors. Let's see how it reads.

src/content/language/constructors.md
atsansone: Understood. I changed the first example to call out what the initializer list was.

src/content/language/constructors.md
```diff
@@ -402,6 +340,211 @@ var logMap = {'name': 'UI'};
...
+Redirecting factories have several advantages:
+
+* An abstract class might provide a constant constructor
+  that uses the constant constructor of another class.
+* A redirecting factory constructor avoids the need for forwarders
+  to repeat the formal parameters and their default values.
+
+## Parameter initialization
+
+Dart can initialize parameters in a constructor in three ways.
```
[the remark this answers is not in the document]
atsansone: Fixed.

src/content/language/constructors.md
atsansone: Fair, but then we should "name" `PointC` constructors to illustrate the point.

src/content/language/constructors.md
atsansone: Not sure. Let me commit these changes first and you can have another look.

atsansone:
PTAL @natebosch

natebosch (approved):
I played around with the layout this afternoon.

 I think it we're pretty close but there are a few things I'd change.
 
 - Move `Non-default superclass constructors` to nest under `Constructor Inheritance`.
 - Move `Super parameters` up one level - nested under `Constructor Inhericance` as a sibling to `Non-default superclass constructors`.
 - Move `Instance Variable Initialization` before `Constructor Inheritance`.

```md
## Types of Constructors
### // all current, except Non-default superclass constructors

## Instance Variable Initialization
### // all current 

## Constructor Inheritance
### Non-default superclass constructors
### Super Parameters
```

I don't think we need much else changed after that. I had started on some changes in the intro, but in the end I didn't find anything I liked better than whats here.

Issue #5768, opened 2024-05-04
Migrate Jekyll `include` to liquidjs `render` where possible


parlough:
The `include` syntax we're using is supported as a fallback for migration, and [`render`](https://liquidjs.com/tags/render.html) is the appropriate replacement tag. Its included content doesn't inherit its parents scope, reducing the chance of mistakes.

parlough:
Closing in favor of https://github.com/dart-lang/site-www/issues/6256

PR #5770, opened 2024-05-04
Migrate from include to render where possible

parlough:
Contributes to https://github.com/dart-lang/site-www/issues/5768

Some includes rely on inheriting the parent still, which `render` purposefully doesn't support. They will need further updates before being able to migrate. Some rely on global data (such as `site.` variables), which 11ty doesn't yet expose to `render`, but likely will in the future. For now, those can keep using `include`.


atsansone (approved):
**question:** Why are only some changed to `render`? Could you explain this in the summary for this PR ahead of merging it?

Issue #5767, opened 2024-05-04
☂️ Move content out of `/guides` directory to more fitting homes


parlough:
The directory was kind of a catch-all. Move content out of it to a more content-related destination, or remove it if not helpful anymore.

dkbast:
It looks like this migration did not create proper redirects - see #6384 - all old links run into https://dart.dev/docs and the search has not yet indexed the new locations. Finding the guides is currently impossible (I only found the location by searching in this repo :/ )

MaryaBelanger:
Hey @dkbast, thanks for bringing this to our attention! We're looking into it ASAP

Issue #5766, opened 2024-05-04
Remove "/articles" directory


parlough:
The content within it should be moved somewhere more appropriate or removed as well.

- [x] `/articles/libraries/creating-streams` (https://github.com/dart-lang/site-www/pull/5771)
- [ ] `/articles/archive/converters-and-codecs` (https://github.com/dart-lang/site-www/issues/5175, https://github.com/dart-lang/site-www/pull/5788)
- [x] `/articles/archive/numeric-computation` (https://github.com/dart-lang/site-www/pull/5776)
- [ ] `/articles/archive/zones` (https://github.com/dart-lang/site-www/issues/4348, https://github.com/dart-lang/site-www/pull/5788)

PR #5771, opened 2024-05-04
Consolidate async docs away from old directories

parlough:
Contributes to https://github.com/dart-lang/site-www/issues/5766 and https://github.com/dart-lang/site-www/issues/5767


atsansone (approved):

MaryaBelanger (approved):
Important step in the right direction, thanks Parker!

Issue #4473, opened 2023-01-02
Add infrastructure to check for optimized images and automate it


parlough:
### Describe the problem

Optimizing images for size helps reduce site build time as well as reduce site load time for end users, so we should make sure all of our images are optimized.

### Expected fix

Implement tooling to easily optimize all images on the site, excluding diagrams which include source information.

Consider adding a check for optimized images if it's quick, or run the check weekly or only check new images somehow.

Issue #3124, opened 2021-04-06
Clarify calling one library in two ways on 'Effective Dart: Usage' page


dark-chocolate:
Page URL: https://dart.dev/guides/language/effective-dart/usage
Page source: https://github.com/dart-lang/site-www/tree/master/src/_guides/language/effective-dart/usage.md

Description of issue:

Quoting [this article](https://dart.dev/guides/language/effective-dart/usage#dont-allow-an-import-path-to-reach-into-or-out-of-lib), 

![image](https://user-images.githubusercontent.com/40364434/113766415-ac919580-971d-11eb-861e-e98d37c1111e.png)

I think the first line is not `bad`, but actually is the right way to import. It is further confirmed in the [next section](https://dart.dev/guides/language/effective-dart/usage#prefer-relative-import-paths):

![image](https://user-images.githubusercontent.com/40364434/113766318-908df400-971d-11eb-8528-414d1487af4f.png)

kwalrath:
It's true that the first line isn't bad; it's the combination of the two lines that's bad. I'm not sure how to make this clearer.

dark-chocolate:
@kwalrath By "combination", do you mean there should be a space between the two lines (as mentioned in best practices)?

kwalrath:
No, I mean that you're including the same library in two different ways, which is very bad.

Maybe we could change the example to something like this:

```dart
// This import by itself would be fine.
import 'package:my_package/api.dart';

// This import causes problems because it imports the same file (api.dart)
// but uses a different path. Avoiding ../lib imports prevents this problem. 
import '../lib/api.dart'; // Reaching into lib.
```

dark-chocolate:
@kwalrath I didn't pay attention to the 2nd import line in 1st screenshot. I thought it would be `import 'test_utils.dart'`. I got your point now.

kwalrath:
> @kwalrath I didn't pay attention to the 2nd import line in 1st screenshot. I thought it would be `import 'test_utils.dart'`. I got your point now.

You aren't the first person to be confused by the example, so I think it needs some improvement.

kwalrath:
Turns out we had a PR from long ago that aims to fix this. I think it's ready to commit, and I hope it helps.

parlough:
I think the addition of a "good" example with the current text after https://github.com/dart-lang/site-www/pull/3066 seems to sufficiently solve this issue. Let me know if you think further additions or changes are needed. 

Thanks again for creating an issue for this!

[...]

PR #5779, opened 2024-05-07
Set up automatic image optimization pipeline

parlough:
Use 11ty's first-party [image transformation plugin](https://www.11ty.dev/docs/plugins/image/) to automatically optimize images, convert them to `png`, `webp`, and `avif`, and then transform the site HTML to use the [`<picture>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture) to select the best one for the user. This only runs in production deploy builds to reduce serving time. Also applies lazy loading and async decoding to all images, even when serving without optimizations. The transform also adds a hash of the image, allowing us to expand caching for images in Firebase hosting.

To account for the transformed HTML structure, some minor HTML and CSS changes were needed as well.

Overall, this reduces page load time on pages that use images, and reduces unnecessary downloads that were due to a relatively short cache time for images. On a page with just a few images, such as https://dart.dev/tools/pub/automated-publishing, the lighthouse perf score increases around 10-15 points.

Fixes https://github.com/dart-lang/site-www/issues/4473
Fixes https://github.com/dart-lang/site-www/issues/3124


atsansone:
@parlough : I see no reason not to make this change, but I'm not seeing any benefit. Both the staged version and live version of https://dart.dev/web/debugging and  https://dart.dev/tools/pub/automated-publishing, Lighthouse scores were 97 for all.

Current
image

Staged
image

parlough:
@atsansone You'll see increased improvements for a mobile report or with more images above the fold (in initial viewport). Another benefit of this change is the **much** improved caching, as the plugin adds hashes to images, allowing us to extend the cache time of images greatly.

As another metric for those pages, visiting the automated publishing page with simulated slow 3g, dropped load time from 28s->21s.

atsansone (approved):

PR #5780, opened 2024-05-07
Add temp go link for json_codable

MaryaBelanger:
To be replaced with dart.dev/language/macros#jsoncodable (tbd)


MaryaBelanger:
> Could you clarify what you mean by temporary? Or what part is temporary?

It's a link to use in blogposts for now, and then once the main macro page lands I'll change it to point to the appropriate section there

PR #5803, opened 2024-05-13
Update pub-lish.md

jonasfj:
Documentation for `dart pub publish --skip-validation`.

### DIFF
--- a/src/content/tools/pub/cmd/pub-lish.md
+++ b/src/content/tools/pub/cmd/pub-lish.md
@@ -36,5 +36,14 @@ In the event of warnings, your package *is* uploaded.
 To ensure that your package has no warnings before uploading,
 either don't use `--force`, or use `--dry-run` first.
 
+### `--skip-validation`
+
+Publishes without going through the validation process or resolving dependencies.
+This is useful for advanced users who knows why the validation might fail and wishes to side step a particularly issues.
+
+**Example:** When publishing to pub.dev it may take a few minutes for a newly published package to become available.
+Hence, if you are publishing two dependent packages, where the second depends on the first.
+You can either wait a few minutes in between publishing the first and the second, or use `--skip-validation`
+to publish the second package immediately, by side-stepping client-side valiation.
 
 {% render 'pub-problems.md' %}
--- a/src/content/tools/pub/cmd/pub-lish.md
+++ b/src/content/tools/pub/cmd/pub-lish.md
@@ -39,7 +39,7 @@ either don't use `--force`, or use `--dry-run` first.
 ### `--skip-validation`
 
 Publishes without going through the client-side validation process or resolving dependencies.
-This is useful for advanced users who knows why the validation might fail and wishes to side step a particularly issues.
+This is useful for advanced users who knows why the validation fails and wishes to side step a particularly issues.
 
 **Example:** When publishing to pub.dev it may take a few minutes for a newly published package to become available.
 Hence, if you are publishing two dependent packages, where the second depends on the first.
--- a/src/content/tools/pub/cmd/pub-lish.md
+++ b/src/content/tools/pub/cmd/pub-lish.md
@@ -38,7 +38,7 @@ either don't use `--force`, or use `--dry-run` first.
 
 ### `--skip-validation`
 
-Publishes without going through the validation process or resolving dependencies.
+Publishes without going through the client-side validation process or resolving dependencies.
 This is useful for advanced users who knows why the validation might fail and wishes to side step a particularly issues.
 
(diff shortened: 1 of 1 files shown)


sigurdm (approved):

src/content/tools/pub/cmd/pub-lish.md
```diff
@@ -36,5 +36,14 @@ In the event of warnings, your package *is* uploaded.
 To ensure that your package has no warnings before uploading,
 either don't use `--force`, or use `--dry-run` first.
 
+### `--skip-validation`
+
+Publishes without going through the validation process or resolving dependencies.
+This is useful for advanced users who knows why the validation might fail and wishes to side step a particularly issues.
```
sigurdm, line 42: Perhaps clarify that this is only the client side validations

jonasfj: Added `client-side` to the first line :D

sigurdm (approved):

src/content/tools/pub/cmd/pub-lish.md
```diff
@@ -36,5 +36,14 @@ In the event of warnings, your package *is* uploaded.
 To ensure that your package has no warnings before uploading,
 either don't use `--force`, or use `--dry-run` first.
 
+### `--skip-validation`
+
+Publishes without going through the client-side validation process or resolving dependencies.
+This is useful for advanced users who knows why the validation might fail and wishes to side step a particularly issues.
```
atsansone, line 42: **suggestion:** This needs semantic breaks and active voice.

```suggestion
To side step a particular issue, you might want to use this option.
Advanced users might know why the validation might fail,
but know the issue would be resolved in other related work.
```

src/content/tools/pub/cmd/pub-lish.md
```diff
@@ -36,5 +36,14 @@ In the event of warnings, your package *is* uploaded.
...
+
+**Example:** When publishing to pub.dev it may take a few minutes for a newly published package to become available.
+Hence, if you are publishing two dependent packages, where the second depends on the first.
+You can either wait a few minutes in between publishing the first and the second, or use `--skip-validation`
+to publish the second package immediately, by side-stepping client-side valiation.
```
atsansone, line 47: **nitpick:** Use [semantic breaks](https://sembr.org/), write in active voice per [GSG](https://developers.google.com/style/voice), and fix spelling.

```suggestion
**Example:** Consider a scenario where you want to publish two packages at the same time. When you publish multiple packages to pub.dev,
a few minutes might pass before one package becomes available to other packages.
So, if package B depends on package A,
this publishing delay would result in a validation error for package B.

To avoid this error, you have two choices.

1. Wait a few minutes in between publishing the package A and package B.
1. Add the `--skip-validation` option to package B.
  This prevents package B from validating the presence of package A.
  As you are publishing them at the same time,
  you know they will both exist on pub.dev
  and not cause any errors for package users.
```

atsansone (commented):
@jonasfj : A few tweaks for clarification.

src/content/tools/pub/cmd/pub-lish.md
jonasfj: I updated it to:

> This is useful for advanced users who knows why the validation fails and wishes to side step a particularly issues.

I can see how it makes sense to write:
> To side step a particular issue, you might want to use this option.

But I kind of wanted to suggest that you should only do this if you are an advanced user and you know what you're doing. And know what particular issue you're side stepping.

The following appears to me like a statement about the knowledge that an "advanced user" may or may not have.

> Advanced users might know why the validation might fail,
> but know the issue would be resolved in other related work.

I don't understand it.

src/content/tools/pub/cmd/pub-lish.md
jonasfj: >  Consider a scenario where you want to publish two packages at the same time. 

I'm not sure I want the reader to "consider" anything. I'm stating an example, I don't want the reader to make up an opinion about the statements I'm about to make.

Maybe, it's possible to write "consider" because you want the reader to think about the example we present. It just seems a bit weird, I don't really want the reader to actively do anything. I'm simply providing information. 

> a scenario where you want to publish two packages at the same time. 

I'm not talking about a scenario where you're publishing two packages at the same time.
There is no support for publishing multiple packages in a single transaction :rofl: 
Such scenarios are also unrelated.

Nor are we talking about a scenario where we're are publishing two packages one after the other. Such scenarios are common an unrelated to this known use-case for `--skip-validation`.

We are talking about the specific scenario where you wish to publish two packages one after the other, where the first package depends on the second package.

PR #5808, opened 2024-05-13
Better prerelease docs

jonasfj:
I couldn't understand what the thing it said before meant.

Then I saw that I wrote it 5 years ago :see_no_evil: :see_no_evil: :see_no_evil: 
Then I saw that the PR was edited twice https://github.com/dart-lang/site-www/pull/1785 :rofl: :rofl: :rofl: 

So what was:

> In certain scenarios with nested dependencies the dependency resolution
> algorithm in pub, may pick a pre-release over a stable release.

Became:

> Sometimes pub might choose a pre-release instead of a stable release;
> this can occur when dependencies are nested.

and eventually what landed was:

> If a stable package in the dependency graph depends on a pre-release,
> then pub chooses that pre-release instead of a stable release.

There is some discussion of the issue here: https://github.com/dart-lang/site-www/pull/1785#discussion_r306253171

I suppose we could also say:

> If a stable package in the dependency graph depends on a pre-release,
> then pub _may_ choose a pre-release even if a stable release is possible.

Which is closer to what it already said, but I'm not sure we want to assuming anyone knows what a "stable package" is, I certainly don't. It's probably a package that has a stable release, but it feels a bit wild just introduce new terms like that with a formal definition.

Also it's nice to have: 
 * The reasoning: "the version solver does not attempt all solutions"
 * The intuition "`pub` prefers stable releases the same way it prefers newer versions"
 * The practical effect: "this rarely happens in practice"

Feel free to use other words, this is really a corner case.


src/content/tools/pub/publishing.md
```diff
@@ -285,12 +285,14 @@ Later, when you release version `2.0.0`, it takes precedence over all
...
-For example, if a user wants to test prereleases of version 2.1, then
+For example, if a user wants to test prereleases of version `2.1.0`, then
 instead of `^2.0.0` or `^2.1.0` they might specify `^2.1.0-dev.1`.
 
 :::note
-If a stable package in the dependency graph depends on a prerelease,
-then pub chooses that prerelease instead of a stable release.
+While `pub` prefers stable releases the same way it prefers newer versions,
+the version solver does not attempt all solutions and may pick a prerelease,
+even when a resolution that doesn't use prelreases exists.
```
sigurdm, line 294:
```suggestion
even when a resolution that doesn't use prereleases exists.
```

sigurdm, line 294: Perhaps we could expand to something like:

```suggestion
While `pub` prefers stable releases the same way it prefers newer versions,
the version solver is applying its heuristics greedily, and stops at the first solution it finds. Thus it might end up stuck at a local maximum, with a solution including prereleases even though a solution without those exists.
```

jonasfj: We could do that, I wouldn't mind -- but it does get rather technical.

sigurdm (approved):

atsansone (approved):

PR #5828, opened 2024-05-15
Add section on conflicting superinterfaces

natebosch:
Towards #562

Add a section discussing conflicting superinterfaces to the common
problems document.

The issue discusses the restrictions around both consistent member
signatures and consistent generics. An example is given only for the
latter.


src/content/guides/language/sound-problems.md
```diff
@@ -494,6 +494,34 @@ var maximumOrNull =
...
+A class cannot implement both `C<A>` and `C<B>`, including indirect
+superinterfaces.
+
+#### Example
+
+In the following code,
+class `C` will have conflicting generic interfaces and it will be impossible to
+define valide overrides for some members.
+
+<?code-excerpt "TODO"?>
```
natebosch, line 513: @atsansone - is there a doc describing how to set these up? Do we have any best practices for organization, or naming, or anything else I should be aware of?

atsansone: Yup. Kinda hidden: https://github.com/dart-lang/site-shared/blob/main/doc/code-excerpts.md

src/content/guides/language/sound-problems.md
```diff
@@ -494,6 +494,35 @@ var maximumOrNull =
 
 <hr>
 
+### Conflicting Superinterfaces
+
+A class which `implements` more than one superinterface must be able to
+implement valid overrides for every member.
+Each member with a given name requires compatible signatures across the
+superinterfaces.
```
atsansone, line 502: **suggestion:** May want to add something to clarify here.
```suggestion
A class which `implements` more than one superinterface, or member,
must be able to implement valid overrides for every member.
Each member with a given name requires compatible signatures across the
superinterfaces.
```

If this is done elsewhere, you don't need to do it here.

src/content/guides/language/sound-problems.md
```diff
@@ -494,6 +494,35 @@ var maximumOrNull =
...
+
+Superinterfaces must not include conflicting generics.
+A class cannot implement both `C<A>` and `C<B>`, including indirect
```
atsansone, line 505: **nitpick:** You can use `can't` here. In fact, [Google Style prefers it](https://developers.google.com/style/contractions#negation-contractions).

src/content/guides/language/sound-problems.md
```diff
@@ -494,6 +494,35 @@ var maximumOrNull =
...
+superinterfaces.
+
+#### Example
+
+In the following code,
+class `C` will have conflicting generic interfaces and it will be impossible to
+define valide overrides for some members.
```
atsansone, line 512: **suggestion:** Recommend a little tweak.

```suggestion
class `C` has conflicting generic interfaces.
Define valid overrides for some members would be impossible.
```

atsansone (approved):
@natebosch : Some suggestions. Great work!

src/content/guides/language/sound-problems.md
natebosch: I don't think this is correct - I'll tweak the wording for clarity. I'll change to.

```suggestion
A class which `implements` more than one superinterface must be able to
implement valid overrides for every member of every superinterface.
Each member with a given name requires compatible signatures across the
superinterfaces.
```

atsansone:
@natebosch : The code excerpts need updating.

natebosch:
Sorry for the delay. Coming back around to this, I'm not sure what argument to pass to `dart run excerpter` to validate that I did this correctly. I tried passing it both the `.md` file and the `.dart` file that I edited. In both cases the output is `Processed 0 out of 0 files: 0 out of 0 excerpts visited need to be updated.`

natebosch:
Using `dart run dart_site refresh-excerpt` works for me.

PR #5838, opened 2024-05-17
Best practices for downgrade testing

jonasfj:
Testing with `dart pub downgrade` is an optional extra thing a package author can do.

There is naturally a huge combination of different versions one can test with. But testing with the latest (`dart pub upgrade`) and oldest (`dart pub downgrade`) is a good choice.

**for package authors**!

For application developers it doesn't make much sense, it's better to simply rely on `pubspec.lock` and perhaps liberally apply `dart pub upgrade --tighten` as there is rarely any need to downgrade dependencies (once they have been upgraded).

For **package authors** allowing a wide dependency constraint is more important, otherwise it becomes too easy for package consumers to run into a dependency conflict. Where constraints from different packages they want to use are mutually incompatible.

Note: `dart pub upgrade` won't give the latest packages, it'll try, but it'll still respect incompatible version constraints. The same is the case for `dart pub downgrade`, it tried to give the oldest dependencies, but it may just give an oldest set of dependencies. The version solver does not search the entire state-space, it makes one decision at the time, and when it reaches a solution it's happy. `upgrade/downgrade` really only determines which branches of the solution space it tries to explore first.


src/content/tools/pub/dependencies.md
```diff
@@ -474,6 +474,34 @@ If you modify the pubspec and update to a new major version,
 then you might encounter breaking changes,
 so you need to test even more thoroughly.
 
+### Test with downgraded dependencies
+
+When developing packages for publication, it is often preferable to allow as a wide dependency constraints as possible.
```
sigurdm, line 479:
```suggestion
When developing packages for publication, it is often preferable to allow as wide dependency constraints as possible.
```

src/content/tools/pub/dependencies.md
```diff
@@ -474,6 +474,34 @@ If you modify the pubspec and update to a new major version,
...
+However, it's easy to forget to bump a dependency constraint when it becomes necessary.
+Thus, it's best practice to test you package against downgraded dependencies before publishing.
+
+Test against downgraded dependencies by running:
+
+ * [`dart pub downgrade`][],
+ * `dart analyze`, and possibly,
+ * `dart test`.
+
+Testing with downgraded dependencies should happen alongside normals tests with latest dependencies.
```
sigurdm, line 494:
```suggestion
Testing with downgraded dependencies should happen alongside normal tests with latest dependencies.
```

src/content/tools/pub/dependencies.md
```diff
@@ -474,6 +474,34 @@ If you modify the pubspec and update to a new major version,
...
+Should you ever find that your dependency constraints needs to be bumped, you can so using `dart pub upgrade --tighten`.
```
sigurdm, line 495:
```suggestion
If dependency constraints needs to be bumped, you can so manually, or by using `dart pub upgrade --tighten` to depend on the latest versions.
```

sigurdm (approved):

src/content/tools/pub/dependencies.md
jakemac53, line 494: Should we also suggest running `pub downgrade` on the oldest SDK you support, because that is the most likely to give you the oldest version solves?

jonasfj: We have no plans to do this in `pana` yet, would be interesting to consider.

I think it's an okay suggestion to make. But will leave it out of this PR, because I want it to land, so that we can point pana error message at it :D

Also the Dart SDK has `@Since('3.0.0')` annotations and there are diagnostics to help you bump the SDK constraint, if it's too low. These annotations and diagnostics aren't perfect, but they are much better than what we have in package land -- where there is nothing telling to bump your constraint when you use a newly added class/method/function/...

jakemac53: My main concern was that if people run `pub downgrade` on a new SDK, they might not actually get the older versions of dependencies which they claim to support (that don't support that newer SDK).

I am less concerned about the actual SDK constraint being tested, because as you say `@Since` etc help a lot with that (when used...).

src/content/tools/pub/dependencies.md
```diff
@@ -474,6 +474,49 @@ If you modify the pubspec and update to a new major version,
...
+However, it's easy to forget to bump a
+dependency constraint when it becomes necessary.
+Therefore, it's a best practice to test your package
+against downgraded dependencies before publishing.
+
+Test against downgraded dependencies by running:
+
+ * [`dart pub downgrade`][],
+ * `dart analyze`, and possibly,
+ * `dart test`.
```
parlough, line 499: Perhaps this can be better formatted? Perhaps with some more context about the steps?

````suggestion
To test against downgraded dependencies, run [`dart pub downgrade`][] and
verify your package still analyzes without errors and passes all tests:

```console
dart pub downgrade
dart analyze
dart test
```
````

\cc @MaryaBelanger For thoughts.

**Review by parlough — APPROVED** [review]
Technical content looks great and makes sense to me! Thanks so much for this and related work on encouraging this best practice.

I pushed a few formatting and copy fixes, but please wait for a final copy approval from Marya or Tony before landing.

File: src/content/tools/pub/dependencies.md [hunk 5]
```diff
@@ -474,6 +474,49 @@ If you modify the pubspec and update to a new major version,
 then you might encounter breaking changes,
 so you need to test even more thoroughly.
 
+### Test with downgraded dependencies
+
+When developing packages for publication, it is often preferable to
+allow as wide dependency constraints as possible.
```
MaryaBelanger, line 480:
```suggestion
allow the widest dependency constraints possible.
```

src/content/tools/pub/dependencies.md
```diff
@@ -474,6 +474,49 @@ If you modify the pubspec and update to a new major version,
...
+A wide dependency constraint reduces the likelihood that
+package consumers face a version resolution conflict.
+
+Hence, if you have a dependency on `foo: ^1.2.3` and
```
MaryaBelanger, line 484:
```suggestion
For example, if you have a dependency on `foo: ^1.2.3` and
```

src/content/tools/pub/dependencies.md
```diff
@@ -474,6 +474,49 @@ If you modify the pubspec and update to a new major version,
...
+version `1.3.0` of `foo` is released, it might be reasonable to
+keep the existing dependency constraint (`^1.2.3`).
+But if your package starts using features that were added in `1.3.0`, then
+you'll need to bump your constraint to `^1.3.0`.
+
+However, it's easy to forget to bump a
+dependency constraint when it becomes necessary.
+Therefore, it's a best practice to test your package
```
MaryaBelanger, line 492:
```suggestion
Therefore, the best practice is to test your package
```

src/content/tools/pub/dependencies.md
MaryaBelanger: I agree!

src/content/tools/pub/dependencies.md
```diff
@@ -474,6 +474,49 @@ If you modify the pubspec and update to a new major version,
...
+:::note
+Testing with `dart pub downgrade` enables you to find incompatibilities that
+you might not otherwise have discovered.
+But it doesn't exclude the possibility of incompatibilities.
+
+There are often so many different combinations of versions that
+testing them all is infeasible.
+There might also be older versions allowed by your dependency constraints that
+can't be resolved due to mutually incompatible version constraints from
+packages themselves or your `dev_dependencies`.
```
MaryaBelanger, line 515:
```suggestion
packages themselves or from your `dev_dependencies`.
```

MaryaBelanger (approved):
LGTM, few small fixes and especially [Parker's suggestion](https://github.com/dart-lang/site-www/pull/5838#discussion_r1620019642)!

src/content/tools/pub/dependencies.md
jonasfj: But there is more than one. We're in a subsection of a section called "best practices".

So I asked gemini :see_no_evil:, it also initially suggested using "the best practice", until I asked if couldn't cause problems if there was more than one. And then when I presented it with 3 examples:
```
I have 3 examples:

Example 1:
Best practices:
 * It is a best practice to put on pants.
 * it is a best practice to put on a shirt.

Example 2:
Best practices:
* It is best practice to put on pants.
* it is best practice to put on a shirt.

Example 3:
Best practices:
* It is the best practice to put on pants.
* it is the best practice to put on a shirt.

which of these examples should be preferred?
```

And it argued:
 * Example 1 is preferred because
   * Consistent use of article.
   * Avoids Unintentional Hierarchy: By not using "the," it avoids the implication that one practice is superior to the other.
   * Grammatically correct.
 * Example 2: was using "best practice" as verb, which is informal.
 * Example 3: Using "the" before each instance of "best practice" could suggest that these are the only or most important practices, potentially overlooking other valid options.

MaryaBelanger: Sounds good, you can skip this one!

I was thinking of it as "testing against downgraded dependencies is _the_ best practice _in a certain situation_ (that the section is describing)". But it sounds like it's not a certain situation that warrants testing with downgraded dependencies, more like just an extra precaution you can take anytime when testing.

jonasfj: Well,
> When developing packages for publication,

Then it is the best practice. But still probably not the only one :see_no_evil: 

-----------------

side note: nit picking this is not easy, because feel myself wanting assert that one cannot claim something to be "the best practice", without having to prove that no better practice can exist :rofl: 

But from what I understand "the best practice", is a figure of speech, for saying that it's widely accepted that something is a good thing to be recommended. Not necessarily that it is optimal.

jonasfj:
Feel free to merge, I don't have the powers.

src/content/tools/pub/dependencies.md
MaryaBelanger: > wanting assert that one cannot claim something to be "the best practice", without having to prove that no better practice can exist 🤣

Oh haha I get it. So, I think the way you wrote it is fine and we'll keep that. I think it's also fine to say "_the_ best practice in /_case_/" even if there are other best practices/cases in the _scope_ (in this case, _all_ the dependency best practices). 

But I can see that I'm assuming that wouldn't confuse anyone, and I shouldn't assume that! Thanks for doing your research with Gemini, that was a good idea (I do it all the time too, for technical stuff 😂 )

PR #5854, opened 2024-05-21
Updated note on arrow notation

### LINKED ISSUE #4582: Statement vs expression in => fat arrow explanation
### Page URL

https://dart.dev/guides/language/language-tour#functions

### Page source

_No response_

### Describe the problem

The docs in the linked page state:

> Note: Only an _expression_—not a _statement_—can appear between the arrow (=>) and the semicolon (;). For example, you can’t put an [if statement](https://dart.dev/guides/language/language-tour#if-and-else) there, but you can use a [conditional expression](https://dart.dev/guides/language/language-tour#conditional-expressions).

In describing what a statement is [in this Stack Overflow answer](https://stackoverflow.com/a/62608498), I listed some examples of statements, including the following:

- `myInt = 1;`
- `print('hello');`
- `return null;`
-  `if (name != null) {
    return name;
  } else {
    return 'Guest';
  }`

[One reader commented](https://stackoverflow.com/questions/53690957/difference-between-an-expression-and-statement-in-dart/62608498?noredirect=1#comment132930811_62608498):

> I found example contradictory to `myInt = 1;` being statement. It's `lambda() => myInt = 1;`. Why? Look at fat arrow explanation in the docs: [dart.dev/guides/language/language-tour#functions](https://dart.dev/guides/language/language-tour#functions). It says "Only an expression—not a statement—can appear between the arrow (=>) and the semicolon (;)". The aforementioned example compiles, linter doesn't complain, so it's an expression, not a statement.

That's true:

```dart
void main() {
  int i = 1;
  final function1 = () => i = 3;
  final function2 = () => print(i);
  function1();
  function2(); // 3
}
```

Am I wrong in my understanding of what a statement is? Is `i = 3` an expression rather than a statement? Is `print(i)` not a statement?

Or is the documentation misleading?

### Expected fix

If I'm wrong about what a statement is, then nothing needs to be changed about the documentation note above. However, we should define the terms **statement** and **expression** somewhere within the docs since these terms are used liberally without a clear explanation.

If `i = 3` and `print(i)` are in fact statements, then we should refine the => fat arrow note in the docs. I'm not exactly sure what that better wording would look like, though.

### Additional context

_No response_

atsansone:
Fixes #4582


src/content/language/functions.md
```diff
@@ -49,9 +49,13 @@ The <code>=> <em>expr</em></code> syntax is a shorthand for
...
-Only an *expression*—not a *statement*—can appear between the arrow (=\>) and
-the semicolon (;). For example, you can't put an [if statement][]
-there, but you can use a [conditional expression][].
+Only _expressions_ can appear between the arrow (=\>) and the semicolon (;).
+Expressions return values.
+This means that you can't write code in a place that expects a value.
+In the previous example,
+`_nobleGases[atomicNumber] != null;` returns a boolean value.
+The function then returns a value if the `atomicNumber` falls into
+the noble gas range. The function does nothing but return a value.
```
loic-sharma, line 58: I'd be careful with the term "return" as one might mistake it as "exit a function with a return value". Perhaps "evaluate" or "produce" might be a better term when referring to expressions?

Also, I think the distinction between an if statement and conditional expression is interesting. It shows an example of what you _can't_ use with the arrow syntax.

Here's a potential alternative:

```
Only _expressions_ can appear between the arrow (=\>) and the semicolon (;).
Expressions are anything that evaluate to a value.
For example, you could use a [conditional expression][] but not an [if statement][].
In the previous example,
`_nobleGases[atomicNumber] != null` is an expression that evaluates
to `true` if `_nobleGases` has an entry with the key `atomicNumber`.
```

src/content/language/functions.md
```diff
@@ -49,9 +49,13 @@ The <code>=> <em>expr</em></code> syntax is a shorthand for
 is sometimes referred to as _arrow_ syntax.
 
 :::note
-Only an *expression*—not a *statement*—can appear between the arrow (=\>) and
-the semicolon (;). For example, you can't put an [if statement][]
-there, but you can use a [conditional expression][].
+Only _expressions_ can appear between the arrow (=\>) and the semicolon (;).
+Expressions return values.
+This means that you can't write code in a place that expects a value.
```
leafpetersen, line 54: I wouldn't say "code".  Expressions *are* code.  Maybe keep the reference to statements? e.g. "...you can't put a statement (for example an [if statement][]) after the arrow."

src/content/language/functions.md
```diff
@@ -49,9 +49,13 @@ The <code>=> <em>expr</em></code> syntax is a shorthand for
...
+In the previous example,
+`_nobleGases[atomicNumber] != null;` returns a boolean value.
+The function then returns a value if the `atomicNumber` falls into
```
leafpetersen, line 57: The function returns a value no matter what, it just returns `true` or `false` depending on `atomicNumber`.

atsansone:
@leafpetersen , @loic-sharma : PTAL.

src/content/language/functions.md
```diff
@@ -49,9 +49,15 @@ The <code>=> <em>expr</em></code> syntax is a shorthand for
...
-the semicolon (;). For example, you can't put an [if statement][]
-there, but you can use a [conditional expression][].
+Only _expressions_ can appear between the arrow (`=\>`) and the semicolon (`;`).
+Expressions evaluate to values.
+This means that you can't write a statement where Dart expects a value.
+For example,
+you could use a [conditional expression][] but not an [if statement][].
+In the previous example,
+`_nobleGases[atomicNumber] != null;` returns a boolean value.
+The function then returns a boolean value
```
leafpetersen, line 59: The function always returns a boolean value.  Maybe "The function then returns a boolean value indicating whether or not the `atomicNumber`..."?

src/content/language/functions.md
```diff
@@ -49,9 +49,15 @@ The <code>=> <em>expr</em></code> syntax is a shorthand for
...
-Only an *expression*—not a *statement*—can appear between the arrow (=\>) and
-the semicolon (;). For example, you can't put an [if statement][]
-there, but you can use a [conditional expression][].
+Only _expressions_ can appear between the arrow (`=\>`) and the semicolon (`;`).
+Expressions evaluate to values.
+This means that you can't write a statement where Dart expects a value.
+For example,
+you could use a [conditional expression][] but not an [if statement][].
+In the previous example,
+`_nobleGases[atomicNumber] != null;` returns a boolean value.
```
leafpetersen, line 58: I would take the `;` off of here.  It's not part of the expression.  `_nobleGases[atomicNumber] != null` is the expression (which produces a value).  The `;` is a terminator which separates the function declaration from the declaration which follows it.  

Also, consider saying "produces a boolean value" or "evaluates to a boolean value" instead of returns, since return is usually associated with a function producing a value (which you capture in the next sentence).

PR #5856, opened 2024-05-22
Remove possibly confusing regex from extension methods syntax

### LINKED ISSUE #5306: Fix syntax in 'Extension methods' example
### Page URL

https://dart.dev/language/extension-methods.html

### Page source

https://github.com/dart-lang/site-www/tree/main/src/language/extension-methods.md

### Describe the problem

In the "[implementing extension methods](https://dart.dev/language/extension-methods#implementing-extension-methods)" section, why does the syntax show a question mark on the extension name?

```
extension <extension name>? on <type> {
  (<member definition>)*
}
```

I don't know all the grammar of describing language syntax, but to me it looks like the `?` is a non-optional part of the syntax, which obviously isn't the case. 

Does it represent something, like `<extension name>` could be nullable? In that case shouldn't it be on `<type>` and also represented as optional? e.g. `<extension name> on <type><?>`

### Expected fix

Just a question, I'd like to know what it means. 

### Additional context

_No response_

MaryaBelanger:
Fixes #5306

### DIFF
--- a/src/content/language/extension-methods.md
+++ b/src/content/language/extension-methods.md
@@ -197,8 +197,8 @@ to avoid a name conflict when invoking an extension explicitly.
 Use the following syntax to create an extension:
 
 ```plaintext
-extension <extension name>? on <type> {
-  (<member definition>)*
+extension <extension name> on <type> { // extension-name is optional
+  (<member definition>) // can provide one or more member-definitions
 }
 ```
 


src/content/language/extension-methods.md
```diff
@@ -197,8 +197,8 @@ to avoid a name conflict when invoking an extension explicitly.
 Use the following syntax to create an extension:
 
 ```plaintext
-extension <extension name>? on <type> {
-  (<member definition>)*
+extension <extension name> on <type> { // extension-name is optional
+  (<member definition>) // can provide one or more member-definitions
```
parlough, line 201: Can we keep the original symbols as well as the comments? Otherwise to those that are familiar with the syntax or skim past the comments, it might look like only one member definition is allowed.

With both the syntax and the comments, all readers are accounted for, plus the readers may become more familiar with the syntax as a result :) 

```suggestion
extension <extension name>? on <type> { // <extension name> is optional.
  (<member definition>)* // Can provide one or more <member definition>.
```

PR #5862, opened 2024-05-24
Link `prefer_const_constructors` with related Effective Dart

MaryaBelanger:
Links the Effect Dart entry "CONSIDER making your constructor const if the class supports it" to the `prefer_const_constructors` lint.

Section staged here: https://dart-dev--pr5862-const-lint-aiwlnawg.web.app/effective-dart/design#consider-making-your-constructor-const-if-the-class-supports-it


MaryaBelanger:
I see now that it's not an exact match. Thank you both for explaining! I'll close this

PR #5864, opened 2024-05-24
Changed all sdk refs to use caret syntax

### LINKED ISSUE #5863: Reset sdk versions to use caret syntax
Change references in `pubspec.yaml` examples to use caret syntax.

atsansone:
Fixes #5863

### DIFF
--- a/src/content/tools/pub/dependencies.md
+++ b/src/content/tools/pub/dependencies.md
@@ -99,7 +99,7 @@ you must use a more verbose `hosted` format:
 
 ```yaml
 environment:
-  sdk: '^[!2.14.0!]'
+  sdk: [!'>=2.14.0 < 3.0.0'!]
 
 dependencies:
   transmogrify:
@@ -248,7 +248,8 @@ The Dart community uses semantic versioning<sup id="fnref:semver"><a
 href="#fn:semver">1</a></sup>.
 
 You can express version constraints using either _traditional syntax_
-or _caret syntax_. Both syntaxes specify a range of compatible versions.
+or _caret syntax_ starting with Dart 2.19.
+Both syntaxes specify a range of compatible versions.
 
 The traditional syntax provides an explicit range like `'>=1.2.3 <2.0.0'`.
 The caret syntax provides an explicit starting version `^1.2.3`
@@ -312,7 +313,7 @@ or the next minor version for any package version earlier than 1.0.
 | Version value | Range covers to | Caret Syntax | Traditional Syntax  |
 |:-------------:|:---------------:|:------------:|:-------------------:|
 | >=1.0         | Next major      | `^1.3.0`     | `'>=1.3.0 <2.0.0'`  |
-| <1.0          | Next minor      | `^0.1.2 `    | `'>=0.1.2 <0.2.0' ` |
+| <1.0          | Next minor      | `^0.1.2`     | `'>=0.1.2 <0.2.0'`  |
 
 {:.table}
 
@@ -341,13 +342,13 @@ libraries—it doesn't actually need `test`. In this case, it specifies
 
 ```yaml
 dev_dependencies:
-  test: '>=0.5.0 <0.12.0'
+  test: ^1.20.0
 ```
 
-Pub gets every package that your package depends on, and everything *those*
+Pub gets every package that your package depends on, and everything _those_
 packages depend on, transitively. It also gets your package's dev dependencies,
 but it _ignores_ the dev dependencies of any dependent packages. Pub only gets
-*your* package's dev dependencies. So when your package depends on
+_your_ package's dev dependencies. So when your package depends on
 `transmogrify` it will get `transmogrify` but not `test`.
 
 The rule for deciding between a regular or dev dependency is simple: If
@@ -476,9 +477,9 @@ so you need to test even more thoroughly.
 
 ### Verify the integrity of downloaded packages
 
-When retrieving new dependencies, use the [`--enforce-lockfile`][]
-option to ensure the extracted package content matches
-the contents of the original archive.
+When retrieving new dependencies,
+use the [`--enforce-lockfile`][enforce-lock] option to ensure
+the extracted package content matches the contents of the original archive.
 Without modifying the [lockfile][],
 this flag only resolves new dependencies if:
 
@@ -486,7 +487,7 @@ this flag only resolves new dependencies if:
 * `pubspec.lock` is not missing
 * The packages' [content hashes][] match
 
-[`--enforce-lockfile`]: /tools/pub/cmd/pub-get#enforce-lockfile
+[enforce-lock]: /tools/pub/cmd/pub-get#enforce-lockfile
 [lockfile]: /tools/pub/glossary#lockfile
 [content hashes]: /tools/pub/glossary#content-hashes
 ---
--- a/src/content/tools/pub/dependencies.md
+++ b/src/content/tools/pub/dependencies.md
@@ -94,7 +94,7 @@ The version constraint is optional but recommended.
 If no version constraint is given, `any` is assumed.
 
 :::version-note
-If your package has a [SDK version][] earlier than 2.19,
+If your package has a [SDK constraints][SDK version] earlier than 2.19,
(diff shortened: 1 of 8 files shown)


parlough (commented):
SDKs before 2.19 didn't support this syntax, that's why a lot of these examples and snippets still use the old syntax. Can you revert the ones that have a lower bound before 2.19?

atsansone:
@parlough : Dart supported caret syntax since Dart 1.24.3: https://github.com/dart-lang/pub/issues/2153

@kevmoo verified and asked that it be done.

parlough:
@atsansone https://github.com/dart-lang/pub/pull/3672 landed in Dart 2.19. So users with an SDK before 2.19, will get an error when using pub with a caret SDK constraint. Versions before 2.19 are meant to be supported by many of these mentions, or are even the focus.

natebosch:
There was some confusion between allowing `^` constraints for `environment: sdk:` keys, vs `dependencies: package_name:` keys, or `environment: flutter:` keys.

For

```yaml
environment:
  sdk:
```

We should only use `^` if the constraints is `>=2.19 <3.0` or higher.

We should _not_ use `^` for any `environment: flutter:` keys.

A well formed and idiomatic pubspec will look like either of

```yaml
environment:
  sdk: ^3.3.0
  flutter: '>=3.19.0'
```

or (for older SDKs)

```yaml
  sdk: '>=2.18.0 <3.0.0'
  flutter: '>=3.3.0'
 ```
 
 It has been the case for a long time that `dependencies: package_name:` keys use `^` in most idiomatic pubpsecs.

**Comment by sigurdm** [comment]
> technically supported but the tool explicitly added an error.

Just to clarify: the error is only there when validating before publishing. Since dart 1.8 we can consume the caret syntax, also in sdk constraints.

**Review by natebosch — APPROVED** [review]

File: src/content/tools/pub/pubspec.md [hunk 1]
```diff
@@ -498,7 +498,7 @@ works with any Dart SDK that's version 3.0.0 or higher:
 
 ```yaml
 environment:
-  sdk: ^3.0.0
+  sdk: ^3.2.0
```
parlough, line 501: This no longer matches the above text on L497 now. Either update that or the snippet.

src/content/tools/pub/pubspec.md
```diff
@@ -534,7 +534,7 @@ under the `environment:` field:
 
 ```yaml
 environment:
-  sdk: '>=1.19.0 <3.0.0'
+  sdk: '^3.2.0'
```
parlough, line 537:
```suggestion
  sdk: ^3.2.0
```

src/content/tools/pub/pubspec.md
```diff
@@ -534,7 +534,7 @@ under the `environment:` field:
...
   flutter: ^0.1.2
```
parlough, line 538: While you're here, can you update this to just be a lower bound, as Nate [said](https://github.com/dart-lang/site-www/pull/5864#issuecomment-2136172953) "We should not use ^ for any environment: flutter: keys." 

```suggestion
  flutter: '>= 3.22.0'
```

Could also note that in the preceding text (only the lower bound should be specified).

src/content/tools/pub/dependencies.md
```diff
@@ -476,17 +487,17 @@ so you need to test even more thoroughly.
 
 ### Verify the integrity of downloaded packages
 
-When retrieving new dependencies, use the [`--enforce-lockfile`][]
-option to ensure the extracted package content matches
-the contents of the original archive.
+When retrieving new dependencies,
+use the [`--enforce-lockfile`][enforce-lock] option to ensure
+the extracted package content matches the contents of the original archive.
```
parlough, line 492:
```suggestion
the extracted package contents match the contents of the original archive.
```

src/content/tools/pub/dependencies.md
```diff
@@ -248,15 +258,16 @@ The Dart community uses semantic versioning<sup id="fnref:semver"><a
...
 
 The traditional syntax provides an explicit range like `'>=1.2.3 <2.0.0'`.
 The caret syntax provides an explicit starting version `^1.2.3`
 
 ```yaml
 environment:
-  # This package must use a 2.x version of the Dart SDK starting with 2.14.
-  sdk: '>=2.14.0 < 3.0.0'
+  # This package must use a 2.x version of the Dart SDK starting with 2.19.
+  sdk: '^2.19.0'
```
parlough, line 270: Hmm consider changing this one to `^3.2.0` as SDK constraints `>= 2.12.0` have one-off behavior of actually supporting 3.x versions of the SDK. That just avoids any potential confusion here.

```suggestion
  # This package must use a 3.x version of the Dart SDK starting with 3.2.
  sdk: ^3.2.0
```

src/content/tools/pub/dependencies.md
```diff
@@ -82,24 +82,34 @@ using the `hosted` source:
...
 
-The version constraint is optional but recommended. If no version constraint is
-given, `any` is assumed.
+The version constraint is optional but recommended.
+If no version constraint is given, `any` is assumed.
 
 :::version-note
-If your package has a [language version][] before 2.15,
-you must use a more verbose `hosted` format:
+If your package has a [SDK version][] earlier than 2.19,
```
parlough, line 97: Perhaps "SDK lower-bound" or "SDK constraint" earlier than?

parlough (approved):
Thanks for following through with this @atsansone and adding all of the clarifications! Super appreciated!

I left a few minor adjustments to consider. Also consider having @MaryaBelanger to review the few small copy changes.

src/content/tools/pub/pubspec.md
atsansone:
```suggestion
  sdk: ^3.0.0
```

src/content/tools/pub/dependencies.md
atsansone:
```suggestion
If your package has a [SDK constraints][SDK version] earlier than 2.19,
```

sigurdm:
Yeah the sdk constraints story is a bit muddled.

A few inline comments to expand a bit on your summary.

> Dart, [since 1.8](https://github.com/dart-lang/pub/issues/2153#issuecomment-1313887503), can consume the caret syntax.

Yes.

The reason that the caret syntax was not used with the Dart sdk constraint from the beginning was to provide backwards compatibility (older pubs would not look at the rest of the pubspec after they decide.) When looking into this while trying to find a solution for allowing caret sdk constraints when publishing to pub dev we found out that this was not working correctly - old pub would anyway read the entire pubspec from the package listing api and crash.

So we removed that restriction and [allowed](https://github.com/dart-lang/pub/pull/3672) caret syntax in the dart sdk constraint from dart 2.19

> The [SDK constraint validator](https://github.com/dart-lang/pub/blob/master/lib/src/validator/sdk_constraint.dart) returns an error when an upper bound constraint is not provided.

This is for the **Dart sdk constraint**. The Flutter sdk constraint is different. Here we ignore the upper bound.

Basically for marketing reasons it was decided to release flutter 2.0.0 without breaking backwards compatibility with flutter 1.x.x. To facilitate this it was decided that the Flutter constraint would [ignore](https://github.com/flutter/flutter/issues/68143) the upper bound. (ie. flutter is not strictly following semver, instead following https://docs.flutter.dev/release/compatibility-policy.)

> The SDK constraint validator error doesn't affect the dependency graph in any substantive way.

Correct.

The validator is only attempting to provide best-practice guidance before publication.

We had a corresponding check on the pub.dev server rejecting packages with a caret dart sdk constraint. That has been removed.

> The preference, expressed by [@kevmoo](https://github.com/kevmoo) and [@jonasfj](https://github.com/dart-lang/pub/issues/2153#issue-457953931) , is to use the caret syntax.

Yes, use the caret syntax for all your package dependencies, and for the dart sdk. That is usually the right thing (works with this version, and any later non-breaking version).

> Flutter can't support the caret syntax.

You can also use the caret syntax for the flutter sdk constraint, but  if you do, the upper bound will just be silently ignored. The flutter team strives for maintaining backwards compatibility forever, and offers to help out and provide migration tools when that is not possible.

PR #5872, opened 2024-05-27
Short-link for dart.dev/go/downgrade-testing

jonasfj:
For use in pana output on pub.dev (scoring page)

image

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -184,6 +184,7 @@
 
       { "source": "/go/next-gen-js-interop", "destination": "/interop/js-interop", "type": 301 },
       { "source": "/go/null-safety-migration", "destination": "/null-safety/migration-guide", "type": 301 },
+      { "source": "/go/downgrade-testing", "destination": "/tools/pub/dependencies#test-with-downgraded-dependencies", "type": 301 },
       { "source": "/go/package-discontinue", "destination": "/tools/pub/publishing#discontinue", "type": 301 },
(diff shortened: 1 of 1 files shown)


jonasfj:
Blocked until https://github.com/dart-lang/site-www/pull/5838 lands.

sigurdm (approved):

parlough (approved):
Thanks! Feel free to land when ready.

Optionally, do consider something like `/go/pub-downgrade-testing` or `/go/package-downgrade-testing` to make it more clear it's about package versions rather than SDK versions.

jonasfj:
I don't mind using either:
 * `/go/pub-downgrade-testing` or,
 * `/go/package-downgrade-testing`

But:
 * (A) I think when we present the link to the user, in the context of a pana report about downgrade testing, `pub-` or `package-` is going to be unnecessary. It'll be clear to the user that we're talking about something to do with packages. So from that perspective I don't think we need it.
 * (B) I don't think that there is a huge risk we'll end up conflicting with other future go-links.

So all we'd do is making the short-link longer :rofl:

parlough:
Makes sense! Keeping the shorter link works for me :)

MaryaBelanger (approved):

MaryaBelanger:
@atsansone this was dependent on another PR (the section doesn't exist yet) as stated in an earlier [comment](https://github.com/dart-lang/site-www/pull/5872#issuecomment-2133349575)

PR #5882, opened 2024-05-29
Update node, pnpm, and npm dependencies

parlough:
After this lands, you might need to update your Node version to the latest v20 LTS, run `corepack install`, and `pnpm upgrade`.

The newer Node version comes with a new API that I'd like that to use in the 11ty configuration.


atsansone (approved):
LGTM! Thanks.

Issue #5251, opened 2023-10-16
☔ Update website for GAR compliance


atsansone:
This project covers updating the docs.flutter.dev site to meet the Google Accessibility Requirements.

This project's details can be found in go/dash-site-gar-report.

This includes the following tasks:

## Assess a11y needs
- [ ] Run multipage a11y assessment tool in HTML
- [ ] Run multipage a11y assessment tool in JSON
- [ ] Aggregate changes from JSON results
## Fix sitewide common issues
- [ ] Request new color set from Design
- [ ] Change base colors to improve contrast
- [ ] Fix form controls in header file
## Fix heading issues
- [ ] Fix empty links
- [ ] Fix heading order
- [ ] Fix visual issues
- [ ] Add alt text
## Fix ARIA IDs and attributes
- [ ] Add titles to embedded iframes (YouTube)
- [ ] Fix duplicate ID attributes
- [ ] Check DOM / tab order
## Fix lexer contrast
- [ ] Fix rouge highlighting colors for contrast
## Visual Aid Check
- [ ] Check screen magnifier
- [ ] Check page zoom
- [ ] Check screen reader
- [ ] Check forced colors

PR #5884, opened 2024-05-29
Support specifying attributes to DartPad

parlough:
Enables the mechanism for adding a `title` attribute to DartPad code blocks.

````markdown
```dartpad run="true" title="The classic hello world program in Dart!" theme="dark"
void main() {
  print('Hello world!');
}
```
````

Contributes to https://github.com/dart-lang/site-www/issues/5251

@atsansone After this lands, feel free to add titles to the embedded DartPads. I'd be happy to review!

### DIFF
--- a/src/_11ty/plugins/highlight.js
+++ b/src/_11ty/plugins/highlight.js
@@ -86,12 +86,15 @@ function _highlight(
   language,
   attributeString,
 ) {
-  // Specially handle DartPad snippets so that inject_embed can convert them.
-  if (language.includes('-dartpad')) {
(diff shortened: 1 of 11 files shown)


atsansone (approved):
LGTM % 2 questions:

1. Why do we not set height on DartPad includes any more?
2. Why default to light theme?

parlough:
> 1. Why do we not set height on DartPad includes any more?

The new embed mechanism hasn't supported it, so removing the mentions to avoid misconceptions that it is supported. A lot of them time, they `run-dartpad` portions were copied from elsewhere without paying too much attention I think. If there's a strong desire to bring height adjustment back, open an issue, and I can definitely do so!

>2. Why default to light theme?

That's just what we've always done for embeds since the site body is light theme. Then in codelabs, dark theme indicates it's a conclusion exercise.

PR #5899, opened 2024-06-07
Fix arrow symbol typo in the functions.md

kamil-krawczyk:
In the first note, in the first sentence, the arrow symbol is written as '=\\>'. The backslash character must be removed.


parlough (approved):
Nice catch, thanks so much!

PR #5914, opened 2024-06-13
Begin incorporating linter rule diagnostic messages

parlough:
Content reviewed as part of the SDK already :)


src/_11ty/plugins/highlight.js
```diff
@@ -534,6 +534,7 @@ function _findMarkedTextAndUpdate(text) {
 
       const endIndex = line.indexOf('!]', startIndex);
       if (endIndex === -1) {
+        console.error(line);
```
parlough, line 537:
```suggestion
```

MaryaBelanger (approved):

PR #5918, opened 2024-06-17
Add lints, refactor Async Await

### LINKED ISSUE #4573: [PAGE ISSUE]: 'Asynchronous programming: futures, async, await'

### LINKED ISSUE #5626: Tweak page presentation on 'Asynchronous programming: futures, async, await' page
### Page URL

https://dart.dev/codelabs/async-await/

### Page source

https://github.com/dart-lang/site-www/tree/main/./src/content/codelabs/async-await.md

#### A minor layout issue

We read at [Working with futures: async and await](https://dart.dev/codelabs/async-await#working-with-futures-async-and-await)

> As the following two examples show, the async and await keywords result in asynchronous code that looks a lot like synchronous code. The only differences are highlighted in the asynchronous example, which—if your window is wide enough—is to the right of the synchronous example.

The examples are now longer side-by-side but one under the other.

Update: Well, it seems this actually depends on what the dynamic layout decides. In the PDF generated from the page, the code blocks are indeed side-by-side.

#### A minor output issue

Also the output by the synchronous code referred to above, is given as

```
Fetching user order...
Your order is: Instance of '_Future<String>'
```

but in Dart SDK 3.3.0 it is actually

```
Fetching user order...
Your order is: Instance of 'Future<String>'
```

#### Should a `main` that calls `await` return `void` or `Future<T>`?

Also `main` is declared as

```
Future<void> main() async { ...
}
```

which is (at least initially) confusing. Who receives the future returned by `main()`? The compiler is also very happy with

```
void main() async { ...
}
```

and indeed a bit below in [Example: Execution within async functions](https://dart.dev/codelabs/async-await#example-execution-within-async-functions) we find:

```
void main() async {
  countSeconds(4);
  await printOrderMessage();
}
```

There should be a word on that, maybe.

In https://github.com/dart-lang/site-www/issues/5624 Eric Ernst writes:

> You can have situations where a top-level function named `main` is called from Dart code. (It may or may not be the same function as the one which is invoked from the runtime in order to run the whole program.) This implies that even though a function named `main` has a special role in a Dart program execution, it can also be just another top-level function.
>
> Hence (and also just for consistency in a broader sense), there's nothing wrong in following all the normal rules about functions when writing the declaration of a top-level function named `main`.
>
> It is true that most `main` functions have return type `void`, even when the body is `async`, but that's basically just breaking the normal rules about good style because "it doesn't matter because no Dart code will ever call this function".
>
> With that in mind it's perfectly fine to use `Future<void>` as the return type of an async `main`. This implies that callers can await the computation performed during an execution of `main`, but the given future will be completed by an object which is of no interest (so the value of `await main(...)` should be ignored and discarded).

This is extremely informative but I don't know whether it would be too much for the codelab page.

#### More output to the console in the "do not do this" example

In [Incorrectly using an asynchronous function](https://dart.dev/codelabs/async-await#example-incorrectly-using-an-asynchronous-function), I suggest instead of just returning `'Large Latte'` in the example's future closure, printing a message to the console too:

```
// This example shows how *not* to write asynchronous Dart code.

String createOrderMessage() {
  var order = fetchUserOrder();
  return 'Your order is: $order';
}

Future<String> fetchUserOrder() =>
    Future.delayed(
      const Duration(seconds: 2),
      () { print("Future completes!") ; return 'Large Latte'; }
    );

void main() {
  print(createOrderMessage());
  print("Exiting main()");
}
```

We go from:

```            
Your order is: Instance of '_Future<String>'
```

to the more informative:

```
Your order is: Instance of '_Future<String>'
Exiting main()
Future completes!
```

#### Extending the code in the simple async example

In [Example: Execution within async functions](https://dart.dev/codelabs/async-await#example-execution-within-async-functions), I suggest extending the `main()` function by:

```
const bool doAwait = true;

void main() async {
  countSeconds(4);
  if (doAwait) {
    await printOrderMessage();
  }
  else {
    printOrderMessage();
  }
  print("Exiting main()");
}
```

And then ask the reader to flip `doAwait` to `false`.

The output goes from

```
Awaiting user order...
1
2
3
4
Your order is: Large Latte
Exiting main()
```

to

```            
Awaiting user order...
Exiting main()
1
2
3
4
Your order is: Large Latte
```

#### Highlight the dynamic nature of `order`

In the example given, the type of `order` actually changes depending on whether you `await` or not. This seems evident but it passed me by on first reading.

- When awaiting: `order` is of type `String` as we recuperate the result of the future.
- When not awaiting: `order` is a `Future<String>` as we recuperate the (likely as yet unfinished) future.
- Moreover the returned value is declared a `Future<String>` but we actually return a proper `String` at the very end of the function. The normal rules of how to declare return values have been suspended, in a sense.

```
Future<String> createOrderMessage() async {
  var order = await fetchUserOrder(); //  <----- "order" is a String (and can be declared as such)
  return 'Your order is: $order'; // <----- The function returns a proper String, not actually a Future<String>
}

Future<String> fetchUserOrder() =>
    // Imagine that this function is
// more complex and slow.
Future.delayed(
  const Duration(seconds: 2),
      () => 'Large Latte',
);

Future<void> main() async {
  print('Fetching user order...');
  print(await createOrderMessage());
}
```

#### More complex example for playing around

As an alternative example for async/await I propose this, which also shows that changing the "context" (what's on the heap), in this case `ingredient`, has effects:

```
String ingredient = "cream";

// createOrderMessage() returns to caller at "await order" with
// a Future<String>. That future:
// - waits for the 'order' future to complete, then
// - continues execution after "await order"
// i.e. the createOrderMessage() after "await order" is wrapped into a
// https://en.wikipedia.org/wiki/Continuation that can be called in the
// returned Future<String>

Future<String> createOrderMessage() async {
  Future<String> order = fetchUserOrder();
  print('Ordered. Waiting for completion in createOrderMessage()...');
  await order;
  print('Order completed in createOrderMessage()');
  return 'Your completed order is: $order';
}

Future<String> fetchUserOrder() =>
    Future.delayed(const Duration(seconds: 2), () {
      final result = 'Large Latte with $ingredient';
      print("Order completed inside the 'future': '$result'");
      return result;
    });

// Change this to 'false' to see the differences.
// If you "await", you will get a 'Large Latte with cream' inside main().
// If you do NOT "await", you will get a 'Large Latte with chocolate chips'
// after main() has already exited.

const bool doAwait = false;

Future<void> main() async {
  print('doAwait is $doAwait');
  print('Fetching user order in main()');
  final orderMsg;
  if (doAwait) {
    orderMsg = await createOrderMessage();
  }
  else {
    orderMsg = createOrderMessage();
  }
  ingredient = "chocolate chips";
  print('Obtained the following orderMsg in main(): $orderMsg');
  print('Exiting from main()');
}
```

Output if one does `await`:

```
doAwait is true
Fetching user order in main()
Ordered. Waiting for completion in createOrderMessage()...
Order completed inside the 'future': 'Large Latte with cream'
Order completed in createOrderMessage()
Obtained the following orderMsg in main(): Your completed order is: Instance of 'Future<String>'
Exiting from main()
```

Output if one does **not** `await`:

```
doAwait is false
Fetching user order in main()
Ordered. Waiting for completion in createOrderMessage()...
Obtained the following orderMsg in main(): Instance of 'Future<String>'
Exiting from main()
Order completed inside the 'future': 'Large Latte with chocolate chips'
Order completed in createOrderMessage()
```

#### `await` is an operator

One may note that "await" is classed as an "operator" according to https://dart.dev/language/operators

#### Name-drop the word `continuations`

Finally, one should maybe direct the reader to [The Wikipedia entry for 'Continuation'](https://en.wikipedia.org/wiki/Continuation), which is quite illuminating (to the moderate specialist).

#### A more verbose for the async/await with try/catch example

This one is a possible extension for [Example: async and await with try-catch](https://dart.dev/codelabs/async-await#example-async-and-await-with-try-catch)

```
Future<void> printOrderMessage() async {
  try {
    var order = await fetchUserOrder();
    print("The order is: $order");
  } catch (err) {
    print('Caught error: $err');
  }
  print('Returning from printOrderMessage()');
}

Future<String> fetchUserOrder() {
  // Imagine that this function is more complex.
  return Future.delayed(
      const Duration(seconds: 4),
      () => throw 'Cannot locate user order');
}

void main() async {
  await printOrderMessage();
  print('Exiting from main()');
}
```

#### Add a diagram for the simple example

The [simple example](https://dart.dev/codelabs/async-await#example-asynchronous-functions), is slightly modified by printing out more and adding printouts on types:

I drew a sort-of, kind-of, "sequence diagram" of not-fully-fleshed out semantics using the yEd editor. I'm not even sure whether the thread starting/ending at the black circles are really correct, but having that would help enormously (I don't seem to be able to attach the graphml file though):

##### Simple example, awaiting in `main()`

```
Future<String> createOrderMessage() async {
  var order = await fetchUserOrder();
  print("Type of 'order' is ${order.runtimeType}");
  return 'Your order is: $order';
}

Future<String> fetchUserOrder() =>
    Future.delayed(const Duration(seconds: 2), () {
      print("Future completes!");
      return 'Large Latte';
    });

Future<void> main() async {
  print('Fetching user order...');
  var msg = await createOrderMessage(); // DO AWAIT
  print("Type of 'msg' is ${msg.runtimeType}");
  print(msg);
}
```

Output:

```
Fetching user order...
Future completes!
Type of 'order' is String
Type of 'msg' is String
Your order is: Large Latte
```

Sequence diagram

##### Simple example, NOT awaiting in `main()`

```
Future<String> createOrderMessage() async {
  var order = await fetchUserOrder();
  print("Type of 'order' is ${order.runtimeType}");
  return 'Your order is: $order';
}

Future<String> fetchUserOrder() =>
    Future.delayed(const Duration(seconds: 2), () {
      print("Future completes!");
      return 'Large Latte';
    });

Future<void> main() async {
  print('Fetching user order...');
  var msg = createOrderMessage(); // DO NOT AWAIT
  print("Type of 'msg' is ${msg.runtimeType}");
  print(msg);
}
```

Output:

```
Fetching user order...
Type of 'msg' is Future<String>
Instance of 'Future<String>'
Future completes!
Type of 'order' is String
```

Sequence diagram 2

### Expected fix

As described above.

### Additional context

_No response_

### LINKED ISSUE #5740: Mention two linter rules not enabled by default on 'Asynchronous programming: futures, async, await' page
### Page URL

https://dart.dev/codelabs/async-await/

### Page source

https://github.com/dart-lang/site-www/tree/main/src/content/codelabs/async-await.md

### Use case

I did not understand correctly how to you async and Futur the right way because these two linter rules are not activated by default
```
  - discarded_futures
  - unawaited_futures
```

### Proposal

I think they should be mentioned in https://dart.dev/codelabs/async-await
that would have avoided a lot of pain in my learning...

### Expected fix

_No response_

### Additional context

_No response_

atsansone:
Fixes #5740: Missing Lints - [lines 1142 to 1153](https://github.com/dart-lang/site-www/pull/5918/files#diff-050056d658cbcff2d9d33ceeac780c8203637d77536b27bc7a9b324848107118R1142-R1153)
Fixes #4573: Bad example in tutorial - All examples
Fixes #5626: Fix reference to page presentation - [lines 317 to 318](https://github.com/dart-lang/site-www/pull/5918/files#diff-050056d658cbcff2d9d33ceeac780c8203637d77536b27bc7a9b324848107118R317-R318)

### DIFF
--- a/examples/async_await/bin/async_example.dart
+++ b/examples/async_await/bin/async_example.dart
@@ -1,9 +1,6 @@
 Future<void> printOrderMessage() async {
   print('Awaiting user order...');
-  // #docregion swap-stmts
   var order = await fetchUserOrder();
-  // print('Awaiting user order...');
-  // #enddocregion swap-stmts
   print('Your order is: $order');
 }
 
@@ -12,9 +9,18 @@ Future<String> fetchUserOrder() {
   return Future.delayed(const Duration(seconds: 4), () => 'Large Latte');
 }
 
+// #docregion awaiting
+const bool doAwait = true;
+// #enddocregion awaiting
+
 void main() async {
   countSeconds(4);
-  await printOrderMessage();
+  if (doAwait) {
+    await printOrderMessage();
+  } else {
+    printOrderMessage();
+  }
+  print("Exiting main()");
 }
 
 // You can ignore this function - it's here to visualize delay time in this example.
--- a/examples/async_await/bin/get_order_sync_bad.dart
+++ b/examples/async_await/bin/get_order_sync_bad.dart
@@ -8,15 +8,16 @@ String createOrderMessage() {
 
 Future<String> fetchUserOrder() =>
     // Imagine that this function is more complex and slow.
-    Future.delayed(
-      const Duration(seconds: 2),
-      () => 'Large Latte',
-    );
+    Future.delayed(const Duration(seconds: 2), () {
+      print('Future completes!');
+      return 'Large Latte';
+    });
 
 // #docregion main-sig
 void main() {
   // #enddocregion main-sig
   print('Fetching user order...');
   print(createOrderMessage());
+  print('Exiting main()');
 }
 // #enddocregion no-warning
--- a/examples/async_await/bin/get_order_sync_no_await.dart
+++ b/examples/async_await/bin/get_order_sync_no_await.dart
@@ -11,13 +11,9 @@ Future<String> fetchUserOrder() =>
       return 'Large Latte';
     });
 
-// #docregion main-sig
 Future<void> main() async {
-  // #enddocregion main-sig
   print('Fetching user order...');
-  // #docregion print-order
-  var msg = await createOrderMessage(); // DO AWAIT
+  var msg = createOrderMessage(); // DO NOT AWAIT
   print("Type of 'msg' is ${msg.runtimeType}");
   print(msg);
-  // #enddocregion print-order
 }
--- a/examples/async_await/test/async_await_test.dart
+++ b/examples/async_await/test/async_await_test.dart
@@ -16,6 +16,7 @@ void main() {
       3
       4
       Your order is: Large Latte
+      Exiting main()
     ''';
     expect(async_example.main, m.printsLines(output));
   });
@@ -37,6 +38,7 @@ void main() {
     final output = '''
       Fetching user order...
       Your order is: Instance of 'Future<String>'
+      Exiting main()
     ''';
     expect(get_order_sync_bad.main, m.printsLines(output));
   });
@@ -44,6 +46,9 @@ void main() {
   test('get_order', () {
     final output = '''
       Fetching user order...
+      Future completes!
+      Type of 'order' is String
+      Type of 'msg' is String
       Your order is: Large Latte
     ''';
     expect(get_order.main, m.printsLines(output));
--- a/src/content/guides/whats-new.md
+++ b/src/content/guides/whats-new.md
@@ -622,7 +622,7 @@ we made the following changes to this site:
 [initializing formal parameters]: /language/constructors#use-initializing-formal-parameters
 [support for packages]: /tools/dartpad#library-support
 [asynchronous programming tutorial]: /libraries/async/async-await
-[why asynchronous code matters]: /libraries/async/async-await#why-asynchronous-code-matters
+[why asynchronous code matters]: /libraries/async/async-await#why-does-asynchronous-code-matter
 [security]: /security
 
 
--- a/src/content/libraries/async/async-await.md
+++ b/src/content/libraries/async/async-await.md
@@ -6,71 +6,113 @@ js: [{url: '/assets/js/inject_dartpad.js', defer: true}]
 <?code-excerpt replace="/ *\/\/\s+ignore_for_file:[^\n]+\n//g; /(^|\n) *\/\/\s+ignore:[^\n]+\n/$1/g; /(\n[^\n]+) *\/\/\s+ignore:[^\n]+\n/$1\n/g"?>
 <?code-excerpt plaster="none"?>
 
-This tutorial teaches you how to write asynchronous code using
-futures and the `async` and `await` keywords. 
-Using embedded DartPad editors, 
-you can test your knowledge by running example code
-and completing exercises.
+This tutorial explains how to write asynchronous code using
+futures and the `async` and `await` keywords.
+It uses embedded copies of the DartPad IDE.
+To test your knowledge,
+complete the exercises in these embedded DartPads then run example code.
 
-To get the most out of this tutorial, you should have the following:
-
-* Knowledge of [basic Dart syntax](/language).
-* Some experience writing asynchronous code in another language.
-
-This tutorial covers the following material:
+:::secondary What you'll learn
 
 * How and when to use the `async` and `await` keywords.
 * How using `async` and `await` affects execution order.
 * How to handle errors from an asynchronous call
   using `try-catch` expressions in `async` functions.
 
-Estimated time to complete this tutorial: 40-60 minutes.
-
-:::note
-This page uses embedded DartPads to display examples and exercises.
-{% render 'dartpads-embedded-troubleshooting.md' %}
 :::
 
-The exercises in this tutorial have partially completed code snippets.
-You can use DartPad to test your knowledge by completing the code and
-clicking the **Run** button.
-**Don't edit the test code in the `main` function or below**.
+## Tutorial parameters
+
+### Requirements
+
+To get the most out of this tutorial, you need the following:
+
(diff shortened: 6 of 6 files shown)


MaryaBelanger:
Can you split this into two PRs, one that addresses #5740 and one for your refactor

atsansone:
@MaryaBelanger : Splitting would be difficult. Issue #5740 is covered in [lines 1142 to 1153](https://github.com/dart-lang/site-www/pull/5918/files#diff-050056d658cbcff2d9d33ceeac780c8203637d77536b27bc7a9b324848107118R1142-R1153) of async-await.md.

PR #5932, opened 2024-06-24
Updated part note

### LINKED ISSUE #2371: Clarify when to use Part directives
Can you please clarify what is meant by:

> Note: You may have heard of the part directive, which allows you to split a library into multiple Dart files. We recommend that you avoid using part and create mini libraries instead.

[https://dart.dev/guides/libraries/create-library-packages#organizing-a-library-package](https://dart.dev/guides/libraries/create-library-packages#organizing-a-library-package) 

It’s a bit unclear if part usage is discouraged in general or just in the context of library creation and if so it would be nice to elaborate on the rationale. Happy to open a PR to improve the docs once this has been clarified. Thanks!

atsansone:
Fixes #2371

### DIFF
--- a/src/content/guides/libraries/create-packages.md
+++ b/src/content/guides/libraries/create-packages.md
@@ -57,7 +57,7 @@ you have a situation where two classes are tightly coupled.
 You might know about the `part` directive.
 This directive allows you to split a library into multiple Dart files.
 Though part files can incorporate generated code into a library,
-The Dart team doesn't recommend using them.
+the Dart team doesn't recommend using them.
 Instead, create small libraries.
 :::
 
--- a/src/content/guides/libraries/create-packages.md
+++ b/src/content/guides/libraries/create-packages.md
@@ -56,8 +56,9 @@ you have a situation where two classes are tightly coupled.
 :::note
 You might know about the `part` directive.
 This directive allows you to split a library into multiple Dart files.
-Some might use part files to incorporate generated code into a library,
-but we recommend creating small libraries rather than using `part`.
+Though part files can incorporate generated code into a library,
+The Dart team doesn't recommend using them.
+Instead, create small libraries.
 :::
 
 Create a "main" library file directly under lib,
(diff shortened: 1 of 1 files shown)


munificent (approved):

src/content/guides/libraries/create-packages.md
```diff
@@ -59,9 +54,10 @@ In most cases, each class should be in its own mini library, unless
 you have a situation where two classes are tightly coupled.
 
 :::note
-You may have heard of the `part` directive, which allows
-you to split a library into multiple Dart files. We recommend
-that you avoid using `part` and create mini libraries instead.
+You might know about the `part` directive.
+This directive allows you to split a library into multiple Dart files.
+Some might use part files to incorporate generated code into a library,
+but we recommend creating small libraries rather than using `part`.
```
MaryaBelanger, line 60: Can we stick with the text Bob originally [suggested](https://github.com/dart-lang/site-www/issues/2371#issuecomment-620261086)? Saying "Some might..." makes it sound like a personal choice rather than the actual "this is the only thing you should ever use them for"

```suggestion
Part files are useful for incorporating generated code into a library,
but we otherwise recommend that you don't use them.
Instead, create small libraries.
```
Or something like that.

src/content/guides/libraries/create-packages.md
```diff
@@ -59,9 +54,11 @@ In most cases, each class should be in its own mini library, unless
 you have a situation where two classes are tightly coupled.
 
 :::note
-You may have heard of the `part` directive, which allows
-you to split a library into multiple Dart files. We recommend
-that you avoid using `part` and create mini libraries instead.
+You might know about the `part` directive.
+This directive allows you to split a library into multiple Dart files.
+Though part files can incorporate generated code into a library,
+The Dart team doesn't recommend using them.
```
atsansone, line 60: **nitpick:** 
```suggestion
the Dart team doesn't recommend using them.
```

Issue #5938, opened 2024-06-26
Indicate Flutter users do not need to install Dart separately


abitofevrything:
### Page URL

https://dart.dev/get-dart

### Describe the problem

Flutter users often expect Flutter to require Dart to be installed separately, as they understand that Dart, the language, is separate from Flutter, the framework. However Flutter bundles its own Dart installation, which surprises users when they later see the warning about having a separate Dart installation in `flutter doctor`.

The installation page does mention that Flutter includes the Dart SDK, but only after the option for installing Dart via a package manager, which is what users on Linux systems often jump to, and therefore do not see the message.

### Expected fix

A warning or information box at the start of the installation pages indicating Flutter users should not install Dart separately, or that they do not _need_ to (sometimes a separate Dart installation is needed - e.g I have a separate Dart installation to get the latest features on Dart's dev branch before they come to Flutter).

### I would like to fix this problem.

- [X] I will try and fix this problem on dart.dev.

PR #5939, opened 2024-06-26
Add note informing users not to install Flutter and Dart separately

abitofevrything:
Fixes https://github.com/dart-lang/site-www/issues/5938

---

<details>

</details>

### DIFF
--- a/src/content/get-dart/index.md
+++ b/src/content/get-dart/index.md
@@ -15,7 +15,7 @@ To learn more about the Dart SDK, consult the [Dart SDK overview](/tools/sdk).
 
 :::tip
 If you've installed or plan to [install the Flutter SDK][install-flutter], it
-includes the full Dart SDK and you do not need to install Dart separately.
+includes the full Dart SDK. You don't need to install Dart separately and can skip this guide.
 :::
 
 ## System requirements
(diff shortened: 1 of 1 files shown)


atsansone:
**question:** This point already exists on this page.

image

Does this not meet your requirement? (Honestly, just asking. Overuse of admonitions is a concern with the docs.)

abitofevrything:
Yes, it is mentioned - but only further down the list than the package manager option. I see a lot of linux users on a flutter support group (FlutterDev on Discord) who see this option and immediately jump to it, skipping over the point about installing flutter underneath.

Another option would therefore be to reorder the list to place the Install Flutter option first, but I also think that having the message stand out in some way (in this case, by being in a tip box) early on in the page can help draw more attention to it as most users of Dart are still currently only using it because of Flutter - and therefore do not need to be visiting the Get Dart page in the first place.

atsansone:
@abitofevrything : Make sense. Let me check the admonitions quick and see if we can get more content above the edge of the first screen.

src/content/get-dart/index.md
```diff
@@ -13,6 +13,11 @@ you need to develop Dart command-line, server, and web apps.
 
 To learn more about the Dart SDK, consult the [Dart SDK overview](/tools/sdk).
 
+:::tip
+If you've installed or plan to [install the Flutter SDK][install-flutter], it
+includes the full Dart SDK and you do not need to install Dart separately.
```
atsansone, line 18: **suggestion:** 

```suggestion
If you've installed or plan to [install the Flutter SDK][install-flutter], it
includes the full Dart SDK. You don't need to install Dart separately and can skip this guide.
```

atsansone (approved):

PR #5950, opened 2024-06-28
Updated Wildcard patterns

### LINKED ISSUE #5926: Add recommendation to Wildcards on 'Pattern types' page
### Page URL

https://dart.dev/language/pattern-types/

### Page source

https://github.com/dart-lang/site-www/tree/main/src/content/language/pattern-types.md

### Describe the problem

In the section 'Wildcards' near the end of the page, it is mentioned that a variable pattern like `int _` is a useful construct when we wish to test the type of the scrutinee (the actual example is a bit more involved, `case (int _, String _):`, but that's just the same thing done twice as part of a record pattern matching).

This is indeed true, but it would probably be helpful if this rather implicit recommendation is accompanied by another recommendation: It is in a sense _safer_ to have a habit of using an object pattern (like `int()` respectively `case (int(), String()):`).

They only differ when the given type is generic, but the object pattern is more concise and convenient, and in a sense safer with generic types, and there is no reason to prefer the variable pattern with non-generic types (so we might just as well recommend that developers use the object pattern as a _habit_).

Here's how they differ:

Consider an object pattern whose target type is generic and whose actual type arguments have been omitted. This pattern will receive actual type arguments based on the matched value type. This means that `case MyType():` will implicitly be inferred to mean `case MyType<Some, TypeArguments>():` where said type arguments are chosen to be as specific as possible for the given matched value type (for instance, if we're switching on an expression of type `Iterable<int>` and matching `case List():` then it is inferred to mean `case List<int>():`).

In contrast, consider the same situation except that we are using a variable pattern, `case MyType _:`. In this case, `MyType` uses instantiation to bound to obtain the omitted type arguments. If there are no declared bounds then this means `case MyType<dynamic, dynamic> _:` (similarly, with a matched value type of `Iterable<int>`, `case List _:` means `case List<dynamic> _:`).

This topic is discussed a bit more in [this comment](https://github.com/dart-lang/language/issues/2047#issuecomment-2180365352).

### Expected fix

The section 'Wildcards' could have a concise description of this topic and a reference to the section 'Object' that occurs a few lines earlier. Perhaps something like this:

> Consider using an object pattern (`MyType()`) to test a type rather than a variable pattern (`MyType _`), especially when the given type is generic. For more details please see the section 'Object'.

The section 'Object' could then present the above line of thinking and recommend using `case MyType():` rather than `case MyType _:` as a habit.

### Additional context

_No response_

atsansone:
Fixes #5926

### DIFF
--- a/src/content/language/pattern-types.md
+++ b/src/content/language/pattern-types.md
@@ -438,13 +438,24 @@ If an object has extra fields that the pattern doesn't destructure, it can still
 
 ## Wildcard
 
-`_`
+`_`  
+`<type>()`
 
-A pattern named `_` is a wildcard, either a [variable pattern](#variable) or
-[identifier pattern](#identifier), that doesn't bind or assign to any variable.
+Wildcard patterns don't bind or assign to any variable.
 
-It's useful as a placeholder in places where you need a subpattern in order to
-destructure later positional values:
+These patterns can either be a [variable pattern](#variable) (`_`) or
+[identifier pattern](#identifier) (`<type>()`).
+
+* Object patterns are more concise and convenient,
+  even more so with generic types.
+  With generic types, an object pattern (`MyType()`) infers the
+  most specific type arguments from the matched value.
+* A variable pattern (`MyType _`) uses less specific, often dynamic,
+  type arguments.
+* With non-generic types, there's no reason to prefer variable patterns.
+
+When you need a subpattern to destructure later positional values,
+use a wildcard as a placeholder. This example uses a variable pattern.
 
 <?code-excerpt "language/lib/patterns/pattern_types.dart (wildcard)"?>
 ```dart
@@ -452,13 +463,14 @@ var list = [1, 2, 3];
 var [_, two, _] = list;
 ```
 
-A wildcard name with a type annotation is useful when you want to test a value's
-type but not bind the value to a name:
+To test a value's type but not bind the value to a name,
+use a wildcard name with a type annotation.
+This example uses an object pattern.
 
 <?code-excerpt "language/lib/patterns/pattern_types.dart (wildcard-typed)"?>
 ```dart
 switch (record) {
-  case (int _, String _):
+  case (int(), String()):
     print('First field is int and second is String.');
 }
 ```
--- a/examples/language/lib/patterns/pattern_types.dart
+++ b/examples/language/lib/patterns/pattern_types.dart
@@ -249,7 +249,7 @@ void miscDeclAnalyzedButNotTested() {
     var record = (1, 'one');
     // #docregion wildcard-typed
     switch (record) {
-      case (int _, String _):
+      case (int(), String()):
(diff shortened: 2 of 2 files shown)


src/content/language/pattern-types.md
```diff
@@ -438,27 +438,39 @@ If an object has extra fields that the pattern doesn't destructure, it can still
...
+This example uses an object pattern.
 
 <?code-excerpt "language/lib/patterns/pattern_types.dart (wildcard-typed)"?>
 ```dart
 switch (record) {
-  case (int _, String _):
+  case (int(), String()):
     print('First field is int and second is String.');
 }
 ```
```
eernstg, line 476: Again, I think it would be OK to keep lines 455-464 from the old text unchanged.

What I wanted to propose in #5926 was actually that there should be an extra paragraph here at the end of this section, perhaps something like the following:

--------------------------------------------------------

This approach works fine for non-generic types like `int` and `String`. However, if you wish to test whether the matched value has a generic type then you need to specify the type arguments explicitly. The section about object patterns shows how you can use object patterns to infer the type arguments, which may be both more convenient and safer. Here is an example:

<?code-excerpt "some magic ;-)"?>
```dart
Iterable<int> iter = [2, 3, 4];
switch (iter) {
  case List<int> _:
    print('It was a list whose first element is ${iter[0].isEven}');
}
```

If we had used `case List _:` rather than `case List<int> _:` then it means `case List<dynamic> _:`, which in turn means that `iter` does not get promoted, and `iter[0]` is a compile-time error.

--------------------------------------------------------

Then we could have something like the following at the end of the 'Object' section:

--------------------------------------------------------

An extreme example is the case where the object pattern doesn't specify any properties at all. In this case the object pattern amounts to a type test. It is particularly useful in the case where the given type is generic, because the actual type arguments are inferred from the static type of the matched value. For example:

<?code-excerpt "some magic ;-)"?>
```dart
Iterable<int> iter = [2, 3, 4];
switch (iter) {
  case List():
    print('It was a list whose first element is ${iter[0].isEven}');
}
```

The object pattern `List()` is inferred to mean `List<int>()` because the static type of `iter` is `Iterable<int>`, and this means that the invocation of `iter[0].isEven` is type correct.

Hence, it may be a good habit to use an "empty" object pattern like `List()` when you just want to test the type of the matched value, rather than using a wildcard like `List<int> _`.

--------------------------------------------------------

eernstg (commented):
I tried to suggest a different ordering of the text. It is a little bit tricky because it would be nice to have the two examples (`List _` vs. `List()`) near to each other. However, it also seems useful to have the object pattern example in the object pattern section, and the wildcard example in the wildcard section.

So my comments are surely not in a usable shape yet, but I hope they can be used to take the next step.

RedBrogdon:
I'm going to close this for now. This change will likely be picked up again at a later date, with attribution back to this PR.

PR #5959, opened 2024-07-01
Add rudimentary freshness tool

parlough:
Added for determining freshness. Nothing happens in CI, this is just for running locally.

Used like:

```
./dash_site freshness
```

with output like:

```
====== Files with missing lastVerified dates ======

❓   missing-date   web/libraries.md
❓   missing-date   web/wasm.md

====== Files with stale lastVerified dates ======

🟥    2021-05-13    resources/google-apis.md
🟧    2022-11-21    resources/coming-from/js-to-dart.md
🟧    2022-12-13    resources/coming-from/swift-to-dart.md
🟨    2023-02-23    guides/environment-declarations.md

```

You can optionally change the cutoff to include a file as stale with the `--cutoff` option which accepts an integer amount of days. It currently defaults to 270 days.

The staleness is determined by the `lastVerified` option in each page's YAML frontmatter. `lastVerified` indicates when the page was last verified as up to date as of:

```yaml
---
title: Document title
lastVerified: 2024-04-21
---
```

If a file shouldn't be checked for freshness, you can specify `skipFreshness: true` in the frontmatter:

```yaml
---
title: Document title
skipFreshness: true
---
```


atsansone (approved):
LGTM to start using.

PR #5978, opened 2024-07-17
Refresh Dart docs home page

### LINKED ISSUE #5339: Update docs landing page
I think the docs "landing page" https://dart.dev/guides should be updated to better represent the contents of the site as laid out in the side nav. The "futures..." box doesn't really fit here either.

So the boxes should be:
- Tutorials and codelabs
- Language
- Effective Dart
- Core libraries
- Packages
- Development
- Tools & techniques
- Resources

And each block should accurately summarize the contents.

atsansone:
Fixes #5339

Specifics: This refactors the docs home page to make it more informative and engaging.

Other pages needed updates based on the change to allow use of different columnar layouts.

[Direct Link](https://dart-dev--pr5978-fix-5339-h3rgs2uv.web.app/guides)


src/content/guides/index.md
```diff
@@ -4,24 +4,38 @@ description: Learn to use the Dart language and libraries.
 toc: false
 ---
 
-Welcome to the Dart documentation!
-For a list of changes to this site—new pages, new guidelines, and more—see
-the [What's new][] page.
-
-[What's new]: /guides/whats-new
```
johnpryan, line 11: Why are we taking this paragraph out?

src/_sass/_site.scss
```diff
@@ -25,6 +25,19 @@
...
+  margin-block-start: 0px;
+  // margin-block-end: 0px;
+  margin-inline-start: 0px;
+  margin-inline-end: 0px;
+  padding-block-start: 0px;
+  padding-block-end: 0px;
+  // padding-inline-start: 0px;
+  padding-inline-end: 0px;
+  unicode-bidi: isolate;
+}
```
johnpryan, line 40: Why does this change require changing the margin of all of the headers on the site?

src/_sass/_site.scss
```diff
@@ -370,32 +383,32 @@ main .content {
...
-  min-height: 100px;
-  padding: 20px;
-  margin: 15px;
+  padding: 1rem;
   overflow: hidden;
-  box-sizing: border-box;
-  background-color: #294058;
-  border-radius: 0;
+  background-color: $gray-lighter;
+  border: 2px solid $flutter-color-dark-blue;
```
johnpryan, line 403: I don't like this new style, I prefer the old one. We worked with a UI designer to get this right.

atsansone: I've worked with the design team on these things too. My concern is that with so many cards, the page might become overwhelmingly dark. I'm happy to change it, but that's the reasoning.

src/_sass/_site.scss
atsansone: All browsers have default settings around margins, etc. If you don't reset them, your styles don't necessarily do what you expect.

src/content/guides/index.md
atsansone: Ah. Originally, I had replaced it with another card. I can still do that. I'll see what feedback I get from @sethladd first.

sethladd:
Hi level feedback is that "it's a log of big dark boxes" and a bit hard for my eye to scan through. I'm not a designer, but I'm guessing there's ways to make it easier to scan through the page, less big dark boxes?

Screenshot 2024-07-18 at 1 02 15 PM

src/_sass/_site.scss
johnpryan: I understand how default CSS styles work, I'm asking why do you need to reset _all_ of the styles for the entire site, when our CSS is assuming the browser defaults? This seems like it's impacting other areas of the site that you didn't intend to change.

src/_sass/_site.scss
```diff
@@ -408,7 +421,7 @@ main .content {
   }
 
   a {
-    color: $site-color-card-link;
+    color: $flutter-color-dark-blue;
```
johnpryan, line 424: Did you mean to change this color? This is darker than before, I would prefer to keep the style we currently have.

src/_sass/core/_variables.scss
```diff
@@ -12,13 +12,15 @@ $flutter-color-light-blue: #E7F8FF;
 $flutter-color-yellow: #FFF275;
 $flutter-color-yellow-300: #FFF59C;
 $flutter-color-grey-500: #E8EAED;
+$flutter-color-darkest-blue: #041E3C;
+$flutter-color-lightest-grey: #F8F9FA;
```
johnpryan, line 16: Could you take out these CSS changes and file an issue for them separately?

sethladd:
I would also consider putting "language tour" as the first item under "Learn about Dart". I would also consider moving "packages" out from "learning dart", as "packages" is a more advanced topic. Under learning, the language tour, the effective dart, and the language syntax fit thematically.

"Packages" would be good under a theme like "Ecosystem".

We could also consider a flow like this:

"New to Dart? Start here"

* Dart Pad and Learning Dart (with the language tour, I assume?)
* Install Dart

(in the above, the order is least friction to most friction)

"Explore Dart"

* Libraries overview
* Patterns
* Records

"Use Dart"

* Command line apps
* Flutter
* Server-side apps
* Interop

"Build your skills"

* Effective Dart
* Advanced concurrency
* etc

"Leverage the ecosystem"

* Packages

sethladd:
Small nit, I see the page is "/guides" but sounds like this is "the homepage for all the docs", so maybe "/docs" is a good fit?

src/_sass/_site.scss
atsansone: I did. That color was not one included in the [Dart brand guidelines](go/dart-brand-guidelines), so I wanted to make it consistent. It is as dark as the header.

atsansone:
Changed design based on feedback from @mit-mit . @mit-mit , have a look and let me know about any copy changes as well. Thanks!

johnpryan:
nit: I have to click on the header to follow the link, it would be great if we could click anywhere on the card to go to the guide. 

@atsansone should we cancel the UX office hours since you seem to be happy with the design feedback you've gotten thus far?

atsansone:
@johnpryan : I can fix that issue with the card (I think?) before we continue. (Great point, btw!)

This is something that I know @mit-mit wants to sign off on.

I have no issue reviewing things together, if you would still like to do so.

RedBrogdon:
I'm going to close this for now. This change will likely be picked up again at a later date, with attribution back to this PR.

PR #6000, opened 2024-07-26
Disable 11ty image processing for now

parlough:
To test locally, checkout this branch (`gh pr checkout 6000`), run `pnpm install`, then `./dash_site serve`.


mit-mit:
Hi Parker, this seems to resolve the issue I was having. `./dash_site serve` now works as expected!

parlough:
> Hi Parker, this seems to resolve the issue I was having. `./dash_site serve` now works as expected!

Thanks for trying it out. I'll land this to avoid the issue for now and revisit reenabling the process later on when I have more time.

sfshaza2 (approved):
lgtm

PR #6006, opened 2024-07-31
Add section on exporting members in usage

srujzs:
Addresses https://github.com/dart-lang/sdk/issues/56327#issuecomment-2257230135

Being able to call Dart code from JS is a common
use case (e.g. event listeners). We should add a
section that shows how to export functions and
links to our section on exporting entire Dart
instances.

### DIFF
--- a/src/content/interop/js-interop/mock.md
+++ b/src/content/interop/js-interop/mock.md
@@ -98,8 +98,8 @@ void main() {
 `@JSExport` allows you to declare a class that can be used in
 `createJSInteropWrapper`. `createJSInteropWrapper` will create an object literal
 that maps each of the class' instance member names (or renames) to a JS
-callback, which is created using [`Function.toJS`]. This JS callback then calls
-the instance member when called. In the above example, getting and setting
+callback, which is created using [`Function.toJS`]. When called, the JS callback
+will in turn call the instance member. In the above example, getting and setting
 `counter.value` gets and sets `fakeCounter.value`.
 
 You can specify only some members of a class to be exported by omitting the
--- a/src/content/interop/js-interop/mock.md
+++ b/src/content/interop/js-interop/mock.md
@@ -97,9 +97,10 @@ void main() {
 
 `@JSExport` allows you to declare a class that can be used in
 `createJSInteropWrapper`. `createJSInteropWrapper` will create an object literal
-that maps each of the class' instance member names (or renames) to a JS callback
-that triggers the instance member when called. In the above example, getting and
-setting `counter.value` gets and sets `fakeCounter.value`.
+that maps each of the class' instance member names (or renames) to a JS
+callback, which is created using [`Function.toJS`]. This JS callback then calls
+the instance member when called. In the above example, getting and setting
+`counter.value` gets and sets `fakeCounter.value`.
 
 You can specify only some members of a class to be exported by omitting the
 annotation from the class and instead only annotate the specific members. You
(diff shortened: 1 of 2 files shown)


src/content/interop/js-interop/usage.md
```diff
@@ -389,6 +389,47 @@ but can often be elided on and within interop types and on extension members as
 the compiler can tell it is a JS interop type from the representation type and
 on-type.
 
+## Exporting Dart code to JS
```
RedBrogdon, line 392: Rather than "exporting Dart code," WYDT about "exporting Dart functions and objects," since those are the two specific techniques you're illustrating in this section?

src/content/interop/js-interop/usage.md
```diff
@@ -389,6 +389,47 @@ but can often be elided on and within interop types and on extension members as
...
+
+The above sections show how to call JS members from Dart. It's also useful to
+*export* Dart code so that it can be used in JS. The mechanism to do this
+involves exporting Dart functions, which requires a conversion like
+[`Function.toJS`] that wraps the Dart function with a JS function. Once
+converted, we can pass the resulting JS function to JS through an interop
+member, which can then be called later in JS.
```
RedBrogdon, line 399:
```suggestion
To export a Dart function to JS, first convert it using [`Function.toJS'], which wraps the Dart function with a JS function. Then, pass the wrapped function to JS through an interop member. At that point, it's ready to be called by other JS code.
```

src/content/interop/js-interop/usage.md
```diff
@@ -389,6 +389,47 @@ but can often be elided on and within interop types and on extension members as
...
+
+Functions that are exported this way have type [restrictions] similar to those
+of interop members.
+
+Sometimes it's useful to export an entire Dart interface so that JS can interact
+with a Dart object instead of individual functions. In order to do this, the
+class should be marked as exportable using [`@JSExport`] and an instance of that
+class should be wrapped using [`createJSInteropWrapper`]. For a more detailed
+look on how to do that and how it can be used to mock JS values, see the
+[mocking tutorial].
```
RedBrogdon, line 431:
```suggestion
Sometimes it's useful to export an entire Dart interface so that JS can interact
with a Dart object. To do this, mark the Dart class as exportable using [`@JSExport`] and wrap instances of that class using [`createJSInteropWrapper`]. For a more detailed
explanation of this technique, including how to mock JS values, see the
[mocking tutorial].
```

src/content/interop/js-interop/mock.md
```diff
@@ -97,9 +97,10 @@ void main() {
 
 `@JSExport` allows you to declare a class that can be used in
 `createJSInteropWrapper`. `createJSInteropWrapper` will create an object literal
-that maps each of the class' instance member names (or renames) to a JS callback
-that triggers the instance member when called. In the above example, getting and
-setting `counter.value` gets and sets `fakeCounter.value`.
+that maps each of the class' instance member names (or renames) to a JS
+callback, which is created using [`Function.toJS`]. This JS callback then calls
+the instance member when called. In the above example, getting and setting
```
RedBrogdon, line 102:
```suggestion
callback, which is created using [`Function.toJS`]. When called, the JS callback will in turn call
the instance member. In the above example, getting and setting
```

RedBrogdon (commented):
Looks great. I suggested a few changes to tighten up the wording, but that's it.

sfshaza2 (approved):
lgtm, once @RedBrogdon's issues are addressed.

src/content/interop/js-interop/usage.md
srujzs: Done.

src/content/interop/js-interop/usage.md
srujzs: Done, but I like keeping the first two sentences here to contrast what this section is doing compared to the rest of interop.

src/content/interop/js-interop/usage.md
```diff
@@ -389,6 +389,45 @@ but can often be elided on and within interop types and on extension members as
 the compiler can tell it is a JS interop type from the representation type and
 on-type.
 
+## Exporting Dart functions and objects to JS
+
+The above sections show how to call JS members from Dart. It's also useful to
+*export* Dart code so that it can be used in JS. To export a Dart function to
+JS, first convert it using [`Function.toJS'], which wraps the Dart function with
```
sigmundch, line 396: minor: seems from the last edits the second backtick was accidentally changed to a quote, causing this not to render as code and mismatch the link below.

sigmundch (approved):
Nice, thank you @srujzs !

src/content/interop/js-interop/usage.md
srujzs: Good catch, thanks!

srujzs:
FYI @RedBrogdon - I'm not authorized to merge, so feel free to merge if you think it's good!

PR #6016, opened 2024-08-01
Move away from deprecated shiki highlighting function

parlough:
https://github.com/shikijs/shiki/pull/702 has more details about the upstream deprecation and change.


sfshaza2 (approved):
lgtm

Issue #5995, opened 2024-07-26
Incorrect expansion of "Uri" and a highlighting issue in Effective Dart style guide


neko-tail:
### Page URL

https://dart.dev/effective-dart/style#do-capitalize-acronyms-and-abbreviations-longer-than-two-letters-like-words

### Page source

https://github.com/dart-lang/site-www/blob/main/src/content/effective-dart/style.md#do-capitalize-acronyms-and-abbreviations-longer-than-two-letters-like-words

### Describe the problem

> URI stands for uniform resource identifier. URLs (uniform resource locators) are a common kind of URI.

good:
```dart
Uri // "uniform resource locator"
```

bad:
```dart
URI // "uniform resource locator"
```

### Expected fix

good:
```dart
Uri // "uniform resource identifier"
```

bad:
```dart
URI // "uniform resource identifier"
```

or

good:
```dart
Url // "uniform resource locator"
```

bad:
```dart
URL // "uniform resource locator"
```

### I would like to fix this problem.

- [ ] I will try and fix this problem on dart.dev.

RedBrogdon:
Hrmm. Looks like there's a highlighting issue with "esq" in that block as well:

<img width="560" alt="image" src="https://github.com/user-attachments/assets/e36e9fbb-cbf7-4f5c-af6e-ae626ec7d121">

PR #6019, opened 2024-08-02
Fix expansion of Uri and mispelling of identifier

parlough:
Fixes https://github.com/dart-lang/site-www/issues/5995


sfshaza2 (approved):
Wow, a lot of mispellings! lgtm

PR #6024, opened 2024-08-04
Add a section to whats new on the site as of 3.5

parlough:
This will need to be updated with the actual date and links in the future, so marking as draft and will revisit after the release.


sfshaza2 (approved):
lgtm! Thanks for being so on top of this!

PR #6033, opened 2024-08-07
Account for removal of lint groups

parlough:
Lints no longer are part of three specific groups, but now have tags/groups (https://github.com/dart-lang/linter/issues/1020). This PR removes the expectation of the three groups so that the lints keep rendering. It also updates the source data to the latest version with no groups.

As part of the work in https://github.com/dart-lang/site-www/issues/4499, we'll surface the new categories/tags later on :) 

Closes https://github.com/dart-lang/sdk/issues/56396

**Staged index page:** https://dart-dev--pr6033-feat-lints-no-more-groups-b2yvk36h.web.app/tools/linter-rules

**Staged js-interop lint page:** https://dart-dev--pr6033-feat-lints-no-more-groups-b2yvk36h.web.app/tools/linter-rules/invalid_runtime_check_with_js_interop_types

### DIFF
--- a/src/_data/linter_rules.json
+++ b/src/_data/linter_rules.json
@@ -5,7 +5,6 @@
     "categories": [
       "style"
     ],
-    "group": "style",
     "state": "stable",
     "incompatible": [],
     "sets": [],
@@ -19,7 +18,6 @@
     "categories": [
       "style"
     ],
-    "group": "style",
     "state": "stable",
     "incompatible": [],
     "sets": [],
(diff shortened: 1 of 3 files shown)


kevmoo (approved):
Which SDK commit is this lints file from?

parlough:
> Which SDK commit is this lints file from?

I pulled it from `main` when I created this PR, so I believe https://github.com/dart-lang/sdk/commit/47180c004b1afb1da8df6dd6eb66bdb5233953ba. This includes my change from https://github.com/dart-lang/sdk/commit/c50c194d419c09bdbc9267eb8ed045819fb0beed.

The reason the JS lint showed as unreleased was because before this change its since information had `-wip` in it.

PR #5944, opened 2024-06-27
Update functions.md

RikinR:
added the " Function type " section in the file

Added the Function types section similar to one in Record file . 

Fixes #5901

---

- [x] I’ve reviewed the contributor guide and applied the relevant portions to this PR.
- [x] This PR doesn't contain automatically generated corrections or text (Grammarly, LLMs, and similar).
- [x] This PR follows the [Google Developer Documentation Style Guidelines](https://developers.google.com/style) — for example, it doesn't use _i.e._ or _e.g._, and it avoids _I_ and _we_ (first person).
- [x] This PR uses [semantic line breaks](https://github.com/dart-lang/site-shared/blob/main/doc/writing-for-dart-and-flutter-websites.md#semantic-line-breaks) of 80 characters or fewer.

<details>
  <summary>Contribution guidelines:</summary><br>

  - See our [contributor guide](https://github.com/dart-lang/site-www/blob/main/CONTRIBUTING.md) for general expectations for PRs.
  - Larger or significant changes should be discussed in an issue before creating a PR.
  - Code changes should generally follow the [Dart style guide](https://dart.dev/effective-dart) and use `dart format`.
  - Updates to [code excerpts](https://github.com/dart-lang/site-shared/blob/main/doc/code-excerpts.md) indicated by `<?code-excerpt` need to be updated in their source `.dart` file as well.
</details>

PR #6037, opened 2024-08-09
Function type declarations

### LINKED ISSUE #5901: Add section on Function type declarations to 'Functions' page
### Page URL

https://dart.dev/language/functions/

### Page source

https://github.com/dart-lang/site-www/tree/main/src/content/language/functions.md

### Describe the problem

The [functions page](https://dart.dev/language/functions) doesn't explain `Function` type declarations.

In contrast, the [records page](https://dart.dev/language/records) does explain [record types](https://dart.dev/language/records#record-types) for example.

### Expected fix

Add some explanation about how to use `Function` type declarations. (Including defining return types, parameter types, etc).

It might also be worth mentioning Function types on the [built-in types page](https://dart.dev/language/built-in-types).

### Additional context

_No response_

mmcdon20:
I believe https://github.com/dart-lang/site-www/pull/5944 is abandoned at this point. I am opening a new PR with the changes discussed in the previous PR.

Fixes #5901


mmcdon20:
Can we get a review on this?

sfshaza2 (approved):
@RedBrogdon, this lgtm to me. Can you give it a once over?

src/content/language/functions.md
```diff
@@ -266,6 +266,41 @@ assert(loudify('hello') == '!!! HELLO !!!');
 This example uses an anonymous function.
 More about those in the next section.
 
+## Function types
+
+We can specify the type of a function, which is known as a *function type*.
+It is actually the type of a first-class function object: An expression whose 
+type is a function type can only evaluate to a first-class function object.
```
RedBrogdon, line 273: These two sentences are difficult for me to parse. What do you think about something like "A function type consists of a return type (including `void`) and a list of zero or more parameters."?

src/content/language/functions.md
```diff
@@ -266,6 +266,41 @@ assert(loudify('hello') == '!!! HELLO !!!');
...
+void printElement(int element) => print(element);
+
+void doPrint(void Function(int) printFunction, {required List<int> list}) {
+  list.forEach(printFunction);
+}
+
+var list = [1, 2, 3];
+
+// Store `doPrint` in a variable and call it.
+void Function(void Function(int), {required List<int> list}) g = doPrint;
```
RedBrogdon, line 290: Is it possible to use a simpler example here, rather than one which involves declaring the type of a function that receives a function as one of its parameters?

It's a good to mention that you can do this, don't get me wrong, but I'm wondering if there's a simpler example to start with? Maybe something out of the SDK, like Flutter's WidgetBuilder typedef?

src/content/language/functions.md
```diff
@@ -266,6 +266,41 @@ assert(loudify('hello') == '!!! HELLO !!!');
...
+g(printElement, list: list);
+```
+
+:::note
+In Dart, functions are first-class objects,
+meaning they can be assigned to variables,
+passed as arguments, and returned from other functions.
+The [`typedef`][] keyword allows defining function types explicitly,
```
RedBrogdon, line 298: nit: "defines" rather than "allows defining"

examples/misc/lib/language_tour/functions.dart
```diff
@@ -26,6 +26,22 @@ void miscDeclAnalyzedButNotTested() {
...
+    void printElement(int element) => print(element);
+
+    void doPrint(void Function(int) printFunction, {required List<int> list}) {
+      list.forEach(printFunction);
+    }
+
+    var list = [1, 2, 3];
+
+    // Store `doPrint` in a variable and call it.
+    void Function(void Function(int), {required List<int> list}) g = doPrint;
```
RedBrogdon, line 40: Any particular reason for `g` as the variable name here?

src/content/language/functions.md
eernstg: Lines 272 and 273 describe the semantics of a function type, that is, what it means. Perhaps there's a need to define 'first-class function object', but that's probably difficult to cram into a few words.

A sentence like "A function type consists of ..." describes the syntax of the function type.

So why don't we just have both?

This would fit quite nicely with the next sentence where the syntax is described in a more indirect fashion.

src/content/language/functions.md
eernstg: Perhaps the example could be adjusted to emphasize the simple function type `void Function(int)` which is used in line 284, as well as the more complex one in line 290?

eernstg (commented):
Looks good! I added a couple of drive-by comments.

examples/misc/lib/language_tour/functions.dart
mmcdon20: Do you have a better suggestion for the variable name?

eernstg (approved):
Looks good!

It has been simplified substantially, and there's a very good, direct correspondence between the natural language explanation and the source code example.

I think it's important to be aware of the fact that any use of a function type essentially implies that someone must create a first-class function object (and that's considerably more expensive than just using a function which is known at compile time), but it doesn't have to be included right here.

mmcdon20:
@RedBrogdon can you review the changes?

src/content/language/functions.md
```diff
@@ -266,6 +266,34 @@ assert(loudify('hello') == '!!! HELLO !!!');
 This example uses an anonymous function.
 More about those in the next section.
 
+## Function types
+
+You can specify the type of a function, which is known as a *function type*.
+A function type is obtained from a function declaration header by replacing 
+the function name by the word `Function`. Moreover, you are allowed to omit
```
parlough, line 273:
```suggestion
the function name by the keyword `Function`. Moreover, you are allowed to omit 
```

src/content/language/functions.md
```diff
@@ -266,6 +266,34 @@ assert(loudify('hello') == '!!! HELLO !!!');
 This example uses an anonymous function.
 More about those in the next section.
 
+## Function types
+
+You can specify the type of a function, which is known as a *function type*.
+A function type is obtained from a function declaration header by replacing 
+the function name by the word `Function`. Moreover, you are allowed to omit 
+the names of positional parameters, but the names of named parameters cannot
```
parlough, line 274:
```suggestion
the names of positional parameters, but the names of named parameters can't 
```

src/content/language/functions.md
```diff
@@ -266,6 +266,34 @@ assert(loudify('hello') == '!!! HELLO !!!');
...
+void Function(String, {String greeting}) g = greet;
+g('Dash', greeting: 'Howdy');
+```
+
+:::note
+In Dart, functions are first-class objects,
+meaning they can be assigned to variables,
+passed as arguments, and returned from other functions.
+The [`typedef`][] keyword defines function types explicitly,
+which can be useful for clarity and reusability.
```
parlough, line 292: Since `typedef` can be used for other type aliases and the keyword doesn't act alone in defining the type, perhaps let's restructure this to focus on what developers can use it for. How does this sound?

```suggestion

You can use a [`typdef`][] declaration to explicitly name function types,
which can be useful for clarity and reusability.
```

parlough (approved):
I'm sorry about the delay on review and landing this @mmcdon20! I really appreciate you keeping this PR and enhancement alive.

Looks good to me, but a few minor suggestions and one question to consider. Thanks!

Issue #6055, opened 2024-08-22
Search on 404 page doesn't work


kevmoo:
Go to https://dart.dev/not_here
Search something: `try a search`
Sends user to `https://dart.dev/search.html?cx=011220921317074318178%3Ai4mscbaxtru&ie=UTF-8&hl=en&q=try+a+search`

Should send user to `https://dart.dev/search?cx=011220921317074318178%3A_yy-tmb5t_i&ie=UTF-8&hl=en&q=try+a+search` ???

RedBrogdon:
@parlough, do you have time to look at this in the next few days? If not, I can give it a go.

PR #6060, opened 2024-08-24
Redirect from path with html extension to base path

parlough:
Fixes https://github.com/dart-lang/site-www/issues/6055 by redirecting all links that end in `.html` to their base equivalent. This will also fix a few other 404s commonly seen in the analytics.


sfshaza2 (approved):
lgtm

Issue #6065, opened 2024-08-27
The Web helpers link is missing


KomosSolutions:
### Page URL

https://dart.dev/interop/js-interop/package-web/

### Page source

https://github.com/dart-lang/site-www/tree/main/src/content/interop/js-interop/package-web.md

### Describe the problem

The Web 'Helpers' link is missing, and apears to be deprecated: "To mitigate these differences, package:web contains [helpers](https://github.com/dart-lang/web/tree/main/lib/src/helpers).."

### Expected fix

Update the web helpers section

### I would like to fix this problem.

- [ ] I will try and fix this problem on dart.dev.

PR #6066, opened 2024-08-27
Update web helpers links to account for directory move

parlough:
Fixes https://github.com/dart-lang/site-www/issues/6065


kevmoo (approved):
you rock. thanks

kevmoo:
I agree this change is better than having 404s, but longer term we likely shouldn't be linking to the source code here.

CC @johnpryan who is pondering web API cleanup

Issue #6076, opened 2024-09-04
What can be referenced in a documentation comment reference?


srawlins:
### What information needs to be added?

There are some corners to what can be referenced in a doc comment reference (`/// [...]`). For example, an instance member can be referenced, qualified by it's enclosing class's name (e.g. `[Future.then]`, even though that is not a static member).

A generic type can be referred to by it's "base" name (`[List]`) but not with type arguments (`[List<int>]`). For a comment on a function, the function's parameters are in-scope. An unnamed constructor can be referred to by it's tear-off name (`[Completer.new]`). Doc imports add a scope. Imported elements can be referenced by their import prefix, if they have one (`[async.Future.value]`). Etc.

CC @dart-lang/analyzer-team 

### Where should this new content appear?

As far as I can tell, outside of Effective Dart, there is only one sentence on comment references between square brackets: https://dart.dev/language/comments#documentation-comments. This page is short, so the content could be another section on the page, or a separate page.

### I would like to fix this problem.

- [X] I will try and fix this problem on dart.dev.

srawlins:
This documentation specifically would be timely now as we're adding the "Documentation imports" feature for the analyzer and dartdoc: https://github.com/dart-lang/sdk/issues/56186.

If this team indicates that documentation would be accepted, I can start working on a PR 😁

parlough:
Yes this is something that I've long wanted, so it would definitely be accepted :D. There are some other dartdoc related things I'd like to document, so I'll give some thought on the best place for it.

Feel free to place the docs on their own page for now, perhaps under `/src/content/tools`.

PR #6080, opened 2024-09-09
Initial documentation for doc comment references

srawlins:
Resolves https://github.com/dart-lang/site-www/issues/6076

This does not discuss "doc imports"; I'd like that to be a separate addition.

**Staged:** https://dart-dev--pr6080-doc-comment-references-cx0x9ulm.web.app/tools/doc-comment-references


srawlins:
CC @dart-lang/analyzer-team for any comments.

src/content/tools/doc-comment-references.md
```diff
@@ -0,0 +1,82 @@
+---
+title: Documentation comment references
+short-title: Comment references
+description: Learn about doc comment references and their syntax.
+---
+
+Doc comments can contain references to various code elements. A library member
+or class member, etc. can be referenced by wrapping its name in square brackets
+(`[...]`) in a doc comment (a comment starting with `///`). Examples:
```
bwilkerson, line 9: Consider adding " or `/**`", given that it's technically correct.

src/content/tools/doc-comment-references.md
```diff
@@ -0,0 +1,82 @@
...
+There are several benefits to referring to a code element in a doc comment as a
+doc comment reference:
+
+### Editor features
+
+Doc comment references enable several IDE features. When an element is renamed
+via an IDE command, it can rewrite uses of that element, including references
+in doc comments. When an IDE lists all "references" to an element, it can
+include references in doc comments. An IDE can also provide Go-to-definition
+support at the location of a doc comment reference.
```
bwilkerson, line 33: It mIght be worth mentioning code completion within the reference.

src/content/tools/doc-comment-references.md
```diff
@@ -0,0 +1,82 @@
...
+
+```dart
+abstract class MyList<E> implements List<E> {
+  /// Refer to [add] and [contains], which is declared on [Iterable].
+  void myMethod() {}
+}
+```
+
+The doc comment for a function can include references to the function's
+parameters.
```
bwilkerson, line 75: Presumably also the type parameters of the function?

src/content/tools/doc-comment-references.md
srawlins: Ah ha! Great catch.

src/content/tools/doc-comment-references.md
srawlins: Yes, that is in the next sentence.

These are a little awkward, I want to say "The type parameters of things-that-have-type-parameters can be referenced" and "The parameters of things-that-have-parameters can be referenced." But it's not as general as referencing, for example, a static method on a class: you can refer to `Future.then` anywhere, but the _only_ place from which you can refer to a parameter or a type parameter when it is in scope; You cannot refer to the `Future.value. onValue` if it is out-of-scope.

OK, I've re-worded these to focus on "parameter" and "type parameter" and expanded with a few sentences.

srawlins:
Ready for review.

src/content/tools/doc-comment-references.md
```diff
@@ -0,0 +1,88 @@
...
+Future<T> g<T>(T object) => Future.value(object);
+```
+
+These doc comments contain references to the `String` class, the `object`
+parameter, and the `Future.value` constructor.
+
+## Features of references
+
+There are several benefits to referring to a code element in a doc comment as a
+doc comment reference:
```
parlough, line 25: I think we don't refer to a code element "as a reference", but "with a reference".



Also can we avoid the double "doc comment"?

```suggestion
There are several benefits to referring to code elements with doc comment references:
```

src/content/tools/doc-comment-references.md
```diff
@@ -0,0 +1,88 @@
...
+
+### Editor features
+
+Doc comment references enable several IDE features. An element's name can be
+code-completed within square brackets. When an element is renamed via an IDE
+command, it can rewrite uses of that element, including references in doc
+comments. When an IDE lists all "references" to an element, it can include
+references in doc comments. An IDE can also provide Go-to-definition support at
+the location of a doc comment reference.
```
parlough, line 34: Optionally consider making this a list, so readers can quickly see the features supported if they know them, and choose to learn more. For example:

```
Doc comment references enable several IDE features:

- **Code completion**
  An element's name can be code-completed within square brackets.
- **Rename refactoring**
  When an element is renamed via an IDE command, it can
  rewrite uses of that element, including references in doc comments.
```

src/content/tools/doc-comment-references.md
```diff
@@ -0,0 +1,88 @@
+---
+title: Documentation comment references
+short-title: Comment references
+description: Learn about doc comment references and their syntax.
+---
+
+Doc comments can contain references to various code elements. A library member
+or class member, etc. can be referenced by wrapping its name in square brackets
```
parlough, line 8: We try to [avoid use of etc](https://developers.google.com/style/word-list#etc) if possible.

```suggestion
Doc comments can contain references to various code elements.
Elements, such as functions and classes, can be referenced by wrapping their name in square brackets
```

src/content/tools/doc-comment-references.md
```diff
@@ -0,0 +1,88 @@
+---
+title: Documentation comment references
+short-title: Comment references
+description: Learn about doc comment references and their syntax.
+---
+
+Doc comments can contain references to various code elements. A library member
```
parlough, line 7: Describing doc comment references as referring to "elements" makes sense to me and I believe is what we've always said, but is that because of its implementation ties to the analyzer and our familiarity with it? I'm not necessarily saying we shouldn't use the term, but I just want to make sure it's the best term to use.

I can find very few cases in user-facing docs where we use the term "element" like this. Never do we say "code element". In fact, we are careful to use the term as it already has a few meanings (elements in a collection, collection literal elements, HTML elements, etc).

As a side note, [`comment_references`](https://dart.dev/lints/comment_references) lint docs mention "in-scope identifiers".

@bwilkerson for thoughts based on writing diagnostic messages.

\cc @leafpetersen For other considerations :)

src/content/tools/doc-comment-references.md
```diff
@@ -0,0 +1,88 @@
...
+
+These doc comments contain references to the `String` class, the `object`
+parameter, and the `Future.value` constructor.
+
+## Features of references
+
+There are several benefits to referring to a code element in a doc comment as a
+doc comment reference:
+
+### Editor features
```
parlough, line 27: Perhaps the end of this section or a tip aside can call out the [`comment_references`](https://dart.dev/lints/comment_references) lint as enabling verification of them?

```
:::tip
<Tip content goes here>
:::
```

Another option would be calling this out in the "What can be referenced" section.

parlough (commented):
Thanks so much @srawlins! It's super exciting to finally have this officially documented.

These docs look great to me and could land in the current state if you prefer, but I have a few comments, questions, and enhancements for you to consider.

Feel free to push back on any of them or defer them for later (or for me).

I'll try to open a separate PR this week to explore how we can layout this page alongside other doc-comment related pages.

src/content/tools/doc-comment-references.md
```diff
@@ -0,0 +1,88 @@
...
+```
+
+Most members of a class, an enum, an extension, an extension type, and a mixin
+can also be referenced. A reference to a member that is not in scope must be
+qualified with its container's name. For example the `wait` static method on
+the `Future` class can be referenced in a doc comment with `[Future.wait]`.
+This is true for instance members as well; the `add` method and the `length`
+property on the `List` class can be referenced with `[List.add]` and
+`[List.length]`. In-scope members like these can even be referenced without the
+qualifying container name:
```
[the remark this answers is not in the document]
srawlins: Great suggestion. Done.

src/content/tools/doc-comment-references.md
```diff
@@ -0,0 +1,88 @@
...
+alias for such a paramter's enclosing function type.
+
+Type parameters can be referenced in a doc comment only when they are in scope.
+Therefore a type parameter of a method, top-level function, or type alias can
+only be referenced within a doc comment on that element, and a type parameter
+of a class, enum, extension, extension type, and mixin can only be referenced
+within a doc comment on that element or on one of its members.
+
+The doc comment for a type alias which aliases a class, enum, extension type,
+or mixin cannot reference any of the aliased type's members, if it has any, as
```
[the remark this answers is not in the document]
srawlins: Done

src/content/tools/doc-comment-references.md
```diff
@@ -0,0 +1,88 @@
...
+linked, if possible, to the documentation page of the element being referenced.
+If the element does not have a documentation page (for example, a function
+parameter, a type variable, or a private class), then no link is created.
+
+## What can be referenced
+
+Most library members can be referenced in a doc comment, including classes,
+constants, enums, named extensions, extension types, functions, mixins, and
+type aliases. This includes all in-scope library members, either declared
+locally, or imported. Library members which are imported with an import prefix
```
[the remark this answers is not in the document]
srawlins: Done

src/content/tools/doc-comment-references.md
```diff
@@ -0,0 +1,88 @@
...
+can be referenced with the prefix. Examples:
```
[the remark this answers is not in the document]
srawlins: Done

src/content/tools/doc-comment-references.md
```diff
@@ -0,0 +1,88 @@
...
+referenced within a doc comment on such a parameter's function or on a type
+alias for such a paramter's enclosing function type.
+
+Type parameters can be referenced in a doc comment only when they are in scope.
+Therefore a type parameter of a method, top-level function, or type alias can
+only be referenced within a doc comment on that element, and a type parameter
+of a class, enum, extension, extension type, and mixin can only be referenced
+within a doc comment on that element or on one of its members.
+
+The doc comment for a type alias which aliases a class, enum, extension type,
```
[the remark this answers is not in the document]
srawlins: Done

src/content/tools/doc-comment-references.md
srawlins: Love it, done.

src/content/tools/doc-comment-references.md
srawlins: Great idea, done.

src/content/tools/doc-comment-references.md
srawlins: Done

src/content/tools/doc-comment-references.md
srawlins: I'll stick to "identifiers" instead of "code elements" for now. Good call. I think "element" might be more correct in other cases. For example, an _identifier_ is not referred to, a _class_ is, via it's _identifier_. It is messy though.

src/content/tools/doc-comment-references.md
srawlins: Done, love it.

src/content/tools/doc-comment-references.md
```diff
@@ -0,0 +1,88 @@
...
+Doc comment references enable several IDE features. An element's name can be
+code-completed within square brackets. When an element is renamed via an IDE
+command, it can rewrite uses of that element, including references in doc
+comments. When an IDE lists all "references" to an element, it can include
+references in doc comments. An IDE can also provide Go-to-definition support at
+the location of a doc comment reference.
+
+### API documentation
+
+In API documentation generated by `dart doc`, a doc comment reference is
```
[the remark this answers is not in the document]
srawlins: Oops, just saw this! Done

src/content/tools/doc-comment-references.md
```diff
@@ -0,0 +1,88 @@
...
+import 'dart:math' as math;
+
+/// [List] is in scope.
+/// So is [math.max].
+int x = 7;
+```
+
+Most members of a class, an enum, an extension, an extension type, and a mixin
+can also be referenced. A reference to a member that is not in scope must be
+qualified with its container's name. For example the `wait` static method on
```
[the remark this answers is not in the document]
srawlins: I like it, thanks!

src/content/tools/doc-comment-references.md
```diff
@@ -0,0 +1,88 @@
...
+abstract class MyList<E> implements List<E> {
+  /// Refer to [add] and [contains], which is declared on [Iterable].
+  void myMethod() {}
+}
+```
+
+Parameters of a function and parameters of a function type can be referenced in
+a doc comment only when they are in scope. They can therefore only be
+referenced within a doc comment on such a parameter's function or on a type
+alias for such a paramter's enclosing function type.
```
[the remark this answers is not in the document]
srawlins: Done

src/content/tools/doc-comment-references.md
```diff
@@ -0,0 +1,88 @@
...
+the location of a doc comment reference.
+
+### API documentation
+
+In API documentation generated by `dart doc`, a doc comment reference is
+linked, if possible, to the documentation page of the element being referenced.
+If the element does not have a documentation page (for example, a function
+parameter, a type variable, or a private class), then no link is created.
+
+## What can be referenced
```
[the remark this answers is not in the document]
srawlins: Great idea, I've started an attempt at this.

src/content/tools/doc-comment-references.md
```diff
@@ -0,0 +1,125 @@
...
+```
+
+Most members of a class, an enum, an extension, an extension type, and a mixin
+can also be referenced. A reference to a member that is not in scope must be
+qualified (prefixed) with its container's name. For example the `wait` static
+method on the `Future` class can be referenced in a doc comment with
+`[Future.wait]`. This is true for instance members as well; the `add` method
+and the `length` property on the `List` class can be referenced with
+`[List.add]` and `[List.length]`. In-scope members like these can even be
+referenced without the qualifying container name:
```
parlough, line 99: "In-scope members like these" makes it sound like the qualified examples in the previous sentence are "in-scope". Perhaps clarify with a conditional clause, something like:

```
When container members are in-scope, such as in an instance method's doc comment, they can be referenced without the qualifying container name:
```

src/content/tools/doc-comment-references.md
```diff
@@ -0,0 +1,125 @@
...
+
+Type parameters can be referenced in a doc comment only when they are in scope.
+Therefore a type parameter of a method, top-level function, or type alias can
+only be referenced within a doc comment on that element, and a type parameter
+of a class, enum, extension, extension type, and mixin can only be referenced
+within a doc comment on that element or on one of its members.
+
+The doc comment for a type alias that aliases a class, enum, extension type,
+or mixin can't reference any of the aliased type's members, if it has any, as
+if they were in scope.
```
parlough, line 125: I feel the aside of "if it has any" separates the "as if they were in scope" too much. I think it's fine to just drop it:

```suggestion
or mixin can't reference any of the aliased type's members as if they were in scope.
```

src/content/tools/doc-comment-references.md
```diff
@@ -0,0 +1,125 @@
...
+mis-uses. Keeping doc comment references valid ensures that these IDE features
+are enabled for each reference.
+:::
+
+### API documentation
+
+In API documentation generated by [`dart doc`](/tools/dart-doc), a doc comment reference is
+linked, if possible, to the documentation page of the element being referenced.
+If the element does not have a documentation page (for example, a function
+parameter, a type variable, or a private class), then no link is created.
```
parlough, line 55:
```suggestion
parameter, a type parameter, or a private class), then no link is created.
```

parlough (approved):
Thanks again for this @srawlins and the adjustments!

I have some final suggestions to consider then should be good to land. I'll work on a followup PR to add the PR to the sidenav with some other doc related entries.

src/content/tools/doc-comment-references.md
```diff
@@ -0,0 +1,125 @@
...
+
+These doc comments contain references to the `String` class, the `object`
+parameter, and the `Future.value` constructor.
+
+## Features of references
+
+There are several benefits to referring to code elements with doc comment
+references:
+
+### Editor features
```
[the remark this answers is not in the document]
srawlins: Love it

src/content/tools/doc-comment-references.md
```diff
@@ -0,0 +1,125 @@
...
+
+Doc comment references enable several IDE features:
+
+- **Code completion**
+  An element's name can be code-completed within square brackets.
+- **Rename refactoring**
+  When an element is renamed via an IDE command, it can rewrite uses of that
+  element, including references in doc comments.
```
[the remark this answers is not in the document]
srawlins: Done

src/content/tools/doc-comment-references.md
```diff
@@ -0,0 +1,125 @@
...
+- **Find references**
+  When an IDE lists all "references" to an element, it can include references in
+  doc comments.
+- **Go to definition**
+  An IDE can also provide Go-to-definition support at the location of a doc
+  comment reference.
+
+:::tip
+The [comment_references](https://dart.dev/lints/comment_references) lint rule
```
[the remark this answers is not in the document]
srawlins: Done

src/content/tools/doc-comment-references.md
srawlins: Done

src/content/tools/doc-comment-references.md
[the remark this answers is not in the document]
srawlins: SG. I'm sure there's a good solution in here. Maybe just experiment with different table layouts.

src/content/tools/doc-comment-references.md
srawlins: I like it. Done.

src/content/tools/doc-comment-references.md
srawlins: Yeah good call. Done.

srawlins (commented):
Thanks again!

parlough:
Thanks again Sam! This is now live with a redirect to it: https://dart.dev/to/doc-comment-references

I hope to get time to work on some related docs soon so it has a place it live in the sidenav.

PR #6089, opened 2024-09-16
Update diagnostic docs as of 2024-09-15

parlough:


src/content/tools/diagnostic-messages.md
```diff
@@ -10710,7 +10710,7 @@ abstract class C {
 
 ### invalid_null_aware_operator
 
-_The receiver can't be null because of short-circuiting, so the null-aware
+_The receiver can't be 'null' because of short-circuiting, so the null-aware
```
domesticmouse, line 10713: Backticks for code font?

domesticmouse (approved):

src/content/tools/diagnostic-messages.md
parlough: This comes from the SDK, but I believe diagnostic messages use single quotes on purpose. I forget exactly why, but I think it is due to how IDEs surface the messages and currently can't or don't apply Markdown formatting.

Issue #5254, opened 2023-10-17
Covariant / Covariance definition


MaryaBelanger:
Probably not too critical, but in the soon to be released diagnostics for [Extension Types](https://github.com/dart-lang/language/blob/main/accepted/future-releases/extension-types/feature-specification.md), one mentions "[type parameters used in a non-covariant position](https://dart-review.googlesource.com/c/sdk/+/322442/4/pkg/analyzer/messages.yaml#11489)". 

I searched dart.dev to explain "covariant positions", but didn't find an explicit "this is what covariance means and these are / aren't the positions".

Might be a matter of adding an explicit definition to the Glossary, or changing the [covariant keyword section's](https://dart.dev/guides/language/sound-problems#the-covariant-keyword) wording slightly to explicitly say "Covariance is...".

I also think that covariant keyword section would probably be better suited somewhere else, like in the language tour, maybe the Types section? Or fitted in somewhere with [the deconstruction of the Type System page](https://github.com/dart-lang/site-www/issues/4745) we're planning.

eernstg:
Covariant/contravariant/invariant positions are defined precisely in section 15.1 of [the bleeding edge language specification](https://spec.dart.dev/DartLangSpecDraft.pdf), page 83++, [here as source](https://github.com/dart-lang/language/blob/3cdbe58a7d30e21d9753d201e9cd3ed4da6dab5a/specification/dartLangSpec.tex#L7361).

That's rather heavy reading, but it might work to have (1) conceptual description, rules of thumb, examples, and (2) a reference to the specification.

MaryaBelanger:
Awesome, that helps a lot! Will definitely link to the spec and try a small, high-level summary for the glossary.

parlough:
Speaking to the keyword location, I also think the current situation is not great. It's not super consistent or discoverable, but maybe that's slightly by design as it is not super common and it does cause a loss of some type safety. 

I think the long term solution is to place it when overhauling our type system documentation, but if you feel it's better in the current type system page, I'd be fine with that for now too :)

parlough:
@eernstg I was wondering if you'd be willing to add some initial definitions to the glossary ([page](https://dart.dev/resources/glossary), [source](https://github.com/dart-lang/site-www/blob/main/src/content/resources/glossary.md)) for these words in the context of Dart?

Even if not fully fleshed out, they would be quite helpful as a reference to point others to in various discussions and proposals, both in the site and otherwise.

[...]

PR #6105, opened 2024-09-24
Add section about variance and positions

eernstg:
Resolves https://github.com/dart-lang/site-www/issues/5254.

**Staged:** https://dart-dev--pr6105-glossary-variance-sep24-fxrxoasd.web.app/resources/glossary#variance-and-variance-positions


parlough:
Sorry about the delay in reviewing this, I hope to get to it soon! Thanks again for working on this :D

src/content/resources/glossary.md
```diff
@@ -336,3 +336,51 @@ class A implements B {} // A is a subtype of B, but NOT a subclass of B.
...
+`Object Function(int)`, but it is a supertype of `int Function(Object)`.
+
+This makes sense if you consider their substitutability (see the section
+about subtypes): If you expect at compile time to work with a function of
+type `int Function(int)` then it is OK to actually work with a function of
+type `int Function(Object)` at run time: You will pass an `int` to it (and
+that's fine, it accepts any `Object`), and it returns an `int` (which is
+exactly what you'd expect). Hence, `int Function(Object)` is a subtype of
+`int Function(int)`. Note that everything is turned upside-down when we
+consider a parameter type.
```
parlough, line 366: By "turned upside-down" do you mean "reversed" (as in the opposite) and what does "everything" here mean? The subtype relationship described in the preceding prose or the logic followed?

src/content/resources/glossary.md
```diff
@@ -336,3 +336,51 @@ class A implements B {} // A is a subtype of B, but NOT a subclass of B.
...
+When we consider a more complex type like `List<int Function(int)>` we need
+to talk about the _positions_ in the type. This is basically just a matter
+of turning one of the parts of the type into a placeholder (let's use `_`
+for that), and then consider what happens to the type when we put different
+types into that location in the type.
+
+For example, consider `List<int Function(_)>` as a template for a type
+where we can put different types into the location where `_` occurs.
+
+This type is contravariant in that position: For example,
```
parlough, line 377: Consider combining these two paragraphs, as the second one heavily relies on the first one. With the separation, it might not be clear what "This type" or "that position" refers to.

src/content/resources/glossary.md
```diff
@@ -336,3 +336,51 @@ class A implements B {} // A is a subtype of B, but NOT a subclass of B.
...
+consider a parameter type.
+
+When we consider a more complex type like `List<int Function(int)>` we need
+to talk about the _positions_ in the type. This is basically just a matter
+of turning one of the parts of the type into a placeholder (let's use `_`
+for that), and then consider what happens to the type when we put different
+types into that location in the type.
+
+For example, consider `List<int Function(_)>` as a template for a type
+where we can put different types into the location where `_` occurs.
```
parlough, line 375:
```suggestion
where you can put different types in place of the placeholder (`_`).
```

src/content/resources/glossary.md
parlough: Also is the following example meant to explain _why_ its contravariant in that position or is it _because_ it's contravariant in that position?

My understanding is that it's an example of understanding _why_ that position is contravariant, but it's not clear to me from reading. Perhaps flipping the order of the because clauses and starting at the `Object` and `int` parameter type relation, might make that more clear.

src/content/resources/glossary.md
```diff
@@ -336,3 +336,51 @@ class A implements B {} // A is a subtype of B, but NOT a subclass of B.
...
+
+This type is contravariant in that position: For example, 
+`List<int Function(Object)>` is a subtype of `List<int Function(int)>`
+because `int Function(Object)` is a subtype of `int Function(int)` because
+`int` is a subtype of `int` (the return type) and `Object` is a _supertype_
+of `int` (the parameter type).
+
+This is the general case. In practice, it's often sufficient to know that
```
parlough, line 383: The general case of what? The position substitution exercise?

parlough (commented):
Thanks so much for working on this @eernstg! I'm not very familiar with these concepts, so I naturally read it from a perspective of someone brand new to these terms. As a result, my questions and suggestions might be incorrect, so please feel free to push back on anything. At the very least, I hope my confusions and potential misunderstandings can at least help you identify where readers might go astray.

I left a few other small comments for the second pass so I can validate my understanding before leaving misleading suggestions. :) 

Thanks again and sorry about the delay in reviewing this! I'm super excited to have something tangible in the docs to reference for these terms, and I can see it helping others as it was already insightful to me.

src/content/resources/glossary.md
```diff
@@ -336,3 +336,51 @@ class A implements B {} // A is a subtype of B, but NOT a subclass of B.
 
 class C extends D {} // C is a subtype AND a subclass of D.
 ```
+
+## Variance and variance positions
+
+A type parameter of a class (or a mixin, etc) is said to be _covariant_
```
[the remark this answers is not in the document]
eernstg: I think that's fine.

I plan to introduce a specific phrase for this in the language specification, perhaps 'type-introducing declaration'. The point is that a class declaration introduces a class, and it also introduces a type (and it introduces a family of classes and types if it's generic). A class is an entity that exists at run-time, and an entity which is capable of creating new objects. On top of that, it also introduces a namespace that contains static members. We don't actually have any (pure) type declarations, it's always declaring several other things plus a type that describes those other things. But I think it'll be OK to say 'type declaration' here.

src/content/resources/glossary.md
```diff
@@ -336,3 +336,51 @@ class A implements B {} // A is a subtype of B, but NOT a subclass of B.
...
+In Dart, all type parameters of all class, mixin, mixin class, and enum
+declarations are covariant.
+
+However, function types are different: A function type is covariant in the
+return type, but the opposite (known as _contravariant_) in its parameter
+types.  For example, the type `int Function(int)` is a subtype of the type
+`Object Function(int)`, but it is a supertype of `int Function(Object)`.
+
+This makes sense if you consider their substitutability (see the section
+about subtypes): If you expect at compile time to work with a function of
```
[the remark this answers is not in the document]
eernstg: Of course, I think I was looking for the actual syntax used to write a link in this context, but forgot. I think it's correct now: `[substitutability](#subtype)`. Done.

src/content/resources/glossary.md
```diff
@@ -336,3 +336,51 @@ class A implements B {} // A is a subtype of B, but NOT a subclass of B.
...
+type `int Function(Object)` at run time: You will pass an `int` to it (and
+that's fine, it accepts any `Object`), and it returns an `int` (which is
+exactly what you'd expect). Hence, `int Function(Object)` is a subtype of
+`int Function(int)`. Note that everything is turned upside-down when we
+consider a parameter type.
+
+When we consider a more complex type like `List<int Function(int)>` we need
+to talk about the _positions_ in the type. This is basically just a matter
+of turning one of the parts of the type into a placeholder (let's use `_`
+for that), and then consider what happens to the type when we put different
```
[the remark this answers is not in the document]
eernstg: Done.

src/content/resources/glossary.md
```diff
@@ -336,3 +336,51 @@ class A implements B {} // A is a subtype of B, but NOT a subclass of B.
 
 class C extends D {} // C is a subtype AND a subclass of D.
 ```
+
+## Variance and variance positions
+
+A type parameter of a class (or a mixin, etc) is said to be _covariant_
+when the type as a whole "co-varies" with the actual type argument, that
+is: if we replace the type argument by a subtype then the type as a whole
+is also a subtype.
```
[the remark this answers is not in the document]
eernstg: Done.

src/content/resources/glossary.md
```diff
@@ -336,3 +336,51 @@ class A implements B {} // A is a subtype of B, but NOT a subclass of B.
...
+
+For example, the type parameter of the class `List` is covariant because
+list types co-vary with their type argument: `List<int>` is a subtype of
+`List<Object>` because `int` is a subtype of `Object`.
+
+In Dart, all type parameters of all class, mixin, mixin class, and enum
+declarations are covariant.
+
+However, function types are different: A function type is covariant in the
```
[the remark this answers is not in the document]
eernstg: Indeed; done.

src/content/resources/glossary.md
```diff
@@ -336,3 +336,51 @@ class A implements B {} // A is a subtype of B, but NOT a subclass of B.
...
+return type, but the opposite (known as _contravariant_) in its parameter
+types.  For example, the type `int Function(int)` is a subtype of the type
+`Object Function(int)`, but it is a supertype of `int Function(Object)`.
+
+This makes sense if you consider their substitutability (see the section
+about subtypes): If you expect at compile time to work with a function of
+type `int Function(int)` then it is OK to actually work with a function of
+type `int Function(Object)` at run time: You will pass an `int` to it (and
+that's fine, it accepts any `Object`), and it returns an `int` (which is
+exactly what you'd expect). Hence, `int Function(Object)` is a subtype of
```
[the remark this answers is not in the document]
eernstg: Used that, and did one more iteration on the text.

src/content/resources/glossary.md
eernstg: The point is that if `X` occurs in a covariant position in `T` then `X` occurs in a contravariant position in `void Function(T)`, and if `Y` occurs in a contravariant position in `T` then `Y` occurs in a covariant position in `void Function(T)`. Finally, if `Z` occurs in an invariant position in `T` then it occurs in an invariant position in `void Function(T)`.

That's rather long and tedious, so I just said "everything" is reversed.

I don't know if it's a good idea to elaborate on this (to make this particular point explicit) or it's better to leave it as imprecise as it is currently. I suspect that the nested positions are much, much rarer than the basic ones (so people will write and/or encounter types like `void Function(T)`, but they are much less likely to encounter `void Function(void Function(T))`). If this is true then they won't ever need to use this "reverse everything" rule.

I expanded a bit on the text here.

src/content/resources/glossary.md
eernstg: Combined the two paragraphs.

I changed the return type from `int` to `void` in the following examples, in order to be able to talk about `int` as a type that is unambiguously associated with a single position in the type (avoiding the need to say "the first `int`" vs. "the second `int`" and such).

> is the following example meant to explain _why_ its contravariant in that position

The text is intended to explain how 'contravariant position' is defined, by example. This is in itself an example, because it is intended to define 'contravariant position' as well as 'covariant position', based on a generalization which the reader would probably perform automatically, based on the previous explanations. I added a paragraph where 'covariant position' is defined briefly, but explicitly, in order to improve on this situation.

The text defines 'contravariant position' by explaining why we must conclude that the overall type `List<void Function(_)>` varies in the opposite direction of the variation at `_`, and then states that this by definition makes the position of `_` contravariant.

src/content/resources/glossary.md
eernstg: Done.

src/content/resources/glossary.md
eernstg: I removed that sentence. I think the new text is clearer. ;-)

eernstg (commented):
Made several changes based on the review.

src/content/resources/glossary.md
```diff
@@ -336,3 +336,68 @@ class A implements B {} // A is a subtype of B, but NOT a subclass of B.
...
+consider the _positions_ in the type. This is basically just a matter of
+turning one of the parts of the type into a placeholder (let's use `_` for
+that), and then consider what happens to the type when different types are
+placed in that position.
+
+For example, consider `List<void Function(_)>` as a template for a type
+where you can put different types in place of the placeholder `_`. This
+type is contravariant in the position where that placeholder occurs.
+
+We can illustrate this as follows. `List<void Function(Object)>` is a
```
parlough, line 385: With your adjustments, it's much more clear what this example is illustrating, thanks!

It might help further if we're extra specific about how we applied the suggested process in this example. Perhaps something like:

```
The following logic illustrates this by substituting `Object` and `int` for `_`.
```

This reference back might also help readers who jumped ahead know they should go back to the previous paragraph for more context.

src/content/resources/glossary.md
```diff
@@ -336,3 +336,68 @@ class A implements B {} // A is a subtype of B, but NOT a subclass of B.
...
+Note that everything is turned upside-down for parameter types. In
+particular, this subtype relation among function types requires that the
+_opposite_ subtype relation exists for the parameter type (for example,
+`void Function(Object)` is a subtype of `void Function(int)` because `int`
+is a subtype of `Object`).
+
+With a more complex type like `List<void Function(int)>`, you have to
+consider the _positions_ in the type. This is basically just a matter of
+turning one of the parts of the type into a placeholder (let's use `_` for
+that), and then consider what happens to the type when different types are
```
parlough, line 378: Generally it's best to avoid phrasing like "basically just" as it don't add much meaning, and this process might not be basic for some readers. https://developers.google.com/style/word-list#simple

How about we be more prescriptive in how to do this:

```suggestion
consider the _positions_ in the type. To accomplish this,
turn one of the parts of the type into a placeholder,
and then consider what happens to the type when different types are
```

In this suggestion I also proposed dropping the "(let's use `_` for that)" to keep this sentence shorter and focused. The next paragraph can introduce that choice.

parlough (approved):
Thanks so much for the answers to my questions and your edits @eernstg! This is looking great :D

I have a few follow-up comments for you to consider, but nothing blocking. I'm excited for us to have this useful reference!

src/content/resources/glossary.md
eernstg: Great point, thanks!

src/content/resources/glossary.md
eernstg: That works, thanks!

eernstg (commented):
That's much better, thanks!

eernstg:
@parlough, I think this could be landed now. WDYT?

PR #6108, opened 2024-09-25
Explain `pub get --enforce-lockfile`

sigurdm:
https://github.com/dart-lang/pub/issues/4093


src/content/guides/packages.md
```diff
@@ -153,6 +153,81 @@ of all of its dependencies.
 Checking in the lockfile also ensures that your deployed app
 uses the same versions of code.
 
+:::note
+In some cases `dart pub get` will not give you the exact versions locked in
+`pubspec.lock`. For example:
+
+* If new dependencies are added to `pubspec.yaml` since `pubspec.lock` was created.
+* If the locked version no longer exists in the package repository.
+* If you upgraded the dart sdk, and some packages no longer are compatible with
```
jonasfj, line 162:
```suggestion
* If you upgraded the Dart SDK, and some packages no longer are compatible with
```

src/content/guides/packages.md
```diff
@@ -153,6 +153,81 @@ of all of its dependencies.
...
+  the sdk.
+  
+In all these cases `dart pub get` will
+
+* Attempt to unlock as few dependencies as possible.
```
jonasfj, line 167:
```suggestion
* Unlock as needed.
```

src/content/guides/packages.md
```diff
@@ -153,6 +153,81 @@ of all of its dependencies.
...
+For more information see:
+https://dart.dev/go/content-hashes
+Would change 1 dependency.
+Unable to satisfy `pubspec.yaml` using `pubspec.lock`.
+
+To update `pubspec.lock` run `dart pub get` without
+`--enforce-lockfile`.
+```
+
+:::
```
jonasfj, line 229: This is a very long `:::note` is it not?

src/content/guides/packages.md
```diff
@@ -153,6 +153,81 @@ of all of its dependencies.
...
+```console
+$ dart pub get
+Resolving dependencies... (1.0s)
+Downloading packages... 
++ retry 3.1.2
+```
+
+Also if the hash a published version of a package does not match the hash in
+`pubspec.lock`, `pub` will warn you, but fix the pubspec.lock to reflect the
+published hash:
```
jonasfj, line 180:
```suggestion
Also if the hash of a published version does not match the hash in
`pubspec.lock`, `pub` will warn you, but fix `pubspec.lock` to reflect the
published version:
```

src/content/guides/packages.md
```diff
@@ -153,6 +153,81 @@ of all of its dependencies.
...
+
+For more information see:
+https://dart.dev/go/content-hashes
+Changed 1 dependency!
+```
+
+For deploying your project to production, or for ensuring consistent CI runs,
+use `dart pub get --enforce-lockfile` - that will fail with a non-zero exit code
+if the current lockfile cannot be exactly fulfilled, and does not match your
+current constraints or if there is a hash mismatch.
```
jonasfj, line 206:
```suggestion
When deploying your project to production use `dart pub get --enforce-lockfile`.
This will fail, if dependencies can't be satisfied with the exact versions and hashes in `pubspec.lock`. Thus, you avoid deploying untested dependencies to production.
```

src/content/guides/packages.md
jonasfj: maybe it's not a problem.

jonasfj (approved):
LGTM; from my perspective.

Not sure if it should be structured some other way -- it is a long `:::note`.

src/content/guides/packages.md
sfshaza2: That is a long note! Maybe turn it into a header/section?

Thanks, @jonasfj, for that thorough review!

sigurdm: Yeah - that note got a bit out of hand :) Upgraded to a section

sigurdm:
@sfshaza2 feel free to land if you are happy.

src/content/guides/packages.md
```diff
@@ -178,6 +178,78 @@ due to conflicting version constraints in the pubspec.
 To identify out-of-date packages that require editing the pubspec,
 use [`dart pub outdated`][outdated].
 
+## Getting dependencies in production - `pub get --enforce-lockfile`
```
parlough, line 181: Can we keep the header simpler so it's easier to parse in the TOC?

Also I'm assuming most developers aren't getting dependencies "in production" but rather "for production".

```suggestion
## Get dependencies for production
```

src/content/guides/packages.md
```diff
@@ -178,6 +178,78 @@ due to conflicting version constraints in the pubspec.
...
+
+In some cases `dart pub get` will not give you the exact versions locked in
+`pubspec.lock`. For example:
```
parlough, line 184: Does the following list include all possible situations or just some? If it covers the major cases, I think it's probably fine to lead directly into the list:

```suggestion
In some situations, `dart pub get` does not retrieve
the exact package versions locked in the `pubspec.lock` file:
```

src/content/guides/packages.md
```diff
@@ -178,6 +178,78 @@ due to conflicting version constraints in the pubspec.
...
+
+* If new dependencies are added to `pubspec.yaml` since `pubspec.lock` was created.
```
parlough, line 186:
```suggestion
* If new dependencies are added to `pubspec.yaml` after
  the `pubspec.lock` file was last updated.
```

src/content/guides/packages.md
```diff
@@ -178,6 +178,78 @@ due to conflicting version constraints in the pubspec.
...
+* If the locked version no longer exists in the package repository.
+* If you upgraded the Dart SDK, and some packages no longer are compatible with
+  the sdk.
```
parlough, line 189: Assuming this can also happen on downgrading SDK versions:

```suggestion
* If you changed to a different version of the Dart SDK,
  and some packages are no longer compatible with that new version.
```

src/content/guides/packages.md
```diff
@@ -178,6 +178,78 @@ due to conflicting version constraints in the pubspec.
...
+  
+In all these cases `dart pub get` will
```
parlough, line 191:
```suggestion
In these cases `dart pub get` will:
```

src/content/guides/packages.md
```diff
@@ -178,6 +178,78 @@ due to conflicting version constraints in the pubspec.
...
+
+* Unlock as needed.
```
parlough, line 193: Can you expand on what this means? Repeat dependency resolution keeping as many other locked versions the same?

src/content/guides/packages.md
```diff
@@ -178,6 +178,78 @@ due to conflicting version constraints in the pubspec.
...
+* Notify about any changes in the output.
```
parlough, line 194: It might be helpful to clarify this is talking about package dependency changes, not changes to the textual output.

Does the following make sense?

```suggestion
* Notify you about differences in retrieved packages and the lockfile.
```

src/content/guides/packages.md
```diff
@@ -178,6 +178,78 @@ due to conflicting version constraints in the pubspec.
...
+
+For example after adding `retry: ^3.0.0` to your dependencies:
```
parlough, line 196:
```suggestion
For example, after adding `retry: ^3.0.0` to your dependencies:

```

src/content/guides/packages.md
```diff
@@ -178,6 +178,78 @@ due to conflicting version constraints in the pubspec.
...
+```console
+$ dart pub get
+Resolving dependencies... (1.0s)
+Downloading packages... 
++ retry 3.1.2
+```
+
+Also if the hash of a published version does not match the hash in
+`pubspec.lock`, `pub` will warn you, but fix `pubspec.lock` to reflect the
+published version:
```
parlough, line 206: How about we cross-link to the glossary entry about pub content hashes?

```suggestion
Also, if the [content hash](/tools/pub/glossary#content-hashes) of a 
published package version differs from the hash in the `pubspec.lock` file,
`pub` will warn you and update the lockfile to reflect the published version:
```

src/content/guides/packages.md
```diff
@@ -178,6 +178,78 @@ due to conflicting version constraints in the pubspec.
...
+
+For example if we manually change a hash in `pubspec.lock`:
```
parlough, line 208: We [avoid use of "we"](https://developers.google.com/style/word-list#we).

```suggestion
For example, if you manually change the hash of `retry` in `pubspec.lock`:
```

src/content/guides/packages.md
```diff
@@ -178,6 +178,78 @@ due to conflicting version constraints in the pubspec.
...
+ * The pubspec.lock has been corrupted.
+
+The content-hashes in pubspec.lock has been updated.
+
+For more information see:
+https://dart.dev/go/content-hashes
+Changed 1 dependency!
+```
+
+When deploying your project to production use `dart pub get --enforce-lockfile`.
```
parlough, line 229:
```suggestion
When deploying your project to production,
use `dart pub get --enforce-lockfile` to retrieve dependencies.
```

src/content/guides/packages.md
```diff
@@ -178,6 +178,78 @@ due to conflicting version constraints in the pubspec.
...
+This will fail, if dependencies can't be satisfied with the exact versions and hashes in `pubspec.lock`. Thus, you avoid deploying untested dependencies to production.
```
parlough, line 230: If possible, put the conditional clause first and be more specific about what "this" is. Would "package retrieval" be an appropriate replacement for "this" here, or perhaps "the command"?

Also, is it worth clarifying "dependencies and dependency versions"? As it seems both would be caught.

A possible alternative with some suggestions:

```suggestion
If your project's dependency constraints can't be satisfied with
the exact versions and content hashes in `pubspec.lock`,
package retrieval and the command will fail.
This helps avoid deploying untested
dependencies and dependency versions to production.
```

parlough (commented):
This is great docs and advice @sigurdm, thanks so much and sorry about the delay in reviewing this!

The change mostly looks good to me, I just have a few questions and some optional suggestions to consider.

src/content/guides/packages.md
sigurdm: I think these (together with the hash mismatch) are the major ones.

src/content/guides/packages.md
sigurdm: Yeah!

src/content/guides/packages.md
sigurdm: Tried clarifying

src/content/guides/packages.md
sigurdm: Tried clarifying

sigurdm:
Not rambling at all, all the comments had great value!

parlough (approved):
Thanks so much for those adjustments @sigurdm! Looks great :D

PR #6159, opened 2024-10-18
Update for new core lint diagnostic docs

parlough:
These docs were copied from, and therefore already reviewed in the SDK.


sfshaza2 (approved):
lgtm

PR #6162, opened 2024-10-22
Move glossary definitions to yaml format

parlough:
Converts the site-wide glossary to be hosted in a YAML file. Initially, only a few of the fields are used:

- `term` - for the header entry and default fragment
- `id` - _(optional)_ to override the default fragment
- `long_description` - (but will fallback to `short_description` if missing)
- `related_links` - For linking to other docs

A few other fields have been added for experimentation but are not yet used or surfaced.

- `short_description` - A shorter description to be shown on hover
- `category` - A primary category for the term
- `labels` - Labels/topics that can be used for filtration
- `alternate` - Alternate and related terms that should lead to this entry. Could be because they are synonyms or that term doesn't yet have a specific entry of its own, but is still relevant enough.

Also new to the page:

- A horizontal line between entries to help visually separate them.
- A related links section after the long description.

The current style of this information will likely be completely replaced as more information is surfaced and filtering is added.

---

This work is roughly based on the design proposed in https://flutter.dev/go/glossary-expansion and the [Fuchsia glossary](https://fuchsia.dev/fuchsia-src/contribute/docs/glossary-entries).

Resolves https://github.com/dart-lang/site-www/issues/6143

---

**Staged:** https://dart-dev--pr6162-feat-6143-bema0uix.web.app/resources/glossary


sfshaza2 (approved):
lgtm

MaryaBelanger (approved):
Looking forward to this coming together!

PR #6165, opened 2024-10-23
Loosen return type guidelines for local functions

natebosch:
Local named functions do infer a return type in current Dart versions.

In our recommendation to annotate the return type of functions change
the set of functions the guideline applies to from "named" functions to
"non-local" functions, anonymous function expressions are also local to
a method body.


munificent (approved):
Good catch!

src/content/effective-dart/design.md
```diff
@@ -1200,10 +1200,10 @@ makeGreeting(String who) {
 }
 ```
 
-Note that this guideline only applies to *named* function declarations:
-top-level functions, methods, and local functions. Anonymous function
-expressions infer a return type from their body. In fact, the syntax doesn't
-even allow a return type annotation.
+Note that this guideline only applies to *non-local* function declarations:
+top-level functions, and methods. Local functions and anonymous function
```
lrhn, line 1204: Maybe "top-level functions, static functions and instance methods".

I don't usually call static functions "methods", although I guess "static method" is commonly used for them. (I think it's a Java way of phrasing it, which began because they don't/didn't have plain functions at all.)

lrhn (approved):

MaryaBelanger (approved):
Lgtm, feel free to wait for Parker's review or merge now that you have approval. Thanks!

natebosch:
@MaryaBelanger - I don't have authorization to merge in this repo. Can you merge for me?

lrhn:
Me neither. @munificent?

parlough (approved):
Thanks for this update!

PR #6176, opened 2024-10-28
Consolidate category and labels fields in glossary

parlough:
@MaryaBelanger I originally felt like the separation was useful, but it quickly became murky as I began to add more entries that fit well into multiple categories. We don't use these yet, but let's keep it simple and have one field. If we want to surface some predetermined "categories" to filter by, we can separately specify important labels.

This PR also adds an extra new line between each entry for some visual separation while editing.


MaryaBelanger (approved):
Sgtm

PR #6206, opened 2024-11-14
Update documentation for generative constructor with optional positio…

danielerat:
Update documentation for generative constructor with optional positional parameters

- Updated the example for generative constructors to reflect the use of optional positional parameters with default values.
- Clarified the behavior of the constructor to indicate that `x` and `y` have default values of 2.0 when not passed explicitly.

### DIFF
--- a/src/content/language/constructors.md
+++ b/src/content/language/constructors.md
@@ -55,8 +55,8 @@ class Point {
   double x;
   double y;
 
-  // Generative constructor with optional positional parameters and default values.
-  Point([this.x = 2.0, this.y = 2.0]);
+  // Generative constructor with initializing formal parameters:
+  Point(this.x, this.y);
 }
 ```
 
--- a/src/content/language/constructors.md
+++ b/src/content/language/constructors.md
@@ -51,12 +51,12 @@ To instantiate a class, use a generative constructor.
 <?code-excerpt "point_alt.dart (idiomatic-constructor)" plaster="none"?>
 ```dart
 class Point {
-  // Initializer list of variables and values
-  double x = 2.0;
-  double y = 2.0;
+  // Instance variables to hold the coordinates of the point.
+  double x;
+  double y;
 
-  // Generative constructor with initializing formal parameters:
-  Point(this.x, this.y);
+  // Generative constructor with optional positional parameters and default values.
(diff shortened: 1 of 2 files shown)


src/content/language/constructors.md
```diff
@@ -51,12 +51,12 @@ To instantiate a class, use a generative constructor.
 <?code-excerpt "point_alt.dart (idiomatic-constructor)" plaster="none"?>
```
MaryaBelanger, line 51: Hey @danielerat, thanks for your PR. Our code-excerpts correspond to Dart files in the `/examples` folder of this repo; this one corresponds to [this "docregion" in this file](https://github.com/dart-lang/site-www/blob/main/examples/misc/lib/language_tour/classes/point_alt.dart#L7). Could you update that to match your changes as well? And then run  `./dash_site refresh-excerpts` in your terminal (instructions in the site readme [here](https://github.com/dart-lang/site-www?tab=readme-ov-file#refresh-code-excerpts)). Thanks!

danielerat: Yes sure.

src/content/language/constructors.md
```diff
@@ -51,12 +51,12 @@ To instantiate a class, use a generative constructor.
...
-  double x = 2.0;
-  double y = 2.0;
+  // Instance variables to hold the coordinates of the point.
+  double x;
+  double y;
 
-  // Generative constructor with initializing formal parameters:
-  Point(this.x, this.y);
+  // Generative constructor with optional positional parameters and default values.
+  Point([this.x = 2.0, this.y = 2.0]);
```
parlough, line 59: What do you think about just dropping the default values and making these required positional parameters?

Optional positional parameters and specifying their default values seem orthogonal to this section on "generative constructors".

parlough (approved):
Thanks for the PR @danielerat and for updating both sources! I think dropping the initializers here makes a lot of sense.

I have one question to consider, but looks good to me :)

src/content/language/constructors.md
danielerat: Hey @parlough, I guess that could work too!

The idea behind keeping default values was to retain the intent of the original code while showcasing optional positional parameters. That said, making the parameters required positional would work just as well.

PR #6211, opened 2024-11-15
Remove /stable/ from api doc urls

### LINKED ISSUE #6210: Problem with api docs links
### Describe the problem

The first two API docs links on the [async page](https://dart.dev/language/async) (`Future` and `Stream`) lead to something that looks like this:
image

The url shows `api.dart.dev/stable/dart-async/Stream-class.html`
Visiting those pages by following the links directly from api.dart.dev doesn't have any problems, and the link is different: `api.dart.dev/dart-async/Stream-class.html`, no `/stable`. 

I'm guessing this isn't the only page with issues, I'll check more.

### Expected fix

I think the links to api docs on the site need to be adjusted for recent updates to the api docs site. 

### On which platform(s) did you experience this issue?

_No response_

### Additional context

_No response_

MaryaBelanger:
Fixes #6210 

I did a find-and-replace to remove the redirects to `/stable/` (`{{site.sdkInfo.channel}}`), based on this note from the new api backend launch comms: 

> - [api.dart.dev/](http://api.dart.dev/) serves the latest stable directly (no longer redirects to /stable/<latest stable version>).
> - [api.dart.dev/(main|dev|beta)](http://api.dart.dev/(main%7Cdev%7Cbeta)) redirects to [api.dart.dev/(main|dev|beta)/latest](http://api.dart.dev/(main%7Cdev%7Cbeta)/latest).

But I didn't account for the second bullet, any possible main/dev/beta redirects, will that be an issue? 

I also came across [this note](https://github.com/dart-lang/site-www/blob/main/src/content/tools/sdk.md?plain=1#L17) that uses `site.sdkInfo.channel` as a condition...So I'm starting to think that blanket-removing all of them might not be the right move. Please advise @parlough!

### DIFF
--- a/src/_includes/tools/dart-compile-js-options.md
+++ b/src/_includes/tools/dart-compile-js-options.md
@@ -78,10 +78,10 @@ Some other handy options include:
 : Displays version information for `dart`.
 
 [Dart package configuration file]: {{site.repo.dart.lang}}/blob/main/accepted/2.8/language-versioning/package-config-file-v2.md
-[`String.fromEnvironment`]: {{site.dart-api}}/{{site.sdkInfo.channel}}/dart-core/String/String.fromEnvironment.html
-[`int.fromEnvironment`]: {{site.dart-api}}/{{site.sdkInfo.channel}}/dart-core/int/int.fromEnvironment.html
-[`bool.fromEnvironment`]: {{site.dart-api}}/{{site.sdkInfo.channel}}/dart-core/bool/bool.fromEnvironment.html
-[`bool.hasEnvironment`]: {{site.dart-api}}/{{site.sdkInfo.channel}}/dart-core/bool/bool.hasEnvironment.html
+[`String.fromEnvironment`]: {{site.dart-api}}/dart-core/String/String.fromEnvironment.html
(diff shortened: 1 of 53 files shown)


parlough:
> But I didn't account for the second bullet, any possible main/dev/beta redirects, will that be an issue?

We haven't used the functionality of deploying the site with beta/dev/main versions (and corresponding api.dart.dev links) in a long long time, so I think dropping these is fine. It will also make adding new links simpler and more consistent. 

Any links to docs that are not yet on stable can manually add `beta` or `dev`.

> I also came across [this note](https://github.com/dart-lang/site-www/blob/main/src/content/tools/sdk.md?plain=1#L17) that uses site.sdkInfo.channel as a condition...So I'm starting to think that blanket-removing all of them might not be the right move. Please advise @parlough!

As mentioned, we haven't used the functionality of the site documenting a beta/dev version, so I think this change is fine. We might consider dropping its usage in the note in a follow-up PR too.

parlough (approved):
Thanks for fixing these links @MaryaBelanger!

MaryaBelanger:
@parlough I just did a quick check and it doesn't look like it, but can you confirm this same kind of thing doesn't need to be applied to the flutter site? I did see some links in a site-shared file ("writing for the dart and flutter websites") that use `{{site.dart-api}}`

PR #6214, opened 2024-11-17
Move pub glossary terms to shared glossary

parlough:


MaryaBelanger:
Might want to discuss with the pub team, maybe once you're further along. I'm guessing (and can also kind of see myself) there's some benefit the team sees to having their dedicated glossary separately.

PR #6216, opened 2024-11-18
Move general packages doc out of guides directory

parlough:
Contributes to https://github.com/dart-lang/site-www/issues/5767


MaryaBelanger (approved):
Excellent! Thank you for doing this Parker

PR #6230, opened 2024-11-22
Remove old formatter page link

MaryaBelanger:
@munificent Instead of just removing the link to the old Formatting rules page, I added a link to your new FAQ page. 

Is that ok or should we hold off on attaching the new page to the site for now? I can just remove the new line I added and make to a to-do to add it back with the release.


munificent (approved):
Linking to the FAQ sounds good to me. Thanks for the fix!

PR #6117, opened 2024-09-30
Document pub workspaces

sigurdm:
Fixes #5891

Should probably only land with 3.6

PR #6239, opened 2024-11-26
Analysis context glossary entry

### LINKED ISSUE #6145: Document "analysis context" and account for workspaces too
Related issue: https://github.com/dart-lang/site-www/issues/5891

References relating to multi-option contexts and workspaces:

- https://github.com/dart-lang/sdk/issues/53875
- https://github.com/dart-lang/sdk/issues/53874
- https://github.com/dart-lang/sdk/issues/53876
- https://flutter.dev/go/pub-workspace

MaryaBelanger:
Fixes #6145 (along with changes made in #6117, specifically this commit: https://github.com/dart-lang/site-www/commit/d017546363b1444795429412d226a2cab7b2dbec)

I took the definition from [here](https://github.com/dart-lang/sdk/blob/main/pkg/analyzer/doc/tutorial/analysis.md), but I'm open to improvements if anyone has ideas

### DIFF
--- a/src/_data/glossary.yml
+++ b/src/_data/glossary.yml
@@ -1,3 +1,19 @@
+- term: "Analysis context"
+  short_description: |-
+    A representation of a set of code that should be analyzed using the same
+    sources of metadata.
+  long_description: |-
+    An _analysis context_ is a representation of a set of code that should be
+    analyzed using the same sources of metadata (same `package_config.json` file).
+    That set of code typically corresponds to a package.
+  related_links:
+    - text: "Customizing static analysis"
+      link: "/tools/analysis/"
+    - text: "Performing analysis"
+      link: "https://github.com/dart-lang/sdk/blob/main/pkg/analyzer/doc/tutorial/analysis.md"
+  labels:
+    - "analyzer"
+
 - term: "Assist"
   short_description: |-
     An automated, local code edit targeted at making common improvements to code.
--- a/src/_data/glossary.yml
+++ b/src/_data/glossary.yml
@@ -4,7 +4,8 @@
     sources of metadata.
   long_description: |-
     An _analysis context_ is a representation of a set of code that should be
-    analyzed using the same sources of metadata (same `package_config.json` file).
+    analyzed using the same sources of metadata
+    (same [`package_config.json`](/tools/pub/glossary#system-cache) file).
     That set of code typically corresponds to a package.
   related_links:
     - text: "Customizing static analysis"


src/_data/glossary.yml
```diff
@@ -1,3 +1,20 @@
+- term: "Analysis context"
+  short_description: |-
+    A representation of a set of code that should be analyzed using the same
+    sources of metadata.
```
srawlins, line 4: I think that "sources of metadata" might not be accurate, or the best terms. For one thing, I think the only source of "metadata" is a `package_config.json` file, so "source of metadata" would be fine.

And at that point you might just use "metadata" or something...

I was going to suggest the word "configuration" like "source configuration" or "package configuration file", but (a) I think most users are not really familiar with the package config file, and might mix it up with an _analysis options_ file, which at this point is quite distinct (and we took pains to separate them). So that's a little dicey. "Metadata" does nicely describe what distinguishes one set of sources from another, and doesn't _require_ users to know about the package config files.

src/_data/glossary.yml
```diff
@@ -1,3 +1,20 @@
...
+  long_description: |-
+    An _analysis context_ is a representation of a set of code that should be
```
srawlins, line 6: I'm not sure I like "set of code." (Though I'm game to use consistent terms wherever we can, if it's a consistent term.)

In my mind this conjures an image of a pile of subroutines, like a 30-line function here and a 100-line class there.

One property that I think would be helpful for users is that the "set of code" is a set of files that are more-or-less co-located in the filesystem. So you have properties like:

1. All of the files in one directory are always in the same analysis context.
2. All of the files in one directory are in the same analysis context as the files in any subdirectories, unless there exists a new "metadata" file (`.dart_tool/package_config.json`). (Hard to write that one concisely and without explicitly introducing the package config file that we don't want users to worry about.)

In practice, as you look at a directory tree of source files, a new analysis context is introduced each time you see a `pubspec.yaml` file, as those generally pair with generated `.dart_tool/package_config.yaml` files. But if you haven't run `dart pub get` yet, that's not really true. And if any pubspec file is using the new Workspaces feature, it's not really true again.

After all that, I think the only suggestion I have is to use something like "set of Dart source files" rather than "set of code." Again, consistency with other text being important.

srawlins (approved):

src/_data/glossary.yml
bwilkerson: > ... without explicitly introducing the package config file that we don't want users to worry about.

I disagree. I think it's important for users to understand the role of the package config file and its relationship to a `pubspec.yaml` file (or potentially multiple `pubspec.yaml` files in the case of workspaces. It's essential to understanding how absolute URIs in directives are resolved.

src/_data/glossary.yml
MaryaBelanger: This PR is very droppable, no worries there. The real reason for it was that I wanted to emphasize the performance improvements and was only thinking about that in terms of analysis contexts, so thought I had to define them somewhere. But the way you worded it:

> the fact that using workspaces for large repositories will reduce the amount of memory used by the Dart language server and hence improve IDE performance

does a good job. Thanks for the useful information @srawlins and @bwilkerson! I will close this and change the reference in the new workspaces PR to _not_ mention analysis contexts specifically, just the potential performance benefits.

PR #6240, opened 2024-11-27
doc(mixins): Add override annotation on some code examples

tilucasoli:
include some override annotations for code examples that are not explicitly assigned.

---

<details>

</details>

### DIFF
--- a/src/content/language/mixins.md
+++ b/src/content/language/mixins.md
@@ -87,6 +87,8 @@ mixin Musician {
 }
 
 class Virtuoso with Musician { 
+
+  @override
   void playInstrument(String instrumentName) { // Subclass must define.
     print('Plays the $instrumentName beautifully');
   }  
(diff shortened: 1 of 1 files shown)


MaryaBelanger:
Thanks for this @tilucasoli! I'm going to push a commit that adds a small note saying that the overrides are implicit, so they _can_ be omitted, but otherwise I agree that stating them explicitly is better for docs

tilucasoli:
Thank you for your response! Where will you add this note? Would it be added for each page?

MaryaBelanger:
I think probably around line 71, at the end of `## Specify members a mixin can call on itself` but before `### Define abstract members in the mixin`. That way we can say it applies to all the examples in the section.

MaryaBelanger (approved):
Decided against the note, so this is good to go! Thanks again @tilucasoli

PR #6245, opened 2024-12-02
Provide alternative to SSH for accessing private repo dependencies

### LINKED ISSUE #6234: Provide info how to access private git dependencies via JWT access token (GitHub App)
### Page URL

https://dart.dev/tools/pub/dependencies#git-packages

### Page source

https://github.com/dart-lang/site-www/tree/main/src/content/tools/pub/dependencies.md

### Describe the problem

For private git dependencies, it is recommended to use SSH. But GitHub themselves disengage SSH and also this is only feasable locally. On CIs for every private repo, a new SSH key has to be generated, which then can be entered in the repos as `Deployment Key` to allow fetching the dependency (one key CANNOT be used for multiple repos, unfortunately).

So we should give the hint to authenticate via access tokens, e.g. one generated with "GitHub App"s.

Here a common workflow:
```
    steps:
      - uses: actions/checkout@v4
      - uses: actions/create-github-app-token@v1
        id: app-token
        with:
          app-id: ${{ vars.APP_ID }}
          private-key: ${{ secrets.SSH_PRIVATE_KEY }}
          owner: ${{ github.repository_owner }}
      - name: Set github url and credentials
        run: |
          git config --global --add url."https://x-access-token:${{ steps.app-token.outputs.token }}@github".insteadOf https://github
          git config --global --add url."https://x-access-token:${{ steps.app-token.outputs.token }}@github.com/".insteadOf ssh://git@github.com:
          git config --global --add url."https://x-access-token:${{ steps.app-token.outputs.token }}@github.com/".insteadOf git@github.com:
```

Would that be a valid approach to link in the docs?

### Expected fix

Provide a link or add an example how to fetch private repository dependencies from git via access tokens.

### Additional context

_No response_

MaryaBelanger:
Fixes #6234 

@Gustl22 Thanks for raising this and let us know if the added lines make sense to you!

### DIFF
--- a/src/content/tools/pub/dependencies.md
+++ b/src/content/tools/pub/dependencies.md
@@ -151,6 +151,12 @@ dependencies:
     git: git@github.com:munificent/kittens.git
 ```
 
+Alternatively, you can access private repos using access tokens via HTTPS.
+This bypasses the need to continously generate SSH keys in certain cases.
+To use HTTPS, you'll need to [configure your git environment][].
+
+[configure your git environment]: https://docs.github.com/en/get-started/getting-started-with-git/caching-your-github-credentials-in-git
+
 If you want to depend on a specific commit, branch, or tag,
 add a `ref` key to the description:
 
--- a/src/content/tools/pub/dependencies.md
+++ b/src/content/tools/pub/dependencies.md
@@ -142,7 +142,7 @@ The `git` here says this package is found using Git, and the URL after that is
 the Git URL that can be used to clone the package.
 
 Even if the package repo is private, 
-you can configure your `git` setup to access the repo via an
+you can configure your `git` setup to access the repo using an
 [HTTPS access key][GitHub HTTPS] or an [SSH key-pair][GitHub SSH].
 Then you can depend on the package by using the repo's corresponding URL:
 
--- a/src/content/tools/pub/dependencies.md
+++ b/src/content/tools/pub/dependencies.md
@@ -141,21 +141,20 @@ dependencies:
 The `git` here says this package is found using Git, and the URL after that is
 the Git URL that can be used to clone the package.
 
-Even if the package repo is private, if you can
-[connect to the repo using SSH,][GitHub SSH]
-then you can depend on the package by using the repo's SSH URL:
+Even if the package repo is private, 
+you can configure your `git` setup to access the repo via an
+[HTTPS access key][GitHub HTTPS] or an [SSH key-pair][GitHub SSH].
+Then you can depend on the package by using the repo's corresponding URL:
 
 ```yaml
 dependencies:
   kittens:
+    # SSH URL:
     git: git@github.com:munificent/kittens.git
 ```
 
-Alternatively, you can access private repos using access tokens via HTTPS.
-This bypasses the need to continously generate SSH keys in certain cases.
-To use HTTPS, you'll need to [configure your git environment][].
-
-[configure your git environment]: https://docs.github.com/en/get-started/getting-started-with-git/caching-your-github-credentials-in-git
+The `dart pub` command calls `git clone` as a subprocess, so all you need to
+provide is a `<url>` that works when `git clone <url>` is executed.
 
 If you want to depend on a specific commit, branch, or tag,
 add a `ref` key to the description:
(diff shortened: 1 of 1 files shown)


src/content/tools/pub/dependencies.md
```diff
@@ -151,6 +151,12 @@ dependencies:
     git: git@github.com:munificent/kittens.git
 ```
 
+Alternatively, you can access private repos using access tokens via HTTPS.
+This bypasses the need to continously generate SSH keys in certain cases.
```
jonasfj, line 155: I don't know of any _sane_ setups where you need to continuously generate SSH keys.

That sounds like a corner case.

src/content/tools/pub/dependencies.md
```diff
@@ -151,6 +151,12 @@ dependencies:
     git: git@github.com:munificent/kittens.git
 ```
 
+Alternatively, you can access private repos using access tokens via HTTPS.
```
jonasfj, line 154: It might be better to be more generic.

--------

If you have
```yaml
dependencies:
  kittens:
    git: <url>
```

Then this will work, so long as `git clone <url>` will work.
 * If your `<url>` uses SSH (example: `git@github.com:munificent/kittens.git`), you may need to [configure an SSH key](https://docs.github.com/en/authentication/connecting-to-github-with-ssh).
 * If your `<url>` uses HTTPS (example: `https://github.com/munificent/kittens.git`), you may need to configure [git credentials](https://docs.github.com/en/get-started/getting-started-with-git/caching-your-github-credentials-in-git)).

Neither of these choices is specific to `dart pub`, which merely runs `git` as a subprocess.

--------

That's probably not what you want to write. My point is that the hint we should give here is probably not that "you can use HTTP to avoid continuously generating SSH keys". Because SSH keys should not be continuously generated!

Normally you generate an SSH key on your machine and copy the public key into your settings on github (or whatever git hosting provider you have).

Instead, I propose that the hint we give is that SSH isn't the only way to access private git repositories. Indeed, `dart pub` is not involved with the authentication when cloning a git repository.

I suppose that by trying to help and give example, we've potentially failed to clearly communicate that `dart pub` just calls `git clone` as a subprocess. And it is your responsibility to:
 * (a) configure `git` such that it can clone the repository your dependency is in.
   This **may** involve:
   * Creating a public/private SSH key-pair on your system, and uploading it to Github.
   * Installing and using `gh` to configure HTTPS access for Github.
   * Using `gcloud init` to configure a git-authentication helper for Cloud Source Repositories.
   * Using gpg to generate a key-pair on a yubikey and configuring `gpg-agent` (I've personally done this, it's rather complicated :rofl)
   * Using a u2f-key (yubikey) to create an SSH-key key-pair with `ssh-agent`-
   * etc... the list goes on and we should probably only provide hints for people using github.
 * (b) provide a `<url>` that works when `git clone <url>` is executed.

So describe the comment being very long -- perhaps what I'm trying to say is:
We should say less, or make it clear that using SSH is just an example, not the only way.

ericwindmill: 💯  comment. I think we should just communicate the end of this comment with something like "`dart pub` calls `git clone` as a subprocess, so you need to provide a url that works when `git clone <url>` is executed. This is most commonly done with HTTPS access to Github or with an SSH key-pair on your system that you've uploaded to Github." 

And then link the second sentence to the [Github authentication page](https://docs.github.com/en/authentication).

I don't think you need to exhaustively discuss all the ways one might connect to Github

MaryaBelanger: Thank you both for the feedback! I changed it to kind of equate SSH and HTTPS just as different options to the same end, and then included the line about conforming to `git clone`

Gustl22:
Looks awesome! Thank you for the effort!
Haven't tested every linked Git/https auth mechanism though, but I suppose, if one works the others should work, too 😅

jonasfj (approved):
:rocket:

src/content/tools/pub/dependencies.md
```diff
@@ -141,16 +141,21 @@ dependencies:
 The `git` here says this package is found using Git, and the URL after that is
 the Git URL that can be used to clone the package.
 
-Even if the package repo is private, if you can
-[connect to the repo using SSH,][GitHub SSH]
-then you can depend on the package by using the repo's SSH URL:
+Even if the package repo is private, 
+you can configure your `git` setup to access the repo via an
```
parlough, line 145: https://developers.google.com/style/word-list#via

```suggestion
you can configure your `git` setup to access the repo with an
```

or

```suggestion
you can configure your `git` setup to access the repo using an
```

or something else :)

parlough (approved):
Thanks @MaryaBelanger!

PR #6259, opened 2024-12-07
Move language and library tour to deprecated folder

parlough:
Contributes to https://github.com/dart-lang/site-www/issues/5767


firebase.json
MaryaBelanger, line 1: Do these `/deprecated/...` destinations just redirect to the new version of the page?

I end up at `/language` and `/libraries` when testing the old source for each. I think I see it jump to `/deprecated` first though. Just checking that that's the intended behavior

MaryaBelanger (approved):

firebase.json
parlough: Yes that's the intended behavior if you don't specify any fragment (`#id`).

It's been a while since we implemented this, but remember we set up client-based redirects to redirect from a fragment to the correct location. As different fragments should lead to different pages, but the server doesn't see the fragments.

For example, [/guides/language/language-tour#enums](https://dart-dev--pr6259-feat-move-tours-to-deprected-folder-tw3vd4q8.web.app/guides/language/language-tour#enums) will redirect to [/deprecated/language-tour#enums](https://dart-dev--pr6259-feat-move-tours-to-deprected-folder-tw3vd4q8.web.app/deprecated/language-tour#enums), then the JS on that page will look at the fragment and redirect to [/language/enums](https://dart-dev--pr6259-feat-move-tours-to-deprected-folder-tw3vd4q8.web.app/language/enums).

PR #6269, opened 2024-12-10
Language evolution 3.6

### LINKED ISSUE #6267: Add language evolution entry for Dart 3.6
Just digit separators, I believe

MaryaBelanger:
Also added an id to the section that talks about digit separators in the language tour, so we have something to point to. 

fixes #6267

### DIFF
--- a/src/content/resources/language/evolution.md
+++ b/src/content/resources/language/evolution.md
@@ -42,6 +42,19 @@ on the Dart language GitHub repo.
 
 ## Changes in each release
 
+### Dart 3.6
+_Released 11 December 2024_
+| [Dart 3.6 announcement](https://medium.com/dartlang/announcing-dart-3-6-778dd7a80983)
+
+Dart 3.6 added the [digit separator][] underscore (`_`) to the language.
+Digit separators improve readability of long number literals.
+
+```dart
+var m = 100__000_000__000_000__000_000
+```
+
+[digit separator]: /language/built-in-types#digit-separators
+
 ### Dart 3.5
 _Released 6 August 2024_
 | [Dart 3.5 announcement](https://medium.com/dartlang/dart-3-5-6ca36259fa2f)
--- a/src/content/language/built-in-types.md
+++ b/src/content/language/built-in-types.md
@@ -173,6 +173,8 @@ const msUntilRetry = secondsUntilRetry * msPerSecond;
 
 For more information, see [Numbers in Dart][dart-numbers].
 
+<a id="digit-separators"></a>
+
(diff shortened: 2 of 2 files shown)


src/content/resources/language/evolution.md
```diff
@@ -42,6 +42,19 @@ on the Dart language GitHub repo.
...
 
+### Dart 3.6
+_Released 11 December 2024_
+| [Dart 3.6 announcement](https://medium.com/dartlang/announcing-dart-3-6-778dd7a80983)
+
+Dart 3.6 added the [digit separator][] underscore (`_`) to the language.
+Digit separators improve readability of long number literals.
+
+```dart
+var m = 100__000_000__000_000__000_000
```
parlough, line 53: The original number here is too big to be represented by Dart ints. Consider making it a bit smaller so it doesn't result in an error.

```suggestion
var m = 1__000_000__000_000__000_000;
```

src/content/resources/language/evolution.md
```diff
@@ -42,6 +42,19 @@ on the Dart language GitHub repo.
 
 ## Changes in each release
 
+### Dart 3.6
+_Released 11 December 2024_
+| [Dart 3.6 announcement](https://medium.com/dartlang/announcing-dart-3-6-778dd7a80983)
+
+Dart 3.6 added the [digit separator][] underscore (`_`) to the language.
+Digit separators improve readability of long number literals.
```
parlough, line 50: Nice intro and good idea to have a short sample!

src/content/resources/language/evolution.md
```diff
@@ -42,6 +42,19 @@ on the Dart language GitHub repo.
 
 ## Changes in each release
 
+### Dart 3.6
+_Released 11 December 2024_
+| [Dart 3.6 announcement](https://medium.com/dartlang/announcing-dart-3-6-778dd7a80983)
+
+Dart 3.6 added the [digit separator][] underscore (`_`) to the language.
```
parlough, line 49: This wording makes it seem like "the [...] underscore" is a language construct, but rather number literals are.

Would something like the following make sense, so it indicates there can be more than one?

```suggestion
Dart 3.6 added support for [digit separator][] underscores (`_`) to the language.
```

parlough (approved):
Thanks @MaryaBelanger!

One small issue with the snippet and a small wording suggestion to consider.

src/content/resources/language/evolution.md
MaryaBelanger: That's a really good point, thank you!

PR #6279, opened 2024-12-16
Use `/to` tooling links for pub redirects

parlough:
In an effort to standardize on `/to` for tooling redirects.


MaryaBelanger (approved):
What's the difference between `/to` and `/go` links? If they're no different should the goal be to move everything to the more common `/go`? If there is a meaningful difference, carry on!

parlough:
> What's the difference between `/to` and `/go` links? If they're no different should the goal be to move everything to the more common `/go`? If there is a meaningful difference, carry on!

Due to historical reasons and clarity, `/go` for Flutter redirects is reserved for design docs. So we started using `/to` for tooling redirects. I wrote https://github.com/flutter/flutter/blob/master/docs/contributing/use-reliable-links.md with some guidelines on when to add and use a `/to` link for Flutter contributors 

For consistency purposes, I felt it would be best to adjust links from tools to use `/to/` on dart.dev as well. Then we could repurpose `/go` for whatever we want/need (perhaps design docs as well or temporary redirects).

Do you think it's worth making that transition? Or unnecessary for dart.dev?

PR #6280, opened 2024-12-16
Fix and clean up a few Firebase hosting redirects

parlough:
- Fix a few redirects that lead to broken fragments
- Expand a few to cover 404s seen in analytics
- Consolidate a few really old ones


MaryaBelanger (approved):
Thank you!

PR #6284, opened 2024-12-18
removed the code already written for exception exercice

00suryavanshi00:
I've removed the pre-written code in Exceptions section of dart cheatsheet and replaced it with a comment just like all the other exercises so far.


MaryaBelanger (approved):
Whoops, we missed that one 😅  Thank you for catching our mistake @00suryavanshi00!

PR #6286, opened 2024-12-19
Tighten guidelines around dynamic

natebosch:
As the language and ecosystem has grown the target for static safety has
moved up. Previously it was considered enough to be explicit about the
places where `dynamic` is introduced, and then it is OK to silently do
dynamic things with it, and write code that keeps references `dynamic`
throughout. It is more common today to want to be explicit about places
where `dynamic` is used, not just where it is introduced. We can see
this in expanded use of stricter static checking and the
`avoid_dynamic_calls` lint.

Rephrase two guidelines to move towards being strict about using
dynamic.
- Recommend annotating with `Object?` instead of `dynamic` when
  inferences fails. Update the code sample to use `Object?` in the
  method arguments are return types for the JSON related API. Update the
  text to explain how `Object?` is treated differently and more strictly
  statically, which makes the unsafe operations visible in syntax.
- Rephrase from using dynamic to "disable static checking" to using it
  to "invoke dynamic members". The wording for the rule already focused
  on using dynamic to "permit all operations". Remove the paragraph
  about retaining the `dynamic` type when working with APIs that use it,
  instead recommending to move away from `dynamic` at those boundaries
  and using `Object?` in most code.

### DIFF
--- a/src/content/effective-dart/design.md
+++ b/src/content/effective-dart/design.md
@@ -1410,47 +1410,36 @@ var completer = Completer<Map<String, int>>();
 ```
 
 
-### DO annotate with `dynamic` instead of letting inference fail
+### DO annotate with `Object?` instead of letting inference fail
 
-When inference doesn't fill in a type, it usually defaults to `dynamic`. If
-`dynamic` is the type you want, this is technically the most terse way to get
-it. However, it's not the most *clear* way. A casual reader of your code who
-sees that an annotation is missing has no way of knowing if you intended it to be
-`dynamic`, expected inference to fill in some other type, or simply forgot to
-write the annotation.
+When inference doesn't fill in a type, it usually defaults to `dynamic`, which
+is rarely the best type to use. A `dynamic` reference allows for unsafe
+operations that use identical syntax to operations that are statically safe when
+the type is not `dynamic`. An `Object?` reference is safer. For example a
+`dynamic` reference may fail a type cast that is not visible in the syntax,
+while an `Object?` reference will guarantee that the `as` cast is written in
+source.
 
-When `dynamic` is the type you want, write that explicitly to make your intent
-clear and highlight that this code has less static safety.
+Use `Object?` to indicate in a signature that any type of object, or null, is
+allowed.
 
-<?code-excerpt "design_good.dart (prefer-dynamic)"?>
+<?code-excerpt "design_good.dart (prefer-object-question)"?>
 ```dart tag=good
-dynamic mergeJson(dynamic original, dynamic changes) => ...
+Object? mergeJson(Object? original, Object? changes) => ...
 ```
 
-<?code-excerpt "design_bad.dart (prefer-dynamic)"?>
+<?code-excerpt "design_bad.dart (prefer-object-question)"?>
 ```dart tag=bad
 mergeJson(original, changes) => ...
 ```
 
-Note that it's OK to omit the type when Dart *successfully* infers `dynamic`.
-
-<?code-excerpt "design_good.dart (infer-dynamic)"?>
-```dart tag=good
-Map<String, dynamic> readJson() => ...
-
-void printUsers() {
-  var json = readJson();
-  var users = json['users'];
-  print(users);
-}
-```
-
-Here, Dart infers `Map<String, dynamic>` for `json` and then from that infers
-`dynamic` for `users`. It's fine to leave `users` without a type annotation. The
-distinction is a little subtle. It's OK to allow inference to *propagate*
-`dynamic` through your code from a `dynamic` type annotation somewhere else, but
-you don't want it to inject a `dynamic` type annotation in a place where your
-code did not specify one.
+In the cases where a dynamic member will be invoked this is technically the most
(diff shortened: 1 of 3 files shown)


natebosch:
I'm not sure how we handle a change in recommendations that impacts a link header... Do we just allow old links to break?

natebosch:
Opening this now so folks can start considering it, but I expect we won't be able to organize all the discussion around it until after the holidays.

cc @munificent what do you think?

I _think_ this change is a real shift in the ecosystem and I'm being descriptive here, but I do have a bias here and I'd want to be prescriptive about this if were writing this doc with that intent. cc @lrhn who might have a preference to keep the older style which is more permissive.

parlough:
> I'm not sure how we handle a change in recommendations that impacts a link header... Do we just allow old links to break?

Since these headers usually last a long time and might be linked to from outside the site, if there's an similar or related header, I tend to add an empty anchor with the old fragment above that.

For example:

```markdown
<a id="do-annotate-with-dynamic-instead-of-letting-inference-fail" aria-hidden="true"></a>

### DO annotate with `Object?` instead of letting inference fail
```

src/content/effective-dart/design.md
```diff
@@ -1678,11 +1667,13 @@ bool convertToBool(Object arg) {
...
 ```
 
-The main exception to this rule is when working with existing APIs that use
-`dynamic`, especially inside a generic type. For example, JSON objects have type
-`Map<String, dynamic>` and your code will need to accept that same type. Even
-so, when using a value from one of these APIs, it's often a good idea to cast it
-to a more precise type before accessing members.
+Prefer using `Object?` over `dynamic` in code that is not invoking a member
+dynamically, even when working with existing APIs that use `dynamic`.
+For example, JSON objects have runtime type `Map<String, dynamic>`, which is an
```
lrhn, line 1672: I'd say "JSON objects traditionally have the static type `Map<String, dynamic>`"

The runtime type doesn't matter, there is no meaningful difference between `dynamic` and `Object?` in a runtime type.

Also "JSON objects" is ambiguous. It can either mean any object created by `jsonDecode`, which is typed as just `dynamic`,  or it can mean a Dart representation of a "JSON Object" (based on a JavaScript Object), which is indeed traditionally represented by a map with static type `Map<String, dynamic>`.
The `dart: convert` code doesn't mandate that you use `Map<String, dynamic>` instead of `Map<String, Object?>` when you match or cast to a `Map` type, that's entirely user habit. It's this just saying that you should cast those to `... Object?...` before using them?

src/content/effective-dart/design.md
```diff
@@ -1678,11 +1667,13 @@ bool convertToBool(Object arg) {
...
+equivalent type to `Map<String, Object?>`.
+When using a value from one of these APIs, it's often a good idea to treat it as
```
lrhn, line 1674: Which APIs is this? (First use of "API" here, but the definite form suggests that it refers to something mentioned earlier.)

Is it something that expects a `Map<String, dynamic>`, or which produces it?
Could the previous paragraph start with
"APIs which accept or produce JSON-object maps traditionally type those as `Map<String, dynamic>` ..."

lrhn (commented):
I'm fine with recommending that you don't write `dynamic`.

Might want to say how you then opt back into dynamic behavior. I expect it's `(… as dynamic).operation() as ExpectedType`.

src/content/effective-dart/design.md
```diff
@@ -1410,47 +1410,37 @@ var completer = Completer<Map<String, int>>();
...
+operations that use identical syntax to operations that are statically safe when
+the type is not `dynamic`. An `Object?` reference is safer. For example a
+`dynamic` reference may fail a type cast that is not visible in the syntax,
+while an `Object?` reference will guarantee that the `as` cast is written in
+source.
 
-When `dynamic` is the type you want, write that explicitly to make your intent
-clear and highlight that this code has less static safety.
+Use `Object?` to indicate in a signature that any type of object, or null, is
+allowed.
```
munificent, line 1425: The overall metarule for the guidelines is "unless stated otherwise, express things the most terse way". If you want some return type or parameter type to have type `dynamic`, the most terse way is to simply write nothing.

The original intent of this guideline was to opt *out* of that default "use the shortest thing" rule. That's why the guideline is about "instead of letting inference fail".

The new guideline here is really about "when should you use `Object?` versus `dynamic`, which is a different thing. I think the second updated guideline below does a pretty good job of covering that.

So for this one, I think it would be better to recast it to something like `DON'T let inference fall back to dynamic".

munificent (commented):
Overall, I really like the direction of this. I'd like to iterate a bit on the first rule though, to make it clear that it's about avoiding inference yielding a default type. (For example, if we changed inference to give you `Object?` when it didn't have any other type, I'd still want to have this guideline.)

antfitch:
Nate, do you have time to move forward on this? It's been almost a year. If not, I suggest creating a ticket for this with the things you want fixed.

natebosch:
Yes sorry, let's close for now and I'll reopen when I have a chance to rethink the approach.

Issue #6265, opened 2024-12-10
Document `covariant` elsewhere then retire "Fixing common type problems" page


parlough:
Page: https://dart.dev/guides/language/sound-problems#troubleshooting
Source: https://github.com/dart-lang/site-www/blob/main/src/content/guides/language/sound-problems.md

The page was introduced in the context of sound Dart being new and before the diagnostic messages were well documented. Now the warnings documented here (and many more) have consistent documentation on the [Diagnostic messages](https://dart.dev/tools/diagnostic-messages) page.

The only parts not covered by that page is the section on [runtime errors](https://dart.dev/guides/language/sound-problems#runtime-errors) and the [`covariant` keyword](https://dart.dev/guides/language/sound-problems#the-covariant-keyword). These docs need a new home before retiring the page, or at least the `covariant` docs do. That probably belongs in the language docs anyway.

MaryaBelanger:
It seems like [Runtime errors](https://dart.dev/deprecated/sound-problems#runtime-errors) can move to the types system page too, under [Runtime checks](https://dart.dev/language/type-system#runtime-checks). What do you think? It can almost replace the existing section 1:1 (though I'd try to shorten it a bit, remove the "Fix: ..." section header, etc).

PR #6299, opened 2025-01-03
Move covariant keyword docs out of sound problems page

parlough:
Contributes to https://github.com/dart-lang/site-www/issues/6265


MaryaBelanger (approved):
Thank you for taking care of this! I think it works here

PR #6305, opened 2025-01-05
Surface glossary terms and related links in expandable cards

parlough:
Introduces a new card-based design for entries on the glossary page. The short description is shown by default and the long description and resource links are shown once expanded. When navigating to the page, the card that corresponds to the current fragment (such as `#variance` in `dart.dev/resources/glossary#variance`) is expanded.

Additionally you can specify the type of a resource to add an icon in front of the resource button. Currently the supported types are:

- `term` _(for other glossary entries or word definitions)_
- `article` or `doc` _(general for doc pages)_
- `tutorial` _(for tutorials, codelabs, etc.)_
- `api` _(for API links)_
- `video` _(for videos, such as links to Youtube videos)_
- `code` or `sample` _(for code samples or repository links)_
- `diagnostic` or `lint` _(for links to specific lints, diagnostic docs, etc.)_

To implement this, I've introduced a new set of shared button styles in `_buttons.scss` and updated the shared set of card styles in `_cards.scss`. That way cards across the site _(currently this and the linter rules page)_ can benefit from a consistent structure and shared styles.

Staged affected docs:

- [Glossary page](https://dart-dev--pr6305-misc-glossary-cards-eefxhuts.web.app/resources/glossary)
- [Linter rules index](https://dart-dev--pr6305-misc-glossary-cards-eefxhuts.web.app/tools/linter-rules#rules) _(Should look the same, but the shared styles are used by it)_


MaryaBelanger (approved):
This is so cool 🥹 The resource icons are so thoughtful. 

I dont see how the lint pages are affected, is there no visible change?

Either way, looks amazing, thank you so much.

parlough:
> I don't see how the lint pages are affected, is there no visible change?

Then that's good. They shouldn't be visually affected, but they use the new shared styles.

PR #6311, opened 2025-01-07
Update macros.md

dishankjindal1:
Hi,

I was going through the macros section, and the jsonCodeable hyperlink was point to a specific version which is not the latest. Which game me an impression that it no longer being worked at. but that is not the case.

so after discussing on dart discord, i had the confirmation it was a mistake. And hence raising this PR with the updated link.

---


MaryaBelanger (approved):
Thank you @dishankjindal1 , great catch!

PR #6333, opened 2025-01-15
Remove remnants of `/guides` directory

parlough:
Finalizes the work completed as part of https://github.com/dart-lang/site-www/issues/5767, removing the remaining mentions of the `/guides` directory. Adds a new [`https://dart.dev/docs`](https://dart-dev--pr6333-feat-remove-guides-directory-gf8u03iz.web.app/docs) page to act as an index into our documentation.

Follow up PRs will adjust and expand the index.

Closes https://github.com/dart-lang/site-www/issues/5767


MaryaBelanger (approved):
Wow, can't believe it's done! Thank you Parker.

Qq: So `/docs` is just the index page? No other pages have `/docs` in their path?

parlough:
> Qq: So /docs is just the index page? No other pages have /docs in their path?

Yes, at least for now. Do you think that's confusing? I don't know where else to put this index :P

MaryaBelanger:
> Yes, at least for now. Do you think that's confusing? I don't know where else to put this index :P

I don't think it's confusing! I was just confirming that we weren't introducing another arbitrary layer (which like, _obviously_ not since that's the point of getting rid of `/guides`, but just quadruple-checking 😅 )

PR #6350, opened 2025-01-23
Simplify implementation of homepage layout and styles

parlough:
- Removes reliance on bootstrap classes for layout and responsiveness.
- Stops using bootstrap-provided mixins and other utilities.
- Remove some unnecessary div nesting.
- Rely on same elements for narrow and wide layouts so that it's not required to update two locations.
- Adds a missing missing aria attributes where necessary to improve accessibility.
- Other general cleanup.

Contributes to https://github.com/dart-lang/site-www/issues/4164
Contributes to https://github.com/dart-lang/site-www/issues/5251


MaryaBelanger (approved):
LGTM!

PR #6356, opened 2025-01-23
Add a new box to docs front page

MaryaBelanger:
Just to even it out :) And the package docs are a big deal!


parlough (approved):
Good idea! Thanks :D

Issue #3849, opened 2022-02-21
Move away from jquery in our own JS files


parlough:
### Describe the problem

We currently use jquery in our [js files](https://github.com/dart-lang/site-www/tree/main/src/assets/js). While we have to currently include jquery anyway for our current Bootstrap use, for our very small uses we don't need to use jquery functionality. Removing most uses will make it easier to remove in the future.

### Expected fix

Replace use of Jquery with more modern JS.

parlough:
If anyone wants to help contribute to this. I recommend updating one function in a pull request at a time. So it is easier to review and to test. Let me know if you have any questions or need any help. Thanks!

brnt-toast:
I'll start taking a look over the weekend

atsansone:
Bootstrap 5.x removed its jQuery dependencies. We may be able to consider a holistic change.

parlough:
Yeah it would be nice to update Bootstrap (or even remove it completely as a lot of styles are custom anyway), but either will likely be a larger undertaking. I've made an issue to track that work: https://github.com/dart-lang/site-www/issues/4164

In the mean time, it is nice to incrementally move our custom JS scripts away from jquery so once we do make that bootstrap change, we can remove jquery easily. 

If you work on this @brnt-toast, let us know if you need any help. [`main.js`](https://github.com/dart-lang/site-www/blob/main/src/assets/js/main.js) is a good starting candidate and even migrating one method would be useful.

PR #6360, opened 2025-01-24
Remove usage and inclusion of jquery

parlough:
Now that bootstrap 4 (which required jquery) has been removed, we can remove our final usages of jquery and the inclusion of the file.

Closes https://github.com/dart-lang/site-www/issues/3849


MaryaBelanger (approved):
Thanks!

PR #6368, opened 2025-01-27
Add tooling to programmatically update diagnostic-related files

parlough:
This mostly moves some tooling from the SDK and doesn't need to be too closely reviewed. It adds `./dash_site generate-diagnostics` to update the `src/_data/linter_rules.json` and `src/content/tools/diagnostic-messages.md` files from sources in the SDK.

**This is temporary and much of it will be expanded or replaced** with logic in the static site generator to generate individual diagnostic message pages. Then it will be automatically ran every night, reducing the manual steps necessary. I'm proposing we land this so we can incrementally get to that point and enable simplifying the SDK implementation.


MaryaBelanger (approved):
Looks good in staging, thanks!

Issue #4839, opened 2023-05-10
Add a record example using a typedef


parlough:
A common use case I've already seen is to declare a typedef for a record.

It might be worth documenting a specific example where this makes sense in the typedef and/or the record documentation.

RandalSchwartz:
Just be careful to emphasize that a typedef is an alias, not a new type.

```dart
typedef A = (int a1, String a2);
typedef B = (int b1, String b2);

final A aThing = (14, 'hello');
final B bThing = aThing; // this works!
```

This works of course because A and B are not really new types... just alias for structs.  And records are the same type as long as they have the same shape.

PR #6382, opened 2025-02-02
feat(docs): Add documentation for using records and typedef

Asncodes-80:
This commit introduces a new section in the documentation showcasing the combination of Dart's record types and `typedef` to create a simple, reusable data structure for dynamic UI rendering.

This approach was originally shared as an answer on Stack Overflow in Spring 2023 to the question: "Is there anything like a struct in Dart?" (https://stackoverflow.com/a/76391196/12158825). The solution received highly positive feedback via emails and comments, highlighting its usefulness for developers.

This feature is now documented as a recommended approach for developers aiming to write cleaner and more maintainable Dart code that wasn't mentioned before.

### DIFF
--- a/src/content/language/records.md
+++ b/src/content/language/records.md
@@ -224,6 +224,56 @@ parallelization of futures of different types, which you can read about in the
 [`dart:async` documentation][].
 :::
 
+## Records as simple data structures
+
+The provided example illustrates how developers can leverage this pattern to
+define dynamic UI widgets in a type-safe manner, similar to interfaces in
+TypeScript but without the overhead of classes. This approach is particularly
+useful for scenarios where simplicity and reusability are prioritized.
+
+```dart
+typedef ButtonItem = ({
+  String label,
+  Widget? icon,
+  void Function()? onPressed,
+});
+```
+
+Here is an example about using `record` and `typedef` as structure of dynamic UI
+widgets render:
+
+```dart
+// Define list from `buttonItem` interface.
+List<ButtonItem> buttons = [
+  (
+    label: "Button I",
+    icon: const Icon(Icons.upload_file),
+    onPressed: () => print("Action -> Button I"),
+  ),
+  (
+    label: "Button II",
+    icon: const Icon(Icons.info),
+    onPressed: () => print("Action -> Button II"),
+  )
+];
+
+List<Container> widget = buttons.map(
+  (ButtonItem button) => Container(
+    margin: const EdgeInsets.all(4.0),
+    child: OutlinedButton.icon(
+      onPressed: button.onPressed,
+      icon: button.icon!,
+      label: Text(button.label),
+    ),
+  )
+)
+.toList();
+```
+
+The example demonstrates how to define a `ButtonItem` structure using records
+and `typedef`, which promotes cleaner and more maintainable code by adhering to
+the DRY (Don't Repeat Yourself) principle.
+
 [language version]: /resources/language/evolution#language-versioning
 [collection types]: /language/collections
 [pattern]: /language/patterns#destructuring-multiple-returns


lrhn:
I think the text is a little too unfocused. It's not clear what it's trying to tell the user, and unclear on why this code would be better or worse than alternatives. It's just an example, and it lacks a motivation for that example, so the reader is likely to not know what their take-away should be.

The text can be improved, but that's just editing. We can fix that.
I want it to have a clear focus and motivation for the section, and I'm not sure what the intended focus and motivation is for the current text, making me feel like asking "Why are you telling me this?".

(The text starts out by referring to the previous section as just "The provided example", which makes a reader think it should be something closer. It mentions UI code in a way that feels out of the blue, as something that the example shows something about, even though this is the first mention of UI in the document, and it's not mentioned in any other section. The reference to JavaScript interfaces may be relevant, but should have a link to a definition or explanation of what it is, a reader cannot be assumed to know already. Claims are made, but are generally unsupported: "without the overhead of classes" (which overhead? syntactic or runtime?), "particularly useful for..." (how?).
As a section, the text should be self-contained, or at least make references to other sections explicit and clear, so the reader knows where to go and look if they haven't read it recently. It should also fit into the document, so if there are any running examples it can use, that would be better than making up ones from scratch, from a domain that isn't otherwise mentioned.)

If the section was presented as just an example, of how you can use records as simple data objects,
without making any claims about whether they are better or worse than any other choice,
then I think it would work.

Something like:
> ## Records as simple data structures
>
> Records only hold data, but if that is all you need, they're immediately available without needing
> to declare any new classes.
> For a simple list of data tuples which all have the same shape, a list of records is the most direct
> representation.
>
> Take this list of "button definitions":
> ```dart
> final buttons = [
>   (
>     label: "Button I",
>     icon: const Icon(Icons.upload_file),
>     onPressed: () => print("Action -> Button I"),
>   ),
>   (
>     label: "Button II",
>     icon: const Icon(Icons.info),
>     onPressed: () => print("Action -> Button II"),
>   )
> ];
> ```
>
> This code can be written directly without needing any further declarations.
>
> One can choose to give the record type itself a name, and use that rather
> than writing out the full record type. That also allows stating that some
> fields can be `null`, even if none of the *current* entries in the list
> have a `null` value.
> ```dart
> typedef ButtonItem = ({String label, Icon icon, void Function()? onPressed});
> final List<ButtonItem> buttons = [
>   // ...
> ];
> ```
> Because record types are structural types, that does not introduce any
> new type names, just another way to refer to the record type.
> However if all code refers to the record type by its alias, it's more likely
> that you can later change the implementation.
>
> Code can work with such button definitions the same it would
> with simple class instances.
> ```dart
>   List<Container> widget = [
>     for (var button in buttons)
>       Container(
>         margin: const EdgeInsets.all(4.0),
>         child: OutlinedButton.icon(
>           onPressed: button.onPressed,
>           icon: button.icon,
>           label: Text(button.label),
>         ),
>       ),
>   ];
> ```
>
> It may be possible to later change the type to a class type to add methods.
> ```dart
> class ButtonItem {
>   final String label;
>   final Icon icon;
>   final void Function()? onPressed;
>   ButtonItem({required this.label, required this.icon, this.onPressed});
>   bool get hasOnpressed => onPressed != null;
> }
> ```
> or to an extension type:
> ```dart
> extension type ButtonItem._(({String label, Icon icon, void Function()? onPressed}) _) {
>   String get label => _.label;
>   Icon get icon => _.icon;
>   void Function()? get onPressed => _.onPressed;
>   ButtonItem({required String label, required Icon icon, void Function()? onPressed})
>       : this._((label: label, icon: icon, onPressed: onPressed));
>   bool get hasOnpressed => _.onPressed != null;
> }
> ```
> and then create the list of button definitions using that type's constructors:
> ```dart
> final List<ButtonItem> buttons =  [
>   ButtonItem(
>     label: "Button I",
>     icon: const Icon(Icons.upload_file),
>     onPressed: () => print("Action -> Button I"),
>   ),
>   ButtonItem(
>     label: "Button II",
>     icon: const Icon(Icons.info),
>     onPressed: () => print("Action -> Button II"),
>   )
> ];
> ```
> while not needing to change the code that uses that list.
>
> Changing any type does require the code that uses it to be very careful about
> not making assumptions. A type alias does not offer any protection from code
> using that a value is a record. An extension type offers a little protections,
> but only a class provides full abstraction and control over access to its state.


Or something :)

I might be over-embellishing, and making this too long for a simple example.
(And since I know what I want to say, it's hard to say if I have actually said it in
a reasonable order.)

MaryaBelanger:
> I might be over-embellishing, and making this too long for a simple example.

Not at all @lrhn, this is great! I was expecting just a review of the code alone, but I feel like you gave me a world-class lesson in technical writing in the process! I really appreciate your consideration of the docs :)

@Asncodes-80, I'll update the PR to incorporate Lasse's guidance and then we can get this published on the site. Thanks so much for bringing this to our attention!

MaryaBelanger:
I think this also fixes #4839 (@parlough?)

src/content/language/records.md
```diff
@@ -220,10 +220,128 @@ parallelization of futures of different types, which you can read about in the
...
+    onPressed: () => print("Action -> Button II"),
+  )
+];
+```
+
+Again, all while not needing to change the code that uses that list.
+
+Changing any type does require the code using it to be very careful about
+not making assumptions. A type alias does not offer any protection or guarantee
+for the code using it that the value is a record.
```
lrhn, line 334: This sentence is a little hard to read. 
I'd maybe put a comma after "it".

src/content/language/records.md
```diff
@@ -220,10 +220,128 @@ parallelization of futures of different types, which you can read about in the
...
+Extension types, also, offer little protection.
+Only a class can provide full abstraction and control over access to its state.
```
lrhn, line 336: I wonder if "encapsulation" is a better word than "abstraction" here. Or maybe both, with "encapsulation" being a shorter way to say "control over access to its state".

lrhn (approved):
Looks good.

src/content/language/records.md
MaryaBelanger, line 334:
```suggestion
not making assumptions. A type alias does not offer any protection or guarantee,
for the code using it as a reference, that the value being aliased is a record.
```

src/content/language/records.md
MaryaBelanger, line 336:
```suggestion
Only a class can provide full abstraction and encapsulation.
```

I like both!

MaryaBelanger (approved):

Asncodes-80:
@MaryaBelanger @lrhn Thank you for your thorough consideration.

PR #6385, opened 2025-02-03
Fix blanket `/guides` redirect

### LINKED ISSUE #6384: Missing redirects for (all?) guides
### Describe the problem

The dart guides seem to have gone - when searching for "define" on google.com or dart.dev (looking for the environment variable parameter) I get this link served in the results: https://dart.dev/guides/environment-declarations - unfortunately this leads just to https://dart.dev/docs - it should redirect to https://dart.dev/libraries/core/environment-declarations 

This seems to have been missed in https://github.com/dart-lang/site-www/issues/5767 and imho should be addressed (urgently) as the new "home" is not yet indexed and thereby not easily accessible anymore resulting in a degraded user experience (which obviously was not the intention).

### Expected fix

All guides which have been "rehomed" should have a 301 redirect to their new homes.

cc @parlough

parlough:
Fixes #6384 by moving the blanket redirect to a fallback redirect

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -208,7 +208,6 @@
       { "source": "/go/unsound-null-safety", "destination": "/null-safety/unsound-null-safety", "type": 301 },
 
       { "source": "/googleapis", "destination": "https://github.com/dart-lang/googleapis", "type": 301 },
-      { "source": "/guides{,/**}", "destination": "/docs", "type": 301 },
       { "source": "/guides/environment-declarations", "destination": "/libraries/core/environment-declarations", "type": 301 },
       { "source": "/guides/get-started", "destination": "/overview", "type": 301 },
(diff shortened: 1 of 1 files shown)


MaryaBelanger (approved):
I'm not sure if this is only because search takes you from staging to the live site, but when I search "environment" from your staging link, it takes me to dart.dev/docs (live site)

I'm just going to approve this because I'm assuming this can't really be tested from staging, but let me know if I can test it some other way!

parlough:
> I'm not sure if this is only because search takes you from staging to the live site, but when I search "environment" from your staging link, it takes me to dart.dev/docs (live site)

The current embedded search experience is essentially embedding Google Search, so it will only take you to the live indexed version.

dkbast:
👍 can confirm that this works - thanks for the quick response!

parlough:
@dkbast Thanks for noticing and calling this out!

PR #6403, opened 2025-02-10
Document F-bounds and inference using bounds for 3.7

### LINKED ISSUE #6258: [3.7] Document language feature inference-using-bounds
Design doc here: https://github.com/dart-lang/language/blob/main/accepted/future-releases/3009-inference-using-bounds/design-document.md

To do: evaluate

MaryaBelanger:
Fixes #6258

### DIFF
--- a/examples/type_system/lib/strong_analysis.dart
+++ b/examples/type_system/lib/strong_analysis.dart
@@ -156,3 +156,18 @@ void _miscDeclAnalyzedButNotTested() {
     // #enddocregion generic-type-assignment-implied-cast
   }
 }
+
+
+// #docregion inference-using-bounds
+class A<X extends A<X>> {}
+class B extends A<B> {}
+class C extends B {}
+
+void f<X extends A<X>>(X x) {}
+
+void main() {
+  f(B()); // OK.
+  f(C()); // Inference fails, compile-time error.
+  f<B>(C()); // OK.
+}
+// #enddocregion inference-using-bounds
--- a/src/content/language/generics.md
+++ b/src/content/language/generics.md
@@ -144,6 +144,7 @@ an object is a List, but you can't test whether it's a `List<String>`.
 When implementing a generic type,
 you might want to limit the types that can be provided as arguments,
 so that the argument must be a subtype of a particular type.
+This restriction is called a bound.
 You can do this using `extends`.
 
 A common use case is ensuring that a type is non-nullable
@@ -193,6 +194,30 @@ Specifying any non-`SomeBaseClass` type results in an error:
 var foo = [!Foo<Object>!]();
 ```
 
+### Self-referential type parameter restrictions
+
+When using bounds to restrict parameter types, you can refer the bound
+back to the type parameter itself. This creates a self-referential constraint,
+or an F-bound. For example:
+
+<?code-excerpt "misc/test/language_tour/generics_test.dart (f-bound)"?>
+```dart
+abstract class Comparable<T> {
+    int compareTo(T o);
+}
+
+class MyClass<T extends Comparable<T>> implements Comparable<MyClass<T>> {
+    @override
+    int compareTo(MyClass<T> other) {
+      // implementation...
+      return 0;
+    }
+}
+```
+
+The F-bound `T extends Comparable<T>` means `T` must be comparable to itself.
+So, `MyClass` can only be used with types that can be compared to
+other instances of the same type.
 
 ## Using generic methods
 
--- a/src/content/language/type-system.md
+++ b/src/content/language/type-system.md
@@ -358,6 +358,40 @@ The return type of the closure is inferred as `int` using upward information.
 Dart uses this return type as upward information when inferring the `map()`
 method's type argument: `<int>`.
 
+#### Inference using bounds
+
+:::version-note
+Inference using bounds requires a [language version][] of at least 3.7.0.
+:::
+
+Dart's type inference algorithm generates constraints by combining
+exisiting lower-bound constraints with the actual type bounds,
+not just best-effort approximations.
+This is especially important for [F-bounded][] types,
+where inference using bounds correctly infers that, in the example below,
+`C` is a subtype of `B`,
+without needing you to explicitly specify the type `f<B>(C())`:
+
+<?code-excerpt "lib/strong_analysis.dart (inference-using-bounds)"?>
+```dart
+class A<X extends A<X>> {}
+class B extends A<B> {}
+class C extends B {}
+
+void f<X extends A<X>>(X x) {}
+
+void main() {
+  f(B()); // OK.
+  f(C()); // Inference fails, compile-time error.
+  f<B>(C()); // OK.
+}
+```
+
+For more information on the inference using bounds algorithm,
+read the [desgin document][]. 
+
+[F-bounded]: /language/generics/#self-referential-type-parameter-restrictions
+[design document]: {{site.repo.dart.lang}}/blob/main/accepted/future-releases/3009-inference-using-bounds/design-document.md#motivating-example
 
 ## Substituting types
 
--- a/examples/misc/test/language_tour/generics_test.dart
+++ b/examples/misc/test/language_tour/generics_test.dart
@@ -65,3 +65,17 @@ void main() {
 }
 
 class View {}
+
+// #docregion f-bound
+abstract class Comparable<T> {
+    int compareTo(T o);
+}
+
+class MyClass<T extends Comparable<T>> implements Comparable<MyClass<T>> {
+    @override
+    int compareTo(MyClass<T> other) {
+      // implementation...
+      return 0;
+    }
+}
+// #enddocregion f-bound


MaryaBelanger:
@parlough The error from `test / analyze" is, I think", just because inference using bounds is not stable yet...?

```
error - examples/type_system/lib/strong_analysis.dart:171:3 - Couldn't infer type parameter 'X'.
          
          Tried to infer 'C' for 'X' which doesn't work:
            Type parameter 'X' is declared to extend 'A<X>' producing 'A<C>'.
          The type 'C' was inferred from:
            Parameter 'x' declared as     'X'
                          but argument is 'C'.
          
          Consider passing explicit type argument(s) to the generic.
          
           - could_not_infer
   info - examples/misc/test/language_tour/generics_test.dart:70:[16](https://github.com/dart-lang/site-www/actions/runs/13251876512/job/36991328626?pr=6403#step:6:17) - Unnecessary use of an abstract class. Try making 'compareTo' a top-level function and removing the class. - one_member_abstracts
```

For the `info` message though, do you have an idea for a better way to write that? That's how I could get it to work in dartpad on main, but lmk if there's something neater I could do. Thanks!

parlough:
> @parlough The error from test / analyze is, I think, just because inference using bounds is not stable yet...?

Yep! Now I've updated `main` to 3.7 and merged it in, analysis seems to be working now 👍 

> For the info message though, do you have an idea for a better way to write that?

No, I think that's good. I pushed a change that adds an `ignore` comment. I also added the `interface` class modifier, which I think might be a good signal to not trigger the lint. I'll explore making that change in the SDK.

chloestefantsova (approved):
The edits LGTM. Thank you so much, Marya! I'd like to known @eernstg's opinion on this too.

MaryaBelanger:
Thank _you_ Chloe! And yes I will definitely wait for Erik's input, looking forward to it!

src/content/language/generics.md
```diff
@@ -195,6 +196,30 @@ Specifying any non-`SomeBaseClass` type results in an error:
 var foo = [!Foo<Object>!]();
 ```
 
+### Self-referential type parameter restrictions
+
+When using bounds to restrict parameter types, you can refer the bound
+back to the type parameter itself. This creates a self-referential constraint,
+or an F-bound. For example:
```
MaryaBelanger, line 203:
```suggestion
or F-bound. For example:
```

src/content/language/type-system.md
```diff
@@ -372,6 +372,42 @@ The return type of the closure is inferred as `int` using upward information.
...
+
+void f<X extends A<X>>(X x) {}
+
+void main() {
+  f(B()); // OK.
+  f(C()); // Inference fails, compile-time error.
+  f<B>(C()); // OK.
+}
+```
+
```
eernstg, line 405: Finally, I think we should mention the probably most important property of this feature: It allows us to _deconstruct_ a type argument (we bind `Y` to a part of `X`). Perhaps something along these lines:

The following two functions illustrate the probably most impactful improvement based on this feature:

```dart
(X, Y) f1<X extends Iterable<Y>, Y>(X x) => (x, x.first);
(Iterable<Y>, Y) f2<Y>(Iterable<Y> x) => (x, x.first);
```

`f1` uses the type parameter `X` in order to preserve the information that the given argument is not just an `Iterable` when it is actually a `List` or a `Set`. However, with `f1`, without the feature, the type variable `Y` always gets inferred as `dynamic`, which is unsafe.

So `f2` was created in order to remember the type argument of the iterable, but with that signature the information about which kind of iterable it is gets lost every time.

```dart
// Behavior without the feature.

void main() {
  var (myList, myInt) = f1([1]);
  myInt.whatever; // No compile-time error, throws at run time.
  
  var (mySet, myString) = f2({'Hello!'});
  mySet.union({}); // Compile-time error, `mySet` has type `Iterable`.
}
```

With the feature, we don't need `f2` because `f1` works as it should:

```dart

  var (myList, myInt) = f1([1]);
  myInt.whatever; // Compile-time error, `myInt` has type `int`.
  
  var (mySet, myString) = f1({'Hello!'});
  mySet.union({}); // Works!
}
```

eernstg (approved):
Thanks for the kind words! I added a bunch of comments, hope they are helpful!

src/content/language/generics.md
```diff
@@ -195,6 +196,30 @@ Specifying any non-`SomeBaseClass` type results in an error:
 var foo = [!Foo<Object>!]();
 ```
 
+### Self-referential type parameter restrictions
```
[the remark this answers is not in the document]
MaryaBelanger: Doesn't hurt at all, thanks for the context!

src/content/language/type-system.md
```diff
@@ -372,6 +372,42 @@ The return type of the closure is inferred as `int` using upward information.
...
 method's type argument: `<int>`.
 
+#### Inference using bounds
+
+:::version-note
+Inference using bounds requires a [language version][] of at least 3.7.0.
+:::
+
+Dart's type inference algorithm generates constraints by combining
+exisiting lower-bound constraints with the actual type bounds,
```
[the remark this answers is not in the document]
MaryaBelanger: Thanks for explaining, I learned something new!

src/content/language/generics.md
```diff
@@ -195,6 +196,30 @@ Specifying any non-`SomeBaseClass` type results in an error:
...
+  int compareTo(T o);
+}
+
+class MyClass<T extends Comparable<T>> implements Comparable<MyClass<T>> {
+  @override
+  int compareTo(MyClass<T> other) {
+    // implementation...
+    return 0;
+  }
+}
```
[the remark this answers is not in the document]
MaryaBelanger: I agree. I wasn't too sure about the background / prior understanding of the audience this is applicable to, so this makes much more sense now

src/content/language/type-system.md
```diff
@@ -372,6 +372,42 @@ The return type of the closure is inferred as `int` using upward information.
...
+
+class B extends A<B> {}
+
+class C extends B {}
+
+void f<X extends A<X>>(X x) {}
+
+void main() {
+  f(B()); // OK.
+  f(C()); // Inference fails, compile-time error.
```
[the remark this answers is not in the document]
MaryaBelanger: Yeah I can see from all your suggestions that I wrote this section kind of on the line between treating this like a new feature/improvement, and trying to make it sound like this is "just the way it works". I'll clean up the whole section with that in mind, thank you for the thorough review!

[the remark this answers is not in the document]
MaryaBelanger: Maybe someone can help with code comment best practices, but I added a longer explanation here:

```dart
  f(C()); // OK. Without using bounds, inference relying on best-effort
  // approximations would fail after detecting that C is not a subtype of A<C>.
```

src/content/language/type-system.md
```diff
@@ -372,6 +372,42 @@ The return type of the closure is inferred as `int` using upward information.
...
+  f<B>(C()); // OK.
+}
+```
```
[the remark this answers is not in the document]
MaryaBelanger: Absolutely, I'll add it

src/content/language/type-system.md
MaryaBelanger: I added another paragraph and code block to the end of the section to capture this point, please let me know if it makes sense.

src/content/language/type-system.md
```diff
@@ -372,6 +372,79 @@ The return type of the closure is inferred as `int` using upward information.
...
+void main() {
+  f(B()); // OK.
+  f(C()); // OK. Without using bounds, inference relying on best-effort
+  // approximations would fail after detecting that `C` is not a subtype of `A<C>`.
+  f<B>(C()); // OK.
+}
+```
+
+Here's a more realistic example using everyday types in Dart like `int` or `num`,
+that typically have their own special properties:
```
eernstg, line 410: It might be a good idea to omit the part about the special properties: When I mentioned the special properties it was just because they could be a distraction. At this point we already have an example that clearly doesn't depend on special built-in-type-ish properties, so we just need to avoid the distraction:
```suggestion
Here's a more realistic example using everyday types in Dart like `int` or `num`:
```

src/content/language/type-system.md
```diff
@@ -372,6 +372,79 @@ The return type of the closure is inferred as `int` using upward information.
...
+
+<?code-excerpt "lib/bounded/instantiate_to_bound.dart (inference-using-bounds-2)"?>
+```dart
+X max<X extends Comparable<X>>(X x1, X x2) => x1.compareTo(x2) > 0 ? x1 : x2;
+
+void main() {
+  max(
+    3,
+    7,
+  ); // Infers `num` with the feature, would have reported an error without it.
```
eernstg, line 420: Is the following suggestion out of style? If it's OK to put the comment in front then it might be more readable:
```suggestion
  // Inferred as `max<num>(3, 7)` with the feature, fails without it.
  max(3, 7);
```

src/content/language/type-system.md
eernstg: Sounds good, added a couple of comments below.

src/content/language/type-system.md
```diff
@@ -372,6 +372,79 @@ The return type of the closure is inferred as `int` using upward information.
...
+}
+```
+
+With inference using bounds, Dart can *deconstruct* type arguments,
+extracting type information from a generic type parameter's bound.
+This allows functions like `f1` in the following example to preserve both the
```
eernstg, line 426: When we don't have `f2`, perhaps a name like `f` would suffice. (Unless there's a name clash with some other declaration named `f` nearby, but I couldn't see any).

src/content/language/type-system.md
```diff
@@ -372,6 +372,79 @@ The return type of the closure is inferred as `int` using upward information.
...
+
+void main() {
+  var (myList, myInt) = f1();
+  myInt.whatever; // Compile-time error, `myInt` has type `int`.
+
+  var (mySet, myString) = f1({'Hello!'});
+  mySet.union({}); // Works, `mySet` has type `Set<String>`.
+}
+```
+
```
eernstg, line 442: It's tempting to say a couple of words about the alternative (that is, the behavior of current Dart with the code above):

```suggestion

Without the feature, `myInt` has type `dynamic`. This implies that an expression like `myInt.whatever` is not detected as being wrong at compile time, it will instead throw at run time. Conversely, `mySet.union({})` is a compile-time error when we don't have the feature, because we can't preserve the information that `mySet` is a `Set`.
```

eernstg (approved):
Added a couple of comments.

src/content/language/generics.md
```diff
@@ -195,6 +196,31 @@ Specifying any non-`SomeBaseClass` type results in an error:
 var foo = [!Foo<Object>!]();
 ```
 
+### Self-referential type parameter restrictions (F-bounds)
```
parlough, line 199: Maybe a shorter ID here would be more manageable and easier to reference?

```suggestion
### Self-referential type parameter restrictions (F-bounds) {:#f-bounds}
```

examples/type_system/lib/strong_analysis.dart
```diff
@@ -156,3 +156,21 @@ void _miscDeclAnalyzedButNotTested() {
...
+
+class C extends B {}
+
+void f<X extends A<X>>(X x) {}
+
+void main() {
+  f(B()); // OK.
+  f(C()); // OK. Without using bounds, inference relying on best-effort
+  // approximations would fail after detecting that `C` is not a subtype of `A<C>`.
+  f<B>(C()); // OK.
```
parlough, line 173: The long comment ends up formatting in a way that makes it hard to recognize which statement the 2nd part of it is attached to. How about something like the following?

```suggestion
  f(B()); // OK.

  // OK. Without using bounds, inference relying on best-effort approximations
  // would fail after detecting that `C` is not a subtype of `A<C>`.
  f(C());

  f<B>(C()); // OK.
```

parlough (approved):
Thanks @MaryaBelanger and everyone for working so hard on figuring out what to include here and how to word it.

Looks good to me! Just one suggestion to consider:

MaryaBelanger:
Ahh sorry I merged it too fast! I'll reopen a pr for your fixes, thank you for checking

examples/type_system/lib/strong_analysis.dart
MaryaBelanger: Thanks, I didn't like how this ended up but wasn't sure what to do, so this is a big help!

PR #6416, opened 2025-02-13
Uncomment digit separators code excerpt

MaryaBelanger:
Just noticed some comments around digit separators that said to uncomment after 3.6!


parlough (approved):
Oh I forgot about this snippet! Thanks for getting these under analysis :)

PR #6423, opened 2025-02-16
Update variables.md incorrect reference to the `final bar` variable

venomdev:
Fixes the comment that mentions omitting `const` from variables defined as `final`.

### DIFF
--- a/src/content/language/variables.md
+++ b/src/content/language/variables.md
@@ -249,8 +249,8 @@ final bar = const [];
 const baz = []; // Equivalent to `const []`
 ```
 
-You can omit `const` from the initializing expression of a `const` declaration,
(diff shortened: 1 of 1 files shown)


parlough (commented):
Thanks for opening a PR @venomdev!

However, I don't think this change is currently correct.

The initializing expression of a `final` variable declaration is not inherently `const`, so you can't omit it, as seen by the `bar` example needing `const` before `[]`.

Could you expand on your reasoning behind the change or what part of the text caused confusion? I'd love to make adjustments that make the text more clear!

Thanks again :)

venomdev:
I think I may have read it incorrectly, I see the meaning of it more clearly now.

Issue #2625, opened 2020-09-07
Add dark theme


feinstein:
Could we have a dark theme for the website? It's a lot better to read at night, something like https://firebase.flutter.dev/docs/overview.

![image](https://user-images.githubusercontent.com/8120472/92414858-7ad73280-f12c-11ea-9ad2-c68885734295.png)

kwalrath:
That's pretty cool! We're on different infrastructure, so I suspect it'd be a lot of work (and thus unlikely to get done), but you never know.

baturax:
this would really be good. it hurts my eyes at night

PR #6430, opened 2025-02-18
Update alert styles to be more accessible and match Flutter docs

parlough:
Also contributes to https://github.com/dart-lang/site-www/issues/2625 as these styles will be easier to support and more accessible in a dark mode.


MaryaBelanger (approved):
Looks great!

PR #6432, opened 2025-02-18
Added version note about wildcard variables language feature of Dart 3.7

ViktorVoloshko:
Dart 3.7 added wildcard variables language feature that prevents `_` identifiers from actually declaring a variable. So multiple `_` in a single callback now possible.

Added version note about this to according guideline.

---

<details>

</details>


MaryaBelanger:
Thank you for noticing this @ViktorVoloshko, great catch! This is definitely important to update.

@munificent, what do you think about me rewriting the section completely? Or maybe removing it?

For a rewrite, it could a bunch of different ways:
- DON'T use `_`, `__`, etc. for unused callback parameters
- DO use wildcard variables for unused callback parameters
- ??

Or, we could leave it as-is with a note like the one in this PR (I'd move it right under the section heading), with a plan to remove/change/update later, when we're further from the latest version release?

MaryaBelanger (approved):
Thank you @ViktorVoloshko! I'm going to merge this PR. Other stakeholders: I might open another PR to change the title of this section all together; will link to here for reference

munificent:
Sorry for the delay, I was out all week. I agree that rewriting it eventually would be good. I would probably change the guideline to "PREFER using `_` for unused callback parameters".

Then in the prose mention that if targeting Dart < 3.7 that you may need to use `__`, etc.

Issue #6439, opened 2025-02-19
Website navigation menu is broken


samujjal-gogoi:
### Page URL

https://dart.dev

### Describe the problem

Prior to the merge of #6436, clicking on the hamburger navigation menu on the site homepage would trigger a dropdown that listed the site pages. Currently, this dropdown no longer appears.

### Additional context

Platform: Android
Browser: Chrome (v133.0.6943.50)

### I would like to fix this problem.

- [ ] I will try and fix this problem on dart.dev.

parlough:
Thanks for reporting this and identifying the cause, I appreciate it.

Sorry about the degradation in behavior. https://github.com/dart-lang/site-www/pull/6440 should fix it and in the mean time, the menu is accessible from a wider screen.

PR #6440, opened 2025-02-19
Fix menu being empty on narrow layouts

parlough:
Fixes https://github.com/dart-lang/site-www/issues/6439 which was caused by a CSS order change in https://github.com/dart-lang/site-www/pull/6436. Also fixes some of the related TODOs left by that change in order to fix the issue without reverting.

Can test by toggling the menu in a narrow layout of the page, ensuring it is visible, then ensuring the page content is still visible as well once the menu is closed.

### DIFF
--- a/src/_sass/base/_layout.scss
+++ b/src/_sass/base/_layout.scss
@@ -5,14 +5,15 @@
 }
 
 #page-content {
+  display: flex;
   min-height: calc(100vh - $site-header-height);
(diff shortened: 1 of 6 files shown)


MaryaBelanger (approved):
> Can test by toggling the menu in a narrow layout of the page, ensuring it is visible, then ensuring the page content is still visible as well once the menu is closed.

Works as described!

PR #6448, opened 2025-02-21
Wrap cookie notice buttons in very narrow layouts

parlough:
Now wraps the buttons to a new line if on a too narrow layout.

### After:
<br>

Screenshot of cookie notice with wrapping

### Before:
<br>

Screenshot of cookie notice without wrapping


MaryaBelanger (approved):
Thank you!

Issue #5019, opened 2023-06-28
Add details on `pubspec_overrides.yaml` to 'Package dependencies' page


Luckey-Elijah:
### Page URL

https://dart.dev/tools/pub/dependencies.html

### Page source

https://github.com/dart-lang/site-www/tree/main/src/tools/pub/dependencies.md

### Describe the problem

The pub dependencies page should provide information on the `pubspec_overrides.yaml` file - adding a few lines to the [dependency overrides](https://dart.dev/tools/pub/dependencies#dependency-overrides) section.

Currently, there is very little information on the `pubspec_overrides.yaml` file. The closest thing to documentation I could find was in the [`CHANGELOG for Dart 2.16`](https://github.com/dart-lang/sdk/blob/main/CHANGELOG.md#pub-3) and some scattered discussion across various issues open on varying repositories.

jonasfj:
@sigurdm this seems pretty easy to do.

We should probably cover it reference documentation, which means:
 * https://dart.dev/tools/pub/package-layout, and,
 * https://dart.dev/tools/pub/pubspec#dependencies

The package-layout should mention that the file exists, and where you can read more about it.
Same thing about pubspec.yaml reference.

Probably the best place to put the actual documentation is in a section at the end of:
https://dart.dev/tools/pub/dependencies

PR #6462, opened 2025-02-27
Document the `pubspec_overrides.yaml` file

sigurdm:
fixes https://github.com/dart-lang/site-www/issues/5019

### DIFF
--- a/src/content/tools/pub/dependencies.md
+++ b/src/content/tools/pub/dependencies.md
@@ -438,7 +438,7 @@ workspace.
 
 ## `pubspec_overrides.yaml`
 
-If you locally want to change certain aspects of your `pubspec.yaml` but do not
+If you want to change certain aspects of your `pubspec.yaml` but do not
 want to have active changes to the actual file, you can place a file called
 `pubspec_overrides.yaml` placed next to `pubspec.yaml`.
 
--- a/src/content/tools/pub/dependencies.md
+++ b/src/content/tools/pub/dependencies.md
@@ -432,9 +432,9 @@ As a result, if you publish a package to pub.dev,
 keep in mind that your package's dependency overrides
 are ignored by all users of your package.
 
-If you are using a [pub workspace][workspaces], you can place `dependency_overrides` in all
-workspace packages, but a single package can only be overridden once in the
-workspace.
+If you are using a [pub workspace][workspaces], you can
+have `dependency_overrides` in each workspace package, but
+a single package can only be overridden once in the workspace.
 
 ## `pubspec_overrides.yaml`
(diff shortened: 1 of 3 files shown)


src/content/tools/pub/dependencies.md
```diff
@@ -432,6 +432,30 @@ As a result, if you publish a package to pub.dev,
 keep in mind that your package's dependency overrides
 are ignored by all users of your package.
 
+If you are using a [pub workspace][workspaces], you can place `dependency_overrides` in all
+workspace packages, but a single package can only be overridden once in the
+workspace.
```
parlough, line 437:
```suggestion
If you are using a [pub workspace][workspaces], you can
have `dependency_overrides` in each workspace package, but
a single package can only be overridden once in the workspace.
```

src/content/tools/pub/dependencies.md
```diff
@@ -432,6 +432,30 @@ As a result, if you publish a package to pub.dev,
...
+
+## `pubspec_overrides.yaml`
+
+If you want to change certain aspects of your `pubspec.yaml` but do not
+want to have active changes to the actual file, you can place a file called
```
parlough, line 442: Could you expand on what you meant by active in "active changes to the actual file"?

Do you mean changes that would affect consumers of the package?

src/content/tools/pub/dependencies.md
```diff
@@ -432,6 +432,30 @@ As a result, if you publish a package to pub.dev,
...
+Attributes from that file will override those from `pubspec.yaml`.
+
+The properties that can be overridden are:
+
+* `dependency_overrides`
+* `workspace`
+* `resolution`
+
+This can be useful to avoid accidentally checking temporary overrides in to 
+version control. Also it can make it easier to generate overrides from a script.
```
parlough, line 454:
```suggestion
version control. It can also make it easier to generate overrides from a script. 
```

parlough (commented):
Mostly looks good to me, thanks so much for tackling this @sigurdm! I appreciate you adding it to the layout index as well.

I do have two minor wording suggestions to consider and a question about the file's introduction:

src/content/tools/pub/dependencies.md
sigurdm: I mean something like "active version control changes" but I tried to simplify, just focus on avoiding changing the main file.

parlough (approved):
Thanks for those updates! Looks good to me :)

I pushed a commit with some minor formatting and copy changes as well as added a redirect from `dart.dev/to/pubspec-overrides` to the new section :)

PR #6464, opened 2025-02-28
Enable searching by term and alternate terms on glossary

parlough:
Enables searching glossary entries by part of a term or exactly matching an alternate version of the word.

Reuses (and now shares) the styles from the search experience on the linter rules index.

**Staged:** https://dart-dev--pr6464-feat-searchable-glossary-brmmfv5o.web.app/resources/glossary


antfitch (approved):
Looks great!

PR #6474, opened 2025-03-06
Fix incorrect code in inference using bounds

mnordine:
The function calls were using the wrong name, first example was missing its argument. I believe the intention was to pass in a `List<int>`, perhaps `[1]`.

### DIFF
--- a/src/content/language/type-system.md
+++ b/src/content/language/type-system.md
@@ -432,10 +432,10 @@ without losing type safety or specific type information.
 (X, Y) f<X extends Iterable<Y>, Y>(X x) => (x, x.first);
 
 void main() {
-  var (myList, myInt) = f1();
+  var (myList, myInt) = f([1]);
   myInt.whatever; // Compile-time error, `myInt` has type `int`.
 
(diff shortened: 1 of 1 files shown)


eernstg (approved):
That's correct! The function was named `f1` because there was another one (`f2`) illustrating the "before" and "after" version of the static analysis. It was abbreviated and the function was renamed, and the mistakes were introduced at that point. Thanks!

Some background can be seen here: https://github.com/dart-lang/site-www/pull/6403/files/020b98ff0944a7368e648ff6ca0c8bc1f43d30f6#diff-d028b328d07abdc9a45ebc82b5227850fac5c6c61725226f87d4ee515e4333c4, in the comment on line 405 of 'type-system.md'.

eernstg:
@antfitch, I'm happy about these changes. However, I'm not in the team that normally makes the final decision about site-www PRs, could you take a look?

mnordine:
> https://dart-dev--pr6474-patch-1-ebmhrdto.web.app/

lgtm

antfitch (commented):
LGTM!

PR #6475, opened 2025-03-06
Add anchors for replaced Effective Dart entries

parlough:
These sections were replaced in https://github.com/dart-lang/site-www/commit/eb1ad755caa1f1fc409eb15043c4632c23d44430. When removing or renaming a section header, especially in Effective Dart, it's important to ensure links to the entries don't break as there might be existing references across the internet, in various chats, LLM training data, etc.


antfitch (approved):
Looks great!

PR #6476, opened 2025-03-07
Cross link from diagnostic message docs to related lint rule

parlough:
Adds a `Lint rule` tag after the header of the diagnostics resulting from enabled lints.

Screenshot of the tag for the prefer_relative_imports diagnostic


antfitch (approved):
This looks great.

PR #6485, opened 2025-03-11
Separate primary build check from other workflows

parlough:
This will highlight that this is the most important check and also fix the broken  build status image in the [README](https://github.com/dart-lang/site-www/blob/main/README.md).

Also removes some unneeded _(from previous infrastructure setups)_ environment configuration from the existing workflows.


antfitch (approved):
LGTM!

PR #6493, opened 2025-03-17
Explain handling of lost domain name ownership wrt. verified publishers

sigurdm:


src/content/tools/pub/verified-publishers.md
```diff
@@ -39,6 +39,17 @@ pub.dev verifies that the user creating the verified publisher has
 admin access to the associated ["Domain Property"][domain-prop],
 based on existing logic in the [Google Search Console.][search-console]
 
+:::note
+Domain name ownership is only verified at the time of publisher creation.
```
antfitch, line 43: Domain name ownership is verified only once when a publisher is created. After that:

- Losing control of a domain does not cause the original publisher owner to lose access to the publisher.
- Acquiring a domain does not grant the new owner any rights to a publisher that was previously associated with it.
- Publisher ownership must be explicitly transferred by the current publisher owner.

antfitch (approved):

mit-mit (approved):

PR #6522, opened 2025-04-08
Documentation style: Document conceptual properties as properties

lrhn:
Say explicitly that a function that is named like a property (noun phrase or non-imperative verb phrase, as described in `design.md`) should also be documented like a property (noun-phrase or "Whether ...", basically).

(I keep seeing people documenting a noun-phrased function with "Returns noun-phrase", just because the style guide does not have any exceptions for using verb-phrases for functions. If it's treated as a property, which is allowed, it should also be documented as a property.)

Tweaks some examples to not contradict other rules. For example, don't have a two-statement first paragraph "good" example after the "single statement first paragraph" rule.)

(Will update good/bad examples too..)

### DIFF
--- a/examples/misc/lib/effective_dart/docs_good.dart
+++ b/examples/misc/lib/effective_dart/docs_good.dart
@@ -31,10 +31,14 @@ void miscDeclAnalyzedButNotTested() {
 
   <Flag>() {
     // #docregion no-annotations
-    /// Defines a flag.
+    /// Defines a flag with the given [name] and [abbr] as abbreviation.
     ///
-    /// Throws an [ArgumentError] if there is already an option named [name] or
-    /// there is already an option using abbreviation [abbr]. Returns the new flag.
+    /// The [name] and [abbr] strings must not be empty.
+    ///
+    /// Returns a new flag.
+    ///
+    /// Throws a [DuplicateFlagException] if there is already an option named [name]
+    /// or there is already an option using the abbreviation [abbr].
     Flag addFlag(String name, String abbr) => ellipsis();
     // #enddocregion no-annotations
   };
@@ -64,19 +68,17 @@ void miscDeclAnalyzedButNotTested() {
 
   <T>() {
     // #docregion third-person
-    /// Returns `true` if every element satisfies the [predicate].
-    bool all(bool predicate(T element)) => ellipsis();
-
+    /// Connects to server and fetches results of query.
+    Stream<QueryResult> fetchResults(Query query) => ellipsis();
+      
     /// Starts the stopwatch if not already running.
-    void start() {
-      ellipsis();
-    }
+    void start() => ellipsis();
 
     // #enddocregion third-person
   };
 
   // #docregion code-sample
-  /// Returns the lesser of two numbers.
+  /// The lesser of two numbers.
   ///
   /// ```dart
   /// min(5, 3) == 3
@@ -157,6 +159,8 @@ void miscDeclAnalyzedButNotTested() {
   // #enddocregion markdown
 }
 
+class DuplicateFlagException implements Exception {}
+
 class IOError {}
 
 class PermissionError {}
@@ -165,8 +169,9 @@ class Widget {}
 
 // #docregion redundant
 class RadioButtonWidget extends Widget {
-  /// Sets the tooltip to [lines], which should have been word wrapped using
-  /// the current font.
+  /// Sets the tooltip to [lines].
+  ///
+  /// The lines should be word wrapped using the current font.
   void tooltip(List<String> lines) {
     ellipsis();
   }
@@ -222,6 +227,18 @@ class Modal {
 
 //----------------------------------------------------------------------------
 
+abstract class Iterable<E> {
+  // @docregion noun-for-func-returning-value
+  /// The [index]th element of this iterable in iteration order.
+  E elementAt(int index);
+  
+  /// Whether this iterable contains an element equal to [element].
+  bool contains(Object? element);
+  // #enddocregion noun-for-func-returning-value
+}
+
+//----------------------------------------------------------------------------
+
 class Pool {
   // #docregion getter-and-setter
   /// The pH level of the water in the pool.
@@ -259,10 +276,10 @@ class ToggleComponent {}
 
 // #docregion this
 class Box {
-  /// The value this wraps.
+  /// The value this box wraps.
   Object? _value;
 
-  /// True if this box contains a value.
+  /// Whether this box contains a value.
   bool get hasValue => _value != null;
 }
 
--- a/examples/misc/lib/effective_dart/docs_good.dart
+++ b/examples/misc/lib/effective_dart/docs_good.dart
@@ -165,6 +165,10 @@ class IOError {}
 
 class PermissionError {}
 
+class Query {}
+
+class QueryResult {}
+
 class Widget {}
 
 // #docregion redundant
@@ -227,7 +231,7 @@ class Modal {
 
 //----------------------------------------------------------------------------
 
-abstract class Iterable<E> {
+abstract class MyIterable<E> {
   // #docregion noun-for-func-returning-value
   /// The [index]th element of this iterable in iteration order.
   E elementAt(int index);
--- a/examples/misc/lib/effective_dart/docs_good.dart
+++ b/examples/misc/lib/effective_dart/docs_good.dart
@@ -37,8 +37,8 @@ void miscDeclAnalyzedButNotTested() {
     ///
     /// Returns a new flag.
     ///
-    /// Throws a [DuplicateFlagException] if there is already an option named [name]
-    /// or there is already an option using the abbreviation [abbr].
+    /// Throws a [DuplicateFlagException] if there is already an option named
+    /// [name] or there is already an option using the abbreviation [abbr].
     Flag addFlag(String name, String abbr) => ellipsis();
     // #enddocregion no-annotations
   };
@@ -70,7 +70,7 @@ void miscDeclAnalyzedButNotTested() {
     // #docregion third-person
     /// Connects to server and fetches results of query.
     Stream<QueryResult> fetchResults(Query query) => ellipsis();
-      
+
     /// Starts the stopwatch if not already running.
     void start() => ellipsis();
 
@@ -231,7 +231,7 @@ abstract class Iterable<E> {
   // #docregion noun-for-func-returning-value
   /// The [index]th element of this iterable in iteration order.
   E elementAt(int index);
-  
+
   /// Whether this iterable contains an element equal to [element].
   bool contains(Object? element);
   // #enddocregion noun-for-func-returning-value
--- a/examples/misc/lib/effective_dart/docs_good.dart
+++ b/examples/misc/lib/effective_dart/docs_good.dart
@@ -68,7 +68,7 @@ void miscDeclAnalyzedButNotTested() {
 
   <T>() {
     // #docregion third-person
-    /// Connects to server and fetches results of query.
+    /// Connects to the server and fetches the query results.
     Stream<QueryResult> fetchResults(Query query) => ellipsis();
 
     /// Starts the stopwatch if not already running.
--- a/examples/misc/lib/effective_dart/docs_good.dart
+++ b/examples/misc/lib/effective_dart/docs_good.dart
@@ -228,7 +228,7 @@ class Modal {
 //----------------------------------------------------------------------------
 
 abstract class Iterable<E> {
-  // @docregion noun-for-func-returning-value
+  // #docregion noun-for-func-returning-value
   /// The [index]th element of this iterable in iteration order.
   E elementAt(int index);
   
--- a/examples/misc/lib/effective_dart/docs_good.dart
+++ b/examples/misc/lib/effective_dart/docs_good.dart
@@ -31,15 +31,15 @@ void miscDeclAnalyzedButNotTested() {
 
   <Flag>() {
     // #docregion no-annotations
-    /// Defines a flag with the given [name] and [abbr] as abbreviation.
+    /// Defines a flag with the given [name] and [abbreviation].
     ///
-    /// The [name] and [abbr] strings must not be empty.
+    /// The [name] and [abbreviation] strings must not be empty.
     ///
     /// Returns a new flag.
     ///
     /// Throws a [DuplicateFlagException] if there is already an option named
-    /// [name] or there is already an option using the abbreviation [abbr].
-    Flag addFlag(String name, String abbr) => ellipsis();
+    /// [name] or there is already an option using the [abbreviation].
+    Flag addFlag(String name, String abbreviation) => ellipsis();
     // #enddocregion no-annotations
   };
 
--- a/src/content/effective-dart/_toc.md
+++ b/src/content/effective-dart/_toc.md
@@ -38,7 +38,7 @@ the project:
 
 * <a href='/effective-dart/style#do-format-your-code-using-dart-format'>DO format your code using <code>dart format</code>.</a>
 * <a href='/effective-dart/style#consider-changing-your-code-to-make-it-more-formatter-friendly'>CONSIDER changing your code to make it more formatter-friendly.</a>
-* <a href='/effective-dart/style#avoid-lines-longer-than-80-characters'>AVOID lines longer than 80 characters.</a>
+* <a href='/effective-dart/style#prefer-lines-80-characters-or-fewer'>PREFER lines 80 characters or fewer.</a>
 * <a href='/effective-dart/style#do-use-curly-braces-for-all-flow-control-statements'>DO use curly braces for all flow control statements.</a>
 
 </div>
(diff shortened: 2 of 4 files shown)


src/content/effective-dart/_toc.md
```diff
@@ -38,7 +38,7 @@ the project:
 
 * <a href='/effective-dart/style#do-format-your-code-using-dart-format'>DO format your code using <code>dart format</code>.</a>
 * <a href='/effective-dart/style#consider-changing-your-code-to-make-it-more-formatter-friendly'>CONSIDER changing your code to make it more formatter-friendly.</a>
-* <a href='/effective-dart/style#avoid-lines-longer-than-80-characters'>AVOID lines longer than 80 characters.</a>
+* <a href='/effective-dart/style#prefer-lines-80-characters-or-fewer'>PREFER lines 80 characters or fewer.</a>
```
lrhn, line 41: (This link was already dead. Running `./dash_site effective-dart`  updated it.

examples/misc/lib/effective_dart/docs_good.dart
```diff
@@ -64,19 +68,17 @@ void miscDeclAnalyzedButNotTested() {
 
   <T>() {
     // #docregion third-person
-    /// Returns `true` if every element satisfies the [predicate].
-    bool all(bool predicate(T element)) => ellipsis();
```
lrhn, line 68: This is precisely the kind of function this change is about, so it's not a "good" example of using a verb-documentation.
(It's also the example people keep pointing to as why they use "Returns".)

src/content/effective-dart/documentation.md
```diff
@@ -231,19 +232,17 @@ It's better to say nothing
...
 
-### PREFER starting function or method comments with third-person verbs
+### PREFER starting comments of a function or method with third-person verbs if its main purpose is a side effect
 
 The doc comment should focus on what the code *does*.
 
 <?code-excerpt "docs_good.dart (third-person)"?>
 ```dart tag=good
-/// Returns `true` if every element satisfies the [predicate].
-bool all(bool predicate(T element)) => ...
```
lrhn, line 241: This is precisely the kind of function this change is about, so it's not a "good" example of using a verb-documentation.
(It's also the example people keep pointing to as why they use "Returns".)

src/content/effective-dart/documentation.md
```diff
@@ -285,6 +284,36 @@ cases, usage of "or not" with "whether" is superfluous and can be omitted,
 especially when used in this context.
 :::
 
+### PREFER a noun phrase or non-imperative verb phrase for a function or method if returning a value is its primary purpose
```
lrhn, line 287: This is the new rule, mathcing the similarly titled rule for function names. Naming and documentation should match up, and this now says so explicitly.

src/content/effective-dart/documentation.md
```diff
@@ -393,12 +424,22 @@ Flag addFlag(String name, String abbr) => ...
 The convention in Dart is to integrate that into the description of the method
 and highlight parameters using square brackets.
 
+Consider having sections starting with "The \[paramameter\]" to describe 
+parameters, with "Returns" for the returned value and "Throws" for exceptions.
+Errors can be documented the same way as exceptions,
+or just as requirements that must be satified, without documenting the
+precise error which will be thrown.
```
lrhn, line 431: Just the guidelines we use in the platform libraries - paragraphs starting with "Returns" or "Throws" are easy to scan for.

src/content/effective-dart/documentation.md
```diff
@@ -548,14 +588,15 @@ think.
 
 When documenting a member for a class, you often need to refer back to the
 object the member is being called on. Using "the" can be ambiguous.
+Prefer having some qualifier after "this", a sole "this" can ambiguous too.
```
lrhn, line 591: (The "good" example of "The value this wraps." doesn't read well. I find that having an unqualifiied "this", "that", or a plain "it" usually makes documentation harder to read.)

examples/misc/lib/effective_dart/docs_good.dart
```diff
@@ -31,10 +31,14 @@ void miscDeclAnalyzedButNotTested() {
 
   <Flag>() {
     // #docregion no-annotations
-    /// Defines a flag.
+    /// Defines a flag with the given [name] and [abbr] as abbreviation.
```
antfitch, line 34:
```suggestion
    /// Creates a new flag.
    ///
    /// Provide a unique [name] for the flag and a short abbreviation ([abbr]).
    /// Both the [name] and [abbr] must be non-empty strings.
    ///
    /// Returns a new flag.
    ///
    /// Throws a [DuplicateFlagException] if a flag with [name] or
    /// [abbr] already exists.
```

examples/misc/lib/effective_dart/docs_good.dart
```diff
@@ -64,19 +68,17 @@ void miscDeclAnalyzedButNotTested() {
 
   <T>() {
     // #docregion third-person
-    /// Returns `true` if every element satisfies the [predicate].
-    bool all(bool predicate(T element)) => ellipsis();
+    /// Connects to server and fetches results of query.
```
antfitch, line 71: YES! We should be doing this everywhere (in my opinion) and especially in this style guide.

examples/misc/lib/effective_dart/docs_good.dart
```diff
@@ -64,19 +68,17 @@ void miscDeclAnalyzedButNotTested() {
...
-      ellipsis();
-    }
+    void start() => ellipsis();
 
     // #enddocregion third-person
   };
 
   // #docregion code-sample
-  /// Returns the lesser of two numbers.
+  /// The lesser of two numbers.
```
antfitch, line 81:
```suggestion
  /// Gets the lesser of two numbers.
```

I think we should always start with a verb BUT we should avoid using `Returns` since it's self explanatory. I would advise that we use `produces` or `gets` when there isn't a stronger verb.

examples/misc/lib/effective_dart/docs_good.dart
```diff
@@ -157,16 +159,23 @@ void miscDeclAnalyzedButNotTested() {
...
+
 class Widget {}
 
 // #docregion redundant
 class RadioButtonWidget extends Widget {
-  /// Sets the tooltip to [lines], which should have been word wrapped using
-  /// the current font.
+  /// Sets the tooltip to [lines].
+  ///
+  /// The lines should be word wrapped using the current font.
```
antfitch, line 178:
```suggestion
  /// Sets the tooltip to [lines].
  /// [lines] represents a list of strings and each string
  /// in this list represents a line of text in the tooltip.
  ///
  /// The lines should be word wrapped using the current font.
```

We really should be defining parameters, in my opinion. I have no idea what `lines` is supposed to represent. A chat for a later date. For now, I've added an update to hopefully capture what `lines` is for.

examples/misc/lib/effective_dart/docs_good.dart
```diff
@@ -157,16 +159,23 @@ void miscDeclAnalyzedButNotTested() {
...
+
+class QueryResult {}
+
 class Widget {}
 
 // #docregion redundant
 class RadioButtonWidget extends Widget {
-  /// Sets the tooltip to [lines], which should have been word wrapped using
-  /// the current font.
+  /// Sets the tooltip to [lines].
```
antfitch, line 176:
```suggestion
  /// Sets the tooltip to [lines].
  /// [lines] represents a list of strings and each string
  /// in this list represents a line of text in the tooltip.
  ///
  /// The lines should be word wrapped using the current font.
```

We really should be defining parameters, in my opinion. I have no idea what `lines` is supposed to represent. A chat for a later date. For now, I've added an update to hopefully capture what `lines` is for.

examples/misc/lib/effective_dart/docs_good.dart
```diff
@@ -259,10 +280,10 @@ class ToggleComponent {}
 
 // #docregion this
 class Box {
-  /// The value this wraps.
+  /// The value this box wraps.
   Object? _value;
 
-  /// True if this box contains a value.
+  /// Whether this box contains a value.
```
antfitch, line 286:
```suggestion
  /// Checks if this box contains a value.
```

src/content/effective-dart/_toc.md
antfitch: Thank you for fixing!

src/content/effective-dart/documentation.md
```diff
@@ -231,19 +232,17 @@ It's better to say nothing
 than waste a reader's time telling them something they already know.
 
 
-### PREFER starting function or method comments with third-person verbs
+### PREFER starting comments of a function or method with third-person verbs if its main purpose is a side effect
```
antfitch, line 235:
```suggestion
### PREFER starting comments of a function or method with third-person verbs
```

I would avoid that last part and always lean towards always using a verb. See my notes in the previous file. Even with functions that don't return something, they are doing something. If we need to chat more, let me know.

src/content/effective-dart/documentation.md
```diff
@@ -231,19 +232,17 @@ It's better to say nothing
...
 
 The doc comment should focus on what the code *does*.
 
 <?code-excerpt "docs_good.dart (third-person)"?>
 ```dart tag=good
-/// Returns `true` if every element satisfies the [predicate].
-bool all(bool predicate(T element)) => ...
+/// Connects to server and fetches results of query.
```
antfitch, line 241:
```suggestion
/// Connects to the server and fetches the query results.
```

src/content/effective-dart/documentation.md
antfitch: I really think we should revisit this... I think we are unnecessary complications by adding this new rule. Why not simply state that 'Returns' should not be used? I'd love to chat about this with whoever agreed on this rule.

src/content/effective-dart/documentation.md
antfitch: I agree. The qualifier helps quite a bit. Nice!

src/content/effective-dart/documentation.md
antfitch: Let's chat. I'd like to learn more about this rule and the history behind the push to include it. I'm worried that we're overcomplicating things to avoid using 'Returns' in our function summaries. I agree that 'Returns' should not be used but that we probably don't need to go to this extent to define a new rule.

src/content/effective-dart/documentation.md
parlough: I believe this is more about functions whose purpose is to return something without side effects. In that case, developers shouldn't care what "something" the function is doing (or how it's doing it), they care about the value it returns.

antfitch:
@munificent looping you into this discussion.

examples/misc/lib/effective_dart/docs_good.dart
parlough: Yes, this is technically a function but it's a property, but developers shouldn't need to care about the code it runs, only what the property it provides access to is.

src/content/effective-dart/documentation.md
antfitch: I totally agree with you. I have a slightly different approach to dealing with these. There is an elegant way of dealing with function summaries in the GoogleSQL reference docs that I feel we should pull over here. Example:

https://cloud.google.com/bigquery/docs/reference/standard-sql/string_functions

examples/misc/lib/effective_dart/docs_good.dart
parlough: We do cover parameters in https://dart.dev/effective-dart/documentation#do-use-prose-to-explain-parameters-return-values-and-exceptions.

In this specific case, I think it's fine to keep this example short to focus on the effective dart entry at hand ([AVOID redundancy with the surrounding context](https://dart.dev/effective-dart/documentation#avoid-redundancy-with-the-surrounding-context)).

examples/misc/lib/effective_dart/docs_good.dart
lrhn: Or, to be consistent with naming, avoid the abbreviation "abbr". (Slightly ironic to abbreviate "abbreviation" :grin:.)

examples/misc/lib/effective_dart/docs_good.dart
lrhn: I disagree very strongly.

My goal here is to break away from the unconditional rule that function docs must start with a verb, because it's inconsistent with treating some functions as properties when naming them.

If the function is *named* using a noun phrase or predicate, it's because it is considered a (parameterized) *property*. Then it should be documented the same way it's named.

We also have a naming rule that says "don't start a function name with `get`."
You shouldn't start a DartDoc with "Gets" either, for all the same reasons.

The *reason* for that rule is that it doing so is a sign that the function is really a property.
We then have another rule which says to name such a function it like a property.
That should not just mean *name* it, but to treat it like a property in *every* way, including documentation, and we have rules for documenting properties that don't start with a verb.
Making one a noun and the other a verb makes the documentation and name feel out of sync.

So just "The foo", it's cleaner. The "Gets" adds nothing, but it weakens the connection between name and doc.

Compare:
```dart
/// The first element of this list.
E get first;

/// The [index]th element of this list.
E operator [])(int index);

/// The [index]th element of this iterable.
E elementAt(int index);
```

Adding "Gets" to the `operator[]` function or `elementAt` function doesn't make anything clearer.
It doesn't say anything other than what "Returns" say.

src/content/effective-dart/documentation.md
lrhn: Then people just start using "Gets the foo" instead of "Returns the foo", when the consistent (with member naming) documentation would just be "The foo". And "Gets" isn't better than "Returns", it's still a no-content verb added just to have a verb, in a place where no verb is needed.

The reason to not use a verb for documentation is the same reason as to not use a verb for naming - it's a property, not an operation, so it should be named for its value, not its computation.
The name and documentation should match.

Maybe *this* rule is unnecessary. If we instead added to the naming rule that "If you do this, document the member like a property too.", then we might get the same effect. I would still want to say something on the documentation page to remind people that "always use a verb phrase for a function" isn't actually a universal rule, it only applies to functions that aren't really (being treated like) a property.

examples/misc/lib/effective_dart/docs_good.dart
lrhn: Agree. I chose to not add more text here, because the original didn't have it, but in a review I would have asked for a description of what `lines` *means*, both its format and how it will be used.

examples/misc/lib/effective_dart/docs_good.dart
lrhn: Agree, and my primary point of this CL: It's a property. It's already recognized and named as one, so it should be documented as one too. And a boolean property would use `Whether`.

Also "Checks if" doesn't say what the result of that check is used for. With that doc, it could also be a `void` founction that throws if the check fails. It would have to be followed by a `Returns ...`, and that would likely be `Returns whether this box contains a value.`, which duplicates almost all of the first line.

src/content/effective-dart/documentation.md
lrhn: Indeed, and that's the change I'm going for here. Don' use verb phrases for functions that are really properties. Not if you already used a non-verb name for them.

antfitch:
Has a proposal for this change already been created and approved? If yes, can you attach it? If not, can you create that and send it through a review?

examples/misc/lib/effective_dart/docs_good.dart
antfitch: I see where you're coming from now. Stepping back from my suggestion.

src/content/effective-dart/documentation.md
antfitch: I see where you're coming from now after I mapped the sections in Documentation and Design. Reverted my suggestion.

src/content/effective-dart/documentation.md
antfitch: I see where you're coming from now after I mapped the sections in Documentation and Design. Reverted my suggestion.

examples/misc/lib/effective_dart/docs_good.dart
antfitch: I agree! Maybe change it if you want to.

src/content/effective-dart/documentation.md
antfitch: Resolving since I have a better idea of how things map with each other now.

examples/misc/lib/effective_dart/docs_good.dart
antfitch: Got it!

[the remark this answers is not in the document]
antfitch: Meh, let's just leave it.

antfitch (approved):
Generally okay with things.

parlough (approved):
I pushed some minor typo and copy fixes and formatting cleanup, but overall looks great to me. I think it makes a lot of sense to clarify these cases this way. The existing examples being more consistent is nice as well. Thanks!

Issue #2093, opened 2019-11-14
Improve  Control Flow Collections section


johnpryan:
## Page

https://dart.dev/language/operators

## Notes

The [feature spec](https://github.com/dart-lang/language/blob/master/accepted/2.3/control-flow-collections/feature-specification.md) for control flow collections has a few more examples than [the language tour](https://dart.dev/guides/language/language-tour#collection-operators). 

It would be great if searching for "Dart ui-as-code" or "dart control flow collections"  brought up a page with a quick explanation and some examples.

MaryaBelanger:
Leaving a TODO to make sure [this link](https://dart.dev/language/collections#collection-operators:~:text=proposal%20and%20the-,control%20flow%20collections%20proposal.,-For%20more%20information) to the proposal at the end of the [Maps](https://dart.dev/language/collections#maps) section gets updated.

Also, I don't think Collection operators (and probably Spread operators too) needs to be under List - they apply to other collections as well. So, another TODO, we should move them to be top level sections on the Collections page (or, top level pages under the "Collections" umbrella, if that's how that content ends up being organized)

atsansone:
This relates to https://github.com/dart-lang/site-www/issues/2631.

PR #6525, opened 2025-04-10
Collection elements

antfitch:
This PR refactors the `Operators` section of our `Collections` guide. This change is to prepare for the addition of the null-aware elements for collections.

Preview:

* https://dart-dev--pr6525-dart-null-awareness-elements-ujtoc8s3.web.app/language/collections#supplementals

Overview of changes:

* Replaced `Operators` with `Collection elements`.

* I have moved each supplemental item into its own section so that they are easier to find in the right TOC.

* The examples were difficult for me to follow and the output of the examples was not provided. I added output and updated the examples to be easy for someone to copy and paste into DartPad and test.

* I think we needed to clarify what type of null is supported by our spread operators. Not supported (list = null), Supported (list = [1, null, 2]).

History:

See https://github.com/dart-lang/site-www/issues/2631#issue-698613262 for some other background.

Resolves https://github.com/dart-lang/site-www/issues/2093

---

<details>

</details>


examples/misc/lib/language_tour/misc/null_aware_spread_operator_in_collection_a.dart
```diff
@@ -0,0 +1,9 @@
+// void main() {
+//   // #docregion code_sample
+//   var a = null;
```
munificent, line 3: This line will need to look something like:

```dart
List<int>? a = null;
```

(Otherwise, type inference infers `a` to have type `Null`, which is true but not super useful.)

examples/misc/lib/language_tour/misc/null_aware_spread_operator_in_collection_a.dart
```diff
@@ -0,0 +1,9 @@
...
+//   var b = [1, 2, null];
```
munificent, line 4: Nit: It might be a little clearer as `1, null, 2` and likewise below. That way it's little more obvious that the `null` in the resulting list comes from the middle of this one.

src/content/language/collections.md
```diff
@@ -242,78 +250,180 @@ final constantMap = const {2: 'helium', 10: 'neon', 18: 'argon'};
...
 
-## Operators
+[Using constructors]: /language/classes#using-constructors
+[`Map`]: {{site.dart-api}}/dart-core/Map-class.html
+[type inference]: /language/type-system#type-inference
+
+## Supplementals
 
-### Spread operators
+The following supplemental materials can be used with
```
munificent, line 259: The language spec calls these "elements". So a list literal contains a series of elements, each of which might be an expression (i.e. `1`), a spread element (`...foo`)`, an if element, or a for element.

We don't *have* to follow the spec terminology. We don't always do a great job of coming up with names for stuff when we specify them. But we try and in this case, I think the name works OK.

It's up to you, though. :)

src/content/language/collections.md
```diff
@@ -242,78 +250,180 @@ final constantMap = const {2: 'helium', 10: 'neon', 18: 'argon'};
...
+<a id="spread-operators"></a>
 
-For example, you can use the spread operator (`...`) to insert
-all the values of a list into another list:
+### Spread operations
+
+With a spread operation, you can
+add values from one collection to another collection. Null
+values inside of a  collection can be passed in, but null
+collections produce an error. A spread operation has this
```
munificent, line 270: I don't know if it's worth pointing out that the thing you spread can't be `null`. There are many places in the language where `null` is an error (like `1 + null` or `null.someMethod`) and we don't usually bother to call that out.

If your intent is to contrast it with null-aware spread, that makes sense.

src/content/language/collections.md
```diff
@@ -242,78 +250,180 @@ final constantMap = const {2: 'helium', 10: 'neon', 18: 'argon'};
...
 
-For more details and examples of using the spread operator, see the
-[spread operator proposal.][spread proposal]
+In the following example, an error is produced because
+`a` is null and can't be passed into `items`.
+
+```dart
+var a = null;
+var items = [0, ...a, 4]; // Error
+```
```
munificent, line 292: If it were me, I would omit the discussion of null here and then talk about null handling under null-aware spread below.

Maybe something like:

---

### Null-aware spread operations

Sometimes you want to include the elements of one collection in another,
but the former is also nullable:

```dart
List<String> buildCommandLine(
  String executable,
  List<String> options, [
  List<String>? extraOptions,
]) {
  return [
    executable,
    ...options,
    ...extraOptions, // <-- Error!   
  ];
}
```

Because of null safety, Dart won't let you perform an operation on a value that might be null, so this has a compile error.

If you want to spread a nullable collection, you can use a null-aware spread:

```dart
List<String> buildCommandLine(
  String executable,
  List<String> options, [
  List<String>? extraOptions,
]) {
  return [
    executable,
    ...options,
    ...?extraOptions, // <-- OK now.
  ];
}
```

When using `...?`, if the collection is `null`, then no elements are added. Note that if the collection itself is not `null`, but it contains elements that are, those `null` elements will be added to the result.

src/content/language/collections.md
```diff
@@ -242,78 +250,180 @@ final constantMap = const {2: 'helium', 10: 'neon', 18: 'argon'};
...
+
+[Spread operator]: /language/operators/#spread-operators
 
 <a id="collection-operators"></a>
-### Control-flow operators
 
-Dart offers **collection if** and **collection for** for use in list, map,
-and set literals. You can use these operators to build collections using
-conditionals (`if`) and repetition (`for`).
+### If statements
```
munificent, line 327: It depends on how precise/pedantic you want to be, but these technically aren't if statements, they're if elements. The distinction matters somewhat in that the thing inside an if statement's then and else branch is always a statement, and can be any statement:

```dart
if (condition) return x; // Return is a statement.
```

But the thing inside an if element must always be an element:

```dart
var list = [
  if (condition) return x; // Error! Can't have a return here.
];
```

src/content/language/collections.md
```diff
@@ -242,78 +250,180 @@ final constantMap = const {2: 'helium', 10: 'neon', 18: 'argon'};
...
+```dart
+// boolean expression must be true to include result
+if (<bool_expression>) <result>
+```
 
-<?code-excerpt "misc/test/language_tour/built_in_types_test.dart (list-if)"?>
 ```dart
-var nav = ['Home', 'Furniture', 'Plants', if (promoActive) 'Outlet'];
+// boolean expression must be false to include result
+if (!<bool_expression>) <result>
```
munificent, line 341: I don't think it's worth showing `!` here because there's really nothing particular to if elements.

But I do think it's worth showing an example of an if element with an else clause:

```dart
var menu = [
  if (isTuesday)
    'tacos'
  else
    'spaghetti',
];
```

src/content/language/collections.md
```diff
@@ -242,78 +250,180 @@ final constantMap = const {2: 'helium', 10: 'neon', 18: 'argon'};
...
+You can use an `if` statement inside of a collection to
+conditionally include values in the collection based on a
+boolean expression. An `if` statement has this signature in
+a collection:
 
-Here's an example of using **collection if**
-to create a list with three or four items in it:
+```dart
+// boolean expression must be true to include result
+if (<bool_expression>) <result>
```
munificent, line 336: I think this might be clearer if the example was wrapped in a collection literal:

```dart
var list = [
  if (<bool_expression>) <result>
];
```

src/content/language/collections.md
```diff
@@ -242,78 +250,180 @@ final constantMap = const {2: 'helium', 10: 'neon', 18: 'argon'};
...
 
-Dart also supports [if-case][] inside collection literals:
+In the following example, `1` is included in `items`
+because `includeItem` evaluates as true, but `2` is not
+included because `!includeItem` evaluates as false.
 
 ```dart
-var nav = ['Home', 'Furniture', 'Plants', if (login case 'Manager') 'Inventory'];
+var includeItem = true;
+var items = [0, if (includeItem) 1, if (!includeItem) 2, 3]; // [0, 1, 3]
```
munificent, line 350: Yeah, in this case, else would certainly be simpler:

```dart
var items = [0, if (includeItem) 1 else 2, 3]; // [0, 1, 3]
```

(Tangent: I was doing a readability review a while back and the author had code exactly like you have here. I pointed out that they could use an `else` instead. They replied "I don't think the language lets you do that." I said, "I'm pretty sure it does... because I designed that language feature." :D)

src/content/language/collections.md
```diff
@@ -242,78 +250,180 @@ final constantMap = const {2: 'helium', 10: 'neon', 18: 'argon'};
...
+conditionally include values in the collection if the
+matching expressions evaluate as true. An `if-case`
+statement has this signature in a collection:
 
-<?code-excerpt "misc/test/language_tour/built_in_types_test.dart (list-for)"?>
 ```dart
-var listOfInts = [1, 2, 3];
-var listOfStrings = ['#0', for (var i in listOfInts) '#$i'];
-assert(listOfStrings[1] == '#1');
+if (<expression> case expression_literal) <result>
```
munificent, line 366: The thing after `case` here is a [pattern](https://dart.dev/language/patterns), not an expression. (It's a little confusing because there are patterns that are simply constant expressions, so you can use an expression here. But you can also use things that aren't expressions. And some things that syntactically look like expressions aren't because pattern syntax overlaps expression syntax.)

src/content/language/collections.md
```diff
@@ -242,78 +250,180 @@ final constantMap = const {2: 'helium', 10: 'neon', 18: 'argon'};
...
+var a = 'one';
+var b = '';
+var items = [
+  0,
+  if (a case 'one') 1,
+  if (b case 'two') 2,
+  3,
+  4,
+]; // [0, 1, 3, 4]
+```
```
munificent, line 384: It might be worth showing an example that you couldn't do just using `if (a == 'one')`. Perhaps:

```dart
var string = 'hello';
var coordinates = [
  1,
  if (string case String(length: var length)) length,
  3,
]; // 1, 5, 3
```

That's a pretty weird, contrived example, though. It's hard to come up with a good but simple example. :-/

src/content/language/collections.md
```diff
@@ -242,78 +250,180 @@ final constantMap = const {2: 'helium', 10: 'neon', 18: 'argon'};
...
+### For loops
+
+You can use a `for` loop inside of a collection to
+programmatically generate multiple values and insert them
+into the collection in a concise and readable way.
+A `for` loop has this signature in a collection:
+
+```dart
+for (<expression> in <collection>) <result>
+```
```
munificent, line 400: There are two kinds of for loops in Dart and we support both of them as elements:

```dart
var fruits = ['apple', 'banana', 'cherry'];
var list = [
  // for-in loop:
  for (var fruit in fruits) fruit.length,
  // Old C-style for loop:
  for (var x = 1; x < 4; x++) x,
]; // [5, 6, 5, 1, 2, 3].
```

src/content/language/collections.md
```diff
@@ -242,78 +250,180 @@ final constantMap = const {2: 'helium', 10: 'neon', 18: 'argon'};
...
+In the following example, the square for each value in
+`numbers` is included in `items`.
+
+<?code-excerpt "misc/lib/language_tour/misc/for_loop_in_collection_a.dart (code_sample)"?>
+```dart
+var numbers = [2, 3, 4];
+var items = [1, for (var n in numbers) n * n, 7]; // [1, 4, 9, 16, 7]
+```
+
+You can include `if` statements in `for` loops.
```
munificent, line 411: You can nest all kinds of elements arbitrarily deeply:

```dart
var ys = [4, 5, 6];
var list = [
  if (condition) ...[
    for (var x = 0; x < 3; x++)
      for (var y in ys)
        if (x < y)
          x + y
  ]
];
```

(In this way, Dart's control flow elements are much more powerful than [list comprehensions](https://en.wikipedia.org/wiki/List_comprehension) are in other languages.)

In particular, it's common and idiomatic to use a spread on a collection literal immediately inside an `if` or `for`. In a statement context, if you want to execute a bunch of statements based on a single if condition, you use a block:

```dart
if (condition) doOneThing();

if (condition) {
  doMultiple();
  things();
}
```

You can't use blocks inside collection literals (because blocks are statements, not elements). The analogous construct is spreading a literal:

```dart
var list = [
  if (condition) oneThing(),

  if (condition) ...[
    multiple(),
    things(),
  ],
];
```

There's no special support in the language for "spread blocks" like this. It just naturally falls out from the fact that elements can contain other elements.

This pattern is common enough that it might be worth documenting here.

munificent (approved):
A bunch of feedback here and there but overall I really like the improvement.

examples/misc/lib/language_tour/misc/null_aware_spread_operator_in_collection_a.dart
antfitch: Updated! One note: I removed this code sample since it was producing analysis errors. @parlough if there is some way to suppress those errors on specific files, let me know and I'll add them back.

examples/misc/lib/language_tour/misc/null_aware_spread_operator_in_collection_a.dart
antfitch: Updated! I used [1, null, 3]

src/content/language/collections.md
antfitch: I was wondering why we didn't use the term 'element' in this guide. I've updated this for `List` and `Set`, but I've kept key-value pair for `Map`. I think we should follow the spec on this. Element is a very common term. To me, it seems like an anti-pattern if we don't use it. Take a close look at my changes.

src/content/language/collections.md
antfitch: Yes, my intent is to contrast with null-aware spread. I'll add a link to it to make this more clear. If more is needed, let me know.

src/content/language/collections.md
antfitch: Added! I included some sample data for grounding. I also included an excerpt for one of them that can also quickly be copied and pasted into DartPad to produce results.

I wanted to leave the super simple example as a baseline, so that's still there for now. If you really want to be rid of it, let me know.

src/content/language/collections.md
antfitch: This is fascinating. I think we should definitely call them if elements. I'm guessing we call these if key-value pairs for maps? I'll change descriptions and titles so that we can test and see what we think.

src/content/language/collections.md
antfitch: I just realized something else. It's very good that you pointed this out because even Gemini didn't mention that the `if` element could include an `else` statement. I'll get these docs updated so that AI model answers are more accurate.

src/content/language/collections.md
antfitch: I've added a few examples. They aren't as fun as this one, though.

src/content/language/collections.md
antfitch: This is pattern? Nice! I had no idea. I'll update and add a note about that.

src/content/language/collections.md
antfitch: Excellent! Adding these signatures and examples.

src/content/language/collections.md
antfitch: I've created a new section called `Nesting elements` to capture all of this.

antfitch:
@parlough I've made some large changes after the first review. Here's an overview:

* We're now using the term 'element' to describe items in collections.
* From the discussion above, it seems like all of these supplementals are actually different types of elements that you can add to a collection. I've updated headings to reflect this.
* I've added a comment above each code sample that does not have an excerpt with a reason. This is so we can keep track of samples that don't have them on purpose.
* Added the various syntax structures that can be used for each element.

src/content/language/collections.md
antfitch: And after a little research, it seems that a map has elements and an element in a map has a key-value pair in it. I think I'm in alignment now.

src/content/language/collections.md
antfitch: Added!

src/content/language/collections.md
antfitch: I decided to leave this one out because it calls out one collection type (I'm trying to avoid that with the generic syntax). If you feel strongly, I could create tabs that show each collection type and what this syntax looks like. 
@munificent or @parlough let me know your thoughts.

examples/misc/lib/language_tour/misc/for_loop_in_collection_a.dart
```diff
@@ -0,0 +1,8 @@
+void main() {
+  // #docregion code_sample
+  var numbers = [2, 3, 4];
+  var items = [1, for (var n in numbers) n * n, 7]; // [1, 4, 9, 16, 7]
```
parlough, line 4: Since you want to include the output, you could include the `print` calls in you snippets as well and put the comments before that.

Then developers could quickly copy and paste the whole snippets to try themselves as well.

Another option is to structure these as tests and use the `equals` matchers. Then we could also run these tests in CI to know the output is correct.

```dart
var numbers = [2, 3, 4];
var items = [1, for (var n in numbers) n * n, 7];

expect(items, equals([1, 4, 9, 16, 7]));
```

src/content/language/collections.md
```diff
@@ -242,78 +252,355 @@ final constantMap = const {2: 'helium', 10: 'neon', 18: 'argon'};
 For more information about maps, refer to the Maps section of the
 [`dart:core` documentation](/libraries/dart-core#maps).
 
-## Operators
+[Using constructors]: /language/classes#using-constructors
+[`Map`]: {{site.dart-api}}/dart-core/Map-class.html
+[type inference]: /language/type-system#type-inference
+
+## Supplementals
```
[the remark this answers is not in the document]
parlough: Then we could also add a basic section about the leaf elements (expression elements and map entry elements).

src/content/language/collections.md
```diff
@@ -242,78 +252,355 @@ final constantMap = const {2: 'helium', 10: 'neon', 18: 'argon'};
...
+
+<?code-excerpt "misc/lib/language_tour/misc/nesting_in_collection_c.dart (code_sample)"?>
+```dart
+var items = [
+  if (condition) oneThing(),
+  if (condition) ...[multiple(), things()],
+]; // [oneThing, [multiple_a, multiple_b], things]
+```
+
+You can nest all kinds of elements arbitrarily deep.
```
parlough, line 591: Is it worth clarifying besides the leaf elements (expressions and map entries)?

parlough (commented):
Thanks for the extensive work on this @antfitch and for your patience on me reviewing it.

I love these additions, especially the extra examples and showing off more features (`else`, `if-case`, nesting, etc).

Overall these additions are a huge improvement. I just have one main concern about the conflation of "element" and some other minor suggestions to consider. Feel free to push back on anything and I'm happy to talk over any of my comments offline as well.

Thanks so much! :D

src/content/language/collections.md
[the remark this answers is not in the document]
antfitch: Just laying the groundwork for our newest element. That PR will come next!

src/content/language/collections.md
```diff
@@ -242,78 +252,355 @@ final constantMap = const {2: 'helium', 10: 'neon', 18: 'argon'};
...
+<?code-excerpt "misc/lib/language_tour/misc/spread_operator_in_collection_a.dart (code_sample)"?>
 ```dart
-var list2 = [0, ...?list];
-assert(list2.length == 1);
+var a = [1, 2, null, 4];
+var items = [0, ...a, 5]; // [0, 1, 2, null, 4, 5]
 ```
 
-For more details and examples of using the spread operator, see the
-[spread operator proposal.][spread proposal]
```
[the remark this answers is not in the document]
antfitch: :D

site-shared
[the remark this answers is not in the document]
antfitch: Updated!

src/content/language/collections.md
```diff
@@ -45,10 +47,10 @@ but it can help prevent copy-paste errors.
 var list = ['Car', 'Boat', 'Plane'];
 ```
 
-Lists use zero-based indexing, where 0 is the index of the first value
-and `list.length - 1` is the index of the last value. 
+Lists use zero-based indexing, where 0 is the index of the first element
+and `list.length - 1` is the index of the last element.
```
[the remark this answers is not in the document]
parlough: I know it's what developers might be used to seeing, but I'm not sure that familiarity is worth the confusion from one term being used to mean both syntactical elements in a collection literal and actual values in a list.

When reading the page with "element" being used for both, it is quite confusing for me and I'm familiar with the distinction.

See https://github.com/dart-lang/site-www/issues/2631#issue-698613262 for some other background.

Maybe we could have a call about it?

src/content/language/collections.md
```diff
@@ -242,78 +252,355 @@ final constantMap = const {2: 'helium', 10: 'neon', 18: 'argon'};
...
+spread element and will produce an error. 
+
+A spread element has this syntax in a collection:
 
-<?code-excerpt "misc/test/language_tour/built_in_types_test.dart (list-spread)"?>
 ```dart
-var list = [1, 2, 3];
-var list2 = [0, ...list];
-assert(list2.length == 4);
+...<collection_name>
```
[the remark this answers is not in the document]
parlough: Sounds good, thanks for updating this!

Since a collection literal evaluates to some collection, I think it's accurate to consider a literal an expression as well :)

src/content/language/collections.md
```diff
@@ -242,78 +252,347 @@ final constantMap = const {2: 'helium', 10: 'neon', 18: 'argon'};
...
+```dart
+// If the expression matches the pattern, include the result.
+if (<expression> case <pattern>) <result>
+```
+
+```dart
+// If the operation such as '(<bool_expression>)'
+// or '(<expression> case <pattern>)' resolves as true,
+// include the first result, otherwise include the second
+// result.
```
parlough, line 405: Consider adjusting the break here to better collocate the "otherwise" clause.

```suggestion
// include the first result, otherwise,
// include the second result.
```

src/content/language/collections.md
```diff
@@ -242,78 +252,347 @@ final constantMap = const {2: 'helium', 10: 'neon', 18: 'argon'};
...
-## Operators
+[Using constructors]: /language/classes#using-constructors
+[`Map`]: {{site.dart-api}}/dart-core/Map-class.html
+[type inference]: /language/type-system#type-inference
+
+## Collection elements {: #collection-elements }
+
+In a collection, an element is an individual item in that
+collection. These types of elements are supported in
+a collection:
```
parlough, line 263: This is where the confusion of "element" that I mentioned particularly shows up.

The following h3 subsections are talking about elements in a collection literal, but this intro conflates items in actual collections and the elements in a literal. This mostly affects the usage of "individual item" as the collection elements can result in multiple items.

```suggestion
Collection elements are the syntactical constructs
used within collection literals to define the
resulting collection's initial contents.

These types of elements are supported in a collection:
```

src/content/language/collections.md
```diff
@@ -242,78 +252,347 @@ final constantMap = const {2: 'helium', 10: 'neon', 18: 'argon'};
...
+
+*   Built-in data types
+*   Functions and methods
```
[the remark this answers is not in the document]
parlough: Perhaps for follow-up, we could also add a short subsection that introduces "Expression elements" that explicitly introduces the leaf elements.

examples/misc/test/language_tour/collections/if_case_operator_in_collection_b.dart
```diff
@@ -0,0 +1,21 @@
+import 'package:test/test.dart';
+
+void main() {
+  // #docregion code_sample
+  var orderDetails = ['Apples', 12, ''];
```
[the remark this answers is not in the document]
antfitch: ^_^

site-shared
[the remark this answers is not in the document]
antfitch: Totally! Keeping this one open in case I run into the problem again.

src/content/language/collections.md
antfitch: I have an alternative for us to take a look at tomorrow.

src/content/language/collections.md
```diff
@@ -242,78 +252,347 @@ final constantMap = const {2: 'helium', 10: 'neon', 18: 'argon'};
 For more information about maps, refer to the Maps section of the
 [`dart:core` documentation](/libraries/dart-core#maps).
 
-## Operators
+[Using constructors]: /language/classes#using-constructors
+[`Map`]: {{site.dart-api}}/dart-core/Map-class.html
+[type inference]: /language/type-system#type-inference
+
+## Collection elements {: #collection-elements }
```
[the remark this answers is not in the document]
antfitch: Saving these for now, but I'll make a note to check on these before I submit this PR.

src/content/language/collections.md
```diff
@@ -242,78 +252,347 @@ final constantMap = const {2: 'helium', 10: 'neon', 18: 'argon'};
...
+```dart tag=fails-sa
+List<String> buildCommandLine(
+  String executable,
+  List<String> options, [
+  List<String>? extraOptions,
+]) {
+  return [
+    executable,
+    ...options,
+    ...extraOptions, // <-- OK now.
```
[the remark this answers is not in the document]
antfitch: Yup! Thanks for catching this. Should be fixed now.

johnpryan (approved):

src/content/language/collections.md
antfitch: Tentatively closing this because that section has changed so much. I'll have an addition revision today that we should look over when its complete. I'll let you know when it's ready.

examples/misc/test/language_tour/collections/for_loop_in_collection_a.dart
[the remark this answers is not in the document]
antfitch: Absolutely. Thank you for that great idea.

munificent (approved):
A bunch of little minor suggestions, but this looks great!

src/content/language/collections.md
```diff
@@ -13,16 +13,18 @@ Dart has built-in support for list, set, and map [collections][].
...
 is the *array*, or ordered group of objects. In Dart, arrays are
 [`List`][] objects, so most people just call them *lists*.
 
-Dart list literals are denoted by
-a comma separated list of expressions or values,
-enclosed in square brackets (`[]`).
-Here's a simple Dart list:
+Dart list literals are denoted by a comma-separated list of
+elements enclosed in square brackets (`[]`). Each element
+contains an expression or value. Here's a simple Dart list:
```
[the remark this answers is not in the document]
antfitch: Done!

src/content/language/collections.md
```diff
@@ -146,12 +151,17 @@ final constantSet = const {
...
 
+[`Set`]: {{site.dart-api}}/dart-core/Set-class.html
+
 ## Maps
 
-In general, a map is an object that associates keys and values. Both
-keys and values can be any type of object. Each *key* occurs only once,
-but you can use the same *value* multiple times. Dart support for maps
-is provided by map literals and the [`Map`][] type.
+In general, a map is an object that contains elements in the
```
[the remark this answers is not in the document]
antfitch: Done!

src/content/language/collections.md
```diff
@@ -242,78 +252,406 @@ final constantMap = const {2: 'helium', 10: 'neon', 18: 'argon'};
...
+        collection. 
+
+*   Control flow element: Conditionally or iteratively
+    constructs collections directly within a
+    collection literal.
 
-### Spread operators
+    *   Spread element: Iterates over a given sequence
+        (collection expression) and inserts all of the
+        resulting values into another collection.
```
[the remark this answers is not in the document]
antfitch: Fixed here & elsewhere.

src/content/language/collections.md
```diff
@@ -242,78 +252,406 @@ final constantMap = const {2: 'helium', 10: 'neon', 18: 'argon'};
...
+To learn more about collection elements, see the following
+sections.
+
+<a id="control-flow-operators" aria-hidden="true"></a>
+<a id="spread-operators" aria-hidden="true"></a>
+<a id="collection-operators" aria-hidden="true"></a>
+
+### Expression elements {: #expression-element }
+
+An expression element evaluates a single general expression
```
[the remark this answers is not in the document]
antfitch: Updated here and elsewhere.

src/content/language/collections.md
```diff
@@ -242,78 +252,406 @@ final constantMap = const {2: 'helium', 10: 'neon', 18: 'argon'};
...
-var list2 = [0, ...?list];
-assert(list2.length == 1);
+var a = [1, 2, null, 4];
+var items = [0, ...a, 5]; // [0, 1, 2, null, 4, 5]
 ```
 
-For more details and examples of using the spread operator, see the
-[spread operator proposal.][spread proposal]
+If you would like the spread element to ignore a null
+collection that is passed into and not produce an
```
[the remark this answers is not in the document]
antfitch: Updated!

src/content/language/collections.md
```diff
@@ -242,78 +252,406 @@ final constantMap = const {2: 'helium', 10: 'neon', 18: 'argon'};
...
+    ...options,
+    ...?extraOptions, // <-- OK now.
+  ];
+}
+
+// Usage:
+//   buildCommandLine('dart', ['run', 'my_script.dart'], null);
+// Result:
+//   [dart, run, my_script.dart]
 ```
```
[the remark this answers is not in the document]
antfitch: ^_^

src/content/language/collections.md
```diff
@@ -242,78 +252,406 @@ final constantMap = const {2: 'helium', 10: 'neon', 18: 'argon'};
...
+// include the second result.
+if (<if_operation>) <result> else <result>
+```
+
+```dart
+// If the first operation such as '(<bool_expression>)'
+// or '(<expression> case <pattern>)' resolves as true,
+// include the first result, otherwise,
+// perform the second operation.
+if (<if_operation>) <result> else if (<if_operation>)
```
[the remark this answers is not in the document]
antfitch: I'll delete this example. Essentially the previous one should cover this. I've swapped this out with `if (<expression> case <pattern>) <result> else <result>`.

src/content/language/collections.md
```diff
@@ -242,78 +252,406 @@ final constantMap = const {2: 'helium', 10: 'neon', 18: 'argon'};
...
+
+```dart
+// If the bool expression is true, include the result.
+if (<bool_expression>) <result>
+```
+
+```dart
+// If the expression matches the pattern, include the result.
+if (<expression> case <pattern>) <result>
+```
```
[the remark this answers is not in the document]
antfitch: I updated this section to be more clear. No need for context switching anymore.

src/content/language/collections.md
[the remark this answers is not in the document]
antfitch: Updated~

src/content/language/collections.md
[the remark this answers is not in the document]
antfitch: Decided to leave. Essentially, following guidance by the editorial team (they want us to almost always use slugs since it's less likely when titles change that something will break).

src/content/language/collections.md
```diff
@@ -242,78 +252,406 @@ final constantMap = const {2: 'helium', 10: 'neon', 18: 'argon'};
...
+
+An expression element has this syntax in a collection:
 
-<?code-excerpt "misc/test/language_tour/built_in_types_test.dart (list-spread)"?>
 ```dart
-var list = [1, 2, 3];
-var list2 = [0, ...list];
-assert(list2.length == 4);
+<general_expression>
+```
```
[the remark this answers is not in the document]
antfitch: I know this is very broad, but I left this here to be consistent with the other sections. I know it's very very simple, but I want to make sure people know that we didn't accidentally leave this out here.

src/content/language/collections.md
```diff
@@ -242,78 +252,406 @@ final constantMap = const {2: 'helium', 10: 'neon', 18: 'argon'};
...
+
+A map entry element has this syntax in a collection:
+
+```dart
+<key_expression>: <value_expression>
+```
+
+### Spread elements {: #spread-element }
+
+The spread element (`...`) iterates over a given sequence
```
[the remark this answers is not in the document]
antfitch: Updated here and also for the null-aware spread element.

src/content/language/collections.md
antfitch: Tentatively leaving as is & closing. If anyone has inspiration for change, re-open and add your thoughts.

src/content/language/collections.md
antfitch: Tentatively closing due to the age of the ask. Feel free to modify this section in a followup PR if you feel like it.

antfitch:
Everyone, I think we are in good enough shape to push this. If anyone has followup changes, go ahead and open a PR we'll get any additional changes added. Whew, this was a big one. Thank you so, so, so much for your help! ^_^

Issue #5614, opened 2024-03-02
Fix named parameters in 'Dart cheatsheet codelab' page


dtonhofer:
### Page URL

https://dart.dev/codelabs/dart-cheatsheet/

### Page source

https://github.com/dart-lang/site-www/tree/main/./src/content/codelabs/dart-cheatsheet.md

### Describe the problem

Dart cheatsheet codelab, "Code Example" for [Named Parameters](https://dart.dev/codelabs/dart-cheatsheet#code-example-9):

The test code does not verify that the no-argument call to `copyWith()` indeed takes over the present values of "this", instead of just setting them to the values used in the default constructor ( I made that error in the code).

### Expected fix

Extend the testing code as follows (here also factoring out those nasty strings):

```
class MyDataObject {
  final int anInt;
  final String aString;
  final double aDouble;

  MyDataObject({
     this.anInt = 1,
     this.aString = 'Old!',
     this.aDouble = 2.0,
  });

  // solved code problem
  MyDataObject copyWith({int? newInt, String? newString, double? newDouble}) {
     var myInt = newInt ?? this.anInt;
     var myString = newString ?? this.aString;
     var myDouble = newDouble ?? this.aDouble;
     return MyDataObject(anInt: myInt, aString: myString, aDouble: myDouble);
  }
}


// Tests your solution (Don't edit!):
void main() {
  final source = MyDataObject();
  final errs = <String>[];
  final insteadTxt = ' rather than the expected value ';

  final callTxt1 = 'copyWith(newInt: 12, newString: \'New!\', newDouble: 3.0)';
  try {
    final copy = source.copyWith(newInt: 12, newString: 'New!', newDouble: 3.0);    
    final callTxtExt = 'Called ' + callTxt1 + ', \n and the new object\'s '; 
    if (copy.anInt != 12) {
      errs.add(callTxtExt + 'anInt was ${copy.anInt}' + insteadTxt + '(12).');
    }
    
    if (copy.aString != 'New!') {
      errs.add(callTxtExt + 'aString was ${copy.aString}' + insteadTxt + '(\'New!\').');
    }
    
    if (copy.aDouble != 3) {
      errs.add(callTxtExt + 'aDouble was ${copy.aDouble}' + insteadTxt + '(3).');
    }
  } catch (e) {
    print('Called ' + callTxt1 + ' \n and got an exception: ${e.runtimeType}');
  }
  
  final callTxt2 = 'copyWith()';
  try {
[...]

PR #6532, opened 2025-04-18
Expand tests for named argument entry in cheatsheet

parlough:
Adds specific tests for the default values from the constructor, in case the developer accidentally uses them as the default value in `copyWith`.

Fixes https://github.com/dart-lang/site-www/issues/5614

### DIFF
--- a/src/content/resources/dart-cheatsheet.md
+++ b/src/content/resources/dart-cheatsheet.md
@@ -1101,43 +1101,66 @@ class MyDataObject {
 void main() {
   final source = MyDataObject();
   final errs = <String>[];
-  
+
   try {
     final copy = source.copyWith(newInt: 12, newString: 'New!', newDouble: 3.0);
-    
+
     if (copy.anInt != 12) {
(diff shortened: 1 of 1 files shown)


antfitch:
This change looks nice, although something large jumped out at me on this page. The code examples don't seem to be code examples. They do seem to be coding challenges that you can interact with to learn how to use certain features.

I think we should consider naming the sections we are calling `Code example` to something else. Here are a few ides:

* Code challenge
* Interactive coding challenge
* Coding exercise
* Coding problem

 It's up to you if you want to fix this here. If you don't, I think we need to do this in a future PR.

antfitch (approved):

parlough:
> The code examples don't seem to be code examples. They do seem to be coding challenges that you can interact with to learn how to use certain features.
> 
> I think we should consider naming the sections we are calling Code example to something else.

I think that makes a lot of sense and updating the names sound good to me. I went ahead and did so in https://github.com/dart-lang/site-www/pull/6534, let me know how the change their sounds.

Do note that we expect the GSE work to replace this doc as it is already out of date and not a great experience due to us dropping some supporting DartPad features.

PR #6547, opened 2025-04-29
6545-fix-reference-wording

### LINKED ISSUE #6545: Incorrect wording: Should say "change reference", not "change value" of a non-final, non-const variable
### Page URL

https://dart.dev/language/variables/

### Page source

https://github.com/dart-lang/site-www/tree/main/src/content/language/variables.md

### Describe the problem

In the Dart documentation, the explanation regarding const initialization is misleading:

"You can change the value of a non-final, non-const variable, even if it used to have a const value:"

Example:
var foo = const [];
foo = [1, 2, 3];  // Was const []
The statement says "Change the value," but this is inaccurate because:

The const value (an empty list const []) is immutable and cannot be changed.

What’s happening here is a change of the reference — foo is reassigned to a new mutable list [1, 2, 3].

### Expected fix

Suggested rewording:

"You can change the reference of a non-final, non-const variable, even if it was initially assigned a const value."

This better reflects the behavior without implying that the original const object was modified.

Thank you for considering this correction!

### Additional context

_No response_

DunyaYousufzai:
Thanks for your contribution! Please replace this text with a description of what this PR is changing or adding and why, list any relevant issues, and review the contribution guidelines below.

Fixes #6545.

This updates a misleading sentence in the variables documentation that incorrectly implies the const value is being changed. The updated wording clarifies that the variable reference is being changed instead.

---

- [ ] I’ve reviewed the contributor guide and applied the relevant portions to this PR.
- [ ] This PR doesn't contain automatically generated corrections or text (Grammarly, LLMs, and similar).
- [ ] This PR follows the [Google Developer Documentation Style Guidelines](https://developers.google.com/style) — for example, it doesn't use _i.e._ or _e.g._, and it avoids _I_ and _we_ (first person).
- [ ] This PR uses [semantic line breaks](https://github.com/dart-lang/site-shared/blob/main/doc/writing-for-dart-and-flutter-websites.md#semantic-line-breaks) of 80 characters or fewer.

<details>

</details>


antfitch:
Nice catch. Make sure to sign the CLA so that we can make this change.

antfitch (approved):

Issue #5404, opened 2023-12-13
Add functionality explanation to the `timeout(...)` method on Streams page


SandPod:
### Page URL

https://dart.dev/libraries/async/using-streams

### Describe the problem

The tutorial for asynchronous streams is missing documentation about the `timeout(...)` method available on Stream objects. It is listed together with `handleError(...)` and `transform(....)` but misses a section describing its functionality.

### Expected fix

A matching description about the `timeout(...)` method describing what it is for and how it is used.

### I would like to fix this problem.

- [ ] I will try and fix this problem on dart.dev.

huycozy:
Hi @SandPod 
Could you check https://api.flutter.dev/flutter/dart-async/Stream/timeout.html to see if it has information that you are looking for?

SandPod:
@huycozy Yes, that documentation was good enough for my use case. Thank you 👍

The information should still probably either be summarized or linked from tutorial page.

PR #6566, opened 2025-05-04
Add explanation for timeout in using-streams tutorial

parlough:
Resolves https://github.com/dart-lang/site-www/issues/5404

### DIFF
--- a/src/content/libraries/async/using-streams.md
+++ b/src/content/libraries/async/using-streams.md
@@ -312,14 +312,46 @@ There is no recovering from that.
 The following code shows how to use `handleError()` to remove errors
 from a stream before using it in an **await for** loop.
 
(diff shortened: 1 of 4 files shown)


src/content/libraries/async/using-streams.md
```diff
@@ -304,27 +304,62 @@ Stream<T> timeout(
...
 ```
 
+In the previous example, an **await for** loop is never
+returned to if no events are emitted by the stream.
+To avoid this, use the `timeout()` function to create
+a new stream. `timeout()` enables you to set a
+time limit and continue emitting events on the returned
+stream.
+
+The following code modifies the previous example.
```
antfitch, line 338: The entire section had a lot of pre-existing wording issues. I've gone through and cleaned things up. Double-check this last part.

Issue #5206, opened 2023-09-19
Clarify ambiguity on 'Sound null safety' page


PranjalPatra:
### Page URL

https://dart.dev/null-safety/

### Page source

https://github.com/dart-lang/site-www/tree/main/src/null-safety/index.md

### Describe the problem

I believe there may be an ambiguity in the statement 'If the type system determines that something isn’t null, then that thing can never be null.' A more accurate phrasing could be, 'If the type system determines that something isn’t **nullable**, then that thing can never be null.'

The reason for this suggestion is that a variable declared as nullable (e.g., `int? a = 10`) can have a value at one point in time and still be set to null later. The current wording might lead to confusion about the variable's ability to hold a null value in the future.

### Expected fix

update the documentation if I am correct

### Additional context

I am just learning

parlough:
Thanks for opening an issue! You're right the text could potentially be construed that way as it stands.

I think the confusion at least partially comes from the unspecific "something" and "that thing" terms. I believe this sentence is not aiming to discuss nullability of variables, but rather intending to discuss the references (to values).

I'm not yet sure how to best clarify that without making the explanation too long, but I'll give it some thought and am happy to hear some more ideas.

Let me know what you think. Thanks again!

PR #6570, opened 2025-05-04
Replace ambigious terminology in 'Fully sound' explanation

parlough:
- Adjust the "fully sound" definition to avoid the potentially confusing "something" and "that thing" terminology
- Switch to a definition list for improved formatting
- Move the blurbs about performance to after the definition list

Fixes https://github.com/dart-lang/site-www/issues/5206


antfitch (approved):

antfitch:
Love this change!

Issue #5252, opened 2023-10-16
Propose new name for "Null assertion operator"


MaryaBelanger:
## Page

https://dart.dev/language/operators#other-operators

## Background 

@dart-lang/language-team @bwilkerson  I would like to settle on a term for "Null assertion operator"/`!` and then update all of our resources (docs, lints, diagnostics, in the SDK, etc) to use that term **exclusively** as part of the effort to make Dart's terminology more consistent (#5247).

There's been a lot of discussion going on (https://github.com/dart-lang/sdk/issues/58535 , https://github.com/dart-lang/sdk/issues/47185) but it doesn't feel conclusive enough yet to make a decision. Summarizing the major points:

- `null assertion operator` (currently standard on dart.dev)
  -  Inaccurate: implies the desired outcome is null, when in reality it does the opposite.
- `non-null assertion operator` 
  - Better, but still possibly misleading: "assert", while its literal definition is appropriate, is already dedicated in Dart and therefore inaccurate. 
- `null check operator` / `non-null check operator` 
  - Agreeable and ubiquitous, but also vague and inaccurate: "null check" is too general and currently used in the docs to encompass a lot of different ways to check null (`... != null`, `as`, etc.).
- Other options like `non-null enforcing`, `non-null cast`, `null failure`, `throw if null`, `unwrap` (Swift), etc are floating around too.

(@parlough's [longer summary of options](https://github.com/dart-lang/sdk/issues/58535))

## Proposal 

Considering the pros/cons people seem to have with the ideas discussed so far, I'm wondering...

### What about "Null _Exception_ Operator"?

Purely from an effective terminology perspective, to me `null exception operator` tells users exactly what they need to know by focusing on the end result that actually affects them (the exception thrown), while bypassing the contention around _how_ exactly it's getting there (cast vs. assertion vs. check, etc.). It's clearly saying "this operator causes an exception on null".

- `null failure` is similar, technically describing the outcome, but "failure" is vague; if we're choosing this we might as well be more explicit and just say what the failure is: an "exception". 
- You could argue `throw-if-null operator` is even more explicit, but I feel it's too far from the original terminology to be a contender and is kind of a mouth-full. 
- `non-null assertion` and `non-null check` are tied for second place for me, as they're the best combination of familiar/recognizable/accurate, but they're not perfect and there's clearly issue taken with both of them.

## Goals

- Definitely should move away from `null assertion operator`
- Can we settle one of these? 
[...]

PR #6571, opened 2025-05-04
Standardize docs on 'not-null assertion operator'

parlough:
In https://github.com/dart-lang/site-www/issues/5252, no complete conclusion was came to. However, the site is currently inconsistent in how it refers to the operator which is even more harmful. I'm going to lean on what I've seen in the community and familiarity with otherwise commonly known languages:

- [TypeScript's non-null assertion operator](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#non-null-assertion-operator-postfix-)
- [Kotlin's not-null assertion operator](https://kotlinlang.org/docs/null-safety.html#not-null-assertion-operator)

Resolves https://github.com/dart-lang/site-www/issues/5252

### DIFF
--- a/src/content/effective-dart/usage.md
+++ b/src/content/effective-dart/usage.md
@@ -436,7 +436,7 @@ class UploadException {
 [final]: /effective-dart/design#prefer-making-fields-and-top-level-variables-final
 [null-check pattern]: /language/pattern-types#null-check
 [`final`]: /effective-dart/usage#do-follow-a-consistent-rule-for-var-and-final-on-local-variables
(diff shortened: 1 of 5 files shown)


antfitch (approved):
Tentatively approving this. I've most often seen non-null in language documentation, but to keep everything consistent, will approve this. This looks like something we need to add to the glossary and to our writing style guide.

antfitch:
I've added a word list to the wiki. This is not a glossary. Instead, it's a style guide that we can use to include what to use, what not to use, and include links to why decisions were made. https://github.com/dart-lang/site-www/wiki/Word-list

PR #6585, opened 2025-05-09
Update usage.md

apulsifer:
Fixed spelling of "too" where it is used to mean "also".


apulsifer:
Realized this proposed edit was incorrect, so it is closed by the author

PR #6588, opened 2025-05-13
Update collections.md

rahulrao0209:
Fix bug in for element example.


antfitch (changes requested):
This content is generated. You'll need to update the source file. To learn how to do this, go here: https://github.com/dart-lang/site-www/wiki/Documentation-guide#code-samples

It looks like you'll need to find this file and update it: 

```
misc/test/language_tour/collections/for_loop_in_collection_c.dart
```

And then when you are done, you'll need to run this script to update the code example in the docs: 

```
dart run excerpter misc/test/language_tour/collections/for_loop_in_collection_c.dart
```

PR #6591, opened 2025-05-13
Include Flutter CLI instructions for Dart experiment flags

loic-sharma:
Explain how you can turn on a Dart experiment when running a Flutter app.

Here's prior art where we explain how a Dart feature can be used on Flutter's tooling: https://dart.dev/libraries/core/environment-declarations#flutter

---

<details>

</details>

### DIFF
--- a/src/content/tools/experiment-flags.md
+++ b/src/content/tools/experiment-flags.md
@@ -20,17 +20,28 @@ pass the corresponding flag to the tool.
 
 For example, to enable the experiments
 `super-mixins` and `no-slow-checks`,
-add those flags to the `dart` command:
+add those flags to the
+`dart` or `flutter` command:
+
+{% tabs "experiment-flags-tabs" %}
+
+{% tab "Dart command" %}
 
 ```console
 $ dart run --enable-experiment=super-mixins,no-slow-checks bin/main.dart
 ```
 
-Or for the `flutter` command:
+{% endtab %}
+
+{% tab "Flutter command" %}
 
 ```console
 $ flutter run --enable-experiment=super-mixins,no-slow-checks
 ```
+
+{% endtab %}
+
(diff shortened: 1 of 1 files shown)


antfitch:
@loic-sharma I was about to submit this, but then I realized that this change would introduce Flutter commands into the Dart documentation. I think we need to capture this, but we should do this in the Flutter documentation.

I believe we need to add this change here:
https://docs.flutter.dev/reference/flutter-cli

The problem is that we don't have a lot of the flutter CLI mapped out. Would you have any interest in working on a project to map out the Flutter CLI?

loic-sharma:
@antfitch Dart does have a few docs that also point to their `flutter` command counterparts. For example:

1. https://dart.dev/libraries/core/environment-declarations#flutter
2. https://dart.dev/tools/analysis
3. https://dart.dev/tools/testing#flutter-testing

I think a breadcrumb here in Dart's docs would be useful. Most folks that want to enable a Dart experiment will want to do so for their Flutter app - that's the way most folks interface with Dart.

When we have a Flutter CLI reference, we could update this section to link to Flutter's docs. What do you think?

> Would you have any interest in working on a project to map out the Flutter CLI? I do have an agenda item for myself to do this, but in 2025 Q3.

Yup, I'd be happy to help! :)

antfitch:
> 1. https://dart.dev/libraries/core/environment-declarations#flutter

@loic-sharma thank you for this context! Okay, this is what I'm going to do: (a) merge this change of yours, (b) update the Flutter CLI in Q3 and then recalibrate the Flutter and Dart CLI command docs as needed.

antfitch:
@loic-sharma I was going to add tabs, but it looks like we don't have markdown support for these on the Dart side yet. Using your original structure.

PR #6604, opened 2025-05-19
Add version note and redirect for cross compilation

parlough:
- Add version note to cross compilation support
- Add `cross-compilation` anchor to use for new `/to/cross-compilation` tooling redirect
- Fix case of double "the"
- Minor formatting changes

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -373,6 +373,7 @@
       { "source": "/support/faq", "destination": "/resources/faq", "type": 301 },
       { "source": "/support{,/**}", "destination": "/community", "type": 301 },
 
+      { "source": "/to/cross-compilation", "destination": "/tools/dart-compile#cross-compilation", "type": 301 },
       { "source": "/to/doc-comment-references", "destination": "/tools/doc-comments/references", "type": 301 },
       { "source": "/to/downgrade-testing", "destination": "/tools/pub/dependencies#test-with-downgraded-dependencies", "type": 301 },
       { "source": "/to/enforce-lockfile", "destination": "/tools/pub/packages#get-dependencies-for-production", "type": 301 },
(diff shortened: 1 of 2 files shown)


antfitch (approved):
LGTM, but I'm find the terms and definitions difficult to see when they don't have bullets. We can chat about that at a later date. I'm assuming that the version note will be removed once 3.8 is released? If not, we should also chat about that.

parlough:
>  I'm assuming that the version note will be removed once 3.8 is released? If not, we should also chat about that.

Generally no. Dart and Flutter developers often stay on a previous version for a while. If we don't have version notes, they'll be confused when features don't work as the CLI will provide no useful information to them about why.

For [language versioned](https://dart.dev/resources/language/evolution#language-versioning) changes, we keep related version notes until the stable SDK no longer supports language versions without them. For example, Dart 3.0 dropped support for language versions before 2.12.

For SDK changes or features, such as this, it's less clear cut. At least for 4 releases (a year), but it depends on the feature and the confusion that could result.

parlough:
> LGTM, but I find the terms and definitions difficult to see when they don't have bullets

That's fair. We can definitely explore alternative/improved styles for definition lists! Let me know if you have any ideas :)

antfitch:
> > I'm assuming that the version note will be removed once 3.8 is released? If not, we should also chat about that.
> 
> Generally no. Dart and Flutter developers often stay on a previous version for a while. If we don't have version notes, they'll be confused when features don't work as the CLI will provide no useful information to them about why.
> 
> For [language versioned](https://dart.dev/resources/language/evolution#language-versioning) changes, we keep related version notes until the stable SDK no longer supports language versions without them. For example, Dart 3.0 dropped support for language versions before 2.12.
> 
> For SDK changes or features, such as this, it's less clear cut. At least for 4 releases (a year), but it depends on the feature and the confusion that could result.

We should definitely chat about this and get a process written down. I'll make a note to circle back around to this.

Issue #6461, opened 2025-02-26
☔ Create glossary entries for important Dart terms


parlough:
### Overview

We now have expanded support for a shared [glossary](https://dart.dev/resources/glossary) on the Dart site. Currently definitions live only on the page, but eventually their short definitions will be shown on hover across the entire site. This helps developers learning Dart to avoid context switching while providing them useful information if they need.

To support that functionality as well as help developers better understand and discover important terms today, we'd like to add glossary entries for terms relevant to Dart, contextualized to Dart and its surrounding ecosystem. We'd love your help to document terms you know well or found important in your Dart journey. We're beyond happy to help review and edit your change, bringing your expertise to Dart developers of all calibers.

Glossary entries are added in the [`glossary.yml`](https://github.com/dart-lang/site-www/blob/main/src/_data/glossary.yml) file.

### Glossary entry format

Each entry in the [`glossary.yml`](https://github.com/dart-lang/site-www/blob/main/src/_data/glossary.yml) file has the following fields:

<details>
<summary><strong>Expand for format, purpose, and options of glossary entries</strong></summary>

- `term` _(required)_
  The term the entry is for. Can be multiple words, but should preferably fit in one line.

- `short_description` _(required)_
  A 1-2 sentence description of the term in the context of Dart. The description should provide enough info to refresh someone previously familiar with the term and enough info for anyone to determine if they want to learn more.'

- `long_description` _(optional, but recommended)_
  A full description of the word in the context of Dart and/or Flutter. Can be written in standard Markdown, with cross-links, code blocks, etc.

- `labels` _(optional)_
  A list of up to 5 relevant labels to enable filtering and categorizing this term. While this is optional, it’s best to provide at least one label. You can reference existing labels in the file to align with them and for inspiration, but feel free to use other ones that you think will aid discovery.

- `related_links` _(optional)_
  A list of links to resources related to the word. Resource don't have to be on the Dart site, but should be reliable and reputable. Each resource has the following information:
  
  - `text` _(required)_
    A descriptive link text, such as the title of a page or video.
  - `link` _(required)_
    The link of the resource.
  - `type` _(optional)_
    The content type of the resource to help readers understand its usefulness to them. Currently the following resource types are supported, but I'm happy to add support for more if necessary:
    
    - `term` _(for other glossary entries or word definitions)_
    - `article` or `doc` _(general for doc pages)_
    - `tutorial` _(for tutorials, codelabs, etc.)_
    - `api` _(for API links)_
[...]

PR #6618, opened 2025-05-24
Glossary entries for important dart terms

IldySilva:
contributes to #6461  by adding Zone and Late terms

### DIFF
--- a/src/_data/glossary.yml
+++ b/src/_data/glossary.yml
@@ -265,7 +265,7 @@
   alternate:
     - "with mixin"
     
-- term: "late"
+- term: "Late"
   id: "late"
   short_description: |-
     A keyword that enables deferred initialization of non-nullable variables.
@@ -758,11 +758,6 @@
     - Maintaining request-specific state across async gaps (e.g., in server apps)
     - Testing and debugging async behavior
 
-    Flutter and some backend frameworks (like `shelf`) use zones internally
-    to track things like user sessions or HTTP request context.
-
-    Zones form a tree, where each zone has a parent. Unless customized, 
-    async operations inherit the current zone of the calling context.
   related_links:
     - text: "Zones API documentation"
       link: "https://api.dart.dev/stable/dart-async/Zone-class.html"
--- a/src/_data/glossary.yml
+++ b/src/_data/glossary.yml
@@ -264,6 +264,67 @@
     - "understanding diagnostics"
   alternate:
     - "with mixin"
+    
+- term: "late"
+  id: "late"
+  short_description: |-
+    A keyword that enables deferred initialization of non-nullable variables.
+
+    In Dart, `late` is used when a variable is guaranteed to be initialized 
+    before it's accessed but cannot be initialized at the point of declaration.
+  long_description: |-
+    The `late` keyword in Dart is used to indicate that a non-nullable variable 
+    will be initialized later, after its declaration, but before it's used. 
+    This helps avoid the need to make a variable nullable (`?`) when you know 
+    it will definitely receive a value, just not immediately.
+
+    Using `late` defers initialization, allowing you to write more flexible 
+    and readable code, especially when dealing with dependencies or complex setup.
+
+    For example:
+    ```dart
+    late String description;
+
+    void setup() {
+      description = 'This will be initialized before use.';
+    }
+    ```
+
+    If you try to read a `late` variable before it's initialized, Dart throws a 
+    `LateInitializationError`.
+
+    You can also use `late final` when the variable should only be set once. 
+    This is common in dependency injection scenarios.
+
+    Example:
+    ```dart
+    class MyService {
+      late final Logger logger;
+
+      MyService(this.logger);
+    }
+    ```
+
+    Be cautious: if the `late` variable is never initialized, it will cause a 
+    runtime error when accessed.
+  related_links:
+    - text: "Null safety in Dart"
+      link: "/null-safety"
+      type: "doc"
+    - text: "Late variables"
+      link: "https://dart.dev/language/variables#late-variables"
+      type: "doc"
+    - text: "Error: LateInitializationError"
+      link: "https://api.dart.dev/stable/dart-core/LateInitializationError-class.html"
+      type: "api"
+  labels:
+    - "language"
+    - "variables"
+    - "null safety"
+  alternate:
+    - "late keyword"
+    - "deferred initialization"
+    - "late final"
 
 - term: "Override inference"
   short_description: |-
--- a/src/_data/glossary.yml
+++ b/src/_data/glossary.yml
@@ -716,3 +716,69 @@
     - "covariant"
     - "contravariance"
     - "contravariant"
+
+
+- term: "Zone"
+  id: "zone"
+  short_description: |-
+    A mechanism in Dart to customize the behavior of asynchronous code 
+    without modifying the async code itself.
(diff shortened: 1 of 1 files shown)


src/_data/glossary.yml
```diff
@@ -264,6 +264,67 @@
     - "understanding diagnostics"
   alternate:
     - "with mixin"
+    
+- term: "Late"
+  id: "late"
+  short_description: |-
+    A keyword that enables deferred initialization of non-nullable variables.
```
lrhn: Consider "of (usually non-nullable) variables".

(The feature works for nullable variables too, it's just not necessary for non-`final` variables since they are implicitly initialized to `null`.)

src/_data/glossary.yml
```diff
@@ -264,6 +264,67 @@
...
+
+    In Dart, `late` is used when a variable is guaranteed to be initialized 
+    before it's accessed but cannot be initialized at the point of declaration.
```
lrhn: Comma before "but"?

src/_data/glossary.yml
```diff
@@ -264,6 +264,67 @@
...
+  long_description: |-
+    The `late` keyword in Dart is used to indicate that a non-nullable variable
```
lrhn: I'd drop "non-nullable" here.

src/_data/glossary.yml
```diff
@@ -264,6 +264,67 @@
...
+    ```dart
+    late String description;
+
+    void setup() {
+      description = 'This will be initialized before use.';
+    }
+    ```
+
+    If you try to read a `late` variable before it's initialized, Dart throws a 
+    `LateInitializationError`.
```
lrhn: I'd emphasize this part, more than what is already done below, because it's what makes `late` a dynamically failing feature, like `dynamic` function calls.
Perhaps adding 
> Be careful with late variables that are part of a public API. If clients of the API can read the variable before it's initialized, they will get a runtime error with a non-descriptive `LateInitializationError`. Having a nullable private variable and a public getter which throws, for example, a `StateError` if read when the value is still `null`, to give the client a clearer error message, may be worth the extra complexity.

Or something. It's a dynamic feature which turns off static safety checks in the language, like `dynamic`, which makes it dangerous if used without caution.
It can fail at runtime if you use it incorrectly, with no static warning, like `dynamic`.
(But with an initializer expression, it's completely risk free, so use that if you possibly can.)

Also, a variable being `late` is not *free*. It has an extra cost in the getter even if you don't write an `int get foo => _foo ?? (throw LateInitializationError());` yourself, the implicit getter does effectively the same thing.

I generally recommend not exposing a late variable without an initializer expression. Only keep them where you can control access to ensure they're not read early.

src/_data/glossary.yml
```diff
@@ -264,6 +264,67 @@
...
+
+    You can also use `late final` when the variable should only be set once. 
+    This is common in dependency injection scenarios.
+
+    Example:
+    ```dart
+    class MyService {
+      late final Logger logger;
+
+      MyService(this.logger);
```
lrhn: (The `this.logger` should be removed, otherwise the `late` has no effect.)

src/_data/glossary.yml
```diff
@@ -264,6 +264,67 @@
...
+    }
```
lrhn, line 314: This is what I'd consider a *bad* example, because it exposes the late variable in the public API, at a point where it might not be initialized yet.

Consider instead something like this, where a cyclic dependency means you cannot create the both of the two objects directly in their constructors:
```dart
class LinkedQueue<T> {
  late final QueueLink<T> _head;
  LinkedQueue() {
    _head = QueueLink<T>._head(owner: this); // Cyclic reference between objects.
  }
  // ...
}
```

Or maybe an example similar to how it's usually used in tests, but that's harder to make a generic version of.

src/_data/glossary.yml
```diff
@@ -264,6 +264,67 @@
...
+    ```
+
+    Be cautious: if the `late` variable is never initialized, it will cause a 
+    runtime error when accessed.
```
lrhn: ... or just if it's accessed too soon.

src/_data/glossary.yml
```diff
@@ -655,3 +716,64 @@
     - "covariant"
     - "contravariance"
     - "contravariant"
+
+
+- term: "Zone"
+  id: "zone"
+  short_description: |-
+    A mechanism in Dart to customize the behavior of asynchronous code 
+    without modifying the async code itself.
```
lrhn: Also a dynamic execution context that is persistant across asynchronous delays.
I find that the *context* is often more important than modifying `dart:async` functionality (which only experts should do, it's too easy to get wrong.)

Maybe start with the execution context:

>     A dynamic execution context that is preserved across an `await`, and remembered
>     by asynchronous callbacks.
>     Can store values and can customize the behavior of the asynchronous functionality
>     of `dart:async`.

Or something like that.

src/_data/glossary.yml
```diff
@@ -655,3 +716,64 @@
...
+  long_description: |-
+    A zone is an execution context that allows you to run code with
+    customized behavior for asynchronous events such as timers,
+    microtasks, and uncaught errors.
+
+    Zones provide a way to track and influence asynchronous execution 
+    without requiring the async code to be aware of it.
+
+    You can create a new zone using `runZoned` (or `runZonedGuarded`)
+    and override zone-specific features like error handling, timers, and print.
```
lrhn: Nit: `print` is not an asynchronous event or functionality. It was just added as an override in `Zone` because it's the only place we had overrides.

src/_data/glossary.yml
```diff
@@ -655,3 +716,64 @@
...
+    }
+    ```
+
+    In this example, the uncaught error inside the async callback
+    is intercepted by the custom zone.
+
+    Zones are useful for:
+    - Logging
+    - Error tracking
+    - Maintaining request-specific state across async gaps (e.g., in server apps)
```
lrhn: I'd put this at the top. This is the primary use-case for zones that most people should use.

src/_data/glossary.yml
IldySilva:
```suggestion
    A keyword that enables deferred initialization of variables, typically used with non-nullable variabels.
```
?

src/_data/glossary.yml
IldySilva:
```suggestion
    before it's accessed, but cannot be initialized at the point of declaration.
```

src/_data/glossary.yml
IldySilva:
```suggestion
    The `late` keyword in Dart is used to indicate that a variable 
```

src/_data/glossary.yml
IldySilva: I'd prefer to keep the  usescase  after the definition, since the goal here is to explain what the term is, as it's meant for a glossary.

src/_data/glossary.yml
antfitch: Closed since this seems to be outdated.

src/_data/glossary.yml
antfitch: It looks like this is outdated now. Closing.

antfitch:
This looks great @IldySilva! I've merged your changes. If you have more entries to add, go for it!

IldySilva:
Awesome ,thank you @antfitch , i'll certainly work on it

PR #6619, opened 2025-05-24
Fix broken link for Glossary card on resources page

IldySilva:
The "Glossary" card on the /resources page was incorrectly linking to `/resources/breaking-changes`. This PR updates the link to point to the correct glossary location at` /resources/glossary`


parlough (approved):
Thanks so much for catching and fixing this, I appreciate it! Looks good to me 🚀

PR #6658, opened 2025-06-10
Glossary entries for important dart terms

IldySilva:
contributes to #6461 by adding Wildcard


antfitch:
@parlough or @lrhn can one of you do a quick technical review? I'll follow up with grammar edits if any are needed.

src/_data/glossary.yml
```diff
@@ -727,6 +727,51 @@
...
+    Wildcards help make the code more intentional by clearly marking
+    values that are not needed in a specific context.
+
+    Example:
+    ```dart
+    // Ignoring the index in a forEach loop
+    var names = ['Alice', 'Bob', 'Charlie'];
+    for (var _ in names) {
+      print('Someone is here!');
+    }
```
lrhn, line 751: Consider also having an example with a wildcard parameter. Something like:
```dart
  map.forEach((key, _) {
    use(key);
  }
```
(Not a great example, you can just do `map.keys.forEach((key) { ... })`.)

src/_data/glossary.yml
```diff
@@ -727,6 +727,51 @@
     - "contravariance"
     - "contravariant"
 
+- term: "Wildcard"
+  id: "wildcard"
+  short_description: |-
+    A symbol (`_`) that represents an unused or unnamed value in patterns and other contexts.
+  long_description: |-
+    A wildcard is the underscore character (`_`) used to ignore values or indicate
```
lrhn: Consider "used ..." -> "used instead of a variable name ..."

lrhn (approved):

src/_data/glossary.yml
IldySilva: I think it might be better to keep things concise here, another example may be unnecessary, since the description already explains the concept clearly @lrhn

antfitch (approved):
Looks great!

PR #6672, opened 2025-06-18
Fix cookie notice background color in dark mode

parlough:
Instead of always assuming it should have a white background, use a css property to respond to the current theme.


antfitch (approved):
Yay!

PR #6676, opened 2025-06-19
Add "immutable" term to glossary

IldySilva:
Contributes to #6461 by adding "immutable" term.

**Staged:** https://dart-dev--pr6676-main-wfby8hmu.web.app/resources/glossary#immutable


antfitch:
You are amazing @IldySilva! Will get to this on Monday.

src/_data/glossary.yml
```diff
@@ -194,6 +194,46 @@
     - "umbrella term"
   alternate:
     - "Procedure"
+- term: "Immutable"
+  short_description: |-
+    An object whose state cannot be changed after it is created.
+  long_description: |-
+    An immutable object is one whose state cannot be modified after it is created.
+    This means all of its fields must be final.
```
lrhn: and the fields' values must also be immutable.

(We generally use "immutable" for deep immutability and "unmodifiable" for shallow, where the fields are final, but their values can be modified. You need deep immutability for the benefits mentioned below.)

src/_data/glossary.yml
```diff
@@ -194,6 +194,46 @@
...
+
+    Immutability helps ensure consistency and enables
+    safer use in concurrent or reactive code.
+
+    In Dart, you can make a class immutable by:
+      - Marking it with the `@immutable` annotation from `meta` package
```
lrhn: Which I presume is to warn you if the field values are not immutable types. It doesn't actually enforce anything in the language, it's just an analyzer warning. 
Maybe phrase it as:

> In Dart, you can have the analyzer help you ensure that a class is immutable by
> marking it with the `@immutable` annotation from the `meta` package.
> That will check that:
> - All fields are `final`, and
> - All fields have immutable types.

src/_data/glossary.yml
```diff
@@ -194,6 +194,46 @@
...
+      class User {
+        final String name;
+        final int age;
+
+        const User(this.name, this.age);
+      }
+      ```
+
+    Once created, you cannot modify a `User` instance. You must create a new one
+    to "change" any data.
```
lrhn: I would *definitely* also mention constants. Something like.

> All Dart constant values are immutable. A `const [v1, v2, v3]` creates an immutable list, and any class
> which has a `const` (non-factory) constructor must have only `final` fields.

antfitch:
Hi @IldySilva it looks like there are a few suggestions in here. When you've looked at them, let us know. Thanks for adding these terms.

IldySilva:
Thanks @lrhn  for the helpful points!

antfitch (approved):
Looks great!

PR #6697, opened 2025-06-29
Fix minor typo in extension-types.md

rahulrao0209:


src/content/language/extension-types.md
```diff
@@ -87,7 +87,7 @@ The declaration also introduces:
 - An implicit constructor: `E(int i) : i = i`.
 
 The representation object gives the extension type access to an object
-at the underlying type.
+of the underlying type.
```
eernstg, line 90: I'm afraid I don't think this is an improvement.

The representation variable (here `i`) gives the extension type (and clients, by the way) access to the representation object using an expression whose static type is the representation type (here `int`). I think it makes sense to say that the representation object is accessed _at_ the representation type, because this terminology indicates that it could also be accessed at other types (for instance, every object can be accessed at the type `Object?`).

This emphasis on the chosen type makes sense because the extension type itself is a different way to access the very same object at a different type, namely the extension type. The type is a _perspective_ on the representation object that we've chosen in a given setting, not an inherent property of the representation object itself. Of course, the only perspectives we're allowed to use are the types that can soundly be used as the static type of the given object, but this doesn't change the fact that it's all about choosing a perspective rather than blindly reporting a property of the representation object itself.

So if we have `var e = E(42);` then the representation object is the value of `42`, but `e` will access that object at the type `E` and `e.i` will access the very same object at the type `int`.

The word 'of' seems to imply that the given type is a property of the object itself. This is misleading because it conflicts with the ability to talk about multiple types that are used to access the same object.

parlough (approved):
Thanks for noticing and fixing this!

eernstg:
Oh, @rahulrao0209 and @parlough, I'd recommend that we do not make this change. Please take a look at my comment above.

parlough:
Happy to revert this or make further changes @eernstg, thanks for calling it out. I can't seem to see your comment though, would you mind reposting it?

eernstg:
>  I can't seem to see your comment though, would you mind reposting it?

Done, thanks!

musvaage:
_off-topic_

A <kbd>Member</kbd> is welcome to resolve pull `#6701`.

My GitHub account does not have a Google account associated with it.

IMHO, the google-cla is unnecessarily restrictive on which see ticket `#6705`.

src/content/language/extension-types.md
rahulrao0209: Thanks a lot for the detailed explanation @eernstg. I am not well-versed enough with Dart yet to grasp this nuance and assumed the word 'at' to be a simple grammatical error while reading through the docs.
@parlough please feel free to revert and apologies for the trouble!

parlough: No troubles at all @rahulrao0209! Bringing attention to this was very helpful as I agree the existing wording was a potential source of confusion as "at a type" isn't super familiar or common phrasing.

In https://github.com/dart-lang/site-www/pull/6710 I landed some cleanup for the page formatting and temporarily reverted to the "at" phrasing.

Then in https://github.com/dart-lang/site-www/pull/6713 I landed an updated version of the text that Erik and I decided on. Consider checking that out and seeing if it makes more sense.

Thanks again for your attention and work here!! I really appreciate it :D

rahulrao0209: Thank you! @parlough 
And yeah, the updated version does make more sense to me.

Issue #6631, opened 2025-05-31
Add a glossary entry for "Context type"


parlough:
As used by [dot shorthands](https://github.com/dart-lang/language/blob/main/working/3616%20-%20enum%20value%20shorthand/proposal-simple-lrhn.md)

parlough:
@kallentu or @lrhn Would one of you be interested in adding this glossary entry? Or able to provide some info/pointers for someone who wants to add it. Thanks :D

lrhn:
Hmmm. I'm not good a being brief, which a really the most important part of a glossary. And correct. Brief and correct. And understandable ... \<Monty Python inquisition meme\>

So let me write _something_, then the wordsmiths can try to tweak it.

```yaml
- term: "Context type"
  short_description: |-
    A type expected of an expression by the code surrounding the expression.
  long_description: |-
    An expression evaluates to a value. Often that value is passed somewhere
    else, like being stored in a variable, passed as argument to a function or
    returned by a function.

    The parameter type of the function being called, the return type of the
    function being returned from, or the declared type of the variable being
    assigned to, is the type *expected* by the expression providing the value.
    Dart calls this type the *context type* of the expression, and in some
    places, the meaning of an expression *depends* on its context type:

    - Type inference can use the context type, that's what's called "downwards inference".
      For example:

       ```dart
           List<int> list = [];
       ```
      The context type of `[]` is `List<int>`, which inference uses to infer a type argument of `<int>[]`.
   
    - Implicit coercions. For example implicit downcast from `dynamic`:
      
      ```dart
      String asString(dynamic value) => value;`
      ```

      The context type of the returned expression `value` is `String`,
      and `value` has type `dynamic`, so there is an implicit downcast of `value as String`.
      (The other implicit coercions, `.call`-tear-off and generic instantiation, also
      use the context type.)
     
    - Double literals without `.` or `e`, like:

      ```dart
      double nul = 0;
      ```

      The context type of `0` is `double`, so it becomes a double literal with the same value as `0.0`.
     
    - Dot-shorthands: 

       ```dart
       int x = .parse(input);
       ```

       The context type of `.parse(input)` is `int`, so `parse` is looked up on `int,
       and it becomes `int.parse(input)`.

  The context type is found and used during *type inference*.

  Not all expressions have context types.
   
  - Some don't need a context type, when the expression is used for its effect, not its value.
    For example: `set.remove(value);` is an expression-statement, an expression followed by `;`,
[...]

Issue #4745, opened 2023-04-06
Overhaul and introduce new type system documentation


parlough:
Currently our [type system docs](https://dart.dev/language/type-system) are old and limited.

We need to introduce new type system, type inference, and similar documentation, potentially including it under **Language > Types**. Some of these details can be gathered from [Understanding null safety](https://dart.dev/null-safety/understanding-null-safety) but that doesn't cover everything, it is written for readers transitioning from a pre null-safety world, and is perhaps too detailed for most users. We will need to find a balance.

This is related to https://github.com/dart-lang/site-www/issues/4603.

PR #6699, opened 2025-06-29
Add context type term  to glossary

IldySilva:
contributes to #6631 by adding Context type term

### DIFF
--- a/src/_data/glossary.yml
+++ b/src/_data/glossary.yml
@@ -17,6 +17,61 @@
   alternate:
     - "Quick assist"
 
+- term: "Context type"
+  short_description: |-
+    A type expected from an expression based on how it's used.
+  long_description: |-
+    A **context type** is the type that surrounding code expects from an expression, 
+    such as a variable type, a parameter type, or a return type.
+
+    Dart uses the context type to interpret and infer meaning from expressions, including:
+
+    - **Type inference** ("downwards inference"):
+      ```dart
+      List<int> list = [];
+      ```
+      The context type `List<int>` lets the compiler infer the list type as `<int>[]`.
+
+    - **Implicit coercion** (e.g., downcast):
+      ```dart
+      String asString(dynamic value) => value;
+      ```
+      The return context is `String`, so Dart inserts an implicit downcast.
+
+    - **Literal interpretation**:
+      ```dart
+      double d = 0;
+      ```
+      The context type `double` makes `0` behave like `0.0`.
+
+    - **Static access shorthand** (dot shorthand):
+      ```dart
+      int x = .parse(input);
+      ```
+      The context type is `int`, so `.parse` resolves to `int.parse(input)`.
+
+    Some expressions have **no context type**:
+
+    - **Expression statements**, like `set.remove(value);`, are used for effect — not value — so no type is expected.
+    - **When type flows from the expression to the context**:
+      ```dart
+      var list = [1];
+      ```
+      `[1]` has no context type; Dart infers `List<int>` from its contents.
+
+    This is sometimes called the **empty context type**.
+
+  related_links:
+    - text: "Type inference"
+      link: "https://dart.dev/language/type-system#type-inference"
+      type: "doc"
+    - text: "Implicit downcast from `dynamic`"
+      link: "https://dart.dev/language/type-system#implicit-downcasts-from-dynamic"
+      type: "doc"
+  labels:
+    - "language"
+    - "type system"
+
 - term: "Constant context"
   short_description: |-
     A region of code where the const keyword is implied and


IldySilva:
based on the [explanation and examples provided by @irhn.](https://github.com/dart-lang/site-www/issues/6631#issuecomment-3004071322) ,  worked on shortening and structuring it for the glossary format.

@parlough  @antfitch

src/_data/glossary.yml
```diff
@@ -17,6 +17,61 @@
...
+
+    Dart uses the context type to interpret and infer meaning from expressions, including:
+
+    - **Type inference** ("downwards inference"):
+      ```dart
+      List<int> list = [];
+      ```
+      The context type `List<int>` lets the compiler infer the list type as `<int>[]`.
+
+    - **Implicit coercion** (e.g., downcast):
```
eernstg: The word 'coercion' typically refers to a computation (e.g., computing a `double` value which is considered to be "the same" as a given `int` value, or computing a non-generic function object based on a generic function object and a context type that determines the actual type arguments).

In this case there is no computation, it is simply checked that the given object has the required type already (here: we check that it is a `String`). If the object doesn't already have the required type then the ongoing computation is cancelled by throwing an error.

For this reason I'd recommend that the word 'coercion' is avoided. Perhaps it could simply be written as follows?:

```suggestion
    - **Implicit downcast**:
```

IldySilva: Great, @eernstg  thanks for the explanation! I’ll switch to “implicit downcast” as suggested.

eernstg: Thanks!

IldySilva: Since this PR is based on @lrhn’s original comment, I’d love to hear if he’s also OK with replacing  ,  just to be sure we stay aligned with his original wording.

eernstg:
Hi @parlough, I was added as a reviewer because I had a single comment. However, I don't think I should finish the process. Could you suggest a reviewer who would be a regular site-www reviewer?

src/_data/glossary.yml
```diff
@@ -17,6 +17,61 @@
...
+    Some expressions have **no context type**:
+
+    - **Expression statements**, like `set.remove(value);`, are used for effect — not value — so no type is expected.
+    - **When type flows from the expression to the context**:
+      ```dart
+      var list = [1];
+      ```
+      `[1]` has no context type; Dart infers `List<int>` from its contents.
+
+    This is sometimes called the **empty context type**.
```
parlough: My feeling is we can drop this sentence, at least in this context of a glossary entry. If it becomes a common term or we use it in the docs, we can revisit it or even add it as its own entry.

src/_data/glossary.yml
```diff
@@ -17,6 +17,61 @@
...
+
+  related_links:
+    - text: "Type inference"
+      link: "https://dart.dev/language/type-system#type-inference"
```
parlough: Thanks for including these links to related resources! 

Can you update these to use root-relative links though, to be consistent with the other glossary entries?

```suggestion
      link: "/language/type-system#type-inference"
```

src/_data/glossary.yml
```diff
@@ -17,6 +17,61 @@
...
+      type: "doc"
+    - text: "Implicit downcast from `dynamic`"
+      link: "https://dart.dev/language/type-system#implicit-downcasts-from-dynamic"
```
```suggestion
      link: "/language/type-system#implicit-downcasts-from-dynamic"
```

src/_data/glossary.yml
```diff
@@ -17,6 +17,61 @@
   alternate:
     - "Quick assist"
 
+- term: "Context type"
+  short_description: |-
+    A type expected from an expression based on how it's used.
```
parlough: I think this short description is great, but do you think it would make sense to adjust it to say "based on how **and where** it's used"? It seems a lot of the later examples relate to location.

src/_data/glossary.yml
```diff
@@ -17,6 +17,61 @@
...
+  long_description: |-
+    A **context type** is the type that surrounding code expects from an expression, 
+    such as a variable type, a parameter type, or a return type.
+
+    Dart uses the context type to interpret and infer meaning from expressions, including:
+
+    - **Type inference** ("downwards inference"):
+      ```dart
```
parlough, line 33: A small nit about formatting, but can you add an extra blank line before and after code blocks? They make code blocks a bit more visually distinct and easy to review, so it's the standard we follow across the site.

````md
Content introduction...

```dart
void main() {
  // ...
}
```

Other content here...
````

File: src/_data/glossary.yml [hunk 7]
```diff
@@ -17,6 +17,61 @@
...
+      String asString(dynamic value) => value;
+      ```
+      The return context is `String`, so Dart inserts an implicit downcast.
+
+    - **Literal interpretation**:
+      ```dart
+      double d = 0;
+      ```
+      The context type `double` makes `0` behave like `0.0`.
```
parlough, line 53: I think this make sense, but I worry "behave like" might cause some readers to misinterpret about when this happens. I think it's okay to be a bit more specific, even though it makes the explanation a bit longer.

How does something like this sound?

```
      The context type of `double` results in `0` being treated as a double literal with the same value as `0.0`.
```

src/_data/glossary.yml
```diff
@@ -17,6 +17,61 @@
...
+    - **Static access shorthand** (dot shorthand):
+      ```dart
+      int x = .parse(input);
+      ```
+      The context type is `int`, so `.parse` resolves to `int.parse(input)`.
+
+    Some expressions have **no context type**:
+
+    - **Expression statements**, like `set.remove(value);`, are used for effect — not value — so no type is expected.
+    - **When type flows from the expression to the context**:
```
parlough: Can you update these two list items to have [parallel construction](https://developers.google.com/style/lists#parallel-syntax).

I'm not exactly sure the best way yet, but I'll give it some more thought or perhaps @antfitch has a suggestion.

Maybe something like the following?

```markdown
Sometimes expressions have **no context type**, including:

- **When used as statements**:
  When expressions are used as statements, such as in `set.remove(value);`, just for their effect — not ...
- **When the context's type are inferred from the expression**:
  ...
```

src/_data/glossary.yml
```diff
@@ -17,6 +17,61 @@
...
+      ```dart
+      var list = [1];
+      ```
+      `[1]` has no context type; Dart infers `List<int>` from its contents.
```
parlough: I think it might be helpful to be specific about what "Dart is inferring `List<int>`" for. I guess it's for both the list itself and the variable, but maybe it's sufficient to just mention for the list.

src/_data/glossary.yml
IldySilva: Yeah @parlough, I really need to pay more attention to those 😅 , thanks for catching it!

src/_data/glossary.yml
IldySilva: based on [Documentation style guide](https://developers.google.com/style/lists#parallel-syntax), make sense
i"m updating with @parlough  suggestion

src/_data/glossary.yml
IldySilva: Sounds good — I’ll remove it for now.

antfitch:
> Thanks again for working on these glossary entries @IldySilva, I really appreciate it. This one will be helpful to have for upcoming diagnostics as well.
> 
> Overall the entry looks good to me. I did leave some comments with minor fixes and cleanup for you consider. I'd also like @antfitch to review the text when she has a chance.
> 
> This entry is pretty long but I think that's probably good for now, as we don't yet have proper type system or type inference documentation where this longer-form content can live. Once we have a chance to do that _(tracked in #4745)_, we can revisit moving some of this content there.

Looking over this now @parlough. @IldySilva I'll make small copy edits and squash this for you if everything looks great.

PR #6721, opened 2025-07-13
Add glossary entry for bottom type

IldySilva:
Adds  "Bottom type" term to glossary.

Preview: https://dart-dev--pr6721-main-ejjxss0q.web.app/resources/glossary#bottom-type

### DIFF
--- a/src/_data/glossary.yml
+++ b/src/_data/glossary.yml
@@ -107,7 +107,7 @@
     Since `fail` never returns, it has type `Never`, which can be assigned to any other type.
 
   related_links:
-    - text: "Builtin types"
+    - text: "Built-in types"
       link: "/language/built-in-types"
       type: "doc"
     - text: "Never for unreachable code"
--- a/src/_data/glossary.yml
+++ b/src/_data/glossary.yml
@@ -85,14 +85,17 @@
   short_description: |-
     A type that has no values and is a subtype of all other types.
   long_description: |-
-    The **bottom type** in Dart is the type that has **no values** and is considered a **subtype of every other type**.
+    The **bottom type** in Dart is the type that has **no values** and is considered a [**subtype**](/resources/glossary#subtype) of every other type.
 
     In Dart, the bottom type is represented by the `Never` type.
 
     This means a value of type `Never` can be used anywhere, because such a value can never actually exist. 
-    It's most often used to represent functions or expressions that **never return**, such as those that throw exceptions or loop forever.
+    It's most often used as the **return type** of functions to indicate they **never return**,  
+    such as for those that throw exceptions or loop forever.
 
     Example:
+
+    The `fail` function below always throws an exception, so it’s declared with return type `Never` to indicate that it never returns.
 
     ```dart
     Never fail(String message) {
@@ -104,7 +107,7 @@
     }
     ```
 
-    Since `fail` never returns, it has type `Never`, which can be assigned to any other type.
+    Since `fail` returns Never, assigning it to a String is allowed    
 
   related_links:
     - text: "Built-in types"
--- a/src/_data/glossary.yml
+++ b/src/_data/glossary.yml
@@ -81,6 +81,44 @@
     - "language"
     - "type system"
 
+- term: "Bottom type"
+  short_description: |-
+    A type that has no values and is a subtype of all other types.
+  long_description: |-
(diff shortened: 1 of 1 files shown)


antfitch (changes requested):
This looks excellent to me. One note is that I'm seeing a site-shared updates in this change. @parlough, do you want to check those and also the technical side of the PR?

parlough:
As always, thanks for this PR @IldySilva! I'll make sure to take a look soon.

As for the `site-shared` git submodule, it looks like you might have commit an outdated version. You can update all submodules to latest with `git submodule foreach git pull origin main` from the repo root directory, then commit that update. If you have trouble with that or hit a problem, let me know.

Also, for future reference, consider working on changes and opening PRs from branches besides `main`. This allows you to work on multiple changes at the same time if ever necessary and is a common  practice. https://git-scm.com/book/en/v2/Git-Branching-Basic-Branching-and-Merging seems to have some details. If you're using the GitHub UI, they have some docs on [Creating and deleting branches within your repository](https://docs.github.com/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-and-deleting-branches-within-your-repository).

IldySilva:
Understood , @parlough , can you check if its ok now?

antfitch (approved):
Looks great @IldySilva and I am LOVING your contributions.

src/_data/glossary.yml
```diff
@@ -81,6 +81,44 @@
...
+    
+    void main() {
+      String result = fail('Oooops'); // OK: Never is a subtype of String
+    }
+    ```
+
+    Since `fail` never returns, it has type `Never`, which can be assigned to any other type.
+
+  related_links:
+    - text: "Builtin types"
```
```suggestion
    - text: "Built-in types"
```

src/_data/glossary.yml
```diff
@@ -81,6 +81,44 @@
...
+      link: "/language/built-in-types"
+      type: "doc"
+    - text: "Never for unreachable code"
```
parlough, line 119: Great thought to link to this section!

src/_data/glossary.yml
```diff
@@ -81,6 +81,44 @@
     - "language"
     - "type system"
 
+- term: "Bottom type"
+  short_description: |-
+    A type that has no values and is a subtype of all other types.
+  long_description: |-
+    The **bottom type** in Dart is the type that has **no values** and is considered a **subtype of every other type**.
```
parlough: Consider cross linking the "subtype" text here to [/resources/glossary#subtype](https://dart.dev/resources/glossary#subtype).

src/_data/glossary.yml
```diff
@@ -81,6 +81,44 @@
...
+
+    In Dart, the bottom type is represented by the `Never` type.
+
+    This means a value of type `Never` can be used anywhere, because such a value can never actually exist. 
+    It's most often used to represent functions or expressions that **never return**, such as those that throw exceptions or loop forever.
```
parlough: I'm not sure `Never` itself represents the functions and expressions that never return, but rather is used by them to indicate they never return.

How about rewording this slightly to something like:

```suggestion
    It's most often used as the return type of functions to indicate they **never return**,
    such as for those that throw exceptions or loop forever.
```

src/_data/glossary.yml
```diff
@@ -81,6 +81,44 @@
...
+
+    Example:
```
parlough: How about add a bit more introduction to the code snippet? That way readers know what they should consider and look for as they read through it. Perhaps something like:

```
For example, the `fail` function in the following snippet always throws an exception.
So it is declared with a return type of `Never` to indicate it will never return.
```

src/_data/glossary.yml
```diff
@@ -81,6 +81,44 @@
...
+    Never fail(String message) {
+      throw Exception(message);
+    }
+    
+    void main() {
+      String result = fail('Oooops'); // OK: Never is a subtype of String
+    }
+    ```
+
+    Since `fail` never returns, it has type `Never`, which can be assigned to any other type.
```
parlough: If you do expand the example introduction, consider updating this as well to focus just on the assignment to `String`.

src/_data/glossary.yml
IldySilva: Good catch

IldySilva:
awesomeee,letss movee on

Issue #6688, opened 2025-06-23
Dart 3.10 dart install


antfitch:

dcharkes:
We didn't merge `dart install` documentation:

* https://github.com/dart-lang/site-www/pull/6725

PR #6725, opened 2025-07-14
Add dart-install guide

antfitch:
Adds a guide for `dart install` to Tools & techniques section.

Fixes [6688](https://github.com/dart-lang/site-www/issues/6688)

---

### DIFF
--- a/src/content/tools/dart-install.md
+++ b/src/content/tools/dart-install.md
@@ -1,0 +1,285 @@
+---
+title: dart install
+description: Install Dart CLI tools for global use.
+---
+
+The `dart install` command installs Dart CLI tools for global use.
+It is a newer alternative to `dart pub global activate`.
+
+## Install a package
+
+There are a few ways to install a package. To learn more,
+see the following sections.
+
+### Install a general package
+
+The following command installs all executables specified in
+a package's `pubspec.yaml` [executables][] section on the
+PATH. If the executables section doesn't exist, it installs all
+`bin/*.dart` entry points as executables.
+
+```console
+$ dart install [arguments] <package> [version-constraint]
+```
+
+For example:
+
+```console
+$ dart install markdown
+```
+
+### Install a pub.dev package
+
+The following command specifies a package on the pub.dev
+site to install.
+
+```console
+$ dart pub install <pub.dev package>
+```
+
+For example:
+
+```console
+$ dart pub install markdown
+```
+
+### Install a Git package
+
+The following commands can both be used to install a package
+in a Git repository.
+
+```console
+$ dart install --source git <Git URL>
+```
+
+```console
+$ dart install -sgit <Git URL>
+```
+
+The following examples, which install the `async_await` package on
+[GitHub][], are equivalent:
+
+```console
+$ dart install --source git https://github.com/dart-lang/async_await.git
+$ dart install -sgit https://github.com/dart-lang/async_await.git
+```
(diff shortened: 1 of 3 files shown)


dcharkes:
Thanks @antfitch! 🙏 

We can't merge this yet, I'm still working on a `dart install` implementation. Once I've implemented it, it will only be available in dev releases (and then the next beta).

I've made the edits that I'm aware of.

dcharkes:
FYI: `dart install` is now available in the `dev` releases.

src/content/tools/dart-install.md
```diff
@@ -0,0 +1,240 @@
...
+## Install a package
+
+There are a few ways to install a package. To learn more,
+see the following sections.
+
+### Install a general package
+
+The following command installs all executables specified in
+a package's `pubspec.yaml` [executables][] section on the
+PATH. If the executables section doesn't exist, it installs all
```
dcharkes: > If the executables section doesn't exist

This got changed, it only installs things from the executables section.

src/content/tools/dart-install.md
```diff
@@ -0,0 +1,240 @@
...
+
+Check which packages are installed
+
+```console
+$ dart installed
+```
+
+For example:
+
+## Options
```
dcharkes: These are the options to `dart install` they are not shared with `dart uninstall` or `dart installed`.

src/content/tools/dart-install.md
```diff
@@ -0,0 +1,240 @@
...
+
+### dart installed
+
+Check which packages are installed
+
+```console
+$ dart installed
+```
+
+For example:
```
dcharkes: Example is missing?

src/content/tools/dart-install.md
antfitch: Okay, great! Deleting this...

src/content/tools/dart-install.md
antfitch: Deleted the example since there's no difference from the command above it. If there's any syntax we're missing for this command, let me know.

src/content/tools/dart-install.md
antfitch: Thanks for catching this. I've updated the heading and the description below it. Take a close look.

antfitch:
@dcharkes, I added your changes. Take a close look.

src/content/tools/dart-install.md
```diff
@@ -0,0 +1,238 @@
+---
+title: dart install
+description: Install Dart CLI tools for global use.
+---
+
```
dcharkes: Do we need a note that it's available from Dart 3.10?

antfitch: I think we need to introduce a Preview note that we can add here and then delete when 3.10 ships. It's something that is commonly used elsewhere in docs at Google, so maybe we introduce that here. I'll check around with the team.

parlough: Since the command is now available on the beta channel and quite likely to land to stable, I think should just use a normal version note mentioning the required version. That way we can land the PR now, we don't have to update it after the release, and future developers continue to benefit from the required version information.

```suggestion
:::version-note
Support for `dart install` was introduced in Dart 3.10.
:::
```

antfitch: I'll add this for now, but I want to find a better way.

PR #6733, opened 2025-07-16
Switch CSP to report only for now

parlough:
The CSP introduced in https://github.com/dart-lang/site-www/pull/6728 breaks various functionality on the site, including the SDK archive, embedded videos, inline analytics, etc. as it isn't configured for the scripts those rely on. Until someone gets a chance to do so, we should disable the policy or switch it to report only for now. This PR switches it to report only for now which will log violations in the browser console.

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -21,7 +21,7 @@
       {
         "source": "**",
         "headers": [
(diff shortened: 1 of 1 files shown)


mit-mit:
Good catch @parlough, sorry my testing wasn't comprehensive enough

PR #6735, opened 2025-07-18
Update links and redirects to updated spec location

parlough:
Adds and uses a tooling redirect [dart.dev/to/spec-draft](https://dart.dev/to/spec-draft) to point to the new spec location on Google Cloud storage.

**Staged:** https://dart-dev--pr6735-misc-update-draft-spec-destination-icix1kwq.web.app/to/spec-draft


eernstg (approved):
The purpose of this PR looks good to me (very good)!

However, I can't verify the actual URLs because I do not know the mechanisms involved well enough to tell exactly what they should be.

I believe @athomas will be back on Monday.

antfitch (approved):

athomas:
Why do we want this? The cloud storage backend is an implementation detail. Ideally, we'd serve the spec from spec.dart.dev directly but Firebase hosting doesn't support GCS backends at this point.

PR #6736, opened 2025-07-18
typeInfo consists only: 'Data is an integer: 123'

Vic-Bonlight:
Screenshot from 2025-07-18 11-55-08

Running on Dart SDK version: 3.8.1 there is no 'Data is a double: 123' in the variable typeInfo  after running the example.

### DIFF
--- a/src/content/language/collections.md
+++ b/src/content/language/collections.md
@@ -597,7 +597,7 @@ var typeInfo = [
   if (data case String s) 'Data is a string: $s',
   if (data case bool b) 'Data is a boolean: $b',
   if (data case double d) 'Data is a double: $d',
-]; // [Data is an integer: 123, Data is a double: 123]
+]; // [Data is an integer: 123]
 ```
 
 <?code-excerpt "misc/test/language_tour/collections/if_case_operator_in_collection_d.dart (code_sample)"?>


antfitch:
Interesting. The default behavior should look like this:

Screenshot 2025-07-18 at 3 52 28 PM

To keep things aligned with DartPad, I'm going to keep things as they are in the docs, but I'll keep an eye on this. @parlough any clues as to why the result is different in DartPad vs terminal?

antfitch:
So, I checked with @leonsenft and he mentioned that the IDE is running the Dart VM, which has distinct numeric types for integer and double, but DartPad is running Dart compiled to JavaScript, which has no distinction between the two types. JavaScript only has a `number` type which is floating point under the hood.

I'll think about what to do here.

parlough:
> @parlough any clues as to why the result is different in DartPad vs terminal?

This is expected due to different number representations on the JS platform versus other ("native") platform. https://dart.dev/resources/language/number-representation has more information.

For this specific case, to avoid the potential for confusion and get the same results on both platforms, we could maybe drop the "is a double" portion and add an "is a num" section instead. Something like:

```
Object data = 123;
var typeInfo = [
  if (data case num i) 'Data is a num: $i',
  if (data case int i) 'Data is an integer: $i',
  if (data case String s) 'Data is a string: $s',
  if (data case bool b) 'Data is a boolean: $b',
]; // [Data is a num: 123, Data is an integer: 123]
print(typeInfo);
```

parlough:
I'm going to close this for now, but if you would like to explore another direction to avoid this discrepancy, feel free to reopen this PR or open a new one. Thanks again for noticing and working on this!

PR #6755, opened 2025-07-25
Add shadowing term to the glossary

IldySilva:

**Staged:** https://dart-dev--pr6755-main-ot10wqva.web.app/glossary/shadowing

### DIFF
--- a/src/_data/glossary.yml
+++ b/src/_data/glossary.yml
@@ -737,7 +737,8 @@
   short_description: |-
     When a local declaration hides another with the same name.
   long_description: |-
-    **Shadowing** occurs when a local declaration (such as a variable or parameter)
+    **Shadowing** occurs when a local declaration,
+    such as a variable or parameter,
     uses the same name as an existing declaration in an outer scope,
     making the outer one inaccessible within the inner scope.
 
@@ -747,8 +748,9 @@
     used deliberately to improve the clarity of your code.
 
     ### Example
-    
-    In this example, the local `message` inside `printMessage` **shadows** the global one:
+
+    In this example, the local `message` variable inside
+    the `printMessage` function **shadows** the top-level `message` variable:
 
     ```dart
     final message = 'Global';


antfitch:
@IldySilva this looks great to me. @parlough can you do a quick tech pass?

src/_data/glossary.yml
```diff
@@ -730,6 +730,55 @@
...
+      int value = 10;
+      if (true) {
+        int value = 20; // Shadows the outer `value`
+        print(value);   // Prints: 20
+      }
+      print(value);     // Prints: 10
+    }
+    ```
+
+    Although legal, excessive shadowing can reduce code readability and should be avoided in most cases.
```
parlough: It seems this mostly duplicates the earlier paragraph "Shadowing is valid in Dart...", I think it's fine to drop this one in favor of that paragraph.

src/_data/glossary.yml
```diff
@@ -730,6 +730,55 @@
...
+- term: "Shadowing"
+  short_description: |-
+    When a local declaration hides another with the same name.
+  long_description: |-
+    **Shadowing** occurs when a local declaration (such as a variable or parameter)
+    uses the same name as an existing declaration in an outer scope,
+    making the outer one inaccessible within the inner scope.
+
+    Shadowing is valid in Dart, but it can lead to confusing or unintended behavior,
+    so it’s generally discouraged unless used deliberately and clearly.
```
parlough: This sentence ends up being quite long. Consider breaking it into two, with the discouraged portion being pulled out. Perhaps something like:

```suggestion
    While valid in Dart, shadowing can lead to
    confusing code or unintended behavior.
    As a result, it's generally discouraged unless
    used deliberately to improve the clarity of your code.
```

src/_data/glossary.yml
```diff
@@ -730,6 +730,55 @@
...
+      print(message); // Prints: Local
+    }
+
+    void main() {
+      printMessage();
+      print(message); // Prints: Global
+    }
+    ```
+
+    In this example, the local `message` inside `printMessage` **shadows** the global one.
```
parlough: For an example introduction like this, move it to before the code snippet. So the reader has the context/intention in mind while they read the code.

src/_data/glossary.yml
```diff
@@ -730,6 +730,55 @@
...
+    uses the same name as an existing declaration in an outer scope,
+    making the outer one inaccessible within the inner scope.
+
+    Shadowing is valid in Dart, but it can lead to confusing or unintended behavior,
+    so it’s generally discouraged unless used deliberately and clearly.
+
+    ### Example
+
+    ```dart
+    String message = 'Global';
```
parlough: Unless a sample is specifically related to typing a variable, we prefer to use type inference for local variables as well as for top-level variables with relatively "[obvious](https://dart.dev/tools/linter-rules/omit_obvious_property_types)".

```suggestion
    final message = 'Global';
```

src/_data/glossary.yml
```diff
@@ -730,6 +730,55 @@
...
+
+    void printMessage() {
+      String message = 'Local'; // Shadows the global `message`
```
```suggestion
      final message = 'Local'; // Shadows the global `message` variable.
```

src/_data/glossary.yml
```diff
@@ -730,6 +730,55 @@
...
+
+    void main() {
+      printMessage();
+      print(message); // Prints: Global
+    }
+    ```
+
+    In this example, the local `message` inside `printMessage` **shadows** the global one.
+
+    Shadowing can also happen in nested blocks:
```
```suggestion
    Shadowing can also occur in nested blocks:
```

src/_data/glossary.yml
```diff
@@ -730,6 +730,55 @@
...
+
+    ```dart
+    void main() {
+      int value = 10;
```
```suggestion
      final value = 10;
```

src/_data/glossary.yml
```diff
@@ -730,6 +730,55 @@
...
+      if (true) {
+        int value = 20; // Shadows the outer `value`
```
```suggestion
        final value = 20; // Shadows the outer `value` variable.
```

parlough (commented):
Thanks @IldySilva! This will be a useful reference to be able to point to when discussing shadowing, such as in documentation for the avoid shadowing lints.

The content looks mostly good to me, just a few minor suggestions to consider. Let me know if you have any questions or concerns about them. Thanks again :)

IldySilva:
Also, do you know of any good reference we could add to related_links for shadowing? @parlough

parlough:
> Also, do you know of any good reference we could add to related_links for shadowing? @parlough

Hmm, great question. I don't think we have any specific docs related to shadowing variables. What do you think about linking to the [Variable shadowing](https://en.wikipedia.org/wiki/Variable_shadowing) Wikipedia article? It has some more details and examples that might be useful for developers. We could give the link a type of "external".

antfitch (approved):

IldySilva:
> > Also, do you know of any good reference we could add to related_links for shadowing? @parlough
> 
> Hmm, great question. I don't think we have any specific docs related to shadowing variables. What do you think about linking to the [Variable shadowing](https://en.wikipedia.org/wiki/Variable_shadowing) Wikipedia article? It has some more details and examples that might be useful for developers. We could give the link a type of "external".

Seems valid, but I think it would be redundant since the glossary already gives a good sense of what shadowing is. While it could be interesting to include, I don't think its absence causes any harm either.

parlough:
> Seems valid, but I think it would be redundant since the glossary already gives a good sense of what shadowing is. While it could be interesting to include, I don't think its absence causes any harm either.

Makes sense to me, thanks for being thoughtful about any additions! Landed 🚀 

Thanks again :)

Issue #6687, opened 2025-06-23
Dart 3.9 @awaitNotRequired annotation


antfitch:
Here is a new annotation that the analyzer can understand: @awaitNotRequired.

https://github.com/dart-lang/sdk/commit/b7cb6c3ed07a63f6ae7adb20683822ac3585b017 introduces the annotation, and has a good CHANGELOG entry for the meta package, and a good doc-comment in the code.

We need to add a link to this new annotation here:
https://dart.dev/language/metadata

Note:
It looks like we may be missing several built-in annotations.

parlough:
> It looks like we may be missing several built-in annotations.

One thing to note is that this page outlines annotations supported by the core Dart libraries, so the `package:meta` annotations have intentionally not been documented there previously.

Perhaps we can just add a note that links to `package:meta` and highlights the Dart analyzer provides valuable diagnostics when using the annotations it provides? As part of that, we can also link to its [API docs](https://pub.dev/documentation/meta/latest/meta/) which includes quite detailed reference docs for each annotation already, so we don't have to reduplicate content.

srawlins:
Thanks for working this @parlough ! That's a huge help! 😁

PR #6731, opened 2025-07-15
[dart 3.9.0] Add @awaitNotRequired annotation to docs

antfitch:
Adds `@awaitNotRequired` to the metadata docs.

Fixes #6687 

Notes:

@srawlins go ahead and review this first. Please feel free to update the files directly with your edits. Check the link I added for the new annotation, as I'm not sure it's correct. @parlough, once we have Sam's okay, this will be ready for you to review.

---

- [x] I’ve reviewed the contributor guide and applied the relevant portions to this PR.
- [x] This PR doesn't contain automatically generated corrections or text (Grammarly, LLMs, and similar).
- [x] This PR follows the [Google Developer Documentation Style Guidelines](https://developers.google.com/style) — for example, it doesn't use _i.e._ or _e.g._, and it avoids _I_ and _we_ (first person).
- [x] This PR uses [semantic line breaks](https://github.com/dart-lang/site-shared/blob/main/doc/writing-for-dart-and-flutter-websites.md#semantic-line-breaks) of 80 characters or fewer.

PR #6758, opened 2025-07-29
Overhaul the metadata and annotation language docs

parlough:
Overhaul the metadata page of the language docs, with a primary goal of including some mention of `package:meta`. I know technically `package:meta` isn't technically part of the language, but for developers reading these docs, the distinction isn't important enough to document this somewhere completely different. However, I did still try to clarify that the support around it is provided "by the Dart analyzer" rather than "by Dart".

- Integrate @antfitch's idea from https://github.com/dart-lang/site-www/pull/6731 of adding sections, particularly one focused on the built-in annotations.
- As began in https://github.com/dart-lang/site-www/pull/6731 by @antfitch, include a short definition and relative links about each of the built-in (`dart:core`) annotations.
- Document how to configure the supported targets of custom annotations using `package:meta/meta_meta.dart`.
- Add a new section on the (analyzer supported) annotations provided by `package:meta`, include some examples, and otherwise link to its (quite thorough) API docs.
- Take advantage of the ability to highlight relevant lines in example code blocks.

Resolves https://github.com/dart-lang/site-www/issues/6687

**Staged:** http://dart-dev--pr6758-feat-metadata-package-meta-325c4pbv.web.app/language/metadata

### DIFF
--- a/src/content/language/metadata.md
+++ b/src/content/language/metadata.md
@@ -127,11 +127,15 @@ void doSomething() {
 }
 ```
 
-To indicate the type of elements that should be annotated with your annotation,
-you can use the [`@Target`][] annotation from [`package:meta`][].
+### Specifying supported targets {:.no_toc}
 
-For example, if you wanted the above `@Todo` annotation to
-only be allowed on functions, you'd add the following annotation:
+To indicate the type of language constructs that
+should be annotated with your annotation,
+use the [`@Target`][] annotation from [`package:meta`][].
+
+For example, if you wanted the earlier `@Todo` annotation to
+only be allowed on functions and methods,
+you'd add the following annotation:
 
 <?code-excerpt "misc/lib/language_tour/metadata/todo.dart (target-kinds)"?>
 ```dart highlightLines=3
--- a/src/content/language/metadata.md
+++ b/src/content/language/metadata.md
@@ -71,6 +71,37 @@ members annotated with `@deprecated` or `@Deprecated`.
 [Extend a class]: /language/extend
 [Dart analyzer]: /tools/analysis
 
+## Analyzer-supported annotations
+
+Beyond providing support and analysis for the [built-in annotations][],
+the [Dart analyzer][] provides additional support and diagnostics for
+a variety of annotations from [`package:meta`][].
+Some commonly used annotations the package provides include:
+
+[`@visibleForTesting`][]
+: Marks a public member of a package as only public so
+  that the package can write tests for it.
+  The analyzer hides the member from autocompletion suggestions
+  and warns if it's used from another package.
+
+[`@awaitNotRequired`][]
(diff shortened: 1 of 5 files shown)


src/content/language/metadata.md
```diff
@@ -11,42 +10,73 @@ nextpage:
...
+
+[`@Deprecated`]: {{site.dart-api}}/dart-core/Deprecated-class.html
+[`@deprecated`]: {{site.dart-api}}/dart-core/deprecated-constant.html
+[`@override`]: {{site.dart-api}}/dart-core/override-constant.html
+[`@pragma`]: {{site.dart-api}}/dart-core/pragma-class.html
+[providing a deprecation message]: /tools/linter-rules/provide_deprecation_message
+[Extend a class]: /language/extend
+[Dart analyzer]: /tools/analysis
+
+## Custom annotations
```
srawlins, line 105: This section is so esoteric, can we move it below "## Analyzer supported annotations"? There are so few people that will ever need to use `@Target`, haha.

src/content/language/metadata.md
```diff
@@ -59,20 +89,62 @@ class Todo {
...
+}
+```
+
+With this configuration, the analyzer will warn if `Todo` is used as
+an annotation on any declaration besides a top-level function or method.
+
+[`@Target`]: {{site.pub-api}}/meta/latest/meta_meta/Target-class.html
+[`package:meta`]: {{site.pub-pkg}}/meta
+
+## Analyzer supported annotations
```
srawlins: I think this should be spelled "Analyzer-supported annotations"

src/content/language/metadata.md
```diff
@@ -59,20 +89,62 @@ class Todo {
...
+
+Beyond providing support and analysis for the [built-in annotations][],
+the [Dart analyzer][] provides additional support and diagnostics for
+a variety of annotations from [`package:meta`][].
+Some commonly used annotations the package provides include:
+
+[`@visibleForTesting`][]
+: Marks a public member of a package as only public so
+  that the package can write tests for it.
```
srawlins: Packages don't write tests. :P Maybe something like "so that the member can be accessed from the package's tests."

src/content/language/metadata.md
parlough: That makes sense, thanks for the suggestion!

Besides being a bit esoteric, after reordering the sections, it seems more natural to put the sections on pre-defined annotations next to each other anyway.

**Edit:** I also added a nested header for the section on target configuration, keeping the most relevant content on how annotations are defined a bit more focused.

src/content/language/metadata.md
parlough: > Packages don't write tests.

Unfortunately for us :)

Thanks for calling this out. I applied your suggestion.

PR #6771, opened 2025-08-05
Add Combinator term to glossary

IldySilva:
Contributes to https://github.com/dart-lang/site-www/issues/6461 by adding "Combinator"


src/_data/glossary.yml
```diff
@@ -64,6 +64,61 @@
...
+      // print(pi); // Error: pi is hidden.
+      print(Random()); // Accessible
+    }
+    ```
+
+    Combinators can also be chained:
+
+    ```dart
+    import 'dart:math' show pi, Random hide sqrt;
+    ```
```
eernstg: True, combinators in Dart can be chained. However, we might not want to spend the reader's time on considering this feature.

The [semantics of multiple combinators](https://github.com/dart-lang/language/blob/7893d138e927f6e397209aa63736de9d483a328f/specification/dartLangSpec.tex#L20632) in the same import or export directive is repeated applications of a subset operation (both `show` and `hide` transforms a namespace whose key set is _K1_ to a namespace whose key set is _K2_ such that _K2 ⊆ K1_). In the example above, 'dart:math' provides a specific namespace that has a binding for the keys `pi`, `Random`, and many others; after `show pi, Random` the namespace only has a binding for `pi` and `Random`; finally, `hide sqrt` makes no difference (the namespace still has a binding for `pi` and `Random`, and no other bindings).

In short, with `show x1, ... xk hide y1, .. ym`, the `hide` clause will have no effect except that it will remove the bindings for names which are members of `x1 .. xk` as well as `y1 .. ym`. But in that case you'd just have a `show` clause where you omit those names which are mentioned in both lists, and drop the `hide` clause entirely.

Similarly, with `hide x1, .. xk show y1, .. ym`, the `hide` clause has no effect unless some elements in `x1, .. xk` occur in `y1, .. ym` as well (and then you might just as well delete them from `y1, .. ym` and drop the `hide` clause).

So we shouldn't imply that chaining of combinators is a useful endeavor. ;-)

I pushed for changes to this semantics several years ago, but the proposal didn't get much support from the rest of the language team, probably because of worries about it being a breaking change.

In any case, it's more useful to focus on a single `show` clause or a single `hide` clause.

We could mention that chained combinators are possible, but useless, but I doubt that's going to be helpful for anyone.

eernstg (commented):
I added a drive-by comment, leaving the full review to the site-www team.

It's good to have this entry! 'Combinator' has a completely different meaning in other contexts (e.g., in functional languages). So there's a lot of potential confusion around this word, and this glossary entry could help with that.

src/_data/glossary.yml
IldySilva: @eernstg  Totally agree , no need to mention chaining if it doesn’t provide practical value.  thanks for bringing that up

antfitch:
@parlough, can you take a look?

parlough (approved):
This is a great entry @IldySilva! I agree that it will be useful to have an entry to point to for this less-often used Dart term. Thanks so much :D

I pushed a fix to the links and a few minor line break changes, but otherwise looks good to me.

Issue #6652, opened 2025-06-07
Define 'obviously typed' in glossary


eernstg:
### Page URL

https://dart.dev/resources/glossary/

### Page source

https://github.com/dart-lang/site-www/tree/main/src/content/resources/glossary.md

### Describe the problem

The lints `omit_obvious_local_variable_types`, `specify_nonobvious_local_variable_types`, `omit_obvious_property_types`, and `specify_nonobvious_property_types` all share the ability to characterize certain expressions as 'obviously typed'.

We should define this concept explicitly. It could perhaps be a glossary entry.

### Expected fix

Here is a definition of the concept:

### Obvious Expression Types

In order to have a mechanically decidable notion of what it takes for a type to be 'obvious' we need a precise definition. The point is that it should recognize a number of types that are indeed obvious, and then all other types are considered non-obvious. An expression `e` has an obvious type in the following cases:

- `e` is a non-collection literal. *For instance, `1`, `true`, `'Hello, $name!'`.*
- `e` is a collection literal with actual type arguments. *For instance, `<int, bool>{}`.*
- `e` is a list literal or a set literal where each element has an obvious type, and all elements have the same type. *For instance, `[1, 2]`, `{ [true, false], <bool>[] }`, but not `[1, 1.5]`.*
- `e` is a map literal where all key-value pair have a key with an obvious type and a value with an obvious type, and all keys have the same type, and all values have the same type. *For instance, `{ #a: <int>[] }`, but not `{1: 1, 2: true}`.*
- `e` is a record literal where every component has an obvious type. *For instance, `(1, enabled: true)`, but not `(1, enabled: foo())`.*
- `e` is a local variable that has not been promoted. *For instance, if `i` is a local variable declared as `var i = 1;` then `i` has an obvious type in the following code, as long as it hasn't been promoted.*
- `e` is a formal parameter that has not been promoted.
- `e` is an instance creation expression whose class part is not raw. *For instance `C(14)` if `C` is a non-generic class, or `C<int>(14)` if `C` accepts one type argument, but not `C(14)` if `C` accepts one or more type arguments.*
- `e` is a cascade whose target has an obvious type. *For instance, `1..isEven..isEven` has an obvious type because `1` has an obvious type.*
- `e` is a type cast. *For instance, `myComplexpression as int`.*
- `e` is a conditional expression where both branches have an obvious type, and they have the same type. *For instance, `condition ? 'yes' : 'no'`.*
[...]

PR #6795, opened 2025-08-14
fix: typo in `omit_obvious_local_variable_types` example

mockturtl:
Page URL: https://dart.dev/tools/linter-rules/omit_obvious_local_variable_types.html

https://github.com/dart-lang/site-www/tree/main/src/content/tools/linter-rules/individual-rules.md

---

 I believe the "GOOD" example in `omit_obvious_local_variable_types` should consider the loop variable `recipe in cookbook` as having an "obvious" type, and omit it.  See highlight:

Image

- I believe `recipe` qualifies as a _local variable that has not been promoted_, per #6652.  If I've misunderstood the definition of "obvious" and "not obvious", apologies -- the distinction is _(ahem)_ not obvious!

- This change is already present in docs for the related lint, [`omit_local_variable_types`](https://dart.dev/tools/linter-rules/omit_local_variable_types).

---


eernstg:
@mockturtl, thanks for your input! However, we should not land this PR.

Here's why:

The specification of what it means for an expression to have an 'obvious' type [here](https://github.com/dart-lang/site-www/issues/6652), and the actual behavior of `omit_obvious_local_variable_types` agree that an iteration variable like `recipe` does not get an obvious type based on whether or not the corresponding iterable has an obvious type. In other words, it isn't even sufficient that `iter` has an obvious type, `v` still doesn't have an obvious type in `for (final T v in iter) ...`. So `omit_obvious_local_variable_types` will not recommend that you make it `final v` rather than `final T v`.

In the given example, the iterable is `cookbook`. This is a constant variable which is declared outside the body of the current function, which means that `cookbook` has a non-obvious type anyway.

On the other hand, `omit_obvious_local_variable_types` won't complain if you remove the type annotation on `recipe`, because it will _never_ ask for a type annotation to be added, it will only ask for some type annotations to be removed. But if you enable `specify_nonobvious_local_variable_types` then it _will_ ask for a type annotation on `recipe`, because its type isn't considered to be obvious.

> I believe `recipe` qualifies as a local variable that has not been promoted

That is true, but this means that if you write `final v = recipe;` then `v` is considered to have an obvious type because `recipe` used as an expression is considered to have an obvious type. In contrast, the _declaration_ of `recipe` can have a type annotation (without complaints from `omit_obvious_local_variable_types`) because of the reason mentioned above (that is, "obviousness doesn't propagate from the iterable to the iteration variable in a `for` statement").

I clarified the commentary text on this item in https://github.com/dart-lang/site-www/issues/6652.

PR #6800, opened 2025-08-17
Add entry for native asset

IldySilva:
Contributes to https://github.com/dart-lang/site-www/issues/6461 by adding "Native asset"

**Staged:** https://dart-dev--pr6800-glossary-1-131apsrb.web.app/glossary/native-asset


src/_data/glossary.yml
```diff
@@ -546,6 +546,42 @@
     - "deferred initialization"
     - "late final"
 
+- term: "Native asset"
+  short_description: |-
+      A non-Dart resource, such as native code or data, that a Dart app can use.
```
dcharkes: We only used the term "native asset" for "code assets". Basically it's short for "native code asset".

Before we introduced the data assets, "native assets" implied build hooks, because that was the one feature for build hooks. Once we decided we want to introduce multiple asset types, we made the "build hook" a more prominent term.

I think it might be more useful too introduce in the glossary multiple terms:

* Hook
* Build hook
* [Link hook]
* [Asset]
* Code asset
* [Data asset (basically the same as "Asset" in Flutter, but reported from a build hook.)]

However, data assets and link hooks haven't fully shipped yet, so we can omit those for now. And maybe we then also omit "asset" for now. And potentially also "hook".

For "code asset" the alternatives can be "native code asset" and "native asset".

antfitch: @IldySilva and @dcharkes how does this sound for this particular PR:

Native asset:
Short for code asset. See code asset for details.

Native code asset:
Short for code asset. See code asset for details.

Code asset
Put the definition here.

I can add the other things to our glossary checklist that @parlough created.

IldySilva: Personally, I think having three separate entries where only one has the actual definition might make the glossary harder to use. Maybe it would be clearer to have a single Code asset entry with native asset and native code asset listed as alternates/aliases inside it. That way we keep everything centralized without duplicating or fragmenting the concept.

Also, it might help to add a short note in the Code asset entry mentioning that the term was previously referred to as native asset, so readers who come across older references still understand the connection.

antfitch: @IldySilva, I like the idea of adding the alternative aliases inside of the definition. Go for it.

IldySilva: **Code asset**
  A non-Dart resource, such as native code or data, that a Dart app can use.

**Short_description:** 
  Compiled native code that a Dart app can use through FFI or linking.

**long_description:** 
  A **code asset** is compiled code that isn't written in Dart but can be
  packaged and used by a Dart application. Examples include shared libraries
  (`.so`, `.dll`, `.dylib`) built in languages like C, C++, or Rust.

  Dart code can call into these assets using the [FFI](/guides/libraries/c-interop).

  > Historically, code assets were referred to as **native assets** or
  **native code assets**. These names are kept as alternates for clarity.

  Code assets are useful for:
  * Integrating with existing native libraries.
  * Accessing system capabilities not directly exposed in Dart.
  * Sharing performance-critical functionality across platforms.

  For example, a Dart application may include a native `.so` library
  that exposes specific functions, and call those functions from Dart.

**alternate:**
  - "Native asset"
  - "Native code asset"

IldySilva: @antfitch @dcharkes how does this look? Does it match what you had in mind?

antfitch: LGTM! @dcharkes, any changes?

dcharkes: Some edits:

 **Code asset**
  A non-Dart resource, such as native code or data, that a Dart app can use.

**Short_description:** 
  Compiled native code that is bundled with Dart app using a build hook can use through `dart:ffi`.

**long_description:** 
  A **code asset** is compiled code that isn't written in Dart but can be
  packaged and used by a Dart application. Examples include shared libraries
  (`.so`, `.dll`, `.dylib`) built in languages such as C, C++, or Rust.

  Code assets can be bundled with a Dart package using a [build hook] (add link to build hooks page on dart.dev).

  Dart code can call into these assets using the [FFI](/guides/libraries/c-interop).

  > Historically, code assets were referred to as **native assets** or
  **native code assets**. These names are kept as alternates for clarity.

  Code assets are useful for:
  * Integrating with existing native libraries.
  * Accessing system capabilities not directly exposed in Dart.
  * Sharing performance-critical functionality across platforms.

  For example, a Dart application may include a native `.so` library
  that exposes specific functions, and call those functions from Dart.

**alternate:**
  - "Native asset"
  - "Native code asset"

src/_data/glossary.yml
```diff
@@ -213,6 +213,43 @@
     - "language"
     - "type system"
 
+- term: "Code asset"
+  short_description: |-
+    Compiled native code that is bundled with Dart app using a build hook can use through `dart:ffi`.
```
dcharkes: can be used through `dart:ffi`

src/_data/glossary.yml
```diff
@@ -213,6 +213,43 @@
...
+  long_description: |-
+    A **code asset** is compiled code that isn't written in Dart but can be
+    packaged and used by a Dart application. Examples include shared libraries
+    (`.so`, `.dll`, `.dylib`) built in languages like C, C++, or Rust.
+
+    Code assets can be bundled with a Dart package using a  [build hook](/tools/hooks)
```
dcharkes: double space

src/_data/glossary.yml
```diff
@@ -797,6 +834,7 @@
   alternate:
     - "pubspec.lock"
 
+
```
dcharkes, line 842: unrelated edit

dcharkes (approved):

parlough (approved):
I pushed some minor formatting clean up and an update to an old c-interop link, but otherwise looks great to me. Thanks for working on this @IldySilva and thanks everyone else for the reviews :D

src/_data/glossary.yml
```diff
@@ -213,6 +213,47 @@
...
+    [FFI]: /interop/c-interop
+  related_links:
+    - text: "C interop using dart:ffi"
+      link: "/interop/c-interop/"
+      type: "doc"
+    - text: "Hooks"
+      link: "/tools/hooks"
+      type: "doc"
+  labels:
+    - "language"
```
parlough: On a second look through, I'm not sure the `language` label fits this one, as it's mostly a library and implementation feature.

Perhaps there's a new label we can put hook and asset related terms under? Maybe just "assets" or "build"? Do you have any thoughts?

dcharkes: * packages
* libraries
* interop

PR #6803, opened 2025-08-18
Fix for Misspelling and Bug in Code Sample

connie-ooi:
Fixed a spelling error and a bug in the code. 

Fixes https://dart.dev/get-started/add-commands

---

### DIFF
--- a/src/content/get-started/add-commands.md
+++ b/src/content/get-started/add-commands.md
@@ -294,12 +294,12 @@ null checks, and string interpolation.
 
     ```dart
     void searchWikipedia(List<String>? arguments) {
-      late String? articleTitle;
+      final String articleTitle;
 
       // If the user didn't pass in arguments, request an article title.
       if (arguments == null || arguments.isEmpty) {
         print('Please provide an article title.');
-        articleTitle = stdin.readLineSync(); // Await input from the user
+        articleTitle = stdin.readLineSync() ?? ''; // Await input from the user
       } else {
         // Otherwise, join the arguments into the CLI into a single string
         articleTitle = arguments.join(' ');
--- a/src/content/get-started/add-commands.md
+++ b/src/content/get-started/add-commands.md
@@ -212,7 +212,7 @@ null checks, and string interpolation.
        be null, even in production. The purpose of null-safety isn't
        to stop you from ever using null in your code, because
        representing the absense of a value can be valuable. Rather,
-       it's to force you to consider nullability and therefor be more
+       it's to force you to consider nullability and therefore be more
        careful about it. Along with the analyzer, this helps prevent
        one of the most common runtime crashes in programming:
        null-pointer errors.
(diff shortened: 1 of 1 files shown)


antfitch:
This looks great! Was there an issue for this in the Issues list? If yes, can you add that issue ID to the "Fixes" line in your description above? If not, great! Nothing else to do here.

src/content/get-started/add-commands.md
```diff
@@ -294,7 +294,7 @@ null checks, and string interpolation.
 
     ```dart
     void searchWikipedia(List<String>? arguments) {
-      late String articleTitle;
+      late String? articleTitle;
```
parlough: Nice catch! Since the type is nullable (`?`), the variable doesn't have to be `late` either, as it will just have an initial value of `null`.

```
      String? articleTitle;
```

It can even be `final` since it is definitely assigned by either the body from the `if` or the `else` :D

```suggestion
      final String? articleTitle;
```

parlough (approved):
Thanks @connie-ooi!!

I have one (optional) follow-up suggestion that can clean up the code further after your fix.

src/content/get-started/add-commands.md
eernstg: Drive-by comment:

I can understand if you think that the following change is too big, but it could make sense to keep the type and declare the variable as `final String articleTitle;`.

The point is that this allows the static analysis to detect that `articleTitle` will be a `String` (and it won't be null), which is in line with the actual code that follows, and it's useful to illustrate this "positive side" of sound null safety. We already have `arguments` showing how to allow null to be included.

Line 302 would then need to have something like `?? "";` at the end, which is yet another little hint about null safety (to be mentioned in line 330).

The print statement will work just fine when `articleTitle` is null, but with the type `String` we know that it won't be null, which could be seen as a useful sanity check.

This also illustrates that the flow analysis can detect that `articleTitle` is guaranteed to be initialized, no matter which path we're taking through this function body, which is a kind of check that we won't get if the variable is nullable (because it might just get initialized to null and then stay unchanged forever).

eernstg: I can't comment on unchanged lines, so I'll put it here:
Line 315 should be updated to have the actual declaration. Also, the following is not correct:

> late is a keyword that tells the analyzer that you, the programmer, promise that this variable won't be null by the time it's used

The late variable would be uninitialized which is not the same as being null. Anyway, it would need to be changed because the variable isn't late any more, but we could say something like this:

> The Dart static analysis proves that the variable articleTitle is guaranteed to be initialized when the print statement is executed.

Finally, a couple of other locations need to be updated to talk about `String` rather than `String?`.

eernstg (commented):
Added a couple of drive-by comments.

antfitch:
@connie-ooi, can you work with @eernstg to make those updates for him? When done, I would submit this.

src/content/get-started/add-commands.md
eernstg: Thanks!

antfitch:
Merged!!!

PR #6813, opened 2025-08-25
Adjust documentation of extension types: Cast/match does not run cons…

eernstg:
This PR extends the description of 'Type Considerations' in the section about extension types, in order to make it more explicit that type casts and pattern matching operations won't invoke a constructor, it just changes the static type based on the representation type of said extension type.

Written in response to https://github.com/dart-lang/sdk/issues/61343#issuecomment-3197242069.

---

<details>

</details>

### DIFF
--- a/src/content/language/extension-types.md
+++ b/src/content/language/extension-types.md
@@ -476,8 +476,8 @@ still refers to the same object, and `v` refers to the same object as `i`, and
 nothing has changed about this object. It is only the static type that changeed
 (and hence the methods that we can call on this object). In particular, the
 change of type does not involve an invocation of a constructor. If you wish to
-execute a constructor (e.g., to perform some kind of validation) then it is
-necessary to write an explicit constructor invocation (such as `NumberE(i)`).
+execute a constructor (for example, to perform some kind of validation) then it
+is necessary to write an explicit constructor invocation (such as `NumberE(i)`).
 
 It's important to be aware of this behavior when using extension types.
 Always keep in mind that an extension type exists and matters at compile time,
--- a/src/content/language/extension-types.md
+++ b/src/content/language/extension-types.md
@@ -477,7 +477,7 @@ nothing has changed about this object. It is only the static type that changeed
 (and hence the methods that we can call on this object). In particular, the
 change of type does not involve an invocation of a constructor. If you wish to
 execute a constructor (e.g., to perform some kind of validation) then it is
-necessary to write an explicit constructor invocation (such as `NumberE(1)`).
+necessary to write an explicit constructor invocation (such as `NumberE(i)`).
 
 It's important to be aware of this behavior when using extension types.
 Always keep in mind that an extension type exists and matters at compile time,
--- a/src/content/language/extension-types.md
+++ b/src/content/language/extension-types.md
@@ -473,7 +473,7 @@ void main() {
 
 When `i` gets the static type `NumberE` based on a cast or a pattern match, `i`
 still refers to the same object, and `v` refers to the same object as `i`, and
-nothing has changed about this object. It is only the static type that changeed
+nothing has changed about this object. It is only the static type that changed
 (and hence the methods that we can call on this object). In particular, the
 change of type does not involve an invocation of a constructor. If you wish to
 execute a constructor (for example, to perform some kind of validation) then it
--- a/src/content/language/extension-types.md
+++ b/src/content/language/extension-types.md
@@ -473,7 +473,7 @@ void main() {
 
 When `i` gets the static type `NumberE` based on a cast or a pattern match, `i`
 still refers to the same object, and `v` refers to the same object as `i`, and
-nothing has changed about this object. It is only the static type that changed
+the object itself remains unchanged. It is only the static type that changed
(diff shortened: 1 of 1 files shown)


src/content/language/extension-types.md
```diff
@@ -446,8 +446,8 @@ and when testing against an extension type (`case MyExtensionType(): ... `).
 void main() {
   var n = NumberE(1);
 
-  // Run-time type of 'n' is representation type 'int'.
-  if (n is int) print(n.value); // Prints 1.
+  // The run-time type of 'n' is the representation type 'int'.
+  if (n is int) print(n); // Prints 1.
```
eernstg, line 450: I changed `n.value` to `n` because it has the same behavior, and `n.value` raises a question which is slightly tricky to explain, and it's not the topic of this text.

The tricky part is that `n` is not promoted to `int` even though we have just tested that `n is int` and the result was `true`. This means that `n` still has the static type `NumberE` when we reach the expression `n.value`. This might very well be confusing to a reader who does not consider the fact that `int` is not a subtype of `NumberE`. A reader who thinks that the "promotion" has occurred will not understand why `n.value` isn't a compile-time error. With `print(n)`, the code works whether or not `n` has been promoted, so we just avoid talking about promotion entirely.

src/content/language/extension-types.md
```diff
@@ -446,8 +446,8 @@ and when testing against an extension type (`case MyExtensionType(): ... `).
 void main() {
   var n = NumberE(1);
 
-  // Run-time type of 'n' is representation type 'int'.
-  if (n is int) print(n.value); // Prints 1.
+  // The run-time type of 'n' is the representation type 'int'.
```
antfitch, line 449:
```suggestion
  // The runtime type of 'n' is the representation type 'int'.
```

src/content/language/extension-types.md
antfitch: Sounds good to me. @parlough any thoughts on this?

src/content/language/extension-types.md
```diff
@@ -471,13 +471,21 @@ void main() {
 }
 ```
 
-It's important to be aware of this quality when using extension types.
+When `i` gets the static type `NumberE` based on a cast or a pattern match, `i`
+still refers to the same object, and `v` refers to the same object as `i`, and
+nothing has changed about this object. It is only the static type that changed
```
antfitch:
```suggestion
the object itself remains unchanged. It is only the static type that changed
```

antfitch (approved):
I think this looks great.

src/content/language/extension-types.md
eernstg: I think it does make sense to keep using 'run-time' here.

Here are the three related words/phrases as I usually classify them: 'Run time' can be used as a noun, denoting the points in time where a given program is being executed ("the error is not detected statically, it only emerges at run time"). 'Run-time' can be used as an adjective, associating some entity with run time (".. so it's a run-time error"). 'Runtime' is a software entity which is used to support the execution of a program; it includes elements like primitive operations that can't be expressed in the language itself, memory management and garbage collection, and other support facilities (".. and the behavior whereby the stack is unrolled and an exception is thrown is handled by the runtime.").

I think it's a pretty standard choice to express noun phrases as separate words and adjectives with hyphens, so that one should not be controversial.

The choice to use 'runtime' as yet another word with its own meaning is not so easy to support with explicit dictionary definitions, but pages like https://learn.microsoft.com/en-us/dotnet/core/diagnostics/built-in-metrics-runtime, https://docs.rs/tokio/latest/tokio/runtime/struct.RuntimeMetrics.html, https://docs.oracle.com/en-us/iaas/jms/doc/java-runtime-metrics.html, and https://pkg.go.dev/runtime/metrics show that at least the .NET, Rust, Java, and Go communities do use 'runtime' to denote the entity that supports program execution (and they also mention garbage collection and similar tasks).

eernstg:
Thanks, PTAL!

src/content/language/extension-types.md
parlough: I think avoiding the promotion discussion here makes a lot of sense, thanks for this adjustment.

src/content/language/extension-types.md
parlough: I agree with Erik here and it's something I've been updating as I come across older, inconsistent usages. We can add this clarification to our own contribution docs.

src/content/language/extension-types.md
```diff
@@ -471,13 +471,21 @@ void main() {
...
 ```
 
-It's important to be aware of this quality when using extension types.
+When `i` gets the static type `NumberE` based on a cast or a pattern match, `i`
+still refers to the same object, and `v` refers to the same object as `i`, and
+the object itself remains unchanged. It is only the static type that changed
+(and hence the methods that we can call on this object). In particular, the
+change of type does not involve an invocation of a constructor. If you wish to
+execute a constructor (for example, to perform some kind of validation) then it
+is necessary to write an explicit constructor invocation (such as `NumberE(i)`).
```
parlough, line 480: Thanks for adding this additional clarification. I've seen incorrect assumptions around this behavior a few times, so I'm glad we call out it specifically now.

Issue #6798, opened 2025-08-15
wrong calculation


Subhamuni:
### Page URL

https://dart.dev/language/methods/

### Page source

https://github.com/dart-lang/site-www/tree/main/src/content/language/methods.md

### Describe the problem

double get bottom => top + height;
set bottom(double value) => top = value - height;

### Expected fix

double get bottom => top - height;
set bottom(double value) => top = value + height;

### I would like to fix this problem.

- [ ] I will try and fix this problem on dart.dev.

PR #6827, opened 2025-09-02
Clarify rectangle property calculation

parlough:
https://github.com/dart-lang/site-www/issues/6798 reported the calculation as wrong, but I believe it's right since we tend to document screen-space where `(0, 0)` is in the upper left and y grows down. To avoid any future misunderstandings, I've added a doc comment to the class to clarify this. I've also highlighted the import lines of the code snippet.

Closes https://github.com/dart-lang/site-www/issues/6798

**Staged:** https://dart-dev--pr6827-fix-6798-hwzhhasy.web.app/language/methods#getters-and-setters


connie-ooi (approved):

parlough:
Thanks for the review @connie-ooi!! I appreciate it :D

Issue #6492, opened 2025-03-17
Extends interface by class


morlang-alex:
### Page URL

https://dart.dev/language/class-modifiers#interface

### Describe the problem

Hello. Why did you write that exstends an interface by a class would be an error?

![Image](https://github.com/user-attachments/assets/ceec8560-fd19-467c-a5ca-b9a3f8b1e51e)

![Image](https://github.com/user-attachments/assets/18457cee-661b-4d8d-be7c-26465e49b743)

![Image](https://github.com/user-attachments/assets/21743bd4-ea3d-4956-86da-bf5d8b3f3908)

![Image](https://github.com/user-attachments/assets/8cc1d126-7bdf-4c0c-9cb6-09d11c75e9fb)

It works without any ERROR!

### I would like to fix this problem.

- [ ] I will try and fix this problem on dart.dev.

parlough:
Thanks for filing an issue about it!

This is because class modifier inheritance restrictions only apply outside the current library (usually Dart file). If you put `Car` in a separate Dart file, extending `Vehicle` would result in an error.

The following text at the beginning of the section aimed to clarify this:

> Libraries outside of the interface's own defining library can implement the interface, but not extend it.

Do you have any suggestions for how we could make this more clear here or in the code snippet?

PR #6829, opened 2025-09-02
Expand comments explaining class modifier errors

parlough:
Since the code snippets are a bit long and developer's have reported issues missing surrounding context, expand on the comments that call out errors related to using class-modifier annotated classes.

Resolves https://github.com/dart-lang/site-www/issues/6492

### DIFF
--- a/examples/language/lib/class_modifiers/ex5/sealed.dart
+++ b/examples/language/lib/class_modifiers/ex5/sealed.dart
@@ -1,4 +1,4 @@
-// ignore_for_file: instantiate_abstract_class,non_exhaustive_switch_expression
+// ignore_for_file: instantiate_abstract_class, non_exhaustive_switch_expression
 sealed class Vehicle {}
 
 class Car extends Vehicle {}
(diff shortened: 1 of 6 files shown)


examples/language/lib/class_modifiers/ex5/sealed.dart
```diff
@@ -7,16 +7,21 @@ class Truck implements Vehicle {}
...
-String getVehicleSound(Vehicle vehicle) {
-  // ERROR: The switch is missing the Bicycle subtype or a default case.
-  return switch (vehicle) {
-    Car() => 'vroom',
-    Truck() => 'VROOOOMM',
-  };
+extension VehicleSounds on Vehicle {
+  String get sound {
+    // ERROR: The switch does not exhaustively account for
+    // all possible instances of a `Vehicle`.
```
eernstg: It is always tricky to describe the situation where an object `o` has run-time type `B`, and `A` is some supertype of `B`. We don't usually say that `o` "is an instance of `A`" because that tends to mean that it has run-time type `A`. I often say that "`o` has type `A`" or "`o` is typable as an `A`". I'm then relying on the ability to claim that this implies that `A myA = o as dynamic;` will succeed, and hence `A` is one of the types that `o` "has". This might suffice:

```suggestion
    // all possible objects of type `Vehicle`.
```

examples/language/lib/class_modifiers/ex5/sealed.dart
```diff
@@ -7,16 +7,21 @@ class Truck implements Vehicle {}
...
+    // In this case, a `Vehicle` with a run-time type of `Bicycle`.
```
eernstg: "does not account for" might be too broad to unambiguously imply the non-match, but we could make it explicit:

```suggestion
    // In this case, a `Vehicle` with a run-time type of `Bicycle` will not
    // match any of the cases.
```

src/content/language/class-modifiers.md
```diff
@@ -289,18 +294,23 @@ class Truck implements Vehicle {}
...
-String getVehicleSound(Vehicle vehicle) {
-  // ERROR: The switch is missing the Bicycle subtype or a default case.
-  return switch (vehicle) {
-    Car() => 'vroom',
-    Truck() => 'VROOOOMM',
-  };
+extension VehicleSounds on Vehicle {
+  String get sound {
+    // ERROR: The switch does not exhaustively account for
+    // all possible instances of a `Vehicle`.
```
eernstg:
```suggestion
    // all possible objects of type `Vehicle`.
```

src/content/language/class-modifiers.md
```diff
@@ -289,18 +294,23 @@ class Truck implements Vehicle {}
...
+    // In this case, a `Vehicle` with a run-time type of `Bicycle`.
```
eernstg:
```suggestion
    // In this case, a `Vehicle` with a run-time type of `Bicycle` will not
    // match any of the cases.
```

parlough:
Thanks for the feedback and explanations @eernstg, I appreciate it. I applied your changes and made some additional minor changes to the sound example.

Issue #5477, opened 2024-01-19
Document getters and setters on the 'Functions' page


parlough:
### Page URL

https://dart.dev/language/functions.html

### Page source

https://github.com/dart-lang/site-www/tree/main/src/language/functions.md

### Describe the problem

We don't currently document getters and setters on this page, only on the [methods page](https://dart.dev/language/methods#getters-and-setters), but they are also allowed as top-level functions.

### Expected fix

Introduce a section to the functions page that documents getters and setters and perhaps modify the docs on the methods page if it make sense to?

CrazyRikin:
hey @parlough  ! is there someone working on this issue? if not then I would like to work on this issue.

PR #6830, opened 2025-09-02
add section for getters and setters

connie-ooi:
Added a section for getters and setters.

Fixes https://github.com/dart-lang/site-www/issues/5477

Preview:
https://dart-dev--pr6830-getters-setters-bcxd3yuk.web.app/language/functions#getters-and-setters

---

<details>

</details>


src/content/language/functions.md
```diff
@@ -620,3 +618,51 @@ is not `final`) an external setter.
 [conditional expression]: /language/operators#conditional-expressions
 [Flutter]: {{site.flutter}}
 [trailing commas]: /language/collections#lists
+
+## Getters and setters
+
+While getters and setters are very commonly used within classes, they are special functions that provide read and write access to an object's properties. Getters read and return the values of the properties. Setter functions take a parameter and assign it to the property. 
+
+Getters and setters control access to a property and are most useful when you want to protect or manage an internal private variable.
```
eernstg: This is perhaps a bit too narrow. Here's one way to give the reader a broader perspective on the topic:

Every property access (top-level, static, instance) amounts to an invocation of a getter or a setter. A variable implicitly induces a getter, and it may also induce a setter (if the variable is mutable). That's the reason why reading a property invokes a getter, and writing a property invokes a setter, even in the case where the property is declared to be a variable.

However, we can also declare a getter or a setter explicitly (using the keywords `get` or `set`). This allows the reading and/or writing of a property to be handled by a computation.

That's the whole point of having getters and setters: The client does not depend on whether the property is stored or computed, so the provider is free to make changes.

For example, the value of the property might not be stored anywhere, it could be computed each time the getter is called. Another example would be that the value is stored in a private variable, and public access occurs by calling a getter and/or a setter. In this case the getter and setter can be used to enforce that every read and write of the underlying variable is accompanied by some extra computations, e.g., validating or post-processing the value which is stored in the private variable:

[the `_secret` example]

eernstg (commented):
Added a drive-by comment.

src/content/language/functions.md
```diff
@@ -620,3 +618,55 @@ is not `final`) an external setter.
...
+
+However, you can also declare a getter or a setter explicitly (using the keywords `get` or `set`). This allows a property's value to be computed when it's read or written.
+
+The purpose of using getters and setters is to create a clear separation between the client (the code that uses the property) and the provider (the class that defines it). The client asks for or sets a value without needing to know if that value is stored in a simple variable or calculated on the spot. This gives the provider the freedom to change how the property works. 
+
+For example, the value of the property might not be stored anywhere, it could be computed each time the getter is called. Another example is when a value is stored in a private variable, and public access is only allowed by calling a getter or a setter. The example below showcases this.
```
antfitch:
```suggestion
For example, because the value of the property might not be stored anywhere, it could be computed each time the getter is called. Another example is that when a value is stored in a private variable, public access is only allowed by calling a getter or a setter. The example below showcases this.
```

src/content/language/functions.md
```diff
@@ -620,3 +618,55 @@ is not `final`) an external setter.
...
+
+```dart
+/// Defines a private variable secret with `_`.
+String _secret = 'Hello';
+
+/// The public top-level GETTER.
```
antfitch: Let's change `GETTER` to `getter` and `SETTER` to `setter` in the sample to be consistent with style, although I get why this was uppercased.

src/content/language/functions.md
```diff
@@ -620,3 +618,55 @@ is not `final`) an external setter.
...
+  print('Setter was used! New secret: "$newMessage"');
+  if (newMessage.isNotEmpty) {
+    _secret = newMessage;
+  }
+}
+
+void main() {
+  // Reading the value calls the GETTER.
+  print('Current message: ${secret}');
+  // Output:
```
antfitch: I wonder if this might look better since it's meant for multi-line comments:

/*
Output:
Getter was used!
Current message: HELLO
*/

I would test and see what the preview looks like. If it looks good, I'd swap this and the other outputs, otherwise what you're using should be fine. I think this is a something we probably need to document in our documentation guide. I'll look into that.

src/content/language/functions.md
```diff
@@ -620,3 +618,55 @@ is not `final`) an external setter.
 [conditional expression]: /language/operators#conditional-expressions
 [Flutter]: {{site.flutter}}
 [trailing commas]: /language/collections#lists
+
+## Getters and setters
```
antfitch: This section looks great. Can you go through the markdown file and make sure that all lines are under 80 characters? There should be a setting in VS Code that let's you track when something goes over that line. If you need help finding it, let me know.

antfitch (changes requested):
A few small nits. Looks awesome!

src/content/language/functions.md
```diff
@@ -620,3 +619,77 @@ is not `final`) an external setter.
...
+simple variable or calculated on the spot. This gives the provider
+the freedom to change how the property works. 
+
+For example, because the value of the property might not be stored anywhere,
+it could be computed each time the getter is called. Another example
+is that when a value is stored in a private variable, and public access
+is only allowed by calling a getter or a setter.
+The example below showcases this.
+
+```dart
```
antfitch: Your code sample looks great! Next step is to move this into a code excerpt. So in Dart docs, most code snippets need to be added to our excerpt library. When in this library, we have a program that checks to make sure our excerpts always run. It's essentially the way we make sure that everything we publish runs.

Take a look at the code snippets above this one. You'll notice they all have a line that looks similar to this:

```
<?code-excerpt "misc/test/language_tour/async_test.dart (async-generator)"?>
```

At the location above, lives a code excerpt called `async_test.dart` that has a tiny Dart program in it with the code excerpt inside of it. To learn how we do this, start here: [Dart documentation guide - Code Samples](https://github.com/dart-lang/site-www/wiki/Documentation-guide#code-samples)

Now doing this for the first time can be a bit confusing, so do you best to figure how to make this work and if you get stuck ping @parlough or myself. @parlough knows the system inside and out.

antfitch (approved):
LGTM from me! @parlough, do you have any changes before this gets submitted?

examples/language/lib/functions/getters_setters.dart
```diff
@@ -0,0 +1,42 @@
+/// Defines a private variable secret with `_`.
```
parlough: I got slightly confused reading this comment as I wasn't sure if "secret" was referring to the variable name or the visibility of the variable.

Perhaps we can be more explicit by referencing the full name of the variable and add more details about the underscore?

Would something like this capture your intent?

```
// Defines a variable `_secret` that is private to the library since
// its identifier starts with an underscore (`_`).
```

examples/language/lib/functions/getters_setters.dart
```diff
@@ -0,0 +1,42 @@
...
+String _secret = 'Hello';
+
+/// The public top-level getter.
+/// It provides read access to the [_secret].
```
parlough: Since this is more intended as context for understanding the code snippet rather than how [doc comments](https://dart.dev/language/comments#documentation-comments) should be written, consider converting this to normal [inline comments](https://dart.dev/language/comments#single-line-comments).

Perhaps similar to this:

```dart
// A public top-level getter that
// provides read access to [_secret].
```

examples/language/lib/functions/getters_setters.dart
```diff
@@ -0,0 +1,42 @@
...
+String get secret {
+  print('Getter was used!');
+  return _secret.toUpperCase();
+}
+
+/// The public top-level setter.
+/// It provides write access to the [_secret].
```
parlough: Similar to my comment on the getter, consider making this an inline comment:

```
// A public top-level setter that
// provides write access to [_secret].
```

examples/language/lib/functions/getters_setters.dart
```diff
@@ -0,0 +1,42 @@
...
+set secret(String newMessage) {
+  print('Setter was used! New secret: "$newMessage"');
+  if (newMessage.isNotEmpty) {
+    _secret = newMessage;
+  }
```
parlough: Good example! However, should the "New secret: ..." portion be printed out separately in the if statement? As otherwise it's not the new secret.

```
  print('Setter was used!');
  if (newMessage.isNotEmpty) {
    _secret = newMessage;
    print('New secret: "$newMessage"');
  }
```

examples/language/lib/functions/getters_setters.dart
```diff
@@ -0,0 +1,42 @@
...
+  /*
+  Output:
+  Getter was used!
+  Current message: HELLO
+  */
+
+  // Assigning a value calls the setter.
+  secret = 'Dart is fun';
+
+  // Reading it again calls the getter to show the new value.
```
parlough: Since it's not just the stored `_secret`, but one made uppercase, would it be worth clarifying that by staying "new computed value" instead of just "new value"?

src/content/language/functions.md
```diff
@@ -620,3 +619,76 @@ is not `final`) an external setter.
 [conditional expression]: /language/operators#conditional-expressions
 [Flutter]: {{site.flutter}}
 [trailing commas]: /language/collections#lists
+
+## Getters and setters
```
parlough: I'd consider moving this section before the preceding "External functions" section as that is a much more niche feature. Perhaps above the generators section even.

src/content/language/functions.md
```diff
@@ -620,3 +619,76 @@ is not `final`) an external setter.
...
+
```
parlough: Don't worry about it for this PR, but for future changes, try to break up your lines of Markdown with [Semantic breaks](https://sembr.org/). Semantic breaks allow easier review, editing, and improves diffs and change history if the text is modified in the future.

Happy to discuss more about semantic breaks if you'd like and Shams is a great resource for understanding them as well!

src/content/language/functions.md
```diff
@@ -620,3 +619,76 @@ is not `final`) an external setter.
...
+Every property access (top-level, static, or instance) is an invocation 
+of a getter or a setter. A variable implicitly creates a getter 
+and, if it's mutable, a setter. This is why when you access a property, 
+you're actually calling a small function in the background. Reading a 
+property calls a getter function, and writing one calls a setter function,
+even in cases where the property is declared a variable. 
+
+However, you can also declare a getter or a setter explicitly (using
+the keywords `get` or `set`). This allows a property's value to be
+computed when it's read or written.
```
parlough: Since `get` and `set` are a primary focus of this text, would it make sense to take their mention out of the parenthetical? Or do you intentionally not want to introduce them here?

Depends on your intentions, but perhaps something like:

```
However, you can also declare getters and setters explicitly with
the `get` and `set` keywords respectively.
This allows a property's value to be computed when it's read or written.
```

src/content/language/functions.md
```diff
@@ -620,3 +619,76 @@ is not `final`) an external setter.
...
+
+The purpose of using getters and setters is to create a clear
+separation between the client (the code that uses the property) and
+the provider (the class that defines it). The client asks for or
```
parlough: Since getters and setters are also supported at the top level of a library, let's clarify that:

```suggestion
the provider (the class or library that defines it). The client asks for or
```

src/content/language/functions.md
```diff
@@ -620,3 +619,76 @@ is not `final`) an external setter.
...
+and, if it's mutable, a setter. This is why when you access a property, 
+you're actually calling a small function in the background. Reading a 
+property calls a getter function, and writing one calls a setter function,
+even in cases where the property is declared a variable. 
+
+However, you can also declare a getter or a setter explicitly (using
+the keywords `get` or `set`). This allows a property's value to be
+computed when it's read or written.
+
+The purpose of using getters and setters is to create a clear
```
parlough: This paragraph is a great explanation!

src/content/language/functions.md
```diff
@@ -620,3 +619,76 @@ is not `final`) an external setter.
...
+separation between the client (the code that uses the property) and
+the provider (the class that defines it). The client asks for or
+sets a value without needing to know if that value is stored in a
+simple variable or calculated on the spot. This gives the provider
+the freedom to change how the property works. 
+
+For example, because the value of the property might not be stored anywhere,
```
parlough: I like that both these use cases are mentioned here, thanks!

src/content/language/functions.md
```diff
@@ -620,3 +619,76 @@ is not `final`) an external setter.
...
+it could be computed each time the getter is called. Another example
+is that when a value is stored in a private variable, and public access
+is only allowed by calling a getter or a setter.
+The example below showcases this.
```
parlough: The example is great because it showcases a sort of combination of both reasonings mentioned here.

Optionally consider providing a more detailed introduction though so developers know what to look for when reading the code snippet.

If you think that'd be valuable, perhaps something like this?

```
The following example showcases this,
with the `secret` getter and setter providing
indirect access to the private variable `_secret` with
its own manipulations on the assigned and retrieved values.
```

If that's too long, perhaps "manipulated" can modify "access" instead of being an extra phrase at the end.

```
The following example showcases this,
with the `secret` getter and setter providing
manipulated access to the private variable `_secret`.
```

src/content/language/functions.md
```diff
@@ -620,3 +619,76 @@ is not `final`) an external setter.
...
+
+<?code-excerpt "language/lib/functions/getters_setters.dart"?>
+```dart
```
parlough: As a question for you, do you think it'd be valuable if any of the lines in the code snippet are highlighted? If so, which ones?

parlough (approved):
This is great @connie-ooi, thanks so much! This is definitely an addition we needed. While we did document getter and setter instance methods a tiny bit, we never provided much detail or explanation on the "why" of them. So I'm quite excited for developers to benefit from this extra context.

The change looks good to me already, but I have some small comments and questions that might be worth considering before landing or at least for the future. Do feel free to ignore or push back on any of my comments.

src/content/language/functions.md
conooi: Thanks Parker! Would love to discuss more about Semantic breaks. I tried breaking it up but maybe was not successful at it.

src/content/language/functions.md
conooi: Great question! I would highlight lines 574-577, 588-587, 591, and 600. They show the getter and setter declaration and their usage.

PR #6832, opened 2025-09-03
Use stable 3.9 version for SDK constraint

parlough:
Also update the validation of the installed Dart version in the `dart_site` script to properly check for 3.9 instead of 3.7.


antfitch (approved):
lgtm!

Issue #6851, opened 2025-09-05
Missing navigation from the list of diagnostics to the pages with details


bwilkerson:
### Page URL

https://dart.dev/tools/diagnostics

### Page source

https://github.com/dart-lang/site-www/blob/main/src/content/tools/diagnostics/index.md

### Describe the problem

The text at the top of the page says "If a diagnostic has extra documentation and guidance, click Learn more to view it.", but there are no Learn more buttons on any of the diagnostic bubbles.

### Expected fix

We either need to generate "Learn more" buttons or we need to change the text and implement another way to navigate.

### I would like to fix this problem.

- [ ] I will try and fix this problem on dart.dev.

parlough:
Thanks for catching this @bwilkerson. It's a recent regression from https://github.com/dart-lang/site-www/commit/fd943b8add36fe5eaa55b594ea6a8501ae667407. I'll prepare a fix.

PR #6852, opened 2025-09-05
Fix missing "Learn more" buttons on diagnostic index page

parlough:
Fixes the "Learn more" link buttons not being added to Diagnostic cards on the diagnostic index page.

The `_DiagnosticInfo` extension type assumed the values were strings instead of booleans, causing the "Learn more" buttons not to be rendered even if the `hasDocumentation` value was set to `true`.

Fixes https://github.com/dart-lang/site-www/issues/6851

**Staged:** https://dart-dev--pr6852-fix-6851-2lf4ba4i.web.app/tools/diagnostics#diagnostic-index

### DIFF
--- a/src/content/tools/diagnostics/deprecated_mixin.md
+++ b/src/content/tools/diagnostics/deprecated_mixin.md
@@ -1,0 +1,45 @@
+---
+title: deprecated_mixin
+description: >-
+  Details about the deprecated_mixin
+  diagnostic produced by the Dart analyzer.
+underscore_breaker_titles: true
+bodyClass: highlight-diagnostics
+---
+
+_Mixing in '{0}' is deprecated._
+
+## Description
+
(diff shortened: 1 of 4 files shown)


src/content/tools/diagnostics/deprecated_mixin.md
antfitch, line 1: @parlough, the deprecated annotation is a project @connie-ooi is assigned to work on. Is that this?

parlough: This diagnostic is related, yes. It's one of many that can be triggered by using an element annotated with one of the new annotations in the deprecated fashion. This page is autogenerated from the docs Sam wrote for these diagnostics in the SDK.

Docs about the annotations and their use is separate and still needed.

antfitch: Ahhh, got it!

PR #6857, opened 2025-09-07
Fix Typo in Monte Carlo method example 

### LINKED ISSUE #6856: Typo in Monte Carlo method example
### Page URL

https://dart.dev/overview

### Page source

https://github.com/dart-lang/site-www/blob/main/src/content/overview.md

examples/misc/lib/overview_pi.dart

### Describe the problem

Mistake in the comment from the Monte Carlo method

```
// So, when given random points with x ∈ <0,1>,
// y ∈ <0,1>, the ratio of those inside a unit circle
// should approach π / 4.
```

### Expected fix

The Monte Carlo ratio formula assumes a full circle inscribed in a square centered at the origin, with x and y values ranging from -1 to 1.

```
// So, when given random points with x ∈ <-1,1>,
// y ∈ <-1,1>, the ratio of those inside a unit circle
// should approach π / 4.
```

### Additional context

_No response_

johanremilien:
The Monte Carlo ratio formula assumes a full circle inscribed in a square centered at the origin, with x and y values ranging from -1 to 1.

Fixes #6856

---

### DIFF
--- a/src/content/overview.md
+++ b/src/content/overview.md
@@ -90,8 +90,10 @@ Stream<double> computePi({int batch = 100000}) async* {
     final ratio = count / total;
 
     // Area of a circle is A = π⋅r², therefore π = A/r².
-    // So, when given random points with x ∈ <-1,1>,
-    // y ∈ <-1,1>, the ratio of those inside a unit circle
+    // We consider only non-negative x and y (that is, the
+    // first quadrant), which doesn't change the ratio.
+    // So, when given random points with x ∈ [0, 1],
+    // y ∈ [0, 1], the ratio of those inside the unit circle
     // should approach π / 4. Therefore, the value of π
     // should be:
     yield ratio * 4;


antfitch:
@eernstg can you double-check or triage this?

examples/misc/lib/overview_pi.dart
```diff
@@ -20,8 +20,8 @@ Stream<double> computePi({int batch = 100000}) async* {
     final ratio = count / total;
 
     // Area of a circle is A = π⋅r², therefore π = A/r².
-    // So, when given random points with x ∈ <0,1>,
-    // y ∈ <0,1>, the ratio of those inside a unit circle
+    // So, when given random points with x ∈ <-1,1>,
+    // y ∈ <-1,1>, the ratio of those inside a unit circle
```
eernstg: The explanation could have been more explicit about this, but the range from 0.0 to (but not including) 1.0 is actually correct: This is the range of values that `random.nextDouble()` can return.

This means that we're computing the ratio of random points _in the first quadrant and inside the unit circle_ against all observed points in the first quadrant and inside the "unit square" (where the absolute value of the coordinates is less than 1.0). It doesn't matter that we're only looking at the first quadrant because it's a ratio &mdash; if we had been looking at all quadrants then we'd just have approximately 4 times as many points on both sides of the ratio.

The notation used to specify the interval is `<0, 1>`, which is not common. A notation like `[0; 1[` (where the first left bracket means "from and including" and the second left bracket means "to, but not including") would be more common. However, the use of `[` as a kind of "right bracket" is probably so weird in a software context that we shouldn't use it. Another rather common notation is `[0; 1)`, but that's also quite weird in the context of source code. In the end, including / not including makes no difference for the computed value.

(Just checking, and https://mathresearch.utsa.edu/wiki/index.php?title=Interval_Notation#:~:text=An%20interval%20is%20said%20to%20be%20left%2Dopen%20if%20and,base%20of%20the%20open%20sets does mention both of these notations.)

Another difficulty is that `[text]` in a DartDoc comment is used to refer to an entity by name in a location where that name is in scope. However, I don't think this makes a difference in any way when the comment is not a DartDoc comment.

Hence, we could probably use this notation: `x ∈ [0,1]`.

```suggestion
    // We consider only non-negative x and y (that is, the
    // first quadrant), which doesn't change the ratio.
    // So, when given random points with x ∈ [0, 1],
    // y ∈ [0, 1], the ratio of those inside the unit circle
```

src/content/overview.md
```diff
@@ -90,8 +90,8 @@ Stream<double> computePi({int batch = 100000}) async* {
     final ratio = count / total;
 
     // Area of a circle is A = π⋅r², therefore π = A/r².
-    // So, when given random points with x ∈ <0,1>,
-    // y ∈ <0,1>, the ratio of those inside a unit circle
+    // So, when given random points with x ∈ <-1,1>,
+    // y ∈ <-1,1>, the ratio of those inside a unit circle
```
eernstg:
```suggestion
    // We consider only non-negative x and y (that is, the
    // first quadrant), which doesn't change the ratio.
    // So, when given random points with x ∈ [0, 1],
    // y ∈ [0, 1], the ratio of those inside the unit circle
```

eernstg (approved):
LGTM, with a comment.

examples/misc/lib/overview_pi.dart
johanremilien: Exactly! Makes perfect sense.

johanremilien:
@eernstg Thank you for your time and insightful review 🎯

antfitch (approved):

Issue #6839, opened 2025-09-03
Simplify sidenav active entry determination


parlough:
The sidenav active entry determination is a bit complicated due to working around performance issues with previous site infrastructures. The Jaspr migration directly converted the logic, but now that's it written in Dart, we should simplify it or at least clean it up.

Issue #6853, opened 2025-09-05
☔ Finalize, clean up, and document Jaspr version of site infrastructure


parlough:

schultek:
@parlough 
> Integrate [inject_dartpad](https://github.com/dart-lang/site-shared/tree/main/pkgs/inject_dartpad) logic directly with Jaspr setup rather than a separately compiled script. (@parlough)

Did you already start with this? Else I could take this.

parlough:
> Did you already start with this? Else I could take this.

@schultek I started it but haven't gotten around to finishing it. If you want to take it over, definitely feel free. If it helps, I pushed my original explorations to https://github.com/dart-lang/site-www/tree/feat/injected-dartpad-replace.

parlough:
The core tasks here have all been completed, with the rest being enhancement work that is now possible. Those tasks are tracked in separate issues and projects already.

Thanks @schultek for helping me finish this off!

PR #6858, opened 2025-09-08
Clean up sidenav implementation and encapsulate active nav logic

parlough:
This PR cleans up the implementation for the sidenav while making minimal functional changes. The primary changes include:

- Move the sidenav data representation to a separate model file.
- Breaks the sidenav implementation into a few components.
- Encapsulate the logic for calculating and using the active nav highlighting path with a new `_ActiveNavigationPath` class.
- Avoid some computations when they weren't needed.
- Only functional change is fix for sidenav offset from navbar to be correct when there is no TOC.

Resolves https://github.com/dart-lang/site-www/issues/6839
Contributes to https://github.com/dart-lang/site-www/issues/6853


antfitch:
@redbrogdon have time to do a quick code review?

parlough:
Thanks for the review :D

As this PR is mostly just clean up and reorganization, I'm going to merge it to unblock and simplify some other work. Andrew (or others), if you'd still like to review or leave comments, please do and I'll open up a follow-up PR with those adjustments. Thanks!

PR #6892, opened 2025-09-23
Add "Tree shaking" entry to glossary

IldySilva:
Adds "tree shaking" term to the site-wide glossary that can be referenced when explaining the behavior of Dart's various compilers.

### DIFF
--- a/src/data/glossary.yml
+++ b/src/data/glossary.yml
@@ -1424,15 +1424,15 @@
   short_description: |-
     A compiler optimization that removes unused code.
   long_description: |-
-    **Tree shaking** is a compiler optimization that analyzes which parts of
-    the code are actually used by an application and removes everything else
-    from the final output.
+    **Tree shaking** is a compiler optimization that
+    analyzes which parts of the code are actually used by an application and
+    removes everything else from the final output.
 
     This makes the compiled program smaller and faster to load, since dead code
     that is never referenced doesn't end up in the binary or bundle.
 
-    Tree shaking is especially important for web and mobile apps, where code size
-    directly impacts download time and runtime performance.
+    Tree shaking is especially important for web and mobile apps, where
+    code size directly impacts download time and runtime performance.
 
     For example, if you import a library that defines ten top-level functions,
     but your app only calls two, tree shaking ensures


src/data/glossary.yml
```diff
@@ -1420,6 +1420,43 @@
...
+    from the final output.
+
+    This makes the compiled program smaller and faster to load, since dead code
+    that is never referenced doesn't end up in the binary or bundle.
+
+    Tree shaking is especially important for web and mobile apps, where code size
+    directly impacts download time and runtime performance.
+
+    Example:
+    If an imported library defines 10 classes but your app only uses 2,
```
antfitch:
```suggestion
    If an imported library defines ten classes but your app only uses two,
```

src/data/glossary.yml
```diff
@@ -1420,6 +1420,43 @@
...
+    tree shaking ensures the other 8 classes are excluded from the compiled output.
```
antfitch:
```suggestion
    tree shaking ensures the other eight classes are excluded from the compiled output.
```

antfitch:
Looks great. @parlough, want to do a tech review?

src/data/glossary.yml
```diff
@@ -1420,6 +1420,43 @@
...
+      link: "/language/libraries"
+      type: "doc"
+    - text:  "Web deployment"
+      link:  "/web/deployment"
+      type: "doc"
+
+  labels:
+    - "language"
+    - "compiler"
+    - "optimization"
```
parlough, line 1453: Nice new label! I can imagine a few other future terms that could be grouped under it as well.

src/data/glossary.yml
```diff
@@ -1420,6 +1420,43 @@
...
+  alternate:
+    - "Dead code elimination"
```
parlough: To match the syntax used in https://en.wikipedia.org/wiki/Dead-code_elimination, let's hyphenate the descriptor.

```suggestion
    - "Dead-code elimination"
```

src/data/glossary.yml
```diff
@@ -1420,6 +1420,43 @@
...
+    - text: "Libraries & imports"
+      link: "/language/libraries"
+      type: "doc"
+    - text:  "Web deployment"
+      link:  "/web/deployment"
+      type: "doc"
+
+  labels:
+    - "language"
+    - "compiler"
```
parlough: To match some of the other labels, let's use the [nominalized](https://en.wikipedia.org/wiki/Nominalization) form of this.

```suggestion
    - "compilation"
```

src/data/glossary.yml
```diff
@@ -1420,6 +1420,43 @@
...
+
+    This makes the compiled program smaller and faster to load, since dead code
+    that is never referenced doesn't end up in the binary or bundle.
+
+    Tree shaking is especially important for web and mobile apps, where code size
+    directly impacts download time and runtime performance.
+
+    Example:
+    If an imported library defines ten classes but your app only uses two,
+    tree shaking ensures the other eight classes are excluded from the compiled output.
```
parlough: I love that you included this example!

I think it's a bit hard to understand how (and when) classes are "included" in compiled output though, so perhaps it'd be better to replace the mention with functions? Developers often have some familiarity with that from imperative logic.

```
    For example, if you import a library that defines ten top-level functions,
    but your app only calls two, tree shaking ensures
    the other eight functions are excluded from the compiled output.
```

parlough (approved):
Love it!! We had a bunch of lose mentions of "tree shaking" but no coherent location.

I do think the term is more familiar to developers now, but not everyone, so it'll be great to have this entry.

Thanks @IldySilva!

parlough:
Thanks for making those adjustments! 🚀

PR #6893, opened 2025-09-23
Add "Closurization" entry to glossary

IldySilva:
Adds "closurization" term to the site-wide glossary.


src/data/glossary.yml
```diff
@@ -141,6 +141,54 @@
...
+
+    void main() {
+      var greeter = Greeter();
+
+      // Tear-off of the instance method `greet`
+      var sayHello = greeter.greet;
+
+      sayHello('Dart'); // Prints: Hello, Dart!
+    }
+    ```
```
antfitch, line 174:
```suggestion
    ```

```

src/data/glossary.yml
```diff
@@ -141,6 +141,54 @@
...
+    In this example, `greeter.greet` is closurized into a closure that
+    captures the instance `greeter`. When `sayHello` is called, it executes
+    as if you had called `greeter.greet(...)` directly.
+
+    Closurization ensures that tear-offs are **first-class function values**
```
antfitch: I'd delete `are **first-class function values** in Dart and `

src/data/glossary.yml
```diff
@@ -141,6 +141,54 @@
...
+    in Dart and can be used wherever a function is expected.
+  related_links:
```
antfitch, line 183: space before this line?

antfitch:
Looks great. @parlough, want to do a tech review?

src/data/glossary.yml
IldySilva: Not totally sure why it’s like that, but the rest of the glossary keeps it this way , probably just to keep things compact. I’ll stick with the same pattern here, and if we ever change it we can do it across all entries.

src/data/glossary.yml
IldySilva: Ah, right ,  a bit redundant 😁, I’ll update

src/data/glossary.yml
```diff
@@ -141,6 +141,54 @@
     - "Function callback"
     - "Callback function"
 
+- term: "Closurization"
+  short_description: |-
+    The process of turning a method or function tear-off into a closure.
+  long_description: |-
+    **Closurization** is the process where Dart turns a method or function
```
parlough: To help better distinguish these two cases, let's specify "instance method" since "method" is sometimes used to reference static methods as well.

```suggestion
    **Closurization** is the process where Dart turns a instance method or function
```

src/data/glossary.yml
```diff
@@ -141,6 +141,54 @@
...
+    tear-off into a closure object that can be stored, passed, or invoked later.
```
parlough: At least to Dart developers, there's not really a concept of a "closure object", so I think it might be better to stick with "function object".

```suggestion
    tear-off into a function object that can be stored, passed, or invoked later.
```

src/data/glossary.yml
```diff
@@ -141,6 +141,54 @@
...
+
+    A tear-off is a reference to an existing function or method without calling it.
+    When you take a tear-off, the runtime (or compiler) wraps it into a closure
```
parlough: Across the docs we tend to use the verbiage "tear off a function".

```suggestion
    When you tear off a function, the runtime (or compiler) wraps it into a closure
```

src/data/glossary.yml
```diff
@@ -141,6 +141,54 @@
...
+    class Greeter {
+      void greet(String name) {
+        print('Hello, $name!');
+      }
+    }
+
+    void main() {
+      var greeter = Greeter();
+
+      // Tear-off of the instance method `greet`
```
```suggestion
      // Tear-off of the instance method `greet`.
```

src/data/glossary.yml
```diff
@@ -141,6 +141,54 @@
...
+- term: "Closurization"
+  short_description: |-
+    The process of turning a method or function tear-off into a closure.
+  long_description: |-
+    **Closurization** is the process where Dart turns a method or function
+    tear-off into a closure object that can be stored, passed, or invoked later.
+
+    A tear-off is a reference to an existing function or method without calling it.
+    When you take a tear-off, the runtime (or compiler) wraps it into a closure
+    so that it behaves like any other function value in Dart.
```
parlough: I don't think we use the term "function value" anywhere else, could we just say "any other function"?

```suggestion
    so that it can be used like any other function in Dart.
```

src/data/glossary.yml
parlough: @eernstg Is "runtime (or compiler)" the accurate actor here?

parlough (commented):
This is really cool! A bit more niche of a term but definitely something Dart developers might come across and now they can get an introduction through this entry.

Generally looks great to me, thanks @IldySilva!

I don't think the entry should be super detailed, but to make sure we are are introducing the semantics correctly, I'd appreciate a quick review from @eernstg as well.

src/data/glossary.yml
```diff
@@ -141,6 +141,54 @@
     - "Function callback"
     - "Callback function"
 
+- term: "Closurization"
+  short_description: |-
+    The process of turning a method or function tear-off into a closure.
+  long_description: |-
+    **Closurization** is the process where Dart turns a method or function
+    tear-off into a function object that can be stored, passed, or invoked later.
```
eernstg: The verb phrase 'to tear off [something]' is pretty common. It's somewhat informal, but that should be fine in the glossary. The specification documents tend to prefer 'a property extraction' (which may be a getter invocation or a closurization). 'A tear-off' may be the function object which is the result of a closurization.

This means that there's a terminology clash when we say 'an instance method or function tear-off'. I think it's better to simply delete 'tear-off' from line 149.

The following text could then say '... into a function object (or tear-off) that can be stored ...'.

src/data/glossary.yml
```diff
@@ -141,6 +141,54 @@
...
+
+    A tear-off is a reference to an existing function or method without calling it.
```
eernstg: There's a coherence issue with 'is a reference ... without calling it'. Perhaps:
```suggestion
    A tear-off is a reference to an existing function or method that evaluates to a
    function object for that existing entity, but does not call it.
```

src/data/glossary.yml
eernstg: How about "wraps it into a function object which can be used like any other function object in Dart."?

It looks like this text is attempting to define/explain the notion of a function object, but we already have similar material in line 149.

parlough:
Thanks for the feedback Erik!

@IldySilva This PR looks good to me once you apply Erik's suggestion (or some form of them). Let me know if you have any questions or issues :D

antfitch:
Before this is merged, please make sure that I or another technical writer approves first.

connie-ooi (approved):

PR #6895, opened 2025-09-23
Add "Obviously typed" entry to glossary

IldySilva:
Adds a glossary entry for the concept of an "obviously typed" expression that will be referenced by related diagnostics.

Resolves https://github.com/dart-lang/site-www/issues/6652

### DIFF
--- a/src/data/glossary.yml
+++ b/src/data/glossary.yml
@@ -928,8 +928,8 @@
       `String`.
 
     In this list, _homogeneous_ means "where every element has the same type".
-    For example, `{1: "one", 2: "two"}` is a homogeneous map,
-    but `{1: "one", 1.5: true}` is not.
+    For example, `{1: 'one', 2: 'two'}` is a homogeneous map,
+    but `{1: 'one', 1.5: true}` is not.
     By contrast, if the type of an expression requires inference (for example
     `[1, 2.5]` → `List<num>`), it is **not** considered obviously typed.
 


src/data/glossary.yml
```diff
@@ -883,6 +883,54 @@
...
+
+    This concept is mainly used in lint rules like
+    `omit_obvious_local_variable_types` and
+    `specify_nonobvious_property_types` to decide whether an explicit type
+    annotation is redundant or necessary.
+
+    ### Examples of obviously typed expressions
+    - Non-collection literals: `1`, `true`, `'Hello'`
+    - Collection literals with explicit type arguments: `<int>[]`, `<String, bool>{}`
+    - Homogeneous list/set literals: `[1, 2, 3]`, `{true, false}`
```
eernstg: This item should also mention that each element in the list must have an obvious type. So `[1, 2, f()]` isn't obviously typed even in the case where `f` has return type `int`.

```suggestion
    - Homogeneous list/set literals whose elements are obviously typed: `[1, 2, 3]`, `{true, false}`
```

src/data/glossary.yml
```diff
@@ -883,6 +883,54 @@
...
+    - Map literals with obvious key and value types: `{1: 10, 2: 20}`
```
eernstg:
```suggestion
    - Homogeneous map literals with obvious key and value types: `{1: 10, 2: 20}`
```

src/data/glossary.yml
```diff
@@ -883,6 +883,54 @@
...
+    - Record literals with obvious types: `(1, enabled: true)`
+    - Instance creation with non-generic classes or explicit type args: `C()`, `C<int>()`
```
eernstg: Two items are missing here:

```suggestion
    - Local variables and formal parameters that have not been promoted.
    - Instance creation with non-generic classes or explicit type arguments: `C()`, `Set<int>()`
    - Cascades with obviously typed targets: `StringBuffer('Hello, ')..write('world!')`.
```

I changed `C<int>()` to `Set<int>()` because it's confusing to have two examples right next to each other (`C()` and `C<int>()`) where it is implied by the first one that `C` is a non-generic class, and implied by the second one that `C` is generic.

src/data/glossary.yml
```diff
@@ -883,6 +883,54 @@
...
+    - Type casts: `x as int`
+    - Type tests: `x is String`
```
eernstg: Missing item:

```suggestion
    - Conditional expressions with obviously typed branches of the same type: `condition ? 1 : 2`.
    - Type tests: `x is String`
```

src/data/glossary.yml
```diff
@@ -883,6 +883,54 @@
...
+    - Throw expressions: `throw Exception()`
+    - `this` and type literals: `String`
```
eernstg: Missing item:

```suggestion
    - A parenthesized expression whose contents is obviously typed: `('Hello!')`.
    - `this` and type literals: `String`
```

src/data/glossary.yml
```diff
@@ -883,6 +883,54 @@
...
+    An expression in Dart is considered **obviously typed** when its type is
+    immediately clear from its syntax, without needing inference or complex
+    analysis.
+
+    This concept is mainly used in lint rules like
+    `omit_obvious_local_variable_types` and
+    `specify_nonobvious_property_types` to decide whether an explicit type
+    annotation is redundant or necessary.
+
+    ### Examples of obviously typed expressions
```
eernstg: This concept is defined by this entry (we don't have plans to define it anywhere else), so it needs to be precise, it's not just about examples.

```suggestion
    ### Definition of obviously typed expressions
```

src/data/glossary.yml
```diff
@@ -883,6 +883,54 @@
...
+    - Collection literals with explicit type arguments: `<int>[]`, `<String, bool>{}`
+    - Homogeneous list/set literals: `[1, 2, 3]`, `{true, false}`
+    - Map literals with obvious key and value types: `{1: 10, 2: 20}`
+    - Record literals with obvious types: `(1, enabled: true)`
+    - Instance creation with non-generic classes or explicit type args: `C()`, `C<int>()`
+    - Type casts: `x as int`
+    - Type tests: `x is String`
+    - Throw expressions: `throw Exception()`
+    - `this` and type literals: `String`
+
```
eernstg, line 929:
```suggestion

In this list, _homogeneous_ means "where every element has the same type". For example, `{1: "one", 2: "two"}` is a homogeneous map, but `{1: "one", 1.5: true}` is not.
```

src/data/glossary.yml
IldySilva: Great Point, thanks

eernstg (approved):
Thanks, LGTM!

src/data/glossary.yml
```diff
@@ -883,6 +883,62 @@
...
+    An expression in Dart is considered **obviously typed** when its type is
+    immediately clear from its syntax, without needing inference or complex
+    analysis.
+
+    This concept is mainly used in lint rules like
+    `omit_obvious_local_variable_types` and
+    `specify_nonobvious_property_types` to decide whether an explicit type
+    annotation is redundant or necessary.
+
+    ### Definition of obviously typed expressions
```
parlough: Consider a simpler titler with a brief introduction to the list.

```suggestion
    ### Obviously typed expressions
    
    The following types of expressions are considered obviously typed:
```

src/data/glossary.yml
```diff
@@ -883,6 +883,62 @@
...
+
+    - Non-collection literals: `1`, `true`, `'Hello'`
```
parlough: Each item in a list should have a [parallel syntax](https://developers.google.com/style/lists#parallel-syntax) with consistent [capitalization and end punctuation](https://developers.google.com/style/lists), but right now there's a mix with some ending in a period and some not.

For the purposes of this list, I think you can end each one with a period.

src/data/glossary.yml
```diff
@@ -883,6 +883,62 @@
...
+    - Homogeneous map literals with obvious key and value types: `{1: 10, 2: 20}`
+    - Record literals with obvious types: `(1, enabled: true)`
+    - Local variables and formal parameters that have not been promoted.
+    - Instance creation with non-generic classes or explicit type arguments: `C()`, `Set<int>()`
+    - Cascades with obviously typed targets: `StringBuffer('Hello, ')..write('world!')`.
+    - Type casts: `x as int`
+    - Conditional expressions with obviously typed branches of the same type: `condition ? 1 : 2`.
+    - Type tests: `x is String`
+    - Throw expressions: `throw Exception()`
+    - A parenthesized expression whose contents is obviously typed: `('Hello!')`.
```
parlough: Since all of the other list entries are plural, update this one to match so the list has a [parallel structure](https://developers.google.com/style/lists):

```suggestion
    - Parenthesized expressions whose contents are obviously typed: `('Hello!')`.
```

src/data/glossary.yml
```diff
@@ -883,6 +883,62 @@
...
+    This distinction helps style rules decide when to enforce or omit type
+    annotations for clarity and consistency.
+
+  related_links:
+    - text: "Effective Dart: Style"
+      link: "/effective-dart/style"
+      type: "doc"
+    - text: "Lint: omit_obvious_local_variable_types"
+      link: "/tools/linter-rules/omit_obvious_local_variable_types"
+      type: "doc"
```
parlough: For lints and diagnostics, we actually have a special type we can use :) 

```suggestion
      type: "diagnostic"
```

src/data/glossary.yml
```diff
@@ -883,6 +883,62 @@
...
+    - text: "Lint: specify_nonobvious_property_types"
+      link: "/tools/linter-rules/specify_nonobvious_property_types"
+      type: "doc"
```
```suggestion
      type: "diagnostic"
```

parlough (commented):
Thanks for working on this large, complicated entry @IldySilva! It'll be really nice to have this to point to from the relevant diagnostic docs.

The content looks good to me considering Erik's approval. I just have some small suggestions to match the style guide.

parlough (approved):
Thanks for making those adjustments! Looks great to me :D

antfitch:
Hey folks, I did not finish my review of this. Just an FYI that I want to make sure a technical writer always stamps writing tasks with an LGTM. I'll see if there is a rule I can set up to enforce this.

parlough:
> Hey folks, I did not finish my review of this. Just an FYI that I want to make sure a technical writer always stamps writing tasks with an LGTM. I'll see if there is a rule I can set up to enforce this.

Ah sorry, that's my fault. I would normally wait for a tech writer reviewer on this, but I mixed this PR up with another PR where you did review.

If you have any feedback, do feel free to follow up up with a PR or let me know and I can do so. If you see anything egregious, feel free to revert and we can reopen.

PR #6898, opened 2025-09-25
Replace site-shared submodule with git dependencies

parlough:
The `inject_dartpad.dart.js` script needs to be inlined now that site-shared is not a submodule. However, this is temporary as we plan to replace this with a new Dart API directly from a Jaspr component. The API is added in https://github.com/dart-lang/site-shared/pull/247.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,0 @@
-[submodule "site-shared"]
-	path = site-shared
-	url = https://github.com/dart-lang/site-shared
-	branch = main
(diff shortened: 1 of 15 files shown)


antfitch (approved):
This is a big one!

mit-mit:
This makes me soooo happy! Can we get ride of all the submodule silliness comments in https://github.com/dart-lang/site-www/blob/main/README.md too?

parlough:
> This makes me soooo happy! Can we get ride of all the submodule silliness comments in https://github.com/dart-lang/site-www/blob/main/README.md too?

Thanks for pushing me to do this! Nice that we can do this now that everything used is just a Dart package.

I forgot to push that commit before this landed. I opened https://github.com/dart-lang/site-www/pull/6904 with the doc updates and workflow cleanup.

PR #6904, opened 2025-09-26
Clean up docs and workflows after submodule removal

parlough:
Follow up that I forgot to push to the https://github.com/dart-lang/site-www/pull/6898 PR.

- Stops setting up the submodule in GitHub workflows.
- Stops checking if the submodule is up to date with dependabot.
- Doesn't try to initialize the submodule in Firebase Studio.
- Remove the instructions around the submodule from the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,10 +33,6 @@ you probably don't need to build this site.
 Often you can make changes using the GitHub UI.
 We can stage the changes automatically in your pull request.
 
-> [!IMPORTANT]  
-> If you are cloning this repository locally,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -66,10 +66,9 @@ and already the latest stable version:
 dart --version
 ```
 
-### Clone this repo and its submodules
+### Clone this repo
 
 > [!NOTE]
```
mit-mit: I suggest we just git rid of this whole note; I don't think this README is the place to teach folks about git

PR #6921, opened 2025-10-02
Show a tooltip when copying code to clipboard 

### LINKED ISSUE #6841: Show a toast or tooltip when copying code to clipboard
When the copy button is pushed in a code block (or elsewhere), we should have some sort of toast to provide confirmation the content was actually copied.

This can be added to https://github.com/dart-lang/site-www/blob/main/site/lib/src/components/copy_button.dart.

IldySilva:
Show temporary "Copied!" label after clicking the CopyButton.

Fixes #6841


site/lib/src/components/copy_button.dart
```diff
@@ -28,20 +28,32 @@ class CopyButton extends StatefulComponent {
...
-
     super.initState();
   }
 
+  void copy() {
+    web.window.navigator.clipboard.writeText(component.toCopy);
+
+    setState(() => _copied = true);
+
+    Future.delayed(const Duration(seconds: 2), () {
```
parlough: To provide static feedback if we unintentionally return something from the delayed closure:

```suggestion
    Future<void>.delayed(const Duration(seconds: 2), () {
```

site/lib/src/components/copy_button.dart
```diff
@@ -28,20 +28,32 @@ class CopyButton extends StatefulComponent {
...
   void initState() {
-    // Unhide the copy button if successfully initialized on the client.
     if (kIsWeb && component.toCopy.isNotEmpty) {
       _hidden = false;
     }
-
     super.initState();
   }
 
+  void copy() {
```
parlough:
```suggestion
  void _copy() {
```

site/lib/src/components/copy_button.dart
```diff
@@ -50,11 +62,9 @@ class _CopyButtonState extends State<CopyButton> {
         ...component.classes,
       ],
       title: component.title ?? 'Copy ${component.toCopy} to your clipboard.',
-      content: component.buttonText,
+      content: _copied ? 'Copied!' : component.buttonText,
       icon: iconButton ? 'content_copy' : null,
-      onClick: () {
-        web.window.navigator.clipboard.writeText(component.toCopy);
-      },
+      onClick: copy,
```
parlough:
```suggestion
      onClick: _copy,
```

site/lib/src/components/copy_button.dart
```diff
@@ -50,11 +62,9 @@ class _CopyButtonState extends State<CopyButton> {
         ...component.classes,
       ],
       title: component.title ?? 'Copy ${component.toCopy} to your clipboard.',
-      content: component.buttonText,
+      content: _copied ? 'Copied!' : component.buttonText,
```
parlough, line 65: I **love** this solution! Works well for all usages of this component without any new styles needed.

PR #6930, opened 2025-10-13
Remove no longer used portions of `install.js`

parlough:
- Removes portions of `install.js` that haven't had any affect for a long time due to changes to the Get Dart page.
- No longer includes `install.js` on the archive page where it has no effect.

These changes result in no functional difference, but can slightly reduce the loading time of these pages.


antfitch:
Goodbye ancient code bits.

Issue #6689, opened 2025-06-23
Dart 3.10 Dot Shorthands


antfitch:
See this issue for details: https://github.com/dart-lang/site-www/issues/6285

kallentu:
We'll be most likely releasing dot shorthands for 3.10, instead of 3.9, if that changes anything.

Issue #6285, opened 2024-12-18
[Dot Shorthands] Add Dot Shorthands to the language docs


kallentu:
This issue tracks the work needed for adding dot shorthands to the language docs

b/395888113 tracks the work needed

### Page

https://dart.dev/language#enums (update)
https://dart.dev/language/dot-shorthand (new - maybe needed)

### References
* [Original issue](https://github.com/dart-lang/language/issues/357)
* [Feature specification](https://github.com/dart-lang/language/blob/main/working/3616%20-%20enum%20value%20shorthand/proposal-simple-lrhn.md)
* [Tracking doc](https://docs.google.com/document/d/1mfxozFP9hXfQUBoeTrSUgw9rucEuNTqY-4KxuKZKcEM/edit?pli=1&tab=t.0#heading=h.xii6f83bmssy)
* [Meta issue](https://github.com/dart-lang/sdk/issues/57036)

Project plan - [go/dot-shorthands-v1-ix](http://goto.google.com/dot-shorthands-v1-ix)

kallentu:
Moving the assignment to @connie-ooi .
Dot shorthands will be releasing for stable 3.10 which is on November 12th.

PR #6941, opened 2025-10-14
[Dart 3.10] Document dot shorthands

conooi:
Resolves https://github.com/dart-lang/site-www/issues/6285
Resolves https://github.com/dart-lang/site-www/issues/6689

Files changed: 
- examples/language/lib/classes/shorthand.dart
- src/content/language/dot-shorthands.md
- src/content/language/enums.md
- src/content/language/extension-methods.md
- src/content/language/index.md
- src/data/sidenav.yml

Link to page previews:
- [Dot shorthands](https://dart-dev--pr6941-dotshorthands-9ckdrfhh.web.app/language/dot-shorthands)
- [Intro to Dart](https://dart-dev--pr6941-dotshorthands-9ckdrfhh.web.app/language)
- [Enums](https://dart-dev--pr6941-dotshorthands-9ckdrfhh.web.app/language/enums)
- [Extension methods](https://dart-dev--pr6941-dotshorthands-9ckdrfhh.web.app/language/extension-methods)


examples/language/lib/classes/shorthand.dart
```diff
@@ -0,0 +1,185 @@
...
+const Point myOrigin = .origin(); // Instead of const Point.origin()
+
+// Using shorthands in a const collection literal
+const List<Point> keyPoints = [ .origin(), .new(1.0, 1.0) ]; // Instead of [const Point.origin(), const Point(1.0, 1.0)]
+// #enddocregion const
+
+// #docregion nested
+// AVOID: Hard to read, types are hidden
+// Widget complex = .container(
+//   child: .padding(
```
kallentu: `.padding`, `.text` and `.container` don't exist? Unless this is a hypothetical example.
Otherwise, this should be more like: 
```
// Widget complex = .new(
//   child: .new(
//     padding: .all(8.0),
//     child: .new('Hello')
//   )
// );

File: examples/language/lib/classes/shorthand.dart [hunk 2]
```diff
@@ -0,0 +1,185 @@
...
+//     padding: .all(8.0),
+//     child: .text('Hello')
+//   )
+// );
+
+// PREFER: Explicit types in nested structures
```
kallentu: I wouldn't say this is exactly the guidance we're giving. You can and should use shorthands in nested structures. In [Loïc's dogfooding doc](https://docs.google.com/document/d/1UcaYBTdP2yLQi8aDgSOLoxqWe1jrzkeo0o9fyNOnOpQ/edit?resourcekey=0-fk1Vz0zUIJQTLQLgFDK__g&tab=t.0#heading=h.q6vfmmxhuhxd), he put his recommendations out as the following:
```
Recommendation: The Flutter framework uses dot shorthands when the target type is obvious.

⚠️ This recommendation disallows the vast majority of potential usages of dot shorthands in the Flutter framework.

⚠️ This recommendations only applies to the Flutter framework. I recommend that app developers use dot shorthands everywhere, except in return statements of long methods.
```

Perhaps the recommendation should be changed to use dot shorthands as much as possible (where it makes sense to), even in nested structures. `.new()` is a little nuanced and perhaps should not be used with a lot of nesting or without obvious types, but we haven't set clear guidance on it.

examples/language/lib/classes/shorthand.dart
```diff
@@ -0,0 +1,185 @@
...
+  return switch (level) {
+    .debug   => 'gray',    // Instead of LogLevel.debug
+    .info    => 'blue',    // Instead of LogLevel.info
+    .warning => 'orange',  // Instead of LogLevel.warning
+    .error   => 'red',     // Instead of LogLevel.error
+  };
+}
+
+// Example usage:
+// var warnColor = colorCode(.warning); // Returns 'orange'
```
kallentu: Uncomment?
Would also be nice to have more examples with using dot shorthands for arguments (like here).

examples/language/lib/classes/shorthand.dart
```diff
@@ -0,0 +1,185 @@
...
+// #docregion tearoff
+class Logger {
+  final String name;
+  Logger(this.name); // Unnamed constructor
+  Logger.verbose() : name = 'VERBOSE'; // Named constructor
+}
+
+void main() {
+  // .new for an UNNAMED constructor tear-off:
+  Logger Function(String) createLogger = .new; // Instead of Logger.new
```
kallentu: `.new` for this context doesn't actually work unfortunately.
I think the _only_ situation where `.new` can be used is `Object x = .new` where the context type is `Object`. Otherwise, there's no tearoff that would satisfy the context type requirement here. Same with 78. We should rewrite/remove these.

examples/language/lib/classes/shorthand.dart
```diff
@@ -0,0 +1,185 @@
...
+ 
+// #docregion constructors
+ class Point {
+  final double x, y;
+  const Point(this.x, this.y);
+  const Point.origin() : x = 0, y = 0; // Named constructor
+  static Point fromList(List<double> list) => Point(list[0], list[1]); // Static method
+}
+
+// Named constructor
```
kallentu: You can also show unnamed constructor examples `.new(<params>)`.

src/content/language/dot-shorthands.md
```diff
@@ -0,0 +1,280 @@
+---
+title: Shorthand syntax
+description: Learn about the shorthand syntax in Dart.
```
kallentu: Perhaps we should call it the dot shorthand syntax. There might be other shorthand syntaxes coming down the line from the language team and it might be confusing to have this broader-sounding title when that happens.

src/content/language/dot-shorthands.md
```diff
@@ -0,0 +1,280 @@
...
+shortTitle: Shorthand
+prevpage:
+  url: /language/enums
+  title: Enums
+nextpage:
+  url: /language/extension-methods
+  title: Extension methods
+---
+
+:::experimental
```
kallentu: Are we planning on removing this once the feature ships?

src/content/language/dot-shorthand.md
```diff
@@ -0,0 +1,280 @@
...
+  url: /language/extension-methods
+  title: Extension methods
+---
+
+:::experimental
+Support for shorthand syntax is experimental and
+can only be used on the `main` channel.
+:::
+
+Write more concise Dart code by using the shorthand syntax `.foo`
```
kallentu: Maybe specify `dot shorthand syntax`?

src/content/language/index.md
```diff
@@ -282,9 +282,33 @@ if (!yourPlanet.isGiant) {
 }
 ```
 
+When the compiler can infer the enum type from the context, 
+you can use a more concise shorthand syntax to access enum values.
```
kallentu: I'd also explicitly write out `concise dot shorthand syntax` here.

src/content/language/dot-shorthand.md
```diff
@@ -0,0 +1,280 @@
...
+---
+
+:::experimental
+Support for shorthand syntax is experimental and
+can only be used on the `main` channel.
+:::
+
+Write more concise Dart code by using the shorthand syntax `.foo` 
+instead of `ContextType.foo`. The compiler can infer the type 
+whenever it's clear from the context, letting you access
```
kallentu: Unfortunately, I feel like if I was a user using this feature, I would want an explicit doc on how this magical "context" and "context type" works. We gloss over it in the doc, but I'm not sure if we would like to be more clear on its definition and how we obtain the context type.

src/content/language/dot-shorthand.md
```diff
@@ -0,0 +1,280 @@
...
+
+// Constructor example
+class Point {
+  final int x, y;
+  Point(this.x, this.y);
+  Point.origin() : x = 0, y = 0;
+}
+Point origin = .origin(); // Instead of Point.origin()
+```
+## Using shorthand syntax
```
kallentu: `Using dot shorthand syntax` and adding `dot` on line 45 as well.

src/content/language/dot-shorthand.md
```diff
@@ -0,0 +1,280 @@
...
+
+### Enums
+A primary and highly recommended use case for shorthands is with enums, 
+especially in assignments and switch statements, where the 
+enum type is very clear.
```
kallentu: Maybe `obvious` instead of the word `clear`?

src/content/language/dot-shorthand.md
```diff
@@ -0,0 +1,280 @@
...
+  return switch (level) {
+    .debug   => 'gray',    // Instead of LogLevel.debug
+    .info    => 'blue',    // Instead of LogLevel.info
+    .warning => 'orange',  // Instead of LogLevel.warning
+    .error   => 'red',     // Instead of LogLevel.error
+  };
+}
+
+// Example usage:
+// var warnColor = colorCode(.warning); // Returns 'orange'
```
kallentu: Same question here, do we want to uncomment this?

src/content/language/dot-shorthand.md
```diff
@@ -0,0 +1,280 @@
...
+```
+
+### Constructors
+Shorthands are useful for invoking named constructors,
```
kallentu: Nit: `Dot shorthands`

src/content/language/dot-shorthand.md
```diff
@@ -0,0 +1,280 @@
...
+}
+
+// Named constructor
+Point origin = .origin(); // Instead of Point.origin()
+
+// Static factory method
+Point p1 = .fromList([1.0, 2.0]); // Instead of Point.fromList([1.0, 2.0])
+
+// With type arguments for generic class constructors
+List<int> intList = .filled(5, 0); // Instead of List<int>.filled(5, 0)
```
kallentu: This isn't the equivalent to the comment of `List<int>.filled(5, 0)` actually. It's equivalent to `List.filled(5,0)` where the compiler then infers the type parameters. It's obvious here, but there are some examples where this makes a difference.

This is a contrived example, but something like this wouldn't work if we pulled the type parameters from the context:
```

class C<T> {
  static C<X> foo<X>(X x) => new C<X>();
  C<U> cast<U>() => new C<U>();
}

void main() {
  C<bool> c = .foo("String").cast(); // C.foo and not C<bool>.foo
}
```

src/content/language/dot-shorthand.md
```diff
@@ -0,0 +1,280 @@
...
+``` 
+
+### Static members
+You can use shorthand syntax to call static methods or 
+access static fields/getters of a class when the class 
+type is providing the context.
```
kallentu: "when the class type is providing the context"
This writes a little weird to me. The class type will always be providing the context, it's not really a matter of _when_.

src/content/language/dot-shorthands.md
```diff
@@ -0,0 +1,280 @@
...
+const Point myOrigin = .origin(); // Instead of const Point.origin()
+
+// Using shorthands in a const collection literal
+const List<Point> keyPoints = [ .origin(), .new(1.0, 1.0) ]; // Instead of [const Point.origin(), const Point(1.0, 1.0)]
+```
+
+
+## Advanced usage
+
+### Chaining
```
kallentu: Not sure if it's more confusing to mention this, but chaining is so particular because we pull the context type and use that as the type for the shorthand, but the type that ends up returning at the end of the chain also needs to match that same initial context type.

src/content/language/dot-shorthand.md
```diff
@@ -0,0 +1,280 @@
...
+method calls or property accesses, just as if you had 
+written out the type name.
+
+<?code-excerpt "language/lib/classes/shorthand.dart (chain)"?>
+```dart
+// .fromCharCode(72) resolves to the String "H",
+// then the instance method .toLowerCase() is called on that String.
+String lowerH = .fromCharCode(72).toLowerCase(); // Instead of String.fromCharCode(72).toLowerCase()
+
+// print(lowerH); // Output: h
```
kallentu: Uncomment?

src/content/language/dot-shorthand.md
```diff
@@ -0,0 +1,280 @@
...
+
+  // ERROR: The type context is lost by casting `myColor` to `Object`.
+  // The compiler no longer knows that `.green` should refer to `Color.green`.
+  if ((myColor as Object) == .green) {
+    print('This will not compile.');
+  }
+}
+```
+
+### Constructor tear-offs
```
kallentu: This also needs changing based on my above comment ^

src/content/language/dot-shorthand.md
```diff
@@ -0,0 +1,280 @@
...
+
+  print(appLog.name);         // Prints "App"
+  print(verboseLog.name);     // Prints "VERBOSE"
+  print(stringSet.isEmpty);   // Prints "true"
+}
+```
+
+
+
+## Best practices and limitations
```
kallentu: Will need changing based on above comment ^

src/content/language/dot-shorthand.md
```diff
@@ -0,0 +1,280 @@
...
+// );
+
+// PREFER: Explicit types in nested structures
+Widget complex = Container( // Assuming types
+  child: Padding(
+    padding: EdgeInsets.all(8.0),
+    child: Text('Hello')
+  )
+);
+```
```
kallentu: I also don't know if we do this, but do we ever link to the actual spec in documentation? I feel like if users wanted a full, exhaustive rundown of the feature, it'd be nice to get to the spec to read. @antfitch Do you have any guidance here?

src/content/language/dot-shorthands.md
antfitch: +1 for `Dart shorthand syntax`

src/content/language/dot-shorthand.md
antfitch: +1, make sure to update this everywhere.

antfitch:
@connie-ooi and @kallentu, after you've finished your review, let me know and I'll start my review. I'll be focusing on flow. If any technical changes are required, @kallentu, I'll pull you in.

src/content/language/dot-shorthands.md
conooi: Yes!

src/content/language/dot-shorthand.md
```diff
@@ -0,0 +1,328 @@
...
+---
+
+:::experimental
+Support for dot shorthand syntax is experimental and
+can only be used on the `main` channel.
+:::
+
+Write more concise Dart code by using the dot shorthand syntax `.foo` 
+instead of `ContextType.foo`. The compiler can infer the type 
+whenever it's clear from the context, letting you access
```
kallentu: Nit: Maybe just "The compiler infers the type from context, ..."

src/content/language/dot-shorthand.md
```diff
@@ -0,0 +1,328 @@
...
+  final int x, y;
+  Point(this.x, this.y);
+  Point.origin() : x = 0, y = 0;
+}
+Point origin = .origin(); // Instead of Point.origin()
+```
+
+## The role of context type 
+Dot shorthand syntax works because the compiler understands 
+the context type. This is the type it expects an expression
```
kallentu: For the first two sentences, maybe something like "Dot shorthands use the context type to determine the member we're resolving to. The context type is the type that Dart expects an expression to have based on its location."

src/content/language/dot-shorthand.md
```diff
@@ -0,0 +1,328 @@
...
+  };
+}
+
+// Example usage:
+var warnColor = colorCode(.warning); // Returns 'orange'
+```
+
+### Named constructors
+Dot shorthands are useful for invoking named constructors, 
+static factory methods, or constructors of generic classes
```
kallentu: I don't think "static factory methods" exist. Factory constructors are probably what you're looking for and you'd probably want to change the `Point.fromList` to be a factory constructor to demonstrate that. And then update the comments in the code snippet.

src/content/language/dot-shorthand.md
```diff
@@ -0,0 +1,328 @@
...
+  static Point fromList(List<double> list) => Point(list[0], list[1]); // Static method
+}
+
+// Named constructor
+Point origin = .origin(); // Instead of Point.origin()
+
+// Static factory method
+Point p1 = .fromList([1.0, 2.0]); // Instead of Point.fromList([1.0, 2.0])
+
+// With type arguments for generic class constructors
```
kallentu: I'm not entirely sure what this comment is supposed to mean, maybe update it? Are we just trying to show a generic class constructor?

src/content/language/dot-shorthand.md
```diff
@@ -0,0 +1,328 @@
...
+  final double x, y;
+  const Point(this.x, this.y);
+  const Point.origin() : x = 0, y = 0; // Named constructor
+  static Point fromList(List<double> list) => Point(list[0], list[1]); // Static method
+}
+
+// Named constructor
+Point origin = .origin(); // Instead of Point.origin()
+
+// Static factory method
```
kallentu: Update this comment.

src/content/language/dot-shorthand.md
```diff
@@ -0,0 +1,328 @@
...
+static factory methods, or constructors of generic classes 
+where type arguments are provided.
+
+<?code-excerpt "language/lib/classes/shorthand.dart (constructors)"?>
+```dart
+ class Point {
+  final double x, y;
+  const Point(this.x, this.y);
+  const Point.origin() : x = 0, y = 0; // Named constructor
+  static Point fromList(List<double> list) => Point(list[0], list[1]); // Static method
```
kallentu: Update this comment too.

src/content/language/dot-shorthands.md
```diff
@@ -0,0 +1,328 @@
...
+List<int> intList = .filled(5, 0); // Instead of List.filled(5, 0)
+``` 
+### Unnamed constructors
+The `.new` dot shorthand provides a concise way to call an
+unnamed constructor of a class. This is useful 
+for instantiating fields and variables where the type is 
+already explicitly declared.
+
+<?code-excerpt "language/lib/classes/shorthand.dart (unnamed)"?>
+```dart
```
kallentu, line 150: Maybe this code snippet would be easier to read if you broke it out into two code blocks (one showing before and one showing after) rather than having it commented as before and after in a single code block.

src/content/language/dot-shorthand.md
```diff
@@ -0,0 +1,328 @@
...
+```dart
+// Static method
+int httpPort = .parse('80'); // Instead of int.parse('80')
+
+// Static field/getter
+BigInt bigIntZero = .zero; // Instead of BigInt.zero
+```
+
+### Constant expressions 
+You can use dot shorthands within `const` contexts
```
kallentu: Nit: "You can use dot shorthands within a constant context..."

src/content/language/dot-shorthand.md
```diff
@@ -0,0 +1,328 @@
...
+```dart
+enum Status { none, running, stopped, paused }
+class Point {
+  final double x, y;
+  const Point(this.x, this.y);
+  const Point.origin() : x = 0.0, y = 0.0;
+}
+
+// Enum values are always constants
+const Status defaultStatus = .running; // Instead of const Status.running
```
kallentu: Nit: In the comments here, you wouldn't actually write `const` twice.
So you wouldn't write:
`const Status defaultStatus = const Status.running;` (That would actually be linted against)
You would probably write:
`const Status defaultStatus = Status.running;`

Meaning, `// Instead of const Status.running` -> `// Instead of Status.running` and all the comments underneath.

src/content/language/dot-shorthand.md
```diff
@@ -0,0 +1,328 @@
...
+
+  // ERROR: The type context is lost by casting `myColor` to `Object`.
+  // The compiler no longer knows that `.green` should refer to `Color.green`.
+  if ((myColor as Object) == .green) {
+    print('This will not compile.');
+  }
+}
+```
+
+## Best practices
```
kallentu: I wonder if we should have best practices at all? I chatted with Bob a bit about it and it seems like we typically don't prescribe anything and just allow people to use it. Maybe once we've seen how people use it, we'd come back and update with more guidance. I'm not sure if we should be telling people to avoid using it in any sort of way right now (even though Loic's dogfooding had specific recs)

Wonder if Amanda has any guidance on this?

conooi: Yeah that makes sense to me! I can take out that section and we can update it later

src/data/sidenav.yml
```diff
@@ -60,6 +60,8 @@
           permalink: /language/mixins
         - title: Enums
           permalink: /language/enums
+        - title: Dot Shorthands
+          permalink: /language/dot-shorthand
```
kallentu: Opinion on making this plural? I think I always refer to this feature as dot shorthands.

conooi: Fixed! Thank you

src/data/sidenav.yml
```diff
@@ -60,6 +60,8 @@
           permalink: /language/mixins
         - title: Enums
           permalink: /language/enums
+        - title: Dot Shorthands
```
parlough: We use [sentence case for titles](https://developers.google.com/style/headings).

```suggestion
        - title: Dot shorthands
```

src/content/language/dot-shorthands.md
```diff
@@ -0,0 +1,266 @@
+---
+title: Dot shorthands
+description: Learn about the dot shorthand syntax in Dart.
+shortTitle: Dot Shorthands
```
parlough: You don't need a short title if it matches the normal one.

```suggestion
```

src/content/language/dot-shorthands.md
```diff
@@ -0,0 +1,266 @@
...
+Support for dot shorthand syntax is experimental and
+can only be used on the `main` channel.
+:::
+
+Write more concise Dart code by using the dot shorthand syntax `.foo` 
+instead of `ContextType.foo`. The compiler infers the type 
+from context, letting you access enum values, static members, 
+or constructors more cleanly.
+
+Here is a quick look:
```
parlough: Prefer [using contractions](https://developers.google.com/style/contractions):

```suggestion
Here's a quick look:
```

examples/language/lib/classes/shorthand.dart
```diff
@@ -0,0 +1,172 @@
...
+// #docregion chain
+// .fromCharCode(72) resolves to the String "H",
+// then the instance method .toLowerCase() is called on that String.
+String lowerH = .fromCharCode(72).toLowerCase(); 
+// Instead of String.fromCharCode(72).toLowerCase()
+
+print(lowerH); // Output: h
+// #enddocregion chain
+
+// #docregion allowedequality
```
parlough: When naming docregions with multiple words, we have a convention of using dashes between the word. So in this case, `allowed-equality`.

examples/language/lib/classes/shorthand.dart
```diff
@@ -0,0 +1,172 @@
...
+
+// #docregion unnamedafter
+class _PageState extends State<Page> {
+  final AnimationController _animationController = .new(vsync: this);
+  final ScrollController _scrollController = .new();
+  final GlobalKey<ScaffoldMessengerState> scaffoldKey = .new();
+  Map<String, Map<String, bool>> properties = .new();
+  // ...
+}
+// #enddocregion unnamedafter
```
parlough: Each file should have one extra new line at the end.

```suggestion
// #enddocregion unnamedafter

```

src/content/language/dot-shorthands.md
```diff
@@ -0,0 +1,266 @@
...
+  title: Enums
+nextpage:
+  url: /language/extension-methods
+  title: Extension methods
+---
+
+:::experimental
+Support for dot shorthand syntax is experimental and
+can only be used on the `main` channel.
+:::
```
parlough: Since we know when this is landing and what the requirement will be, let's add a language version note instead. That way it will continue to be helpful and relevant and not need a second update.

```suggestion
:::version-note
Dot shorthands require a [language version][] of at least 3.10.
:::

[language version]: /resources/language/evolution#language-versioning
```

src/content/language/dot-shorthands.md
```diff
@@ -0,0 +1,266 @@
...
+Dot shorthands use the context type to determine the member 
+the complier resolves to. The context type is the type that Dart 
+expects an expression to have based on its location. 
+For example, in `Status currentStatus = .running`, 
+the compiler knows a `Status` is expected, so it infers 
+`.running` to mean `Status.running`.
+
+## Using dot shorthand syntax
+
+### Enums
```
parlough, line 79: Here and elsewhere, after each header add an empty new line:

```suggestion

### Enums
```

src/content/language/index.md
```diff
@@ -282,9 +282,33 @@ if (!yourPlanet.isGiant) {
...
+```dart
+// Instead of the full, explicit syntax:
+Planet myPlanet = Planet.venus;
+
+// You can use the dot shorthand:
+Planet myPlanet = .venus;
+```
+Dot shorthand isn't limited to variable declarations. 
+It can also be used in contexts like function arguments 
+and switch cases where the enum type is clear to the compiler.
```
parlough, line 307:
```
and switch cases where Dart can infer the enum type.
```

src/content/language/index.md
```diff
@@ -282,9 +282,33 @@ if (!yourPlanet.isGiant) {
...
+
+```dart
+// Instead of the full, explicit syntax:
+Planet myPlanet = Planet.venus;
+
+// You can use the dot shorthand:
+Planet myPlanet = .venus;
+```
+Dot shorthand isn't limited to variable declarations. 
+It can also be used in contexts like function arguments
```
parlough: It seems we tend to use the plural form to refer to the feature:

```suggestion

Dot shorthands aren't limited to variable declarations. 
They can also be used in contexts like function arguments 
```

src/content/language/index.md
```diff
@@ -282,9 +282,33 @@ if (!yourPlanet.isGiant) {
...
+When the compiler can infer the enum type from the context, 
+you can use a more concise dot shorthand syntax to access enum values. 
+Instead of writing the full `EnumName.value`, 
+you can simply write `.value`. This can make your code cleaner 
+and easier to read.
+
+
+For example, when declaring a variable with an explicit enum type, 
+you can omit the Planet class name because 
+the type of Planet is already established:
```
parlough:
```suggestion
For example, when declaring a variable with an explicit type of `Planet`, 
you can omit the enum name because 
the type of `Planet` is already established:
```

src/content/language/index.md
```diff
@@ -282,9 +282,33 @@ if (!yourPlanet.isGiant) {
 }
 ```
 
+When the compiler can infer the enum type from the context, 
+you can use a more concise dot shorthand syntax to access enum values.
```
parlough:
```suggestion
you can use the more concise dot-shorthand syntax to access enum values. 
```

src/content/language/index.md
```diff
@@ -282,9 +282,33 @@ if (!yourPlanet.isGiant) {
 }
 ```
 
+When the compiler can infer the enum type from the context, 
+you can use a more concise dot shorthand syntax to access enum values. 
+Instead of writing the full `EnumName.value`, 
+you can simply write `.value`. This can make your code cleaner 
+and easier to read.
```
parlough: Following [this guideline from the style guide](https://developers.google.com/style/word-list#just):

> Sometimes, just is useful for conveying that one approach is simpler than another. In those cases, use just instead of simply.

```suggestion
you can just write `.value`. This can make your code cleaner 
and easier to read.
```

src/content/language/index.md
```diff
@@ -282,9 +282,33 @@ if (!yourPlanet.isGiant) {
 }
 ```
 
+When the compiler can infer the enum type from the context,
```
parlough, line 285: Perhaps we should just use "Dart" here as the analyzer does this inference to, but we don't have a single word to combine both tools?

```suggestion
When Dart can infer the enum type from the context, 
```

src/content/language/dot-shorthands.md
```diff
@@ -0,0 +1,266 @@
...
+// Invoking a const named constructor
+const Point myOrigin = .origin(); // Instead of Point.origin()
+
+// Using shorthands in a const collection literal
+const List<Point> keyPoints = [ .origin(), .new(1.0, 1.0) ];
+// Instead of [Point.origin(), Point(1.0, 1.0)]
+```
+
+
+## Advanced usage
```
parlough: "Advanced" is a bit overloaded a word, potentially misleading developers that this section is only for advanced developers. Would an alternative name like "Specialized scenarios" work here?

src/content/language/dot-shorthands.md
```diff
@@ -0,0 +1,266 @@
...
+// Factory constructor
+Point p1 = .fromList([1.0, 2.0]); // Instead of Point.fromList([1.0, 2.0])
+
+// Generic class constructor
+List<int> intList = .filled(5, 0); // Instead of List.filled(5, 0)
+``` 
+### Unnamed constructors
+The `.new` dot shorthand provides a concise way to call an
+unnamed constructor of a class. This is useful 
+for instantiating fields and variables where the type is
```
parlough: "Instantiating" made me a bit confused here as I was imagining it was talking about within a constructor declaration itself. How about "Assingning"

```suggestion
for assigning fields or variables where the type is 
```

src/content/language/dot-shorthands.md
```diff
@@ -0,0 +1,266 @@
...
+// Named constructor
+Point origin = .origin(); // Instead of Point.origin()
+
+// Factory constructor
+Point p1 = .fromList([1.0, 2.0]); // Instead of Point.fromList([1.0, 2.0])
+
+// Generic class constructor
+List<int> intList = .filled(5, 0); // Instead of List.filled(5, 0)
+``` 
+### Unnamed constructors
```
parlough, line 133:
```suggestion

### Unnamed constructors

```

src/content/language/dot-shorthands.md
```diff
@@ -0,0 +1,266 @@
...
+}
+
+// Example usage:
+var warnColor = colorCode(.warning); // Returns 'orange'
+```
+
+### Named constructors
+Dot shorthands are useful for invoking named constructors 
+or factory constructors. This syntax also works when providing 
+type arguments to a generic class's constructor.
```
parlough, line 108: It's not clear to me if this is intending to say that the syntax supports specifying type arguments or its able to infer the type arguments from the context type.

I'm assuming it's intending to say the latter?

src/content/language/extension-methods.md
```diff
@@ -2,8 +2,8 @@
 title: Extension methods
 description: Learn how to add to existing APIs.
 prevpage:
-  url: /language/enums
-  title: Enums
+  url: /language/dot-shorthands
+  title: Dot Shorthands
```
parlough:
```suggestion
  title: Dot shorthands
```

src/content/language/enums.md
```diff
@@ -6,8 +6,8 @@ prevpage:
   url: /language/mixins
   title: Mixins
 nextpage:
-  url: /language/extension-methods
-  title: Extension methods
+  url: /language/dot-shorthands
+  title: Dot Shorthands
```
parlough:
```suggestion
  title: Dot shorthands
```

examples/language/lib/classes/shorthand.dart
```diff
@@ -0,0 +1,172 @@
...
+// #docregion methods
+// Static method
+int httpPort = .parse('80'); // Instead of int.parse('80')
+
+// Static field/getter
+BigInt bigIntZero = .zero; // Instead of BigInt.zero
+// #enddocregion methods
+ 
+// #docregion constructors
+ class Point {
```
parlough: Since this file already has a `Point` class, it can't have a second one (or the third one later on). Consider splitting some of these examples to more focused files. Or I can do so for you if you prefer, just let me know!

parlough (commented):
Thanks for the hard work on this @connie-ooi! I love it so far :D 

I haven't looked through all of the code yet, but I wanted to leave some initial comments and questions until I have time to do so:

src/content/language/dot-shorthand.md
antfitch: Yes, I absolutely think we should do this. Spec: https://github.com/dart-lang/language/blob/main/accepted/3.10/dot-shorthands/feature-specification.md

src/content/language/dot-shorthands.md
```diff
@@ -0,0 +1,266 @@
...
+  url: /language/extension-methods
+  title: Extension methods
+---
+
+:::experimental
+Support for dot shorthand syntax is experimental and
+can only be used on the `main` channel.
+:::
+
+Write more concise Dart code by using the dot shorthand syntax `.foo`
```
antfitch:
```suggestion
## Overview

Write more concise Dart code by using the dot shorthand syntax `.foo` 
```

We should start adding Overview sections to our Dart topics. I know this isn't common right now, but without it, this information gets lost in the right navigation pane.

src/content/language/dot-shorthands.md
```diff
@@ -0,0 +1,266 @@
...
+
+## The role of context type 
+Dot shorthands use the context type to determine the member 
+the complier resolves to. The context type is the type that Dart 
+expects an expression to have based on its location. 
+For example, in `Status currentStatus = .running`, 
+the compiler knows a `Status` is expected, so it infers 
+`.running` to mean `Status.running`.
+
+## Using dot shorthand syntax
```
antfitch:
```suggestion
## Lexical structure and syntax

The dot shorthand feature introduces a new expression, the _static member shorthand_, which begins with a leading dot (`.`) and provides a concise way to access static members, constructors, and enum values when the compiler can infer the type from the surrounding code.
```

Avoid using -ing words in headings and stick to active voice. Because this is a reference, be as declarative as possible.

Always make sure to add an opening sentence or paragraph under headings.

Because I worked on the GoogleSQL project for years, I will probably be referencing it a lot. Here's an example of how headings are used in that reference guide: https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax

src/content/language/dot-shorthands.md
antfitch: Let's call this Rules and limitations.

```suggestion
## Rules and limitations

Add an intro sentence here.
```

src/content/language/dot-shorthands.md
```diff
@@ -0,0 +1,266 @@
...
+```dart
+// .fromCharCode(72) resolves to the String "H",
+// then the instance method .toLowerCase() is called on that String.
+String lowerH = .fromCharCode(72).toLowerCase();
+// Instead of String.fromCharCode(72).toLowerCase()
+
+print(lowerH); // Output: h
+```
+
+### Equality operators
```
antfitch:
```suggestion
### Asymmetric equality checks
```

src/content/language/dot-shorthands.md
```diff
@@ -0,0 +1,266 @@
...
+
+// Using shorthands in a const collection literal
+const List<Point> keyPoints = [ .origin(), .new(1.0, 1.0) ];
+// Instead of [Point.origin(), Point(1.0, 1.0)]
+```
+
+
+## Advanced usage
+
+### Chaining
```
antfitch:
```suggestion
### Clear context type required in chains
```

src/content/language/dot-shorthands.md
```diff
@@ -0,0 +1,266 @@
...
+  }
+
+  // ERROR: The type context is lost by casting `myColor` to `Object`.
+  // The compiler no longer knows that `.green` should refer to `Color.green`.
+  if ((myColor as Object) == .green) {
+    print('This will not compile.');
+  }
+}
+```
+
```
antfitch, line 317: I think we are missing a few rules here. I would add the following, but no need to add examples.

```suggestion
### Expression statements can't start with `.`

To avoid potential parsing ambiguities in the future, an
expression statement is not allowed to begin with a
`.` token.

### No generic constructor tear-offs with type arguments

You can't use type arguments on a `.new` dot shorthand to
create an instantiated tear-off of a constructor. This is 
consistent with the existing behavior for explicit constructor
tear-offs (`MyClass.new<int>`) and produces a
compile-time error.

### Limited declaration kinds

Dot shorthands is designed to work with types that have a 
clear static scope, namely `class`, `mixin`, `enum`, and
`extension` type declarations. This can be problematic for
some platform types (like `Function`) whose declaration
kind is not explicitly specified.

### Delayed resolution during compilation

Unlike regular static member access, which can be resolved
lexically before type inference, dot shorthands can only be
resolved after the compiler has inferred the types. This 
delayed resolution is a complication for the Dart
implementation and could potentially affect tooling, like
static analysis and tree-shaking, that might expect all static
targets to be known earlier.

### Limited handling of union types

While there is special handling for nullable types (`T?`) and
`FutureOr<T>`, support is limited.

*  For a nullable type (T?), you can access static members
   of `T`, but not of `Null`.

*  For `FutureOr<T>`, you can access static members of `T`
  (primarily to support `async` function returns), but you
  can't access static members of the `Future` class itself.
```

src/content/language/dot-shorthands.md
```diff
@@ -0,0 +1,266 @@
...
+const Point myOrigin = .origin(); // Instead of Point.origin()
+
+// Using shorthands in a const collection literal
+const List<Point> keyPoints = [ .origin(), .new(1.0, 1.0) ];
+// Instead of [Point.origin(), Point(1.0, 1.0)]
+```
+
+
+## Advanced usage
+
```
antfitch, line 225: I've taken a close look at the spec (https://github.com/dart-lang/language/blob/main/accepted/3.10/dot-shorthands/feature-specification.md) and used an LLM extension to grab the rules and caveats of dot shorthands notation. I've added a few things that were missing.

src/content/language/dot-shorthands.md
antfitch: @kallentu check my addition.

src/content/language/dot-shorthands.md
antfitch:
```suggestion
## Grammar

Write more concise Dart code by using the dot shorthand syntax `.foo` 
```

Let's turn this into a section called grammar. Add a paragraph like this somewhere:

```
In essence, the dot shorthands grammar allows an expression to start with one of the following and then optionally chain other operations onto it.

*   `.` and an identifier (`.myValue`)
*   Constructor (`.new()`)
*   Constant creation (`const .myValue()`)
```

antfitch (commented):
Looking good! I have a few structural changes and additions related to rules.

examples/language/lib/classes/shorthand.dart
conooi: I would appreciate if you could help me out with this @parlough! :D

src/content/language/dot-shorthands.md
kallentu: > No generic constructor tear-offs with type arguments

I wouldn't add this section if it's only about tear-offs, I think it's not very useful in practice since the only constructor tear-off that really works for dot shorthands is `Object x = .new`. You could mention it for general constructor invocations - "No generic constructor invocations with type arguments", and if you do need them to, then you'd put it in the context type.

> Limited declaration kinds

Dot shorthands still do work with `Function` (with the `apply(..)` static member). I'm not sure this section is useful to explain. It's mentioned in the spec as a point of discussion for how to technically define what a dot shorthand can be used on, but I think this only adds extra confusion. If we wanted to talk about the types that dot shorthands can't use (and keep this section somewhat), we could use `dynamic`, `void`, or function types as the example instead.

> Delayed resolution during compilation

This section in the spec was a warning to the implementation teams (me, I guess haha) that we need to change the implementation to handle this resolution after type inference, which typically, you could do before. I don't think this has any implications on our users and it's not something they need to know.

> Limited handling of union types

Nice, I think we should keep this!

examples/language/lib/classes/shorthand.dart
antfitch: I advise splitting this into multiple example files. It will be easier to track and move them around in the future.

src/content/language/dot-shorthands.md
```diff
@@ -0,0 +1,311 @@
...
+<?code-excerpt "language/lib/classes/shorthand_methods.dart"?>
+```dart
+// Static method
+int httpPort = .parse('80'); // Instead of int.parse('80')
+
+// Static field/getter
+BigInt bigIntZero = .zero; // Instead of BigInt.zero
+```
+
+### Constant expressions
```
antfitch, line 194:
```suggestion
### Constant expressions 

```

Make sure to add an empty line between headings and paragraphs. I've added an extra line here, but you'll need to with the other headings.

src/content/language/dot-shorthands.md
```diff
@@ -0,0 +1,311 @@
...
+    print('The color is not blue.');
+  }
+
+  // OK: The context for the ternary is the variable `inferredColor`
+  // being assigned to, which has a type of `Color`.
+  Color inferredColor = condition ? .green : .blue;
+  print('Inferred color is $inferredColor');
+}
+```
+The dot shorthand must be on the right-hand side of the `==`
```
antfitch, line 284:
```suggestion

The dot shorthand must be on the right-hand side of the `==` 
```

antfitch: Make sure to add a space between code samples and the next paragraph.

src/content/tools/analyzer-plugins.md
antfitch, line 1: This file snuck in there. You'll want to delete before merging this.

antfitch (approved):
LGTM! When @parlough has given final signoff, merge this ASAP. Great job!

src/content/tools/analyzer-plugins.md
conooi: Thanks for catching it!

src/content/language/dot-shorthands.md
```diff
@@ -0,0 +1,324 @@
...
+  url: /language/extension-methods
+  title: Extension methods
+---
+
+<?code-excerpt replace="/ *\/\/\s+ignore_for_file:[^\n]+\n//g; /(^|\n) *\/\/\s+ignore:[^\n]+\n/$1/g; /(\n[^\n]+) *\/\/\s+ignore:[^\n]+\n/$1\n/g"?>
+
+:::experimental
+Support for dot shorthand syntax is experimental and
+can only be used on the `main` channel.
+:::
```
parlough: This feature needs a language version note anyway (and will for a while), so you can replace this note with one:

```suggestion
:::version-note
Dot shorthands require a [language version][] of at least 3.10.
:::

[language version]: /resources/language/evolution#language-versioning
```

src/content/language/dot-shorthands.md
```diff
@@ -0,0 +1,324 @@
...
+  Point(this.x, this.y);
+  Point.origin() : x = 0, y = 0;
+}
+
+Point origin = .origin(); // Instead of Point.origin()
+```
+
+## The role of context type 
+
+Dot shorthands use the context type to determine the member
```
parlough: Consider cross linking the first usage of context type to its glossary entry: https://dart.dev/resources/glossary#context-type

```suggestion
Dot shorthands use the [context type][] to determine the member 
```

Then at the end of the section:

```
[context type]: /resources/glossary#context-type
```

src/content/language/index.md
parlough: It seems we tend to use the plural form to refer to the feature:

```suggestion

Dot shorthands aren't limited to variable declarations. 
They can also be used in contexts like function arguments 
```

src/content/language/index.md
parlough: Reposting this suggestion since I think it got lost:

```suggestion
For example, when declaring a variable with an explicit type of `Planet`, 
you can omit the enum name because 
the type of `Planet` is already established:
```

src/content/language/index.md
```diff
@@ -282,9 +282,33 @@ if (!yourPlanet.isGiant) {
...
+For example, when declaring a variable with an explicit enum type, 
+you can omit the Planet class name because 
+the type of Planet is already established:
+
+
+```dart
+// Instead of the full, explicit syntax:
+Planet myPlanet = Planet.venus;
+
+// You can use the dot shorthand:
```
parlough:
```suggestion
// You can use a dot shorthand:
```

src/content/language/index.md
parlough: Reposting since I think this suggestion got lost:

```suggestion
you can use the more concise dot-shorthand syntax to access enum values. 
```

src/content/language/dot-shorthands.md
parlough: Bumping this suggestion. Any thoughts here?

src/content/language/dot-shorthands.md
```diff
@@ -0,0 +1,324 @@
...
+already explicitly declared.
+
+This syntax is particularly effective for cleaning up 
+repetitive class field initializers. As shown in the 
+"After" example, it can be used for constructors both with 
+and without arguments. It also infers any generic type arguments 
+from the context. 
+
+Before:
+<?code-excerpt "language/lib/shorthands/unnamed_constructors.dart (unnamed-before)"?>
```
parlough: Consider using bold syntax for "before" and "after". Also consider being more specific about what these states are:

```suggestion
**Without dot shorthands:**

<?code-excerpt "language/lib/shorthands/unnamed_constructors.dart (unnamed-before)"?>
```

src/content/language/dot-shorthands.md
```diff
@@ -0,0 +1,324 @@
...
+  final GlobalKey<ScaffoldMessengerState> scaffoldKey =
+      GlobalKey<ScaffoldMessengerState>();
+
+  Map<String, Map<String, bool>> properties = <String, Map<String, bool>>{};
+  // ...
+}
+```
+
+After:
+<?code-excerpt "language/lib/shorthands/unnamed_constructors.dart (unnamed-after)" replace="/_PageStateAfter/_PageState/g;"?>
```
parlough:
```suggestion
**Using dot shorthands:**

<?code-excerpt "language/lib/shorthands/unnamed_constructors.dart (unnamed-after)" replace="/_PageStateAfter/_PageState/g;"?>
```

src/content/language/dot-shorthands.md
```diff
@@ -0,0 +1,324 @@
...
+String lowerH = .fromCharCode(72).toLowerCase();
+// Instead of String.fromCharCode(72).toLowerCase()
+
+print(lowerH); // Output: h
+```
+
+### Asymmetric equality checks
+
+The `==` and `!=` operators have a special rule for dot shorthands. 
+When dot shorthand syntax is used directly on the right-hand side (RHS)
```
parlough: Since the "RHS" acronym isn't used after this parenthetical, I think you can just remove it:

```suggestion
When dot shorthand syntax is used directly on the right-hand side
```

src/content/language/dot-shorthands.md
```diff
@@ -0,0 +1,324 @@
...
+  // OK: The context for the ternary is the variable `inferredColor`
+  // being assigned to, which has a type of `Color`.
+  Color inferredColor = condition ? .green : .blue;
+  print('Inferred color is $inferredColor');
+}
+```
+
+The dot shorthand must be on the right-hand side of the `==` 
+or `!=` operator. Comparing against a more complex expression, 
+like a ternary, is also not allowed.
```
parlough: We tend to not use the word ternary to describe this construct but rather "conditional expression".

```suggestion
like a conditional expression, is also not allowed.
```

src/content/language/dot-shorthands.md
```diff
@@ -0,0 +1,324 @@
...
+    print('This will not compile.');
+  }
+}
+```
+
+### Expression statements can't start with `.`
+
+To avoid potential parsing ambiguities in the future, an
+expression statement is not allowed to begin with a
+`.` token.
```
parlough, line 320: I think for this to be worth including we should add a basic example of what this would look like, otherwise readers might not know what it's referring to.

src/content/language/dot-shorthands.md
```diff
@@ -0,0 +1,324 @@
...
+
+### Limited handling of union types
+
+While there is special handling for nullable types (`T?`) and
+`FutureOr<T>`, support is limited.
+
+*  For a nullable type (T?), you can access static members
+ of `T`, but not of `Null`.
```
parlough:
```suggestion
*  For a nullable type (T?), you can access static members
   of `T`, but not of `Null`.
```

src/content/language/dot-shorthands.md
```diff
@@ -0,0 +1,324 @@
...
+  print('Inferred color is $inferredColor');
+}
+```
+
+The dot shorthand must be on the right-hand side of the `==` 
+or `!=` operator. Comparing against a more complex expression, 
+like a ternary, is also not allowed.
+
+<?code-excerpt "language/lib/shorthands/equality_with_errors.dart"?>
+```dart
```
parlough: You can add the `fails-sa` tag to clarify this example fails static analysis and won't compile successfully. 

````suggestion
```dart tag=fails-sa
````

File: examples/language/lib/shorthands/equality_with_errors.dart [hunk 71]
```diff
@@ -0,0 +1,28 @@
...
+  bool condition = true;
+
+  // ERROR: The shorthand must be on the right side of `==`.
+  // Dart's `==` operator is not symmetric for this feature.
+  // ignore: dot_shorthand_missing_context
+  if (.red == myColor) {
+    print('This will not compile.');
+  }
+
+  // ERROR: The right-hand side is a complex expression (a ternary),
```
parlough:
```suggestion
  // ERROR: The right-hand side is a complex expression (a conditional expression),
```

examples/language/lib/shorthands/enums.dart
```diff
@@ -0,0 +1,15 @@
+// Enum example
```
parlough: Since the introductory text for this snippet makes this clear, I think you can drop this comment:

```suggestion
```

examples/language/lib/shorthands/enums.dart
```diff
@@ -0,0 +1,15 @@
...
+enum LogLevel { debug, info, warning, error }
+
+// Function to get a color code based on log level
```
parlough: This one can directly be a [documentation comment](https://dart.dev/language/comments#documentation-comments):

```suggestion
/// Returns the color code to use for the specified log [level].
```

examples/language/lib/shorthands/intro.dart
```diff
@@ -0,0 +1,16 @@
+// Enum example
```
parlough:
```suggestion
// Enum example:
```

examples/language/lib/shorthands/intro.dart
```diff
@@ -0,0 +1,16 @@
...
+enum Status { none, running, stopped, paused }
+
+Status currentStatus = .running; // Instead of Status.running
+
+// Static method example
```
parlough:
```suggestion
// Static method example:
```

examples/language/lib/shorthands/intro.dart
```diff
@@ -0,0 +1,16 @@
...
+int port = .parse('8080'); // Instead of int.parse('8080')
+
+// Constructor example
```
parlough:
```suggestion
// Constructor example:
```

examples/language/lib/shorthands/static_members.dart
```diff
@@ -0,0 +1,5 @@
+// Static method
+int httpPort = .parse('80'); // Instead of int.parse('80')
+
+// Static field/getter
+BigInt bigIntZero = .zero; // Instead of BigInt.zero
```
parlough: Maybe we can expand the comments a bit here. Otherwise it's a bit ambigious if the comment is trying to say the following line is a static method or usage of one.

How about something sort-of like this?

```suggestion
// Use dot-shorthand syntax to invoke a static method:
int httpPort = .parse('80'); // Instead of int.parse('80')

// Use dot-shorthand syntax to access a static field or getter:
BigInt bigIntZero = .zero; // Instead of BigInt.zero
```

Then applying whatever structure you use here to the other example files as well.

src/content/language/dot-shorthands.md
```diff
@@ -0,0 +1,324 @@
...
+the compiler knows a `Status` is expected, so it infers 
+`.running` to mean `Status.running`.
+
+## Lexical structure and syntax
+
+The dot shorthand feature introduces a new expression, 
+the _static member shorthand_, which begins with a leading dot (`.`) 
+and provides a concise way to access static members, 
+constructors, and enum values when the compiler can infer 
+the type from the surrounding code.
```
parlough: This phrasing is good for an announcement blog post or similar, but hopefully this documentation can be timeless. So can we reword this to avoid describing dot shorthands as a "feature" or something being introduced and instead just talk about it as a part of the language.

Also the sentence is a bit long, so it might benefit from being split into two.

Do whatever you think is best, but a rough suggestion could be:

```suggestion
A _static member shorthand_ is an expression that
begins with a leading dot (`.`).
When the type is known from the surrounding context
this syntax provides a concise way to access
static members, constructors, and enum values.
```

src/content/language/dot-shorthands.md
```diff
@@ -0,0 +1,324 @@
...
+---
+
+<?code-excerpt replace="/ *\/\/\s+ignore_for_file:[^\n]+\n//g; /(^|\n) *\/\/\s+ignore:[^\n]+\n/$1/g; /(\n[^\n]+) *\/\/\s+ignore:[^\n]+\n/$1\n/g"?>
+
+:::experimental
+Support for dot shorthand syntax is experimental and
+can only be used on the `main` channel.
+:::
+
+## Grammar
```
parlough: Later on there's a section called "Lexical structure and syntax", which has mostly the same meaning as "grammar". Perhaps this one should just be called "Overview" or "Introduction"?

src/content/language/dot-shorthands.md
```diff
@@ -0,0 +1,324 @@
...
+to start with one of the following and then optionally chain 
+other operations onto it:
+
+* and an identifier `.myValue`
+
+* Constructor `.new()`
+
+* Constant creation `const .myValue()`
+
+Here’s a quick look at how it simplifies an enum assignment:
```
parlough:
```suggestion
Here's a quick look at how it simplifies an enum assignment:
```

src/content/language/dot-shorthands.md
```diff
@@ -0,0 +1,324 @@
...
+code by omitting the type when the compiler can infer it 
+from context. This provides a clean alternative to writing 
+the full `ContextType.foo` when accessing enum values, 
+static members, or constructors.
+
+In essence, the dot shorthands grammar allows an expression 
+to start with one of the following and then optionally chain 
+other operations onto it:
+
+* and an identifier `.myValue`
```
parlough: What is this "and" referring to? Perhaps this was meant to just be "An identifier" or just "Identifier"?

src/content/language/dot-shorthands.md
```diff
@@ -0,0 +1,324 @@
...
+
+## Grammar
+
+Dot shorthand syntax `.foo` lets you write more concise Dart 
+code by omitting the type when the compiler can infer it 
+from context. This provides a clean alternative to writing 
+the full `ContextType.foo` when accessing enum values, 
+static members, or constructors.
+
+In essence, the dot shorthands grammar allows an expression
```
parlough: Does it make sense to just say "dot shorthands" here, dropping the "grammar"? We don't mention the lower-level concept of a grammar a lot (or at all maybe) in the `/language` docs.

```suggestion
In essence, dot shorthands allow an expression 
```

parlough (approved):
I love it, thank you @connie-ooi!

I pushed some fixes and clean up to the excerpt sources, so make sure to run `git pull` before working on any more local changes. I also recommend checking my commits to see the changes I made.

I have a few suggestions and questions remaining, but feel free to push back on anything you don't think makes sense or file an issue if you prefer to address something after this lands. Thanks again! Let me know if you have any questions or issues :D 

_Note some review comments are hidden by default. I think some of the comments from my last review got lost from that. In the middle of all the comments, there should be a "Load more..." button you can click to see them all._

src/content/language/dot-shorthands.md
```diff
@@ -0,0 +1,347 @@
...
+the compiler knows a `Status` is expected, so it infers 
+`.running` to mean `Status.running`.
+
+[context type]: /resources/glossary#context-type
+
+## Lexical structure and syntax
+
+A _static member shorthand_ is an expression that
+begins with a leading dot (`.`).
+When the type is known from the surrounding context
```
parlough:
```suggestion
When the type is known from the surrounding context,
```

src/content/language/dot-shorthands.md
```diff
@@ -0,0 +1,347 @@
...
+
+The `.new` dot shorthand provides a concise way to call an
+unnamed constructor of a class. This is useful 
+for assigning fields or variables where the type is 
+already explicitly declared.
+
+This syntax is particularly effective for cleaning up 
+repetitive class field initializers. As shown in the 
+"After" example, it can be used for constructors both with 
+and without arguments. It also infers any generic type arguments
```
parlough: I forgot this text referenced the following samples. Since we changed their labels, we need to update "After" here too.

Does this work?

```suggestion
repetitive class field initializers.
As shown in the following "after" example using dot shorthands,
it can be used for constructors both with and without arguments.
tt also infers any generic type arguments 
```

PR #6953, opened 2025-10-17
Compile client code with dart2wasm

simolus3:
This configures `build_web_compilers` to use ddc on dev builds (`dash_site serve`) while compiling with both `dart2js` and `dart2wasm` on release modes (`dash_site build`, or the added `--release` flag to `serve` to experiment with this more easily).

Using jaspr CLI options to switch to dart2wasm would _only_ compile sources with `dart2wasm`, this uses both compilers and a feature detection snippet emitted by `dart2wasm` to determine which variant is used. That seems to be in spirit of https://github.com/dart-lang/site-www/issues/6854.

To stop jaspr from overriding options passed to `build_web_compilers`, one needs to use the `--no-managed-build-options` flag. A downside is that the `PRODUCTION` define set on release modes would not be forwarded to client code any longer. The constant is only used in the `robots.txt` generator so this should be fine, forwarding this to client code would require more changes to `jaspr_cli`.

Closes https://github.com/dart-lang/site-www/issues/6854.

### DIFF
--- a/site/build.yaml
+++ b/site/build.yaml
@@ -22,4 +22,6 @@ targets:
             dart2js:
               args: [-O4, -Djaspr.flags.release=true]
             dart2wasm:
-              args: [-O4, -Djaspr.flags.release=true]
+              # Use -O2 to catch more potential runtime issues, it is not much
+              # slower than -O4: https://github.com/dart-lang/site-www/pull/6953#discussion_r2453318668
+              args: [-O2, -Djaspr.flags.release=true]


kevmoo:
Ooooo!!!

site/build.yaml
```diff
@@ -7,3 +7,19 @@ builders:
...
+          compilers:
+            dartdevc:
+        release_options:
+          # Use production compilers for release builds, and ensure
+          # kReleaseMode in jaspr is enabled.
+          compilers:
+            dart2js:
+              args: [-O4, -Djaspr.flags.release=true]
+            dart2wasm:
+              args: [-O4, -Djaspr.flags.release=true]
```
kevmoo: We should ponder `-O2` – it'll make sure we catch more runtime issues and we've noticed very little impact to code size and speed!

kevmoo: thanks!

PR #6961, opened 2025-10-23
Add blog link to site switcher

parlough:
This makes it more consistent with other Dart-related site properties. It is still linked to from the top navigation as well.

**Screenshot with the added button:**
Screenshot of the added link button to the blog.


site/lib/src/components/layout/header.dart
```diff
@@ -69,16 +69,6 @@ class DashHeader extends StatelessComponent {
               ],
             ),
           ]),
-          li([
```
antfitch, line 72: Let's keep the link to the blog in this top menu.

parlough: Sure, I added it back to the header.

antfitch:
Hi, I feel very strongly that we need the blog at the top and not hidden, so this is a change I'd like for us not to make. If there is a project plan somewhere with a larger vision, let's review it.

parlough:
> Hi, I feel very strongly that we need the blog at the top and not hidden, so this is a change I'd like for us not to make. If there is a project plan somewhere with a larger vision, let's review it.

@antfitch I went ahead and added it back to the header. Do you think it's fine to live there and in the site switcher? I was hoping to have it there for consistency.

PR #6962, opened 2025-10-24
Keep track of active site section in sidenav buttons

parlough:
Similar to the top navigation menu on widescreen layouts, keep track of the current active site section on the narrow-screen version of the sidenav.

This provides mobile users better insight into where on the site they are and will helps prepare for multiple sidenav support.

---

**After:**

Screenshot of the new buttons highlighting the Docs section<br>

**Before:**

Screenshot of the old buttons with no highlighting

### DIFF
--- a/site/lib/src/components/layout/sidenav.dart
+++ b/site/lib/src/components/layout/sidenav.dart
@@ -53,14 +53,24 @@ final class DashSideNav extends StatelessComponent {
         _TopNavItem(
           href: '/docs',
           label: 'Docs',
-          iconId: 'article',
+          iconId: 'docs',
           active: activeEntry == ActiveNavEntry.docs,
+        ),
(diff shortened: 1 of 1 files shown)


parlough:
@antfitch Thanks for taking a look!

> I like the addition of the icons, but I think we need to revisit what's in our top navigation pane. Let's sync to figure out what should go in that top menu moving forward.

Do you mind if we land this in the current state? A next step (which I'll discuss as part of a design) will be getting rid of the top navigation in its current form. Landing this will help clear the way for making that possible alongside some other design fixes before then.

antfitch:
Yes, we can land this in its current state, but before we do so, can you add Blog and Community into the drop down? That way nothing gets lost when we switch from horizontal to vertical top navigation.

PR #6969, opened 2025-10-27
feat: Add tooltips to glossary term links with their definition

schultek:
Shows a tooltip for all glossary links with their term and definition. Glossary links are styled differently than normal links to give a visual indication to the visitor.

- On non-touch devices: shows the tooltip on hover.
- On touch devices: toggles the tooltip on click.

Always positions the tooltip inside the content container with no horizontal overflow.

Bildschirmfoto 2025-10-27 um 14 48 08

### DIFF
--- a/site/lib/_sass/components/_glossary.scss
+++ b/site/lib/_sass/components/_glossary.scss
@@ -13,7 +13,6 @@ body.glossary-page main {
       min-height: 8rem;
 
       .initial-content {
-
         // Only show the first paragraph if collapsed.
         > :not(:first-child) {
           display: none;
@@ -30,7 +29,6 @@ body.glossary-page main {
     }
 
     .expand-button {
-
       &:hover,
       &:focus-within {
         transition: transform .25s ease-out;
--- a/site/lib/src/extensions/glossary_link_processor.dart
+++ b/site/lib/src/extensions/glossary_link_processor.dart
@@ -83,7 +83,15 @@ class GlossaryTooltip extends StatelessComponent {
       span(classes: 'tooltip-content', [
         text(entry.shortDescription),
         text(' '),
-        a(href: '/resources/glossary#${entry.id}', [text('Learn more')]),
+        a(
+          href: '/resources/glossary#${entry.id}',
+          attributes: {
+            'title':
+                'Learn more about \'${entry.term}\' and '
+                'find related resources.',
+          },
+          [text('Learn more')],
+        ),
       ]),
     ]);
   }


parlough:
I love the functionality @schultek! Thanks :D 

What about keyboard only users though? Would it make sense to also show the tooltip if the link is focused?

schultek:
Good catch, I will add focus support as well

site/lib/_sass/components/_tooltip.scss
parlough, line 1: Can you update this file to use 2 space indentation?

site/lib/src/extensions/glossary_link_processor.dart
```diff
@@ -0,0 +1,89 @@
...
+
+        processedNodes.add(
+          ElementNode(
+            'span',
+            {'class': 'tooltip-wrapper'},
+            [
+              ElementNode('a', {
+                ...node.attributes,
+                'class': '${node.attributes['class'] ?? ''} tooltip-target'
+                    .trim(),
```
parlough:
```suggestion
                'class': [
                  ?node.attributes['class'],
                  'tooltip-target',
                ].toClasses,
```

site/lib/src/extensions/glossary_link_processor.dart
```diff
@@ -0,0 +1,89 @@
...
+  final GlossaryEntry entry;
+
+  @override
+  Component build(BuildContext context) {
+    return span(classes: 'tooltip', [
+      span(classes: 'tooltip-header', [text(entry.term)]),
+      span(classes: 'tooltip-content', [
+        text(entry.shortDescription),
+        text(' '),
+        a(href: '/resources/glossary#${entry.id}', [text('Learn more')]),
```
parlough: Should these two be a div?

```suggestion
    return div(classes: 'tooltip', [
      span(classes: 'tooltip-header', [text(entry.term)]),
      div(classes: 'tooltip-content', [
        text(entry.shortDescription),
        text(' '),
        a(href: '/resources/glossary#${entry.id}', [text('Learn more')]),
```

site/lib/src/extensions/glossary_link_processor.dart
```diff
@@ -0,0 +1,89 @@
...
+  const GlossaryTooltip({required this.entry});
+
+  final GlossaryEntry entry;
+
+  @override
+  Component build(BuildContext context) {
+    return span(classes: 'tooltip', [
+      span(classes: 'tooltip-header', [text(entry.term)]),
+      span(classes: 'tooltip-content', [
+        text(entry.shortDescription),
```
parlough, line 84: Consider using `DashMarkdown` with `inline: true` here. Some of the  short descriptions have inline Markdown.

site/lib/src/extensions/glossary_link_processor.dart
```diff
@@ -0,0 +1,89 @@
...
+  @override
+  Future<List<Node>> apply(Page page, List<Node> nodes) async {
+    final glossary = Glossary.fromList(page.data['glossary'] as List<Object?>);
+    return _processNodes(nodes, glossary);
+  }
+
+  List<Node> _processNodes(List<Node> nodes, Glossary glossary) {
+    final processedNodes = <Node>[];
+
+    for (var i = 0; i < nodes.length; i++) {
```
parlough: Can this just be a for-each loop?

site/lib/_sass/components/_tooltip.scss
```diff
@@ -0,0 +1,57 @@
+.tooltip-wrapper {
+    position: relative;
+
+    a.tooltip-target {
+        color: inherit;
+        text-decoration: underline dotted;
+    }
+
+    .tooltip {
```
parlough: If the originally tooltip target was in a bold or italic span, it ends up affecting the contents of the tooltip as well.

Perhaps we should reset the font weight and style here?

site/lib/_sass/components/_tooltip.scss
```diff
@@ -0,0 +1,57 @@
...
+        visibility: hidden;
+
+        display: flex;
+        position: absolute;
+        z-index: 100;
```
parlough: Consider using on the of the z indices from `_root.scss`:

```suggestion
        z-index: var(--site-z-dropdown);
```

site/lib/src/extensions/glossary_link_processor.dart
parlough: Can you add a new `tooltip-footer` section for the link so it has additional separation from the content?

site/lib/_sass/components/_tooltip.scss
```diff
@@ -0,0 +1,57 @@
...
+        }
+    }
+
+    // On non-touch devices, show tooltip on hover or focus.
+    @media all and not (pointer: coarse) {
+        &:hover .tooltip {
+            visibility: visible;
+        }
+
+        &:has(.tooltip-target:focus) .tooltip {
```
parlough: Can this just use [`:focus-within`](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-within) that way it is also visible if focus is within the tooltip?

```suggestion
        &:focus-within .tooltip {
```

parlough (commented):
This is super exciting. Thanks for working on this @schultek!

Works super well! I just have some small suggestions and questions.

site/lib/src/extensions/glossary_link_processor.dart
schultek: They can't because they are inside a `<p>`.

site/lib/src/extensions/glossary_link_processor.dart
schultek: Seems like `inline` is not implemented.

site/lib/_sass/components/_tooltip.scss
schultek: TIL

site/lib/src/extensions/glossary_link_processor.dart
schultek: I though it's nicer to be inline with the content.

site/lib/_sass/components/_tooltip.scss
schultek: Done

site/lib/_sass/components/_tooltip.scss
schultek: Done

site/lib/src/extensions/glossary_link_processor.dart
schultek: Done

site/lib/_sass/components/_tooltip.scss
schultek: Done

site/lib/src/extensions/glossary_link_processor.dart
schultek: Done

site/lib/_sass/components/_tooltip.scss
schultek: Done

site/lib/_sass/components/_tooltip.scss
```diff
@@ -0,0 +1,61 @@
+.tooltip-wrapper {
+  position: relative;
+
+  a.tooltip-target {
+    color: inherit;
+    text-decoration: underline dotted;
```
parlough: I don't know why, but this seems to not work on Safari unless you separate them:

```suggestion
    text-decoration: underline;
    text-decoration-style: dotted;
```

site/lib/_sass/components/_tooltip.scss
```diff
@@ -0,0 +1,61 @@
...
+    }
+  }
+
+  // On touch devices, show tooltip on click (see global_scripts.dart).
+  @media all and (pointer: coarse) {
+    .tooltip.visible {
+      visibility: visible;
+    }
+  }
+}
```
parlough:
```suggestion
}

```

site/lib/_sass/components/_glossary.scss
```diff
@@ -77,4 +81,4 @@ body.glossary-page main {
     align-items: center;
     gap: 0.25rem;
   }
-}
+}
```
parlough:
```suggestion
}

```

site/lib/src/client/global_scripts.dart
```diff
@@ -543,3 +544,93 @@ void _setUpReleaseTags() {
   fetchVersion('beta');
   fetchVersion('dev');
 }
+
+void _setUpTooltips() {
+  final tooltipWrappers = web.document.querySelectorAll('.tooltip-wrapper');
+
+  final isTouchscreen = web.window.matchMedia('(pointer: coarse)').matches;
+
+  void setup(bool setUpClickListener) {
```
parlough: Consider making this is a named parameter, as otherwise it's not  clear (without inline hints) what the `true` or `false` do.

parlough (approved):
Thanks for making those adjustments @schultek! Looks good to me after some small suggestions:

schultek:
@parlough all done

parlough:
@schultek This still looks good to me, but before we land it, can you investigate a pre-existing issue? It seems navigating to the full glossary terms scrolls to the wrong section of the page (which is a bit confusing from the learn more links). Without looking, I'm guessing because they are being closed on load and the location of the fragment is changing?

site/lib/src/extensions/glossary_link_processor.dart
```diff
@@ -0,0 +1,90 @@
...
+  final GlossaryEntry entry;
+
+  @override
+  Component build(BuildContext context) {
+    return span(classes: 'tooltip', [
+      span(classes: 'tooltip-header', [text(entry.term)]),
+      span(classes: 'tooltip-content', [
+        text(entry.shortDescription),
+        text(' '),
+        a(href: '/resources/glossary#${entry.id}', [text('Learn more')]),
```
parlough: Can you add a html `title` to the link with some more detailed text like "Learn more about 'term' and find related resources."

schultek: Not sure what you mean? `<title>` sets the document title.

schultek:
> It seems navigating to the full glossary terms scrolls to the wrong section of the page (which is a bit confusing from the learn more links). Without looking, I'm guessing because they are being closed on load and the location of the fragment is changing?

Fixed

parlough:
Thanks for making that adjustment! Everything looks great :D

PR #6982, opened 2025-10-30
Wrap the page header and add share button

parlough:
Begin each doc with a wrapped card of the breadcrumbs (if they exist), the page title, the page description, and a more options button to share the page.

Example on [`/language/functions`](https://dart-dev--pr6982-feat-wrapped-page-header-uenifb34.web.app/language/libraries):

Screenshot of wrapped page header

As follow-up, we can explore adding other operations to the more options dropdown, such as "View as Markdown", "Open in AI Studio", etc.


antfitch:
I think this looks great. How easy would it be to add a 'Summary' to the right-pane TOC when there isn't an H2 heading with this title? My reason is that the summary often has really great content, but there is no way to get to it through the right TOC. This is a problem in many docsets, not just Dart & Flutter.

PR #7010, opened 2025-11-11
[3.10] Update SDK constraints and documented Dart version

parlough:


pubspec.yaml
```diff
@@ -3,7 +3,7 @@ publish_to: none
 homepage: https://dart.dev
 
 environment:
-  sdk: ^3.9.0
+  sdk: ^3.10.0-0
```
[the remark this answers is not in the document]
parlough: We can't as the beta 3.10 builds are still a prerelease. Once the Dart and Flutter beta channels have moved to 3.11 or 3.10.0 stable, we can simplify the constraint to 3.10.

site/lib/src/markdown/alert_syntax.dart
```diff
@@ -157,7 +157,8 @@ enum _AlertType {
   ),
   secondary(
     cssClass: 'alert-secondary',
-  );
+  )
+  ;
```
kevmoo, line 161: does this mean the trailing semicolon is not needed?

parlough: The trailing semicolon is still needed since this enum has a class definition.

However, I guess a trailing comma wasn't needed for `dart format` to move the semicolon to the next line. So I went ahead and [added the trailing commas](https://github.com/dart-lang/site-www/pull/7010/commits/19590ef7179cd9d7f3d4732c47354549c89a6386) now :)

PR #7026, opened 2025-11-18
Overhaul lint details setup

parlough:
Intermediate step as part of https://github.com/dart-lang/site-www/issues/7027. **This change doesn't result in any changes to the site output.**

This PR moves the generation of a new `src/data/lint-info.json` file to the site rather than rely on the deprecated `linter_rules.json` previously generated in the SDK. To accomplish this, temporarily duplicate the descriptions of each lint until they can be accessed from `messages.yaml` or some other data file.


antfitch (approved):

sfshaza2:
@parlough, can we either finish this PR or close it?

PR #7040, opened 2025-11-23
add more examples to package:web migration

fsw:
I think adding a few more examples and info will help with the migration process.

---

### DIFF
--- a/src/content/interop/js-interop/package-web.md
+++ b/src/content/interop/js-interop/package-web.md
@@ -160,7 +160,7 @@ context.
 
 `Element` is now `HTMLElement` but the extension type `Element` also exists as
 a base type for `HTMLElement`, `SVGElement`, etc.
-See [Element MDN documentation].
+See [Element MDN documentation][].
 `querySelector` returns an Element as it can return other subtypes of `Element`
 besides `HTMLElement` like `SVGElement`. A downcast to `HTMLElement` is needed
 if you need to access its methods.


antfitch (changes requested):
Will review after our technical reviewers have finished first.

IldySilva (approved):

src/content/interop/js-interop/package-web.md
```diff
@@ -156,6 +156,78 @@ Node append(Node node) native;
 `native` is an internal keyword that means the same as `external` in this
 context.
 
+### Element and HTMLElement
+
+`Element` is now `HTMLElement` but `Element` class also exists as base class for
```
srujzs: nit: "but the extension type `Element` also exists as a base type for `HTMLElement`, `SVGElement`, etc."

src/content/interop/js-interop/package-web.md
```diff
@@ -156,6 +156,78 @@ Node append(Node node) native;
...
+HTMLElement, SVGElement etc. See [mdn element].
```
srujzs: nit: Let's expand the link alias so it flows better e.g. "See [the Element documentation]."

src/content/interop/js-interop/package-web.md
```diff
@@ -156,6 +156,78 @@ Node append(Node node) native;
 `native` is an internal keyword that means the same as `external` in this
 context.
 
+### Element and HTMLElement
```
srujzs: nit: Probably want to backtick these.

src/content/interop/js-interop/package-web.md
```diff
@@ -156,6 +156,78 @@ Node append(Node node) native;
...
+
+`Element` is now `HTMLElement` but `Element` class also exists as base class for
+HTMLElement, SVGElement etc. See [mdn element].
+`querySelector` returns `Element` as it can query in SVG as well so cast is
```
srujzs: I think we can reword this to something like:

"`querySelector` returns an `Element` as it can return other subtypes of `Element` besides `HTMLElement` like `SVGElement`. A downcast to `HTMLElement` is needed if you need to access its methods."

----

This is sort of self-explanatory with how types/extension types works, but I think it's okay being explicit because of the type differences between `package:web` and `dart:html`.

src/content/interop/js-interop/package-web.md
```diff
@@ -156,6 +156,78 @@ Node append(Node node) native;
...
+required to access `HTMLElement` methods.
+
+```dart
+element.querySelector('#selectme')!.className = 'test'; // valid in both
+
+element.querySelector('#selectme')!.style.color = 'red'; // Remove
```
srujzs, line 171: I don't quite remember the style in the site docs, but I believe you can label an example as "bad" or "good". We should maybe do that instead of the comments. The same applies everywhere else. Maybe @antfitch might know more on how to format that.

src/content/interop/js-interop/package-web.md
```diff
@@ -156,6 +156,78 @@ Node append(Node node) native;
...
+(element.querySelector('#selectme') as HTMLElement).style.color = 'red'; // Add
+```
+
+### List operations
+
+Methods like `Element::querySelectorAll` or `Element::children` return values no
```
srujzs: Slight reword:

"Unlike `dart:html`, `package:web` methods like `Element.querySelectorAll` and `Element.children` return values that don't implement `List`. If a `List` is required, you'll need to wrap it with a class. For immutable operations, ..."

src/content/interop/js-interop/package-web.md
```diff
@@ -156,6 +156,78 @@ Node append(Node node) native;
...
+for (final child in parent.children) {} // Remove
+for (final child in JSImmutableListWrapper(parent.children)) {} // Add
+```
+
+You need to add implementation for simple mutable operations like `removeWhere`:
+
+```dart
+parent.children.removeWhere(test); // Remove
+
+for (var i = parent.children.length - 1; i >= 0; --i) {
```
srujzs, line 197: Unrelated, but it might be beneficial for us (the `package:web` owners) to have a mutable list implementation to support things like this.

src/content/interop/js-interop/package-web.md
```diff
@@ -156,6 +156,78 @@ Node append(Node node) native;
...
+  if (test(parent.children.item(i)!)) {
+    parent.children.item(i)!.remove();
+  }
+}  // Add
+
+```
+
+### Common DOM manipulation examples
+
+All non standard `dart:html` fields and helpers not present in IDL definitions
```
srujzs: nit: I think we can just reword this to something like:

Here are some common examples of trivial changes that need to be made when migrating from `dart:html` to `package:web`:

src/content/interop/js-interop/package-web.md
fsw: fixed

src/content/interop/js-interop/package-web.md
fsw: fixed

src/content/interop/js-interop/package-web.md
fsw: sure, fixed

src/content/interop/js-interop/package-web.md
fsw: I matched other examples on this page. I think this makes sense as good/bad are clear for conventions while here we are migrating so both lines are technically "good" but in different libraries.

src/content/interop/js-interop/package-web.md
fsw: fixed

src/content/interop/js-interop/package-web.md
fsw: Yes, I was also thinking if it would be better to implement this instead of documenting. I will try to open a PR to package:web later and link this PR to update this example. In my code base I used an extension to fix this. I am not sure if any other method than removeWhere would be usable here though.

src/content/interop/js-interop/package-web.md
fsw: fixed

src/content/interop/js-interop/package-web.md
srujzs: > I will try to open a PR to package:web later

Sounds great, thank you!

> if any other method than removeWhere would be usable here though.

On reflection, `children` was exposed as a List in `dart:html` but `HTMLCollection` is not a list-like, so we'd need something bespoke. An alternative is to have something specific for `children` instead. I'm also not sure if there are any examples of mutable list-likes in the browser to warrant a general mutable List implementation.

src/content/interop/js-interop/package-web.md
srujzs: > I matched other examples on this page.

Maybe I should update the other examples. :) But acknowledged, either way is fine for me.

src/content/interop/js-interop/package-web.md
fsw: Indeed, it seems general implementation is not possible. I have worked out something like this
https://github.com/dart-lang/web/pull/490 to work with children and childNodes.

antfitch (approved):

PR #7056, opened 2025-12-04
Changelog System for Dart Docs

antfitch:
**Description:**

This PR refactors the changelog implementation to use a hybrid static/dynamic architecture and migrates the Dart SDK Chagelog, Breaking Changes page, Language Evolution page to this new system.

**Preview:**

+ Changelog page: https://dart-dev--pr7056-breaking-changes-migration-f9xiaksk.web.app/changelog
+ Language versioning page: https://dart-dev--pr7056-breaking-changes-migration-f9xiaksk.web.app/language/versions

**Tasks before we merge:**

- [x] Parker to review the new Changelog section & look for gaps
- [x] Add Changelog to the TOC
- [x] Remove from the TOC, but do not delete the topics: `Dart SDK Changelog`, `Breaking Changes`, `Language Evolution`
- [x] Add a deprecated message to the top of the topics with a link to the new Changelog: `Dart SDK Changelog`, `Breaking Changes`, `Language Evolution`
- [x]  Moved the language versioning content from `Language Evolution` to `Language versioning`

**Tasks left after we merge**
- [ ] Tasks in https://github.com/dart-lang/site-www/pull/7205
- [ ] 6 months after we launch the Changelist, delete the old topics & update shortcuts in firebase.json.


site/lib/_sass/base/_utils.scss
```diff
@@ -18,3 +18,7 @@
 .text-center {
   text-align: center;
 }
+
+.hidden {
+  display: none !important;
+}
```
kevmoo: baby kittens love ending files with a trailing newline 🐱

site/lib/src/components/pages/changelog/changelog_index.dart
```diff
@@ -0,0 +1,109 @@
...
+  const _ChangelogEntryCard(this.entry);
+
+  final ChangelogEntry entry;
+
+  @override
+  Component build(BuildContext context) {
+    // Generate a somewhat unique ID for DOM filtering
+    final uniqueId = slugify(
+      '${entry.version}-${entry.area}-${entry.subArea ?? ''}-${entry.description.substring(0, min(20, entry.description.length))}',
+    );
```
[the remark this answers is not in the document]
antfitch: updated.

site/lib/_sass/pages/_changelog-index.scss
```diff
@@ -0,0 +1,676 @@
...
+  .filter-header {
+    display: block;
+  }
+
+  #changelog-filter-group-wrapper {
+    position: fixed;
+    top: 11rem;
+    bottom: 0;
+    right: -220px;
+    width: 220px;
```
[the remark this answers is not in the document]
antfitch: updated.

antfitch:
This is ready for review. @parlough, @ericwindmill, and @kevmoo if you need any changes, let me know.

kevmoo:
I didn't review the code, but I noticed the staged version:

image

See the double title thingy?

site/lib/main.dart
```diff
@@ -103,4 +103,12 @@ List<CustomComponent> get _embeddableComponents => [
...
     builder: (_, attrs, child) => Card.fromAttributes(attrs, child),
   ),
+  CustomComponent(
+    pattern: RegExp('ChangelogIndex', caseSensitive: false),
+    builder: (_, _, _) => const ChangelogIndex(),
+  ),
+  // CustomComponent(
+  //   pattern: RegExp('LearningResourceIndex', caseSensitive: false),
+  //   builder: (_, _, _) => LearningResourceIndex(),
+  // ),
```
ericwindmill: We shouldn't leave commented out code in the codebase

site/lib/src/models/changelog_model.dart
```diff
@@ -0,0 +1,137 @@
...
+  newTag('new', 'New'),
+  breaking('breaking', 'Breaking'),
+  fixed('fixed', 'Fixed'),
+  changed('changed', 'Changed'),
+  experimental('experimental', 'Experimental'),
+  versioned('versioned', 'Language versioned'),
+  deprecated('deprecated', 'Deprecated'),
+  removed('removed', 'Removed'),
+  none('none', 'None')
+  ;
```
ericwindmill, line 126: nit: This semicolon on it's own line is weird? Maybe run `dart fmt`

site/lib/_sass/pages/_changelog-index.scss
```diff
@@ -0,0 +1,762 @@
...
+  z-index: 1000; // Ensure it's on top of everything
+}
+
+.legend-dialog {
+  background-color: var(--site-base-bgColor);
+  border-radius: 8px;
+  box-shadow: 0 4px 24px rgba(0, 0, 0, 0.2);
+  width: 90%;
+  max-width: 600px;
+  max-height: 90vh;
```
ericwindmill, line 591: I think the max-width should have a larger max width, it looks cramped when the desktop window is full screen.

site/lib/_sass/pages/_changelog-index.scss
```diff
@@ -0,0 +1,762 @@
...
+    overflow-y: auto;
+
+    h4 {
+      margin-top: 1.5rem;
+      margin-bottom: 1rem;
+    }
+
+    dl {
+      display: grid;
+      grid-template-columns: max-content 1fr;
```
ericwindmill: Similar to my above comment, I think this should be something like `2fr 3fr`. The content looks cramped because the first column takes up most of the space, but has less content.  Alternatively, you could explicitly make the content in column 1 be on two lines, so the `max-content` width isn't so long.

site/lib/_sass/pages/_changelog-index.scss
```diff
@@ -0,0 +1,762 @@
...
+    padding: 1.5rem;
+    overflow-y: auto;
+
+    h4 {
+      margin-top: 1.5rem;
+      margin-bottom: 1rem;
+    }
+
+    dl {
+      display: grid;
```
ericwindmill, line 633: The legend popup is pretty is hard to read on mobile. Consider using `grid-template-area` to make the grid shape change based on the width of the screen. For example, the "type of change" could be listed above the paragraph of explainer text on mobile, and in columns (as it is now) on larger screens

ericwindmill (changes requested):
I left a few comments, all of which are small and/or just style opinions. 

I am not a web expert or Jaspr expert, so @parlough's opinion would be valuable here.

antfitch:
Let's choose a data serialization strategy for Dash sites before committing this.

site/lib/src/components/pages/changelog/breaking_changes_legend.dart
```diff
@@ -0,0 +1,95 @@
...
+            ),
+            text('.'),
+          ]),
+        ]),
+      ]),
+    ]);
+  }
+}
+
+// Helper functions for definition list elements if not available in jaspr
```
parlough: In the latest release (v0.22) of Jaspr (which we use after https://github.com/dart-lang/site-www/commit/97c6cddb0cfb4fbfdb27f94e657536867b01bb09), I added supported for these dom components. So once you update with `main`, you can remove these functions and use the built-in ones :)

site/lib/src/markdown/markdown_parser.dart
```diff
@@ -56,14 +56,22 @@ class DashMarkdown extends AsyncStatelessComponent {
...
+      var nodes = DashMarkdownParser.buildNodes(markdownNodes);
+      for (final extension in allNodeProcessingExtensions) {
+        nodes = await extension.apply(currentPage, nodes);
+      }
 
-    return _nodeBuilder.build(nodes);
+      return _nodeBuilder.build(nodes);
+    } catch (e, st) {
+      print('Error rendering markdown: $e\n$st');
+      return text('Error rendering markdown');
```
parlough: I'd prefer if we don't catch errors here as it might cause issues to be lost. If some Markdown fails to be parsed, we likely should fix it right away rather than potentially not noticing.

tool/dash_site/pubspec.yaml
```diff
@@ -23,6 +23,7 @@ dependencies:
   markdown: ^7.3.0
   path: ^1.9.1
   yaml: ^3.1.3
+  yaml_edit: ^2.2.2
```
parlough: It seems this dependency isn't used:


```suggestion
```

site/lib/src/models/changelog_model.dart
```diff
@@ -0,0 +1,137 @@
...
+    var description = map['description'] as String;
+    var link = map['link'] as String?;
+
+    // Resolve liquid variables
+    for (final entry in _liquidVariables.entries) {
+      description = description.replaceAll(entry.key, entry.value);
+      if (link != null) {
+        link = link.replaceAll(entry.key, entry.value);
+      }
+    }
```
parlough: This manual replacement of Liquid variables doesn't feel too great and I worry it'll cause confusion around why some are working while others aren't.

Since it likely doesn't make sense to run the full templating logic over these strings either, perhaps it's best to just avoid using the template variables in the data and just use the expected values.

I don't expect the current variables used here to change much and if ever needed, they are pretty easy to update with find and replace.

site/lib/_sass/pages/_changelog-index.scss
```diff
@@ -0,0 +1,762 @@
...
+        font-size: 0.9rem;
+
+        &:focus {
+          outline: 2px solid var(--site-primary-color);
+          border-color: transparent;
+        }
+      }
+    }
+  }
+}
```
parlough:
```suggestion
}

```

site/lib/src/models/changelog_model.dart
```diff
@@ -0,0 +1,137 @@
...
+
+    return ChangelogEntry(
+      description: description,
+      version: map['version']
+          .toString()
+          .trim(), // Ensure version is string even if e.g. 3.0
```
parlough: A pain point of YAML 😢

It might be worth it to only accept strings and the full `major.minor.patch` syntax.

Since we already depend on `package:pub_semver` anyway, we could even use `Version.parse` to validate the format and get a type that's easier to work with.

site/lib/src/models/changelog_model.dart
```diff
@@ -0,0 +1,137 @@
...
+      releaseDate: map['releaseDate']?.toString(),
+      area: map['area'] as String,
+      subArea: map['subArea'] as String?,
+      tags:
+          (map['tags'] as List<Object?>?)
+              ?.cast<String>()
+              .map(ChangelogTag.fromId)
+              .toList() ??
+          [],
```
parlough:
```suggestion
          const [],
```

site/lib/src/models/changelog_model.dart
```diff
@@ -0,0 +1,137 @@
...
+    '{{site.pub}}': 'https://pub.dev',
+    '{{site.repo.dart.org}}': 'https://github.com/dart-lang',
+    '{{site.redirect.go}}': 'https://dart.dev/go',
+  };
+
+  factory ChangelogEntry.fromElement(web.Element element) {
+    final dataTags = element.getAttribute('data-tags') ?? '';
+    return ChangelogEntry(
+      description: element.getAttribute('data-description') ?? '',
+      version: (element.getAttribute('data-version') ?? '').trim(),
```
parlough: Doesn't seem this one needs to be trimmed.

```suggestion
      version: element.getAttribute('data-version') ?? '',
```

site/lib/src/models/changelog_model.dart
```diff
@@ -0,0 +1,137 @@
...
+
+  final String description;
+  final String version;
+  final String? releaseDate;
+  final String area;
+  final String? subArea;
+  final List<ChangelogTag> tags;
+  final String? link;
+
+  static String getShortVersion(String version) {
```
parlough: If you switch `version` to be represented with `Version` from `package:pub_semver`, this would be a bit simpler and could an extension on that type.

site/lib/src/models/changelog_model.dart
```diff
@@ -0,0 +1,137 @@
...
+
+  @override
+  int get hashCode =>
+      description.hashCode ^
+      version.hashCode ^
+      releaseDate.hashCode ^
+      area.hashCode ^
+      subArea.hashCode ^
+      link.hashCode ^
+      Object.hashAll(tags);
```
parlough: Potentially a bit more reliable and easier to maintain by just using a single `hashAll`:

```suggestion
  int get hashCode => Object.hashAll([
    description,
    version,
    releaseDate,
    area,
    subArea,
    link,
    ...tags,
  ]);
```

site/lib/src/models/changelog_model.dart
```diff
@@ -0,0 +1,137 @@
...
+      other is ChangelogEntry &&
+          runtimeType == other.runtimeType &&
+          description == other.description &&
+          version == other.version &&
+          releaseDate == other.releaseDate &&
+          area == other.area &&
+          subArea == other.subArea &&
+          link == other.link &&
+          tags.length == other.tags.length &&
+          tags.asMap().entries.every((e) => e.value == other.tags[e.key]);
```
parlough: Consider using [`MapEquality`](https://pub.dev/documentation/collection/latest/collection/MapEquality-class.html) from `package:collection` so we don't need to maintain/duplicate that logic.

site/lib/src/models/changelog_model.dart
```diff
@@ -0,0 +1,137 @@
...
+      Object.hashAll(tags);
+}
+
+enum ChangelogTag {
+  newTag('new', 'New'),
+  breaking('breaking', 'Breaking'),
+  fixed('fixed', 'Fixed'),
+  changed('changed', 'Changed'),
+  experimental('experimental', 'Experimental'),
+  versioned('versioned', 'Language versioned'),
```
parlough: I'd consider making this specifically `languageVersioned` so its meaning is more clear.

```suggestion
  languageVersioned('versioned', 'Language versioned'),
```

site/lib/src/components/pages/changelog/changelog_index.dart
```diff
@@ -0,0 +1,109 @@
...
+
+class _ChangelogEntryCard extends StatelessComponent {
+  const _ChangelogEntryCard(this.entry);
+
+  final ChangelogEntry entry;
+
+  @override
+  Component build(BuildContext context) {
+    // Generate a somewhat unique ID for DOM filtering
+    final uniqueId = slugify(
```
parlough: For something similar on flutter/website, @schultek used https://pub.dev/packages/nanoid2. Perhaps would make sense to use here as well?

site/lib/src/components/pages/changelog/changelog_index.dart
```diff
@@ -0,0 +1,109 @@
...
+      '${entry.version}-${entry.area}-${entry.subArea ?? ''}-${entry.description.substring(0, min(20, entry.description.length))}-${entry.hashCode}',
+    );
+
+    return div(
+      classes: 'changelog-card card',
+      id: uniqueId,
+      attributes: {
+        'data-version': entry.version,
+        if (entry.releaseDate != null) 'data-releasedate': entry.releaseDate!,
```
parlough: Can use null-aware elements here :D 

```suggestion
        'data-releasedate': ?entry.releaseDate,
```

site/lib/src/components/pages/changelog/changelog_index.dart
```diff
@@ -0,0 +1,109 @@
...
+        'data-area': entry.area,
+        if (entry.subArea != null) 'data-subarea': entry.subArea!,
```
parlough:
```suggestion
        'data-subarea': ?entry.subArea,
```

site/lib/src/components/pages/changelog/changelog_index.dart
```diff
@@ -0,0 +1,109 @@
...
+        'data-tags': entry.tags.map((t) => t.id).join(','),
+        // Store description for client-side search.
+        // Warning: This might be large.
+        'data-description': entry.description,
+        if (entry.link != null) 'data-link': entry.link!,
```
parlough:
```suggestion
        'data-link': ?entry.link,
```

site/lib/src/components/pages/changelog/changelog_index.dart
```diff
@@ -0,0 +1,109 @@
...
+        ]),
+        div(classes: 'card-content', [
+          if (entry.releaseDate != null)
+            p(classes: 'release-date', [
+              em([text('Released: ${entry.releaseDate} • v${entry.version}')]),
+            ]),
+          DashMarkdown(content: entry.description),
+          if (entry.link != null)
+            div(classes: 'read-more', [
+              a(href: entry.link!, target: Target.blank, [
```
parlough:
```suggestion
          if (entry.link case final entryLink?)
            div(classes: 'read-more', [
              a(href: entryLink, target: Target.blank, [
```

site/lib/src/components/pages/changelog/changelog_index.dart
```diff
@@ -0,0 +1,109 @@
...
+                text('Read more'),
+                span(classes: 'material-symbols-outlined', [
+                  text('open_in_new'),
+                ]),
```
parlough: We have a shared component for this.

```suggestion
                MaterialIcon('open_in_new'),
```

site/lib/src/components/pages/changelog/changelog_index.dart
```diff
@@ -0,0 +1,109 @@
...
+          ]),
+        ]),
+        div(classes: 'card-content', [
+          if (entry.releaseDate != null)
+            p(classes: 'release-date', [
+              em([text('Released: ${entry.releaseDate} • v${entry.version}')]),
+            ]),
+          DashMarkdown(content: entry.description),
+          if (entry.link != null)
+            div(classes: 'read-more', [
```
parlough, line 113: Consider updating this to use the already existing `Button` component. If not, consider adding the `text-button` class to make the styling and interaction handling more consistent.

site/lib/src/components/pages/changelog/changelog_index.dart
```diff
@@ -0,0 +1,109 @@
...
+      classes: 'changelog-card card',
+      id: uniqueId,
+      attributes: {
+        'data-version': entry.version,
+        if (entry.releaseDate != null) 'data-releasedate': entry.releaseDate!,
+        'data-area': entry.area,
+        if (entry.subArea != null) 'data-subarea': entry.subArea!,
+        'data-tags': entry.tags.map((t) => t.id).join(','),
+        // Store description for client-side search.
+        // Warning: This might be large.
```
parlough: I'm not sure we should do this. It bloats the page size quite a bit since everything is loaded.

Could the search instead check the already existing content? If we do keep it in the data, we need to escape any potential HTML in the description.

site/lib/src/components/pages/changelog/changelog_index.dart
```diff
@@ -0,0 +1,109 @@
...
+    final changesData = context.page.data['changelog'] as List<Object?>?;
+
+    final changelogEntries = <ChangelogEntry>[];
+    if (changesData != null) {
+      for (final change in changesData) {
+        changelogEntries.add(
+          ChangelogEntry.fromMap(change as Map<String, Object?>),
+        );
+      }
+    }
```
parlough, line 27: Currently this will silently fail if there are no changelog entries, rendering an empty index.

Since that likely indicates a mistake or other regression, consider adding a check that if `changesData` is `null` or is empty, throws an error.

parlough (commented):
This is exciting @antfitch! Thanks for diving into this.

I haven't had a chance to do a full review, but thought I would leave my in-progress comments. I'll finish up soon.

site/lib/main.dart
antfitch: Removing.

site/lib/src/models/changelog_model.dart
antfitch: The dart formatter is doing this automatically. Here's the message I get back about why this is here:

The standard Dart formatter (dart format) places this semicolon on its own line to clearly distinguish the list of enum cases from the class-like logic that follows. If you didn't have the constructor or fields, the semicolon wouldn't be necessary at all.

And yeah, that's all I've got. So leaving for now, but maybe something to ask the Dart team to reconsider if it seems too weird?

site/lib/_sass/pages/_changelog-index.scss
antfitch: Aye! I last tested on mobile and forgot to test again on desktop. Changed the desktop version to 900px. Hope that's enough.

site/lib/_sass/pages/_changelog-index.scss
antfitch: Added a media query to deal with screen width larger than 600px. Added 2fr 3fr to that.

site/lib/_sass/pages/_changelog-index.scss
antfitch: This should be fixed now. Can you take another look?

site/lib/src/markdown/markdown_parser.dart
antfitch: Check out my update. Does it work?

tool/dash_site/pubspec.yaml
antfitch: deleted

site/lib/src/models/changelog_model.dart
antfitch: Had Antigravity clean this up. How did it do?

site/lib/_sass/pages/_changelog-index.scss
antfitch: My IDE was deleting these by default on safe. *face palm*. Should be fixed now.

site/lib/src/components/pages/changelog/changelog_index.dart
antfitch: I ended up having Antigravity fix this. I'm actually worried that we may need to update our documentation to make sure we have the most recent information and we're using the more modern pattern that you suggested. Capturing this here so I can find this comment later.

site/lib/src/components/pages/changelog/changelog_index.dart
antfitch: Also flagging Antigravity here. We need to circle back and make sure it's suggesting the most up-to-date solutions.

site/lib/src/components/pages/changelog/changelog_index.dart
antfitch: Updated! Part of this update included adding a trailing icon to the Button class. Look closely.

site/lib/src/components/pages/changelog/changelog_index.dart
antfitch: Removed the 'data-description' attribute which should help with page size bloat and updated the search logic in changelog_model.dart. Take a close look.

site/lib/src/components/pages/changelog/changelog_index.dart
antfitch: Double-check my change. Is it what you need?

site/lib/_sass/pages/_changelog-index.scss
```diff
@@ -0,0 +1,781 @@
...
+      }
+
+      dd {
+        margin: 0;
+        grid-column: 2;
+        margin-bottom: 0.5rem;
+      }
+    }
+
+    .tag-label {
```
parlough, line 654: I haven't taken a close look at the style implementation, but I noticed these tag label styles (and the colors for each type) are mostly duplicated with the styles at L472-L520.

Perhaps these can be abstracted out and shared, with just the differences added to this situation?

site/lib/src/components/pages/changelog/changelog_filters_sidebar.dart
```diff
@@ -0,0 +1,331 @@
+import 'dart:math';
```
parlough:
```suggestion
// Copyright (c) 2025, the Dart project authors. Please see the AUTHORS file
// for details. All rights reserved. Use of this source code is governed by a
// BSD-style license that can be found in the LICENSE file.

import 'dart:math';
```

site/lib/src/components/common/client/filtering.dart
```diff
@@ -0,0 +1,149 @@
...
+    return div([
+      h4([.text(title)]),
+      ul([
+        for (final item in items)
+          li([
+            input(
+              type: InputType.checkbox,
+              attributes: {
+                'role': 'checkbox',
+                'name': 'filter-${idProvider(item)}',
```
parlough: Consider pulling this call to `idProvider(item)` out to a new variable before the `return`. So we can avoid calling it multiple times in the same `build`.

site/lib/src/components/common/client/filtering.dart
```diff
@@ -0,0 +1,149 @@
...
+  Component build(BuildContext context) {
+    return div([
+      h4([.text(title)]),
+      ul([
+        for (final item in items)
+          li([
+            input(
+              type: InputType.checkbox,
+              attributes: {
+                'role': 'checkbox',
```
parlough: No need to set the role to checkbox if its input type is already checkbox. Generally you can prefer the native handling if it's correct :)


```suggestion
```

site/lib/src/components/common/client/filtering.dart
```diff
@@ -0,0 +1,149 @@
+// Copyright 2025 The Dart Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:jaspr/dom.dart';
+import 'package:jaspr/jaspr.dart';
+import 'package:universal_web/web.dart' as web;
+
+class CheckboxFilterGroup<T> extends StatelessComponent {
```
parlough, line 13: Consider adding some basic doc comments to the members here, at least the class. Specifically it might be worth mentioning that `T` should implement `hashCode` and `==`.

site/lib/src/components/common/client/filtering.dart
```diff
@@ -0,0 +1,149 @@
...
+    return details(
+      [
+        summary([.text(component.title)]),
+        ul([
+          for (final item in component.items)
+            li([
+              input(
+                type: InputType.checkbox,
+                attributes: {
+                  'role': 'checkbox',
```
parlough:
```suggestion
```

site/lib/src/components/common/client/filtering.dart
```diff
@@ -0,0 +1,149 @@
...
+              ),
+              label(
+                attributes: {'for': 'filter-${component.idProvider(item)}'},
+                [.text(component.labelProvider(item))],
+              ),
+            ]),
+        ]),
+      ],
+      classes: 'version-group',
+      attributes: _isOpen ? {'open': ''} : null,
```
parlough: It seems the `details` Jaspr component already has handling for `open`, so no need to handle the attributes manually:


```suggestion
      open: _isOpen,
```

site/lib/src/components/common/client/filtering.dart
```diff
@@ -0,0 +1,149 @@
...
+  final List<Component> actions;
+  final List<Component> bottomActions;
+  final String? id;
+
+  @override
+  Component build(BuildContext context) {
+    return div(id: id, classes: 'chip-filters-group', [
+      div(classes: 'top-row', [
+        searchBar,
+        if (actions.isNotEmpty) ...actions,
```
parlough: No need for checks like this before spreads. If `actions` is empty, there's nothing that will be spread into the created list.

```suggestion
        ...actions,
```

site/lib/src/components/common/client/filtering.dart
```diff
@@ -0,0 +1,149 @@
...
+      ]),
+      div(classes: 'label-row', [
+        resultCount,
+        if (bottomActions.isNotEmpty) ...bottomActions,
```
parlough: No need for checks like this before spreads. If `bottomActions` is empty, there's nothing that will be spread into the created list.

```suggestion
        ...bottomActions,
```

site/lib/src/components/common/client/filtering.dart
```diff
@@ -0,0 +1,149 @@
...
+    this.actions = const [],
+    this.bottomActions = const [],
+    this.id,
+    super.key,
+  });
+
+  final Component searchBar;
+  final Component resultCount;
+  final List<Component> actions;
+  final List<Component> bottomActions;
```
parlough: Perhaps `trailingActions` is a more accurate (and flexible) name here?

site/lib/src/components/common/button.dart
```diff
@@ -24,12 +25,13 @@ class Button extends StatelessComponent {
     this.title,
     this.asRaw = false,
     this.onClick,
-  }) : assert(content != null || icon != null);
+  }) : assert(content != null || icon != null || trailingIcon != null);
```
parlough: I think even if `trailingIcon` is specified, either `content` or `icon` should be as well. If an icon button without other content is desired, then `icon` should be used instead. 

```suggestion
  }) : assert(content != null || icon != null);
```

Or a more specific handling a message could be added for that case.

site/lib/src/components/pages/changelog/breaking_changes_legend.dart
```diff
@@ -0,0 +1,104 @@
...
+  const BreakingChangesLegend({required this.onClose, super.key});
+
+  final VoidCallback onClose;
+
+  @override
+  Component build(BuildContext context) {
+    return div(classes: 'legend-overlay', [
+      div(classes: 'legend-dialog', [
+        div(classes: 'legend-header', [
+          const h3([.text('About Breaking Changes')]),
```
parlough: I guess this page expanded in purpose since you started, but should this index instead become "Types of changes", have an updated description, and add explanations for the other types ("Fix", "New", "Changed")?

site/lib/src/components/pages/changelog/breaking_changes_legend.dart
```diff
@@ -0,0 +1,104 @@
...
+          button(
+            classes: 'close-button',
+            onClick: onClose,
+            [
+              const span(classes: 'material-symbols-outlined', [
+                .text('close'),
+              ]),
+            ],
+          ),
```
parlough: Can reuse our existing `Button` component here, which includes handling for an icon button.

```suggestion
          Button(
            icon: 'close',
            classes: ['close-button'],
            onClick: onClose,
          ),
```

site/lib/src/components/pages/changelog/breaking_changes_legend.dart
```diff
@@ -0,0 +1,104 @@
...
+        ]),
+        const div(classes: 'legend-content', [
+          p([
+            .text(
+              'This page lists all breaking changes and deprecations in '
```
parlough: Should this introduction be at the top of the page before the index? Then also be updated to cover the expanded range of the page.

site/lib/src/components/pages/changelog/breaking_changes_legend.dart
```diff
@@ -0,0 +1,104 @@
...
+
+class BreakingChangesLegend extends StatelessComponent {
+  const BreakingChangesLegend({required this.onClose, super.key});
+
+  final VoidCallback onClose;
+
+  @override
+  Component build(BuildContext context) {
+    return div(classes: 'legend-overlay', [
+      div(classes: 'legend-dialog', [
```
parlough: Optionally consider extracting this dialog handling out to a shared component.

site/lib/src/components/pages/changelog/breaking_changes_legend.dart
```diff
@@ -0,0 +1,104 @@
...
+import 'package:jaspr/jaspr.dart';
+
+class BreakingChangesLegend extends StatelessComponent {
+  const BreakingChangesLegend({required this.onClose, super.key});
+
+  final VoidCallback onClose;
+
+  @override
+  Component build(BuildContext context) {
+    return div(classes: 'legend-overlay', [
```
parlough: To match common modal behavior, perhaps this overlay should have an on-click listener to check if outside the dialog is clicked, close it.

site/lib/src/components/pages/changelog/breaking_changes_legend.dart
```diff
@@ -0,0 +1,104 @@
...
+                "The Dart SDK doesn't maintain backward compatibility, "
+                'and code might break as soon as you upgrade your SDK version '
+                'if it relies on the previous behavior. '
+                'These are the majority of changes and '
+                "aren't specially marked.",
+              ),
+            ]),
+            dt([
+              .text('Language versioned '),
+              span(classes: 'tag-label versioned-tag', [
```
parlough: Perhaps these tag labels should be a new shared component to avoid needing to repeat the setup here and in the index cards.

site/lib/src/components/pages/changelog/changelog_filters.dart
```diff
@@ -0,0 +1,157 @@
...
+      if (listContainer == null) {
+        return;
+      }
+
+      final entryCards = listContainer.querySelectorAll('.changelog-card');
+      recreateEntries(entryCards);
+    }
+  }
+
+  void recreateEntries(web.NodeList entryCards) {
```
parlough:
```suggestion
      _recreateEntries(entryCards);
    }
  }

  void _recreateEntries(web.NodeList entryCards) {
```

site/lib/src/components/pages/changelog/changelog_filters.dart
```diff
@@ -0,0 +1,157 @@
...
+      final entry = ChangelogEntry.fromElement(element);
+      entries.add(entry);
+    }
+    filteredEntriesCount = entries.length;
+  }
+
+  void setFilters([void Function()? callback]) {
+    setState(callback ?? () {});
+
+    final entriesToShow = filters.filterEntries(entries, searchQuery);
```
parlough: It might make sense to convert this to a set as it's only used for `contains` checks.

```suggestion
    final entriesToShow = filters.filterEntries(entries, searchQuery).toSet();
```

site/lib/src/components/pages/changelog/changelog_filters.dart
```diff
@@ -0,0 +1,157 @@
...
+    filteredEntriesCount = entriesToShow.length;
+
+    final listContainer = web.document.getElementById('all-changelog-list');
+    if (listContainer != null) {
```
parlough: Consider flipping this condition to reduce nesting.


```
    if (listContainer == null) {
      return;
    }

    // ...
```

site/lib/src/components/pages/changelog/changelog_filters.dart
```diff
@@ -0,0 +1,157 @@
...
+            filters.reset();
+            setFilters();
+          },
+          [
+            const span(classes: 'material-symbols-outlined', [
+              .text('close_small'),
+            ]),
+            const span([.text('Clear filters')]),
+          ],
+        ),
```
parlough: Can reuse our existing `Button` component here to simplify the implementation and avoid reimplementing anything:


```suggestion
        Button(
          icon: 'close_small',
          content: 'Clear filters',
          disabled:
              searchQuery.isEmpty &&
              filters.selectedTags.isEmpty &&
              filters.selectedAreas.isEmpty &&
              filters.selectedVersions.isEmpty,
          onClick: () {
            // Update search query and reset filters.
            // We call setFilters to ensure the UI updates even if
            // filters.reset() doesn't trigger a change,
            // such as if only search query was active.
            searchQuery = '';
            filters.reset();
            setFilters();
          },
        ),
```

site/lib/src/components/pages/changelog/changelog_filters.dart
```diff
@@ -0,0 +1,157 @@
...
+            if (toggle != null) {
+              toggle.checked = !toggle.checked;
+            }
+          },
+          [
+            const span(classes: 'material-symbols-outlined', [
+              .text('filter_list'),
+            ]),
+          ],
+        ),
```
parlough: Can reuse our existing `Button` component here to simplify the implementation and avoid reimplementing anything:


```suggestion
        Button(
          icon: 'filter_list',
          classes: ['show-filters-button'],
          onClick: () {
            final toggle =
                web.document.getElementById('open-filter-toggle')
                    as web.HTMLInputElement?;
            if (toggle != null) {
              toggle.checked = !toggle.checked;
            }
          },
        ),
```

site/lib/src/components/pages/changelog/changelog_filters_sidebar.dart
```diff
@@ -0,0 +1,331 @@
...
+        attributes: {'hidden': 'true'},
+      ),
+      div(id: 'changelog-filter-group-wrapper', [
+        div(id: 'changelog-filter-group', [
+          const div(classes: 'filter-header', [
+            label(
+              attributes: {'for': 'open-filter-toggle', 'aria-hidden': 'true'},
+              classes: 'close-icon',
+              [
+                span(classes: 'material-symbols-outlined', [.text('close')]),
```
parlough:
```suggestion
                MaterialIcon('close', title: 'Close filters'),
```

site/lib/src/components/pages/changelog/changelog_filters_sidebar.dart
```diff
@@ -0,0 +1,331 @@
...
+              onClick: () {
+                setState(() => _showLegend = true);
+              },
+              attributes: {'aria-label': 'About changelog'},
+              [
+                const span(classes: 'material-symbols-outlined', [
+                  .text('info'),
+                ]),
+              ],
+            ),
```
parlough:
```suggestion
            Button(
              icon: 'info',
              classes: ['info-icon'],
              title: 'About changelog',
              onClick: () {
                setState(() => _showLegend = true);
              },
            ),
```

site/lib/src/components/pages/changelog/changelog_filters_sidebar.dart
```diff
@@ -0,0 +1,331 @@
...
+              // Ensure we have populated available filters if running on web.
+              if (kIsWeb &&
+                  (ChangelogFiltersSidebar.filters.availableAreas.isEmpty ||
+                      ChangelogFiltersSidebar
+                          .filters
+                          .availableVersions
+                          .isEmpty)) {
+                // Do not populate here to avoid side effects during build.
+                // It should be done in initState.
+              }
```
parlough: It seems this if statement doesn't do anything. Perhaps a remnant from development or debugging?


```suggestion
```

site/lib/src/components/pages/changelog/changelog_filters_sidebar.dart
```diff
@@ -0,0 +1,331 @@
...
+              for (final group in groups.values) {
+                group.sort(
+                  (left, right) => ChangelogFiltersSidebar.filters
+                      .compareVersions(right, left),
+                );
+              }
+              final sortedMajors = groups.keys.toList()
+                ..sort(
+                  (left, right) => int.parse(right).compareTo(int.parse(left)),
+                );
```
parlough:
```suggestion
              final sortedMajors = groups.keys.sorted(
                (left, right) => int.parse(right).compareTo(int.parse(left)),
              );
```

site/lib/src/components/pages/changelog/changelog_filters_sidebar.dart
```diff
@@ -0,0 +1,331 @@
...
+                  labelProvider: (tag) => tag.label,
+                  idProvider: (tag) => tag.id,
+                ),
+                CheckboxFilterGroup<String>(
+                  title: 'Area',
+                  items:
+                      ChangelogFiltersSidebar.filters.availableAreas
+                          .where((area) => area != 'Docs')
+                          .toList()
+                        ..sort(ChangelogFiltersSidebar.filters.compareAreas),
```
parlough:
```suggestion
                  items: ChangelogFiltersSidebar.filters.availableAreas
                      .where((area) => area != 'Docs')
                      .sorted(ChangelogFiltersSidebar.filters.compareAreas),
```

site/lib/src/components/pages/changelog/changelog_filters_sidebar.dart
```diff
@@ -0,0 +1,331 @@
...
+                  selectedItems: ChangelogFiltersSidebar.filters.selectedAreas,
+                  onToggle: (area, checked) =>
+                      ChangelogFiltersSidebar.filters.setArea(area, checked),
+                  labelProvider: (area) => area,
+                  idProvider: (area) => 'area-${slugify(area)}',
+                ),
+                div([
+                  const h4([.text('Version')]),
+                  for (final major in sortedMajors)
+                    CollapsibleFilterGroup<String>(
```
parlough: These end up with quite a lot of entries. Since I imagine filtering by a specific minor version will be one of the most common ways to filter, we might want to explore ways to not hide them by default. Definitely can be done as part of follow-up improvements though.

site/lib/src/components/pages/changelog/changelog_filters_sidebar.dart
```diff
@@ -0,0 +1,331 @@
...
+  Set<String> selectedVersions = {};
+
+  Set<String> availableAreas = {};
+  Set<String> availableVersions = {};
+
+  void populateAvailableFiltersFromDOM() {
+    final listContainer = web.document.getElementById('all-changelog-list');
+    if (listContainer == null) return;
+
+    var changed = availableAreas.add('Docs');
```
parlough: It's not super clear why changed is initially set to the result of "Docs" being in the set of available areas. Consider adding a doc comment to this method or some inline comments to explain the logic here?

site/lib/src/components/pages/changelog/changelog_filters_sidebar.dart
```diff
@@ -0,0 +1,331 @@
...
+    if (isSelected) {
+      selectedAreas.add(area);
+    } else {
+      selectedAreas.remove(area);
+    }
+    notifyListeners();
+  }
+
+  void setVersion(String version, bool isSelected) {
+    if (!version.contains('.')) {
```
parlough: All of this logic here gives me another reason to think we should consider using the `Version` type from `package:pub_semver` rather than trying to reimplement version comparisons and handling.

site/lib/src/components/pages/changelog/changelog_filters_sidebar.dart
```diff
@@ -0,0 +1,331 @@
...
+
+  List<ChangelogEntry> filterEntries(
+    List<ChangelogEntry> entries,
+    String searchQuery,
+  ) {
+    if (searchQuery.isEmpty &&
+        selectedTags.isEmpty &&
+        selectedAreas.isEmpty &&
+        selectedVersions.isEmpty) {
+      return List.from(entries);
```
parlough: Does this need to return a new list? Doesn't look like callers should modify it.


```suggestion
      return entries.
```

If it does, then use a spread, `List.of`, or `.toList()` as `List.from` has some typing issues.


```suggestion
      return [...entries];
```

site/lib/src/components/pages/changelog/changelog_filters_sidebar.dart
```diff
@@ -0,0 +1,331 @@
...
+    }
+
+    final entriesToShow = <ChangelogEntry>[];
+    searchQuery = searchQuery.trim().toLowerCase();
```
parlough: Consider using a new variable here and updating the usages to avoid reassigning parameters which can be confusing (and is often linted against):

```
final normalizedQuery = searchQuery.trim().toLowerCase();
```

site/lib/src/components/pages/changelog/changelog_index.dart
```diff
@@ -0,0 +1,116 @@
...
+
+    for (final change in changesData) {
+      changelogEntries.add(
+        ChangelogEntry.fromMap(change as Map<String, Object?>),
+      );
+    }
+
+    if (changelogEntries.isEmpty) {
+      throw Exception('No changelog entries found.');
+    }
```
parlough, line 39: Can simplify this with a collection-for element :D 

```suggestion
    final changesData = context.page.data['changelog'] as List<Object?>?;
    if (changesData == null || changesData.isEmpty) {
      throw Exception('Changelog data is missing or invalid.');
    }

    final changelogEntries = <ChangelogEntry>[
      for (final change in changesData)
        ChangelogEntry.fromMap(change as Map<String, Object?>),
    ];
```

site/lib/src/components/pages/changelog/changelog_index.dart
parlough: I still recommend considering something like this.

site/lib/src/models/changelog_model.dart
```diff
@@ -0,0 +1,126 @@
...
+      link: element.getAttribute('data-link'),
+    );
+  }
+
+  final String description;
+  final String version;
+  final String? releaseDate;
+  final String area;
+  final String? subArea;
+  final List<ChangelogTag> tags;
```
parlough: Should this be a `Set`? Or do you want the file to be conscious about the order it specifies tags?

site/lib/src/components/pages/changelog/changelog_index.dart
```diff
@@ -0,0 +1,116 @@
...
+      },
+      [
+        div(classes: 'card-header', [
+          div(classes: 'header-left', [
+            span(classes: 'version-badge', [.text(entry.shortVersion)]),
+
+            span(classes: 'area-badge', [
+              .text(
+                entry.area +
+                    (entry.subArea != null ? ': ${entry.subArea}' : ''),
```
parlough: These entries look like the title of each change rather than labels of the area of the change. This is a bit confusing as there could be other changes with the same area-subarea.

Perhaps we can explore alternative designs for the this label or group changes with the same version and area in the same card?

site/lib/src/components/pages/changelog/changelog_index.dart
```diff
@@ -0,0 +1,116 @@
...
+              ),
+            ]),
+          ]),
+          div(classes: 'tags', [
+            for (final tag in entry.tags)
+              span(classes: 'tag-label ${tag.id}-tag', [.text(tag.label)]),
```
parlough: Perhaps we can add some `title` to these with a short explanation of what the tag means.

parlough (commented):
Thanks for making those adjustments @antfitch! Still super cool and generally looks great. Note I pushed some changes to update with `main` and migrate to the new version of Jaspr. I also have a few more suggestions to consider and some questions:

1. Each change card contains a version tag at the beginning of the title. It's attracts a lot of visual attention more than the rest of the information and feels a bit duplicated, particularly considering the most common situation is likely looking at all the changes in a single release. Could/should this badge be pulled out to be before the cards in that version? Then it acts as a larger grouping.

2. This page has been expanded to incorporate all changes which is really nice, but could you provide more details on how we plan to keep this up to date? Do we want the source file to actually live in the SDK and require developers update that instead of/in addition the `CHANGELOG.md` file?

3. Doesn't necessarily have to be done in this initial PR, but once we switch to this index, we lose the ability to directly link to a focused collection of breaking changes or language-versioned changes. To support those use cases and others, it'd be nice to support specifying types from the page query parameters.

site/lib/src/components/pages/changelog/changelog_index.dart
antfitch: Missed this request. Will get it in.

site/lib/src/components/pages/changelog/changelog_filters_sidebar.dart
```diff
@@ -0,0 +1,318 @@
...
+    selectedAreas.clear();
+    selectedVersions.clear();
+    availableAreas.clear();
+    availableVersions.clear();
+    if (notify) {
+      notifyListeners();
+    }
+  }
+
+  int compareVersions(String v1, String v2) {
```
schultek: This could use `Version` class from `package:pub_semver`

site/lib/src/components/pages/changelog/changelog_index.dart
```diff
@@ -0,0 +1,109 @@
...
+import 'changelog_filters.dart';
+import 'changelog_filters_sidebar.dart';
+
+final class ChangelogIndex extends StatelessComponent {
+  const ChangelogIndex({super.key});
+
+  @override
+  Component build(BuildContext context) {
+    // Reset static filters state to prevent SSR state leaks in dev mode.
+    ChangelogFiltersSidebar.filters.disposeState(notify: false);
```
schultek: I don't think this is necessary. During SSR the filter state shouldn't be used anyways.

site/lib/_sass/pages/_changelog-index.scss
antfitch: I've refactored this file to clear out redundancies I could find. There is one note: we have a _tags.scss file and it looks like we could move the tag logic for the changelog over, but it looks like it might be a hefty refactor, so leaving that out for now unless you want me just go ahead and do that.

site/lib/src/components/common/client/filtering.dart
antfitch: I'm a huge fan of having comments. Done! Check closely.

site/lib/src/components/common/client/filtering.dart
antfitch: Created a variable called `itemId`. Also, refactored this file to make it more modular and remove redundancies.

site/lib/src/components/common/client/filtering.dart
antfitch: Done!

site/lib/src/components/pages/changelog/breaking_changes_legend.dart
antfitch: Yeah, this makes a lot of sense. Updated this so that the dialog is closed when the overlay is clicked on. Also, I refactored the code to make it more modular.

site/lib/src/components/pages/changelog/breaking_changes_legend.dart
antfitch: Done. Added a new component called dialog.dart. I'm a bit worried that this is too broad a term, so let me know if I should change it to something else.

site/lib/src/components/pages/changelog/breaking_changes_legend.dart
antfitch: I removed the intro text in this dialog, updated the title, and added new sections. Can you look closely at the wording? I'm not sure that some of these are strictly for the SDK.

site/lib/src/components/pages/changelog/breaking_changes_legend.dart
antfitch: Deleted this to reduce words on the page.

site/lib/src/components/pages/changelog/breaking_changes_legend.dart
antfitch: Definitely. Take a look at the update. One note is that I kept the class `_TagLabel` inside of this current file because we already have a tags component that is a bit different. Probably that component could be refactored to include this new information, but keeping things separate for now so you can decide whether we do that now or something else.

site/lib/src/components/pages/changelog/changelog_filters.dart
antfitch: Done!

site/lib/src/components/pages/changelog/changelog_filters_sidebar.dart
antfitch: Sounds good. Added a doc comment here. Also, I separated the UI code from the state code (we now have changelog_filters_sidebar.dart and changelog_filters_notifier.dart).

site/lib/src/components/pages/changelog/changelog_filters_sidebar.dart
antfitch: I've updated this code to use the package:pub_semver package. Also updated changelog_model.dart.

site/lib/src/components/pages/changelog/changelog_filters_sidebar.dart
antfitch: Definitely. Updated.

site/lib/src/components/pages/changelog/changelog_filters_sidebar.dart
antfitch: Nope, we don't need the new list. Updated `List.from(entries)` to `entries`.

site/lib/src/components/pages/changelog/changelog_filters_sidebar.dart
antfitch: I think that's a VERY good idea. Thank you for catching this. Updated!

site/lib/src/components/pages/changelog/changelog_index.dart
antfitch: Thanks for catching this. I've removed this line.

site/lib/src/components/pages/changelog/changelog_index.dart
antfitch: That reduced a lot of lines, but it caused the server to crash. Not sure why, but investigating... And fixed!

site/lib/src/components/pages/changelog/changelog_filters_sidebar.dart
antfitch: Sounds good. I'll leave this as something one of us can follow up with in a future PR.

site/lib/src/components/pages/changelog/changelog_index.dart
antfitch: Great idea. It also makes it less necessary to open the legend. Do you think we still need the legend?

site/lib/src/models/changelog_model.dart
antfitch: Set makes more sense, and nope, I don't think order should matter here. Updated.

site/lib/src/models/changelog_model.dart
antfitch: Added this change. It reduced quite a bit of code.

site/lib/src/models/changelog_model.dart
antfitch: Updated. Take a close look at my changes.

antfitch (commented):
Went through the files and refactored. Take a close look at how I re-organized content under version numbers.

site/lib/src/components/common/dialog.dart
```diff
@@ -0,0 +1,60 @@
...
+  /// The title displayed in the dialog header.
+  final String title;
+
+  /// The content of the dialog.
+  final List<Component> children;
+
+  @override
+  Component build(BuildContext context) {
+    return div(
+      // TODO: Rename these classes to be generic (e.g. dialog-overlay) in global styles
```
schultek: We should resolve this TODO before merging.

site/lib/src/components/pages/changelog/changelog_filters_sidebar.dart
```diff
@@ -0,0 +1,150 @@
...
+                ]),
+              ]);
+            },
+          ),
+        ]),
+      ]),
+      if (_showLegend)
+        BreakingChangesLegend(
+          onClose: () => setState(() => _showLegend = false),
+        ),
```
schultek: Instead of conditional rendering, we should toggle a className on the dialog (e.g. `show`). Then the content is already prerendered in the html and better accessible for screenreaders etc.

site/lib/src/components/pages/changelog/changelog_filters_sidebar.dart
```diff
@@ -0,0 +1,150 @@
...
+                      ChangelogFiltersSidebar.filters.setArea(area, checked),
+                  labelProvider: (area) => area,
+                  idProvider: (area) => 'area-${slugify(area)}',
+                ),
+                div([
+                  const h4([.text('Version')]),
+                  for (final major in sortedMajors)
+                    CollapsibleFilterGroup<String>(
+                      key: Key(major),
+                      title: '$major.x',
```
schultek, line 129: We could improve the ux here by adding a checkbox to the title for signaling the 'any major.x version' case, and then have this sync with the sub-version selection: 
- checked when all sub-versions are selected
- indeterminate when only some sub-versions are selected
- unchecked when no sub-versions are selected

Mockup: 

Image

site/lib/src/components/pages/changelog/changelog_filters_notifier.dart
```diff
@@ -0,0 +1,247 @@
...
+/// Manages the state of the changelog filters.
+class ChangelogFiltersNotifier extends ChangeNotifier {
+  static final _versionXyRegex = RegExp(r'^\d+\.\d+$');
+
+  Set<ChangelogTag> selectedTags = {};
+  Set<String> selectedAreas = {};
+  Set<String> selectedVersions = {};
+
+  Set<String> availableAreas = {};
+  Set<String> availableVersions = {};
```
schultek: We should probably store the `Version` objects itself in here and `selectedVersions`, which would save us some parsing back and forth.

site/lib/src/components/pages/changelog/changelog_filters_sidebar.dart
antfitch: I specifically want to avoid this. My reasoning is that the user has to be super careful where they click. When I created this part a long time ago, I noticed that it was easy to accidentally click or unclick the version when I actually wanted to expand or close a section.

site/lib/src/components/pages/changelog/changelog_index.dart
antfitch: Sounds good. Added a slight design change.

site/lib/src/components/common/dialog.dart
antfitch: Thank you for catching. Fixed!

site/lib/src/components/pages/changelog/changelog_filters_sidebar.dart
antfitch: Updated!

site/lib/src/components/pages/changelog/changelog_filters_notifier.dart
antfitch: Updated, but check carefully.

site/lib/src/components/pages/changelog/changelog_filters_notifier.dart
```diff
@@ -0,0 +1,191 @@
...
+      selectedAreas.remove(area);
+    }
+    notifyListeners();
+  }
+
+  void setVersion(Version version, bool isSelected) {
+    if (isSelected) {
+      selectedVersions.add(version);
+      // Check if all versions of this major are selected
+      // Logic for major "check" is now computed in UI, so we just add the individual version
```
schultek: This comment seems outdated.

site/lib/src/components/pages/changelog/changelog_filters_sidebar.dart
```diff
@@ -0,0 +1,191 @@
...
+                                    .contains,
+                              );
+                        } else if (item is Version) {
+                          return ChangelogFiltersSidebar
+                              .filters
+                              .selectedVersions
+                              .contains(item);
+                        }
+                        return false;
+                      },
```
schultek: We should use the `selectedItems` set here and remove the `isCheckedProvider` method, to be declarative where possible.

could be:


```suggestion
                      selectedItems: {
                        if (ChangelogFiltersSidebar.filters.isMajorSelected(major)) major,
                        ...ChangelogFiltersSidebar.filters.selectedVersions,
                      },
```

With `isMajorSelected` doing the major version check as in the current callback.

site/lib/src/components/pages/changelog/breaking_changes_legend.dart
```diff
@@ -0,0 +1,107 @@
...
+  /// Callback triggered when the close button or overlay is clicked.
+  final VoidCallback onClose;
+
+  /// Whether the legend is visible.
+  final bool visible;
+
+  @override
+  Component build(BuildContext context) {
+    return Dialog(
+      classes: visible ? 'show' : null,
```
schultek: The `Dialog` should also have the `visible` bool itself instead of the `classes` String, so that the class name logic is contained inside the dialog.

site/lib/src/components/pages/changelog/changelog_filters_notifier.dart
antfitch: Tentatively closing.

site/lib/src/components/pages/changelog/changelog_filters_sidebar.dart
antfitch: Tentatively closing.

site/lib/src/components/pages/changelog/changelog_filters_sidebar.dart
antfitch: Tentatively closing.

site/lib/src/components/pages/changelog/changelog_index.dart
antfitch: Tentatively closing.

site/lib/src/components/pages/changelog/changelog_filters_sidebar.dart
antfitch: Tentatively closing.

site/lib/src/components/pages/changelog/changelog_filters_notifier.dart
antfitch: Totally. Deleted it. This section seems pretty self explanatory now.

site/lib/src/components/pages/changelog/breaking_changes_legend.dart
antfitch: Absolutely. Done.

site/lib/src/components/pages/changelog/changelog_filters_sidebar.dart
antfitch: Updated and tested.

site/lib/src/components/pages/changelog/breaking_changes_legend.dart
antfitch: Tentatively closing. Feel free to re-open if this is still a problem.

site/lib/src/components/pages/changelog/changelog_filters_sidebar.dart
antfitch: Tentatively closing. Feel free to re-open if this is still a problem.

site/lib/src/components/pages/changelog/changelog_filters_notifier.dart
antfitch: Tentatively closing. Feel free to re-open if this is still a problem.

antfitch:
@parlough, I've cleaned up the description for this PR. It should be much easier to read now. If a step is missing, feel free to update the description and add it directly.

src/data/sidenav/default.yml
```diff
@@ -350,12 +350,8 @@
   children:
     - title: Dart Blog
       permalink: https://blog.dart.dev
-    - title: Dart SDK Changelog
-      permalink: https://github.com/dart-lang/sdk/blob/main/CHANGELOG.md
-    - title: Breaking changes
-      permalink: /resources/breaking-changes
-    - title: Language evolution
-      permalink: /resources/language/evolution
```
parlough, line 358: This language evolution page still covers important content, mostly around language versioning (https://dart.dev/resources/language/evolution#language-versioning and below).

Until that's covered elsewhere, I'd prefer it to still have an entry in the sidenav.

Perhaps we can extract that content to its own page before this PR lands?

parlough, line 358: Can we keep this in the sidenav for now or first move out the language versioning contents (https://dart.dev/resources/language/evolution#language-versioning and below) to their own page?

It's important content that I want to be sure is discoverable and doesn't get lost.

site/lib/src/components/pages/changelog/changelog_index.dart
```diff
@@ -0,0 +1,123 @@
...
+                classes: 'tag-label ${tag.id}-tag',
+                attributes: {'title': tag.description},
+                [.text(tag.label)],
+              ),
+          ]),
+        ]),
+        div(classes: 'card-content', [
+          if (entry.releaseDate != null)
+            p(classes: 'release-date', [
+              em([.text('Released: ${entry.releaseDate}')]),
```
parlough, line 109: Perhaps this information should instead (or also) be alongside the version header?

site/lib/src/components/pages/changelog/changelog_index.dart
```diff
@@ -0,0 +1,123 @@
...
+        'data-link': ?entry.link,
+      },
+      [
+        div(classes: 'card-header', [
+          div(classes: 'header-left', [
+            span(classes: 'area-badge', [
+              .text(entry.area),
+            ]),
+            if (entry.subArea case final subArea?)
+              span(classes: 'entry-title', [.text(subArea)]),
```
parlough: I'm still not sure about these titles. Rather than a subarea, they look like a title describing the change. For now we can land it as is, but something to consider.

parlough: Looks like these titles often contain Markdown formatting.


```suggestion
              span(classes: 'entry-title', [
                DashMarkdown(content: subArea, inline: true),
              ]),
```

src/data/changelog.yml
parlough, line 1: I'm curious, do you think it makes sense to first group by version, reducing duplication and providing some structure? That also makes it so we don't need to update a bunch of releaseDates when a version is released.

Or do you see entries as potentially not being tied to an existing or future release?

src/data/changelog.yml
```diff
@@ -0,0 +1,4652 @@
+# --------------------------------------------------
+- version: 3.11.0
+  releaseDate: TBD
+  area: SDK
+  subArea: Platform Updates
```
parlough, line 5: Not necessary for this PR, but I'd consider updating these sub areas to be sentence case for consistency.

src/content/changelog.md
```diff
@@ -0,0 +1,8 @@
+---
+title: Dart changelog
+description: A complete changelog for Dart.
+bodyClass: changelog-page
+showBreadcrumbs: false
+---
+
```
parlough, line 7: Do you think it'd be valuable to add a short introduction of the page here? Or perhaps a more detailed description above?

site/lib/src/models/changelog_model.dart
```diff
@@ -0,0 +1,140 @@
+// Copyright (c) 2026, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:collection/collection.dart';
+import 'package:pub_semver/pub_semver.dart';
```
parlough, line 6: Switching to this made everything much easier to review and maintain, thanks for that update!

site/lib/src/components/common/dialog.dart
parlough, line 1: If convenient, can you make it so this dialog closes when you press `Esc`? Will potentially be more intuitive and convenient for keyboard users.

src/data/changelog.yml
```diff
@@ -0,0 +1,4652 @@
...
+  description: |
+    Dart 3.11 adds support for Unix domain sockets on Windows in `dart:io`. It
```
parlough, line 7: Not important for these pre-existing entries, but for future entries we definitely should have guidelines on these descriptions. For example, in this one, the repetition of Dart 3.11 seems a bit awkward.

parlough (approved):
In general, this is super cool and very exciting.

The primary concern I have before landing this version is losing the language versioning page from the sidenav (I recommend extracting that content out). I have a few other small comments you can consider now as well.

There's some other logic and style changes I'd like to explore as well, but that can happen as follow up that I can propose when I have more time.

Thanks for working so hard on this @antfitch!!!

src/data/sidenav/default.yml
antfitch: I completely missed that. I've create a new topic called `Language versioning` and moved it to the top, under `Comments` in the TOC (it seems like a high-level thing).

I've also added a deprecated warning to the top of each deprecated topic with links to the new changelog and language versioning page.

antfitch: Moved out the language versioning to a new page, and it's looking pretty good, I think! Check out the change under Language > Language versioning

src/data/changelog.yml
antfitch: I thought about it, but sometimes within a version, some dates are not all the same. For example, we might have some patches that fall under a specific version number but were released at a later date. 3.9 was a good example, and I suspect we may need to update 3.10 with a few additional fixes. I'm thinking we'll also merge in doc updates here, and some of those will land within a specific release timeframe.

I tried a few ways to group, but it just never looked quite right. My thoughts: after we land this, maybe you can play around with the layout of this and see if you can find something new and interesting for the team to look at and review?

src/content/changelog.md
antfitch: On it!

site/lib/src/models/changelog_model.dart
antfitch: Totally. Thank you for keeping an eye on these things!

site/lib/src/components/pages/changelog/changelog_index.dart
antfitch: Updated!

antfitch: I'll land this, but if you have a design idea, can you bring it to an upcoming meeting with a demo? Would love to see what cool thing you come up with.

site/lib/src/components/pages/changelog/changelog_index.dart
antfitch: Going to leave this as is for now, but only because I think you may have a vision and we need to sync on that. Added a few other comments about creating an updated demo.

ericwindmill (approved):
LGTM The code looks good, the page looks good!

schultek (approved):
LGTM, and the page looks great!

src/data/sidenav/default.yml
antfitch: Tentatively closing, since this should be fixed now. Feel free to re-open or add a new comment if edits are needed.

src/data/changelog.yml
antfitch: Totally get this. Will clean up in a followup CL.

src/data/changelog.yml
antfitch: Will fix in the next PR. Added to the checklist above in this PR description so it doesn't get missed.

src/data/sidenav/default.yml
antfitch: Tentatively closing, since this should be fixed now. Feel free to re-open or add a new comment if edits are needed.

site/lib/src/components/common/dialog.dart
antfitch: We're having a review problem on this site right now, so adding this to the todo fix list in the description so we don't miss this, and I'm also able to push this PR.

antfitch:
Merging this! All, I've updated the todo list at the top this PR to catch any smaller issues. I'll focus on those next in a new PR.

kevmoo:
https://dart.dev/changelog – SWEET!

Are we going to advertise this?

PR #7061, opened 2025-12-07
Generate the diagnostic pages from data files

parlough:
- Updates `dart run dash_site generate-diagnostics` to generate a `src/data/diagnostics.json` file from data in the SDK. This file is not meant to be read or manually edited. It's mostly just a mechanism to cache the results calculated from the SDK data. 
- Stops manually generating the individual diagnostic doc pages found at `/tools/diagnostics/<diagnostic_name>`. Instead use the data from the intermediate `src/data/diagnostics.json` file to generate each page from code. The content of the generated files remains the same with this PR. 

These changes reduce site cloning and full build time, make updates easier to implement and review, and help avoid confusion about those files being the source of the docs.

**Staged example page** to show they remain functional and the same content wise: https://dart-dev--pr7061-misc-generate-diagnostic-pages-gr1n2k31.web.app/tools/diagnostics/abstract_super_member_reference

### DIFF
--- a/site/lib/src/client/global_scripts.dart
+++ b/site/lib/src/client/global_scripts.dart
@@ -21,6 +21,7 @@ void setUpSite() {
   _setUpTableOfContents();
   _setUpReleaseTags();
   _setUpTooltips();
+  _setUpSteppers();
 }
 
 void _setUpSidenav() {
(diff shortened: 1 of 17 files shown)


src/data/diagnostics.json
antfitch, line 1: Before we move to JSON, let's discuss the architectural structure we are going to use moving forward for this and other components like the glossary and changelog at our next infrastructure meeting on Thursday.

Take a close look at the structure of changelog.yml here: https://github.com/dart-lang/site-www/pull/7056/files#diff-e22c5b326abca90af21f295ded86832bb7ac29f663523d5abe98d368c13928e9

Top of mind for me is how we handle markdown. I can see that with YML, this is possible with our markdown parser (see the changelog PR). Would you like to test to see if markdown support can be added to JSON?

Other general thoughts:
- Make sure markdown is supported in the JSON, YML, or whatever text format we decide to use.
- Let's enforce the 80-line rule for strings (thinking especially about the descriptions).
- Let's keep meta content at the top of each entry and move the summary & description to the bottom since they will likely require multiple lines.

parlough: Thanks for taking a look @antfitch!

> Before we move to JSON, let's discuss the architectural structure we are going to use moving forward for this and other components like the glossary and changelog at our next infrastructure meeting on Thursday.

This one is a bit different than the glossary and changelog. This JSON file is just for intermediate storage, not meant to be edited or reviewed by hand. YAML is more painful to serialize and deserialize from and generally should be avoided unless it's for human editing. Happy to discuss more at the meeting if needed.

> Would you like to test to see if markdown support can be added to JSON?

I'm not sure what you mean? The Markdown strings stored in this file are rendered as Markdown content.

> Let's enforce the 80-line rule for strings 

I generally agree, but since this file is just for storage, not editing or viewing, I'm not sure it's worth figuring something out there, particularly with JSON not supporting line breaks within string literals.

> Let's keep meta content at the top of each entry and move the summary & description to the bottom since they will likely require multiple lines.

Could you explain what you mean here? Do you have an example?

ericwindmill: I agree with Parker. We should use JSON for ephemeral data and data that isn't meant to be read or edited by a human. We should use YAML in the other case.

antfitch: Per the meeting today, JSON sounds good. @parlough I think it would be better to move the longer values to the bottom for each entry, but this is personal preference. If no one will be looking at or editing this data and it will be automatically created, it really doesn't matter. Closing this.

PR #7074, opened 2025-12-13
[js-interop] move the New/Old API page to the legacy page

kevmoo:


kevmoo:
@antfitch – finally figured out that I broke the anchor. Fixed!

parlough (approved):
I pushed some small fixes and cleanup, but otherwise looks good to me. Thanks!

Since you have requested changes, please take another look @antfitch. Thank you :D

PR #7099, opened 2026-01-07
fix: make heading anchors copy full URL to clipboard fixes #6756

### LINKED ISSUE #6756: [user experience] Copy anchor link to clipboard in clicking header
### Describe the problem

Steps to reproduce:

1. browse to https://dart.dev/tools/pub/packages
2. scroll down to "Upgrading a dependency"
3. click "#" behind "Upgrading a dependency"
4. Go to a different tab with a text field
5. Hit ctrl+v.

Expected behavior

* Pastes in https://dart.dev/tools/pub/packages#upgrading-a-dependency

Actual behavior

* Pastes in whatever you had on your clipboard, not the link of the anchor on the website.

Anushreebasics:
### Summary
This PR fixes issue #6756 where clicking the '#' anchor next to headings
did not copy the section link to the clipboard.  

### Changes
- Updated `HeaderWrapperExtension` to add an inline `onclick` handler
  for the heading anchor `<a>` elements.
- Clicking '#' now:
  - Copies the full URL including fragment to the clipboard
  - Updates the URL in the address bar
  - Prevents default page jump behavior
  - Temporarily replaces '#' with a smaller 'Copied!' text for 1.2 seconds to provide visual feedback

### Testing
- Verified that clicking '#' on h2–h5 headings copies the correct URL
- Confirmed that the page URL updates without scrolling
- Works across multiple headings on the same page

### Notes
- This approach adds behavior directly in the anchor element.
- Alternative approach would be a global client-side handler, but this
  is safe and consistent with current Jaspr HTML generation.

Fixes #6756 

reference video: 

https://github.com/user-attachments/assets/06411e8d-07c2-4057-a20e-ace125897a16

### DIFF
--- a/src/content/tools/hooks.md
+++ b/src/content/tools/hooks.md
@@ -235,12 +235,21 @@ create a `build.dart` script similar to the following:
 [`BuildInput`]: {{site.pub-api}}/hooks/latest/hooks/BuildInput-class.html
 [`BuildOutputBuilder`]: {{site.pub-api}}/hooks/latest/hooks/BuildOutputBuilder-class.html
 
+:::note
+When naming your dynamic libraries, ensure the name is
+consistent across all target architectures and SDKs.
+On Apple platforms, Flutter's build system relies on
+consistent filenames to correctly generate frameworks and XCFrameworks.
+For more details, check out [Binding to native code][flutter-native]
+on the Flutter documentation site.
+:::
+
 ### Automatically bundled assets {: #generate-assets }
 
 The hooks are run automatically when invoking the `run`, `build`, or `test`
 commands. The resulting assets are stored in the output directory specified in
 the hook input. The Dart SDK then automatically bundles those assets with
-your Dart app so that that they can be accessed at run time.
+your Dart app so that they can be accessed at run time.
 
 ### Use assets {: #reference-assets }
 
@@ -252,6 +261,15 @@ only output assets in their own package. `CBuilder` in
 the build hook in the previous example outputs the asset ID
 `package:native_add_library/native_add_library.dart`, and is
 based on the `packageName` and `assetName`.
+
+:::tip
+As a best practice, use the Dart library URI where you
+intend to use the code asset as the asset ID.
+While you can use any asset ID that starts with your package name,
+using the library URI enables Dart to
+automatically connect the asset at run time without
+needing to specify an `assetId` in the `@Native` annotation.
+:::
 
 The following example illustrates how to bind to the native
 C function `add` from `native_add_library.c` and call it:
@@ -340,9 +358,11 @@ See the following links for more information:
 * [Code assets package][]
 * [Code assets library reference][]
 * [C interop][]
+* [Binding to native code in Flutter apps][flutter-native]
 
 [Hooks package]: {{site.pub-pkg}}/hooks
 [Hooks library reference]: {{site.pub-api}}/hooks/latest/hooks/
 [Code assets package]: {{site.pub-pkg}}/code_assets
 [Code assets library reference]: {{site.pub-api}}/code_assets/latest/code_assets/
 [C interop]: /interop/c-interop
+[flutter-native]: {{site.flutter-docs}}/platform-integration/bind-native-code


site/lib/src/extensions/header_processor.dart
```diff
@@ -50,6 +50,18 @@ final class HeaderWrapperExtension implements PageExtension {
             'class': 'heading-link',
             'href': '#$headerId',
             'aria-label': "Link to '$headerText' section",
+            // 👇 COPY TO CLIPBOARD LOGIC
+            'onclick': '''
```
parlough: Implementing this as an inline JS string is quite repetitive in the output HTML, harder to maintain, and causes us to lose out on static checks.

Perhaps a more robust approach would be to add a new Dart function to https://github.com/dart-lang/site-www/blob/main/site/lib/src/client/global_scripts.dart that finds the relevant anchors and adds the behaviors to them?

Another option would be a hydrated Jaspr client component, but I think the single global handler should be sufficient for this case.

parlough (changes requested):
Thanks for the PR, I appreciate it! However I have some concerns about the inline-JS approach.

Could you migrate it to a setup function written in Dart? Or a Jaspr client component if you prefer that? Let me know if you have any concerns or hit any issues though. Thanks again!

Anushreebasics:
@parlough please review the changes:

Moved the heading anchor click behavior from inline JavaScript to a single global handler in global_scripts.dart.
This avoids duplicated JS in the generated HTML, improves maintainability, and keeps content generation separate from client-side behavior. The new handler copies the section URL to the clipboard, updates the hash without jumping, and shows brief visual feedback.

site/lib/src/client/global_scripts.dart
```diff
@@ -24,6 +25,46 @@ void setUpSite() {
...
+    anchor.addEventListener(
+      'click',
+      ((web.Event event) {
+        event.preventDefault();
+
+        final headingId = anchor.dataset['headingId'];
+        if (headingId == null || headingId.isEmpty) return;
+
+        final url =
+            '${web.window.location.origin}${web.window.location.pathname}#$headingId';
```
parlough: To make this a bit safer and also easier to read, we could use the `Uri` class provided by Dart:


```suggestion
        final url = Uri.parse(web.window.location.href)
            .replace(fragment: headingId)
            .toString();
```

site/lib/src/client/global_scripts.dart
```diff
@@ -24,6 +25,46 @@ void setUpSite() {
...
+  for (var i = 0; i < anchors.length; i++) {
+    final anchor = anchors.item(i) as web.HTMLAnchorElement;
+
+    anchor.addEventListener(
+      'click',
+      ((web.Event event) {
+        event.preventDefault();
+
+        final headingId = anchor.dataset['headingId'];
+        if (headingId == null || headingId.isEmpty) return;
```
parlough: Rather than adding a new data entry, would it make sense to just reuse the `href` already present on the anchor?

site/lib/src/client/global_scripts.dart
```diff
@@ -24,6 +25,46 @@ void setUpSite() {
...
+
+        // Temporary visual feedback
+        final originalText = anchor.textContent;
+        anchor.textContent = 'Copied!';
+        web.window.setTimeout(
+          (() {
+            anchor.textContent = originalText;
+          }).toJS,
+          1200 as JSAny,
+        );
```
parlough: Rather than using `setTimeout` which is a bit awkward through Dart, we can instead use `Future.delayed` which is built in :D 


```suggestion
        Future<void>.delayed(const Duration(milliseconds: 1200), () {
          anchor.textContent = originalText;
        });
```

Anushreebasics:
@parlough please review the changes

Anushreebasics:
@antfitch  @conooi, please review this PR

Anushreebasics:
@parlough please review this PR

conooi:
Hi @Anushreebasics! Thanks for all your help. Adding in @khanhnwin to review your code.

Anushreebasics:
@khanhnwin please review

khanhnwin:
This looks great, thanks @Anushreebasics!

Anushreebasics:
@conooi the branch is up to date with the main branch and squashed them to one commit, kindly review

Anushreebasics:
@conooi please review, I have fixed the formatting errors

conooi:
@parlough please review when you have time :)

Anushreebasics:
@parlough please review

parlough (approved):
Sorry about the delay in reviewing this again!

Looks good to me. I did push a small change adding a fallback behavior incase the clipboard access failed.

Thanks so much @Anushreebasics!

PR #7136, opened 2026-01-31
[prototype] Dart Docs Skills

antfitch:
Note: this skill I added here was updated by @parlough and landed in https://github.com/dart-lang/site-www/pull/7159

This change introduces the "Agent Skill" architecture to the repository, replacing the experimental Model Context Protocol (MCP) server. This lightweight approach allows both human developers and AI agents (like Antigravity) to discover and execute maintenance tasks using standard CLI tools and natural language triggers.

There is one skill in this initial release to updated the SDK changelog.

How to Test

1. [user] Open the site-www workspace in your AI IDE.
2. [user] In the chat, simply type:
3. [user] "skill: update sdk changelog 3.11"
4. [agent] Discover the skill in the registry.
5. [agent] Execute the script (dart run .../sync.dart --version 3.11.0).
6. [agent] Update src/data/changelog.yml with correctly tagged entries. 

Key Changes

- Lightweight Architecture: Removed the heavy dart-site-mcp-server. Maintenance tasks are now correctly implemented as standalone CLI scripts wrapped in an "Agent Skill" definition.
- "Agent Skill" Standard: Introduced tool/skills/ as a registry for capabilities.
- Registry: tool/skills/README.md allows agents to auto-discover capabilities.
- Manifests: Each skill (e.g., tool/skills/update_sdk_changelog/SKILL.md) contains step-by-step instructions and natural language triggers.
- Smart CLI Tool: Refactored the changelog logic into tool/skills/update_sdk_changelog/scripts/sync.dart.
- Smart Tagging: Implemented regex-based inference to automatically detect and apply tags like breaking, removed, new, and fixed based on the entry description (including recognizing phrases like "no longer compile").
- Auto-Discovery: Added .agent/workflows/maintenance.md, ensuring that any agent opening this repository implicitly knows how to find and use these tools without manual configuration.


tool/mcp_server/lib/tools/changelog.dart
```diff
@@ -0,0 +1,304 @@
...
+        if (description.isNotEmpty) {
+          entries.add(
+            ChangelogEntry(
+              version: targetVersion,
+              releaseDate: 'TBD',
+              area: currentArea ?? 'SDK',
+              subArea: currentSubArea,
+              description: description,
+              tags: _inferTags(description),
+              link: pendingLink,
```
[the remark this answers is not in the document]
antfitch: Updated.

tool/mcp_server/lib/tools/changelog.dart
```diff
@@ -0,0 +1,304 @@
...
+
+  List<String> _inferTags(String description) {
+    final tags = <String>[];
+    final lower = description.toLowerCase();
+    if (lower.contains('breaking change')) {
+      tags.add('breaking');
+    }
+    if (tags.isEmpty) tags.add('changed');
+    return tags;
+  }
```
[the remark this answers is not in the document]
antfitch: Updated.

tool/mcp_server/bin/server.dart
```diff
@@ -0,0 +1,61 @@
...
+    // Register prompts
+    addPrompt(
+      Prompt(
+        name: 'update_sdk_changelog',
+        description:
+            'Workflow instructions for updating the Dart SDK changelog. Use this whenever asked to update, sync, or modify the changelog for a new version.',
+      ),
+      (request) async {
+        final content = await File(
+          'tool/mcp_server/update_sdk_changelog.md',
```
jakemac53: This assumes a cwd of site-www which isn't always going to be the case. You can use Isolate.resolvePackageUri to find the root of this package. I am on a phone currently but you can ping me for more details if needed.

tool/mcp_server/tool/run_changelog_sync.dart
```diff
@@ -0,0 +1,93 @@
+import 'dart:convert';
+import 'dart:io';
+
+void main(List<String> args) async {
```
jakemac53: This file and the ones below are basically implementing a simple MCP client that does one tool call - the dart_mcp package does come with an McpClient implementation which you could use. But given the simplicity here I am not sure it would be less code.

Is this intended mostly as just a test of the MCP server? If not I would just extract the actual implementation out to a separate file and call it directly instead of going through MCP for it.

jakemac53: Also if these are intended to just be testing things, I would just write them as tests, see the https://github.com/dart-lang/ai/blob/main/pkgs/dart_mcp/test/client_and_server_test.dart test as an example and corresponding TestEnvironment in test_utils.dart which lets you unit test these things very quickly all within the same process.

tool/mcp_server/bin/server.dart
antfitch: Thanks for catching this. I've updated the code. Take a close look. Tentatively closing this, but if the issue isn't fix, feel free to reopen.

tool/mcp_server/tool/run_changelog_sync.dart
antfitch: These aren't for testing. MCP Server will call these as part of the changelog update. There will me more tasks soon. My thoughts... as we add new tasks, we revisit. Tentatively closing, but feel free to open if more work is needed.

tool/mcp_server/bin/server.dart
```diff
@@ -46,6 +49,9 @@ final class SiteMcpServer extends MCPServer with ToolsSupport, PromptsSupport {
       },
     );
 
+    // Calling super.initialize is required to set up capabilities
+    final result = await super.initialize(request);
+
     return result;
```
jakemac53:
```suggestion
    return super.initialize(request);
```

tool/mcp_server/tool/run_changelog_sync.dart
jakemac53: This isn't what the MCP Server invokes, this is just a script that invokes the MCP server and makes a tool call. These files could all be deleted and the MCP Server would work just fine... (these looks like scripts that AGY generated to do its own verification of the MCP server or something).

antfitch:
Per a conversation with @parlough, I'm going to try something a bit different. No MCP Server (@jakemac53 this is similar to what you were suggesting). I'll have an update soon.

antfitch:
Switched the architecture over to skills. @johnpryan adding you since we're moving into skills territory and this is your specialty.

tool/mcp_server/bin/server.dart
antfitch: Removed the server code.

tool/mcp_server/tool/run_changelog_sync.dart
antfitch: Moved over to skills. On my machine the agent was hitting these. It was difficult for me to tell if they were part of the flow or not. The agent reported that it was. I'm guessing it was confused or taking the lazy route during testing. Closing this since we've moved away from the MCP Server and to skills. The skills are directly calling these scripts.

antfitch:
@parlough, the dart-website-staging is failing. It looks like it's due to the datetime format in data_processor.dart. Not sure if there was a change to this file recently. In the sync.dart file part of this PR, there is a `releaseDate` field, but this is a string type. It's weird, but wanted to bring this to your attention just incase.

.agent/workflows/maintenance.md
parlough, line 1: I think this direction is much easier to maintain than the previous MCP server, thanks for making this change! However, it seems this use of a skill discovery workflow (alongside the `tool/skills/README.md` document) is reimplementing the already built-in support for discovering and following [agent skills](https://antigravity.google/docs/skills).

Any particular reason to use a workflow like this? If not, it'd be nice to use the more broadly supported skill mechanism. Based on this implementation, I opened https://github.com/dart-lang/site-www/pull/7159 as an alternative proposal with some other changes if you want to take a look.

Let me know what you think. Thanks again @antfitch!

Issue #7145, opened 2026-02-03
Dedicated location for pub source descriptors as command-line arguments


dcharkes:
In https://dart-review.googlesource.com/c/sdk/+/477560, we expand the use of pub source descriptors from `dart pub add <package-name>:<descriptor>` to `dart run <package-name>@<descriptor>` and `dart install <package-name>@<descriptor>`.

I believe it would be worthwhile for https://dart.dev/tools/pub/cmd/pub-add#source-descriptor to have a dedicated section on dart.dev. One option:

* https://dart.dev/tools/dart-tool after the section https://dart.dev/tools/dart-tool#available-commands. Reasoning there are 3 available commands after we land the change that accept pub source descriptor command-line arguments.

Side note: The source descriptors are not full YAML. They are "YAML flow style". (E.g. I don't think we support throwing YAML block style with newlines in command-line arguments?)

cc @sigurdm @antfitch

sigurdm:
I guess there is a big overlap with https://dart.dev/tools/pub/dependencies#dependency-sources.

dcharkes:
Yeah, but I suppose you want to use YAML block style there. 😄 

But definitely the new page/section documenting the source descriptors with YAML flow style for command-line arguments should link to https://dart.dev/tools/pub/dependencies#dependency-sources, and https://dart.dev/tools/pub/dependencies#dependency-sources can link back to the new page/section we add for source descriptors for command-line arguments.

sigurdm:
Technically nothing prevents you from using block style yaml on the command-line or flow style in a pubspec. This works

<img width="756" height="218" alt="Image" src="https://github.com/user-attachments/assets/1a770b71-9156-4d46-bba8-3ed0bb8c6b7a" />

dcharkes:
True, but to make the commands more readable, I presume all our examples are going to use flow-style right?

I don't care a whole lot if we have a package for source descriptors for command-line args separate from the pubspec dependency sources. My gut feeling says two separate pages that detail the default way to do things as command-line arguments and the default way to do things in a YAML file is going to be the most clear.

Happy to roll whatever you, @antfitch and @parlough think will be the clearest for our users.

PR #7157, opened 2026-02-08
Centralize documentation for pub 'package descriptors'

parlough:
- Adds a new section to the [Package dependencies](https://dart.dev/tools/pub/dependencies) page which describes and outlines "Package descriptors" (previously referenced as source descriptors), covering the simple version-constraint form and the more complex YAML style.
- Then updates the existing docs for `dart pub add` and `dart pub unpack` to cross-reference the new section.
- Removes the old version notes now that the syntax has been supported for over 3 years now.
- Adds a new `/to/package-descriptors` tooling redirect that points to the new section and can be used from the CLI tool.

**Open question:** I currently included this in the existing [Package dependencies](https://dart.dev/tools/pub/dependencies) page since it overlaps a lot with it and also includes the discussion of the dependency source types and version constraints. Do you think this ended up fine or do you think this is better covered on a standalone page?

Resolves https://github.com/dart-lang/site-www/issues/7145

**Staged:** https://dart-dev--pr7157-feat-package-descriptors-ncoslalk.web.app/to/package-descriptors


src/content/tools/pub/dependencies.md
```diff
@@ -275,6 +282,86 @@ If it's `flutter`, the dependency is satisfiable as long as:
 
 If it's an unknown identifier, the dependency is always considered unsatisfied.
 
+<a id="source-descriptors" aria-hidden="true"></a>
+
+## Package descriptors
+
+Several [`dart` tool][dart-tool] commands, such as
+[`dart pub add`][pub-add] and [`dart pub unpack`][pub-unpack],
```
dcharkes, line 290: Maybe add `dart run` and `dart install` with links to those pages.

See https://dart-review.googlesource.com/c/sdk/+/477560

src/content/tools/pub/dependencies.md
```diff
@@ -275,6 +282,86 @@ If it's `flutter`, the dependency is satisfiable as long as:
...
+to specify the version or dependency source of a package.
+
+The simplest form of a package descriptor is
+a [version constraint](#version-constraints),
+assuming the default source of the [pub.dev site][pubsite].
+For example, the following command uses a descriptor of `^1.2.3`
+to [add a dependency][pub-add] on `package:foo` with a constraint of `^1.2.3`.
+
+```console
+dart pub add foo:^1.2.3
```
dcharkes, line 301: and then here `dart install foo@^1.2.3` and `dart run foo@^1.2.3`

dcharkes (approved):
> **Open question:** I currently included this in the existing [Package dependencies](https://dart.dev/tools/pub/dependencies) page since it overlaps a lot with it and also includes the discussion of the dependency source types and version constraints. Do you think this ended up fine or do you think this is better covered on a standalone page?

A 'package descriptor' is used outside the `dart pub` command. However, if you use them in `dart run` or `dart install` you're still going through Pub to fetch those versions. So "package descriptor" is still very much a Pub concept. So, I think nesting it under the Pub tool is fine.

Thanks @parlough 🙏

src/content/tools/pub/dependencies.md
parlough: Since that CL just landed in the main channel, I'll likely wait a bit and add those mentions alongside the other documentation for those additions as a separate follow-up.

Tracking in https://github.com/dart-lang/site-www/issues/7171

PR #7169, opened 2026-02-12
Remove accidentally introduced empty file

parlough:
I'm not sure what might have caused it, but was introduced in https://github.com/dart-lang/site-www/commit/cad932f854fd17625acc623459f489d436f2378b. I believe this was an accident, but if it has a purpose, feel free to close this and let me know. Thanks!


conooi:
Thanks for catching this @parlough!!

PR #7175, opened 2026-02-16
Documented that pub workspaces can be nested

### LINKED ISSUE #6890: Mention that pub workspaces can nest.

SoufianElfouzari:
Adds documentation clarifying that pub workspaces can be nested.

This change explains that a package within a workspace can define its own
workspace by including a `workspace` section in its `pubspec.yaml` file.

Fixes #6890

---

- [x] This PR follows the Google Developer Documentation Style Guidelines.
- [x] This PR uses semantic line breaks of 80 characters or fewer.


src/content/tools/pub/workspaces.md
```diff
@@ -40,7 +40,14 @@ There are some downsides to this setup:
   separate analysis contexts for each package, increasing memory usage.
 
 Pub allows you to organize your repository as a _workspace_ using a single
-shared resolution for all your packages. 
+shared resolution for all your packages.
+
+Pub workspaces can also be nested. A package that is part of a workspace can
+define its own workspace by including a `workspace` section in its
+`pubspec.yaml` file. This allows for a hierarchical organization of large
```
parlough: I'm not sure this is the best location for this mention as the `workspace` pubspec field hasn't been introduced at this point, as this is just the initial overview.

What do you think about moving the new content you wrote to its own section after the "Glob-pattern support" section? Then we could even add more details, such as an example if helpful.


````
## Glob-pattern support

// ...

## Nested workspaces

Description of nesting support here...
```

parlough (commented):
Thanks so much for the PR @SoufianElfouzari. I'm sorry about our delay in reviewing it.

Overall your added content looks like a good, brief introduction to the capability. However, I'm not so sure we should provide so many details so early on in the introduction. What do you think about moving the clarification to its own section?

parlough (approved):
I went ahead and some additions and moved the reference to a new standalone section. Thanks again for tackling this!

PR #7177, opened 2026-02-17
Add "sound" glossary entry for type soundness

parlough:
Mostly duplicated from https://dart.dev/language/type-system#what-is-soundness before I begin an overhaul of that page. If you have thoughts about what should be covered in addition or instead, do let me know. Thanks!

**Staged:** https://dart-dev--pr7177-feat-sound-glossary-entry-wtvukc96.web.app/glossary/sound

Contributes to https://github.com/dart-lang/site-www/issues/4745

### DIFF
--- a/src/data/glossary.yml
+++ b/src/data/glossary.yml
@@ -1602,7 +1602,7 @@
       When you change one piece of code,
       the sound type system enables Dart tooling to
       warn you about other code that broke.
-    - **Better ahead-of-time (AOT) compilation.**
+    - **Better [ahead-of-time (AOT)](#aot) compilation.**
       Sound types help the compiler generate
       smaller, more efficient native code.
 


eernstg (approved):
This is very well written, LGTM!

One thing came to mind, though:

```dart
class A<X> {
  void Function(X) f; // This is a "non-covariant" member.
  A(this.f);
}

void main() {
  A<num> a = A<int>((arg) {});
  a.f; // Checks the type of the actual value, and throws.
}
```

Non-covariant members (https://github.com/dart-lang/language/issues/296) are subject to a client-side run-time type check. That is, the call site will evaluate the expression and then perform the dynamic type check which is needed, and then throw if the resulting value doesn't have the statically known type.

This is a very special situation because the value _is there_ (`(a as dynamic).f` works fine!), but we aren't allowed to access it if the static type is violated, and this is something which is never possible except when a variable has a type where a type variable of the enclosing class/mixin/... occurs in the variable type in a non-covariant position (here: as a parameter type of a function type).

You can still argue that we have this nice property: "An expression of type `T` will never evaluate to an object which isn't typable as a `T`", but it's kind of cheating if we're just checking the type immediately before we give the value of the expression to the client&mdash;if this is considered OK then you could have a completely unsound type system and then just perform this kind of check for every expression evaluation. So it's crucial that non-covariant members are really rare, so we basically don't have to worry about this behavior in normal code.

(This kind of situation can be flagged by enabling `unsafe_variance`, so it's possible to avoid it, making it even rarer, but the real solution is to have statically checked variance.)

It is probably not useful to mention non-covariant members in this glossary entry, but I thought it would be useful to make sure it's on the table.

PR #7178, opened 2026-02-17
Add an entry for "top type" to the glossary

parlough:
Adds an entry for "top type" to the glossary, as we already have an entry for "bottom type", and I will reference this in upcoming type system documentation.

**Staged:** https://dart-dev--pr7178-glossary-top-type-p2l0m99j.web.app/glossary/top-type

Contributes to https://github.com/dart-lang/site-www/issues/4745

### DIFF
--- a/src/data/glossary.yml
+++ b/src/data/glossary.yml
@@ -1873,14 +1873,16 @@
 
 - term: "Top type"
   short_description: |-
-    A type that is a supertype of all other types.
-    In Dart, `Object?`, `dynamic`, and `void` are top types.
+    A type that is a supertype of every other type,
+    including `Object?`.
   long_description: |-
-    A **top type** is a type that sits at the top of the type hierarchy.
-    Every other type is a [subtype](#subtype) of a top type,
+    A **top type** is any type that is a supertype of `Object?`.
+    Because `Object?` is itself a supertype of every other Dart type,
+    a top type sits at the very top of the type hierarchy,
     so you can assign any value to a variable with a top type.
 
-    Dart has three top types:
+    Dart has many top types,
+    but the primary ones are the following three:
 
     - **`Object?`** -
       The nullable supertype of all types.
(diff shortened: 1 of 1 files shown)


src/data/glossary.yml
```diff
@@ -1767,6 +1767,69 @@
...
+- term: "Top type"
+  short_description: |-
+    A type that is a supertype of all other types.
+    In Dart, `Object?`, `dynamic`, and `void` are top types.
+  long_description: |-
+    A **top type** is a type that sits at the top of the type hierarchy.
+    Every other type is a [subtype](#subtype) of a top type,
+    so you can assign any value to a variable with a top type.
+
+    Dart has three top types:
```
eernstg: There are many other top types: Whenever `T` is a top type, `FutureOr<T>` is a top type as well, and so is `F?` where `F` is a type alias whose expansion is `Object` or `dynamic`, or any other supertype of `Object`, and `FutureOr<Object>?`, and more.

We have a syntactic characterization of top types in the type predicate `TOP(_)`, which is defined [here](https://github.com/dart-lang/language/blob/master/accepted/2.12/nnbd/feature-specification.md#helper-predicates).

I guess it would be fine to just leave the door open: "Dart has many top types, but the most important ones are the following three:". Right after the itemized list a couple of others like `FutureOr<Object?>`, `FutureOr<Object>?`, and `dynamic?` could be mentioned.

parlough: Hmm, I'm glad I requested your review because I didn't know to consider these cases. I'll give some thought to how I want to incorporate these. Thanks @eernstg!

Besides the `FutureOr` case, is it accurate to say those other types are considered top types after normalization to one of the core/bare top types?

eernstg: Thanks! All those types are top types, before or after normalization. It might be true that all the composite top types (using `FutureOr`, `?`, etc.) will be normalized into one of the three main top types, but that would be a proof by induction in the rules about `TOP(_)` and `NORM(_)`, and I won't promise that there are no funny corner cases. ;-)

In any case, it's probably useful to know that `FutureOr<Object?>` is also a top type, and that there are others as well, because these types might occur in actual code. For example, `FutureOr<X>` could occur in a situation where `X` is a type parameter, and we happen to know that `X` is a top type, and then `FutureOr<X>` will also be a top type.

The starting point is that a top type is _any_ type which is a supertype of `Object?`, so maybe that should be mentioned first. Then the specific main top types could be mentioned, and then just a hint about all the remaining forms.

PR #7195, opened 2026-02-27
Miscellaneous agent skill cleanup

parlough:
- Removes the symlink to the `.agent` directory now that most tools respect `.agents` now or have a way to do so with config options. This includes the latest stable release of Gemini CLI which even prioritizes the `.agents` directory skills.
- Adds the new `update-whats-new` skill to our `.agents` readme which has an index of our skills.
- Update the new `update-whats-new` skill to follow our formatting guidelines so it's easier to read in source as well as easier to update and review changes to in the future.
- Slightly adjust the wording of some instructions in the `update-whats-new` skill.


conooi (approved):
Thank you, Parker!!

PR #7211, opened 2026-03-10
Introduce two new docs covering codecs and converters

### LINKED ISSUE #5175: Update or remove 'Converters and codecs' article
### Page URL

https://dart.dev/articles/archive/converters-and-codecs

### Page source

https://github.com/dart-lang/site-www/tree/main/src/_articles/archive/converters-and-codecs.md

### Describe the problem

The article has been marked as obsolete for a long time, but still is hosted and provides valuable information.

### Expected fix

Either remove the article or update it for modern standards:

- Remove extra "article information (publish date, up-front authors, etc)
- Update text and code samples for modern Dart
- Add code excerpts **AND** tests for code samples
- Reduce amount of long text (more focused)
- Copy edit and update for modern GSG/our writing guidelines
- Update formatting/style for our current site guidelines
- After previous steps, remove obsolete labels
- Find home in sidenav

### Additional context

_No response_

parlough:
Inspired by the pre-existing [Converters and codecs page](https://dart.dev/libraries/convert/converters-and-codecs), introduce two new articles that are more focused than the original and follow our current standards for writing and structure.

- [Use converters and codecs](https://dart-dev--pr7211-feat-converters-and-codecs-6xdboazt.web.app/libraries/convert/converters-and-codecs): Introduces Dart developers to codecs and converters, how to use them on their own, how to compose them, and how to use them with streams.
- [Build custom converters and codecs](https://dart-dev--pr7211-feat-converters-and-codecs-6xdboazt.web.app/libraries/convert/build-custom-codecs): A guide on how to build your own converters and codecs and some considerations to make when doing so.

Resolves https://github.com/dart-lang/site-www/issues/5175

### DIFF
--- a/examples/libraries/lib/convert/build_custom_codecs.dart
+++ b/examples/libraries/lib/convert/build_custom_codecs.dart
@@ -1,0 +1,164 @@
+import 'dart:convert';
+
+// #docregion caesar-encoder-sink
+/// A [StringConversionSink] that applies a Caesar cipher [_shift] and
+/// forwards the result to the [_output] sink.
+class _CaesarEncoderSink extends StringConversionSinkBase {
+  final int _shift;
+  final StringConversionSink _output;
+
+  _CaesarEncoderSink(this._shift, this._output);
+
+  @override
+  void addSlice(String chunk, int start, int end, bool isLast) {
+    final buffer = StringBuffer();
+    for (var i = start; i < end; i++) {
+      buffer.writeCharCode(_shiftCodeUnit(chunk.codeUnitAt(i), _shift));
+    }
+    _output.add(buffer.toString());
+    if (isLast) {
+      _output.close();
+    }
+  }
+
+  @override
+  void close() => _output.close();
+}
+// #enddocregion caesar-encoder-sink
+
+// #docregion basic-encoder, caesar-encoder
+/// Encodes a string by shifting each letter forward in the alphabet.
+class CaesarEncoder extends Converter<String, String> {
+  /// The number of positions to shift each letter forward.
+  final int shift;
+
+  /// Creates an encoder that shifts each letter by [shift] positions.
+  const CaesarEncoder(this.shift);
+
+  @override
+  String convert(String input) {
+    final buffer = StringBuffer();
+    for (final codeUnit in input.codeUnits) {
+      buffer.writeCharCode(_shiftCodeUnit(codeUnit, shift));
+    }
+    return buffer.toString();
+  }
+  // #enddocregion basic-encoder
+
+  // #docregion start-chunked-conversion
+  @override
+  StringConversionSink startChunkedConversion(Sink<String> sink) {
+    // Wrap the output sink if it isn't already
+    // a StringConversionSink.
+    // ignore: close_sinks
+    final stringSink = sink is StringConversionSink
+        ? sink
+        : StringConversionSink.from(sink);
+    return _CaesarEncoderSink(shift, stringSink);
+  }
+
+  // #enddocregion start-chunked-conversion
+  // #docregion basic-encoder
+}
+// #enddocregion caesar-encoder
+
+// #docregion shift-code-unit
+/// Shifts the specified [codeUnit] by
+/// [shift] positions in the alphabet.
+///
+/// Only shifts lowercase ASCII letters (a-z).
+/// All other characters are returned unchanged.
+int _shiftCodeUnit(int codeUnit, int shift) {
+  const a = 0x61;
+  const z = 0x7A;
+  if (codeUnit >= a && codeUnit <= z) {
+    return a + (codeUnit - a + shift) % 26;
+  }
+  return codeUnit;
+}
+// #enddocregion basic-encoder, shift-code-unit
+
+// #docregion basic-decoder
+/// Decodes a Caesar-cipher-encoded string by
+/// shifting each letter backward in the alphabet.
+class CaesarDecoder extends Converter<String, String> {
+  /// The number of positions to shift each letter backward.
+  final int shift;
+
+  /// Creates a decoder that reverses a Caesar cipher with [shift] positions.
+  const CaesarDecoder(this.shift);
+
+  @override
+  String convert(String input) {
+    final buffer = StringBuffer();
+    for (final codeUnit in input.codeUnits) {
+      // Shift backward by shifting forward by (26 - shift).
+      buffer.writeCharCode(_shiftCodeUnit(codeUnit, 26 - shift));
+    }
+    return buffer.toString();
+  }
+  // #enddocregion basic-decoder
+
+  @override
+  StringConversionSink startChunkedConversion(Sink<String> sink) {
+    // ignore: close_sinks
+    final stringSink = sink is StringConversionSink
+        ? sink
+        : StringConversionSink.from(sink);
+    return _CaesarEncoderSink(26 - shift, stringSink);
+  }
+
+  // #docregion basic-decoder
+}
+// #enddocregion basic-decoder
+
+// #docregion caesar-codec
+/// A codec that encodes and decodes strings using a
+/// [Caesar cipher](https://wikipedia.org/wiki/Caesar_cipher).
+class CaesarCodec extends Codec<String, String> {
+  /// The number of positions to shift each letter.
+  final int shift;
+
+  @override
+  CaesarEncoder get encoder => CaesarEncoder(shift);
+
+  @override
+  CaesarDecoder get decoder => CaesarDecoder(shift);
+
+  /// Creates a Caesar cipher codec with the given [shift].
+  const CaesarCodec(this.shift);
+
+  /// Creates a [CaesarCodec] that uses ROT13 encoding.
+  const CaesarCodec.rot13() : shift = 13;
+}
+// #enddocregion caesar-codec
+
+// #docregion use-encoder
+void useEncoderExample() {
+  const encoder = CaesarEncoder(3);
+  print(encoder.convert('hello')); // khoor
+  print(encoder.convert('xyz')); // abc
+}
+// #enddocregion use-encoder
+
+// #docregion use-codec
+void useCodecExample() {
+  const caesar = CaesarCodec(3);
+
+  final encoded = caesar.encode('hello');
+  print(encoded); // khoor
+
+  final decoded = caesar.decode(encoded);
+  print(decoded); // hello
+
+  // The `inverted` getter returns a new codec that
+  // applies converts in the inverse direction of the codec.
+  final inverted = caesar.inverted;
+  print(inverted.encode('khoor')); // hello
+}
+// #enddocregion use-codec
+
+// #docregion top-level-instance
+/// A [CaesarCodec] with the standard shift of 13 (ROT13).
+const rot13 = CaesarCodec.rot13();
+// #enddocregion top-level-instance
--- a/examples/libraries/lib/convert/build_custom_codecs.dart
+++ b/examples/libraries/lib/convert/build_custom_codecs.dart
@@ -28,7 +28,9 @@ class _CaesarEncoderSink extends StringConversionSinkBase {
 
 // #docregion basic-encoder, caesar-encoder
 /// Encodes a string by shifting each letter forward in the alphabet.
+// #docregion chunked-encoder
 class CaesarEncoder extends Converter<String, String> {
+  // #enddocregion chunked-encoder
   /// The number of positions to shift each letter forward.
   final int shift;
 
@@ -45,7 +47,8 @@ class CaesarEncoder extends Converter<String, String> {
   }
   // #enddocregion basic-encoder
 
-  // #docregion start-chunked-conversion
+  // #docregion chunked-encoder
+
   @override
   StringConversionSink startChunkedConversion(Sink<String> sink) {
     // Wrap the output sink if it isn't already
@@ -57,10 +60,9 @@ class CaesarEncoder extends Converter<String, String> {
     return _CaesarEncoderSink(shift, stringSink);
   }
 
-  // #enddocregion start-chunked-conversion
   // #docregion basic-encoder
 }
-// #enddocregion caesar-encoder
+// #enddocregion caesar-encoder, chunked-encoder
 
 // #docregion shift-code-unit
 /// Shifts the specified [codeUnit] by
@@ -81,7 +83,9 @@ int _shiftCodeUnit(int codeUnit, int shift) {
 // #docregion basic-decoder
 /// Decodes a Caesar-cipher-encoded string by
 /// shifting each letter backward in the alphabet.
+// #docregion chunked-decoder
 class CaesarDecoder extends Converter<String, String> {
+  //#enddocregion chunked-decoder
   /// The number of positions to shift each letter backward.
   final int shift;
 
@@ -99,8 +103,12 @@ class CaesarDecoder extends Converter<String, String> {
   }
   // #enddocregion basic-decoder
 
+  // #docregion chunked-decoder
+
   @override
   StringConversionSink startChunkedConversion(Sink<String> sink) {
+    // Wrap the output sink if it isn't already
+    // a StringConversionSink.
     // ignore: close_sinks
     final stringSink = sink is StringConversionSink
         ? sink
@@ -110,7 +118,7 @@ class CaesarDecoder extends Converter<String, String> {
 
   // #docregion basic-decoder
 }
-// #enddocregion basic-decoder
+// #enddocregion basic-decoder, chunked-decoder
 
 // #docregion caesar-codec
 /// A codec that encodes and decodes strings using a
@@ -119,17 +127,17 @@ class CaesarCodec extends Codec<String, String> {
   /// The number of positions to shift each letter.
   final int shift;
 
-  @override
-  CaesarEncoder get encoder => CaesarEncoder(shift);
-
-  @override
-  CaesarDecoder get decoder => CaesarDecoder(shift);
-
   /// Creates a Caesar cipher codec with the given [shift].
   const CaesarCodec(this.shift);
 
   /// Creates a [CaesarCodec] that uses ROT13 encoding.
   const CaesarCodec.rot13() : shift = 13;
+
+  @override
+  CaesarEncoder get encoder => CaesarEncoder(shift);
+
+  @override
+  CaesarDecoder get decoder => CaesarDecoder(shift);
 }
 // #enddocregion caesar-codec
 
@@ -143,17 +151,17 @@ void useEncoderExample() {
 
 // #docregion use-codec
 void useCodecExample() {
-  const caesar = CaesarCodec(3);
+  const cipher = CaesarCodec(3);
 
-  final encoded = caesar.encode('hello');
+  final encoded = cipher.encode('hello');
   print(encoded); // khoor
 
-  final decoded = caesar.decode(encoded);
+  final decoded = cipher.decode(encoded);
   print(decoded); // hello
 
   // The `inverted` getter returns a new codec that
   // applies converts in the inverse direction of the codec.
-  final inverted = caesar.inverted;
+  final inverted = cipher.inverted;
   print(inverted.encode('khoor')); // hello
 }
 // #enddocregion use-codec
--- a/examples/libraries/lib/convert/build_custom_codecs.dart
+++ b/examples/libraries/lib/convert/build_custom_codecs.dart
@@ -160,7 +160,7 @@ void useCodecExample() {
   print(decoded); // hello
 
   // The `inverted` getter returns a new codec that
-  // applies converts in the inverse direction of the codec.
+  // converts in the inverse direction of the original codec.
   final inverted = cipher.inverted;
   print(inverted.encode('khoor')); // hello
 }
--- a/examples/libraries/lib/convert/build_custom_codecs_complete.dart
+++ b/examples/libraries/lib/convert/build_custom_codecs_complete.dart
@@ -1,19 +1,4 @@
-// #docregion complete
 import 'dart:convert';
-
-/// Shifts the specified [codeUnit] by
-/// [shift] positions in the alphabet.
-///
-/// Only shifts lowercase ASCII letters (a-z).
-/// All other characters are returned unchanged.
-int _shiftCodeUnit(int codeUnit, int shift) {
-  const a = 0x61;
-  const z = 0x7A;
-  if (codeUnit >= a && codeUnit <= z) {
-    return a + (codeUnit - a + shift) % 26;
-  }
-  return codeUnit;
-}
 
 /// A [StringConversionSink] that applies a Caesar cipher shift
 /// and forwards the result to [_output].
@@ -58,6 +43,8 @@ class CaesarEncoder extends Converter<String, String> {
 
   @override
   StringConversionSink startChunkedConversion(Sink<String> sink) {
+    // Wrap the output sink if it isn't already
+    // a StringConversionSink.
     // ignore: close_sinks
     final stringSink = sink is StringConversionSink
         ? sink
@@ -86,6 +73,8 @@ class CaesarDecoder extends Converter<String, String> {
 
   @override
   StringConversionSink startChunkedConversion(Sink<String> sink) {
+    // Wrap the output sink if it isn't already
+    // a StringConversionSink.
     // ignore: close_sinks
     final stringSink = sink is StringConversionSink
         ? sink
@@ -100,17 +89,17 @@ class CaesarCodec extends Codec<String, String> {
   /// The number of positions to shift each letter.
   final int shift;
 
-  @override
-  CaesarEncoder get encoder => CaesarEncoder(shift);
-
-  @override
-  CaesarDecoder get decoder => CaesarDecoder(shift);
-
   /// Creates a Caesar cipher codec with the given [shift].
   const CaesarCodec(this.shift);
 
   /// Creates a [CaesarCodec] that uses ROT13 encoding.
   const CaesarCodec.rot13() : shift = 13;
+
+  @override
+  CaesarEncoder get encoder => CaesarEncoder(shift);
+
+  @override
+  CaesarDecoder get decoder => CaesarDecoder(shift);
 }
 
 void main() {
@@ -123,4 +112,16 @@ void main() {
   print(decoded); // the quick brown fox
 }
 
-// #enddocregion complete
+/// Shifts the specified [codeUnit] by
+/// [shift] positions in the alphabet.
+///
+/// Only shifts lowercase ASCII letters (a-z).
+/// All other characters are returned unchanged.
+int _shiftCodeUnit(int codeUnit, int shift) {
+  const a = 0x61;
+  const z = 0x7A;
+  if (codeUnit >= a && codeUnit <= z) {
+    return a + (codeUnit - a + shift) % 26;
+  }
+  return codeUnit;
+}
--- a/examples/libraries/lib/convert/build_custom_codecs_complete.dart
+++ b/examples/libraries/lib/convert/build_custom_codecs_complete.dart
@@ -1,0 +1,126 @@
+// #docregion complete
+import 'dart:convert';
+
+/// Shifts the specified [codeUnit] by
+/// [shift] positions in the alphabet.
+///
+/// Only shifts lowercase ASCII letters (a-z).
+/// All other characters are returned unchanged.
+int _shiftCodeUnit(int codeUnit, int shift) {
+  const a = 0x61;
+  const z = 0x7A;
+  if (codeUnit >= a && codeUnit <= z) {
+    return a + (codeUnit - a + shift) % 26;
+  }
+  return codeUnit;
+}
+
+/// A [StringConversionSink] that applies a Caesar cipher shift
+/// and forwards the result to [_output].
+class _CaesarEncoderSink extends StringConversionSinkBase {
+  final int _shift;
+  final StringConversionSink _output;
+
+  _CaesarEncoderSink(this._shift, this._output);
+
+  @override
+  void addSlice(String chunk, int start, int end, bool isLast) {
+    final buffer = StringBuffer();
+    for (var i = start; i < end; i++) {
+      buffer.writeCharCode(_shiftCodeUnit(chunk.codeUnitAt(i), _shift));
+    }
+    _output.add(buffer.toString());
+    if (isLast) {
+      _output.close();
+    }
+  }
+
+  @override
+  void close() => _output.close();
+}
+
+/// Encodes a string by shifting each letter forward in the alphabet.
+class CaesarEncoder extends Converter<String, String> {
+  /// The number of positions to shift each letter forward.
+  final int shift;
+
+  /// Creates an encoder that shifts each letter by [shift] positions.
+  const CaesarEncoder(this.shift);
+
+  @override
+  String convert(String input) {
+    final buffer = StringBuffer();
+    for (final codeUnit in input.codeUnits) {
+      buffer.writeCharCode(_shiftCodeUnit(codeUnit, shift));
+    }
+    return buffer.toString();
+  }
+
+  @override
+  StringConversionSink startChunkedConversion(Sink<String> sink) {
+    // ignore: close_sinks
+    final stringSink = sink is StringConversionSink
+        ? sink
+        : StringConversionSink.from(sink);
+    return _CaesarEncoderSink(shift, stringSink);
+  }
+}
+
+/// Decodes a Caesar-cipher-encoded string
+/// by shifting each letter backward in the alphabet.
+class CaesarDecoder extends Converter<String, String> {
+  /// The number of positions to shift each letter backward.
+  final int shift;
+
+  /// Creates a decoder that reverses a Caesar cipher with [shift] positions.
+  const CaesarDecoder(this.shift);
+
+  @override
+  String convert(String input) {
+    final buffer = StringBuffer();
+    for (final codeUnit in input.codeUnits) {
+      buffer.writeCharCode(_shiftCodeUnit(codeUnit, 26 - shift));
+    }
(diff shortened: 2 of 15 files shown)


parlough:
@lrhn would you be able to take a look at these two pages covering converters and codecs? I'd appreciate any insights you have!

@conooi When you have a chance, this PR is ready for an initial look.

@lamek This PR includes a variety of code excerpts if you'd like to take a look.

Thanks all! I appreciate it :)

conooi (approved):
lgtm! Thanks, Parker!!

examples/libraries/lib/convert/build_custom_codecs.dart
```diff
@@ -0,0 +1,172 @@
...
+
+  // #docregion basic-encoder
+}
+// #enddocregion caesar-encoder, chunked-encoder
+
+// #docregion shift-code-unit
+/// Shifts the specified [codeUnit] by
+/// [shift] positions in the alphabet.
+///
+/// Only shifts lowercase ASCII letters (a-z).
```
lrhn, line 71: (Can farily easily do upper case too, just create
```dart
  const a = 0x61;
  const z = 0x7a;
  const caseDelta = 0x20;
  final lowerCodeUnit = codeUnit | caseDelta;
  if (lowerCodeUnit >= a && lowerCodeUnit <= z) {
    final letterNumber = lowerCodeUnit - a;
    return codeUnit - letterNumber + (letterNumber + shift).remainder(26);
  }
```
Easier than having to explain why it doesn't do it.)

examples/libraries/lib/convert/build_custom_codecs.dart
```diff
@@ -0,0 +1,172 @@
...
+  }
+  return codeUnit;
+}
+// #enddocregion basic-encoder, shift-code-unit
+
+// #docregion basic-decoder
+/// Decodes a Caesar-cipher-encoded string by
+/// shifting each letter backward in the alphabet.
+// #docregion chunked-decoder
+class CaesarDecoder extends Converter<String, String> {
```
lrhn: It's repeated code.
If someone understands the Ceasar cipher, they'll know that encoding and decoding is the same operation with a different key.
If they don't, the exact example won't mean much.

Could have 
```dart
final class CaesarConverter extends Converter<String, String> { 
  // ...
  const CaesarConverter.encoder(int shift) 
      : assert(0 <= shift && shift < 26, throw RangeError.range(shift, 0, 25, 'shift')),
        _shift = shift;
  const CaesarConverter.decoder(int shift) 
      : assert(0 <= shift && shift < 26, throw RangeError.range(shift, 0, 25, 'shift')),
        _shift = (26 - shift) % 26;
  // ...
 }`
```

examples/libraries/lib/convert/build_custom_codecs.dart
```diff
@@ -0,0 +1,172 @@
...
+  final int _shift;
+  final StringConversionSink _output;
+
+  _CaesarEncoderSink(this._shift, this._output);
+
+  @override
+  void addSlice(String chunk, int start, int end, bool isLast) {
+    final buffer = StringBuffer();
+    for (var i = start; i < end; i++) {
+      buffer.writeCharCode(_shiftCodeUnit(chunk.codeUnitAt(i), _shift));
```
lrhn, line 16: (I'd have a `String _shiftCodeUnits(String input, int shift) {...}` helper. You write this code repeatedly. Well, at least twice.)

examples/libraries/lib/convert/build_custom_codecs.dart
```diff
@@ -0,0 +1,172 @@
...
+/// Shifts the specified [codeUnit] by
+/// [shift] positions in the alphabet.
+///
+/// Only shifts lowercase ASCII letters (a-z).
+/// All other characters are returned unchanged.
+int _shiftCodeUnit(int codeUnit, int shift) {
+  const a = 0x61;
+  const z = 0x7A;
+  if (codeUnit >= a && codeUnit <= z) {
+    return a + (codeUnit - a + shift) % 26;
```
lrhn, line 77: Use `.remainder(26)` instead of `% 26`.

Using `.remainder(26)` guarantees a result in the range 0..25.

There is no guard on the `shift` value, so it can potentially be negative, which can make `(codeUnit - a + shift)` negative and then `%26` gives a negative result.

examples/libraries/lib/convert/build_custom_codecs.dart
```diff
@@ -0,0 +1,172 @@
+import 'dart:convert';
+
+// #docregion caesar-encoder-sink
```
lrhn, line 3: (For readability of the file itself, I'd put the `CaesarCodec` first, then `CaesarConverter`, and then the private classes they depend on. But that's just, like, my way of structuring things: Most important first, things it depends on below. If you don't need to know the details of the dependencies, you can just not scroll down, and if you need to find a detail, you only need to look below, not above.

If this file only exists as a file to extract chunks from for the article, then its readability probably isn't important.)

examples/libraries/lib/convert/build_custom_codecs.dart
```diff
@@ -0,0 +1,172 @@
...
+/// A [StringConversionSink] that applies a Caesar cipher [_shift] and
+/// forwards the result to the [_output] sink.
```
lrhn, line 5: Single-line first paragraph.
Consider:
```dart
/// A [StringConversionSink] that applies a Casar cipher.
///
/// Shifts inputs by [_shift] and emits the shifted strings on
/// [_output].
```

examples/libraries/lib/convert/build_custom_codecs.dart
```diff
@@ -0,0 +1,172 @@
...
+    }
+  }
+
+  @override
+  void close() => _output.close();
+}
+// #enddocregion caesar-encoder-sink
+
+// #docregion basic-encoder, caesar-encoder
+/// Encodes a string by shifting each letter forward in the alphabet.
```
lrhn, line 30: Noun phrases for classes.
Consider (making it one `CaesarConverter` class instead of both encoder and decoder):
```dart
/// Casear cipher converter that shifts letters forward by [shift].
///
/// Converts strings to strings by shifting every ASCII letter
/// forwards in the alphabet by [shift] positions, wrapping over
/// from `z` to `a` and `Z` to `A` if shifting past the end of the
/// alphabet.
```

examples/libraries/lib/convert/build_custom_codecs.dart
```diff
@@ -0,0 +1,172 @@
...
+  }
+  // #enddocregion basic-encoder
+
+  // #docregion chunked-encoder
+
+  @override
+  StringConversionSink startChunkedConversion(Sink<String> sink) {
+    // Wrap the output sink if it isn't already
+    // a StringConversionSink.
+    // ignore: close_sinks
```
lrhn, line 56: This ignore-comment is visible to readers of the article.
(`build-custom-codecs.md` line 276, and one in 298 too.)

I'd not put that ignore here. It's not something a reader of the example needs to see.
If the lint was in the recommended set, then everybody should be assumed to have to adhere to it, but it isn't. 
Example code should not assume something that we don't assume for all users, that's just confusing.
(If the lint needs to be ignored here, then it's not a good enough lint. A sink that is returned is clearly for later use, and should not need to be closed here, but at the call-site.)

If the ignore is needed in this repo, make it an `// ignore_for_file:`, so the example code included in the article doesn't contain it.

examples/libraries/lib/convert/build_custom_codecs.dart
```diff
@@ -0,0 +1,172 @@
...
+    return _CaesarEncoderSink(shift, stringSink);
+  }
+
+  // #docregion basic-encoder
+}
+// #enddocregion caesar-encoder, chunked-encoder
+
+// #docregion shift-code-unit
+/// Shifts the specified [codeUnit] by
+/// [shift] positions in the alphabet.
```
lrhn, line 69: Single line first paragraph.

Looks like it should fit on one line.

examples/libraries/lib/convert/build_custom_codecs.dart
```diff
@@ -0,0 +1,172 @@
...
+  const z = 0x7A;
+  if (codeUnit >= a && codeUnit <= z) {
+    return a + (codeUnit - a + shift) % 26;
+  }
+  return codeUnit;
+}
+// #enddocregion basic-encoder, shift-code-unit
+
+// #docregion basic-decoder
+/// Decodes a Caesar-cipher-encoded string by
```
lrhn, line 84: (If keeping separate decoder, all the same comments apply as to the encoder.)

examples/libraries/lib/convert/build_custom_codecs.dart
```diff
@@ -0,0 +1,172 @@
...
+
+  @override
+  void close() => _output.close();
+}
+// #enddocregion caesar-encoder-sink
+
+// #docregion basic-encoder, caesar-encoder
+/// Encodes a string by shifting each letter forward in the alphabet.
+// #docregion chunked-encoder
+class CaesarEncoder extends Converter<String, String> {
```
lrhn, line 32: This class is basically repeated as `CaesarDecoder`, with a single `26 - shift` as the only change.

If someone understands the Ceasar cipher, they'll know that encoding and decoding is the same operation with a different key.
If they don't, the exact example won't mean much.

Could have 
```dart
final class CaesarConverter extends Converter<String, String> { 
  // ...
  const CaesarConverter(int shift) 
      : assert(0 <= shift && shift < 26, '$shift should be in the range 0..25'),
        _shift = shift;
  const CaesarConverter.inverted(int shift) 
      : assert(0 <= shift && shift < 26, '$shift should be in the range 0..25'),
        _shift = (26 - shift) % 26;
  // ...
 }`
```
(Uses `%26` instead of `.remainder(26)` because the latter sadly isn't potentially constant. If the assert runs, the value will be in the 0..25 range, which ensures we won't have more than 26 constant instances. If assert doesn't run, and the converter is created from a runtime value, the stored shift can be negative, but using `.remainder(26)` in the actual conversion corrects for that.)

examples/libraries/lib/convert/build_custom_codecs.dart
```diff
@@ -0,0 +1,172 @@
...
+  // #docregion chunked-encoder
+
+  @override
+  StringConversionSink startChunkedConversion(Sink<String> sink) {
+    // Wrap the output sink if it isn't already
+    // a StringConversionSink.
+    // ignore: close_sinks
+    final stringSink = sink is StringConversionSink
+        ? sink
+        : StringConversionSink.from(sink);
```
lrhn, line 59: (Could, and IMO should, be:
```dart
if (sink is! StringConversionSink) {
  sink = StringConversionSink.from(sink);
}
```
The code here is probably copying existing code that predates Dart 2's flow-based promotion. We can do better today.)

examples/libraries/lib/convert/build_custom_codecs.dart
```diff
@@ -0,0 +1,172 @@
...
+/// [Caesar cipher](https://wikipedia.org/wiki/Caesar_cipher).
+class CaesarCodec extends Codec<String, String> {
+  /// The number of positions to shift each letter.
+  final int shift;
+
+  /// Creates a Caesar cipher codec with the given [shift].
+  const CaesarCodec(this.shift);
+
+  /// Creates a [CaesarCodec] that uses ROT13 encoding.
+  const CaesarCodec.rot13() : shift = 13;
```
lrhn, line 134: (Consider making this a forwarding constructor:
```dart
  const CaesarCodec.rot13() : this(13);
```
I personally prefer when a class has only one initializing constructor, that means I can check that to see how fields are initialized and any restrictions that apply to them.

And, more importantly than my personal preferences, when Dart gets primary constructors, this class can be re-written as:
```darrt
class CaesarCodec(final int shift) extends Codec<String, String> {
  //
  const CaesarCodec.rot13() : this(13);
  //
}
```

without needing any change to the redirecting constructor.
Be prepared for primary constructors. 😉 )

examples/libraries/lib/convert/build_custom_codecs.dart
```diff
@@ -0,0 +1,172 @@
...
+
+  // #docregion basic-decoder
+}
+// #enddocregion basic-decoder, chunked-decoder
+
+// #docregion caesar-codec
+/// A codec that encodes and decodes strings using a
+/// [Caesar cipher](https://wikipedia.org/wiki/Caesar_cipher).
+class CaesarCodec extends Codec<String, String> {
+  /// The number of positions to shift each letter.
```
lrhn, line 127: (Can continue:
```dart
  ///
  /// Is shifted forwards for encoding and backwards for
  /// decoding.
```

examples/libraries/lib/convert/build_custom_codecs.dart
```diff
@@ -0,0 +1,172 @@
...
+    return buffer.toString();
+  }
+  // #enddocregion basic-encoder
+
+  // #docregion chunked-encoder
+
+  @override
+  StringConversionSink startChunkedConversion(Sink<String> sink) {
+    // Wrap the output sink if it isn't already
+    // a StringConversionSink.
```
lrhn, line 55: Can this fit on one line? I find it abrupt to read the line-break.
Consider
```dart 
    /// Make sink a StringConversionSink, if it isn't already one.
```

examples/libraries/lib/convert/build_custom_codecs_complete.dart
```diff
@@ -0,0 +1,127 @@
...
+      buffer.writeCharCode(_shiftCodeUnit(codeUnit, shift));
+    }
+    return buffer.toString();
+  }
+
+  @override
+  StringConversionSink startChunkedConversion(Sink<String> sink) {
+    // Wrap the output sink if it isn't already
+    // a StringConversionSink.
+    // ignore: close_sinks
```
lrhn, line 48: (I'd drop the ignore completely in this file. It's still not relevant to a reader. So maybe just disable the lint entirely for the repo. No big loss.)

examples/libraries/lib/convert/converters_and_codecs.dart
```diff
@@ -0,0 +1,76 @@
...
+// #enddocregion encode-decode, encoder-decoder, custom-config, fuse-codec, fuse-converter, encoding-lookup
+
+// #docregion encode-decode
+void encodeDecodeExample() {
+  // Encode a Dart object to a JSON string.
+  final jsonString = json.encode({'name': 'Dash', 'age': 5});
+  print(jsonString); // {"name":"Dash","age":5}
+
+  // Decode a JSON string back to a Dart object.
+  final data = json.decode(jsonString) as Map<String, Object?>;
```
lrhn, line 13: (Thanks for using `Object?` instead of `dynamic`!! 😄)

examples/libraries/lib/convert/converters_and_codecs_io.dart
```diff
@@ -0,0 +1,66 @@
+// #docregion stream-lines, stream-json, stream-gzip, decode-stream, encoding-param
+import 'dart:convert';
+import 'dart:io';
+
+// #enddocregion stream-lines, stream-json, stream-gzip, decode-stream, encoding-param
+
+// #docregion stream-lines
+void streamLinesExample() async {
+  // Read a file as a stream of bytes, decode UTF-8 into a string,
+  // then split the single string into lines.
```
lrhn, line 10: Consider adding:
```dart
  // (See also `File.readAsLines`.)
```

Don't want someone to come away from this example and thinking this is how they *should* read lines from a file. It's more of an example of how `readAsLines` might do it internally, in case they ever need to do something _similar_.

examples/libraries/lib/convert/converters_and_codecs_io.dart
```diff
@@ -0,0 +1,66 @@
...
+
+// #docregion stream-json
+void streamJsonExample() async {
+  // Read a JSON file and decode it in a single streaming pipeline.
+  final stream = File('data.json')
+      .openRead() //
+      .transform(utf8.decoder)
+      .transform(json.decoder);
+
+  await for (final value in stream) {
```
lrhn, line 30: The `json.decoder` always emits a single value.

I'd do:
```dart
  final value = await stream.first; // or even .single
  print(value);
```

examples/libraries/test/convert/build_custom_codecs_test_example.dart
```diff
@@ -0,0 +1,65 @@
...
+void main() {
+  const codec = CaesarCodec(3);
+
+  group('CaesarCodec', () {
+    test('encode shifts letters forward', () {
+      expect(codec.encode('abc'), equals('def'));
+      expect(codec.encode('xyz'), equals('abc'));
+    });
+
+    test('decode reverses encode', () {
```
lrhn, line 17: Another test of that could be:
```dart
  var roundtrip = codec.fuse(codec.inverted);
  expect(roundtrip.encode(original), equals(original));

  var convertRoundtrip = codec.encoder.fuse(codec.decoder);
  expect(convertRoundtrip.convert(original), equals(original));
```
Good logic to test for all converters, that their encoder and decoder combines to the indentity. (When it does, some may throw away redundant information or correct errors.)

Which, as a complete tangent, also means that you _could_ choose to override `fuse`:
```dart
  Codec<String, R> fuse<R>(Codec<String, R> next) =>
    if (next case CaesarCodec next) {
      final shift = (this.shift + next.shift).remainder(26);
      return CaesarCodec(shift) as Codec<String, R>;
    }
    return super.fuse(next);
  }
```
and similar for the converter
```dart
  Converter<String, R> fuse<R>(Converter<String, R> next) =>
    if (next case CaesarConverter next) {
      final shift = (this.shift + next.shift).remainder(26);
      return CaesarConverter._(shift) as Converter<String, R>;
    }
    return super.fuse(next);
  }
```
I don't know if that's a good *example* to have. If it is, it may be worth adding.)

src/content/libraries/convert/build-custom-codecs.md
```diff
@@ -0,0 +1,743 @@
...
+
+[Caesar cipher]: https://wikipedia.org/wiki/Caesar_cipher
+
+[Use converters and codecs]: /libraries/convert/converters-and-codecs
+
+## When to build a custom codec
+
+Build a custom `Codec` or `Converter` when your conversion:
+
+- Is **bidirectional** and you need both encoding and decoding.
```
lrhn, line 40: That's only `Codec`. A `Converter` can be only one-way.

src/content/libraries/convert/build-custom-codecs.md
```diff
@@ -0,0 +1,743 @@
...
+[`Codec`]: {{site.dart-api}}/dart-convert/Codec-class.html
+[`fuse`]: {{site.dart-api}}/dart-convert/Codec/fuse.html
+
+[Caesar cipher]: https://wikipedia.org/wiki/Caesar_cipher
+
+[Use converters and codecs]: /libraries/convert/converters-and-codecs
+
+## When to build a custom codec
+
+Build a custom `Codec` or `Converter` when your conversion:
```
lrhn, line 38: (Should the list be read as "and" or "or"?)

src/content/libraries/convert/build-custom-codecs.md
```diff
@@ -0,0 +1,743 @@
...
+
+- Is **bidirectional** and you need both encoding and decoding.
+- Benefits from **streaming** because the data might be
+  too large for a single in-memory transformation.
+- Should **compose** with other codecs through `fuse`.
+- Is meant to be **reused** across projects or published as a package.
+
+For one-directional or internal conversions,
+a purpose-built function is often simpler.
```
lrhn, line 47: The directionality isn't really important, more the streaming aspect (works on long seqeuences of something, and can do so in a chunk-based way).
You can have `Converter`s that have no inverse, and they'll work fine when fused with other converters and applied to streams.

What would an example of a non-`Convert`-based conversion be?
(Pondering ...)

I think it's often objects converting/encoding/serializing themselves, and then a separate `parse` function to parse the serialized form back. Nobody would create an `IntCodec` that encodes a single integer to a numeral, and decodes it back to an integer. It doesn't _scale_, which is what `Converter` and `Codec` is about.

Or as you say below: `toJson` and `fromJson`.

I think that means *compositional* conversion, not *sequential*. If one long sequence of homogenic units is converted to another long sequence of units, sequentially, then `Converter` is good. If your input is a tree-structure, then you want recursion, and `Converter` isn't great for that.

Not at all sure how to say that, but I'd drop the focus in _directionality_ and focus on the _structure_ of what is being converted to or from. If neither is a linear sequence, streaming conversion doesn't matter, and then there is no real benefit from a `Converter` (or two converter as a `Codec`).

src/content/libraries/convert/build-custom-codecs.md
```diff
@@ -0,0 +1,743 @@
...
+For application-level object serialization,
+the `fromJson` and `toJson` convention with code generation is more idiomatic.
```
lrhn, line 49: (Then you combine the JSON-structured objects and use the JsonCodec to encode the final result.

But that might be because `JsonCodec` is actually a _bad_ example of a codec. It can only encode a single object, it's not streaming, and it can only decode into a single object. It doesn't work use streaming encoding at all. ... Maybe it should, as JSON-lines, but that's a different issue.)

src/content/libraries/convert/build-custom-codecs.md
```diff
@@ -0,0 +1,743 @@
...
+
+[`convert`]: {{site.dart-api}}/dart-convert/Converter/convert.html
+
+## Build the decoder
+
+The inverse of a Caesar encoder with a shift of `N`
+is a Caesar encoder with a shift of `26 - N`.
+You could reuse `CaesarEncoder` directly,
+but creating a separate class keeps the intent clear and
+lets you customize behavior if needed:
```
lrhn, line 118: (OK, so this duplication was intentional.
Then Caesar-ciphers might not be the best example, because there is no real behavior to customize. And nitpickers like me will fuzz over that instead of reading what you write 🫤 .

I also liked to have an example that converters can be reused, ... but this doesn't have to be it.)

src/content/libraries/convert/build-custom-codecs.md
```diff
@@ -0,0 +1,743 @@
...
+  final output = File('message.gz').openWrite();
+  output.add(encryptAndCompress.encode('Secret message.'));
+  await output.close();
+
+  // Read and decrypt.
+  final bytes = await File('message.gz').readAsBytes();
+  final decrypted = encryptAndCompress.decode(bytes);
+  print(decrypted); // Secret message.
+}
+```
```
lrhn, line 381: Is it worth making the comparison between:
```dart
  source.transform(converter1).transform(converter2)
```
and 
```dart
  source.transform(converter1.fuse(converter2))
```
?

They should output the same thing, but the latter _may_ be more efficient in a few cases, where `converter1` knows `converter2` and can provide a more efficient combined transformation without needing an intermediate representation.

src/content/libraries/convert/build-custom-codecs.md
```diff
@@ -0,0 +1,743 @@
...
+      expect(codec.encode('hello, world!'), equals('khoor, zruog!'));
+    });
+
+    test('empty string encodes to empty string', () {
+      expect(codec.encode(''), equals(''));
+    });
+
+    test('inverted codec swaps encode and decode', () {
+      final inverted = codec.inverted;
+      expect(inverted.encode('def'), equals('abc'));
```
lrhn, line 429: Could be:
```dart
  expect(inverted.encode('def'), equals(codec.decode('def')));
  expect(inverted.decode('def'), equals(codec.encode('def')));
```
to be extra explicit.

src/content/libraries/convert/build-custom-codecs.md
```diff
@@ -0,0 +1,743 @@
...
+      expect(inverted.decode('abc'), equals('def'));
+    });
```
lrhn, line 431: As suggested above, maybe also check that `codec.fuse(codec, code.inverted)`, `codec.fuse(code.inverted, codec)`, `codec.encoder.fuse(codec.decoder)` and `codec.decoder.fuse(codec.encoder)` are all identity-transformations, at least when they should be.

Or maybe I'm overcomplicating things. I probably am. I am.
The perfect is the enemy of the good. Most people won't care about `fuse`. Sorry :)

src/content/libraries/convert/build-custom-codecs.md
```diff
@@ -0,0 +1,743 @@
...
+When designing your own codecs and converters,
+keep the following best practices in mind.
+
+### Make codecs const-constructable when possible
+
+If your codec's configuration is able to be immutable,
+specify its constructors as `const`.
+This lets callers create compile-time constant instances
+and follows the pattern set by the built-in codecs.
+
```
lrhn, line 611: A little strong on the `const` recommendation. There are tradeoffs.
Consider:

----

Being a constant allows being used as default value, which is particulary useful for `Encoding`s like `utf8`.

Constant constructors are very limited in how they can validate their arguments. Sometimes being able to validate constructor inputs is more important than being constant.

A codec like `Utf8Codec` has only very little state, a  `allowMalformed` boolean, and both values are valid. It has no need to validate its inputs, and can easily have a constant constructor.

The `CaesarCodec` stores an integer. It could make sense to check that the integer is in the range 0..25. The codec is written to work with any value, working the same modulo 26, but values outside of that range is _probably_ a user error.
It's not strictly necessary to validate values, but if you don't, the code needs to be written to be safe for any integer input.

A more complicated codec, for example a general 256-character [code page][Code Page] encoding configured with a 256-character conversion table, may want (or need) to check
that its conversion table does not contain duplicates.
That's is outside the capabilities of a constant constructor,
and such a constructor would have to be non-constant.

Always do what gives the user the best and safest user experience.

[Code Page]: https://en.wikipedia.org/wiki/Code_page

src/content/libraries/convert/build-custom-codecs.md
```diff
@@ -0,0 +1,743 @@
...
+  // Constructor sets defaults.
+  const JsonCodec({this.reviver, this.toEncodable});
+
+  // Method overrides accept per-call options.
+
+  @override
+  dynamic decode(String source, {
+    Object? Function(Object?, Object?)? reviver,
+  }) { /* ... */ }
```
lrhn, line 633: You can even start the body with
```dart
}) {
  reviver ??= this.reveiver;
  /* ... */
}
```
to suggest how it's used.

src/content/libraries/convert/build-custom-codecs.md
```diff
@@ -0,0 +1,743 @@
...
+
+[`JsonCodec`]: {{site.dart-api}}/dart-convert/JsonCodec-class.html
+
+### Override fuse for optimized composition
+
+If your codec is commonly composed with another specific codec,
+override the [`fuse`][] method to return an optimized implementation.
+The built-in [`JsonCodec`][] does this: when fused with [`Utf8Codec`][],
+it returns a codec that uses [`JsonUtf8Encoder`][] to
+bypass the intermediate string representation.
```
lrhn, line 652: (And `CaesarCodec` would fuse efficiently with another `CaesarCodec` by adding their shifts, so you could even have an example here.)

src/content/libraries/convert/build-custom-codecs.md
```diff
@@ -0,0 +1,743 @@
...
+[`Utf8Codec`]: {{site.dart-api}}/dart-convert/Utf8Codec-class.html
+[`JsonUtf8Encoder`]: {{site.dart-api}}/dart-convert/JsonUtf8Encoder-class.html
+
+### Choose the right sink base class
+
+Dart provides specialized sink base classes that can
+improve performance by avoiding unnecessary conversions:
+
+[`StringConversionSink`][]
+: For converters that input or output strings.
```
lrhn, line 665: A little vague on where you use them.

Sinks are used for chunked conversion.
The converter itself only controls what it returns
from `startChunkedConversion`, so here I'd probably
only say that `StringConversionSink` is for converters
that _input_ strings.

A converter that outputs strings can perform more efficiently if the argument to `startChunkedConversion` is a  `StringConversionSink`, but whether it is that, is 
determined by client of the conversion, not the converter.

(Converting an incoming `Sink<String>` to a `StringConversionSink` is just an implementation detail for the converter's own convenience, so the internal logic only has to care about one case. It's not actually more efficient.
Which also means that `CaesarEncoder` doesn't actually _need_ to do that wrapping. It only uses `add` and `close` on the sink anyway.)

src/content/libraries/convert/build-custom-codecs.md
```diff
@@ -0,0 +1,743 @@
...
+[`ByteConversionSink`]: {{site.dart-api}}/dart-convert/ByteConversionSink-class.html
+[`ChunkedConversionSink`]: {{site.dart-api}}/dart-convert/ChunkedConversionSink-class.html
+
+### Respect data ownership in sinks
+
+Data passed to a sink's `add` method shouldn't be
+modified by the caller afterward.
+Sinks are allowed to hold references to the data
+rather than copying it immediately.
+Since they're immutable, this is naturally safe for strings,
```
lrhn, line 695: -> Since strings are immutable, this is naturally safe for them,
but ...

(When I read "Since they're ..." the most recent plural noun in context was "Sinks", so I jumped to thinking "immutable sinks?" before reading any further.)

src/content/libraries/convert/index.md
```diff
@@ -0,0 +1,12 @@
+---
+title: Dart's support for data conversion
+shortTitle: dart:convert library
+breadcrumb: Convert
+description: >-
+  Learn about Dart's library support for
+  converting between and transforming different data representations
```
lrhn, line 7: "converting between and transforming ...."

I stumbled a little reading that.
I think the "between and" combination was the stumbler.

Consider:

"transforming and converting between ..."

Or

"converting between, and transforming, ..."

src/content/libraries/convert/converters-and-codecs.md
```diff
@@ -1,615 +1,401 @@
...
-    _outSink.add(_converter.convert(data));
-  }
+## Transform streams with converters
 
-  void close() {
-    _outSink.close();
-  }
-}
-```
+Because every `Converter` implements [`StreamTransformer`][],
```
lrhn, line 236: (if they implement `startChunkedConversion`, which all platform converters do)

src/content/libraries/convert/converters-and-codecs.md
```diff
@@ -1,615 +1,401 @@
...
-  int key = int.parse(args[2]);
-  new File(inFile)
-    .openRead()
-    .transform(new RotConverter(key))
-    .pipe(new File(outFile).openWrite());
+void main() async {
+  // Read a file as a stream of bytes, decode UTF-8 into a string,
+  // then split the single string into lines.
+  final lines = File('data.txt')
+      .openRead() //
```
lrhn, line 252: Why `//` here? Trying to coerce the formatter?

examples/libraries/lib/convert/converters_and_codecs_io.dart
lrhn, line 10: Not really a single string. That would be the description of:
```dart
    final bytes = File('data.txt').readAsBytesSync();
    final string = utf8.decode(bytes);
    final lines = const LineSplitter().convert(string);
```

The streaming version specifically doesn't create a single string.

examples/libraries/lib/convert/converters_and_codecs.dart
```diff
@@ -0,0 +1,76 @@
...
+  // Chain two converters: JSON encode, then UTF-8 encode.
+  final encoder = json.encoder.fuse(utf8.encoder);
+  final bytes = encoder.convert({'key': 'value'});
+  print(bytes); // [123, 34, 107, ...]
+}
+// #enddocregion fuse-converter
+
+// #docregion encoding-lookup
+void encodingLookupExample() {
+  final encoding = Encoding.getByName('utf-8');
```
lrhn, line 72: I'd put a `!` after the lookup instead of a `?.` in the use below.

src/content/libraries/convert/converters-and-codecs.md
```diff
@@ -1,615 +1,401 @@
...
-```
+For other character sets such as UTF-16, GBK, or Shift-JIS,
+check out third-party packages like [`package:charset`][] that
+provide many codecs implementing the `Encoding` interface.
 
-_CipherSinkAdapter is simply:
+`Encoding` also provides a default implementation of [`decodeStream`][].
+In contrast to `Stream.transform` which returns a `Stream<String>`,
+the `decodeStream` method reads a byte stream and
+returns a single decoded `String` (wrapped in a `Future`).
```
lrhn, line 338: returns ... (wrapped in a `Future`) -> asynchronously returns ...

Or just -> return ...

People know that "returns `Future<String>`" means "returns `String`, but later".

src/content/libraries/convert/converters-and-codecs.md
```diff
@@ -1,615 +1,401 @@
...
+The `Codec` and `Converter` classes serve a different purpose.
+They're designed for data format transformations:
+encoding schemes, compression, character sets, and
+other conversions where streaming support, bidirectional symmetry,
+and composability through `fuse` matter.
+
+If the conversion is about how data is _represented_,
+such as with bytes, strings, or compressed data,
+the codec pattern is a natural fit.
+If the conversion is about how data is _structured_,
```
lrhn, line 392: (Well, that is a very good description of what I tried to say above. Say the same thing there!)

lrhn (approved):
LGTM! Super update.

I tend to want to over-communicate all the details, so take my suggestions with an unhealthy amount of salt.

examples/libraries/lib/convert/build_custom_codecs.dart
parlough: I'm coming back to this a few months later (thanks for the thorough review), and I wanted to confirm this suggestion.

https://api.dart.dev/dart-core/num/operator_modulo.html suggests for the `%` operator, "The sign of the returned value r is always positive". Doesn't that match what we want here? A bit of a contrast compared to many other languages like Java and JS though, so perhaps it's worth me explicitly calling out the behavior of %?

PR #7217, opened 2026-03-18
Dart Blog

schultek:
Adds a space for blog posts at `/blog` and moves over all posts from medium.

**Staged:** https://dart-dev--pr7217-dart-blog-ahwpcg6k.web.app/blog

### DIFF
--- a/site/lib/src/components/common/dash_image.dart
+++ b/site/lib/src/components/common/dash_image.dart
@@ -21,7 +21,8 @@ class DashImage extends StatelessComponent {
   final String? caption;
 
   static DashImage fromAttributes(Map<String, String> attributes) {
-    if (attributes['src'] == null) {
+    final imageSrc = attributes['src'];
+    if (imageSrc == null) {
       throw ArgumentError.value(
         attributes,
         'src',
@@ -29,7 +30,7 @@ class DashImage extends StatelessComponent {
       );
     }
     return DashImage(
-      src: attributes['src']!,
+      src: imageSrc,
       alt: attributes['alt'],
       caption: attributes['caption'],
     );
@@ -39,14 +40,15 @@ class DashImage extends StatelessComponent {
   Component build(BuildContext context) {
     if (alt == null) {
       print(
-        '[WARNING] DashImage is missing an "alt" attribute for ${context.page.url}:$src',
+        '[WARNING] DashImage is missing an "alt" attribute for '
+        '${context.page.url}:$src',
       );
     }
     return figure([
       img(src: context.resolveAsset(src), alt: alt),
-      if (caption != null && caption!.isNotEmpty)
+      if (caption case final caption? when caption.isNotEmpty)
         figcaption([
-          DashMarkdown(content: caption!, inline: true),
+          DashMarkdown(content: caption, inline: true),
         ], classes: 'figure-caption'),
     ]);
   }


parlough:
Love it! :D 

I still plan to review the implementation, but from navigating around the staged site, here's some things I noticed so far. I'm posting as a comment as GitHub doesn't handle the large amount of changes well.

Issues:
- [x] Some extracted code blocks have extra back ticks at the beginning and end.
  https://github.com/dart-lang/site-www/blob/d5e24574612f6b0417bc8b5d555024e08c645db4/src/content/blog/why-nullable-types/index.md?plain=1#L64
- [x] "Copy Link" should be "Copy link" in the share menu.
- [x] When in a narrow layout, the dot separator before the date is unnecessary: [Screenshot of author info](https://github.com/user-attachments/assets/29547d5b-b2e0-4a70-a7dc-3a9ea587fb5c)
- [x] The hover style for the share menu button is quite bright in dark mode: [Screenshot of hovering](https://github.com/user-attachments/assets/6d138669-5417-4bef-ad72-d4b29708233c)
- [x] In narrow layouts, the hamburger menu button is broken. I'm not sure if showing the default sidenav is the best option long term, but mobile users should at least be able to navigate to the other site sections.
- [x] The focus style for the root page's category switcher is partially covered: [Screenshot of focused category](https://github.com/user-attachments/assets/219471d1-bb27-400b-9f85-5b417f6d83c8)
- [x] Consider validating social links, if specified, are not empty. For example, if I click on the author info for Marya ([Dart 3.7 example](https://dart-dev--pr7217-dart-blog-ahwpcg6k.web.app/blog/announcing-dart-3-7)), it just goes to the X homepage.

Suggestions and questions to optionally consider:

- For the breadcrumbs, should we drop the current page title?
- Maybe the page `title` should append `- The Dart Blog` or something similar?
- Should we remove "Dart blog" from the site switcher now that it has its own entry? [Screenshot of entry in site switcher](https://github.com/user-attachments/assets/247db17e-01ba-420b-b788-2fc63da67cbc)
- At the end of the blog, it might be nice to have some action to do next if the reader wants to dive deeper or keep going. Perhaps we can use the previous and next cards based on date? Or at least a quick way to navigate to the blog homepage.

abdallahshaban557:
Some feedback:

- I don't quite get why we have the seperation between "announcement" and "all" and then "other". Does it really matter to have these be separate?
- some of the images are not sized to the space they are occupying - for example like the Announcing Dart 3 blog entry in the list of blogs.
- should images have some border to seperate them from the rest of the content? check out this page for example which has the pub.dev screenshot https://dart-dev--pr7217-dart-blog-ahwpcg6k.web.app/blog/preparing-the-dart-and-flutter-ecosystem-for-null-safety
- The font style for the share button dropdown menu feels pretty different than the rest of the site.
- I think for code snippets - the language should probably be at the right left hand side, so that it isn't hidden behind the copy button.

This is already great by the way!

schultek:
> Should we remove "Dart blog" from the site switcher now that it has its own entry?

Since in the last meeting we talked about making a clear visual separation between docs and blog, how about something like this for the header?

Bildschirmfoto 2026-03-23 um 13 12 10

So we keep the site switcher and remove the header links (or maybe keep 1/2 header links to main site and docs).

Also do we want to show the banner? And the search? 

---

> I don't quite get why we have the seperation between "announcement" and "all" and then "other". Does it really matter to have these be separate?

I took this over from medium. I don't have a strong opinion on it. We can either remove it, or add more categories next to announcements. (E.g. "All   |   Announcements | Articles | ...).
I do think categories itself make sense, we just need to define which ones we want.

schultek:
Changes:

- Header is not styled differently, plus the "Blog" tag added to the logo
  - Also docs now have "Docs" tag and learn section "Learn" tag
- Share menu style fixed
- More categories added based on feedback
- Breadcrumbs now contain the category
- Banner removed

@parlough  @abdallahshaban557

schultek:
- [x] License headers updated
- [x] Using `Object?` everywhere
- [x] Changed tagline
- [x] Made `Blog` and other subtypes link to their pages
- [x] Added "More from Dart" section under each post

Open points:

- There are about 200 images in blog posts without an alt attribute, so I can't make it required.
- For the "next posts" I made it that the latest 2 posts in the same category are shown. Does that work for you?
- I'd leave category in sidebar and cta out for now.

parlough:
This still looks good to me. @antfitch Any concerns with landing now and following up later with any further changes, including whatever's needed for funneling the old traffic?

antfitch:
@parlough no concerns on my side. We've got a funnel plan and you, I, and @schultek can work out the fine details later this week.

Issue #7043, opened 2025-11-24
[Primary Constructors] Add primary constructors to the language docs


kallentu:
Source of truth: [b/465472806](http://b/465472806)

GTM Notes: https://docs.google.com/document/d/1TruAZikz-v4VDcoxgQbRScOaRrNhxgE1nvRVwWfEHt8/edit?usp=sharing

### Pages

- [ ] https://dart.dev/language/primary-constructors (new)
- [ ] Add this snippet to a general doc explaining the new changes with `final` and `var`: https://github.com/dart-lang/sdk/issues/62044#issuecomment-3562463986

### References
* [Updating lint docs issue](https://github.com/dart-lang/sdk/issues/62044)  (@pq)

* [Original issue](https://github.com/dart-lang/language/issues/2364)
* [Feature specification](https://github.com/dart-lang/language/blob/main/accepted/future-releases/primary-constructors/feature-specification.md)
* [Tracking doc (go/dart-primary-constructors)](https://docs.google.com/document/d/1zup1_ocbVizW2kKnfMo4ov2eG6R1o8GAxSnyEUfA6MY/edit?tab=t.0#heading=h.6dyyhz5krl9v)
* [Meta issue](https://github.com/dart-lang/sdk/issues/61524)

cc. @antfitch @conooi

kallentu:
Not entirely sure what the process for GTM x language features will look like. I think we're still working on that.
In the meantime, creating an issue to track some existing thoughts we had on documentation changes. We can change/close this issue in the future if we're consolidating the work elsewhere.

kallentu:
Filed the documentation request at b/465472806. 
Will continue to keep this issue open for my own tracking.

parlough:
Thanks @kallentu and @conooi, the docs look great!

We should consider updating some other examples throughout the site to use the new syntax, but we should revisit that after the stable release or even a few releases.

PR #7236, opened 2026-04-02
Add docs for primary constructors

conooi:
Add docs for primary constructor feature. 

Contributes to https://github.com/dart-lang/site-www/issues/7043

**Staged:** https://dart-dev--pr7236-primary-constructors-vzvv9u3e.web.app/language/primary-constructors


src/content/language/constructors.md
```diff
@@ -263,6 +263,45 @@ For more discussion, watch this Decoding Flutter video on tear-offs.
 
 <YouTubeEmbed id="OmCaloD7sis" title="Dart Tear-offs | Decoding Flutter"></YouTubeEmbed>
 
+## Concise constructor syntax
+
+In Dart 3.13 and later, you can omit the class name
+when declaring a generative or factory constructor inside the class body.
+Instead, use the modifier `new` or `factory` directly:
```
kallentu: The word `Instead` probably doesn't make sense here. It's not really countering a point.

Probably something like `inside the class body by using the modifier...`

src/content/language/constructors.md
```diff
@@ -263,6 +263,45 @@ For more discussion, watch this Decoding Flutter video on tear-offs.
...
+  double x, y;
+
+  // Concise unnamed generative constructor
+  new(this.x, this.y);
+
+  // Concise named generative constructor
+  new origin() : x = 0, y = 0;
+
+  // Concise named factory constructor
+  factory fromMap(Map<String, double> map) => Point(map['x']!, map['y']!);
```
kallentu: Maybe use this instead:
```
  // Equivalent to factory Point.clone(Point other)
  factory clone(Point other) => Point(other.x, other.y);
```

The `fromMap` factory is just using null-asserts and has an assumption with no checks and I'd rather us use a simpler, safer example.

src/content/language/constructors.md
```diff
@@ -215,10 +215,10 @@ class to use whenever someone makes a call to the redirecting constructor.
 factory Listenable.merge(List<Listenable> listenables) = _MergingListenable
 ```
 
-It might appear that ordinary factory constructors
-could create and return instances of other classes.
-This would make redirecting factories unnecessary.
-Redirecting factories have several advantages:
+You might think that ordinary factory constructors can create
```
kallentu: Maybe keep most of the previous tone:
`Since ordinary factory constructors can create and return instances of other classes as well, redirecting factories seem unnecessary. However, redirecting factories have several advantages:`

src/content/language/functions.md
```diff
@@ -61,6 +61,7 @@ that indicates whether the `atomicNumber` falls into the noble gas range.
 
 ## Parameters
 
+
```
kallentu: Extra newline here?

src/content/language/primary-constructors.md
```diff
@@ -0,0 +1,190 @@
...
+```
+
+Declaring a parameter in the primary constructor with `var` or `final`
+implicitly induces an instance variable for that parameter.
+
+To ensure this constructor executes on every new instance,
+a class, mixin class, or enum with a primary constructor
+cannot have any other non-redirecting generative constructors.
+
+## Declare parameters
```
kallentu: Rename to `Declaring parameters`

src/content/language/primary-constructors.md
```diff
@@ -0,0 +1,190 @@
...
+
+Placing a parameter list just after the class name
```
kallentu: Maybe remove line 57-58 and for line 59 start with "Parameters in primary constructors with the `var` or `final` modifier..."

src/content/language/primary-constructors.md
```diff
@@ -0,0 +1,190 @@
...
+specifies both a constructor and instance variables.
+Parameters with the `var` or `final` modifier, called **declaring parameters**,
+implicitly induce a field.
+
+Omitting the modifier creates a **non-declaring parameter**,
+which does not induce an implicit field.
+Use non-declaring parameters for custom initialization logic.
```
kallentu: Not sure if we need this sentence. Non-declaring parameters are just what we're familiar with existing normal constructor parameters. This section might be more confusing and could use some simplification.

src/content/language/primary-constructors.md
```diff
@@ -0,0 +1,190 @@
...
+class Point(var int x, var int y) {
+  this : assert(x >= 0 && y >= 0) {
+    print('Point initialized at ($x, $y)');
+  }
+}
+```
+
+This block can specify an initializer list after `this` and/or a function body.
+
+## Private named parameters
```
kallentu: I don't think this section title is accurate. Private named parameters is specifically for _named_ parameters so we could explain that in here, but I don't think it's the right title for this section with its current contents.

Then with the contents itself, since the parameter isn't named, you might as well have written:
```dart
class User(final String _name);
```
which would declare a field `final String _named` with the value that the user puts in at the call site (e.g. `User("Kallen")`)
You wouldn't need to have a parameter with a different name here.

src/content/language/primary-constructors.md
```diff
@@ -0,0 +1,190 @@
...
+}
+```
+
+Dart also supports private named parameters in traditional constructors.
+To learn more, visit
+[Private named parameters](/language/constructors#private-named-parameters).
+
+## Empty bodies
+
+An empty body of a class, mixin class, or extension type (`{}`)
```
kallentu: Extensions can also have `;` bodies.

And they don't need to be using a primary constructor. So we can probably remove the `when using a primary constructor` bit.
Although, then perhaps we should reconsider where this `Empty bodies` section should be.

src/content/language/primary-constructors.md
```diff
@@ -0,0 +1,190 @@
...
+<?code-excerpt "language/lib/primary_constructors/super_parameters.dart"?>
+```dart
+class A(final int a);
+
+class B(super.a) extends A;
+```
+
+## Constraints and breaking changes
+
+To avoid unexpected compiler errors and understand the impact of this feature,
```
kallentu: I wouldn't say these are unexpected, but maybe instead we can write something like "These are some of the constraints and errors you should be aware of:"

src/content/language/primary-constructors.md
```diff
@@ -0,0 +1,190 @@
...
+keep these constraints and edge cases in mind:
+
+*   **No `late` or `external` fields**: The `late` and `external` modifiers
```
kallentu: Perhaps more like "No `late` or `external` parameters in primary constructors"?
Late or external fields are allowed in general.

src/content/language/primary-constructors.md
```diff
@@ -0,0 +1,190 @@
...
+    are not allowed on parameters in the primary constructor header.
+    To use these modifiers, declare the fields in the class body as usual.
+*   **Name collisions**: Declaring a parameter in the primary constructor
+    with the same name as a method or another field in the class body
+    results in a compile-time error.
+
+:::warning
+**Important breaking changes and edge cases:**
+
+*   **Restriction on `final` in normal function parameters**:
```
kallentu: Also, `var`.
Enforced with the `var_with_no_type_annotation` lint.
Also this lint and the `avoid_final_parameters` lint only work in 3.12 and below which might be worth mentioning.

examples/language/lib/primary_constructors/modifier_class.dart
```diff
@@ -0,0 +1,7 @@
+// #docregion modifier-class
+// Using a primary constructor.
+class ModifierClass(this.x) {
+  late int x;
```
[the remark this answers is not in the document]
eernstg: This is not true, there is no such error. I think the example was specifically intended to show that certain variable declarations (in particular, `late` and `external` ones) cannot be declared by a declaring parameter. We _can_ use an initializing formal to initialize the `late` variable, and the example does compile and run without errors. You could claim that it is silly to declare `x` as `late` when it will always be initialized when the enclosing object is created, but it seems useful to illustrate that primary constructors and these kinds of variables _can_ be used together even though those variables can't be introduced by a declaring parameter.

src/content/language/primary-constructors.md
```diff
@@ -0,0 +1,190 @@
...
+Parameters with the `var` or `final` modifier, called **declaring parameters**,
+implicitly induce a field.
+
+Omitting the modifier creates a **non-declaring parameter**,
+which does not induce an implicit field.
+Use non-declaring parameters for custom initialization logic.
+
+Because `final` and `var` modifiers on parameters are reserved exclusively
+for declaring parameters in primary constructors,
+you cannot use them on parameters in other kinds of functions.
```
eernstg: Perhaps move this paragraph down after the code excerpt (the code illustrates things explained above, and does not give any examples of the error mentioned in this paragraph).

src/content/language/primary-constructors.md
eernstg: Perhaps this would be more helpful if the section had two examples showing the same outcome achieved in the old/standard way as well as the new 'private named' way:

```dart
// Variant not using a private named parameter.
class User({required String name}) {
  String _name = name;
}

// Variant using a private named parameter.
class User({required var String _name});
```

.. and then mention that the call sites are identical, `User(name: 'John Doe')`.

src/content/language/primary-constructors.md
eernstg: Could also be "A declaring parameter cannot be `late` or `external` (those variables must be declared in the body, like today)."

src/content/language/primary-constructors.md
```diff
@@ -0,0 +1,190 @@
...
+    becomes a compile-time error.
+    It is reserved exclusively for declaring parameters in primary constructors.
+    To enforce immutable parameters as a style choice,
+    use the [avoid_final_parameters](https://dart.dev/tools/linter-rules/avoid_final_parameters)
+    or [parameter_assignments](https://dart.dev/tools/linter-rules/parameter_assignments) linter rules.
+*   **The `factory` method edge case**:
+    If you have a method named `factory` with no return type
+    (for example, `factory() {}`),
+    it will be parsed as a factory constructor after primary constructors ship.
+    Ensure your methods have explicit return types to avoid this conflict.
```
eernstg: Perhaps:
```suggestion
    Make sure such methods have explicit return types to avoid this conflict.
```
Otherwise it might be interpreted to mean that developers must add explicit return types to all methods, which could be a much bigger effort. Also, it would be counterproductive to add explicit return types on instance methods whose return type is obtained by override inference.

eernstg (commented):
Added a few drive-by comments.

src/content/language/constructors.md
conooi: Oops I ran the proofreading skill and I think it made this suggestion! Thanks for catching it

src/content/language/primary-constructors.md
conooi: Thanks for the suggestion! However, the [Google developer documentation style guide](https://developers.google.com/style/headings#ing-verb-forms) recommends avoiding -ing verb forms as the first word in headings when possible.

How about `Parameter declarations` instead?

src/content/language/primary-constructors.md
conooi: Changing it to : "Keep these constraints and potential errors in mind when using primary constructors:" 

To make the sentence more active.

src/content/language/constructors.md
```diff
@@ -263,6 +263,46 @@ For more discussion, watch this Decoding Flutter video on tear-offs.
...
+concise named constructors do not use a dot
+between the keyword (modifier) and the name.
+
+<?code-excerpt "concise_syntax.dart (concise-syntax)"?>
+```dart
+class Point {
+  double x, y;
+
+  // Concise unnamed generative constructor
+  new(this.x, this.y);
```
eernstg, line 285: Standard formatting will include a space between the keyword `new` or `factory` and the formal parameter list, as a reminder that the keyword is not an identifier. A bit like `if (condition) ...` rather than `if(condition) ...` which shouldn't read as a function call.

src/content/language/primary-constructors.md
```diff
@@ -0,0 +1,206 @@
...
+<?code-excerpt "language/lib/primary_constructors/primary_constructors.dart (constructor-bodies)" replace="/PointWithBody/Point/g"?>
+```dart
+class Point(var int x, var int y) {
+  this : assert(x >= 0 && y >= 0) {
+    print('Point initialized at ($x, $y)');
+  }
+}
+```
+
+This block can specify an initializer list after `this` and/or a function body.
```
eernstg, line 121: Perhaps extend the text slightly to cover forms like `this : assert(x >= 0);` and `@metadata this;` as well?

eernstg (approved):
LGTM!

src/content/language/primary-constructors.md
kallentu: The "declaring" in declaring parameters is moreso of a adjective rather than a verb here, but I see that we'd probably still want to avoid that. It's less so about the section being a section for how to declare parameters and more about "declaring parameters" which allows the users to declare fields in parameters.

I don't think `Parameter declarations` is accurate. Hmm, maybe something like "Field declarations in parameters"?

src/content/language/mixins.md
```diff
@@ -208,7 +208,7 @@ class Novice extends Musician { // Use Musician as a class
 ```
 
 Any restrictions that apply to classes or mixins also apply to mixin classes:
-
+
```
kallentu: Nit: There's an extra space here.

kallentu (approved):

src/content/language/primary-constructors.md
conooi: Sounds good! I like "Field declarations in parameters"

src/content/language/primary-constructors.md
```diff
@@ -0,0 +1,214 @@
...
+  url: /language/constructors
+  title: Constructors
+nextpage:
+  url: /language/methods
+  title: Methods
+---
+
+:::version-note
+Primary constructors require a language version of at least 3.13.
+:::
```
parlough, line 14:
```suggestion
:::version-note
Primary constructors require a [language version][] of at least 3.13.
:::

[language version]: /language/versioning
```

src/content/language/primary-constructors.md
```diff
@@ -0,0 +1,214 @@
+---
+title: Primary constructors
+description: Create classes concisely by declaring parameters and fields in the class header.
```
parlough, line 3: To keep this under 80 characters:

```suggestion
description: >-
    Create classes concisely by
    declaring parameters and fields in the class header.
```

src/content/language/primary-constructors.md
```diff
@@ -0,0 +1,214 @@
...
+    [parameter_assignments](https://dart.dev/tools/linter-rules/parameter_assignments)
+    linter rule.
+*   **The `factory` method edge case**:
+    If you have a method named `factory` with no return type
+    (for example, `factory() {}`),
+    the compiler parses it as a factory constructor
+    after primary constructors ship.
+    Make sure such methods have explicit return types
+    to avoid this conflict.
+:::
```
parlough, line 213: Let's put the lint in code font and use a link definition.

```suggestion
    To enforce immutable parameters as a style choice in Dart 3.13 and later,
    use the [`parameter_assignments`][] linter rule.
*   **The `factory` method edge case**:
    If you have a method named `factory` with no return type
    (for example, `factory() {}`),
    the compiler parses it as a factory constructor
    after primary constructors ship.
    Make sure such methods have explicit return types
    to avoid this conflict.
:::

[`parameter_assignments`]: /tools/linter-rules/parameter_assignments
```

src/content/language/primary-constructors.md
```diff
@@ -0,0 +1,214 @@
...
+
+In both cases, the caller uses the public name `name`
+at the call site: `User(name: 'John Doe')`.
+
+## Empty bodies
+
+An empty body of a class, mixin class, extension, or extension type (`{}`)
+can be replaced by a semicolon (`;`).
+While this is true in general for these declarations, it is particularly
+useful when using a primary constructor to keep the entire declaration on a single line.
```
parlough, line 160:
```suggestion
While this is true in general for these declarations,
it is particularly useful when using a primary constructor to
keep the entire declaration on a single line.
```

src/content/language/functions.md
```diff
@@ -211,6 +211,14 @@ assert(say('Bob', 'Howdy') == 'Bob says Howdy with a carrier pigeon');
 
 <a id="the-main-function" aria-hidden="true"></a>
 
+### Parameter modifiers
+
+In Dart 3.13 and later, you cannot use modifiers like `final` or `var` for normal function parameters.
+These keywords are now reserved exclusively for [primary constructors](/language/primary-constructors) to declare instance fields.
+
+If you wish to enforce immutability for function parameters,
+use lints like [`parameter_assignments`](https://dart.dev/tools/linter-rules/parameter_assignments) instead of the `final` keyword in the signature.
```
parlough, line 220: - Add breaks to keep under 80 characters.
- Use link definitions.
- Use "Can't" instead of "cannot"
- Use the site relative path instead of full absolute url.

```suggestion
In Dart 3.13 and later,
you can't use modifiers like `final` or `var` for normal function parameters.
These keywords are now reserved exclusively for
[primary constructors][] to declare instance fields.

If you wish to enforce immutability for function parameters,
use lints like [`parameter_assignments`][] instead
of the `final` keyword in the signature.

[primary constructors]: /language/primary-constructors
[`parameter_assignments`]: /tools/linter-rules/parameter_assignments
```

src/content/language/constructors.md
```diff
@@ -263,6 +263,46 @@ For more discussion, watch this Decoding Flutter video on tear-offs.
...
+
+In Dart 3.13 and later, you can omit the class name
+when declaring a generative or factory constructor inside the class body
+by using the modifier `new` or `factory` directly:
+
+-   `new` or `new named`
+-   `factory` or `factory named`
+
+Unlike traditional named constructors (such as `Point.origin()`),
+concise named constructors do not use a dot
```
parlough, line 276:
```suggestion
concise named constructors don't use a dot
```

src/content/language/primary-constructors.md
```diff
@@ -0,0 +1,214 @@
...
+:::
+
+## Overview
+
+Primary constructors provide a concise way to declare a class's fields
+and its main constructor in a single line.
+They reduce the boilerplate of declaring fields, passing parameters,
+and assigning them in the constructor body.
+This shorthand is purely "syntax sugar"
+and does not introduce new runtime semantics.
```
parlough, line 23: Maybe rather than put "syntax sugar" in quotes as a sort of explanation, we could just be more direct about what that means?

```suggestion
This shorthand changes how you write the declaration,
but it doesn't change runtime behavior.
```

src/content/language/primary-constructors.md
```diff
@@ -0,0 +1,214 @@
...
+<?code-excerpt "language/lib/primary_constructors/primary_constructors.dart (constructor-bodies)" replace="/PointWithBody/Point/g"?>
+```dart
+class Point(var int x, var int y) {
+  this : assert(x >= 0 && y >= 0) {
+    print('Point initialized at ($x, $y)');
+  }
+}
+```
+
+This block can specify an initializer list after `this` and/or a function body.
```
parlough, line 121: It's this type of block, not necessarily this exact block right? Perhaps "the" can expand it to similar situations.

```suggestion
The block can specify an initializer list after `this` and/or a function body.
```

src/content/language/primary-constructors.md
```diff
@@ -0,0 +1,214 @@
...
+It also supports forms with just an initializer list followed by a semicolon
+(for example, `this : assert(x >= 0);`),
+and you can apply metadata annotations to it
+(for example, `@metadata this;`).
```
parlough, line 125: Instead of talking about what it supports, would it make sense to focus on how to complete certain actions/tasks?


```suggestion
To provide only an initializer list,
end it with a semicolon, such as `this : assert(x >= 0);`.
You can also annotate the block with metadata,
such as `@metadata this;`.
```

parlough (commented):
Love it! Some feedback to consider:

src/content/language/functions.md
conooi: Oops sorry was going a bit too quick with this one! Thanks for catching all this

PR #7250, opened 2026-04-15
Add private named parameters section

conooi:
Create a new section on the constructors page for the new 3.12 feature Private Named Parameters. 

---


examples/analysis_options.yaml
```diff
@@ -5,6 +5,8 @@ analyzer:
   language:
     strict-casts: true
     strict-raw-types: true
+  enable-experiment:
+    - private-named-parameters
```
munificent: If you bump the SDK constraint in the repo's pubspec.yaml file to `^3.12.0`, then you shouldn't have to enable this experiment.

site/lib/src/pages/glossary.dart
```diff
@@ -21,7 +21,8 @@ enum ResourceType {
   video,
   code,
   diagnostic,
-  external;
+  external
+  ;
```
munificent: Nit: I would keep this `;` on the same line as `external`. Did the formatter do this?

src/content/language/constructors.md
```diff
@@ -393,6 +380,121 @@ class PointD {
...
+}
+```
+
+#### Constraints
+
+* **No conflicts:** Neither the private name nor the generated public name
+  can match any other parameter name in the same constructor.
+* **Initializing formals only:** Named parameters in Dart generally
+  cannot be private. This feature is an exception that only applies
+  to named parameters that are initializing formals (`this._field`).
```
munificent, line 454: Just a note for the future: When primary constructors arrive in Dart 3.13, you will also be able to use private named parameters for declaring parameters like:

```dart
class Point({final int _x = 0});
```

So it's correct now that it's only for initializing formals, but for Dart 3.13, it will be initializing formals and declaring parameters. :)

src/content/language/constructors.md
```diff
@@ -265,7 +265,12 @@ For more discussion, watch this Decoding Flutter video on tear-offs.
 
 ## Instance variable initialization
 
-Dart can initialize variables in three ways.
+Dart provides several ways to initialize instance variables.
+You can assign values in the declaration,
+use initializing formal parameters,
+or use an initializer list.
+For private fields,
+Dart also supports a dedicated syntax using private named parameters.
```
munificent: I'm not sure if this last sentence is worth calling out here. It's not really a separate feature so much as if you have an initializing formal that happens to be a named parameter and have a private name, then it does this little extra thing where the parameter gets a public name.

src/content/language/constructors.md
```diff
@@ -393,6 +380,121 @@ class PointD {
...
+
+  @override
+  String toString() => 'PointPrivate($_x, $_y)';
+}
+
+void testPrivate() {
+  var p = PointPrivate(x: 1.0, y: 2.0);
+  print(p);
+}
+```
```
munificent, line 446: These examples are all excellent!

munificent (approved):
This looks great! \o/

site/lib/src/pages/glossary.dart
conooi: Yeah, the formatter did this D:

examples/analysis_options.yaml
parlough: As Bob mentioned: Rather than enabling this experiment, consider changing the SDK constraint in `/examples/misc/pubspec.yaml` to `^3.12.0-0` to enable 3.12 language features.

src/content/language/constructors.md
```diff
@@ -393,6 +380,121 @@ class PointD {
...
+
+In both cases,
+the caller uses the public name `x` at the call site:
+
+<?code-excerpt "point_private_new.dart (initialize-private-named-usage)" plaster="none"?>
+```dart
+var p = Point(x: 1.0);
+```
+
+Like regular named parameters, you can make private named parameters optional or required.
```
parlough: Optionally consider cross linking named parameters to https://dart.dev/language/functions#named-parameters.

parlough:
```suggestion
Like regular named parameters, you can
make private named parameters optional or required.
```

src/content/language/constructors.md
```diff
@@ -393,6 +380,121 @@ class PointD {
...
+You can also provide explicit default values.
+
+In the following example, the `_x` parameter is optional and defaults to `null`.
+The `_y` parameter is also optional but uses an explicit default value of `0.0`:
```
parlough:
```suggestion
The `_y` parameter is also optional but has an explicit default value of `0.0`:
```

src/content/language/constructors.md
```diff
@@ -393,6 +380,121 @@ class PointD {
...
+```
+
+#### Constraints
+
+* **No conflicts:** Neither the private name nor the generated public name
+  can match any other parameter name in the same constructor.
+* **Initializing formals only:** Named parameters in Dart generally
+  cannot be private. This feature is an exception that only applies
+  to named parameters that are initializing formals (`this._field`).
+  You cannot use private names for regular named parameters.
```
parlough: Generally we use "can't" and similar for [negation contractions](https://developers.google.com/style/contractions#negation-contractions).


```suggestion
  can't be private. This feature is an exception that only applies
  to named parameters that are initializing formals (`this._field`).
  You can't use private names for regular named parameters.
```

parlough: Maybe "identifiers" is more accurate than "names"? I'm not sure, but it seems like a name can't be private but an identifier(/something with an identifier) can be.

src/content/language/constructors.md
```diff
@@ -393,6 +380,121 @@ class PointD {
...
+* **Valid public name:** The private name must map to a valid public identifier.
```
parlough, line 456: Good call out!

src/content/language/constructors.md
```diff
@@ -393,6 +380,121 @@ class PointD {
...
+  print(p);
+}
+```
+
+#### Constraints
+
+* **No conflicts:** Neither the private name nor the generated public name
+  can match any other parameter name in the same constructor.
+* **Initializing formals only:** Named parameters in Dart generally
+  cannot be private. This feature is an exception that only applies
```
parlough: Consider changing "feature" to "capability". It seems less ambiguous and more consistent for the future, where this "new feature" just becomes a part of the language.

src/content/language/constructors.md
```diff
@@ -393,6 +380,121 @@ class PointD {
...
+  to named parameters that are initializing formals (`this._field`).
+  You cannot use private names for regular named parameters.
+* **Valid public name:** The private name must map to a valid public identifier.
+  For example, `this._` or `this._2x` are invalid
+  because they don't have valid public counterparts.
+
+#### Usage in initializer lists
+
+Within the constructor's initializer list,
+you refer to the parameter using its private name:
```
parlough:
```suggestion
reference the parameter using its private name:
```

src/content/language/constructors.md
parlough: I like the expansion of this introduction!

However, I think this additional mention might not be needed as it's just a part of (the already mentioned) initializing formal parameters support.

src/content/language/constructors.md
```diff
@@ -393,6 +380,121 @@ class PointD {
...
+}
+```
+
+#### Interaction with super parameters
+
+When extending a class that uses private named parameters,
+subclasses use the public name for [super parameters](/resources/glossary#super-parameter).
+
+In the following example, the `Tool` class defines the private field `_price`.
+Even though the field is private, the compiler exposes the parameter name publicly.
```
parlough:
```suggestion
Even though the field is private,
the compiler exposes the parameter name publicly.
```

src/content/language/constructors.md
```diff
@@ -393,6 +380,121 @@ class PointD {
...
+  final double _x;
+
+  PointPrivateAssert({required this._x}) : assert(_x >= 0);
+}
+```
+
+#### Interaction with super parameters
+
+When extending a class that uses private named parameters,
+subclasses use the public name for [super parameters](/resources/glossary#super-parameter).
```
parlough: Consider using a link reference for this and declaring the definition at the end of the section.


```
subclasses use the public name for [super parameters][].

[super parameters]: /resources/glossary#super-parameter
```

examples/misc/lib/language_tour/classes/point_alt.dart
```diff
@@ -96,3 +96,42 @@ class PointD {
...
+// #enddocregion initialize-private-named-assert
+
+// #docregion initialize-private-named-super
+class Tool {
+  final int _price;
+  Tool({required this._price});
+}
+
+class Hammer extends Tool {
+  // Forwards to the public 'price' argument
```
parlough, line 134:
```suggestion
  // Forwards to the public `price` parameter:
```

src/content/language/constructors.md
parlough: I'm not sure a better way to word this, but to me, "the compiler" being the actor here seems a bit complicated or not fully accurate. It's a language feature. It's also not clear to me until the next sentence what "exposing a parameter name publicly" exactly means.

In absence of a better actor, would it be fine to just say the parameter is public? Something like the following:

```
Even though the field is private,
its corresponding named parameter is public (`price` not `_price`).
```

src/data/glossary.yml
```diff
@@ -1921,6 +1921,29 @@
     - "subtyping"
     - "subtype polymorphism"
 
+- term: "Super parameter"
```
parlough, line 1924: Love this new glossary addition!

parlough, line 1924: Some of the descriptions in this entry go beyond our standard <= 80 characters per line. Can you adjust them, adding some semantic breaks where necessary? Otherwise the content looks great to me! :D

src/content/language/constructors.md
```diff
@@ -393,6 +380,121 @@ class PointD {
...
+
+:::version-note
+Using private named parameters as initializing formals
+requires a [language version][] of at least 3.12.
+:::
+
+In Dart, fields starting with an underscore are private to their library.
+To initialize a private field using a named parameter,
+you previously had to write manual assignment boilerplate
+in the initializer list:
```
parlough: I think this framing is fine, but eventually "previously" will get a bit out dated, with future readers not writing code before the new syntax existed. Optionally consider adjusting this to a "you can" structure or leaving a comment to revisit this in the future.

src/content/language/constructors.md
```diff
@@ -393,6 +380,121 @@ class PointD {
 }
 ```
 
+### Private named parameters
```
parlough, line 381: Should we name this section and references to this capability as "private-named parameters" to clarify the parameter itself isn't private, but rather the name is?

parlough (approved):
Love all these updates @conooi!

A few suggestions to consider, but generally looks great to me.

src/content/language/constructors.md
conooi: @munificent thoughts?

munificent: It's not really just about parameters whose name happens to be private. Dart has already allowed those for positional parameters:

```dart
class C {
  int _x;
  C(this._x);
}
```

It's really about named parameters whose name is private. I thought about calling the feature "private-named named parameters" but that's obviously no fun for anyone to read or say. :) For what it's worth, I think of the feature as "private (named parameters)" more than "(private named) parameters".

But if you think the hyphen will help clarify things for readers, I'm OK with it.

conooi: I think we should keep the title as is to remain consistent with the feature spec (if that matters). Do you have strong opinions about the hyphen @parlough?

parlough: Nope, I'm happy with keeping it as is and being consistent! Thanks for considering this though :)

PR #7257, opened 2026-04-20
Update Jaspr to 0.23 as well as related dependencies

parlough:
These updates come with fixes to some issues we've been hitting, including double template rendering and running out of file descriptors.


conooi (approved):
Thank you, @parlough!!

PR #7269, opened 2026-05-02
fix(site): prevent docs page title overflow on small screens

Asif-Zaman-Suvo:
### Summary

Fixes the docs **page header card** layout on very narrow viewports so the title stays inside the card, wraps at word boundaries, and leaves room for the **page options (⋮)** control.

### Problem

- On small screens, the raised header (`#site-content-title.wrap`) could **overflow horizontally**.
- Long titles could break in the **middle of a word** (e.g. “documentation” split oddly).
- The **⋮** button in the top-right could **overlap** the title without enough inline padding.

### Changes

**`site/lib/_sass/components/_content.scss`**

- `min-width: 0` on the wrapped header and `h1` so flex items can shrink instead of overflowing.
- Extra `padding-inline-end` on the card so the title clears the options control.
- `word-break: normal`, `overflow-wrap: normal`, and `text-wrap: pretty` so line breaks prefer **spaces between words**.
- **≤360px width:** page title `h1` uses **`2.375rem`**; wider viewports keep the default **`2.75rem`** from `_base.scss`.

### Screenshots

Screenshot 2026-05-03 at 1 15 29 AM
Screenshot 2026-05-03 at 1 15 16 AM

### Testing

- Resize to a narrow width (e.g. ~360px) on a docs page with the card header (e.g. `/docs`) and confirm no horizontal overflow and sensible title wrapping.

### DIFF
--- a/site/lib/_sass/components/_content.scss
+++ b/site/lib/_sass/components/_content.scss
@@ -68,7 +68,7 @@ article {
       min-width: 0;
       max-width: 100%;
       word-break: normal;
-      overflow-wrap: normal;
+      overflow-wrap: break-word;
 
       @media (max-width: 390px) {
         font-size: 2.375rem;
--- a/site/lib/_sass/components/_content.scss
+++ b/site/lib/_sass/components/_content.scss
@@ -55,14 +55,27 @@ article {
       display: flex;
       flex-direction: column;
       gap: 0.35rem;
+      min-width: 0;
 
       background-color: var(--site-raised-bgColor);
       padding: 1rem;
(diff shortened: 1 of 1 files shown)


parlough (approved):
Thanks for investigating and improving the behavior of the page headers!

Asif-Zaman-Suvo:
> Thanks for investigating and improving the behavior of the page headers!

Thank you @parlough for reviewing and merging the PR.
I'm happy to contribute to the Dart documentation and hope to contribute more going forward

PR #7274, opened 2026-05-07
Update lints and diagnostics for latest updates

### LINKED ISSUE #63329: Syntax error in the example code for the lint `simplify_variable_pattern`
The lint documentation: https://dart.dev/tools/linter-rules/simplify_variable_pattern
The bad example is currently this:

```dart
void f(Object o) {
  if (o case String(length: length)) {
    print('string is $length characters long');
  }
}
```

It misses `var` or `final` after the colon and doesn't build:

```
  error • temp.dart:2:29 • Undefined name 'length'. Try correcting the
          name to one that is defined, or defining the name. •
          undefined_identifier
```

Need to change to something like this:

```dart
void f(Object o) {
  if (o case String(length: var length)) {
    print('string is $length characters long');
  }
}
```

By the way, that's the example in https://dart.dev/tools/diagnostics/simplify_variable_pattern
It's unclear why we have both, how they are related, and why they have diverged.

parlough:
These changes have already been reviewed in the SDK, this just updates the vendored files with the latest source data.

Also enables proper styling for marking diagnostic code spans as errors with `[! !]`.

Fixes https://github.com/dart-lang/sdk/issues/63329

### DIFF
--- a/site/lib/src/pages/custom_pages.dart
+++ b/site/lib/src/pages/custom_pages.dart
@@ -156,6 +156,7 @@ List<MemoryPage> get _lintMemoryPages {
               'title': lint.name,
               'underscore_breaker_titles': true,
               'description': 'Learn about the ${lint.name} linter rule.',
+              'bodyClass': 'highlight-diagnostics',
               // If this lint has a version not fully lower case,
               // specify that version as the canonical destination.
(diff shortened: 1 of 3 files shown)


conooi (approved):
lgtm thank you!!

PR #7280, opened 2026-05-09
Temporarily disable format check in CI

parlough:
While we move between branches over the next few weeks, to get the build passing (which still checks for valid analysis and passing tests).


conooi (approved):
Thanks, @parlough !!

antfitch:
Thank you so much for this!

PR #7286, opened 2026-05-13
Add helpful alt text to every blog post image

parlough:
Also update the `DashImage` component to require non-empty alt text for ensuring better accessibility of images.

@conooi No need to review this too closely as the existing blog posts are all historical at this point. The main thing is accessible `alt` text now being required for all blog images.


conooi (approved):
@parlough thank you!!

PR #7292, opened 2026-05-14
Work around aspect-ratio issue on older WebKit versions

### LINKED ISSUE #7289: Users are reporting issues with content being rendered for the blog
### Describe the problem

We have received reports on our blog having rendering issues when viewing it on iPads, the layout looks off

Image

### Expected fix

The layout info for the images and content needs to be fixed

### On which browser(s) did you experience this issue?

_No response_

### Additional context

_No response_

parlough:
Work around an issue with `aspect-ratio` on older WebKit versions by moving the relevant styles to the image's parent element.

Fixes https://github.com/dart-lang/site-www/issues/7289

I verified the issue existed on an iPad Pro 10.5 inch simulator that had iOS 17.5 installed and validated this fix on the simulator as well:

Screenshot of blog index page with simulator in portrait mode
Screenshot of blog index page with simulator in landscape mode

Before:

Screenshot of blog index page with simulator in portrait mode before the fix

### DIFF
--- a/site/lib/_sass/components/_blog.scss
+++ b/site/lib/_sass/components/_blog.scss
@@ -214,10 +214,6 @@
       .blog-card-image {
         flex: 1.5;
         border-radius: 4px;
-
-        img {
-          border-radius: 4px;
-        }
       }
 
       .blog-card-content {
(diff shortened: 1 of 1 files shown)


schultek (approved):
LGTM, thanks for this fix

PR #7293, opened 2026-05-14
Upgrade build_web_compilers to 4.5.1

biggs0125:
Includes change in build_web_compilers to deduplicate wasm deferrred loading module loading.


parlough (approved):
Thanks for investigating and improving this :D - Looks good to me.

PR #7306, opened 2026-05-19
Add 3.12 entries to deprecated breaking change and evolution pages

parlough:
Since we haven't finalized the deprecation and removal of these pages, add 3.12 entries so they are still up to date.


conooi (approved):
thank you!!

PR #7308, opened 2026-05-19
Add Dart 3.12 banner 

conooi:


parlough (approved):
I pushed an alternative implementation based on Markdown for supporting multiple links, but looks good to me. Thanks!

PR #7320, opened 2026-05-31
Update http package version to ^1.4.0 to match the output

SMMousaviSP:
The output which is the result of resolving dependencies does not match the version specified.


parlough (approved):
Both are fine as it should resolve to the newer version if available anyway, but updating this works for me. Thanks!

PR #7336, opened 2026-06-12
docs: clarify async program exit behavior in tutorial

### LINKED ISSUE #7332: Reference to program exit error in chapter 3 that isn't fully explained
### Page URL

https://dart.dev/learn/tutorial/packages-libs

### Page source

https://github.com/dart-lang/site-www/blob/main/src/content/learn/tutorial/packages-libs.md

### Describe the problem

Chapter 4, "Organize code with packages and libraries," references multiple times the description of a program exit issue in the code of chapter 3, "Write asynchronous code," but from what I could tell this description isn't actually in chapter 3, leaving me with the knowledge that there is a program exit issue, but not exactly why that issue exists.

### Expected fix

I would say the ideal fix is to write the clear description of the program exit issue and put it towards the end of chapter 3, where chapter 4 says the description is located.

### Additional context

_No response_

harshyadavDeveloper:
Updates the packages and libraries tutorial to directly explain the asynchronous program exit issue instead of referring to a discussion in Chapter 3 that no longer exists.

This makes the tutorial self-contained and avoids confusion for readers following the lesson sequence.

Fixes #7332


src/content/learn/tutorial/packages-libs.md
```diff
@@ -179,8 +179,8 @@ you can use a [path dependency][].
 Now that you've added `command_runner` as a dependency,
 you can import it into your `cli` application and
 replace your existing argument-handling logic with
-the new `CommandRunner` class. This step also
-fixes the program exit behavior discussed at the end of Chapter 3.
+the new `CommandRunner` class. This step also 
+fixes an issue where the program can exit before asynchronous operations complete.
```
parlough: I think we can leave this part out since it's a bit abstract at this point and it'll be covered below. What do you think?


```suggestion
the new `CommandRunner` class.
```

src/content/learn/tutorial/packages-libs.md
parlough: This is pre-existing, but let's consider adjusting this. Two things:

1. It's not the function being `async` that addresses the issue, but it does enable the use of `await`.
2. I believe adding the `await` isn't actually necessary to ensure the runner's completion. Perhaps we can reframe this around how it makes completion explicit and improves things like error propagation.

How about something like the following?

```suggestion
    - Declaring `main` as `async` lets it use `await`.
      Because `runner.run()` returns a `Future`,
      awaiting it makes `main` explicitly wait for the runner's completion,
      ensuring clear logic order and improving error propagation.
```

parlough (commented):
Thanks for tackling this @harshyadavDeveloper! One optional suggestion to a consider and a larger adjustment to improve the accuracy of the pre-existing statement. Let me know what you think.

harshyadavDeveloper:
Thanks for the review! I've applied the suggestion above and updated the explanation to focus on `async` enabling `await` and on explicit completion/error propagation.

parlough (commented):
Thanks for making those adjustments! Looks good to me.

parlough (approved):

PR #7338, opened 2026-06-14
Update diagnostics and lints as of 2026-06-14

parlough:
The updates are generated from content that's already been reviewed landing in the SDK.


conooi (approved):
tysm @parlough !!

PR #7342, opened 2026-06-16
Add enums, const, and more sections to primary constructor docs

### LINKED ISSUE #7331: [Primary Constructors] Add section on const primary constructors, enums, and link new syntax.
(Based on feedback from the recent bug bash)

The current documentation seems to be missing information on:
- Using primary constructors with enums
  - This is making agents assume that you can't use the feature on enums and our bashers couldn't tell if they could or couldn't based on the current docs.
  - Maybe we should add some examples or an extra minor section?
- Const primary constructors `class const Foo() {}`
  - Add a small example.
- Perhaps link the new constructor syntax? Not sure about this one, but I wonder if it would be worth mentioning if people are looking for that syntax and they search up the primary constructor feature in the docs. Not sure what the best layout is here.

The first two are definitely something we'd want to add to the docs.
@conooi Thoughts?

conooi:
Added deep-dive sections and compiling code examples for constructor scopes, constant/named constructors, enhanced enums, and super parameters. Also documented all spec-defined compile-time constraints and established the correct ["in-body constructor" terminology](https://github.com/dart-lang/sdk/issues/63564#issuecomment-4664006209).

Fixes https://github.com/dart-lang/site-www/issues/7331

### DIFF
--- a/examples/language/lib/primary_constructors/primary_constructors.dart
+++ b/examples/language/lib/primary_constructors/primary_constructors.dart
@@ -62,6 +62,4 @@ enum Color(final String hex) {
   green('#00FF00'),
   blue('#0000FF');
 }
-
 // #enddocregion enums
-
--- a/examples/language/lib/primary_constructors/primary_constructors.dart
+++ b/examples/language/lib/primary_constructors/primary_constructors.dart
@@ -6,6 +6,7 @@ class Point(var int x, var int y);
 
 // Doesn't declare a field.
 class User(String name);
+
 // #enddocregion declaring-parameters
 
 // #docregion initializer-scope
@@ -13,6 +14,7 @@ class DeltaPoint(final int x, int delta) {
   // Accesses 'x' and 'delta' parameters directly!
   final int y = x + delta;
 }
+
 // #enddocregion initializer-scope
 
 // #docregion constructor-bodies
@@ -21,6 +23,7 @@ class PointWithBody(var int x, var int y) {
     print('Point initialized at ($x, $y)');
   }
 }
+
 // #enddocregion constructor-bodies
 
 // #docregion private-fields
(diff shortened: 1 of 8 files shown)


examples/language/lib/primary_constructors/super_parameters.dart
```diff
@@ -1,5 +1,23 @@
+abstract class Widget {
```
kallentu: Is there a doc region for this?
And what are these examples for?

examples/language/lib/primary_constructors/primary_constructors.dart
```diff
@@ -34,11 +34,32 @@ class UserWithPrivateField(String name) {
 class EmptyBodyPoint(var int x, var int y);
 // #enddocregion empty-bodies
 
+// #docregion scoping-shadowing
+class ScopingDemo(var String x) {
+  // In a field initializer, 'x' refers to the parameter 'x'.
```
kallentu: Because `captureAtDeclaration` is late, it refers to the field and not the parameter `x`. Might also be worth renaming these fields because we removed the "capturing" part which were the function expressions.

src/content/language/primary-constructors.md
```diff
@@ -192,6 +310,31 @@ when using primary constructors:
 *   **Name collisions**: Declaring a parameter in the primary constructor
     with the same name as a method or another field in the class body
     results in a compile-time error.
+*   **No assignments to primary parameters**:
```
kallentu: "primary parameters" -> "primary constructor parameters"
(I see that it's written like this in the spec, but I believe we're trying to write primary constructor parameters)

examples/language/lib/primary_constructors/super_parameters.dart
conooi: Hi @kallentu! Those were mock classes just to make a Flutter-like widget example compile under static analysis.

To avoid confusion, I decided to change the example to a clean, self-contained Person and Employee hierarchy instead.

kallentu (approved):

parlough (approved):
Thanks Connie! These additions look great.

Note I made various adjustments to increase consistency with the rest of the page and better align with the documentation style guide. Do consider completing a pass of the updated content when you have a chance.

PR #7343, opened 2026-06-18
Standardize and consolidate font setup

parlough:
Corresponds to https://github.com/flutter/website/pull/13518, updating the Dart site to use newer fonts, consistent with the Flutter docs site. These Google Sans Flex and Code fonts are publicly available and support variable axes. This increases consistency across the sites and enables compliant use of the same fonts on forked/translated sites.

- Replaces many manual font-weights to use one of the updated CSS properties.
- Converts uses of Google Sans and Google Sans Text to Google Sans Flex.
- Converts uses of Google Sans Mono to Google Sans Code.
- Combines some of the requests for slightly faster font loading.
- Switches the blog from Noto Serif to Roboto Serif to be more consistent with the nearby sans serif fonts.

### DIFF
--- a/site/lib/_sass/base/_base.scss
+++ b/site/lib/_sass/base/_base.scss
@@ -3,8 +3,9 @@
 body {
   font-family: var(--site-body-fontFamily);
   font-size: 1rem;
-  font-weight: 400;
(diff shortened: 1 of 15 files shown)


conooi (approved):
Thanks!!

PR #7349, opened 2026-06-22
Add keyboard input styles and usage

parlough:
Adds custom styles for `<kbd>` (migrated from `flutter/website`) and use it in more locations for better semantics and styling of keybindings. Also add support for a `special-key` class that replaces `Cmd/Ctrl` with the OS-appropriate key on load.

Some example usages: https://dart-dev--pr7349-feat-kbd-styles-8lpgp5fe.web.app/web/debugging#overview


conooi (approved):
thanks @parlough !!

PR #7351, opened 2026-06-24
Update functions.md

tagir404:
Anonymous functions example mentions forEach, but the code uses a for-in loop
image

### DIFF
--- a/src/content/language/functions.md
+++ b/src/content/language/functions.md
@@ -347,9 +347,7 @@ var uppercaseList = list.map((item) {
 }).toList();
 // Convert to list after mapping
 
-for (var item in uppercaseList) {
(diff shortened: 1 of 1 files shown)


conooi:
@tagir404 Thanks for catching this discrepancy! The code was intentionally updated to a for-in loop to align with the Effective Dart guideline [AVOID using forEach with a function literal](https://dart.dev/effective-dart/usage#avoid-using-foreach-with-a-function-literal).

Because of that, we want to keep the for-in loop and instead update the explanatory prose on lines 338–339 to match the code (as well as update excerpt sources in examples/).

I'll close this PR and open a quick fix for the text. Thanks again for flagging this!

PR #7362, opened 2026-07-06
Add potentially constant glossary term

### LINKED ISSUE #7354: Add glossary entry for "potentially constant"
And use it in the primary constructor docs

conooi:
Fixes https://github.com/dart-lang/site-www/issues/7354

### DIFF
--- a/src/content/language/primary-constructors.md
+++ b/src/content/language/primary-constructors.md
@@ -247,9 +247,9 @@ Constant primary constructors have these important constraints:
     every instance variable must be `final`, can't be `late`,
     and must be definitely initialized by a declaring parameter,
     field initializer, or the primary constructor's initializer list.
-*   **Potentially constant initializers**:
+*   **[Potentially constant][] initializers**:
     The initializing expressions for each instance variable
-    must be able to be evaluated at compile time.
+    must be potentially constant.
     This includes field initializers and
     expressions in the primary constructor's initializer list.
 
@@ -382,3 +382,4 @@ when using primary constructors:
 
 [language version]: /language/versioning
 [`parameter_assignments`]: /tools/linter-rules/parameter_assignments
+[potentially constant]: /resources/glossary#potentially-constant
--- a/src/data/glossary.yml
+++ b/src/data/glossary.yml
@@ -1433,17 +1433,22 @@
 
 - term: "Potentially constant"
   short_description: |-
-    An expression that can be evaluated at compile time if all constructor parameters
-    and type parameters of the enclosing constant constructor are replaced by constant values.
+    An expression that can be evaluated at compile time
+    if all constructor parameters of the enclosing constant constructor
+    are replaced by constant values,
+    and all type parameters are replaced by concrete types.
   long_description: |-
-    A **potentially constant expression** is an expression that is valid in contexts
-    where a constant is required, such as initializer lists of constant constructors,
-    even though it might not be a compile-time constant itself because it references
-    constructor parameters or type parameters.
+    A **potentially constant expression** is an expression
+    that is valid in contexts where a constant is required,
+    such as initializer lists of constant constructors,
+    even though it might not be a compile-time constant itself
+    because it references constructor parameters or type parameters.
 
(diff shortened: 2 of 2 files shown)


src/content/language/primary-constructors.md
```diff
@@ -247,9 +247,9 @@ Constant primary constructors have these important constraints:
     every instance variable must be `final`, can't be `late`,
     and must be definitely initialized by a declaring parameter,
     field initializer, or the primary constructor's initializer list.
-*   **Potentially constant initializers**:
+*   **[Potentially constant][] initializers**:
     The initializing expressions for each instance variable
-    must be able to be evaluated at compile time.
+    must be potentially constant.
```
sfshaza2, line 252: "Potentially constant"? Interesting verbage...

sfshaza2 (approved):
Very interesting! lgtm

conooi:
@parlough I know this has been merged already but can you take a quick look at the technical details?

eernstg:
I think one part of the text could be made more precise, I'll add a comment.

src/data/glossary.yml
```diff
@@ -1431,6 +1431,56 @@
...
 
+- term: "Potentially constant"
+  short_description: |-
+    An expression that can be evaluated at compile time
+    if all constructor parameters of the enclosing constant constructor
+    are replaced by constant values,
+    and all type parameters are replaced by concrete types.
+  long_description: |-
+    A **potentially constant expression** is an expression
+    that is valid in contexts where a constant is required,
```
eernstg, line 1442: It's tricky to get around this corner without scratches. ;-)

An expression which is potentially constant, but not constant, is an error if it occurs in a location where a constant expression is required.

I think it would be safer to say that it is sufficient for an expression to be potentially constant (that is, it can be constant as well, but that's not required) when it occurs in a `const` constructor initializer list, or it occurs in the initializing expression of a non-late instance variable in a class or enum with a `const` primary constructor.

With that, the reader will know exactly where a potentially constant expression can occur, and we haven't said anything like "a constant expression is required, but it doesn't have to be constant".

src/data/glossary.yml
```diff
@@ -1431,6 +1431,56 @@
...
+    such as initializer lists of constant constructors,
+    even though it might not be a compile-time constant itself
+    because it references constructor parameters or type parameters.
+
+    If all constructor parameters of the enclosing constant constructor
+    are replaced by constant values of the appropriate type,
+    and all type parameters are replaced by concrete types,
+    a potentially constant expression
+    becomes a compile-time constant expression.
```
eernstg, line 1451: We can't know that for sure. Being a potentially constant expression is a rather loose check that helps avoiding expressions which will certainly not be constant after substitution. However, the substitutions can also give rise to an expression which is not constant. For example:

```dart
class A {
  final int i;
  const new(dynamic d): i = d.length;
}

const a1 = A(''); // OK.
const a2 = A([]); // Error, `[].length` is not constant.
const a3 = A(true); // Error, `true` does not have a `length`.
```

We might be able to say that `a3` is irrelevant because we already said that we're only substituting expressions whose type is appropriate. However, we need to say that the substitutions _might_ yield a constant expression (otherwise, the constant evaluation of a constructor invocation that produced the substitutions is rejected), and this already covers `a3` without further explanations.

Next, the phrase 'concrete types' is slightly misleading. It could mean something like "a type derived from a declaration of a class which isn't abstract" (plus several other cases). The phrase which is used in specification documents is 'a constant type expression' (but that's not a very good phrase because it isn't an expression at all).

Perhaps 'a type which is known at compile time' could be used. This implies that it is not a type variable and it does not contain a type variable, and that's the core of it.

So, perhaps:

    If all constructor parameters of the enclosing constant constructor
    are replaced by constant values of the appropriate type,
    and all type parameters are replaced by types which are known at compile time,
    a potentially constant expression
    may become a compile-time constant expression.
    It may also become an expression which is not constant, and it may
    be a compile-time error for other reasons, in which case the constant
    constructor invocation that determined the replacements is flagged
    as an error.

eernstg (commented):
A couple of drive-by comments added.

PR #7373, opened 2026-07-10
Secure GitHub Actions workflow

conooi:
Secure GitHub Actions workflows:

restrict permissions and disable checkout credential persistence

- Change global workflow permissions to `contents: read` to restrict default token scope.
- Move write permissions down to the job level in `no-response.yaml` to resolve an excessive-permissions security warning.
- Set `persist-credentials: false` on all checkouts to prevent git token persistence on runners.
- Resolves all medium/high zizmor security findings.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,7 +13,8 @@ on:
     - cron: "0 11 * * 0"
 
 # Declare default permissions as read only.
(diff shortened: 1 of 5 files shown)


parlough (approved):
Thanks for tackling this @conooi! Looks good to me :)

PR #7376, opened 2026-07-13
Add prefer using abbreviates constructor syntax section

### LINKED ISSUE #7370: Add guideline for new in-body constructor syntax
### Page URL

https://dart.dev/effective-dart/usage#constructors

### Describe the problem

Dart 3.13 adds support for primary constructors. Bundled with that feature is also a new syntax for constructors inside the class body. Instead of repeating the class name, you can use `new` for a generative constructor and `factory` for a factory constructor:

```
Original Dart syntax                     New abbreviated syntax
---------------------------------------  --------------------------
LongClassName() {}                       new() {}
LongClassName.name() {}                  new name() {}
const LongClassName();                   const new();
const LongClassName.name();              const new name();
factory LongClassName() { ... }          factory() { ... }
factory LongClassName.name() { ... }     factory name() { ... }
```

The old syntax is still fully supported, but the long-term goal of the language team is to have the entire ecosystem move to the new syntax. Some (distant) future may remove support for the old syntax.

Because of that, we want to guide users to the new syntax. When 3.13 ships, it will include [a lint rule that does that](https://dart.dev/tools/linter-rules/unnecessary_type_name_in_constructor). I'd also like to add a corresponding guideline to "Effective Dart" that encourages the same thing and provides rationale.

I'm busy writing a blog post for 3.13 now and have some vacation coming up, but I'm happy to do this if no one else gets to it first.

conooi:
Fixes https://github.com/dart-lang/site-www/issues/7370

### DIFF
--- a/examples/misc/lib/effective_dart/usage_good.dart
+++ b/examples/misc/lib/effective_dart/usage_good.dart
@@ -580,3 +580,16 @@ Widget build(BuildContext context) {
 }
 
 // #enddocregion no-new
+
+// #docregion abbreviated-constructor-syntax
+class Logger {
+  final String name;
+
+  factory(String name) => _cache[name] ??= Logger._internal(name);
+
+  new _internal(this.name);
+  new fromJson(Map<String, Object?> json) : name = json['name'] as String;
+
+  static final Map<String, Logger> _cache = {};
+}
+// #enddocregion abbreviated-constructor-syntax
--- a/src/content/effective-dart/usage.md
+++ b/src/content/effective-dart/usage.md
@@ -1385,6 +1385,65 @@ class Point {
 }
 ```
 
+### PREFER using concise constructor syntax
+
+{% render 'linter-rule-mention.md', rules:'unnecessary_type_name_in_constructor' %}
+
+When declaring constructors inside the class body,
+use the concise constructor syntax instead of repeating the class name.
+Use `factory` for factory constructors
+and `new` for all other constructors.
+
+For a complete reference of the concise syntax forms,
+see the [concise constructor syntax table][].
+
+[concise constructor syntax table]: /language/constructors#concise-constructor-syntax
+
+:::note
+This guideline only applies to code at language version 3.13 or higher
+where the concise constructor syntax is available.
+:::
+
+<?code-excerpt "usage_good.dart (abbreviated-constructor-syntax)"?>
+```dart tag=good
+class Logger {
+  final String name;
+
+  factory(String name) => _cache[name] ??= Logger._internal(name);
+
+  new _internal(this.name);
+  new fromJson(Map<String, Object?> json) : name = json['name'] as String;
+
+  static final Map<String, Logger> _cache = {};
+}
+```
+
+<?code-excerpt "usage_bad.dart (abbreviated-constructor-syntax)"?>
+```dart tag=bad
+class Logger {
+  final String name;
+
+  factory Logger(String name) => _cache[name] ??= Logger._internal(name);
+
+  Logger._internal(this.name);
+  Logger.fromJson(Map<String, Object?> json) : name = json['name'] as String;
+
+  static final Map<String, Logger> _cache = {};
+}
+```
+
+Repeating the class name is redundant
+and makes refactoring harder if the class is renamed.
+Using `new` and `factory` keeps constructor declarations concise
+and consistent.
+
+:::note
+This guideline applies only to constructor *declarations*.
+For constructor *invocations*,
+see the [DON'T use `new`](#dont-use-new) guideline below.
+:::
+
+
 ### DON'T use `new`
 
 {% render 'linter-rule-mention.md', rules:'unnecessary_new' %}
--- a/src/content/effective-dart/usage.md
+++ b/src/content/effective-dart/usage.md
@@ -1440,7 +1440,7 @@ and consistent.
 :::note
 This guideline applies only to constructor *declarations*.
 For constructor *invocations*,
-see the [DON'T use `new`](#dont-use-new) guideline below.
+see the [DON'T use `new`](#dont-use-new) guideline.
 :::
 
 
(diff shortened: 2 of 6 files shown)


examples/misc/lib/effective_dart/usage_good.dart
```diff
@@ -580,3 +580,20 @@ Widget build(BuildContext context) {
...
+  const new constant();
+  const new constantNamed();
+}
+
+class Factory {
+  factory() => const Factory._();
+  factory named() => const Factory._();
+
+  const new _();
+}
```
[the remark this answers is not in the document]
munificent: @gemini-code-assist, you are incorrect. No dot after `new` or `factory` for a named constructor. Source: I designed this syntax.

src/content/effective-dart/usage.md
```diff
@@ -1385,6 +1385,65 @@ class Point {
 }
 ```
 
+### PREFER using concise constructor syntax
+
+{% render 'linter-rule-mention.md', rules:'unnecessary_type_name_in_constructor' %}
+
+When declaring generative or factory constructors inside the class body,
```
munificent: I would remove "generative or factory". I think most users don't know what "generative" means. (It's lingo we use on the language team, but almost no one else does.) Also, the new syntax applies to all constructors, so it's simpler to not distinguish.

src/content/effective-dart/usage.md
```diff
@@ -1385,6 +1385,65 @@ class Point {
...
+use the concise constructor syntax instead of repeating the class name.
+Use `new` for generative constructors
```
munificent: Instead of "generative" here, maybe:

> Use `factory` for factory constructors and `new` for all other constructors.

src/content/effective-dart/usage.md
```diff
@@ -1385,6 +1385,65 @@ class Point {
...
+  const new constant();
+  const new constantNamed();
+}
+
+class Factory {
+  factory() => const Factory._();
+  factory named() => const Factory._();
+
+  const new _();
+}
```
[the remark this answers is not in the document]
munificent: Incorrect.

src/content/effective-dart/usage.md
```diff
@@ -1385,6 +1385,65 @@ class Point {
...
```
munificent, line 1421: As much as possible, I try to come up with example code that looks concrete and real-world. People understandably get tired of reading "foo" and "bar" over and over. :)

I don't think the guideline needs to exhaustively cover all the kinds of constructors. Maybe just pick a couple and then point to a table that shows all of them? The proposal has [this table](https://github.com/dart-lang/language/blob/main/accepted/3.13/primary-constructors/feature-specification.md#abbreviations-of-in-body-constructor-declarations). We could potentially take that and incorporate it into [the language docs on constructors](https://dart.dev/language/constructors) (which will also need to be updated to the new syntax at some point).

I think of "Effective Dart's" goal as to be prescriptive about what you should do, but not necessarily be comprehensive about teaching every corner of the language.

src/content/effective-dart/usage.md
```diff
@@ -1385,6 +1385,65 @@ class Point {
 }
 ```
 
+### PREFER using concise constructor syntax
+
+{% render 'linter-rule-mention.md', rules:'unnecessary_type_name_in_constructor' %}
```
munificent, line 1390: We should probably mention somewhere in here that this guideline only applies to code at language version 3.13 or higher where the new syntax is available.

examples/misc/lib/language_tour/built_in_types.dart
```diff
@@ -106,7 +106,7 @@ void miscDeclAnalyzedButNotTested() {
 
   {
     // #docregion trailing-commas
-    var list = ['Car', 'Boat', 'Plane',];
+    var list = ['Car', 'Boat', 'Plane'];
```
munificent: Looks like the formatter removed the trailing comma that is supposed to be here. :)

You can turn off the formatter like:

```dart
    // dart format off
    // #docregion trailing-commas
    var list = ['Car', 'Boat', 'Plane',];
    // #enddocregion trailing-commas
    // dart format on
```

munificent (approved):
I love the new example code. :)

examples/misc/lib/language_tour/built_in_types.dart
conooi: oops fixed! along with all the other empty lines the formatter added 

Thanks!

src/content/effective-dart/usage.md
```diff
@@ -1385,6 +1385,65 @@ class Point {
...
+
+For a complete reference of the concise syntax forms,
+see the [concise constructor syntax table][].
+
+[concise constructor syntax table]: /language/constructors#concise-constructor-syntax
+
+:::note
+This guideline only applies to code at language version 3.13 or higher
+where the concise constructor syntax is available.
+:::
```
parlough: Consider using a version-note here and cross-linking "language version" in a way consistent with our other language version notes.

```markdown
:::version-note
This guideline only applies to libraries with a
[language version][] of at least 3.13,
where the concise constructor syntax is available.
:::

[language version]: /language/versioning
```

src/content/effective-dart/usage.md
```diff
@@ -1385,6 +1385,65 @@ class Point {
...
+
+Repeating the class name is redundant
+and makes refactoring harder if the class is renamed.
+Using `new` and `factory` keeps constructor declarations concise
+and consistent.
+
+:::note
+This guideline applies only to constructor *declarations*.
+For constructor *invocations*,
+see the [DON'T use `new`](#dont-use-new) guideline below.
```
parlough: Great idea to call out this out.

I'd consider dropping the "below" in case these entries shuffle or anything is introduced between these later.

```suggestion
see the [DON'T use `new`](#dont-use-new) guideline.
```

parlough (approved):
Looks great @conooi! Thanks for tackling this new entry :D

Two small suggestions, but otherwise seems ready to land.

Issue #7378, opened 2026-07-14
Document `dart run <package>@` syntax for dynamic remote execution


kevmoo:
The `dart run <package>@[<descriptor>]` syntax was recently introduced (unifying descriptor parsing, see dart-lang/sdk#62123) to allow dynamic, one-off execution of remote packages from pub.dev, similar to `npx` in the Node ecosystem.

Currently, this feature is completely undocumented on dart.dev. 

Specifically:
1. **`src/content/tools/dart-run.md`**: This page currently explicitly directs users to the legacy command: _"To run a program from an arbitrary location, use the pub global command."_ This should be updated to explain the `dart run pkg@` syntax.
2. **Tutorials / Guides**: High-traffic guides like `dart-doc.md` that currently instruct users to install one-off servers (e.g., `dart install dhttpd`) should be updated to recommend `dart run dhttpd@` to avoid polluting the user's `PATH` with temporary tools.

harshyadavDeveloper:
I've opened a PR that addresses this by documenting the `dart run <package>@[<descriptor>]` syntax in `dart-run.md` and updating the `dart-doc.md` guide to use the new one-off execution syntax for `dhttpd`.

Looking forward to any feedback.

mit-mit:
Hi @harshyadavDeveloper are you still working on this? If so, please address feedback in https://github.com/dart-lang/site-www/pull/7379

harshyadavDeveloper:
@mit-mit Yes i will reflect on the suggestions in some time. Sorry for the delay

PR #7379, opened 2026-07-15
Document dart run <package>@[<descriptor>] syntax

harshyadavDeveloper:
Contributes to #7378.

## Summary

This PR updates the documentation to cover the `dart run <package>@[<descriptor>]` syntax for executing packages directly from pub.dev.

### Changes

- Documented the `dart run <package>@[<descriptor>]` syntax in `dart-run.md`.
- Added an example of running an executable directly from pub.dev.
- Updated `dart-doc.md` to recommend `dart run dhttpd@` for one-off usage instead of globally activating `dhttpd`.


harshyadavDeveloper:
This PR documents the `dart run <package>@[<descriptor>]` syntax in the `dart run` documentation and updates the `dart doc` guide to recommend `dart run dhttpd@` for one-off usage instead of globally activating the package.

Feedback is welcome. Thanks!

kevmoo:
please update against latest, please!

src/content/tools/dart-run.md
```diff
@@ -3,13 +3,16 @@ title: dart run
...
 or in one of the dependencies of the current package—from the command line.
 This command provides functionality that was previously in `pub run`
 and the Dart VM tool.
-To run a program from an arbitrary location,
-use the [pub global](/tools/pub/cmd/pub-global) command.
+
+To run an executable from a package published on pub.dev without installing it
+globally, use the `dart run <package>@[<descriptor>]` syntax.
+If you use a tool frequently, consider installing it with the
+[pub global](/tools/pub/cmd/pub-global) command.
```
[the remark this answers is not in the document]
kevmoo: See https://github.com/dart-lang/site-www/issues/7464 - we should avoid talking about `dart pub global` going forward...

parlough, line 15: I think you can probably drop this whole paragraph since there's the new section that talks about the feature and we generally don't want to add new references to `dart pub global` anymore.

src/content/tools/dart-run.md
```diff
@@ -23,6 +26,24 @@ $ cd myapp
 $ dart run
 ```
 
+## Running an executable from pub.dev
```
parlough, line 29: We prefer to start headings with a verb in its infinitive form, so in this case "Run".

I'd also consider expanding this to "from a remote package" as the developer could technically have a different remote configured.


```suggestion
## Run an executable from a remote package
```

src/content/tools/dart-run.md
```diff
@@ -23,6 +26,24 @@ $ cd myapp
...
+
```
parlough, line 30: Consider adding a version note to the top of the section since this feature is still quite new.

```suggestion

:::version-note
Support for running remote package executables was introduced in Dart 3.12.
:::

```

src/content/tools/dart-run.md
```diff
@@ -23,6 +26,24 @@ $ cd myapp
...
+```console
+$ dart run dhttpd@
+```
+
+This downloads the package if necessary, runs its default executable,
+and reuses the cached package on subsequent runs.
+
+Use this form for one-off tools and temporary utilities.
+If you use a tool regularly, consider installing it with
+[`dart pub global activate`](/tools/pub/cmd/pub-global).
```
parlough, line 45: Feel free to be more prescriptive here.

Also reference `dart install` rather than `dart pub global activate`. So something like:


```suggestion
For a tool you use regularly,
install it with [`dart install`][] instead.
```

src/content/tools/dart-run.md
```diff
@@ -23,6 +26,24 @@ $ cd myapp
...
+without adding it to your current package or installing it globally.
+
+For example:
+
+```console
+$ dart run dhttpd@
+```
+
+This downloads the package if necessary, runs its default executable,
+and reuses the cached package on subsequent runs.
```
parlough, line 41: I think we should call out that this compiles the selected executable as well, not just running it.

src/content/tools/dart-run.md
```diff
@@ -23,6 +26,24 @@ $ cd myapp
 $ dart run
 ```
 
+## Running an executable from pub.dev
+
+You can run an executable directly from a package published on pub.dev
+without adding it to your current package or installing it globally.
```
parlough, line 32: Rather than structuring this introduction as "You can", consider restructuring it to a more direct suggestion, like "To do X, do Y". 

In this case, X would be running a remote package executable without adding it as a dependency or installing it, while Y would be specifying the package name followed by an `@` symbol.

src/content/tools/dart-run.md
```diff
@@ -23,6 +26,24 @@ $ cd myapp
...
+
+For example:
```
parlough, line 34: Consider adding an introduction/explanation of what the example does:

```
For example, the following command resolves `package:dhttpd` from pub.dev,
selects the latest stable version that's compatible with your SDK,
and runs the package's default executable:
```

src/content/tools/dart-run.md
```diff
@@ -23,6 +26,24 @@ $ cd myapp
...
+
+```console
+$ dart run dhttpd@
+```
+
```
parlough, line 39: Feel free to tackle this now or we can handle it in a follow-up PR, but after the basic example we should call out configuring the run executable.

The full syntax is.

```
dart run <PACKAGE>[:<EXECUTABLE>]@[<DESCRIPTOR>] [args]
```

Executable and descriptor are both optional, with the important part being the descriptor which is documented in https://dart.dev/tools/pub/dependencies#package-descriptors

parlough (commented):
Thanks for tackling this and sorry about the super long delay on this review.

Here's some feedback to consider and some additional functionality you might want to tackle as well.

Thanks again!

conooi:
Hi @harshyadavDeveloper, just checking in to see if you've had a chance to look over the feedback above? Let us know if you have any questions or need a hand with anything!

harshyadavDeveloper:
Hi, I’ve implemented the requested changes and updated the PR accordingly. Apologies for the delay in getting back to this. Thanks for the feedback and guidance.

src/content/tools/dart-run.md
sfshaza2: Not exactly the same as suggested, but perhaps close enough...

sfshaza2 (approved):
Thanks for this PR, @harshyadavDeveloper, and for following up with the feedback. Thanks also, @kevmoo and @parlough for your thorough reviews.

PR #7381, opened 2026-07-15
Add variable can't be promoted section

### LINKED ISSUE #7355: Please add http://dart.dev/go/non-promo-suspension page
https://github.com/dart-lang/sdk/commit/a41aa1847c2554a0f6d3a139db34315ae838107a fixes a soundness bug in Dart's type promotion logic. To aid users affected by the breakage, the error message contains a link to http://dart.dev/go/non-promo-suspension. Prior to shipping Dart 3.13, we should add a web page at this address, with a helpful explanation of the situation and recommendations for what to do.

Here are the technical details:

For the breakage to occur, the following conditions all need to be met:
- There needs to be a function (or method) that contains a write to a local variable (or parameter).
- That function needs to contain an inner function (or function expression) that type promotes that variable.
- Then, the inner function needs to suspend itself using an `await` expression (or `yield` statement), allowing the outer function to potentially continue running.
- And then, the inner function needs to try to make use of the local variable/parameter again.

Before https://github.com/dart-lang/sdk/commit/a41aa1847c2554a0f6d3a139db34315ae838107a, at the site of the final use of the local variable, it would have its promoted type. This was unsound, because it was possible that during the time the inner function was suspended, the outer function might have continued running and changed the value of the variable.

To fix the unsoundness, https://github.com/dart-lang/sdk/commit/a41aa1847c2554a0f6d3a139db34315ae838107a causes the promotion to be dropped at the time the inner function suspends itself.

Judging by Google's internal code base, it's quite rare for this situation to arise (which is why we were able to make this fix as a breaking change, rather than having to condition it on the language version).

However, it's possible for this fix to be overzealous; it can cause a promotion to be dropped even if the programmer can tell that it's safe to retain the promotion.

Here's an example of how this could happen. It's somewhat contrived but it's based on real-world code we've seen internally at Google:
```dart
import 'dart:async';

Future<void> example(String? extraInfo) async {
  extraInfo ??= 'No extra info'; // (3)
  unawaited(() async {
    log('Doing some asynchronous task...');
    log(extraInfo!); // (1)
    await longTask(); // (2)
    log('Done!');
    log(extraInfo); // (4)
  }());
}

Future<void> longTask() async {
  await Future<void>.delayed(Duration(seconds: 10));
}

void log(String s) {
  print(s);
}

void main() {
  unawaited(example(null));
}
```

In this example, the promotion happens at `(1)` (`extraInfo` is promoted from `String?` to `String`), but then at `(2)`, execution of the inner function is suspended. The compiler and analyzer reason that while the inner function is suspended, the outer function might reach `(3)`, writing a value to `extraInfo` that invalidates the promotion. So, to be on the safe side, they drop the promotion. Consequently, when execution reaches `(4)`, the type of `extraInfo` is `String?`, which is not compatible with the `log` function, so the user sees a compile-time error.

In this particular case, the programmer can tell that code is safe for two reasons:
- Since there's no looping constructs in the outer function, at the time the inner function is suspended, `(3)` is guaranteed to have already been executed, and it won't execute again.
- Besides, the value written to `extraInfo` at `(3)` has type `String`, so even if it were possible for `(3)` to execute again, it would not cause problems with the type promotion.

When a situation like this occurs, the compiler (or analyzer) outputs a context message to let the user know that the promotion was lost due to a suspension. Example compiler output:
```
example.dart:10:9: Error: The argument type 'String?' can't be assigned to the parameter type 'String'.
    log(extraInfo);
        ^
example.dart:8:5: Context: Variable 'extraInfo' could not be promoted due to an 'await' or 'yield'.
Try checking the type of the variable after the 'await' or 'yield'.  See http://dart.dev/go/non-promo-suspension
    await longTask();
    ^
```

Example screenshot from the analyzer (as seen in VSCode):
Image

The solution is to add an extra null check:
```dart
    log(extraInfo!); // (4)
```

conooi:
https://github.com/dart-lang/sdk/commit/a41aa1847c2554a0f6d3a139db34315ae838107a fixes a soundness bug in Dart's type promotion logic. To aid users affected by the breakage, the error message contains a link to http://dart.dev/go/non-promo-suspension. 

Add a web page at this address, with a helpful explanation of the situation and recommendations for what to do.

Resolves https://github.com/dart-lang/site-www/issues/7355


src/content/tools/non-promotion-reasons.md
```diff
@@ -1211,4 +1211,118 @@ void f(int? i, int? j) {
...
 ```
 
+### Variable can't be promoted due to an 'await' or 'yield' {:#suspension}
+
+**The cause:**
+Flow analysis drops type promotion when execution suspends
+across an `await` expression or `yield` statement.
+
+For this promotion failure to occur,
+all of the following conditions are met:
```
stereotype441: Would "all of the following conditions must be met" be clearer?

src/content/tools/non-promotion-reasons.md
```diff
@@ -1211,4 +1211,118 @@ void f(int? i, int? j) {
...
+   letting the outer function continue execution.
+1. The inner function attempts to use the local variable again
+   after the suspension.
+
+In earlier releases, the final access of the local variable
+retained its promoted type.
+However, that behavior was unsound:
+while the inner function is suspended,
+the outer function can continue running and modify the variable's value.
+To guarantee soundness,
```
stereotype441: Consider adding the text "Starting in Dart 3.13" to contrast with "In earlier releases" above. I.e. "Starting in Dart 3.13, to guarantee soundness, flow analysis drops the promotion..."

src/content/tools/non-promotion-reasons.md
```diff
@@ -1211,4 +1211,118 @@ void f(int? i, int? j) {
...
+
+<?code-excerpt "non_promotion/lib/non_promotion.dart (suspension)" replace="/extraInfo!/[!extraInfo!]/g"?>
+```dart tag=good
+import 'dart:async';
+// ···
+Future<void> example(String? extraInfo) async {
+  extraInfo ??= 'No extra info';
+  unawaited(() async {
+    log('Doing some asynchronous task...');
+    log([!extraInfo!]);
```
stereotype441: In the preview (https://dart-dev--pr7381-non-promo-suspension-zxtrs6in.web.app/tools/non-promotion-reasons#suspension), this is rendering as `log(extraInfo);`. This is incorrect; it should be `log(extraInfo!);`.

src/content/tools/non-promotion-reasons.md
```diff
@@ -1211,4 +1211,118 @@ void f(int? i, int? j) {
...
+    await longTask();
+    log('Done!');
+    log([!extraInfo!]);
```
stereotype441: Same issue here.

stereotype441 (commented):
Thanks Connie! This all looks great. I left a couple of minor suggestions about how to phrase things; you can take them or leave them as you see fit.

The only comments that I care deeply about are the comments on lines 1308 and 1311 (about `log(extraInfo);` vs `log(extraInfo!);`).

stereotype441 (approved):
lgtm, thanks!

src/content/tools/non-promotion-reasons.md
```diff
@@ -1211,4 +1211,119 @@ void f(int? i, int? j) {
...
+import 'dart:async';
+
+Future<void> example(String? extraInfo) async {
+  extraInfo ??= 'No extra info'; // (3)
+  unawaited(() async {
+    log('Doing some asynchronous task...');
+    log(extraInfo!); // (1)
+    await longTask(); // (2)
+    log('Done!');
+    log(extraInfo); // (4) ERROR
```
parlough: Consider reordering these to be in source order. Otherwise readers of the following text have to keep in mind the different order of the source and these labels. Unless there's another reason behind the order?

Also, can the current `(1)` be dropped? Or should it be referenced in the text? Seems like it's not currently.

src/content/tools/non-promotion-reasons.md
```diff
@@ -1211,4 +1211,119 @@ void f(int? i, int? j) {
...
 ```
 
+### Variable can't be promoted due to an 'await' or 'yield' {:#suspension}
+
+**The cause:**
+Flow analysis drops type promotion when execution suspends
+across an `await` expression or `yield` statement.
+
+For this promotion failure to occur,
+all of the following conditions must be met:
```
parlough: "All of the" makes it seem like they're separate, but if I'm understanding correctly, it's an ordered relationship. I believe simplifying to just "the" makes this a bit clearer:

```suggestion
For this promotion failure to occur,
the following conditions must be met:
```

Or we could explicitly say something more explicit like "The following sequence causes this promotion failure:". Whatever you think's best!

src/content/tools/non-promotion-reasons.md
```diff
@@ -1211,4 +1211,119 @@ void f(int? i, int? j) {
...
+after the suspension point.
+
+Add an extra null check or `!`
+at the post-suspension call site:
+
+<?code-excerpt "non_promotion/lib/non_promotion.dart (suspension)" replace="/(Done!'\);\n +log\()extraInfo!/$1[!extraInfo!!]/g"?>
+```dart tag=good
+import 'dart:async';
+// ···
+Future<void> example(String? extraInfo) async {
```
parlough: You can set `plaster=""` to keep this snippet a bit more focused and consistent (removing the ... comment).

````suggestion
<?code-excerpt "non_promotion/lib/non_promotion.dart (suspension)" plaster="" replace="/(Done!'\);\n +log\()extraInfo!/$1[!extraInfo!!]/g"?>
```dart tag=good
import 'dart:async';

Future<void> example(String? extraInfo) async {
````

File: src/content/tools/non-promotion-reasons.md [hunk 8]
```diff
@@ -1211,4 +1211,119 @@ void f(int? i, int? j) {
...
+```
+
+**Solution:**
+
+To fix this compile-time error,
+re-verify the variable's type or add an explicit null check
+after the suspension point.
+
+Add an extra null check or `!`
+at the post-suspension call site:
```
parlough: It seems the second paragraph here restates a lot of the preceding one. Can we combine or simplify them?

Maybe this is too long, but I gave it a shot:

```markdown
Re-verify the variable's non-null type, add an explicit not-null check,
or use a not-null assertion (`!`) after the suspension point:
```

src/content/tools/non-promotion-reasons.md
```diff
@@ -1211,4 +1211,119 @@ void f(int? i, int? j) {
...
+
+1. A function or method contains a write to a local variable or parameter.
+1. That function contains an inner function or closure
+   that promotes the variable.
+1. The inner function suspends itself using `await` or `yield`,
+   letting the outer function continue execution.
+1. The inner function attempts to use the local variable again
+   after the suspension.
+
+In earlier releases, the final access of the local variable
```
parlough: Since this change isn't language versioned, consider clarifying this is talking about SDK releases, not language versions.


```suggestion
In earlier releases of the SDK, the final access of the local variable
```

parlough: What is meant by "final" here? The usage mentioned in final entry of the earlier list or the actual final usage of the variable in the function?

If I'm understanding correctly, it's referring to any/all accesses to the variable after the inner function resumes execution. If so, I'd make that more more explicit.

\cc @stereotype441 can you confirm?

src/content/tools/non-promotion-reasons.md
```diff
@@ -1211,4 +1211,119 @@ void f(int? i, int? j) {
 }
 ```
 
+### Variable can't be promoted due to an 'await' or 'yield' {:#suspension}
```
parlough: Since every section on this page is about variables that can't be promoted, consider using a tighter header like the other entries. Perhaps something like:

```suggestion
### Promotion lost across an `await` or `yield` {:#suspension}
```

parlough (approved):
Love it @conooi! Overall, looks ready to land. Just some optional clarity and consistency improvements to consider. Thank you!

src/content/tools/non-promotion-reasons.md
stereotype441: Good point. What is meant here is "the access to the variable in item 4 of the above list"

src/content/tools/non-promotion-reasons.md
conooi: Updated the header to 'Promotion lost across an `await` or `yield`'. Thanks!

src/content/tools/non-promotion-reasons.md
conooi: Updated to clarify that this applies to earlier SDK releases and specifically refers to using the variable after the suspension point (item 4). Thanks!

src/content/tools/non-promotion-reasons.md
conooi: Combined and simplified the solution text as suggested. Thanks!

src/content/tools/non-promotion-reasons.md
conooi: Updated to 'the following conditions must be met:'. Thanks!

src/content/tools/non-promotion-reasons.md
conooi: Reordered code example annotations (1), (2), (3) to be sequential in source order and updated the explanation text. Thanks!

src/content/tools/non-promotion-reasons.md
conooi: Added plaster="" to keep the snippet focused. Thanks!

src/content/tools/non-promotion-reasons.md
conooi: Further updated the phrasing to 'using the local variable after the suspension' for extra clarity. Thanks!

src/content/tools/non-promotion-reasons.md
```diff
@@ -1211,4 +1211,116 @@ void f(int? i, int? j) {
...
+1. The inner function attempts to use the local variable again
+   after the suspension.
+
+In earlier SDK releases, using the local variable after the suspension
+retained its promoted type.
+However, that behavior was unsound:
+while the inner function is suspended,
+the outer function can continue running and modify the variable's value.
+Starting in Dart 3.13, to guarantee soundness,
+flow analysis drops the promotion when the inner function suspends.
```
eernstg, line 1240: It seems appropriate to have this explanation at this time, but after a while it will be less helpful because those older versions of the language are not being used (or even remembered in detail) any more.

Is there a mechanism which would serve to notify someone that this paragraph should then be deleted?

eernstg (commented):
Looks good! I added a single drive-by comment about the new section.

I'll leave it to the site-www team to mark the PR as approved when needed.

src/content/tools/non-promotion-reasons.md
conooi: Good point! We can add an inline comment (such as `<!-- TODO: Remove historical note in future release audit -->`) or filing a doc maintenance issue. 

What do you think @parlough?

parlough: Generally I review `:::version-note` asides as well as "Dart 2.x" and "Dart 3.x" mentions every few releases to determine if each should be left as is, updated, or removed. Perhaps that can be part of your standard release process as well?

If you want this to stand on its own for longer, you could rephrase it to be documented agnostic of the comparison to earlier version and just leave a version note. Keeping it as is and revisiting it later is also fine though.

parlough (approved):
Thanks for those adjustments @conooi! Looks good to me.

src/content/tools/non-promotion-reasons.md
conooi: Yes, will do! I think we can keep as it is for now and revisit this.

PR #7387, opened 2026-07-18
Add glossary terms

conooi:
List of terms added:

Dot shorthand
Extension type
Primary constructor
Private named parameter
WebAssembly (Wasm)

### DIFF
--- a/src/data/glossary.yml
+++ b/src/data/glossary.yml
@@ -689,6 +689,44 @@
     - "pub"
     - "packages"
 
+- term: "Dot shorthand"
+  short_description: |-
+    A syntax shorthand that lets you omit the type name
+    when accessing enums, constructors, or static members.
+  long_description: |-
+    A **dot shorthand** is a language feature that allows for more concise
+    code by omitting the type name when accessing static members,
+    constructors, or enum values.
+
+    This feature is only allowed when the expected type of the expression
+    can be uniquely inferred from the surrounding context (the context type).
+
+    For example, instead of writing `Status.running`, you can write `.running`
+    if the receiver type is already known to be `Status`:
+
+    ```dart
+    enum Status { running, stopped }
+
+    void updateStatus(Status s) {}
+
+    void main() {
+      updateStatus(.running); // Resolves to Status.running
+    }
+    ```
+  related_links:
+    - text: "Dot shorthands"
+      link: "/language/dot-shorthands"
+      type: "doc"
+    - text: "Context type"
+      link: "#context-type"
+      type: "term"
+  labels:
+    - "language"
+    - "syntax"
+  alternate:
+    - "Static access shorthand"
+    - "dot-shorthand"
+
 - term: "Entrypoint"
   short_description: |-
     A Dart library that is directly invoked by a Dart implementation.
@@ -746,6 +784,44 @@
   labels:
     - "pub"
     - "packages"
+
+- term: "Extension type"
+  short_description: |-
+    A compile-time wrapper that provides a zero-cost abstraction
+    over an underlying type.
+  long_description: |-
+    An **extension type** allows you to wrap an existing type with new APIs
+    without incurring the runtime performance or memory overhead of a
+    traditional wrapper class.
+
+    Because the extension type is resolved entirely at compile time, the
+    compiled code behaves as if you used the underlying type directly. This is
+    highly useful for creating distinct type-safe representations of existing
+    data (such as database IDs or unit measurements like meters and feet).
+
+    For example, you can wrap an `int` to represent meters:
+
+    ```dart
+    extension type Meters(int value) {
+      String get label => '${value}m';
+      Meters operator +(Meters other) => Meters(value + other.value);
+    }
+
+    void main() {
+      var distance = Meters(100);
+      print(distance.label); // Prints: 100m
+    }
+    ```
+  related_links:
+    - text: "Extension types"
+      link: "/language/extension-types"
+      type: "doc"
+  labels:
+    - "language"
+    - "type system"
+  alternate:
+    - "extension-type"
+    - "zero-cost wrapper"
 
 - term: "Function"
   short_description: |-
@@ -1511,6 +1587,73 @@
     - "understanding diagnostics"
   alternate:
     - "potentially nullable"
+
+- term: "Primary constructor"
+  short_description: |-
+    A concise way to declare a class's fields
+    and its main constructor in a single line.
+  long_description: |-
+    A **primary constructor** is a syntax shorthand that declares a class's
+    fields and its main constructor directly in the class header.
+
+    This reduces the boilerplate of declaring fields, passing parameters,
+    and assigning them in the constructor body.
+
+    For example, declaring a parameter with the `var` or `final` modifier
+    implicitly creates that field on the class:
+
+    ```dart
+    class Point(var int x, var int y);
+    ```
+  related_links:
+    - text: "Primary constructors"
+      link: "/language/primary-constructors"
+      type: "doc"
+  labels:
+    - "language"
+    - "constructors"
+    - "syntax"
+  alternate:
+    - "primary-constructor"
+
+- term: "Private named parameter"
+  short_description: |-
+    A named constructor parameter that initializes a private field
+    using initializing formal syntax.
+  long_description: |-
+    A **private named parameter** is a constructor feature that relaxes the
+    restriction against private named parameters for initializing formals
+    (`this._field`).
+
+    By declaring a private field as a named parameter in the constructor
+    signature, the compiler automatically exposes the public parameter name
+    (omitting the leading underscore) at the call site while still initializing
+    the private field directly.
+
+    For example, you can write:
+
+    ```dart
+    class Point {
+      final double _x;
+      final double _y;
+
+      // Callers use public names 'x' and 'y'
+      Point.named({required this._x, required this._y});
+    }
+
+    void main() {
+      var p = Point.named(x: 10.0, y: 20.0); // Initializes _x and _y
+    }
+    ```
+  related_links:
+    - text: "Private named parameters"
+      link: "/language/constructors#private-named-parameters"
+      type: "doc"
+  labels:
+    - "language"
+    - "constructors"
+  alternate:
+    - "private-named-parameters"
 
 - term: "Public library"
   short_description: |-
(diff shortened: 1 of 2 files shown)


src/data/glossary.yml
```diff
@@ -689,6 +689,44 @@
...
+  short_description: |-
+    A syntax shorthand that lets you omit the type name
+    when accessing enums, constructors, or static members.
+  long_description: |-
+    A **dot shorthand** is a language feature that allows for more concise
+    code by omitting the type name when accessing static members,
+    constructors, or enum values.
+
+    This feature is only allowed when the expected type of the expression
+    can be uniquely inferred from the surrounding context (the context type).
```
parlough: Consider cross linking "context type" to its glossary entry.

src/data/glossary.yml
```diff
@@ -689,6 +689,44 @@
...
+
+    For example, instead of writing `Status.running`, you can write `.running`
+    if the receiver type is already known to be `Status`:
```
```suggestion
    if the context type is already known to be `Status`:
```

src/data/glossary.yml
```diff
@@ -689,6 +689,44 @@
...
+
+    ```dart
```
parlough: Since the snippet is a bit longer, consider highlighting the relevant line:


````suggestion
    ```dart highlightLines=6
````

File: src/data/glossary.yml [hunk 4]
```diff
@@ -1512,6 +1588,73 @@
...
+    - "primary-constructor"
+
+- term: "Private named parameter"
+  short_description: |-
+    A named constructor parameter that initializes a private field
+    using initializing formal syntax.
+  long_description: |-
+    A **private named parameter** is a constructor feature that relaxes the
+    restriction against private named parameters for initializing formals
+    (`this._field`).
```
parlough: I'd recommend updating the entry to document private named parameters from a fresh perspective, rather than in relation to or as a relaxation (of a no longer existing restriction).

src/data/glossary.yml
```diff
@@ -1512,6 +1588,73 @@
...
+    - "language"
+    - "constructors"
+    - "syntax"
+  alternate:
+    - "primary-constructor"
+
+- term: "Private named parameter"
+  short_description: |-
+    A named constructor parameter that initializes a private field
+    using initializing formal syntax.
```
parlough: When hovering this term and seeing the definition, readers might not know what "initializing formal syntax" means. Consider adjusting the short description to be as simple as possible and most relevant to readers outside this context.

Would something like this work/be accurate?

```
    A named constructor parameter with a private name in the declaration
    and a public name at call sites.
```

src/data/glossary.yml
```diff
@@ -1512,6 +1588,73 @@
...
+    the private field directly.
+
+    For example, you can write:
+
+    ```dart
+    class Point {
+      final double _x;
+      final double _y;
+
+      // Callers use public names 'x' and 'y'
```
```suggestion
      // Callers use the public names 'x' and 'y'.
```

src/data/glossary.yml
```diff
@@ -1512,6 +1588,73 @@
...
+      Point.named({required this._x, required this._y});
+    }
+
+    void main() {
+      var p = Point.named(x: 10.0, y: 20.0); // Initializes _x and _y
```
parlough: To be consistent with the earlier comment format:

```suggestion
      var p = Point.named(x: 10, y: 20); // Initializes '_x' and '_y'.
```

src/data/glossary.yml
```diff
@@ -2332,6 +2475,34 @@
     - "contravariance"
     - "contravariant"
 
+- term: "WebAssembly (Wasm)"
+  short_description: |-
+    A binary instruction format (Wasm) that enables
```
```suggestion
    A binary instruction format that enables
```

src/data/glossary.yml
```diff
@@ -2332,6 +2475,34 @@
...
+    high-performance execution of code on web browsers.
+  long_description: |-
+    **WebAssembly** (commonly abbreviated to **Wasm**) is a portable binary
+    format designed as a compilation target for programming languages, enabling
+    close-to-native execution speeds on the web.
+
+    Dart supports compiling code directly to WebAssembly (using the
+    `dart compile wasm` tool). Compiling Dart to Wasm is especially
+    beneficial for Flutter web apps, improving frame rates, startup times,
+    and application size compared to standard JavaScript compilation.
```
parlough, line 2482: I'd consider softening this declaration as it might be conditional on details of their app, browser and machine types, etc.

```suggestion
    `dart compile wasm` tool). Compiling Dart to Wasm can be especially
    beneficial for Flutter web apps, improving frame rates, startup times,
    and application size compared to standard JavaScript compilation.
```

src/data/glossary.yml
```diff
@@ -1512,6 +1588,73 @@
   alternate:
     - "potentially nullable"
 
+- term: "Primary constructor"
+  short_description: |-
+    A concise way to declare a class's fields
+    and its main constructor in a single line.
```
parlough: It might not end up on a single line due to formatting.

Also primary constructors can be used for other declarations like enums and extension types.

Consider focusing the short description on it being a concise constructor declaration in the header of a class, enum,  or extension type and discussing declaring parameters in the long description.

src/data/glossary.yml
```diff
@@ -1512,6 +1588,73 @@
...
+  long_description: |-
+    A **primary constructor** is a syntax shorthand that declares a class's
+    fields and its main constructor directly in the class header.
```
parlough: Same here, this needs to be expanded beyond just classes.

src/data/glossary.yml
```diff
@@ -747,6 +785,44 @@
     - "pub"
     - "packages"
 
+- term: "Extension type"
```
parlough, line 791: I'm not sure we should try to define extension type in the glossary. It's a finicky topic that has a lot of nuance and it has its own standalone doc already. I worry to make this one fully accurate, we'd have to copy a lot from the primary page, resulting in potential mismatch.

Would you be fine with dropping this one? Or perhaps we can just have a short definition or something like that?

parlough: On the other hand, ["representation type"](https://dart.dev/language/extension-types#:~:text=type%2C%20called%20the-,representation%20type,-.%20When%20defining%20the) might be a more focused term to consider for inclusion in the future.

parlough (commented):
Thanks for tackling these entries @conooi! I left some initial comments, but for the language entries, I'd consider also getting a review from @eernstg or @kallentu.

src/data/glossary.yml
```diff
@@ -689,6 +689,44 @@
     - "pub"
     - "packages"
 
+- term: "Dot shorthand"
+  short_description: |-
+    A syntax shorthand that lets you omit the type name
+    when accessing enums, constructors, or static members.
```
eernstg: Perhaps:
```suggestion
    when accessing constructors or static members, including enum values.
```
The phrase "when accessing enums" is somewhat ambiguous&mdash;my immediate reaction was "what does that mean?", and it might mean things like "calling a method on an enum value", which is not intended.

Next, the phrase "accessing constructors or static members" is actually precisely what dot shorthands can do, and it is also true that the values of an `enum` declaration are (constant) static variables, even though the syntax doesn't clearly show this. I'm using "including" to imply that enum values are indeed static members.

It's not easy to find the right words when we only have very few words to spend, and it might still be somewhat confusing (e.g., some developers might not think that enum values are static members), but I hope the main text and the examples will help.

src/data/glossary.yml
eernstg: Sounds like a good idea to have just a very short section here, mainly to show the link to the existing page on this topic.

src/data/glossary.yml
```diff
@@ -747,6 +785,44 @@
...
+    extension type Meters(int value) {
+      String get label => '${value}m';
+      Meters operator +(Meters other) => Meters(value + other.value);
+    }
+
+    void main() {
+      var distance = Meters(100);
+      print(distance.label); // Prints: 100m
+    }
+    ```
```
eernstg: OK, perhaps that's not considered useful, but the body of this section could be empty, which would presumably mean that it would just contain the links below.

src/data/glossary.yml
```diff
@@ -1512,6 +1588,73 @@
   alternate:
     - "potentially nullable"
 
+- term: "Primary constructor"
+  short_description: |-
+    A concise way to declare a class's fields
+    and its main constructor in a single line.
+  long_description: |-
```
eernstg, line 1581: This could then be an empty body again, just showing the link to the separate section about this topic.

src/data/glossary.yml
eernstg: I think it's useful to omit the parameter kind here. It could be either initializing or declaring, and it gets too wordy to explain both.

src/data/glossary.yml
eernstg: Perhaps:
```suggestion
    A **private named parameter** is a named parameter whose name
    is private. It is only allowed when the parameter is initializing (e.g.,
    `this._x`) or declaring (e.g., `final T _x` or `var T _x` where `T`
    is a type). They only occur in a few places: Initializing parameters
    are only allowed in constructors, and declaring parameters are only
    allowed in primary constructors.
```

src/data/glossary.yml
```diff
@@ -1512,6 +1588,73 @@
...
+    using initializing formal syntax.
+  long_description: |-
+    A **private named parameter** is a constructor feature that relaxes the
+    restriction against private named parameters for initializing formals
+    (`this._field`).
+
+    By declaring a private field as a named parameter in the constructor
+    signature, the compiler automatically exposes the public parameter name
+    (omitting the leading underscore) at the call site while still initializing
+    the private field directly.
```
eernstg: Perhaps:
```suggestion
    An initializing parameter will initialize the instance variable with the
    given private name. A declaring parameter will both implicitly declare
    and initialize an instance variable with the given private name. At
    call sites, arguments are passed to private named parameters by
    specifying the corresponding public name.
```

eernstg (commented):
Looks good! I added a few drive-by comments. I'll leave it to @conooi and @parlough to approve the PR when it's ready.

src/data/glossary.yml
```diff
@@ -689,6 +689,44 @@
...
+    This feature is only allowed when the expected type of the expression
+    can be uniquely inferred from the surrounding context (the context type).
+
+    For example, instead of writing `Status.running`, you can write `.running`
+    if the receiver type is already known to be `Status`:
+
+    ```dart
+    enum Status { running, stopped }
+
+    void updateStatus(Status s) {}
```
kallentu: Just a nit, but maybe make this some useable code instead of an empty body to replicate a real example.

src/data/glossary.yml
```diff
@@ -1512,6 +1588,73 @@
...
+    A concise way to declare a class's fields
+    and its main constructor in a single line.
+  long_description: |-
+    A **primary constructor** is a syntax shorthand that declares a class's
+    fields and its main constructor directly in the class header.
+
+    This reduces the boilerplate of declaring fields, passing parameters,
+    and assigning them in the constructor body.
+
+    For example, declaring a parameter with the `var` or `final` modifier
```
kallentu: Remove the "For example," part. I think it reads better as just a description of declaring parameters and what to do.

kallentu (approved):

kallentu:
I've approved it from my end, but I'll also leave final approval to @parlough 
Thanks y'all! Looking good.

conooi:
@eernstg @parlough do you guys mind taking another look at this? Thanks!

eernstg (approved):
Looks great!

PR #7389, opened 2026-07-19
Clarify return type guidance for operator []=

### LINKED ISSUE #7252: In "DON'T specify a return type for a setter", note that "[]=" is also a setter
### Page URL

https://dart.dev/effective-dart/design#dont-specify-a-return-type-for-a-setter

### Describe the problem

A user-defined `[]=` operator behaves like a setter when it comes to return type: If you write a return type, the only type you can write is `void`. Therefore, there's no point in writing a return type.

Thus, this guideline applies equally well to `[]=` operators and should note that. See https://github.com/dart-lang/sdk/issues/62621 for more context.

enlorik:
Clarifies that `operator []=` follows the same return-type guidance as setters.
Adds matching bad and good examples and keeps their source excerpts in sync.

Fixes #7252


conooi (approved):
Thanks for the contribution @enlorik! This is a great addition following Bob's issue. I pushed a quick commit to refresh the code excerpts so the CI check passes. 

LGTM!

PR #7405, opened 2026-07-24
Fix overly specific type promotion description (#7395)

### LINKED ISSUE #7395: Overly specific type promotion characterization
### Page URL

https://dart.dev/tools/non-promotion-reasons

### Page source

https://github.com/dart-lang/site-www/blob/main/src/content/tools/non-promotion-reasons.md

### Describe the problem

At the beginning of this page, it says

> [Type promotion](https://dart.dev/null-safety/understanding-null-safety#type-promotion-on-null-checks) occurs when flow analysis can soundly confirm a variable with a [nullable type](https://dart.dev/null-safety/understanding-null-safety#non-nullable-and-nullable-types) is not null, and that it will not change from that point on.

This description outlines a number of situations where promotions can occur, but they may also occur when each of those things are false:

- A type promotion may change the type of a local variable `v` to any subtype (e.g., from `num` to `int`), not just from a nullable type to the corresponding non-nullable type. For example: `num n = 1; n as int;` promotes `n` to `int`.
- A local variable can be promoted in spite of the fact that it's changed later on. For example:

```dart
void main() {
  num n = 1; 
  n as int; // `n` is promoted to `int`.
  n = 1.5; // `n` is demoted to `num` and immediately promoted to `double`.
}
```

### Expected fix

Perhaps:

```markdown
[Type promotion][] occurs when flow analysis can soundly confirm
that a variable with a type `T` actually refers to an object of a subtype
`S` of `T`. The type of this variable will then be promoted to `S`.
As a special case, the variable may have a [nullable type][], and
flow analysis may confirm that it is *not null*. The variable is then
promoted to have the corresponding non-nullable type.
```

### Additional context

_No response_

conooi:
Fixes https://github.com/dart-lang/site-www/issues/7395

### DIFF
--- a/src/content/tools/non-promotion-reasons.md
+++ b/src/content/tools/non-promotion-reasons.md
@@ -6,8 +6,11 @@ description: >-
 ---
 
 [Type promotion][] occurs when flow analysis can soundly confirm
-a variable with a [nullable type][] is *not null*, and
-that it will not change from that point on.
+that a variable with a type `T` actually refers to an object of a subtype
+`S` of `T`. The type of this variable will then be promoted to `S`.
+As a special case, the variable may have a [nullable type][], and
+flow analysis may confirm that it is *not null*. The variable is then
+promoted to have the corresponding non-nullable type.
 Many circumstances can weaken a type's soundness,
 causing type promotion to fail.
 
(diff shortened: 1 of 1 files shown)


eernstg (approved):
LGTM, including the adjustment suggested by Gemini!

PR #7407, opened 2026-07-24
Use analyzer not CFE codes for diagnostic pages

### LINKED ISSUE #7406: Confusion in diagnostic documentation generation
### Page URL

https://dart.dev/tools/diagnostics

### Page source

_No response_

### Describe the problem

The diagnostic documentation page has an entry for `record_literal_one_positional_field_no_trailing_comma`, which is defined here: https://github.com/dart-lang/sdk/blob/main/pkg/_fe_analyzer_shared/messages.yaml#L34. The problem is that this is the name used by the CFE, while the analyzer uses the name `record_literal_one_positional_no_trailing_comma` (note the absence of `_field`).

To the best of my knowledge, the user will never see a link to the CFE's name, but will see a link to the analyzer's name, which will give them a 'page not found' error.

### Expected fix

It would be great if the generator could use the analyzer's name when generating documentation from `_fe_analyzer_shared/messages.yaml`.

### Additional context

@parlough 

parlough:
Fixes https://github.com/dart-lang/site-www/issues/7406

**Staged diagnostic page now with correct name:** https://dart-dev--pr7407-fix-analyzer-codes-qrwnx43t.web.app/tools/diagnostics/record_literal_one_positional_no_trailing_comma

### DIFF
--- a/src/data/diagnostics.json
+++ b/src/data/diagnostics.json
@@ -99,7 +99,7 @@
   {
     "id": "abstract_super_member_reference",
     "description": "_The {0} '{1}' is always abstract in the supertype._",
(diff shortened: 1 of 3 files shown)


bwilkerson (approved):
I'm not sure I understand what I'm looking at, not being familiar with this code.

I'm guessing that the changes to `src/data/diagnostics.json` is the generated output and doesn't really require any review.

The other changes look reasonable.

PR #7419, opened 2026-07-30
Add note about deferred loading to dart.dev/web/wasm

mkustermann:
We have now experimental deferred loading support and should mention it in the documentation.

### DIFF
--- a/src/content/web/wasm.md
+++ b/src/content/web/wasm.md
@@ -43,8 +43,8 @@ restrictions:
 
 :::note
 By default, compiled applications do not use [deferred loading][].
-Deferred loading in Wasm is experimental and can be enabled using
-the hidden `--enable-deferred-loading` flag:
+Deferred loading in Wasm is experimental. To enable it, use the
+`--enable-deferred-loading` flag:
 
 ```console
 $ dart compile wasm --enable-deferred-loading ...
@@ -64,6 +64,8 @@ const instantiatedApp = await app.instantiate({}, {
 });
 instantiatedApp.invokeMain();
 ```
+
+This is available from the 3.13 Dart stable release.
 :::
 
 ### Supported packages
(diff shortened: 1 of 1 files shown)


biggs0125 (approved):

kevmoo:
Do we want to advertise this before we make the flag public?
Also: what version was this added? We should tell folks the min Dart version where it's available.

mkustermann:
> Do we want to advertise this before we make the flag public?

Someone using dart2js with deferred loading today **wouldn't even think to deploy their app in wasm mode** because it would bundle everything in the main module (possibly much bigger than dart2js main part). So I think it makes sense to surface that wasm also has this support now.

There's no known bugs in it atm - would love to see more people trying it out and possibly provide feedback
(We did have 2 users who actually reported bugs already, which was great)

Though there may be small things changing, we may:  need to specify the `loadids.json` format somewhere, finalize the embedder api, possibly have a default for embedder api module loading, etc. Furthermore, it would be nice to have the "works great in production" signal from our big g3 app we experiment atm before declaring it stable.

> Also: what version was this added? We should tell folks the min Dart version where it's available.

I believe with the release of the new stable it will be available on all branches (main/dev/beta/stable), not sure if it makes sense to mention version numbers specifically.

src/content/web/wasm.md
```diff
@@ -41,6 +41,31 @@ restrictions:
    contain a temporary workaround. For details, see
    [webdev issue 2206]({{site.repo.dart.org}}/webdev/issues/2296).
 
+:::note
+By default, compiled applications do not use [deferred loading][].
+Deferred loading in Wasm is experimental and can be enabled using
+the hidden `--enable-deferred-loading` flag:
```
conooi:
```suggestion
Deferred loading in Wasm is experimental. To enable it, use the
`--enable-deferred-loading` flag:
```

conooi (approved):
one suggestion but otherwise lgtm! thanks @mkustermann !

src/content/web/wasm.md
mkustermann: Done.

mkustermann:
> I believe with the release of the new stable it will be available on all branches (main/dev/beta/stable), not sure if it makes sense to mention version numbers specifically.

Okay, I've added now a note about the upcoming release:

_This experimental support will be available in the 3.13 dart stable release._

PR #7422, opened 2026-07-31
Update and clean up blog links

parlough:
- Update `medium.com/dartlang` and `blog.dart.dev` links to point to `/blog` instead.
- Add `{{site.flutter-blog}}` and update `medium.com/flutter` and `blog.flutter.dev` to use it.
- Use the new simplified slugs/paths of the articles.
- Correct fragments for their new header-based versions.

Contributes to https://github.com/flutter/website/issues/13507


conooi (approved):
Thank you, @parlough! 

Question: for these posts, do they need to follow semantic breaks?

parlough:
Thanks for taking a look :D

> Question: for these posts, do they need to follow semantic breaks?

For the blog posts? The existing posts didn't because they were imported mostly mechanically. For new posts, I would would personally consider and prefer it, but not require it. Up to you though!

conooi:
@parlough just for these old ones! I was going to comment and suggest semantic breaks for all the changes on these files but wanted to run it with you to see if it was necessary 😆

PR #7433, opened 2026-08-07
[pub] Add SemVer compatibility and breaking changes reference guide

mosuem:
This PR adds a comprehensive SemVer compatibility and breaking changes reference guide for Dart packages (similar to [Cargo's SemVer reference](https://doc.rust-lang.org/cargo/reference/semver.html)).

Related to discussion in https://github.com/dart-lang/sdk/issues/63970.

This is a first draft - AI generated and then manually reviewed and wordsmithed a bit.

The current state can be viewed at https://dart-dev--pr7433-semver-rules-doc-3cv6edzw.web.app/tools/pub/semver-rules

cc @sigurdm @brianquinlan @devmil

### DIFF
--- a/src/content/tools/pub/pubspec.md
+++ b/src/content/tools/pub/pubspec.md
@@ -180,8 +180,11 @@ Once that's been done, consider it hermetically sealed: you can't touch it
 anymore. To make more changes, you'll need a new version.
 
 When you select a version, follow [semantic versioning.][semantic versioning]
+For a reference on which API changes in Dart require major, minor, or patch
+version bumps, see [SemVer compatibility rules][semver-rules].
 
 [semantic versioning]: https://semver.org/spec/v2.0.0-rc.1.html
+[semver-rules]: /tools/pub/semver-rules
 
 ### Description
 
--- a/src/content/tools/pub/pubspec.md
+++ b/src/content/tools/pub/pubspec.md
@@ -181,7 +181,7 @@ anymore. To make more changes, you'll need a new version.
 
 When you select a version, follow [semantic versioning.][semantic versioning]
 For a reference on which API changes in Dart require major, minor, or patch
-version bumps, see [SemVer compatibility rules][semver-rules].
+version bumps, check out [SemVer compatibility rules][semver-rules].
 
 [semantic versioning]: https://semver.org/spec/v2.0.0-rc.1.html
 [semver-rules]: /tools/pub/semver-rules
--- a/src/content/tools/pub/semver-rules.md
+++ b/src/content/tools/pub/semver-rules.md
@@ -1,5 +1,5 @@
 ---
-title: SemVer compatibility and breaking changes
+title: API versioning guidelines
 breadcrumb: SemVer rules
 description: >-
   A comprehensive guide to what changes are breaking (major), non-breaking
@@ -79,9 +79,19 @@ Renaming a public symbol is functionally equivalent to removing the original nam
 
 Moving a declaration from the public `lib/` root to `lib/src/` (without an `export` in a public library file) makes it inaccessible to external packages, breaking all existing consumers.
 
-### MAJOR: Change the type of a top-level variable or constant
+### MAJOR: Widen or change the type of a `const` or `final` variable
 
-Changing the static type of a top-level variable or constant causes type errors for any callers expecting the previous type.
+Changing a `const` or `final` variable to a broader type (for example, `const int maxRetries` to `const num maxRetries`) or to an incompatible type breaks callers expecting members of the more specific type.
+
+### MINOR: Narrow the type of a `const` or `final` variable
+
+Narrowing the static type of a read-only `const` or `final` variable to a more specific subtype (for example, `const num timeout` to `const int timeout`, or `const Object key` to `const String key`) is backward-compatible. Because consumers only read the value, a more specific subtype satisfies all existing type expectations while providing additional type safety.
+
+### MAJOR: Change the type of a mutable (`var`) top-level variable
+
+For mutable variables (which can be both read from and written to), changing the static type in either direction is a breaking change:
+* Narrowing the type (e.g., `num` to `int`) breaks callers assigning a `double` into the variable.
+* Widening the type (e.g., `int` to `num`) breaks callers expecting `int` methods when reading the variable.
 
 ### MAJOR: Change a top-level variable from `var` to `final`
 
@@ -93,7 +103,41 @@ Making a mutable variable `final` breaks any consumer assigning values to that v
 
 Dart 3 introduced **class modifiers** (`final`, `base`, `interface`, `sealed`, and `mixin`), allowing package authors to explicitly define how classes can be used by consumers. The SemVer impact of adding or modifying members depends on these modifiers.
 
-### Unmodified classes (`class` and `abstract class`)
+### Changing modifiers on existing classes
+
+Changing the modifier on an already-published class alters what external packages are permitted to do with it:
+
+#### MAJOR: Change a concrete class to an abstract class
+
+Changing a concrete class to `abstract class` prevents external consumers from instantiating the class directly (`new MyClass()`), causing compile errors.
+
+#### MINOR: Change an abstract class to a concrete class
+
+Making an abstract class concrete enables external instantiation without breaking existing subclasses or implementers.
+
+#### MAJOR: Add the `base` modifier to an existing class
+
+Prevents external consumers from implementing the class (`implements MyClass`), breaking all existing external implementers.
+
+#### MAJOR: Add the `interface` modifier to an existing class
+
+Prevents external consumers from extending the class (`extends MyClass`), breaking all existing external subclasses.
+
+#### MAJOR: Add the `final` or `sealed` modifier to an existing class
+
+Prevents external consumers from extending, implementing, or mixing in the class, breaking all existing external subtypes.
+
+#### MAJOR: Remove the `mixin` modifier from a `mixin class`
+
+Prevents external consumers from using the class as a mixin (`with MyClass`).
+
+#### MINOR: Remove `base`, `interface`, or `final` from a class
+
+Loosens restrictions on consumers, granting new capabilities without invalidating existing code.
+
+---
+
+### Unmodified classes: `class` and `abstract class`
 
 An unmodified class in Dart allows external packages to construct, extend (`extends`), and implement (`implements`) it.
 
@@ -225,6 +269,31 @@ Restricting the superclass requirements (for example, from `on Object` to `on Wi
 
 ---
 
+## Type aliases (`typedef`)
+
+Type aliases (`typedef`) define names for function types, records, or existing class types.
+
+### MINOR: Add a new type alias
+
+Adding a new public `typedef` exposes a new type name without affecting existing code.
+
+### MAJOR: Remove or rename a type alias
+
+Any downstream code referencing the removed `typedef` name will fail to compile.
+
+### MAJOR: Change the aliased type in a type alias
+
+Changing the underlying type that a type alias references (for example, changing `typedef Callback = void Function(int);` to `typedef Callback = void Function(String);`) alters the expected type everywhere the alias is used, breaking consumers.
+
+### MAJOR: Replace a `typedef` with a class or vice versa
+
+Replacing a type alias with a class (or class with a type alias) of the same name is a breaking change:
+
+* **Function type alias $\rightarrow$ class:** In Dart, closure literals are function types, not class instances. Code assigning a closure (`Callback cb = (x) => ...;`) will fail to compile against `class Callback`.
+* **Class type alias $\rightarrow$ subclass:** If `typedef A = B;` is replaced with `class A extends B {}`, `B` is no longer assignable to `A` ($B \not\le A$), and bidirectional type identity is lost.
+
+---
+
 ## Constructors
 
 ### MINOR: Add a new public constructor or factory
@@ -265,7 +334,7 @@ If a class previously had the default implicit constructor `MyClass()`, adding `
 
 Existing call sites omitting the parameter will fail to compile.
 
-#### MINOR: Add an optional parameter (positional or named)
+#### MINOR: Add an optional parameter
 
 Existing call sites can omit the new optional parameter without issue.
 
@@ -285,16 +354,16 @@ Call sites passing `func(paramName: value)` will fail to compile when `paramName
 
 Reordering positional parameters changes the expected argument positions at call sites. Even if the parameter types are identical, it causes callers to pass arguments to the wrong parameters, altering behavior or causing compile-time type errors.
 
-#### MAJOR: Narrow a parameter type (more specific / subtype)
+#### MAJOR: Narrow a parameter type
 
 Changing a parameter from `num` to `int` breaks callers that passed `double`.
 
-#### MINOR / MAJOR: Widen a parameter type (more general / supertype)
+#### Widen a parameter type
 
-* **MINOR** for top-level functions and `final` class methods, because callers can pass a wider variety of inputs.
+* **MINOR** for top-level functions and `final` class methods, because callers can pass a wider variety of inputs (for example, widening a parameter from `Future<T>` to `FutureOr<T>` or from `int` to `num`).
 * **MAJOR** for methods on an implementable class, because it changes the signature required for external overrides.
 
-#### PATCH / MINOR: Change the default value of an optional parameter
+#### MINOR: Change the default value of an optional parameter
 
 Does not break compilation or static type checks, but alters runtime behavior for callers omitting the argument.
 
@@ -302,38 +371,47 @@ Does not break compilation or static type checks, but alters runtime behavior fo
 
 ### Return types
 
-#### MINOR / MAJOR: Narrow a return type (more specific / subtype)
+Return types in Dart are **covariant**: callers expect a return value conforming to at least the declared type, while subclasses overriding a method must return a subtype of the superclass method's return type.
 
-* **MINOR** for callers of top-level functions and `final` class methods (for example, returning `int` instead of `num` is covariant and safe for callers).
-* **MAJOR** if the method is overridden in external subclasses or implementers.
+#### Narrow a return type
+
+Moving down the type lattice to a more specific subtype (e.g., `num` $\rightarrow$ `int`, `Object?` $\rightarrow$ `String`, `void` $\rightarrow$ `T`, or `T` $\rightarrow$ `Never`):
+
+* **MINOR for callers of top-level functions and `final` class methods:**
+  * **Direct invocations:** Callers receive a more specific type with additional available members and tighter type safety.
+  * **Top types (`void` $\rightarrow$ `T`):** Changing a return type from `void` to a specific type is a form of narrowing from the top type. Callers ignoring the return value continue to work, and callbacks passed to `void Function(...)` continue to compile because Dart treats `T Function(...)` as a subtype of `void Function(...)` for any type `T`.
+  * **Bottom type (`T` $\rightarrow$ `Never`):** For functions that unconditionally throw or exit, narrowing the return type to `Never` is backward-compatible because `Never` is a universal subtype of all Dart types.
+* **MAJOR if the method is overridden in external subclasses or implementers:**
+  * External subclasses that declared the previous wider return type (e.g., `@override num method()` or `@override void method()`) will fail to compile with an override error, because the subclass's wider return type is not a valid subtype of the new narrower return type.
 
 #### MAJOR: Widen a return type (more general / supertype)
 
-Changing a return type from `int` to `num` breaks callers expecting `int` methods and properties on the returned object.
+Moving up the type lattice to a broader supertype (e.g., `int` $\rightarrow$ `num`, `Future<T>` $\rightarrow$ `FutureOr<T>`, or specific type `T` $\rightarrow$ `void`):
 
-#### MINOR: Change return type from `void` to a specific type
-
-Existing callers ignoring the return value continue to work.
-
-#### MAJOR: Change return type from a specific type to `void`
-
-Callers using the return value will fail to compile.
+* **MAJOR for callers:**
+  * Callers expecting specific members on the returned object (such as `int` methods, or `.then()` on a `Future<T>`) will fail to compile.
+  * Changing a return type from a specific type `T` to `void` is a form of widening to the top type, breaking any callers attempting to read or use the returned value.
+* Subclasses that already returned a narrower subtype remain valid overrides.
 
 ---
 
 ### Generics and type parameters
 
-#### MAJOR: Add a type parameter without a default bound
+#### MAJOR: Add a bound to an unconstrained type parameter
 
-Call sites or type annotations lacking the type argument may fail to compile or change type inference behavior.
+Adding a type parameter constraint (for example, changing `class Store<T>` to `class Store<T extends State>`) breaks existing consumers using type arguments that do not conform to the new bound (such as `Store<String>`).
 
 #### MAJOR: Tighten a type parameter bound
 
 Changing `<T extends Object>` to `<T extends num>` breaks callers using non-number type arguments.
 
-#### MINOR: Loosen a type parameter bound
+#### MINOR: Loosen or remove a type parameter bound
 
-Changing `<T extends int>` to `<T extends num>` permits more types while remaining compatible with existing valid usages.
+Changing `<T extends int>` to `<T extends num>` or removing a bound permits more types while remaining compatible with existing valid usages.
+
+#### MAJOR: Add a type parameter without a default bound
+
+Call sites or type annotations lacking the type argument may fail to compile or change type inference behavior.
 
 ---
 
@@ -346,6 +424,10 @@ Adding a value to an `enum` breaks exhaustive `switch` statements and pattern ma
 ### MAJOR: Remove or rename an enum value
 
 Any code referencing `MyEnum.oldValue` will fail to compile.
+
+### MAJOR: Reorder enum values
+
+Reordering enum declarations does not cause static compilation errors, but alters each value's `.index` property at runtime. If your package or downstream users serialize enums by index (such as in databases or network protocols), reordering values is a breaking runtime change.
 
 ### MINOR: Add a member to an enhanced enum
 
@@ -364,6 +446,14 @@ Adding extension methods is backward-compatible. (In rare cases, static ambiguit
 #### MAJOR: Remove or rename an extension method
 
 Call sites invoking the extension member will fail to compile.
+
+#### MAJOR: Change the extended type (`on` clause)
+
+Changing the target type of an extension (for example, `extension Helpers on String` to `extension Helpers on int`) removes the extension members from instances of the original type, breaking all existing call sites.
+
+#### MINOR: Widen the extended type
+
+Widening the extended type (for example, from `on int` to `on num`) allows the extension to be used on more types while preserving all existing call sites.
 
 ---
 
--- a/src/content/tools/pub/semver-rules.md
+++ b/src/content/tools/pub/semver-rules.md
@@ -77,8 +77,14 @@ constant is backward-compatible. Existing consumer code will continue to compile
 and function as before.
 
 ```dart tag=good
-// Added to lib/my_package.dart in a minor release:
-void newHelperFunction() {}
+// Before (v1.0.0):
+// lib/my_package.dart
+void calculateTotal() {}
+
+// After (v1.1.0 - MINOR):
+// lib/my_package.dart
+void calculateTotal() {}
+void calculateTax() {} // Added: safe because existing callers are unaffected.
 ```
 
 *Note:* While introducing a new top-level name can theoretically create a naming
@@ -90,11 +96,27 @@ downstream code, SemVer treats top-level additions as non-breaking.
 Removing any public top-level declaration is a breaking change. Any downstream
 code referencing the removed symbol will fail to compile.
 
+```dart tag=bad
+// Before (v1.0.0):
+void calculateTax() {}
+
+// After (v2.0.0 - MAJOR):
+// `calculateTax` was removed. Existing callers fail to compile.
+```
+
 ### MAJOR: Rename a public top-level declaration
 
 Renaming a public symbol is functionally equivalent to removing the original
 name and introducing a new one. Downstream code referencing the original name
 will break.
+
+```dart tag=bad
+// Before (v1.0.0):
+void performTask() {}
+
+// After (v2.0.0 - MAJOR):
+void executeTask() {} // Renamed: callers referencing `performTask` fail to compile.
+```
 
 ### MAJOR: Move a declaration to `lib/src/` without re-exporting
 
@@ -102,11 +124,29 @@ Moving a declaration from the public `lib/` root to `lib/src/`, without an
 `export` in a public library file, makes it inaccessible to external packages,
 breaking all existing consumers.
 
+```dart tag=bad
+// Before (v1.0.0):
+// lib/my_package.dart
+export 'src/engine.dart';
+
+// After (v2.0.0 - MAJOR):
+// lib/my_package.dart
+// `src/engine.dart` is no longer exported: callers importing `package:my_package/my_package.dart` fail to access engine symbols.
+```
+
 ### MAJOR: Widen or change the type of a `const` or `final` variable
 
 Changing a `const` or `final` variable to a broader type, such as changing
 `const int maxRetries` to `const num maxRetries`, or to an incompatible type
 breaks callers expecting members of the more specific type.
+
+```dart tag=bad
+// Before (v1.0.0):
+const int maxRetries = 3;
+
+// After (v2.0.0 - MAJOR):
+const num maxRetries = 3; // Widened: callers expecting `int` methods break.
+```
 
 ### MINOR: Narrow the type of a `const` or `final` variable
 
@@ -115,6 +155,14 @@ specific subtype, such as changing `const num timeout` to `const int timeout` or
 `const Object key` to `const String key`, is backward-compatible. Because
 consumers only read the value, a more specific subtype satisfies all existing
 type expectations while providing additional type safety.
+
+```dart tag=good
+// Before (v1.0.0):
+const num timeout = 30;
+
+// After (v1.1.0 - MINOR):
+const int timeout = 30; // Narrowed: consumers only read the value, receiving a more specific subtype.
+```
 
 ### MAJOR: Change the type of a mutable top-level variable
 
@@ -125,10 +173,26 @@ the static type in either direction is a breaking change:
 * Widening the type, such as `int` to `num`, breaks callers expecting `int`
   methods when reading the variable.
 
+```dart tag=bad
+// Before (v1.0.0):
+num threshold = 10.5;
+
+// After (v2.0.0 - MAJOR):
+int threshold = 10; // Narrowed: callers assigning a double (e.g. `threshold = 1.5;`) fail to compile.
+```
+
 ### MAJOR: Change a top-level variable from `var` to `final`
 
 Making a mutable variable `final` breaks any consumer assigning values to that
 variable.
+
+```dart tag=bad
+// Before (v1.0.0):
+String globalConfig = 'default';
+
+// After (v2.0.0 - MAJOR):
+final String globalConfig = 'default'; // Callers assigning `globalConfig = 'custom';` fail to compile.
+```
 
 ---
 
@@ -149,35 +213,106 @@ packages are permitted to do with it:
 Changing a concrete class to `abstract class` prevents external consumers from
 instantiating the class directly using `new MyClass()`, causing compile errors.
 
+```dart tag=bad
+// Before (v1.0.0):
+class Formatter {}
+
+// Downstream package:
+final formatter = Formatter();
+
+// After (v2.0.0 - MAJOR):
+abstract class Formatter {} // Breaks: `Formatter()` can no longer be directly instantiated.
+```
+
 #### MINOR: Change an abstract class to a concrete class
 
 Making an abstract class concrete enables external instantiation without
 breaking existing subclasses or implementers.
+
+```dart tag=good
+// Before (v1.0.0):
+abstract class Formatter {}
+
+// After (v1.1.0 - MINOR):
+class Formatter {} // Safe: allows direct instantiation without breaking existing subtypes.
+```
 
 #### MAJOR: Add the `base` modifier to an existing class
 
 Prevents external consumers from implementing the class using `implements
 MyClass`, breaking all existing external implementers.
 
+```dart tag=bad
+// Before (v1.0.0):
+class Worker {}
+
+// Downstream package:
+class CustomWorker implements Worker {}
+
+// After (v2.0.0 - MAJOR):
+base class Worker {} // Breaks: downstream implementers (`implements Worker`) fail to compile.
+```
+
 #### MAJOR: Add the `interface` modifier to an existing class
 
 Prevents external consumers from extending the class using `extends MyClass`,
 breaking all existing external subclasses.
+
+```dart tag=bad
+// Before (v1.0.0):
+class Service {}
+
+// Downstream package:
+class CustomService extends Service {}
+
+// After (v2.0.0 - MAJOR):
+interface class Service {} // Breaks: downstream subclasses (`extends Service`) fail to compile.
+```
 
 #### MAJOR: Add the `final` or `sealed` modifier to an existing class
 
 Prevents external consumers from extending, implementing, or mixing in the
 class, breaking all existing external subtypes.
 
+```dart tag=bad
+// Before (v1.0.0):
+class Database {}
+
+// Downstream package:
+class MockDatabase extends Database {}
+
+// After (v2.0.0 - MAJOR):
+final class Database {} // Breaks: downstream subclasses fail to compile.
+```
+
 #### MAJOR: Remove the `mixin` modifier from a `mixin class`
 
 Prevents external consumers from using the class as a mixin using `with
 MyClass`.
 
+```dart tag=bad
+// Before (v1.0.0):
+mixin class TaskRunner {}
+
+// Downstream package:
+class Worker with TaskRunner {}
+
+// After (v2.0.0 - MAJOR):
+class TaskRunner {} // Breaks: downstream mixin uses (`with TaskRunner`) fail to compile.
+```
+
 #### MINOR: Remove `base`, `interface`, or `final` from a class
 
 Loosens restrictions on consumers, granting new capabilities without
 invalidating existing code.
+
+```dart tag=good
+// Before (v1.0.0):
+final class Database {}
+
+// After (v1.1.0 - MINOR):
+class Database {} // Safe: loosens restrictions and allows consumers to extend or implement.
+```
 
 ---
 
@@ -194,10 +329,21 @@ to write `class MyImpl implements Foo`, any new member in `Foo` breaks existing
 implementers due to missing overrides.
 
 ```dart tag=bad
-// In package:foo (v1.1.0) - BREAKING for external implementers:
+// Before (v1.0.0):
 abstract class HttpClient {
   void get(String url);
-  void post(String url, Object body); // External `implements HttpClient` fails!
+}
+
+// Downstream package (v1.0.0):
+class MockHttpClient implements HttpClient {
+  @override
+  void get(String url) {}
+}
+
+// After (v2.0.0 - MAJOR, or broken if released as MINOR):
+abstract class HttpClient {
+  void get(String url);
+  void post(String url, Object body); // Added: MockHttpClient fails to compile due to missing `post`.
 }
 ```
 
@@ -206,14 +352,51 @@ abstract class HttpClient {
 Removing or renaming an instance method, getter, setter, or field breaks all
 callers, extenders, and implementers.
 
+```dart tag=bad
+// Before (v1.0.0):
+class Account {
+  void close() {}
+}
+
+// After (v2.0.0 - MAJOR):
+class Account {
+  // `close()` was removed or renamed: existing callers fail to compile.
+}
+```
+
 #### MINOR: Add a static member to a class
 
 Static members are namespaced to the class, such as `Class.member`, and do not
 affect subclasses or external implementations.
 
+```dart tag=good
+// Before (v1.0.0):
+class MathUtils {
+  static double add(double a, double b) => a + b;
+}
+
+// After (v1.1.0 - MINOR):
+class MathUtils {
+  static double add(double a, double b) => a + b;
+  static double multiply(double a, double b) => a * b; // Safe: namespaced static member addition.
+}
+```
+
 #### MAJOR: Remove or rename a static member
 
 Breaks call sites referencing `Class.staticMember`.
+
+```dart tag=bad
+// Before (v1.0.0):
+class MathUtils {
+  static double multiply(double a, double b) => a * b;
+}
+
+// After (v2.0.0 - MAJOR):
+class MathUtils {
+  // `multiply` was removed: call sites referencing `MathUtils.multiply` fail to compile.
+}
+```
 
 ---
 
@@ -229,19 +412,42 @@ instance methods, getters, setters, or fields is completely backward-compatible.
 Consumers only call existing members and cannot be broken by added members.
 
 ```dart tag=good
-// Safe to add in a minor release:
+// Before (v1.0.0):
+final class Config {
+  final String host;
+  final int port;
+  Config(this.host, this.port);
+}
+
+// After (v1.1.0 - MINOR):
 final class Config {
   final String host;
   final int port;
   Config(this.host, this.port);
 
-  bool get isSecure => port == 443;
+  bool get isSecure => port == 443; // Added: safe because consumers cannot implement or extend Config.
 }
 ```
 
 #### MAJOR: Remove or rename an instance member of a final class
 
 Existing callers referencing the removed member will fail to compile.
+
+```dart tag=bad
+// Before (v1.0.0):
+final class Config {
+  final String host;
+  final int port;
+  Config(this.host, this.port);
+}
+
+// After (v2.0.0 - MAJOR):
+final class Config {
+  final String host;
+  // `port` removed: callers referencing `config.port` fail to compile.
+  Config(this.host);
+}
+```
 
 ---
 
@@ -256,10 +462,45 @@ Because external libraries can only extend rather than implement a `base class`,
 subclasses automatically inherit the new default implementation without needing
 to override it.
 
+```dart tag=good
+// Before (v1.0.0):
+base class Controller {
+  void start() {}
+}
+
+// Downstream package:
+class AppController extends Controller {}
+
+// After (v1.1.0 - MINOR):
+base class Controller {
+  void start() {}
+  void stop() {} // Added: AppController automatically inherits the concrete implementation.
+}
+```
+
 #### MAJOR: Add an abstract member to a base class
 
 Subclasses in external libraries will fail to compile because they do not
 provide an implementation for the new abstract member.
+
+```dart tag=bad
+// Before (v1.0.0):
+abstract base class Controller {
+  void start();
+}
+
+// Downstream package:
+class AppController extends Controller {
+  @override
+  void start() {}
+}
+
+// After (v2.0.0 - MAJOR):
+abstract base class Controller {
+  void start();
+  void stop(); // Added abstract: AppController fails to compile (missing `stop`).
+}
+```
 
 #### MAJOR: Remove or rename an instance member of a base class
 
@@ -277,6 +518,25 @@ extended using `extends`.
 External classes can implement an `interface class`. Adding any new method,
 getter, setter, or field requires implementers to provide the new member,
 breaking compilation.
+
+```dart tag=bad
+// Before (v1.0.0):
+interface class Repository {
+  void fetch();
+}
+
+// Downstream package:
+class CustomRepository implements Repository {
+  @override
+  void fetch() {}
+}
+
+// After (v2.0.0 - MAJOR):
+interface class Repository {
+  void fetch();
+  void save(); // Added: CustomRepository fails to compile due to missing `save()` implementation.
+}
+```
 
 #### MAJOR: Remove or rename an instance member of an interface class
 
@@ -296,22 +556,38 @@ exhaustive `switch` statements or pattern matches in consumer code that do not
 have a wildcard `_` or `default` case will fail to compile.
 
 ```dart tag=bad
+// Before (v1.0.0):
 sealed class Result {}
 class Success extends Result {}
 class Failure extends Result {}
 
-// Adding `class Pending extends Result {}` breaks exhaustive switches:
-// switch (result) {
-//   case Success(): ...
-//   case Failure(): ...
-//   // Error: Switch is no longer exhaustive!
-// }
+// Downstream package:
+String describe(Result result) => switch (result) {
+  Success() => 'Success',
+  Failure() => 'Failure',
+};
+
+// After (v2.0.0 - MAJOR):
+sealed class Result {}
+class Success extends Result {}
+class Failure extends Result {}
+class Pending extends Result {} // Added: downstream `describe` switch is no longer exhaustive and fails to compile.
 ```
 
 #### MINOR: Add a member to a sealed class
 
 External code cannot implement or extend a `sealed` class, so adding instance or
 static members is non-breaking.
+
+```dart tag=good
+// Before (v1.0.0):
+sealed class Result {}
+
+// After (v1.1.0 - MINOR):
+sealed class Result {
+  bool get isSuccess => this is Success; // Added: safe because consumers cannot implement or extend sealed classes.
+}
+```
 
 ---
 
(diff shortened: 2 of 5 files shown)


src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,382 @@
...
+---
+
+When developing and publishing Dart packages, adhering to
+[Semantic Versioning (SemVer)][semver] is essential for maintaining a healthy
+and predictable package ecosystem.
+
+This guide provides a detailed reference on how various changes to your Dart
+code affect the public API surface of your package and what version bump
+(`MAJOR`, `MINOR`, or `PATCH`) is required.
+
```
sigurdm, line 23: You could also reference package pub_semver, and perhaps mention how it treats version numbers differently before 1.0.0.

https://pub.dev/packages/pub_semver.

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,382 @@
+---
+title: SemVer compatibility and breaking changes
+breadcrumb: SemVer rules
+description: >-
+  A comprehensive guide to what changes are breaking (major), non-breaking
+  features (minor), or bug fixes (patch) in Dart packages.
+---
+
```
sigurdm, line 8: This page could be linked from https://github.com/dart-lang/site-www/blob/559bdaae5ceeb575b9ba9117d2d35b412684344b/src/content/tools/pub/publishing.md and https://github.com/dart-lang/site-www/blob/01f6c86867e35cd5100df0597f1e03b0b55aaba0/src/content/tools/pub/versioning.md#L6 and
https://github.com/dart-lang/site-www/blob/1d6d38b227013e7d87fc8c6fa56ff060d7d90dfa/src/content/tools/pub/pubspec.md#L179

mosuem:
@parlough do you know who approves PRs in this repo? There is no code owners file, so I am not sure.. I saw you filed a lot of PRs here.

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,405 @@
...
+* **Public:** Any library directly inside `lib/` (e.g., `lib/my_package.dart`)
+  and any declarations re-exported through `export` directives.
+* **Private / Internal:** Any file inside `lib/src/` that is **not**
+  re-exported by a public library in `lib/`. Changes to unexported files in
+  `lib/src/` are considered internal and do not affect the public SemVer contract.
+
+---
+
+## Top-level declarations
```
sigurdm, line 71: It seems to me all these rules should follow from some more basic definitions about what a breaking change is. Maybe we formulate something can state them up front.

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,405 @@
...
+description: >-
+  A comprehensive guide to what changes are breaking (major), non-breaking
+  features (minor), or bug fixes (patch) in Dart packages.
+---
+
+When developing and publishing Dart packages, adhering to
+[Semantic Versioning (SemVer)][semver] is essential for maintaining a healthy
+and predictable package ecosystem.
+
+This guide provides a detailed reference on how various changes to your Dart
```
sigurdm: This page is focused on API shape. We should also give some more general observations or guidelines of what behavioral changes are considered breaking. What constitutes a bugfix.

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,405 @@
...
+Re-exporting another package's library (`export 'package:foo/foo.dart';`) exposes new public API surface.
+
+### MAJOR: Remove an exported dependency
+
+Consumers relying on your package re-exporting those symbols will fail to compile.
+
+### PATCH / MINOR: Update dependencies within existing constraint ranges
+
+Normal maintenance as long as your package's own public API surface is unchanged.
+
```
sigurdm, line 693: We should also describe what happens if you start throwing more/less kinds of exceptions, what does that do to the contract.

mosuem: Good point. I think Dart is interesting there as we don't specify in the signature what a function throws, so it's definitely more on the informal side.

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,405 @@
+---
+title: SemVer compatibility and breaking changes
```
jonasfj, line 2: We could also call it something like:
 * "API versioning guidelines", or,
 * "Semantic versioning guidelines"

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,405 @@
...
+### MINOR: Add a new top-level declaration
+
+Adding a new public top-level function, class, typedef, enum, extension, or constant is backward-compatible. Existing consumer code will continue to compile and function as before.
+
+```dart tag=good
+// Added to lib/my_package.dart in a minor release:
+void newHelperFunction() {}
+```
+
+*Note:* While introducing a new top-level name can theoretically create a naming conflict with a wildcard import (`import 'package:foo/foo.dart';`) in downstream code, SemVer treats top-level additions as non-breaking.
```
jonasfj: > *Note:* While introducing a new top-level name can theoretically create a naming conflict with a wildcard import (`import 'package:foo/foo.dart';`) in downstream code, SemVer treats top-level additions as non-breaking.

* (A) It's not theoretical,
* (B) SemVer is not a person with opinions, not is it a specification that says any such thing as claimed here.

What you are stating here is that:
 * (1) Introducing a top-level name can break downstream consumers.
 * (2) Best-practices is to do a minor bump when introducing a new top-level name. Because:
   * New top-level names imply new functionality, and,
   * New top-level names only break downstream consumers when there is a naming conflict, which is rarely a problem.


-----------------

I think it might be worth while writing this a bit like "Effective Dart".
With DO, DON'T, CONSIDER, PREFER, AVOID.

We could even make a "Versioning Guide" to Effective Dart, granted it's less broadly relevant than the other guides. 


----------------

We could write this as:

### New top-level declaration
Adding a new public top-level function, class, typedef, enum, extension, or constant is backward-compatible, unless it causes naming conflicts for an existing user.

#### DO minor version bump
A new top-level declaration typically implies new functionality, rather than a mere bug fix (patch), hence, a minor version bump is appropriate.

#### CONSIDER major version bump
If the new top-level name is likely to conflict with existing names from other packages commonly used along side it, then consider a major version bump.

-----------

Possibly even with before/after examples for a `lib/library.dart` or something like that.

Just an idea.

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,405 @@
...
+
+Renaming a public symbol is functionally equivalent to removing the original name and introducing a new one. Downstream code referencing the original name will break.
+
+### MAJOR: Move a declaration to `lib/src/` without re-exporting
+
+Moving a declaration from the public `lib/` root to `lib/src/` (without an `export` in a public library file) makes it inaccessible to external packages, breaking all existing consumers.
+
+### MAJOR: Change the type of a top-level variable or constant
+
+Changing the static type of a top-level variable or constant causes type errors for any callers expecting the previous type.
```
jonasfj: It's perfectly fine to make a `const num` into a `const int` :D

Or a `const Object` into a `const String`.

A more specific type is fine, if you are expecting users to be reading it and not writing it.

src/content/tools/pub/semver-rules.md
mosuem: I added some paragraphs.

src/content/tools/pub/semver-rules.md
mosuem: Good idea. I tried it out, but almost everything said `DO:`, so it didn't give much in terms of readability or information... reverted it again afterwards.

src/content/tools/pub/semver-rules.md
mosuem: Do you have any idea of what that would look like? I like the idea but don't quite know what to write.

src/content/tools/pub/semver-rules.md
mosuem: Same as above - happy for suggestions to expand on..

src/content/tools/pub/pubspec.md
```diff
@@ -180,8 +180,11 @@ Once that's been done, consider it hermetically sealed: you can't touch it
 anymore. To make more changes, you'll need a new version.
 
 When you select a version, follow [semantic versioning.][semantic versioning]
+For a reference on which API changes in Dart require major, minor, or patch
+version bumps, see [SemVer compatibility rules][semver-rules].
```
conooi:
```suggestion
version bumps, check out [SemVer compatibility rules][semver-rules].
```

src/content/tools/pub/versioning.md
```diff
@@ -190,6 +192,12 @@ slot: going from `0.1.2` to `0.2.0` indicates a breaking change, going to
 doesn't affect the public API.
 For simplicity's sake, avoid using `+` after the version reaches `1.0.0`.
 
+:::tip
+For a detailed guide on what constitutes a breaking change versus a minor
+or patch update across Dart language features, see
```
conooi:
```suggestion
or patch update across Dart language features, check out
```

src/content/tools/pub/versioning.md
```diff
@@ -11,6 +11,8 @@ approach to it.
 Consider this to be advanced information.
 To learn _why_ pub was designed the way it was, keep reading.
 If you want to _use_ pub, consult the [other docs][pub].
+For a reference guide on which API changes require major, minor, or patch version bumps,
+see [SemVer compatibility rules][semver-rules].
```
conooi:
```suggestion
check out [SemVer compatibility rules][semver-rules].
```

conooi (approved):

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,540 @@
...
+
+Reordering positional parameters changes the expected argument positions at call sites. Even if the parameter types are identical, it causes callers to pass arguments to the wrong parameters, altering behavior or causing compile-time type errors.
+
+#### MAJOR: Narrow a parameter type
+
+Changing a parameter from `num` to `int` breaks callers that passed `double`.
+
+#### Widen a parameter type
+
+* **MINOR** for top-level functions and `final` class methods, because callers can pass a wider variety of inputs, such as widening a parameter from `Future<T>` to `FutureOr<T>` or from `int` to `num`.
```
natebosch: How nuanced do we want to be? Widening a parameter type _can_ be breaking in Dart, it's just very rare. 

```dart
void main() {
  var callback = doStuff;
  callback = (String s) {};
  callback('');
}

void doStuff(String s) {}
```

If we widen the `s` parameter to `Object` there is a static error.

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,540 @@
...
+
+Call sites invoking the extension member will fail to compile.
+
+#### MAJOR: Change the extended type: the `on` clause
+
+Changing the target type of an extension, such as changing `extension Helpers on String` to `extension Helpers on int`, removes the extension members from instances of the original type, breaking all existing call sites.
+
+#### MINOR: Widen the extended type
+
+Widening the extended type, such as from `on int` to `on num`, allows the extension to be used on more types while preserving all existing call sites.
```
natebosch: May occasionally change which extension member is resolved when there is ambiguity

natebosch (commented):
Do we want to talk about mitigation strategies on this page? For instance, replacing an instance member with an extension member may limit the breaking change to narrow enough corners (impacts uses of the member through `dynamic` or in a library that had no import to the definition).

src/content/tools/pub/semver-rules.md
sigurdm: Maybe we can explain something in terms of contracts, and providing/requiring (you are allowed to provide more and require less, but not vice-versa)

src/content/tools/pub/semver-rules.md
sigurdm: Perhaps we can give some examples? Perhaps a cautionary reminder that one person's bug is another person's feature.
We could write that generally documented limitations override the rules here (if it is documented that a class should not be implemented, then it is a bugfix to make it final)

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,696 @@
...
+Adding any instance method, getter, setter, or field to an unmodified `class` or
+`abstract class` is a breaking change. Because external libraries are permitted
+to write `class MyImpl implements Foo`, any new member in `Foo` breaks existing
+implementers due to missing overrides.
+
+```dart tag=bad
+// In package:foo (v1.1.0) - BREAKING for external implementers:
+abstract class HttpClient {
+  void get(String url);
+  void post(String url, Object body); // External `implements HttpClient` fails!
```
brianquinlan, line 200: Is it clear that this is the new method?

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,696 @@
...
+
+#### MINOR: Add an instance member to a final class
+
+Since external code cannot extend or implement a `final class`, adding new
+instance methods, getters, setters, or fields is completely backward-compatible.
+Consumers only call existing members and cannot be broken by added members.
+
+```dart tag=good
+// Safe to add in a minor release:
+final class Config {
```
brianquinlan, line 233: Are these examples useful? It's not clear what was added.

src/content/tools/pub/semver-rules.md
mosuem: Added a note.

src/content/tools/pub/semver-rules.md
mosuem: Added a note about name clashing in general.

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1457 @@
...
+SemVer impact of changing thrown errors or exceptions depends on the distinction
+between [programmatic errors][effective_dart_errors] and [runtime
+exceptions][effective_dart_errors], as well as your documented API contract.
+
+### Documented exceptions
+
+When a library explicitly documents that a function or method throws a specific
+exception, such as `/// Throws [AuthException] on invalid credentials.`:
+
+#### MAJOR: Change or remove a documented exception type
```
sigurdm: Is removing an exception also a breaking change? I don't see it in the section text, and would assume it to be non-breaking. Maybe just leave it out (or should it be "adding".
```suggestion
#### MAJOR: Change or add a documented exception type
```

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1457 @@
...
+affect the public API surface of your package and what version bump is required.
+
+:::tip
+For a conceptual overview of pub's version solving algorithm and lockfiles, see
+the [Package versioning](/tools/pub/versioning) guide.
+:::
+
+## Semantic versioning and `package:pub_semver`
+
+Dart's package manager, pub, follows [Semantic Versioning 2.0.0][semver] as
```
sigurdm: (comment from claude) The link goes to `2.0.0-rc.1`. That's what the rest of the site links too, but then the text shouldn't say 2.0.0.

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1457 @@
...
+
+While standard SemVer 2.0.0 allows any change before version `1.0.0`, pub and
+[`pub_semver`][pub_semver] enforce a stricter convention so that consumers can
+safely depend on pre-1.0.0 packages using [caret
+syntax](/tools/pub/dependencies#caret-syntax):
+
+* **`0.y.z` versions where $y > 0$, such as `0.2.0`:**
+  * Bumping `y` from `0.2.0` to `0.3.0` is treated as a **breaking change**,
+    which is equivalent to a major bump.
+  * Bumping `z` from `0.2.0` to `0.2.1` is treated as a **backward-compatible
```
sigurdm: (comment from claude) Worth saying that pre-1.0 you can't tell MINOR from PATCH at all, and pointing to the `+1` build-suffix convention from versioning.md.

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1457 @@
...
+* **`0.0.z` versions, such as `0.0.1`:**
+  * Bumping `z` from `0.0.1` to `0.0.2` is treated as a **breaking change**.
+  * Caret syntax `^0.0.1` allows only `>=0.0.1 <0.0.2`.
+
+## The public API boundary
+
+In Dart packages, the **public API** comprises all declarations accessible to
+consumers of your package:
+
+* **Public:** Any library directly inside `lib/`, such as `lib/my_package.dart`,
```
sigurdm: (comment from claude) Libraries in `lib/foo/bar.dart` are public too. Only `lib/src/` is private.

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1457 @@
...
+  and any declarations re-exported through `export` directives.
+* **Private and internal:** Any file inside `lib/src/` that is **not**
```
sigurdm: (comment from claude) Maybe mention that declarations marked `@internal` / `@visibleForTesting` (package:meta) aren't part of the contract even if they're reachable.

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1457 @@
...
+
+```dart tag=bad
+// Before (v1.0.0):
+const int maxRetries = 3;
+
+// After (v2.0.0 - MAJOR):
+const num maxRetries = 3; // Widened: callers expecting `int` methods break.
+```
+
+### MINOR: Narrow the type of a `const` or `final` variable
```
sigurdm: (comment from claude) Same inference caveat as the one added for widening parameters:
```dart
var t = timeout; // now inferred as int
t = 1.5;         // error
```
Fine to call it MINOR, but then the caveat applies everywhere, not just in the tear-off note.

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1457 @@
...
+
+```dart tag=bad
+// Before (v1.0.0):
+num threshold = 10.5;
+
+// After (v2.0.0 - MAJOR):
+int threshold = 10; // Narrowed: callers assigning a double (e.g. `threshold = 1.5;`) fail to compile.
+```
+
+### MAJOR: Change a top-level variable from `var` to `final`
```
sigurdm: (comment from claude) Also missing: `const` → `final` is MAJOR (breaks const contexts). Changing the *value* of a `const` can break too (switch cases, default values, annotations).

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1457 @@
...
+// Before (v1.0.0):
+abstract class Formatter {}
+
+// After (v1.1.0 - MINOR):
+class Formatter {} // Safe: allows direct instantiation without breaking existing subtypes.
+```
+
+#### MAJOR: Add the `base` modifier to an existing class
+
+Prevents external consumers from implementing the class using `implements
```
sigurdm: (comment from claude) Extenders break too, not just implementers:
`class CustomWorker extends Worker {}` → "The type 'CustomWorker' must be 'base', 'final' or 'sealed' because the supertype 'Worker' is 'base'."

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1457 @@
...
+class Service {}
+
+// Downstream package:
+class CustomService extends Service {}
+
+// After (v2.0.0 - MAJOR):
+interface class Service {} // Breaks: downstream subclasses (`extends Service`) fail to compile.
+```
+
+#### MAJOR: Add the `final` or `sealed` modifier to an existing class
```
sigurdm: (comment from claude) `sealed` also makes the class abstract, so construction breaks as well.

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1457 @@
...
+mixin class TaskRunner {}
+
+// Downstream package:
+class Worker with TaskRunner {}
+
+// After (v2.0.0 - MAJOR):
+class TaskRunner {} // Breaks: downstream mixin uses (`with TaskRunner`) fail to compile.
+```
+
+#### MINOR: Remove `base`, `interface`, or `final` from a class
```
sigurdm: (comment from claude) - Removing `sealed` isn't covered. It's MAJOR, because exhaustive switches over the subtypes stop being exhaustive.
- Worth a sentence that removing these is a one-way door: afterwards, adding members is MAJOR by this doc's own rules.
- Adding `mixin` to a class (`class` → `mixin class`) is MINOR and could go here.

sigurdm (commented):
(comment from claude) Thanks, this is a useful catalogue. The main thing I'd like fixed before landing is consistency. Right now some rules say "technically breaking, so MAJOR" (adding members to plain classes, enum values, sealed subtypes) and others say "technically breaking but rare, so MINOR" (top-level name clashes, extension ambiguity, optional params vs. tear-offs, inference). There's no stated reason for the difference. I think a short definitions section up front would solve most of it: something like "breaking = code that uses the API as documented/intended stops compiling (errors, not lints) or changes behavior outside the documented contract", plus "you may provide more and require less". Then say explicitly how documented intent ("not meant to be implemented", "may add enum values") changes the verdict.

Also missing:
- how to make changes *without* breaking (deprecated aliases, `final` classes, documenting intent, don't serialize enums by `.index`)
- deprecation (`@Deprecated`) and `package:meta` annotations (`@internal`, `@visibleForTesting`, `@protected`, `@mustCallSuper`, `@nonVirtual`)
- non-Dart API surface: `bin/` executables and their flags, assets, dropping platform support (e.g. starting to import `dart:io`)
- MINOR vs PATCH guidance. Almost every rule is MAJOR vs MINOR.

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1457 @@
...
+```
+
+---
+
+### Unmodified classes: `class` and `abstract class`
+
+An unmodified class in Dart allows external packages to construct, extend using
+`extends`, and implement using `implements`.
+
+#### MAJOR: Add an instance member to an unmodified class
```
sigurdm: (comment from claude) This contradicts the mixin rule below (L596), which says adding a member is technically breaking for `implements`, but intended usage is `with`, so MINOR. Taken literally, this rule makes most minor releases in the ecosystem wrong, the SDK and Flutter included. If that's really the policy, say so explicitly. Otherwise apply the same "intended use / documented intent" reasoning. Adding enum values and sealed subtypes needs the same treatment.

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1457 @@
...
+```
+
+---
+
+### `base class`
+
+A `base` class can be extended outside the library, but cannot be implemented
+using `implements`.
+
+#### MINOR: Add a concrete instance member to a base class
```
sigurdm: (comment from claude) A new concrete member can still break extenders that already declare a member with the same name and an incompatible signature. With a compatible signature, their member silently overrides it. Same for mixins (next comment) and, via shadowing of extension members, for every "add a member" rule in the doc.

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1457 @@
...
+Breaks both callers and extending subclasses.
+
+---
+
+### `interface class` and `abstract interface class`
+
+An `interface` class can be implemented outside the library, but cannot be
+extended using `extends`.
+
+#### MAJOR: Add any member to an interface class
```
sigurdm: (comment from claude) Static members are fine, cf. L367.

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1457 @@
...
+
+External classes can implement an `interface class`. Adding any new method,
+getter, setter, or field requires implementers to provide the new member,
+breaking compilation.
+
+```dart tag=bad
+// Before (v1.0.0):
+interface class Repository {
+  void fetch();
```
sigurdm: (comment from claude) This doesn't compile: "'fetch' must have a method body because 'Repository' isn't abstract". Needs `abstract interface class`.

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1457 @@
...
+sealed class Result {
+  bool get isSuccess => this is Success; // Added: safe because consumers cannot implement or extend sealed classes.
+}
+```
+
+---
+
+### `mixin` and `mixin class`
+
+#### MINOR: Add a concrete member to a mixin
```
sigurdm: (comment from claude) Besides `implements`: with `class C extends S with M`, a new `M.who()` silently overrides an existing `S.who()`. It compiles and changes behavior.

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1457 @@
...
+mixin Serializer {
+  String serialize();
+  void deserialize(String json); // Added abstract: User fails to compile due to missing `deserialize`.
+}
+```
+
+#### MAJOR: Tighten the `on` constraint on a mixin
+
+Restricting the superclass requirements, such as from `on Object` to `on
+Widget`, prevents existing classes that do not meet the new requirement from
```
sigurdm: (comment from claude) The example uses `State`.

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1457 @@
...
+abstract class Callback {
+  void call();
+} // Function literals are not instances of class Callback: downstream `cb = () => ...` fails to compile.
+```
+
+---
+
+## Constructors
+
+### MINOR: Add a new public constructor or factory
```
sigurdm: (comment from claude) Not in general. If the class only had the implicit default constructor, adding `Request.json(...)` removes `Request()` ("The class 'Request' doesn't have an unnamed constructor"). That's the same issue as the private-constructor rule at L848, but for any constructor.
Also missing: changing a generative constructor to `factory` is MAJOR for extendable classes (subclasses need a generative super constructor).

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1457 @@
...
+
+```dart tag=bad
+// Before (v1.0.0):
+void logMessage(String message) {}
+
+// After (v2.0.0 - MAJOR):
+void logMessage(String message, {required String level}) {} // Callers using `logMessage('data')` fail to compile.
+```
+
+#### MINOR: Add an optional parameter
```
sigurdm: (comment from claude) For methods on extendable/implementable classes, this breaks every existing override ("… isn't a valid override"). That's probably the most common real-world break. The doc makes this distinction for widening parameter types but not here.

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1457 @@
...
+
+```dart tag=bad
+// Before (v1.0.0):
+void connect(String host, int port) {}
+
+// After (v2.0.0 - MAJOR):
+void connect(String host) {} // Callers passing two arguments fail to compile.
+```
+
+#### MAJOR: Rename a named parameter
```
sigurdm: (comment from claude) Related cases that aren't covered:
- renaming a positional parameter: non-breaking
- required positional → optional, or `required` named → optional: fine for callers, MAJOR for overriders
- positional ↔ named: MAJOR
- `covariant`

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1457 @@
...
+:::note Tear-offs and widened parameter types
+Widening a parameter type changes the static type of a tear-off. Because
+function parameters are contravariant, tearing off a function with a widened
+parameter type (like `void Function(Object)`) can prevent assigning narrower
+callbacks into an inferred variable (like `var callback = doStuff; callback =
+(String s) {};`). In practice, such situations are rare, and widening parameter
+types on top-level functions is treated as non-breaking.
+:::
+
+#### MINOR: Change the default value of an optional parameter
```
sigurdm: (comment from claude) This is a behavior change for every caller that omits the argument. The exceptions section says throwing on previously succeeding input is MAJOR. It's hard to see why this is less breaking. I'd say it depends on the contract, often MAJOR.

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1457 @@
...
+Return types in Dart are **covariant**: callers expect a return value conforming
+to at least the declared type, while subclasses overriding a method must return
+a subtype of the superclass method's return type.
+
+#### Narrow a return type
+
+Moving down the type lattice to a more specific subtype, such as `num` to `int`,
+`Object?` to `String`, `void` to a specific type, or any type to `Never`:
+
+* **MINOR for callers of top-level functions and `final` class methods:**
```
sigurdm: (comment from claude) Same inference issue as L151: `var x = getScale(); x = 1.5;` stops compiling.

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1457 @@
...
+
+```dart tag=good
+// Before (v1.0.0):
+class NumericList<T extends int> {}
+
+// After (v1.1.0 - MINOR):
+class NumericList<T extends num> {} // Safe: existing `NumericList<int>` remains valid, and `NumericList<double>` is now supported.
+```
+
+#### MAJOR: Add a type parameter without a default bound
```
sigurdm: (comment from claude) Dart doesn't have default type arguments. What happens: raw references (`Store x`) get instantiated to bounds and keep working, while explicit type arguments (`Store<String>`, `f<int>()`) get the wrong arity. An example would help.

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1457 @@
...
+}
+
+// After (v1.1.0 - MINOR):
+extension StringUtils on String {
+  bool get isBlank => trim().isEmpty;
+  bool get isNotBlank => !isBlank; // Added: backward-compatible new extension member.
+}
+```
+
+#### MAJOR: Remove or rename an extension method
```
sigurdm: (comment from claude) Renaming the extension itself also breaks explicit `StringUtils(s).isBlank` and `show`/`hide`.

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1457 @@
...
+}
+```
+
+:::note Extension ambiguity and name collisions
+Widening an extension's target type (or adding a new extension) can
+occasionally introduce static ambiguity if a downstream consumer already has
+another extension with the same member name in scope, because Dart's extension
+resolution rules may now rank them with equal specificity.
+
+In SemVer, potential naming collisions and resolution ambiguities in downstream
```
sigurdm: (comment from claude) As Jonas said above, SemVer doesn't say this. It's our recommendation. Suggest "We recommend treating …".

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1457 @@
...
+wildcard scopes are treated as non-breaking (`MINOR`), because consumers can
+disambiguate them using explicit extension invocation syntax (such as
+`NumberUtils(x).doubleValue()`) or `hide` directives.
+:::
+
+---
+
+### Extension types: `extension type`
+
+#### MAJOR: Change the underlying representation type
```
sigurdm: (comment from claude) Renaming the representation field (`Id(int value)` → `Id(int id)`) is also breaking, because `value` is a public getter. Adding an `implements` clause is MINOR; removing one is MAJOR.

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1457 @@
...
+// Before (v1.0.0):
+extension type Id(int value) {}
+
+// After (v2.0.0 - MAJOR):
+extension type Id(String value) {} // Breaks callers passing `int` to `Id(...)`.
+```
+
+#### MINOR: Add a member to an extension type
+
+Extension types cannot be implemented or extended with subtyping, so adding
```
sigurdm: (comment from claude) They can, by other extension types: `extension type MyId(int value) implements Id`. The MINOR verdict mostly holds because redeclaration is allowed, but the stated reason is wrong.

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1457 @@
...
+}
+
+// After (v1.1.0 - MINOR):
+/// Throws [ExpiredTokenException] on invalid credentials.
+void authenticate(String token) {
+  throw ExpiredTokenException('Token expired'); // Safe: `ExpiredTokenException extends AuthException`, so `on AuthException` still catches it.
+}
+```
+
+#### MAJOR: Throw an exception on previously succeeding inputs or states
```
sigurdm: (comment from claude) Also: switching between a synchronous throw and a failed `Future` (or stream error event) breaks `try`/`catch` without `await`. And starting to throw an exception that isn't documented at all.

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1457 @@
...
+null values where an `ArgumentError` was previously thrown, is a
+backward-compatible feature addition.
+
+---
+
+## Dependencies and SDK constraints
+
+### MINOR: Increase the Dart SDK constraint lower bound
+
+Increasing the SDK constraint, such as from `sdk: ^3.5.0` to `sdk: ^3.6.0`, is
```
sigurdm: (comment from claude) This also bumps the package's language version, which can change the API. For example, going from <3.0 to ≥3.0 means plain `class`es can no longer be used as mixins unless marked `mixin class`. That's MAJOR if anyone used them that way.

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1457 @@
...
+standard for minor releases to adopt new language features, provided consumers
+on supported SDKs can resolve it.
+
+### MINOR: Export a new dependency
+
+Re-exporting another package's library using `export 'package:foo/foo.dart';`
```
sigurdm: (comment from claude) Worth warning that this ties your versioning to theirs. More generally: if a dependency's types appear in your public API, or you re-export it, allowing a new major of that dependency is MAJOR for you.

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1457 @@
...
+exposes new public API surface.
+
+### MAJOR: Remove an exported dependency
+
+Consumers relying on your package re-exporting those symbols will fail to
+compile.
+
+### PATCH: Update dependencies within existing constraint ranges
+
+Normal maintenance as long as your package's own public API surface is
```
sigurdm: (comment from claude) Not sure what this means. If the resolved version changes within your constraints, there's nothing to publish. Maybe it means raising lower bounds?

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1551 @@
...
+breadcrumb: SemVer rules
+description: >-
+  A comprehensive guide to what changes are breaking major changes, non-breaking
+  minor features, or patch bug fixes in Dart packages.
+---
+
+When developing and publishing Dart packages, adhering to [Semantic
+Versioning][semver] helps create a healthy and predictable package ecosystem.
+
+Any Dart package provides a set of features—and with that comes a
```
lrhn: (I'd personally use comma here, not mdash.)

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1551 @@
...
+contract that the user can rely on this set of features to continue working.
+Packages sometimes get changed, and there are two categories of changes: adding
+or removing functionality. Some changes do one or the other, some do both. While
```
lrhn: Or changing functionality. If you have complex behavior, say a function that reflows text, you can change the algorithm, and therefore the functionality, without removing or adding anything. 

(And also woth saying that API is not the same as functionality.)

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1551 @@
...
+removing functionality may sometimes be necessary, doing so violates the
+contract with the user. To accommodate this, Dart uses semantic versioning to
+help users recognize whether a new version of a package adheres to the old
+contract or presents a new one.
```
lrhn: This paragraph is completely abstract. It doesn't say what a "feature" is or what a "change" is.

It basically says "some changes are breaking" and "SemVer should reflect breaking changes".
Which is the definition of SemVer.

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1551 @@
...
+
+This guide provides a detailed reference on which changes to your Dart code
```
lrhn: This sounds like the list is complete. Let's not assume that :)

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1551 @@
...
+affect the public API surface of your package and what version bump is required.
+
+:::tip
+For a conceptual overview of pub's version solving algorithm and lockfiles, see
+the [Package versioning](/tools/pub/versioning) guide.
```
lrhn: (Why would I want to? 😉 
This is the first mention of "pub", and most concepts of this sentence, in this document. So basically "here is a lot of new unexplained words, you can read more about them here:...", which leaves a first-time reader without the context needed to even know why they should follow the link.
If there are assumptions made about what the reader already knows, it's worth saying them up front, with links, so a reader who doesn't know those things can follow the links and learn before continuing.)

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1551 @@
...
+:::
+
+## Semantic versioning and `package:pub_semver`
+
+Dart's package manager, pub, follows [Semantic Versioning][semver] as
+implemented by [`package:pub_semver`][pub_semver].
+
+A standard version number is formatted as `MAJOR.MINOR.PATCH`, such as `1.2.3`:
+
+* **`MAJOR`**: Incompatible breaking changes to the public API.
```
lrhn: I'd use more text here, to at least say that these are hierarchical (minor versions are only comparable if they have the same major version, patch only if same minor version, and the text after the `:`s here is what a larger, but comparable, number means.

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1551 @@
...
+* **`MINOR`**: Backward-compatible new functionality or features.
+* **`PATCH`**: Backward-compatible bug fixes.
+
+### Pre-1.0.0 versions
```
lrhn: (Conside moving this to the end and adding a comment like "Versions before 1.0.0, so with major version 0, are interpreted specially. See [section below](#pre-1-0-0-version) for detauks.
It's important, but it's not urgent, and it's not the details the reader came to _this_ document for. Especially if they already know SemVer.

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1551 @@
...
+    are occasionally used for tiny fixes prior to `1.0.0`).
+  * Caret syntax `^0.2.0` allows `>=0.2.0 <0.3.0`.
+* **`0.0.z` versions, such as `0.0.1`:**
+  * Bumping `z` from `0.0.1` to `0.0.2` is treated as a **breaking change**.
+  * Caret syntax `^0.0.1` allows only `>=0.0.1 <0.0.2`.
+
+## The public API boundary
+
+In Dart packages, the **public API** comprises all declarations that are part of
+your package's supported contract for external consumers:
```
lrhn: This reads like "public API is the following list", but that's actually only the first item, the next item is exactly what is not public API.

I'd write this differently, and focus on declarations, not libraries.
Mentioning reexports is a red herring, that's an implementation detail.
What matters is which declarations a public library exposes.

> The **public API** of a your package are the declarations which are exported by each **public library**, and which are not explicitly marked as non-public.
> A library is public if the library file is in `lib/`, but not inside `lib/src/`.
> A declaration is marked as non-public if it's marked with annotations
> like `@internal`, `@visibleForTemplate` or `@visiblForTesting` from
> [`package:meta`]({{site.pub-pkg}}/meta), or if it's explicitly documented
> as only for internal use.
>
> Anything else is private to the package, and not intended for use by
> clients of the package.
>
> _Nothing prevents a user from importing a library from inside `lib/src/` of another
> package (as `import 'package:example/src/secret.dart';`) or using
> an otherwise visible member marked `@internal`, but being non-public means that
> doing so is heavily discouraged. The author has made **no promises** that using such
> declarations is safe or will keep working._

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1551 @@
...
+
+* **Public:** Any library inside `lib/` outside of `lib/src/`, including
+  subdirectories such as `lib/my_package.dart` or `lib/foo/bar.dart`, and any
+  declarations re-exported through `export` directives.
+* **Private and internal:** Any file inside `lib/src/` that is **not**
+  re-exported by a public library in `lib/`, as well as declarations annotated
+  with `@internal` or `@visibleForTesting` from [`package:meta`]({{site.pub-pkg}}/meta).
+  Changes to internal declarations do not affect the public SemVer contract.
```
lrhn: (If "Private an internal" is not "everything else", then I'd have to ask what the rest is.
And if it's "everything else", it's better to just say that.)

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1551 @@
...
+   return type or adding a method to a `final` class). A change is breaking
+   (`MAJOR`) when it requires more or provides less.
+2. **Documented contracts and intended usage:** If a declaration explicitly
+   documents restrictions on its use—for example, stating in doc comments that a
+   class must not be implemented externally, or that an enum may gain new values
+   in minor releases—then changes that only affect unsupported usage are
+   non-breaking (`MINOR` or `PATCH`).
+3. **Pragmatic ecosystem conventions:** In Dart, almost any addition can
+   theoretically cause a static error in rare consumer patterns, such as
+   unprefixed wildcard imports, local variable type inference (`var x = fn();`),
```
lrhn: What's a "wildcard import"? (Not a common Dart term. I know what it is in Java, but Dart doesn't have any equivalent.)

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1551 @@
...
+   method tear-offs, or extension resolution ambiguity. By convention, Dart
+   package versioning treats these edge cases as non-breaking (`MINOR`) so that
+   packages can evolve without constant major version bumps.
```
lrhn: Yep. Some of these can be summarized as "Promises are only about this package's API, not about their integration into another program. Name conflicts with other packages are outside the scope of SemVer guarantees."

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1551 @@
...
+
+// After (v2.0.0 - MAJOR):
+void executeTask() {} // Renamed: callers referencing `performTask` fail to compile.
+```
+
+### MAJOR: Move a declaration to `lib/src/` without re-exporting
+
+Moving a declaration from the public `lib/` root to `lib/src/`, without an
+`export` in a public library file, makes it inaccessible to external packages,
+breaking all existing consumers.
```
lrhn: I'd ignore `lib/src/` here.

The problem is _removing a declaration from the exports of a public library_.

It doesn't matter what happens elsewhere. Moving is removing and adding elsewhere, but even moving the declaration to another public library is still breaking, because it is no longer exported by the original library, so any code that refers to the name as imported from the original library will break.

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1551 @@
...
+export 'src/engine.dart';
+
+// After (v2.0.0 - MAJOR):
+// lib/my_package.dart
+// `src/engine.dart` is no longer exported: callers importing `package:my_package/my_package.dart` fail to access engine symbols.
+```
+
+### MAJOR: Widen or change the type of a `const` or `final` variable
+
+Changing a `const` or `final` variable to a broader type, such as changing
```
lrhn: Also an error if it's not a `const` or `final` variable. Maybe just "a variable"?

Changing the return type of any function, _including a getter_, to something that is not a subtype, is breaking.

src/content/tools/pub/semver-rules.md
lrhn: True, but it's one of the exceptions mentioned above as "pragmatic ecosystem conventions".
We don't guarantee that code can't break. We just guarantee that you _can_ do the same thing as before. You may just have to write the types out.
(Basically, if all inferred types had been explicit, would it break? If not, it's a safe change.)

I'd probably accept a change like this in a patch releases if it's considered a bug fix.
If it's a deliberate API change, then there is probably more behind it, and it'll be a minor change anyway.

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1551 @@
...
+// Before (v1.0.0):
+const num timeout = 30;
+
+// After (v1.1.0 - MINOR):
+const int timeout = 30; // Narrowed: consumers only read the value, receiving a more specific subtype.
+```
+
+### MAJOR: Change the type of a mutable top-level variable
+
+For mutable `var` variables that can be both read from and written to, changing
```
lrhn: I think most public mutable variables are not declared using `var`.
(Style guide says type-annotate public API!)
It's more likely to be `String? someName;`.

Just "mutable variables", it's cleaner. (The "that can be both read from and written to" is perfect.)

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1551 @@
...
+the static type in either direction is a breaking change:
+* Narrowing the type, such as `num` to `int`, breaks callers assigning a
+  `double` into the variable.
+* Widening the type, such as `int` to `num`, breaks callers expecting `int`
+  methods when reading the variable.
```
lrhn: (And this is because widening return type of a getter is widening a return type, and narrowing the type of a setter is narrowing a parameter type, and those are breaking changes. Consider having those general rules first: Returns types, parameter types, widening and narrowing. Then variables can refer to those with their getters and setters.)

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1551 @@
...
+
+```dart tag=bad
+// Before (v1.0.0):
+num threshold = 10.5;
+
+// After (v2.0.0 - MAJOR):
+int threshold = 10; // Narrowed: callers assigning a double (e.g. `threshold = 1.5;`) fail to compile.
+```
+
+### MAJOR: Change a top-level variable from `var` to `final` or from `const` to `final`
```
lrhn: This is really two different cases.

" From mutable to not" just removes a setter from public API, and removing a declaration is breaking.

"From const to non-const" (not just final, from const to mutable is also bad), is a new category of breaking: "breaking constants".
If something can be used in a constant expression, changing it so that it no longer can is breaking. There are a number of rules in this group (`const` variable -> not const, `const` constructor -> not `const`, ... well, maybe that's it. Maybe it takes up a lot of headroom because I've had to think about it alot. Anytime someone asks "can't we infer that this could be const" I say no because if the author hasn't _promised_ it can be used with const, they should be free to refactor in a way that is not const compatible.)

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1551 @@
...
+`mixin`. These allow package authors to explicitly define how classes can be
+used by consumers. The SemVer impact of adding or modifying members depends on
+these modifiers.
+
+### Changing modifiers on existing classes
+
+Changing the modifier on an already-published class alters what external
+packages are permitted to do with it:
+
+#### MAJOR: Change a concrete class to an abstract class
```
lrhn: But _only_ if there is a public generative construtor that someone could use to do `new Myclass()`. if there isn't, it's not a change at all, and can happen in a patch.
Factory constructors are fine.

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1551 @@
...
+class Formatter {}
+
+// Downstream package:
+final formatter = Formatter();
+
+// After (v2.0.0 - MAJOR):
+abstract class Formatter {} // Breaks: `Formatter()` can no longer be directly instantiated.
+```
+
+#### MINOR: Change an abstract class to a concrete class
```
lrhn: (And if there are not public generative constructors, it's not even a minor change.)

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1551 @@
...
+// Before (v1.0.0):
+abstract class Formatter {}
+
+// After (v1.1.0 - MINOR):
+class Formatter {} // Safe: allows direct instantiation without breaking existing subtypes.
+```
+
+#### MAJOR: Add the `base` modifier to an existing class
+
+Prevents external consumers from implementing the class using `implements
```
lrhn: "external consumers" is a long way of say "users".

Nothing in this document applies inside the same package. This is about the public API exposed to others. There is no need to sidstinguish between internal and external consumers, so it's just "consumers", and that's still a long word for "users". 

I'dp pick either "users" or "clients" (of the package), and use it consistently.

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1551 @@
...
+class Database {}
+
+// Downstream package:
+class MockDatabase extends Database {}
+
+// After (v2.0.0 - MAJOR):
+final class Database {} // Breaks: downstream subclasses fail to compile.
+```
+
+#### MAJOR: Remove the `mixin` modifier from a `mixin class`
```
lrhn: (There are two things one can remove from an API: declarations and capabilities. So ... maybe say something early about that: Declarations are easy, they're either there or not, they have a signature or interface. Capabilities is about _how_ you can use a declaration. Const-compatibility is one thing. Class modifiers is another, which controls how you can use the declaration as a superclass, mixin or superinterface.)

src/content/tools/pub/semver-rules.md
```diff
@@ -0,0 +1,1551 @@
...
+// Before (v1.0.0):
+final class Database {}
+
+// After (v1.1.0 - MINOR):
+class Database {} // Safe: loosens restrictions and allows consumers to extend or implement.
+```
+
+---
+
+### Unmodified classes: `class` and `abstract class`
```
lrhn: That's not what "Unmodified" means! 😉 

I'd use "Unrestricted" or just "classes with no access modifiers".
Or even "open", but that never really caught on.

This is a new section. The previous one was "Changing modifiers on existing classes", which was about _changes_.

This headline doesn't say what changes it's about.
Maybe 
> ### Changing members of unrestricted non-mixin classes.
>
> Unrestricted classes are `class` declarations with no `interface`, `base`, `final`,
> or `sealed` modifier restricting their use.
> They can be `abstract` or not.

lrhn (commented):
Partial review, only got to my last comment.

src/content/tools/pub/semver-rules.md
mosuem: Done—removed "or remove" from the heading (`#### MAJOR: Change a documented exception type`), as stopping throwing an exception is covered under `MINOR` right below.

src/content/tools/pub/semver-rules.md
mosuem: Done—updated the text to say "Semantic Versioning" / "SemVer" without "2.0.0".

src/content/tools/pub/semver-rules.md
mosuem: Done—clarified that `0.y.z` does not distinguish `MINOR` from `PATCH` and linked to the `+` build suffix convention in `versioning.md`.

src/content/tools/pub/semver-rules.md
mosuem: Done—updated the public API boundary definition to include any library under `lib/` outside `lib/src/` (including subdirectories like `lib/foo/bar.dart`).

src/content/tools/pub/semver-rules.md
mosuem: Done—explicitly mentioned `@internal` and `@visibleForTesting` (`package:meta`) under "Private and internal".

src/content/tools/pub/semver-rules.md
mosuem: Addressed in the new "Guiding principles: Contracts and intended usage" section at the top, which notes that local variable type inference (`var x = fn();`) and similar edge cases are treated as `MINOR` by Dart ecosystem convention.

src/content/tools/pub/semver-rules.md
mosuem: Done—updated the rule to cover changing `const` to `final` as `MAJOR` (breaking `const` contexts).

src/content/tools/pub/semver-rules.md
mosuem: Done—updated `Add the base modifier` to note that unmarked subclasses (`extends`) also break unless marked `base`, `final`, or `sealed`.

src/content/tools/pub/semver-rules.md
mosuem: Done—noted that `sealed` implicitly makes the class `abstract` and prevents direct construction.

src/content/tools/pub/semver-rules.md
mosuem: Done—added `MAJOR: Remove the sealed modifier from a class`, noted that removing `final`/`base` means future member additions become `MAJOR`, and noted that adding `mixin` (`class` -> `mixin class`) is `MINOR`.

src/content/tools/pub/semver-rules.md
mosuem: Done—added the "Guiding principles: Contracts and intended usage" section upfront and updated this rule to clarify that if a class documents that it must not be implemented externally, adding concrete members is non-breaking.

src/content/tools/pub/semver-rules.md
mosuem: Covered under the new upfront "Guiding principles: Contracts and intended usage" section regarding downstream name collisions and shadowing.

src/content/tools/pub/semver-rules.md
mosuem: Done—scoped the heading to `MAJOR: Add an instance member to an interface class` and noted that adding `static` members is `MINOR`.

src/content/tools/pub/semver-rules.md
mosuem: Done—updated the example to use `abstract interface class Repository`.

src/content/tools/pub/semver-rules.md
mosuem: Done—expanded the note on `Add a concrete member to a mixin` to mention superclass member signature conflicts.

src/content/tools/pub/semver-rules.md
mosuem: Done—aligned the prose with `on State` to match the example.

src/content/tools/pub/semver-rules.md
mosuem: Done—added a note explaining that adding a named constructor to a class that only had an implicit default constructor removes `MyClass()` (`MAJOR` unless `MyClass()` is explicitly kept), and that changing a generative constructor to `factory` on an extendable class is `MAJOR`.

src/content/tools/pub/semver-rules.md
mosuem: Done—split `Add an optional parameter` into `MINOR` (top-level functions, constructors, `static` methods, and `final`/`sealed` classes) vs. `MAJOR` (instance methods on extendable or implementable classes).

src/content/tools/pub/semver-rules.md
mosuem: Done—noted under `Rename a named parameter` that renaming a positional parameter is non-breaking.

src/content/tools/pub/semver-rules.md
mosuem: Done—updated `Change the default value of an optional parameter` to distinguish between contract-preserving tweaks (`MINOR` or `PATCH`) and changes that alter the observable behavioral contract (`MAJOR`).

src/content/tools/pub/semver-rules.md
mosuem: Addressed by the general "Guiding principles: Contracts and intended usage" section covering local variable type inference (`var x = fn();`).

src/content/tools/pub/semver-rules.md
mosuem: Done—renamed to `MAJOR: Add or remove a type parameter`, explained instantiation-to-bounds vs. explicit type arguments, and added a code example.

src/content/tools/pub/semver-rules.md
mosuem: Done—updated to `MAJOR: Remove or rename an extension or extension member` and noted explicit extension invocations and `show`/`hide` combinators.

src/content/tools/pub/semver-rules.md
mosuem: Done—changed the phrasing from "In SemVer..." to "By convention...".

src/content/tools/pub/semver-rules.md
mosuem: Done—updated to `MAJOR: Change or rename the underlying representation field` and explained that the representation field name defines a public getter.

src/content/tools/pub/semver-rules.md
mosuem: Done—updated the explanation to state that extension type members are resolved statically and allow implementing extension types (`implements Id`) to redeclare members without virtual override conflicts.

src/content/tools/pub/semver-rules.md
mosuem: Done—updated `Throw an exception on previously succeeding inputs or states` to cover undocumented exceptions as well as switching between synchronous throws and failed `Future`s.

src/content/tools/pub/semver-rules.md
mosuem: Done—added a note explaining that bumping the minimum SDK constraint across a major language version (e.g., `<3.0.0` to `>=3.0.0`) changes the package's default language version and can alter class semantics such as implicit mixins.

src/content/tools/pub/semver-rules.md
mosuem: Done—added a warning under `Export a new dependency` that re-exporting a dependency (or exposing its types in public signatures) means upgrading to a new major version of that dependency requires a `MAJOR` bump for your package.

src/content/tools/pub/semver-rules.md
mosuem: Done—clarified the heading and description as `PATCH: Raise minimum dependency constraints or update internal dependencies`.

PR #7435, opened 2026-08-10
Update changelog for Dart 3.13

conooi:
Update changelog for Dart 3.13


parlough (approved):
Thanks @conooi!

I pushed a commit to fix the build and with some other small adjustments, otherwise looks good to me.

PR #7443, opened 2026-08-12
[blog] Use static cover image for Dart 3.13 release post

conooi:
Add static social preview image for Dart 3.13 blog post

- Extract static first frame from dart3-13.gif to dart3-13-cover.webp
- Update blog frontmatter to use static cover image for social cards
- Update publish-blog skill with static image workflow for GIFs


.agents/skills/publish-blog/SKILL.md
parlough, line 1: Thanks for updating this and providing it with instructions to handle both possibilities!

parlough (approved):
Thanks @conooi! Looks good to me.

Note, as discussed, I pushed a commit that adds support for an additional, optional `socialImage` frontmatter field and updated the skill to account for it.

PR #7471, opened 2026-08-26
Add Package skills docs

conooi:
Add and updates docs for package skills


src/content/tools/package-skills.md
```diff
@@ -0,0 +1,156 @@
...
+
+`--version`
+: Print the tool version and exit.
+
+## Commands
+
+### `skills get`
+
+Installs or updates skills provided by dependencies in your `pubspec.yaml`.
+Running `dart run skills@` without a command
```
jakemac53: This isn't accurate (you will just get the usage text).

src/content/tools/package-skills.md
```diff
@@ -0,0 +1,156 @@
...
+discovers, installs, and manages AI agent skills
+for Dart and Flutter packages.
+
+You can execute `package:skills` directly using `dart run`:
+
+```console
+$ dart run skills@ get
+```
+
+## Global options
```
jakemac53: I would at least move this section down below, but maybe just remove it. Most people probably don't care about these (other than --help).

src/content/tools/package-skills.md
```diff
@@ -0,0 +1,156 @@
+---
```
jakemac53, line 1: This page reads basically like a "man page" which isn't necessarily bad, but it isn't how the other pages read next to it (like the build_runner/webdev pages). Those are more like the README content, describing more what the package actually is and how to use it.

More than anything, its just the inconsistency between these pages that I noticed though.

src/content/ai/package-skills.md
```diff
@@ -0,0 +1,220 @@
...
+* Hallucinate APIs that don't exist in the specific version you're using.
+
+When you install package skills,
+your AI agent gains immediate, authoritative knowledge
+written and maintained by the package authors themselves.
+
+## Install skills for dependencies
+
+To install and update skills provided by your project's dependencies,
+use `dart run skills@ get` (or run `dart run skills@` directly):
```
jakemac53: The `get` command is required (same comment elsewhere)

src/content/ai/package-skills.md
```diff
@@ -0,0 +1,220 @@
...
+```console
+Scanning dependencies for available skills...
+? Select skills to install:
+  [x] shelf: routing - Define and handle HTTP routes
+  [ ] shelf: middleware - Add logging and CORS middleware
+  [x] dio: error-handling - Configure retry interceptors
+```
+
+Use the arrow keys and spacebar to select the skills you want to install,
+then press <kbd>Enter</kbd>.
```
jakemac53, line 79: It might be worth noting you can also use Ctrl+A to select/deselect all

src/content/ai/package-skills.md
```diff
@@ -0,0 +1,220 @@
...
+without interactive prompts, pass the `-a` (or `--all`) flag:
+
+```console
+$ dart run skills@ get --all
+```
+
+## Install skills from a Git repository
+
+You can also install skills directly from any Git repository,
+even if the repository is not published on pub.dev,
```
jakemac53: Nit, repositories are not published to pub.dev, maybe just remove that part:

```suggestion
You can also install skills directly from any Git repository
```

src/content/ai/package-skills.md
```diff
@@ -0,0 +1,220 @@
...
+```
+
+Once added as a source,
+you can update skills from your configured Git repositories at any time
+by running `dart run skills@ get`.
+
+## Skill directory structure
+
+When you install skills,
+they are copied into your project's `.agents/skills/` directory:
```
jakemac53: Technically they are installed wherever appropriate for the agent that you select.

src/content/ai/package-skills.md
```diff
@@ -0,0 +1,220 @@
...
+```
+
+Popular AI coding agents automatically scan `.agents/skills/`
+in your workspace.
+Once installed, the skills are immediately available
+to your agent during code generation, refactoring, and chat.
+
+### Version control recommendations
+
+We recommend committing the `.agents/skills/` directory to source control.
```
jakemac53: I don't know that we have actually had a discussion about what our recommendations are here.

It is actually pretty nuanced and I don't think there is any obvious one size fits all solution.... but if you do check in your skills you should also check in your `.config/dart-skills` dir.

If you do not check in your skills,  you probably should not check in that dir.

src/content/tools/pub/package-skills.md
```diff
@@ -0,0 +1,235 @@
...
+description: A powerful package for Dart developers.
+version: 1.0.0
+
+topics:
+  - skills
+  - server
+```
+
+Packages tagged with `skills` are indexed and discoverable in the
+[skills topic on pub.dev][pub-topics].
```
jakemac53: cc @jonasfj @sigurdm  wdyt about using skills tags like this?

These packages aren't actually about skills, they just include some.

I think maybe if a package _only_ contains skills that might make more sense, but I am not sure about including the tag for any package that contains skills as it will make it harder to find packages on pub that are actually primarily about skills.

conooi: I will remove this for now while the team discusses about it

sigurdm: We could have a tag `#has-skills` or something.
But perhaps this is better to auto-detect in pana, rather than relying on everyone getting it right.
Pana can add a tag `has:skills` or something, which will get indexed for search.

What is the right way of detecting if a package contains skills?
@szakarias might also have insights

jakemac53: You can just check for a non-empty top level `skills/` dir in the package.

You could also potentially use an API from package:skills (we could add one to expose whatever exact functionality you wanted?). That would be more forwards compatible potentially.

jakemac53 (approved):

jesskuras (approved):

PR #7486, opened 2026-08-31
Highlight parameter modifier restriction in primary constructors doc

### LINKED ISSUE #7475: Make parameter modifier breaking change ('final' / 'var') more prominent in primary constructors doc
### Page URL

https://dart.dev/language/primary-constructors

### Page source

_No response_

### Summary
In Dart 3.13, primary constructors reserve `final` and `var` exclusively for declaring parameters, which makes using `final` on ordinary parameters (in functions, methods, and in-body constructors) a compile-time error (`extraneous_modifier`). 
While this is noted at the bottom of the [Primary constructors](https://dart.dev/language/primary-constructors#constraints-and-breaking-changes) page, we should make this breaking change and its migration path much more prominent for developers landing on the page.
### Proposed Changes to `primary-constructors.md`
1. **Add a prominent callout earlier on the page:**
   - Place a visible note/alert earlier (such as under *Field declarations in parameters* or *Overview*) highlighting that `final` and `var` on formal parameters in non-primary constructors and regular functions/methods are now disallowed.
2. **Add a clear code snippet illustrating the error:**
   - Show a direct before/after or valid vs. invalid comparison so readers immediately recognize why their existing code broke:
     ```dart
     // Valid: Primary constructor declaring parameter
     class Point(final int x, final int y);
     // Compile-time error in Dart 3.13+ (extraneous_modifier):
     void printValue(final int x) => print(x); 
### Context / References
- Reported SDK issue: [dart-lang/sdk#64151](https://github.com/dart-lang/sdk/issues/64151)
- Existing warning section: [Constraints and breaking changes](https://dart.dev/language/primary-constructors#constraints-and-breaking-changes)

### Expected fix

_No response_

### Additional context

_No response_

conooi:
Add a prominent warning callout and code excerpt under 'Field declarations in parameters' explaining that \`final\` and \`var\` are reserved exclusively for declaring parameters in Dart 3.13+.
Show a valid vs. invalid comparison highlighting the \`extraneous_modifier\` error on ordinary function parameters, and provide migration guidance pointing to \`dart fix\` and the \`parameter_assignments\` lint rule.

Fixes #7475

### DIFF
--- a/src/content/language/functions.md
+++ b/src/content/language/functions.md
@@ -220,7 +220,7 @@ a compile-time error (`extraneous_modifier`).
 For code examples and migration steps, see
 [Parameter modifier restrictions][].
 
-To enforce immutable parameters as a style choice,
+To enforce non-reassignable parameters as a style choice,
 use the [`parameter_assignments`][] lint rule instead.
 
 ## The main() function {:#main}
--- a/src/content/language/functions.md
+++ b/src/content/language/functions.md
@@ -213,14 +213,15 @@ assert(say('Bob', 'Howdy') == 'Bob says Howdy with a carrier pigeon');
 
 ### Parameter modifiers
 
-In Dart 3.13 and later,
-you can't use modifiers like `final` or `var` for normal function parameters.
-These keywords are now reserved exclusively for
+The `final` and `var` parameter modifiers are reserved exclusively for
 [primary constructors][] to declare instance fields.
+Using them on formal parameters in any other declaration produces
+a compile-time error (`extraneous_modifier`).
+For code examples and migration steps, see
+[Parameter modifier restrictions][].
 
-If you wish to enforce immutability for function parameters,
-use lints like [`parameter_assignments`][] instead
-of the `final` keyword in the signature.
+To enforce immutable parameters as a style choice,
+use the [`parameter_assignments`][] lint rule instead.
 
 ## The main() function {:#main}
 
@@ -706,5 +707,6 @@ is not `final`) an external setter.
 [trailing commas]: /language/collections#lists
 [primary constructors]: /language/primary-constructors
 [`parameter_assignments`]: /tools/linter-rules/parameter_assignments
+[Parameter modifier restrictions]: /language/primary-constructors#parameter-modifier-restrictions
 
 
--- a/src/content/language/primary-constructors.md
+++ b/src/content/language/primary-constructors.md
@@ -82,9 +82,26 @@ class Point(var int x, var int y);
 class User(String name);
 ```
 
-Because `final` and `var` modifiers on parameters are reserved exclusively
-for declaring parameters in primary constructors,
-you can't use them on parameters in other kinds of functions.
+:::warning Breaking change: Parameter modifiers
+In Dart 3.13 and later, `final` and `var` parameter modifiers are
+reserved exclusively for declaring parameters in primary constructors.
+Using `final` or `var` on formal parameters in any other declaration—including
+top-level functions, methods, closures, and in-body constructors—produces
+a compile-time error (`extraneous_modifier`).
+
+<?code-excerpt "language/lib/primary_constructors/primary_constructors.dart (declaring-parameters-error)" replace="/PointFinal/Point/g; /void printValue\(int x\)/void printValue(final int x)/g"?>
+```dart
+// Valid: Primary constructor declaring parameter.
+class Point(final int x, final int y);
+
+// Compile-time error in Dart 3.13+ (extraneous_modifier):
+void printValue(final int x) => print(x);
+```
+
+To fix this error, remove the modifier or run [`dart fix`][].
+To enforce immutable parameters as a style choice in Dart 3.13 and later,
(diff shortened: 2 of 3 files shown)


src/content/language/primary-constructors.md
```diff
@@ -82,17 +82,34 @@ class Point(var int x, var int y);
...
-you can't use them on parameters in other kinds of functions.
-
 For extension types, the primary constructor must have exactly one parameter.
 This parameter is always a declaring parameter, even if you omit the modifier.
 You can use the `final` modifier, but it's an error to use `var`.
 
 Mixin classes can only have a primary constructor
 with no parameters, body, or initializer list.
 
+:::warning Breaking change: Parameter modifiers
```
kallentu: I wonder if this section is more useful in the methods page? Like in the bug, if someone gets the extraneous modifier error from adding these modifiers to a method and is trying to search it up in the docs, they'd probably head to the Methods page first.

It's hard to know to look in this page if you don't know primary constructors exist haha.

src/content/language/primary-constructors.md
```diff
@@ -363,14 +380,16 @@ when using primary constructors:
 *   **Restriction on `final` and `var` in normal function parameters**:
     With primary constructors,
     using `final` or `var` on formal parameters in normal functions
-    becomes a compile-time error.
+    becomes a compile-time error (`extraneous_modifier`).
     They are reserved exclusively for
     declaring parameters in primary constructors.
+    To migrate your code, remove `final` or `var` from affected parameters,
+    or run [`dart fix`][].
     Note that the lints `avoid_final_parameters` and
```
kallentu: Maybe we can rephrase 388 to 390 somehow? The reason why the lints don't work in 3.13 is because there's a compile-time error instead. Users should be using these lints in 3.12 or lower to fix any instances of a final/var usage in a method.

Perhaps suggest users to use those lints in 3.12 and below to preemptively fix these errors ahead of 3.13?

src/content/language/primary-constructors.md
parlough: Good point. Interestingly, we do have a section like that on the functions page: https://dart.dev/language/functions#parameter-modifiers

@conooi I like this new content but it seems we're duplicating this explanation in a lot of places. Rather than adding more locations, how about the contents of this new explanatory note could be a full subsection under https://dart.dev/language/primary-constructors#constraints-and-breaking-changes?

Then keep a shorter warning note here that links to it. Same for the functions page and wherever else that needs to reference it.

Eventually we can drop the warnings and cross references.

src/content/language/primary-constructors.md
```diff
@@ -354,32 +358,37 @@ when using primary constructors:
...
-    using `final` or `var` on formal parameters in normal functions
-    becomes a compile-time error.
-    They are reserved exclusively for
-    declaring parameters in primary constructors.
-    Note that the lints `avoid_final_parameters` and
-    `var_with_no_type_annotation` only work with
-    a [language version][] of 3.12 or lower.
-    To enforce immutable parameters as a style choice in Dart 3.13 and later,
-    use the [`parameter_assignments`][] linter rule.
-*   **The `factory` method edge case**:
```
kallentu, line 374: Did we mean to remove this? Could we document the `factory` breaking change as well?

conooi: Definitely! I had removed it while restructuring the parameter modifier subsection, but I've added it back.

kallentu (approved):

src/content/language/functions.md
```diff
@@ -213,14 +213,15 @@ assert(say('Bob', 'Howdy') == 'Bob says Howdy with a carrier pigeon');
...
 [primary constructors][] to declare instance fields.
+Using them on formal parameters in any other declaration produces
+a compile-time error (`extraneous_modifier`).
+For code examples and migration steps, see
+[Parameter modifier restrictions][].
 
-If you wish to enforce immutability for function parameters,
-use lints like [`parameter_assignments`][] instead
-of the `final` keyword in the signature.
+To enforce immutable parameters as a style choice,
```
parlough: I know this is pre-existing, but perhaps "non-reassignable" is better here as "immutable" could be misunderstood to mean the value is immutable.

```suggestion
To enforce non-reassignable parameters as a style choice,
```

src/content/language/primary-constructors.md
```diff
@@ -82,17 +82,21 @@ class Point(var int x, var int y);
...
-
 For extension types, the primary constructor must have exactly one parameter.
 This parameter is always a declaring parameter, even if you omit the modifier.
-You can use the `final` modifier, but it's an error to use `var`.
 
 Mixin classes can only have a primary constructor
 with no parameters, body, or initializer list.
 
+:::warning
+**Breaking change: Parameter modifiers**
```
parlough: I personally recommend dropping the "Breaking change" text, as we technically don't consider language versioned changes like these as breaking.

```suggestion
:::warning New parameter modifier restrictions
```

parlough (approved):
Love this consolidation and clean up, thanks @conooi!

PR #7489, opened 2026-09-02
Clarify for in loop explanation

conooi:
Updates the anonymous functions section on `/language/functions` to match the example code:
- Updates the prose to describe passing the anonymous function to `map()` and using a `for-in` loop to print each item, removing the outdated reference to `forEach`. Follows up on #7351.
- Trims the `anon-func` excerpt in `functions_test.dart` to just the arrow function definition, ensuring it matches the instruction ("Paste the following line...") while keeping the example compliant with Effective Dart's `avoid_function_literals_in_foreach_calls`.


sfshaza2 (approved):
lgtm!

PR #7500, opened 2026-09-09
Migrate to managed cookie bar script

### LINKED ISSUE #7492: show cookiebar popup only in EU
On pub.dev we use:
```html
<link rel="stylesheet" href="https://www.gstatic.com/glue/v25_0/ccb.min.css"/>
<script
    data-autoload-cookie-consent-bar="true"
    src="https://www.gstatic.com/brandstudio/kato/cookie_choice_component/cookie_consent_bar.v3.js">
</script>
```

Note: to style this use:
```css
// both body and #cookieBar is needed to give higher CSS specificity that the rules already in the ccb.min.css
body #cookieBar {
  background-color: ...
}
```

This is an internal solution that relies on some internal serving magic I don't understand to serve the JS file differently depending on location.

Note: for googlers, you'll need to use incognito mode in your phone, because corp desktops usually won't trigger this popup.

--------------

There is also a newer variant, but I haven't tested it yet:

```html
  <link rel="preconnect" href="https://fonts.googleapis.com">
  <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>

  <!-- fonts -->
  <link rel="preload" href="https://fonts.googleapis.com/css2?family=Google+Sans+Text:wght@400&family=Google+Sans:wght@500&display=swap" as="style">
  <link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Google+Sans+Text:wght@400&family=Google+Sans:wght@500&display=swap">

  <!-- style -->
  <link href="https://www.gstatic.com/glue/cookienotificationbar/cookienotificationbar.min.css" rel="stylesheet">

  <script src="https://www.gstatic.com/glue/cookienotificationbar/cookienotificationbar.min.js"
          data-glue-cookie-notification-bar-category="2B"
          data-glue-cookie-notification-bar-site-id="dart.dev">
  </script>
```

I think `data-glue-cookie-notification-bar-site-id` defaults to `location.host` so we might not even need that.

A "cookie management controls" button can be injected with:
```
<button aria-hidden="true" class="glue-cookie-notification-bar-control">
  Manage cookies
</button>
```

But afaik only required for 2A sites, not 2B which is afaik what dart.dev is.

parlough:
- Replaces custom cookie banner and handling with managed Google script.
- Adjust custom styles to ensure matches with site and is responsive to our dark mode.

Fixes https://github.com/dart-lang/site-www/issues/7492


site/lib/_sass/components/_cookie-notice.scss
```diff
@@ -1,58 +1,63 @@
-#cookie-notice {
-  display: none;
-  justify-content: center;
+@use '../base/mixins' as mixins;
+
+// Override some of the the managed banner's styles to
+// better align with our site and support dark mode.
```
jonasfj, line 4: I don't know how stable the style and classes are...

Like I don't know if they could change over time.

But I'd they do, it probably worn matter much.

jonasfj (approved):
I checked the preview url in the EU, worked fine for me.

site/lib/_sass/components/_cookie-notice.scss
parlough: Yeah I imagine it'll still be properly functional, just a bit odd looking. The [firebase docs site](https://firebase.google.com/docs) has some similar overrides so I imagine it's common or even expected.

PR #7513, opened 2026-09-14
Move away from committed style_hash.dart file

parlough:
Migrates https://github.com/flutter/website/pull/13852 from `flutter/website`

When building the full, static site, check the built `main.css` for its hash during the normal Jaspr site build rather than relying on a custom builder and a committed Dart file.


conooi (approved):
thanks @parlough!

PR #7522, opened 2026-09-18
Add prebuilt/ directory to package layout conventions

jonasfj:
Document the prebuilt/ directory for pre-built binary libraries intended to be used or referenced by hook/build.dart.


src/content/tools/pub/package-layout.md
```diff
@@ -553,6 +555,18 @@ Because these hooks are invoked by the
...
+
+- enchilada/
+  - prebuilt/
+    - salsa.so
+
+</FileTree>
+
+If your package includes pre-built binary libraries
+intended to be used or referenced by `hook/build.dart`,
+place them in a directory called `prebuilt`.
```
sigurdm, line 568: Do we want to discuss checking binaries in vs not? and something about git not allowing empty directories to be checked in?

sigurdm (approved):

src/content/tools/pub/package-layout.md
```diff
@@ -553,6 +555,18 @@ Because these hooks are invoked by the
...
 
+<FileTree>
+
+- enchilada/
+  - prebuilt/
+    - salsa.so
+
+</FileTree>
+
+If your package includes pre-built binary libraries
```
dcharkes, line 566: I think we should call out that these should be git ignored?

